PR #3035, opened 2015-08-20
Rename helm-match-plugin.el to helm-multi-match.el

thierryvolpiatto:
* recipes/helm-core: Update recipe.


purcell:
Thanks.

thierryvolpiatto:

> Merged #3035.

Great, thanks.

-- 
Thierry
Get my Gnupg key:
gpg --keyserver pgp.mit.edu --recv-keys 59F29997

PR #3038, opened 2015-08-22
get wiki last-modified date directly from http headers

steckerhalter:
Addresses https://github.com/milkypostman/melpa/issues/3033

It could be solved better I guess (with some more refactoring) but I want to do some other things too :)

The good thing about it is that it doesn't need two wiki requests any longer and the whole stamp-file business is also not necessary.

### DIFF
--- a/package-build.el
+++ b/package-build.el
@@ -285,6 +285,7 @@ seconds; the server cuts off after 10 requests in 20 seconds.")
 
 (defun package-build--url-copy-file (url newname &optional ok-if-already-exists)
   "Copy URL to NEWNAME.  Both args must be strings.
+Returns the http request's header as a string.
 Like `url-copy-file', but it produces an error if the http response is not 200.
 Signals a `file-already-exists' error if file NEWNAME already exists,
 unless a third argument OK-IF-ALREADY-EXISTS is supplied and non-nil.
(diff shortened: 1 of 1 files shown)


purcell:
Looks like a big improvement - thanks a lot for this!

milkypostman:
only thing to note is that this b0rked all the wiki packages to have a different date. i don't know why. i'm guessing something to do with timezone maybe? anyways, i'm taking this opportunity to do a clean of the temp checkouts. this should resolve my errors. there should be no issue because any new changes will overwrite even the newer dates. but just an FYI

steckerhalter:
ok. sorry, I should have checked that before making the PR

steckerhalter:
Are the dates higher or lower?

milkypostman:
lower. but it's ok. it doesn't matter really.

steckerhalter:
allright. have a nice day then :)

milkypostman:
thanks for the PR!

steckerhalter:
YW. I had another look at the [code](https://github.com/milkypostman/melpa/blob/master/package-build.el#L167):

```cl
(defun package-build--parse-time (str)
  "Parse STR as a time, and format as a YYYYMMDD.HHMM string."
  ;; We remove zero-padding the HH portion, as it is lost
  ;; when stored in the archive-contents
  (let* ((s (substring-no-properties str))
         (time (date-to-time
                (if (string-match "^\\([0-9]\\{4\\}\\)/\\([0-9]\\{2\\}\\)/\\([0-9]\\{2\\}\\) \\([0-9]\\{2\\}:[0-9]\\{2\\}:[0-9]\\{2\\}\\)$" str)
                    (concat (match-string 1 str) "-" (match-string 2 str) "-"
                            (match-string 3 str) " " (match-string 4 str))
                  str))))
    (concat (format-time-string "%Y%m%d." time)
            (format "%d" (or (string-to-number (format-time-string "%H%M" time)) 0)))))
```

format-time-string doc says:

```
The third, optional, argument UNIVERSAL, if non-nil, means describe TIME
as Universal Time; nil means describe TIME in the local time zone.
```

It looks like we currently get localized time out of this so it would probably be better to make it return universal time.

purcell:
Yes. IIRC there's another issue in which we suggested using UTC throughout, but decided that it would be disruptive to do so, and therefore not desirable to rush into... :-) But it's clearly the right thing to do.

steckerhalter:
I understand what you mean :)
Cheers

milkypostman:
I think we should just do it. People already hate our versioning system.
Why not cause upgrades to al packages?

steckerhalter:
Sounds like a plan :)

If it should turn out nasty I could help you out if possible somehow.

purcell:
> I think we should just do it. People already hate our versioning system. Why not cause upgrades to al packages?

I continue to like your style...

milkypostman:
ok so what happened here is that we actually STOPPED pulling UTC and that is why all the new times were < the old times. see https://github.com/milkypostman/melpa/issues/1661

steckerhalter:
right, I think it is now getting GMT in the date string from the headers and before it was getting UTC which wasn't understood by `date-to-time`

PR #3041, opened 2015-08-22
Add recipe for JDEE

pwojnowski:
It's Java Development Environment for Emacs.
The repository is here: https://github.com/jdee-emacs/jdee
I'm a maintainer of the package.
I've tested the package and it works.

### DIFF
--- a/recipes/jdee
+++ b/recipes/jdee
@@ -1,0 +1,1 @@
+(jdee :fetcher github :repo "jdee-emacs/jdee")


purcell:
Yay, that's great! Happy to see JDEE getting neatly packaged. I'll go ahead and merge this, but `which.el` could benefit from a couple of fixes. Firstly, perhaps the `executable-find` does the right thing on Windows? And secondly, the `windows-suffixes` var should be `which-windows-suffixes`. :-)

pwojnowski:
I've created issues for these items. Thanks! :-)

PR #3042, opened 2015-08-23
Add recipe for helm-ctest

danlamanna:
This package allows for selecting individual tests to run through ctest using the helm interface.

Here is the repository (I am the maintainer): https://github.com/danlamanna/helm-ctest


purcell:
Nice. Thanks.

PR #3044, opened 2015-08-23
add recipe for dynamic-ruler

rocher:
This package displays a dynamic ruler at point that can be freely moved around the buffer, for measuring and positioning text.

See nice animations at http://rocher.github.io/dynamic-ruler/


purcell:
Looks good! Quick feedback:

- You should prefix `temporary-invisible-change` with the package name, and I'm pretty certain it doesn't need be autoloaded.
- Please remove the `:files` info from the recipe (it's not needed), and force-push a single commit back to this branch.

Hope that helps. Looks good, otherwise. :-)

-Steve


P.S. The link from that page to the .el file is broken: it goes to https://raw.gitusercontent.com/rocher/dynamic-ruler/masterldyanmic-ruler.el.

rocher:
Thanks for the feedback!!

I've applied the changes, renamed to `dynamic-ruler-temporary-invisible-change` and removed autoload.
Additional, fixed some docs and formatting.

- Francesc

purcell:
Thanks. Locally, though, the code doesn't actually run for me: `M-x dynamic-ruler` results in `Symbol's value as variable is void: x-initialized
reference to free variable `x-initialized'`.

rocher:
Hmmm... I suspect this variable is new in Emacs 24.5. Ok, using `(display-graph-p)` works fine.

But after testing from the repo (I think local copies confused my previous tests), I found that `dynamic-ruler-temporary-invisible-change` must be auto-loaded, otherwise `dynamic-ruler-vertical` doesn't works.
With this, compilation emits a warning, *dynamic-ruler.el:350:1:Warning: the function `dynamic-ruler-momentary-column' might not be defined at runtime.*, but that function is defined at runtime, so I think that's not a serious problem.

Thanks again for your valuable feedback!
-- Francesc

purcell:
Thanks, applied in a single commit in 926c438.



> But after testing from the repo (I think local copies confused my previous tests), I found that dynamic-ruler-temporary-invisible-change must be auto-loaded, otherwise dynamic-ruler-vertical doesn't works.
> With this, compilation emits a warning, dynamic-ruler.el:350:1:Warning: the function `dynamic-ruler-momentary-column' might not be defined at runtime., but that function is defined at runtime, so I think that's not a serious problem.

No, I just think you need to define the macro before defining the function. I think the autoload is just masking that problem for you.

mrkkrp:
This package is still unavailable for some reason.

rocher:
Don't know how long it takes to a packages uploaded to melpa with 'git
commit' to appear in Emacs 'M-x list-packages' ...


-- *Francesc Rocher*

purcell:
It's there. Packages usually get built within a few hours.

mrkkrp:
@purcell, It's OK, I see it now :-)

PR #3045, opened 2015-08-24
Added org-if recipe

Elzair:
I have added a recipe to install [org-if](https://gitlab.com/elzair/org-if), an interactive-fiction authoring system for org-mode.


purcell:
Thanks for this!

Please remove the `:files` spec: `*.el` is included by default. You should amend the commit and force-push it back here.

Looks good otherwise. :-)

Elzair:
Alright. I have done just that.

purcell:
Thanks, merged. :-)

PR #3047, opened 2015-08-24
Add ac-alchemist

syohex:
ac-alchemist provides auto-complete source of alchemist.

https://github.com/syohex/emacs-ac-alchemist


purcell:
Super, thanks.

syohex:
Thanks always :+1:

PR #3049, opened 2015-08-24
Adding git-record-mode

alakra:
Minor mode that automatically adds keystroke changes on a buffer to a git repository via `magit`.

See https://github.com/alakra/git-record-mode for package details.

NOTE: This differs from `git-auto-commit-mode` by hooking into buffer changes, not just when a buffer is saved.

### DIFF
--- a/recipes/git-record-mode
+++ b/recipes/git-record-mode
@@ -1,0 +1,1 @@
+(git-record-mode :fetcher github :repo "alakra/git-record-mode")


purcell:
Thanks, interesting.

> NOTE: This differs from git-auto-commit-mode by hooking into buffer changes, not just when a buffer is saved.

What's the difference in practice? This still just commits the contents of the buffer's file on disk. Also, sometimes `(buffer-file-name)` will return `nil`, which will break this code.

alakra:
The difference is that each buffer change is timestamped and given a commit. In practice, these changes would just be my inserts/deletes. I know this is esoteric, but it's intended for eventual playback of my keystrokes in realtime in another application.

I will look into the `(buffer-file-name)` issue.

Thanks for taking your time to look at this. :)

alakra:
Upon closer inspection, I remembered that I added `(set-visited-file-name)` that appears to prevent `(buffer-file-name)` from returning nil by forcing the buffer to be associated with an actual file. 

See https://github.com/alakra/git-record-mode/blob/master/git-record-mode.el#L51

alakra:
Hi! Any more opinions or feedback on this? Thanks again for your time.

purcell:
- `(remove-hook 'before-save-hook 'delete-trailing-whitespace)` seems a bit wrong to me -- this will modify the global version of `before-save-hook`, won't it?
- `(set-visited-file-name (buffer-name))` seems unnecessary, and potentially dangerous - you should probably just call `save-buffer`, so that the user will be prompted to choose a filename if the buffer is not already associated with one. You'd then trigger an error if a filename had not been selected.
- What happens if the file has already been added? Is `(magit-call-git "add" (buffer-file-name))` idempotent, or does it raise an error when called twice?
- It seems like there are a number of things a user could do to a buffer which would break this, e.g. writing the buffer to an alternate file.

Overall, while I can see this code might be convenient for you, I really think this isn't going to be widely useful to other people in its current form -- and that's perfectly okay: I've got personal minor modes in my own emacs config which I don't think would make a good package, so I haven't tried to distribute them. So my suggestion in this case would be that we don't add it to MELPA.

alakra:
All right.  I won't push the issue then.  Thanks for the feedback!

PR #3050, opened 2015-08-25
new recipe: homebrew-mode

dunn:
- Summary: minor mode for editing Homebrew formulae
- Repository: https://github.com/dunn/homebrew-mode
- Association: I'm the author

Feedback is very welcome, since this is the first mode I've written.

### DIFF
--- a/recipes/homebrew-mode
+++ b/recipes/homebrew-mode
@@ -1,0 +1,1 @@
+(homebrew-mode :repo "dunn/homebrew-mode" :fetcher github)


purcell:
Thanks. Quick feedback:

- `(kbd "C-c h")` -- this keystroke is [reserved for user customisations](https://www.gnu.org/software/emacs/manual/html_node/elisp/Key-Binding-Conventions.html).
- The `homebrew-mode-default-hooks` stuff is not idiomatic, and seems unnecessary. Put the `homebrew-mode-hook` code directly in the `define-minor-mode` block. Define a global mode additionally, and do the `(homebrew--formula-file-p (current-buffer))` check there. Then just tell users to enable the global mode.

Hope that helps,

-Steve

dunn:
Oh, didn't know that about `C-c`. I got the idea from flycheck. Magit does it too, smh.

The obvious alternative prefix is `C-x h`.  Is there an easy way to check if that binding is already used by other minor modes that might be active at the same time?

purcell:
Well, note that `C-c` is acceptable, but not followed by a letter. `C-x h` is used for the help system by default. How about `C-c C-h`?

dunn:
Oh! I see, thanks.

> `C-x h` is used for the help system by default.

That's odd; it's unbound for me. `C-h` is bound correctly as a help prefix, though.

In any case `C-c C-h` sounds fine. Thanks again for all your help!

purcell:
> That's odd; it's unbound for me. C-h is bound correctly as a help prefix, though.

Sorry, you're right -- `C-x h` is `mark-whole-buffer`, another handy keystroke.

Anyhow, looks great - merging now.

dunn:
Forgot I remapped `mark-whole-buffer` :see_no_evil: 

Thank you!

PR #3051, opened 2015-08-27
Add recipe kroman

cheunghy:
Kroman is a package that romanize Korean hangul.

### DIFF
--- a/recipes/kroman
+++ b/recipes/kroman
@@ -1,0 +1,1 @@
+(kroman :repo "cheunghy/kroman" :fetcher github)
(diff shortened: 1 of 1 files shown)


purcell:
I don't see any elisp in that repo...

syohex:
I guess the repository is https://github.com/cheunghy/kroman-el

purcell:
Ah, right. In that case, quick feedback:

- Needs to declare a dependency on `(emacs "24")` because it declares `lexical-binding: t`
- `(setq debug-on-message t)` doesn't belong in there at all

cheunghy:
Thanks.
I've fixed repo name in the recipe.
I removed ```lexical-binding: t```.
It should work for older emacs.

purcell:
Thanks - added as a single commit in 07ff16d.

PR #3052, opened 2015-08-27
Add recipe for keyword-search

akuroiwa:
* Summary: Browser style keyword search for Emacs based on [browse-apropos-url](http://www.emacswiki.org/emacs/BrowseAproposURL).
* Repository: https://github.com/keyword-search/keyword-search
* Association: I'm a provisional maintainer.

### DIFF
--- a/recipes/keyword-search
+++ b/recipes/keyword-search
@@ -1,0 +1,1 @@
+(keyword-search :repo "keyword-search/keyword-search" :fetcher github)


purcell:
Thanks. Quick feedback:

- You should escape the search strings before inserting them into the URL with `format`
- Consider using symbols rather than strings for the search engine names in the `alist`.

Otherwise, looks good. :)

akuroiwa:
Thank you for your advice.
I tried the second suggestion. It works fine.
Could you verify these codes?

```lisp
(completing-read
	  (format "Keyword search at point (default %s): " keyword-search-default)
	  keyword-search-alist nil t nil nil (symbol-name keyword-search-default))
```

```lisp
(url (cdr (assoc (intern-soft key) keyword-search-alist)))
```

I have a question about the first suggestion.
Do you mean to replace white space character with plus character or to escape white space character?
Do you have any idea about `(replace-regexp-in-string)`?

purcell:
> Could you verify these codes?

Looks okay to me, but I haven't tested it.

> I have a question about the first suggestion.
> Do you mean to replace white space character with plus character or to escape white space character?

Yes, use something like `(format search-url (url-hexify-string search-term))` so that reserved characters like `&` and `?` in the search term don't break the URL.

akuroiwa:
Thank you very much.
You are just like a good doctor.
Please see [the latest commit](https://github.com/keyword-search/keyword-search/commit/87274dd29835f720d8cae06256a39a650e396d2a).

I cannot decide whether I should append the following code.
What do you think of it?

```lisp
(url-hexify-string
       (replace-regexp-in-string
	"\\(\s-\\|\t\\|\n\\)+" "\s"
	(buffer-substring (region-beginning) (region-end))))
```

purcell:
> I cannot decide whether I should append the following code.
> What do you think of it?

No, I don't think that's necessary. But I think you're calling `url-hexify-string` in an unusual place: I would suggest that you just call it when you are building the URL, so instead of:

```
(browse-url (format url thing) new-window))
```

you would have:

```
(browse-url (format url (url-hexify-string thing)) new-window))
```

It makes sense to keep the logic of building the URL all together.

akuroiwa:
I have just modified my mistake. Thank you very much.

purcell:
Great, merged - thanks!

akuroiwa:
We feel honored that `keyword-search` recipe was merged into Melpa.

purcell:
Thank you for taking the time to submit this.

PR #3053, opened 2015-08-27
add speech-tagger recipe

cosmicexplorer:
speech-tagger is an extension to tag parts of speech in an emacs buffer
using stanford coreNLP. It can be found at
https://github.com/cosmicexplorer/speech-tagger. I am the maintainer of
this package.

### DIFF
--- a/recipes/speech-tagger
+++ b/recipes/speech-tagger
@@ -1,0 +1,3 @@
+(speech-tagger :fetcher github :repo "cosmicexplorer/speech-tagger"
+               :files ("emacs/speech-tagger.el"
+                       "emacs/penn_treebank_tags.json"))
--- a/recipes/speech-tagger
+++ b/recipes/speech-tagger
@@ -1,5 +1,3 @@
 (speech-tagger :fetcher github :repo "cosmicexplorer/speech-tagger"
                :files ("emacs/speech-tagger.el"
-                       "emacs/penn_treebank_tags.json"
-                       "emacs/speech-tagger.jar")
-               :branch "melpa-build")
+                       "emacs/penn_treebank_tags.json"))
--- a/recipes/speech-tagger
+++ b/recipes/speech-tagger
@@ -1,3 +1,5 @@
 (speech-tagger :fetcher github :repo "cosmicexplorer/speech-tagger"
                :files ("emacs/speech-tagger.el"
-                       "emacs/penn_treebank_tags.json"))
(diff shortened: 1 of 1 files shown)


purcell:
Thanks, looks good. I'd strongly encourage you to follow the [elisp coding conventions](https://www.gnu.org/s/emacs/manual/html_node/elisp/Coding-Conventions.html) for symbol naming, though, because the clojure-style naming here is not at all idiomatic.

Additionally, triple-semicolon comments have a specific meaning, so only use them for standard section headings, e.g. `Commentary:`, `Code:`, `Usage:`, plus the first and last comment lines.

Hope that helps,

-Steve

cosmicexplorer:
Fixed the clojure naming and kept triple-semicolon lines to bookends only. Thanks for the link to those conventions.

I've also made a change (which you can see in this pull request) to the packaging; instead of downloading the jar at runtime, it pulls it in as part of the package, which hopefully takes less people by surprise.

cosmicexplorer:
Just fixed an issue where the "loading" overlay wasn't being removed by the clear-tags function, and made the jar path a `defcustom` to facilitate easier development.

purcell:
Sorry for the slow feedback. The `.el` file looks super now, but we'd strongly prefer to build from your dev branch, not a build branch that is manually updated and may one day get stale without anybody noticing. So to that extent, it would be preferable for the code to come from `master`, and for the package to download the jar at runtime (or, probably, load time).

cosmicexplorer:
Alright, that's what I had before. I'm not familiar with `eval-when` but I'll make it check the jar path for existence and download if not. That does make versioning the jar a tad more difficult, I'll try to see if I can quickly add a hashing scheme to detect versioning without too much hackery.

cosmicexplorer:
Ok, it's proving harder than I thought to download a file over http; `url-retrieve-synchronously` is either dropping bytes, or emacs is converting the coding system incorrectly or something, but I'm still working on this. I could make every speech tag command shoot a network request to my own server, and then I wouldn't have to deal with this at all, but the latency would be super annoying. I'll comment again when I can make this work.

cosmicexplorer:
Ok, turned out the coding system was the problem; specifying `'raw-text` as the coding system in the buffer returned by `url-retrieve-synchronously` seems to have resolved the hashing issues. I've run this through a few tests on an empty emacs and it *seems* to work just fine.

purcell:
Cool, thanks - merged. :-)

PR #3059, opened 2015-08-30
Add recipe for rustfmt

fbergroth:
[rustfmt.el](https://github.com/fbergroth/emacs-rustfmt) formats code in rust buffers using [rustfmt](https://github.com/nrc/rustfmt).

Cheers

### DIFF
--- a/recipes/rustfmt
+++ b/recipes/rustfmt
@@ -1,0 +1,2 @@
+(rustfmt :fetcher github
+         :repo "fbergroth/emacs-rustfmt")


purcell:
Thanks. Quick feedback:

- If the mode needs `lexical-binding`, you should declare a `Package-Requires` declaration listing `(emacs "24")`. (Note that `flycheck-package` would have warned you about this if you were using it.)
- Using `grep-mode` seems a bit hacky. `compilation-minor-mode` might work the same, and would probably be a bit cleaner if it does.

fbergroth:
Thanks for the feedback! I was actually using `flycheck-package`, but since I didn't have a `Package-Requires` or a `Package-Version` header, it was silent.

The issues you mentioned should be fixed now, I chose to not use any mode for the output buffer.

purcell:
> Thanks for the feedback! I was actually using flycheck-package, but since I didn't have a Package-Requires or a Package-Version header, it was silent.

Ah, that figures. I should make it enable itself a little more aggressively, perhaps.

> The issues you mentioned should be fixed now, I chose to not use any mode for the output buffer.

:+1: 

Merging now.

PR #3061, opened 2015-08-31
Add recipe for what-the-commit

danielbarbarito:
[what-the-commit.el](https://github.com/danielbarbarito/what-the-commit.el) fetches a commit from [whatthecommit.com](http://www.whatthecommit.com) and adds it to your kill ring.

### DIFF
--- a/recipes/what-the-commit
+++ b/recipes/what-the-commit
@@ -1,0 +1,2 @@
+(what-the-commit :fetcher github
+         :repo "danielbarbarito/what-the-commit.el")


purcell:
It'd be worth breaking up the 310+ character line for legibility, and reindenting everything consistently. :-)  Also, the docstring for the command would be a great place to mention that it copies the resulting message to the kill ring.

Hope that helps,

-Steve

danielbarbarito:
@purcell thanks for the tips, I'm extremely new at this. Will my pull request not approved until those changes are made or can I just do them after it's accepted?

purcell:
I'd suggest just fixing them up first, so that when people see the new package pop up and check it out, they are more likely to like what they see, and then use it. :-)  I therefore tend to hold off on merging recipes pending code quality changes like this.

danielbarbarito:
@purcell makes sense to me! I cleaned up the code and added that info to the docstring. I'm not 100% sure if its indented properly.

purcell:
Cool - it still wasn't indented correctly. `C-x h` then `TAB` will usually fix that. :-)  But I sent you some changes in https://github.com/danielbarbarito/what-the-commit.el/pull/2 anyway.

danielbarbarito:
@purcell thanks for pointing out the `C-x h` trick, thats very useful. Once again, I'm really new at this so any pointer you give me is going to help immensely. Merged your changes.

purcell:
There's also `C-M-\` (`indent-region`) which will generally reindent the function you're working on. :)

PR #3062, opened 2015-08-31
Add recipe for `desktop+`

ffevotte:
[`desktop+`](https://github.com/ffevotte/desktop-plus)  extends `desktop` by providing more features related to sessions persistance:

1. Instead of relying on Emacs' starting directory to choose the session Emacs restarts, sessions are manipulated by name. All information about them is stored into a centralized directory.

2. Desktop sessions by default save only buffers associated to "real" files. Desktop+ extends this by handling also "special buffers", such as those in `compilation-mode` or `term-mode`.

### DIFF
--- a/recipes/desktop+
+++ b/recipes/desktop+
@@ -1,0 +1,1 @@
+(desktop+ :fetcher github :repo "ffevotte/desktop-plus")


purcell:
Thanks. Quick feedback:

- "improved sessions" isn't a very informative package description
- Please declare dependencies on the latest stable versions of `f` and `dash`, not `"0"`
- There doesn't seem to be a good reason to use the `desktop-` prefix for any of the symbols: `desktop+-` would be more in line with the [elisp conventions](https://www.gnu.org/s/emacs/manual/html_node/elisp/Coding-Conventions.html) and will avoid any possible name clashes.
- When did `nadvice` get added? If it was later than Emacs 24.1, you should adjust your `emacs` dependency in `Package-Requires`

Hope that helps,

-Steve

ffevotte:
Thanks for your feedback. I just pushed a new version of desktop+ fixing these issues:

- the prefix is now consistently `desktop+-` (and is handled by [`names`](/Malabarba/names))
- the package header is fixed:
  - better description
  - correct version numbers for dependencies

And you're right: the package is now in a much better shape. Thanks again (and please don't hesitate if you've further comments to make)

purcell:
Yay, nice!

ffevotte:
Thanks!

PR #3063, opened 2015-08-31
Add nand2tetris recipes

CestDiego:
I had to modularized my packages, so I had to make some changes to the
recipes.c

### DIFF
--- a/recipes/nand2tetris
+++ b/recipes/nand2tetris
@@ -1,3 +1,3 @@
 (nand2tetris :fetcher github
              :repo "CestDiego/nand2tetris.el"
-             :files (:defaults "snippets"))
+             :files ("nand2tetris.el" "nand2tetris-core.el" "snippets"))
--- a/recipes/company-nand2tetris
+++ b/recipes/company-nand2tetris
@@ -1,0 +1,3 @@
+(company-nand2tetris :fetcher github
+                     :repo "CestDiego/nand2tetris.el"
+                     :files ("company-nand2tetris.el"))
--- a/recipes/nand2tetris-assembler
+++ b/recipes/nand2tetris-assembler
@@ -1,0 +1,3 @@
+(nand2tetris-assembler :fetcher github
+                       :repo "CestDiego/nand2tetris.el"
+                       :files ("nand2tetris-assembler.el"))


recipes/nand2tetris
```diff
@@ -1,3 +1,5 @@
 (nand2tetris :fetcher github
              :repo "CestDiego/nand2tetris.el"
-             :files (:defaults "snippets"))
+             :files ("nand2tetris.el"
+                     ("snippets"
+                      ("nand2tetris-mode" "snippets/nand2tetris-mode/*"))))
```
purcell: Just use `"snippets"` here to coy the whole subdir recursively.

recipes/nand2tetris-core
```diff
@@ -0,0 +1,3 @@
+(nand2tetris-core :fetcher github
+                  :repo "CestDiego/nand2tetris.el"
+                  :files ("nand2tetris-core.el"))
```
purcell: Would anybody use `nand2tetris-core` without `nand2tetris` or `nand2tetris-assembler`? This packaging feels too fine-grained to me, with *perhaps* the exception of `company-nand2tetris`.

CestDiego: no, but how else do I  require the package? should I include core with nand2tetris...and then in company-nand2tetris I should require nand2tetris? (which requires core)

purcell: You don't need package requirements between files in the same package. So put all the related files in one package, and then they can just `require` each other. The package metadata would go in the "main" `.el` file - the one with the name matching the package name. Hope that helps.

CestDiego: I modified it so the core is included in the company then :D

purcell:
In company-nand2tetris.el, this is invalid because you need to use proper version numbers:
```
;; Package-Requires: ((names "Lookup.Latest.Version") (nand2tetris "0.0.1") (company "0.5") (cl-lib "0.5.0"))
```

Same in nand2tetris.el:
```
;; Package-Requires: ((names "Lookup.Latest.Version") (yasnippet "0.8.0"))
```

and nand2tetris-assembler.el:
```
;; Package-Requires: ((names "Lookup.Latest.Version") (nand2tetris "0.0.1"))
```

Don't depend on `yasnippet` because it's not core to your package, and not everybody wants to have it installed. Instead, just soft-require it: `(eval-when-compile (require 'yasnippet nil t))`.

purcell:
You can test all these recipes locally using `C-c C-c` -- this will let you check they all install cleanly. :-)

milkypostman:
ping @CestDiego

CestDiego:
@purcell @milkypostman I'm sorry for my absence, well it turns out that I thought "Lookup.Latest.Version" was actually not a function that Melpa had that would make it search for latest version of the package and then install it :P

purcell:
:+1:

CestDiego:
Thank you for your help :) I will take into consideration compilation errors in the next packages :D

PR #3064, opened 2015-08-31
Add rjs recipe.

joeheyming:


purcell:
Quick feedback:

- A slightly longer name would be preferable to avoid confusion. `rjs` is also a file type in Rails.
- Have you seen `requirejs-mode`? Is there any overlap? Could this functionality be submitted as an improvement there?
- Please depend on the stable versions of other packages, not snapshot versions. Look on stable.melpa.org to find them.
- All the symbols you define should start with the package prefix, so those `js2-*` symbols are invalid. See the [elisp conventions](https://www.gnu.org/s/emacs/manual/html_node/elisp/Coding-Conventions.html) for more info.
- `M-x checkdoc` will help you get your docstrings formatted conventionally.
- You're missing the required `(provide ...)` form at the end of the file, and the mandatory `;;; rjs.el ends here` trailing comment line.

FWIW, `flycheck-package` would provide some of this feedback directly inside Emacs if you were to install it.

Hope that helps!

-Steve

joeheyming:
@purcell I'm not sure if this module needs to be a full on emacs mode.  I did just see the requirejs-mode module.  Also, I'm not convinced that our approaches can be merged since mine is more on-demand and his tries to get a list of all current files.  Also, I am providing more customization around sorting the define block and other features like text! style imports.

Any ideas on a different name?  requirejs-bindings? requirejs-manager? requirejs-nav?

purcell:
Perhaps `requirejs-manager`, `requirejs-helpers` or even just `requirejs`?

joeheyming:
I like requirejs too, I just didn't want to be greedy :-)

purcell:
Just go for it. :-)

joeheyming:
Ok, I renamed the repo to requirejs-emacs, and made the requirejs recipe.  Thanks for all your feedback.

joeheyming:
going to resubmit this

PR #3066, opened 2015-08-31
Add recipe for company-ansible

krzysztof-magosa:


purcell:
Don't depend on version `1.0` of `cl-lib`, because that version is not installable separately from Emacs. Since you are already requiring a recent Emacs version, I suggest dropping the `cl-lib` dependency declaration.

purcell:
Also, please put the `Keywords`/`Author` etc between the `Copyright` and the license info in order to fit with conventions. (Enabling `auto-insert-mode` will help you to create the correct boilerplate in future.)

krzysztof-magosa:
Thank you for your feedback.
Issues fixed, please review again.

Cheers.

purcell:
Thanks! Merged.

PR #3068, opened 2015-09-01
Added (and tested) recipe for `subemacs.el'

kbauer:
`subemacs.el` provides functions for evaluating emacs-lisp code in a new `emacs -Q` subprocess, that inherits only the `load-path` of the current session.

This allows conveniently testing commands or compiling files unaffected by user configuration and loaded packages and e.g. helps identifying missing `require` statements.

Repository https://github.com/kbauer/subemacs

The package was created by me.


purcell:
Thanks.

PR #3069, opened 2015-09-01
Add requirejs recipe.

joeheyming:
Renamed rjs to requirejs-emacs.  Hopefully,fixed all the suggestions in the previous thread:

https://github.com/milkypostman/melpa/pull/3064

### DIFF
--- a/recipes/requirejs
+++ b/recipes/requirejs
@@ -1,0 +1,2 @@
+(requirejs :fetcher github :repo "joeheyming/requirejs-emacs")
+


purcell:
Thanks, merged.

You might like to reconsider using `expand-region` here, which is designed as a set of customisable user actions rather than a general-purpose library. Specifically, I believe it's quite easy to implement `(er/mark-outside-pairs)` yourself, e.g. by combining `backward-up-list` and then `mark-sexp`.

joeheyming:
Thanks for the tip.  I agree that I can do it without expand region.

You were right about the overlap with require-mode.  There is also amd-mode.  We all essentially wrote the same thing :-p

joeheyming:
Sweeeet, I didn't know about backward-up-list, it worked like a charm

purcell:
> You were right about the overlap with require-mode. There is also amd-mode. We all essentially wrote the same thing :-p

If one of these packages is clearly the best for end users, it might be worth joining forces, because it's confusing to have multiple overlapping packages.

PR #3071, opened 2015-09-01
Update racer recipe to reflect new project home

phildawes:
We've just moved the racer emacs mode out of the racer repo into it's own project. I hope I've done this right!  -Thanks

### DIFF
--- a/recipes/racer
+++ b/recipes/racer
@@ -1,3 +1,2 @@
-(racer :repo "phildawes/racer"
-       :fetcher github
-       :files ("editors/emacs/racer.el"))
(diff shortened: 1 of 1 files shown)


Wilfred:
Looks good to me! I believe MELPA picks up all *.el files by default, so we don't need to explicitly specify it here.

purcell:
Correct, the `:files` spec should be removed here please. Ideally, amend the commit and force-push it to keep the history clean.

fbergroth:
@phildawes ping :)

phildawes:
Oops sorry, I've updated the patch to remove the :files section

PR #3072, opened 2015-09-01
Add Recipe for evil-texobj-anyblock.el

noctuid:
[evil-textobj-anyblock](https://github.com/noctuid/evil-textobj-anyblock) is a package written by myself that is a port of the vim-textobj-anyblock plugin. It provide a text object that will act on the closest block from a user-defined list that includes blocks such as `()`, `[]`, `<>`, and `""` by default.

### DIFF
--- a/recipes/evil-textobj-anyblock
+++ b/recipes/evil-textobj-anyblock
@@ -1,0 +1,4 @@
+(evil-textobj-anyblock
+  :fetcher github
+  :repo "noctuid/evil-textobj-anyblock")
+


purcell:
Thanks. I think you should arrange to use either `evil-anyobj` or `evil-textobj-anyblock` throughout in order to be consistent. Other than that, looks good.

noctuid:
Thank you. I've made this change now.

purcell:
Thanks!

PR #3075, opened 2015-09-02
New recipe, Python 3 Info Manuals

dvhansen:
Same as the python-info package but for the Python 3 manual.


purcell:
Thanks. Pending https://github.com/dvhansen/python3-info/pull/1

PR #3080, opened 2015-09-03
use master branch for req-package, not develop

edvorg:


purcell:
This misses the point of MELPA: we explicitly want to track the development version of your code. To ship stable packages, apply version tags and corresponding packages will be built on `stable.melpa.org`.

edvorg:
Ok, I got it, thank you.

PR #3082, opened 2015-09-03
use master branch for flycheck-protobuf, not develop

edvorg:


xuchunyang:
There is no develop branch in your repo (https://github.com/edvorg/flycheck-protobuf/branches).

edvorg:
But now it is :)

edvorg:
I found that, melpa uploads my packages from every branch which I push into.
So I need to specify manually which branch to use.

purcell:
Please don't specify the branch. We always use the default branch for your repo, to avoid confusion, so that users who visit your github page will see the same code in the corresponding MELPA package. In the case of this repo, we will use `master`.

PR #3083, opened 2015-09-04
add `ido-exit-target` recipe

waymondo:
Commands and keys for selecting other window and frame targets within ido

### DIFF
--- a/recipes/ido-exit-target
+++ b/recipes/ido-exit-target
@@ -1,0 +1,1 @@
+(ido-exit-target :repo "waymondo/ido-exit-target" :fetcher github)


purcell:
Thanks. I feel like the initialization here is done in a slightly weird way. You've autoloaded the commands, but they'll never really be accessible until the file has been loaded anyway. So those autoloads should go away, and you should probably just provide an autoloaded "ido-exit-target-enable" command which the user can call to install the keybinding in `ido-common-completion-map`.

purcell:
Cool idea, by the way.

waymondo:
Perhaps I could just explain in the README that you can set the `ido-exit-target-keymap-prefix` to `nil` if you want to add your own custom bindings? I removed those autoloads btw.

purcell:
I don't think the `setup` defun is strictly necessary, but without it the autoloads didn't really make sense. So the current state is now probably fine, I think, with the minor downside that it would be fiddly to load the code on-demand.

waymondo:
Ok cool. Thanks for the insight as always.

PR #3087, opened 2015-09-05
use https for github repos

steckerhalter:
Use secure protocol (https) for github

### DIFF
--- a/package-build.el
+++ b/package-build.el
@@ -546,7 +546,7 @@ Return a cons cell whose `car' is the root and whose `cdr' is the repository."
 
 (defun package-build--checkout-github (name config dir)
   "Check package NAME with config CONFIG out of github into DIR."
-  (let* ((url (format "git://github.com/%s.git" (plist-get config :repo))))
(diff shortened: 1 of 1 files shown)


purcell:
I'm sure you've submitted this change for a good reason, but you haven't provided a word of information here about what that is, so it's hard for me to know how to handle this PR. :-)

steckerhalter:
Sorry, I forgot to add the description. Since git:// is insecure it would be better to use https:// as it is the case already for the gitlab fetcher

purcell:
Makes sense, thanks!

milkypostman:
@purcell related to this I just added a rewrite to the server to change git@ to https:// paths. someone else  had mentioned this idea before and a submodule was breaking a recipe because it was using the git@ address.

purcell:
Oh, a rewrite? How does that work?

milkypostman:
It's a fit config thing. You tell it to rewrite any matching git url.

purcell:
Aha! Clever.

PR #3091, opened 2015-09-06
Don't include python sources within anaconda-mode package.

proofit404:
Anaconda mode currently install all its python side dependencies from PyPI.


purcell:
Thanks - much neater!

proofit404:
Glad you like it :smile:

PR #3092, opened 2015-09-06
Add recipe for synonymous

toroidal-code:
This commit adds a recipe for a thesaurus tool. Advantages over current synonym/thesaurus plugins are that it does not require a local thesaurus package (as synosaurus does for English) and uses a web API that provides significantly more information for word lookup and suggestions (see an example [here](http://synonymous.heroku.com/talk)).

### DIFF
--- a/recipes/synonymous
+++ b/recipes/synonymous
@@ -1,0 +1,1 @@
+(synonymous :fetcher github :repo "toroidal-code/synonymous.el")


purcell:
Thanks. Please adjust the naming of your top-level symbols so they follow the [elisp conventions](https://www.gnu.org/software/emacs/manual/html_node/elisp/Coding-Conventions.html), ie. so that they all begin with the prefix `synonymous-`. You should also probably `(require 'cl-lib)` for the sake of clean byte compilation.

toroidal-code:
Alright! Edited and pushed!

purcell:
Awesome, thanks - this looks great.

PR #3093, opened 2015-09-07
Add recipe for ‘vimish-fold’

mrkkrp:
Inspired by this question: https://emacs.stackexchange.com/questions/14781/how-can-i-replicate-vims-code-folding.

It seems that I've solved all the problems with this little package: https://github.com/mrkkrp/vimish-fold


purcell:
Thanks

PR #3095, opened 2015-09-08
Add fstar-mode

cpitclaudel:
F* is a functional, dependently typed programming language from MSR
(https://www.fstar-lang.org/). This is a basic editing mode for F*, with live
verification support (Flycheck) and interactive mode (like Proof-General)
support as well.

In the long run I think I'll move the Flycheck part to a separate package (or
even move the checker definition to Flycheck itself), but for now the
other verification mode (interactive) is a bit too experimental.

Cheers and thanks as always for your hard work,
Clément.

### DIFF
--- a/recipes/fstar-mode
+++ b/recipes/fstar-mode
@@ -1,0 +1,3 @@
+(fstar-mode
+ :fetcher github
+ :repo "FStarLang/fstar-mode.el")


purcell:
Thanks. This mostly looks good, but depending on flycheck is definitely bogus for a major mode. It would be better to provide a separate flycheck-fstar package.

cpitclaudel:
Thanks for the quick reply! 

I've turned the dependency on Flycheck into a soft one; `fstar-mode` will use Flycheck if it is there, but won't complain or fail if it isn't, and won't force-install it. 

At this point, however, the mode really is a lot more useful with Flycheck, so I haven't moved the flycheck code to a separate package. The reason for that is that F* is a verification language, used to write programs but also proofs of programs. Real-time verification is extremely useful in this context, since it lets the verifier check your proofs in the background. This is similar to other languages such as Dafny or Boogie.

In the particular case of F*, however, I've also implemented support for a different verification style more akin to that used for Coq proofs (an interactive, non-automatic mode). Thus I agree that pulling in Flycheck is not always necessary.

Cheers,
Clément

purcell:
:+1:

cpitclaudel:
Awesome, thanks!

PR #3097, opened 2015-09-09
remark-mode: Add missing remark.html file

torgeir:
Be specific about which files to include. This file is needed as it is the html skeleton for the slideshow remark-mode produces.


torgeir:
Sorry for the multiple prs, I'm not that well wandered in melpa packages :)

purcell:
Could you please amend this to `:files (:defaults "remark.html")`? It's a little more general because `:defaults` includes `"*.el"`, and therefore preferred.

torgeir:
Indeed, done!

torgeir:
Anything else missing?

purcell:
Thanks for the reminder - this one fell through the cracks. :-)

torgeir:
:thumbsup:

PR #3098, opened 2015-09-09
Update ac-html 0.4.alpha

cheunghy:
ac-html 0.4 is a breaking new version of ac-html with data provider and language adaptor API.
This version brings back better formatted completion data, too.


purcell:
Thanks! Please use the special value `:defaults` instead of `"*.el"`. You can amend the commit and force-push it here to update the PR cleanly.

cheunghy:
Thanks!
I've updated with the special value `:defaults`.

PR #3099, opened 2015-09-10
add gmail2bbdb recipe

redguardtoo:
Import email and name into bbdb from vcard (a format Gmail supports). works out of the box.


purcell:
Thanks!

PR #3100, opened 2015-09-10
Add recipe for ‘modalka’

mrkkrp:
This is Modalka, a package to easily introduce (almost) native modal editing to Emacs. It helped me already because I've been typing a lot in recent days and since I started using it all the problems are gone. I studied all existing solutions before writing this. You can read readme file if you're interested.

It's still under testing, but already very useful, so it think others can benefit from it.

Here is the link: https://github.com/mrkkrp/modalka

### DIFF
--- a/recipes/modalka
+++ b/recipes/modalka
@@ -1,0 +1,1 @@
+(modalka :repo "mrkkrp/modalka" :fetcher github)


purcell:
Thanks!

PR #3101, opened 2015-09-11
Add recipe smart-mark

cheunghy:
`smart-mark` restores cursor position when `C-g` after mark.

### DIFF
--- a/recipes/smart-mark
+++ b/recipes/smart-mark
@@ -1,0 +1,1 @@
+(smart-mark :repo "cheunghy/smart-mark" :fetcher github)


recipes/smart-mark
```diff
@@ -0,0 +1 @@
+(ac-html :repo "cheunghy/smart-mark" :fetcher github)
```
syohex: `ac-html` -> `smart-mark`

cheunghy: I'm sorry.

cheunghy:
I've fixed the mistake.

purcell:
- Since you've got a list of `mark-*` functions, you can create the defadvice automatically using the list, rather than explicitly writing every `defadvice` form.
- Prefer `when` to `(if ... (progn ...))`

cheunghy:
How to make defadvice work with list?
I tried this, but not work.
``` elisp
(defmacro smart-mark-set-advice (func)
  `(defadvice ,func (before smart-mark-set-restore-before-mark activate)
     (setq smart-mark-point-before-mark (point))))

(mapc (lambda (f)
        (smart-mark-set-advice f))
      smart-mark-mark-functions)
```

cheunghy:
I make a global mode `smart-mark-mode`, then it should be activated and deactivated using the mode.

However, seems `advice-remove` is not work as expected
``` elisp
(defun smart-mark-remove-advices ()
  (advice-remove 'mark-page #'smart-mark-set-restore-before-mark)
  (advice-remove 'mark-paragraph #'smart-mark-set-restore-before-mark)
  (advice-remove 'mark-whole-buffer #'smart-mark-set-restore-before-mark)
  (advice-remove 'mark-sexp #'smart-mark-set-restore-before-mark)
  (advice-remove 'mark-defun #'smart-mark-set-restore-before-mark)
  (advice-remove 'mark-word #'smart-mark-set-restore-before-mark)
  (advice-remove 'keyboard-quit #'smart-mark-restore-cursor-cg-mark))
```
After calling the function, the advice seems not being removed.

How can I make the `advice-remove` work?

purcell:
How about this: https://github.com/cheunghy/smart-mark/pull/1 ? :-)

purcell:
What was wrong with your attempt was that "func" was a quote, so the macro expanded to:

```
(defadvice 'mark-word (...) ...)
```

instead of

```
(defadvice mark-word (...) ...)
```

purcell:
My pull request might not be the simplest way to do this, but hopefully you get the idea. :-)

purcell:
In any case, I'll merge this recipe for now. :-)

cheunghy:
Thanks!

PR #3102, opened 2015-09-11
Add recipe expand-line

cheunghy:
`expand-line` package works similar to `expand-region`, instead of region, it expands line.

Repository for this package is at https://github.com/cheunghy/expand-line

### DIFF
--- a/recipes/expand-line
+++ b/recipes/expand-line
@@ -1,0 +1,1 @@
+(expand-line :repo "cheunghy/expand-line" :fetcher github)


purcell:
Thanks. Quick feedback:

- You must not [call global-set-key](https://github.com/cheunghy/expand-line/blob/master/expand-line.el#L109-L110) for users
- One of your docstrings refers to "Projectile mode"

cheunghy:
Thanks!
The two issues are fixed.

purcell:
Thanks, merged.

PR #3103, opened 2015-09-11
Add recipe for elake

lujun9972:
Add recipe for elake

### DIFF
--- a/recipes/elake
+++ b/recipes/elake
@@ -1,0 +1,1 @@
+(elake :fetcher github :repo "lujun9972/elake")


purcell:
Thanks, but `elake.el` is not a valid elisp package, because it doesn't have the standard header/footer etc and a `provide` form.

Additionally, this recipe will include `elakefile.el` and `make-blog.el`, which seems undesirable.

tarsius:
Please reopen if you are still interested in getting this into Melpa.

PR #3105, opened 2015-09-11
Add flycheck-nim recipe

ALSchwalm:
`flycheck-nim.el` is a package that adds support for the [nim programming language](http://nim-lang.org/) to flycheck. It is available [here](https://github.com/ALSchwalm/flycheck-nim). I am the author and maintainer.

### DIFF
--- a/recipes/flycheck-nim
+++ b/recipes/flycheck-nim
@@ -1,0 +1,1 @@
+(flycheck-nim :repo "ALSchwalm/flycheck-nim" :fetcher github)


purcell:
Thanks, looks good. You might consider getting this added to flycheck core.

Your `(eval-after-load 'flycheck '(add-to-list 'flycheck-checkers 'nim))` can just be `(add-to-list 'flycheck-checkers 'nim)`, btw, because you're explicitly loading `flycheck` via the `require` form at top.

ALSchwalm:
You're right, thanks for the tip.

PR #3106, opened 2015-09-12
Update php-mode recipe

syohex:
Add skeleton files. [php-mode](https://github.com/ejmr/php-mode/) provides skeletons for PHP class, function, try-catch etc now.


purcell:
Please use `:defaults` instead of `"php-mode.el"` in this case. I think the result should be the same, but `:defaults` is more flexible for the future.

syohex:
Thanks for advice. I have updated.

purcell:
super, thanks

PR #3107, opened 2015-09-12
js2-highlight-vars

unhammer:
from http://mihai.bazon.net/projects/editing-javascript-with-emacs-js2-mode/js2-highlight-vars-mode

### DIFF
--- a/recipes/js2-highlight-vars
+++ b/recipes/js2-highlight-vars
@@ -1,0 +1,1 @@
+(js2-highlight-vars :fetcher github :repo "unhammer/js2-highlight-vars")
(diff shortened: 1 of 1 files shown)


purcell:
The recipe points to a nonexistent repo. Did you test it locally before submitting? Please do so - there are corresponding instructions in the README, and it's very easy. :-)

xuchunyang:
Maybe melpa can also check some common mistakes like this one with Travis CI or should not trigger Travis CI checks for a recipe summiting PR since now it is misleading.

purcell:
@xuchunyang Yes, that would be great, but in this case it would involve trying to build the recipe, so we would need a way to detect which recipes had been changed or added since the last successful build -- do you know of a way to do that?

unhammer:
Wow, that was a lot easier to test than I feared; sorry for the trouble.

purcell:
Looks good, thanks.

purcell:
Glad you found the testing easy in the end. I'd be pleased to have feedback if anything wasn't clear.

PR #3110, opened 2015-09-13
Add eshell-git-prompt recipe

xuchunyang:
Repo: https://github.com/xuchunyang/eshell-git-prompt


purcell:
Thanks. You should add `(cl-lib "0.5")` to the list of package dependencies, since it is not present in Emacs < 24.3.

xuchunyang:
Done in commit https://github.com/xuchunyang/eshell-git-prompt/commit/a4284abf137eb876cf8036787348c52a4484bec5

purcell:
Thanks!

PR #3112, opened 2015-09-13
Update ranger package location.

ralesi:
Based on some requests, have been asked to change repo name to designate as an elisp package, since the tool this is based on is also called just ranger.


purcell:
Makes sense, thanks.

PR #3130, opened 2015-09-14
flycheck-mypy recipe

lbolla:
Flycheck interface to MyPy checker for Python.
Code available at https://github.com/lbolla/emacs-flycheck-mypy
Author: Lorenzo Bolla <lbolla@gmail.com>


purcell:
Thanks.

PR #3132, opened 2015-09-14
update directory scheme for haskell-emacs

knupfer:
- tests are now in a seperate directory
- additional .hs files are in Foreign/
- .org file will be used for a help function


purcell:
Thanks.

PR #3133, opened 2015-09-14
Added sexp-move recipe

Elzair:
I added a recipe for a package that helps one navigate S-Expression based languages (and probably others too).

### DIFF
--- a/recipes/sexp-move
+++ b/recipes/sexp-move
@@ -1,0 +1,2 @@
+(sexp-move :repo "elzair/sexp-move"
+           :fetcher gitlab)


purcell:
Hi - thanks for this, but that repo does not exist afaict. Did you mean "sexp-mode"? Please build and test this locally following the guidelines in the README.

syohex:
@purcell The repository is in gitlab, not github.

https://gitlab.com/elzair/sexp-mode

purcell:
@syohex Yes, but the recipe says "sexp-move", not "sexp-mode".

syohex:
Oh, sorry.

Silex:
@Elzair: just curious, how does this differ from https://github.com/Fuco1/smartparens ?

Elzair:
I renamed the git repository to sexp-move. I apologize for the typo.

Elzair:
Well, I have never used **smart-parens** (or most of those packages). 

   I wrote **sexp-move** to make it easier to tab-through my code. I use [ParEdit](http://emacswiki.org/emacs/ParEdit) to automatically add matching parentheses, brackets, and quotation marks to my code, and I want to press *TAB* to get out of them. However, I found inadequate the [default movement commands of ParEdit](http://pub.gajendra.net/src/paredit-refcard.pdf). Here is an example of **paredit-forward**.

```elisp
(foo| (bar baz) quux)
(foo (bar baz)| quux)
```
   
   The command **paredit-forward** skips over the nested S-Expression *(bar baz)*. ParEdit has two "undocumented" commands, **paredit-forward-down** and **paredit-forward-up** that move into and out of nested expression, but it does not have any command to DWIM when traversing the code.

To fill this need, I wrote **sexp-move**.

Elzair:
Would changing the name to **lexeme-move** help to clarify its purpose?

purcell:
Thanks, merged. The `char-code-property` code looks a bit odd to me. Seems like `syntax-ppss` should give you all the information you would need, though I could be wrong.

PR #3134, opened 2015-09-15
added recipe "move-lines"

Kaligule:
I really tried to do the tests and think everything is ok.

### DIFF
--- a/recipes/move-lines
+++ b/recipes/move-lines
@@ -1,0 +1,1 @@
+(move-lines :repo "targzeta/move-lines" :fetcher github)


purcell:
Thanks. This looks the same http://melpa.org/#/move-dup and http://melpa.org/#/move-text, right? It'd be best for users if we don't add another package which does the same thing.

Kaligule:
Hey @purcell 
while move-text doesn't work very well with multiple cursers (thats just my experience), move-dup seems to work well. Thank you.

purcell:
Cool, thanks - `move-dup` is also what I use myself.

PR #3136, opened 2015-09-16
Added stack-mode package for Haskell Stack.

LukeHoersten:


purcell:
Thanks Luke, but the upstream source is not yet in good shape for packaging:

- The line-wrapped description line is invalid
- None of its many package dependencies are declared (cl-lib, flycheck, haskell-mode)
- there are no `;;;###autoload` cookies
- There's no trailing comment line
- There's a spurious "checklist.el" in the directory which is apparently unused
- There's a "fifo.el" which should either be a separate package (if it's widely usable) or renamed to "stack-fifo.el" or similar

The integration with `flycheck` is also decidedly questionable IMO, and it's hard to see why a `stack` user would want or need to have flycheck installed and enabled (by default) just because they installed `stack-mode`.

purcell:
An informative `Commentary:` section would also be good. :-)

LukeHoersten:
Thanks for taking the time to provide feedback. I'll get things in shape
and submit another PR.

As for flycheck integration: this mode is more than just stack integration.
This is 'stack-ide' which uses 'stack' plus 'ide-backend' to find and
compile source on the fly for checking and code nav.

I think your expectation that this wouldn't have anything to do with
flycheck is coming from the name. I'll see if renaming the mode to
'stack-ide-mode' is acceptable.

purcell:
Thanks. If you're to propose renaming it, then "stack-ide" would be better than "stack-ide-mode", because the "-mode" prefix is deprecated for minor modes.

I'm a `stack`, `flycheck` and `haskell-mode` user, and I'm all for better tooling, so it's great to see some work taking shape in this area.

LukeHoersten:
@purcell sounds good. Again thanks for the feedback and quick response.

LukeHoersten:
@purcell I think we've addressed all the issues you've pointed out. Can we re-open this PR? Thanks!

purcell:
Yep, looks good, thanks.

LukeHoersten:
Thanks Steve. Getting this in melpa I think is critical for growing this
softwares usage. Have a good week!

LukeHoersten:
Hmm looks like something is wrong with the merge/CI.

purcell:
Nah, I think it's fine.

LukeHoersten:
Yeah it's in now. Thanks.

PR #3138, opened 2015-09-18
Add recipes/jabber-otr

legoscia:
This is an add-on for OTR encryption in jabber.el.

The upstream repository is [legoscia/emacs-jabber-otr](https://github.com/legoscia/emacs-jabber-otr/) on GitHub. I am the maintainer.

### DIFF
--- a/recipes/jabber-otr
+++ b/recipes/jabber-otr
@@ -1,0 +1,3 @@
+(jabber-otr
+ :fetcher github :repo "legoscia/emacs-jabber-otr"
+ :files (:defaults "emacs-otr.py"))


purcell:
Looks good, but the recipe should be adjusted here please:

- Use `:defaults` instead of the `jabber-otr.el` entry in `:files` -- it includes `"*.el"` and is more future-proof
- Remove `README` from that list. We don't ship non-GNUinfo docs with packages, because they're not easy to find. Instead, put any important information in the `Commentary` section of the `.el` file's header

legoscia:
There, updated the recipe.

purcell:
Awesome, thanks.

PR #3139, opened 2015-09-18
Added pc-bufsw recipe

ibukanov:
I am the maintainer and original author of https://github.com/ibukanov/pc-bufsw .

This switches Emacs buffers according to most-recently-used/least-recently-used order using C-tab and C-S-tab keys. It is similar to window or tab switchers that are available in PC desktop environments or applications.

### DIFF
--- a/recipes/pc-bufsw
+++ b/recipes/pc-bufsw
@@ -1,0 +1,1 @@
+(pc-bufsw :fetcher github :repo "ibukanov/pc-bufsw")


purcell:
Hi - the autoloads here unconditionally change the global keybindings, which is not acceptable behaviour for a package: it should be possible to install a package but never use it. Instead, either leave the keybindings to the user, or provide an autoloaded global minor mode which has those binding in its keymap: then the user can just enable the minor mode to get the keybindings.

ibukanov:
The reason for those autoloads is that I wanted to trigger loading of the package *only* when the user presses the corresponding keybindings. With the minor mode I do not see how to get this. It seems activating the minor mode unconditionally loads the package. Am I mistaken?

purcell:
> The reason for those autoloads is that I wanted to trigger loading of the package only when the user presses the corresponding keybindings.

That's why normally you would simply provide autoloaded commands, and then leave the user to bind them according to their wishes: this will then allow the library to be autoloaded when used. An elisp library shouldn't be calling `global-set-key` unconditionally in any case.

`windmove` is a built-in library which provides a similar type of functionality to `pc-bufsw` - it provides autoloaded commands which the user can bind, and it also provides an autoloaded command called `windmove-default-keybindings` which installs the global keybindings. That command will load the package unconditionally, but that's just the price for following the rules in this case.

Hope that helps.

ibukanov:
Could you clarify if the rules against putting any code into autoloads beyond explicitly supported constructs or just against automatic keybindings? I.e. if I change the package to load the keybindings in autoloads only if the user sets a defcustom flag, would it be OK? 

I am curious if I can have keybindings that can be set entirely through the customization interface and yet that would allow to delay code loading until the keys are pressed even after customization.

purcell:
> I.e. if I change the package to load the keybindings in autoloads only if the user sets a defcustom flag, would it be OK?

But then are the autoloads evaluated before or after the custom settings are loaded? It would make a difference.

> I am curious if I can have keybindings that can be set entirely through the customization interface and yet that would allow to delay code loading until the keys are pressed even after customization.

No, not really.

I feel like you're swimming against the stream here and trying to make things more complicated than they need to be. If there were an idiomatic way to do this, it would already be used by `windmove` etc. This is a tiny package, and if the user sets the keybindings himself, then the package will already be fully autoloaded.

ibukanov:
I updated https://github.com/ibukanov/pc-bufsw so it enables keybindings only after the user explicitly sets the `pc-bufsw-keys-enable` flag. The new code still avoids loading the whole files even if the flag is set to true and provides convenient `pc-bufsw-default-keybindings` to activate default keybindings manually.

This required to copy few defcustom calls into the autoload file, see https://stackoverflow.com/questions/32693757/emacs-package-customization-and-autoloads for the reason to do so. I wish there would be a less hackish way to do that, but at least this allows to activate/customize the package without touching `init.el` while minimizing loading time.

purcell:
Cool, thanks.

PR #3140, opened 2015-09-18
Added spotlight recipe

benmaughan:
This PR adds the package spotlight which provides functions to search files using the OS X spotlight utility. Leverages ivy to give incremental search and narrowing of results.

I am the author of the package, and it is hosted here:
https://github.com/benmaughan/spotlight.el

### DIFF
--- a/recipes/spotlight
+++ b/recipes/spotlight
@@ -1,0 +1,1 @@
+(spotlight :repo "benmaughan/spotlight.el" :fetcher github)


purcell:
Thanks. Quick feedback:

- Your `defcustom` forms should have their `:group` specified
- You're [modifying `ivy-minibuffer-map` globally](https://github.com/benmaughan/spotlight.el/blob/master/spotlight.el#L101), not just for *your* use of ivy: that seems wrong to me.
- Having both `spotlight-base-dir` and `spotlight-user-base-dir` seems unnecessary to me. Just use `read-directory` when no previous directory has been specified and cached.
- You might also need to use `shell-quote-argument` to escape the queries when constructing command lines

Also, stylistic tip -- don't do this:

```el
(if arg
      ;; prompt for dir
      (setq spotlight-user-base-dir (read-directory-name "base directory: "))
    ;; else use default
    (setq spotlight-user-base-dir spotlight-base-dir))
```

when you can just do this:

```el
(setq spotlight-user-base-dir (if arg
                                   (read-directory-name "base directory: ")
                                 spotlight-base-dir))
```

Hope that helps!

benmaughan:
Thanks for taking the time to look and comment! I'll look at fixing those and submit and updated version.

purcell:
Yep, just leave this open and comment back here when ready. :-)

benmaughan:
Hi,

I've taken on your comments and the only one I don't think I can include is:

> Having both spotlight-base-dir and spotlight-user-base-dir seems unnecessary to me. Just use read-directory when no previous directory has been specified and cached.

I've renamed `spotlight-base-dir` to `spotlight-default-base-dir` to reflect the fact that I want it to act as the default base directory if none is specified using the prefix argument. I couldn't see a way to get this behavior without storing the default in a separate variable.

Let me know if you have any additional comments - your previous ones were really useful!

Cheers,
 Ben

purcell:
Makes sense - merging now. Thanks for your patience!

PR #3141, opened 2015-09-19
add recipe for fzf.el

bling:
created a small package for integrating with https://github.com/junegunn/fzf

### DIFF
--- a/recipes/fzf
+++ b/recipes/fzf
@@ -1,0 +1,2 @@
+(fzf :repo "bling/fzf.el" :fetcher github)
+


purcell:
The file uses the functions from `nadvice.el`, which was only added in Emacs 24.4, so `fzf.el` needs to have a `;; Package-Requires: ((emacs "24.4"))` header.

bling:
thanks for the code review.  i've added the dependency declaration.

purcell:
Thanks, merged.

PR #3142, opened 2015-09-20
Add recipe for mgalgs/diffview-mode

mgalgs:
Project summary: Renders unified diff buffers in a side-by-side format using two buffers
Project URL: https://github.com/mgalgs/diffview-mode
I am the author.

### DIFF
--- a/recipes/diffview
+++ b/recipes/diffview
@@ -1,0 +1,1 @@
+(diffview :fetcher github :repo "mgalgs/diffview-mode")


purcell:
Thanks, looks good. I'm kinda surprised this can't be done already with `ediff`. :-)

I suggest you run `M-x checkdoc` and fix up the docstring comments according to its suggestions. There's also a lot of spurious markdown in the `Commentary` section, which won't render in the MELPA package page, and leaves the text clunky to read inside Emacs. If you're generating the README from that content - or vice versa - you might like to figure out how to avoid this.

mgalgs:
Cool, I had never heard of `checkdoc`.  I've fixed all the warnings.

Regarding the markdown, I am indeed generating the `README.md` from the file comments, using another little tool I wrote: [`make-readme-markdown`](https://github.com/mgalgs/make-readme-markdown).  I'm working on adding support to automatically insert badges (like the license badge) without needing to mark it up in the file comments.  I should also be able to teach `make-readme-markdown` how to insert `img` tags from raw links, but I guess those will still be stray text on MELPA.  They would look something like:

    Screenshots:

    Before:
        https://raw.github.com/mgalgs/diffview-mode/master/screenshots/diffview-before.png

    After:
        https://raw.github.com/mgalgs/diffview-mode/master/screenshots/diffview-after.png


Is that a show-stopper?

purcell:
> Is that a show-stopper?

Nope, not at all. BTW, I think there are a couple of similar tools for README/Commentary conversion. Have a feeling one was by Matthew Fidler...

Anyhow, I'll go ahead and merge this now. :-)

Silex:
For information, `diff-ediff-patch` or `epatch`more or less achieves the same, except it patches the file and saves the original one with `.orig`.

purcell:
Cool -- if so, perhaps the implementation could even use that functionality to make its code simpler.

Silex:
Ah, just noticed it `diffview-mode` seems to work with multifile patches... Not sure ediff can do that.

It's probably possible to reuse ediff's internal functions tho.

mgalgs:
AFAICT, all of `ediff`'s functionality relies on a physical checkout of whatever code base is being diff'd.  `diffview-mode` works on raw patch files directly, no need for an underlying source tree where the patch can be cleanly applied.  This is especially useful for reviewing lots of patches from mailing lists since things don't always apply cleanly to what you have locally.

PR #3143, opened 2015-09-21
Added find-and-ctags recipe

redguardtoo:
use find and ctags to update TAGS. It supports windows.

### DIFF
--- a/recipes/find-and-ctags
+++ b/recipes/find-and-ctags
@@ -1,0 +1,1 @@
+(find-and-ctags :fetcher github :repo "redguardtoo/find-and-ctags")


purcell:
Thanks for this, but I don't think the code is ready for distribution yet:

- The package name should match up with the symbol prefix here, so either `fctags.el`, or use `find-and-ctags-*` for the symbols.
- There's a typo (`fctags-get-hostnmae`)
- Some code is unnecessarily [repetitive](https://github.com/redguardtoo/find-and-ctags/blob/master/find-and-ctags.el#L100-L111)
- You shouldn't be calling `cd`: instead, use `let` to temporarily set `default-directory` to the desired value while running commands
- Calling `executable-find` on a full path is also incorrect: you probably want to use `file-executable-p`
- Building up command strings with `format` and `%s` without also using `shell-quote-argument` is dangerous, since the interpolated paths could easily contain spaces or other characters (e.g. `"; rm -rf / ;"`)
- Interactive commands should have docstrings, and they should be valid according to `M-x checkdoc`

redguardtoo:
Thanks. I update the code. `shell-quote-argument` doesn't work in my case, so I escape my options in my own function.

purcell:
> shell-quote-argument doesn't work in my case, so I escape my options in my own function.

But that doesn't handle `'` and `"`. I don't see why `shell-quote-argument` wouldn't be the right thing to use here.

Wanted to mention that you can write
```el
(if find-and-ctags-gnu-find-executable
                      find-and-ctags-gnu-find-executable
                    (find-and-ctags--guess-executable "find"))
```
more neatly as
```el
(or find-and-ctags-gnu-find-executable
    (find-and-ctags--guess-executable "find"))
```

I also noticed that `find-and-ctags-get-hostname` isn't used anywhere, and doesn't really belong with the other code.

The remaining little bit of code feels like personal configuration, to be honest -- I'm not sure it's a widely reusable package in its current form, so my suggestion would be that we don't merge this into MELPA.

redguardtoo:
`shell-quote-argument` won't work because all the options are passed as one string.

If I execute `(shell-command (format "find . %s" (shell-quote-argument "-name '*.el' -not -iwholename '*.elc'")))`, I will get the error message:
```
find: unknown predicate `-name '*.el' -not -iwholename '*.elc''
1 (#o1, #x1, ?\C-a)
```

dgutov:
> because all the options are passed as one string

Why not change that?

tarsius:
> The remaining little bit of code feels like personal configuration, to be honest -- I'm not sure it's a widely reusable package in its current form, so my suggestion would be that we don't merge this into MELPA.

Shall we close this then? (It has been half a year.)

purcell:
Yep, thanks.

PR #3147, opened 2015-09-23
add `use-package-chords` recipe

waymondo:
adds a `key-chord` keyword for `use-package`

### DIFF
--- a/recipes/bind-chord
+++ b/recipes/bind-chord
@@ -1,0 +1,3 @@
+(bind-chord :repo "waymondo/use-package-chords"
+            :fetcher github
+            :files ("bind-chord.el"))
--- a/recipes/use-package-chords
+++ b/recipes/use-package-chords
@@ -1,0 +1,1 @@
+(use-package-chords :repo "waymondo/use-package-chords" :fetcher github)
--- a/recipes/use-package-chords
+++ b/recipes/use-package-chords
@@ -1,3 +1,3 @@
 (use-package-chords :repo "waymondo/use-package-chords"
                     :fetcher github
-                    :files ("use-package-chords.el" "bind-chord.el"))
+                    :files ("use-package-chords.el"))
--- a/recipes/use-package-chords
+++ b/recipes/use-package-chords
@@ -1,1 +1,3 @@
-(use-package-chords :repo "waymondo/use-package-chords" :fetcher github)
+(use-package-chords :repo "waymondo/use-package-chords"
+                    :fetcher github
+                    :files ("use-package-chords.el" "bind-chords.el"))
--- a/recipes/use-package-chords
+++ b/recipes/use-package-chords
@@ -1,3 +1,3 @@
 (use-package-chords :repo "waymondo/use-package-chords"
                     :fetcher github
-                    :files ("use-package-chords.el" "bind-chords.el"))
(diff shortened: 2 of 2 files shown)


purcell:
From the contents of this file I'd say the right name for this would be `bind-chord` or similar, right? The symbols a package defines should have a common prefix matching up with the package name.

waymondo:
The package's primary use is as an extension to `use-package` though. The `bind-chord` defun and `bind-chords` macro exist to feed the `use-package-handler/:chords`, but could also be used by itself. In `use-package`, these two components are split between `bind-key` and `use-package`, but I didn't feel like there was enough code to justify two separate files or to break the naming convention. Does that seem ok?

waymondo:
@jwiegley do you have a preference on the naming convention for `use-package` extensions?

jwiegley:
My preference would probably be `up-chords`, since `use-package-` makes for a terrible prefix.  Another possibility would be `use-chords`.

Silex:
`up-chord` is not descriptive tho... I prefer `use-package-chords`.

jwiegley:
There is that....

waymondo:
Yeah I'm conflicted here.. I agree that it is usually best practice is
to always use a common prefix, but given the extension support for
use-package requires you to define handlers using
`use-package-normalize/*` and `use-package-handler/*`, this kind of
package already can't follow that, which I think this is fine in this
case.

I also think `bind-chords` is the right wording for the autoloaded
defun, as I know it is pretty common for `use-package` users to call
`bind-keys` within a config block, i.e.:

``` elisp
(use-package css-mode
  :config
  (bind-keys :map css-mode-map
             (":" . smart-css-colon)
             ("," . pad-comma)
             ("{" . open-brackets-newline-and-indent)))

```

This would allow for consistency for those who would want to call
`bind-chords` in the same manner.

As per the packages name itself, I think `use-package-chords` is still
the most intuitive and descriptive. To me it seems clear that its
purpose is to extend `use-package` and that it must add support for
`chords`, whatever that may be.

So I guess I would vote that everything still seems ok me as is. Hope
this makes sense!

jwiegley:
Ok, I'll go with that.  `use-package-chords` it is!

waymondo:
@purcell Please review my comment above when you get a chance. Thanks as always.

purcell:
> The package's primary use is as an extension to use-package though. The bind-chord defun and bind-chords macro exist to feed the use-package-handler/:chords, but could also be used by itself. In use-package, these two components are split between bind-key and use-package, but I didn't feel like there was enough code to justify two separate files or to break the naming convention. Does that seem ok?

I'd still split them up, and then `use-package-chords` would be the package which provides the wiring between `bind-chords` and `use-package`.

waymondo:
Ok, done and updated.

purcell:
Okay, but then let's make this 2 packages rather than 1, like `bind-key` and `use-package`. The files can still live in the same repo if that's more convenient for you: we'd just need appropriate `:files` specs.

waymondo:
Done

purcell:
Cool, but now you can also remove `bind-chord.el` from the `:files` list in the `use-package-chords` recipe. :-)

waymondo:
Alrighty, makes sense. Updated.

purcell:
Nice, thanks for your patience. Merged!

PR #3149, opened 2015-09-26
Add recipe for spaceline

TheBB:
Spaceline is the mode-line configuration toolbox used in the popular Spacemacs distribution, which has been extracted as an independent package due to general interest.

* Upstream: https://github.com/TheBB/spaceline
* Spacemacs: https://github.com/syl20bnr/spacemacs

I'm the maintainer of Spaceline, and a collaborator with Spacemacs. @syl20bnr (CC) is the maintainer of Spacemacs.

### DIFF
--- a/recipes/spaceline
+++ b/recipes/spaceline
@@ -1,0 +1,1 @@
+(spaceline :fetcher github :repo "TheBB/spaceline")


purcell:
Thanks. Couple of things I'd point out with the code here:

- Code like `:foreground ,(face-background 'mode-line)` which defines faces by inspecting currently-applied faces is hostile to theming, since it means the default face definition will vary according to the capabilities of the current frame and the order in which this package is loaded relative to the user's theme being set. Better to just choose a default and complete face definition which will be kinda okay for everyone, e.g. `:inherit mode-line :inverse-video t`, or even `:inherit highlight`.
- `defstruct sl--seg` should be `defstruct spaceline--seg`. If everyone chose 2-letter abbreviations for package prefixes, we'd have run out of unique prefixes when MELPA was only a quarter of its current size. :-)

Is spacemacs going to switch to using this package?

syl20bnr:
Great work!
Yes, Spacemacs will use it.

TheBB:
Thanks, I have addressed your comments. There will be a PR for Spacemacs as soon as this is on MELPA.

purcell:
Yay, great - merged!

PR #3151, opened 2015-09-27
Add recipe for restart-emacs

iqbalansari:
This package adds a single command `resem-restart-emacs` which can be used restart emacs from within emacs. The code lives in https://github.com/iqbalansari/restart-emacs. I am the author and the maintainer.

### DIFF
--- a/recipes/restart-emacs
+++ b/recipes/restart-emacs
@@ -1,0 +1,1 @@
+(restart-emacs :fetcher github :repo "iqbalansari/restart-emacs")


rekenerd:
:+1:

purcell:
Cool, thanks. Since the package name and file name is `restart-emacs`, the prefix for the symbols you define should be `restart-emacs-`, not `resem`, as per the [coding conventions](https://www.gnu.org/software/emacs/manual/html_node/elisp/Coding-Conventions.html). The main command can be called `restart-emacs`.

iqbalansari:
Thanks for the feedback, actually I [had](https://github.com/iqbalansari/restart-emacs/blob/e6cbdb75f6f91864f389ea7272f0a17f1ee49073/restart-emacs.el) restart-emacs as the prefix but I changed it `resem-` because it did not seem appropriate to me to introduce a command that looks like a core command (think `kill-emacs`, `suspend-emacs`).  I have reverted the code back to use `restart-emacs` as the prefix.

Also wanted to thank you the awesome work you and @milkypostman  have done with melpa. So **thank you** guys.

purcell:
:+1: The other way to do things would have been to call the package `resem`, and name the symbols accordingly, but I think this is fine.

iqbalansari:
Thanks!

PR #3154, opened 2015-09-28
Add recipe for julia-shell

dennisog:
This package adds better emacs integration with an interactive [Julia](http://julialang.org/) shell. [The original julia-mode](https://github.com/JuliaLang/julia/blob/master/contrib/julia-mode.el) has some shell features, but it lacks completions and other features. As of right now, I'm the sole maintainer and the code sits [here](https://github.com/dennisog/julia-shell-mode).

### DIFF
--- a/recipes/julia-shell
+++ b/recipes/julia-shell
@@ -1,3 +1,3 @@
 (julia-shell :fetcher github
              :repo "dennisog/julia-shell-mode"
-             :files ("*.el" "*.jl"))
+             :files (:defaults "*.jl"))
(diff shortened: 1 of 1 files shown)


purcell:
Thanks. Please use `(:defaults "*.jl")` rather than `("*.el" "*.jl")` - it's a little more future-proof, and therefore preferred. Everything looks good otherwise.

dennisog:
Thanks, I committed the requested changes.

purcell:
Great, thanks - merged!

PR #3158, opened 2015-09-30
Add some helm plugin recipe

myuhe:
Revert https://github.com/milkypostman/melpa/pull/3157
Repo has changed.
`helm-gist` is no loger work correctly , So no recipe is comitted.


purcell:
Thanks.

PR #3161, opened 2015-09-30
Add defproject recipe

danlamanna:
[defproject](https://github.com/kotfic/defproject) lets you define dir-locals in elisp in a cleaner way using use-package like macros.

I'm a contributor with push access.

I contributed [57c0bcb](https://github.com/kotfic/defproject/commit/57c0bcb4adfa630e669dbd6f76e846465da6a149) and [5b5f6](https://github.com/kotfic/defproject/commit/5b5f6283ed72a430664bb7ca8f9382d0ac22dc21)  to make compatible with MELPA.

### DIFF
--- a/recipes/defproject
+++ b/recipes/defproject
@@ -1,0 +1,1 @@
+(defproject :fetcher github :repo "kotfic/defproject")


purcell:
Cool. I'd suggest a few cleanups:

- Many of the `;;;` in the header should be `;;` -- check out what gets inserted by `M-x auto-insert` in a fresh empty `.el` file to see the right format.
- It'd be worth running `M-x checkdoc` and fixing up some of the docstrings accordingly
- It seems wasteful to depend on `dash` just for `-map`, when `mapcar` is built in and does the same thing
- Similarly, `s-chop-prefix` is easily done with `replace-regexp-in-string`, and `s-contains?` is trivially performed with `string-match-p`, so you could arguably drop the `s` dependency

Hope that helps!

danlamanna:
These should be set - [diff](https://github.com/kotfic/defproject/compare/93ea...master).

milkypostman:
last thing, you either need to add `;; Package-Requires: ((emacs "24"))` or remove the `-*- lexical-binding: t; -*-` at the top.

kotfic:
Fixed upstream,   can you update the PR @danlamanna ?

Thanks for your patience guys -  can you tell this is my first rodeo?

milkypostman:
@kotfic since you removed the lexical binding I don't think you need to require emacs 24 anymore.

merging nonetheless.

PR #3163, opened 2015-09-30
Add recipe for jscs

papaeye:
[jscs.el](https://github.com/papaeye/emacs-jscs) is a elisp for [JSCS (JavaScript Code Style)](http://jscs.info/):

- Apply indentation rules from JSCS config file to the current buffer
- Run `jscs --fix` borrowed code from gofmt of go-mode.el

I'm a maintainer (papaeye/emacs-jscs#1).

### DIFF
--- a/recipes/jscs
+++ b/recipes/jscs
@@ -1,0 +1,1 @@
+(jscs :repo "papaeye/emacs-jscs" :fetcher github)


purcell:
Cool. Exposing the `jscs-fix-before-save` function is a bad idea: you never want someone to just add a global `before-save-hook` entry, and then try to be responsible for checking the current mode and doing the right thing. I'd strongly suggest removing that, because `jscs-fix-run-before-save` is the right thing for users to use.

Looks good otherwise!

papaeye:
Removed. Thank you for reviewing!

https://github.com/papaeye/emacs-jscs/commit/06bc9dc29e9c05062ecaafb63ba0a2bb349901d2

purcell:
Merged - thanks!

PR #3164, opened 2015-09-30
Add recipe for wordnut

gromnitsky:


syohex:
Your package does not comply Emacs packaging convention. There are no package headers(`Author`, `Version`), CopyRight, License, Comment section etc(Example is here https://github.com/milkypostman/melpa/blob/master/package-build.el#L1-L40).

Please see
- https://www.gnu.org/software/emacs/manual/html_node/elisp/Simple-Packages.html#Simple-Packages
- https://www.gnu.org/software/emacs/manual/html_node/elisp/Packaging-Basics.html#Packaging-Basics

recipes/wordnut
```diff
@@ -0,0 +1,3 @@
+(wordnut :repo "gromnitsky/wordnut"
+	:fetcher github
+	:files ("*.el"))
```
syohex: You need not specify `:files` property. `*.el` is default value.

Please see item `:files` of https://github.com/milkypostman/melpa#recipe-format

purcell:
> Your package does not comply Emacs packaging convention.

@syohex It's actually okay in this case, technically since a `wordnut-pkg.el` file is provided, allowing MELPA to find the package metadata. It'd still be nice if the `.el` files followed the usual boilerplate, though, just as a matter of style.

The `:files` spec in the recipe should certainly be removed though - @gromnitsky, can you please amend and force-push the recipe accordingly?

gromnitsky:
@purcell: the `:files` spec is removed.

@syohex: I think the comments boilerplate is required for single file packages only. Elisp library header conventions are unbelievably verbose, thus I try to avoid them if it's technically possible.

purcell:
Thanks, merged.

You might not like the boilerplate, but it's not there for an author's benefit, it's there for the benefit of the community, so that important information about source code is recorded somewhere in a consistent format. So it's worth reconsidering, IMO.

PR #3169, opened 2015-10-01
Add Recipe for Mingus

noctuid:
[Mingus](https://github.com/pft/mingus) is a Music Player Daemon frontend. I am making the pull request since the author is busy. The full discussion with the author about adding mingus to MELPA can be found at pft/mingus#9. I tested it in the sandbox and had no problems.

### DIFF
--- a/recipes/mingus
+++ b/recipes/mingus
@@ -1,0 +1,1 @@
+(mingus :repo "pft/mingus" :fetcher github)


purcell:
Nice - I might use this myself. Some feedback on the code which you might like to take upstream:

- There's a spurious leading quote: https://github.com/pft/mingus/blob/master/mingus.el#L553
- Consider using `cl-lib` now instead of `cl` + `cl-macs`: it's mostly the same, and now preferred. Usually the only change is prefixing the corresponding symbols with `cl-`, e.g. `defstruct` -> `cl-defstruct`.
- Rather than listing all the keybindings (which might get changed) in the help text, there should be a way to interpolate the current keybindings with the `\\{...\\}` syntax: worth looking at. :-)

PR #3170, opened 2015-10-02
Add recipe for nameframe

john2x:
https://github.com/john2x/nameframe

Nameframe is a package to manage frames by name.

It was designed to be used with Projectile and/or perspective.el.
It effectively allows users to have dedicated frames for their projects.

Demo is available in the project's README.

### DIFF
--- a/recipes/nameframe
+++ b/recipes/nameframe
@@ -1,0 +1,4 @@
+(nameframe
+ :fetcher github
+ :repo "john2x/nameframe"
+ :files ("nameframe.el"))


purcell:
Thanks. This recipe will include all the `.el` files, which I don't think is what's desired here based on the current metadata headers: `nameframe-projectile` and `nameframe-perspective` appear to be designed as separate packages. Can you please add an appropriate `:files` spec to the recipe by amending the recipe commit, then force-push it back here to update the PR cleanly? You'd then also need to submit separate recipes for those other two packages. Let me know if you have any questions.

john2x:
Thanks. That's what I was confused about. The package was created/intended to be used together with Projectile and perspective.el. But at the same time, `nameframe.el` works on its own.

I like your suggestion about having `nameframe-projectile` and `nameframe-perspective` as separate packages. I'll update the README and this PR.

john2x:
I updated this PR. I'll create new PR's for the other packages once this gets accepted.

purcell:
Super, thanks.

PR #3172, opened 2015-10-03
Add recipe tab-jump-out

cheunghy:
tab-jump-out is a package for jump out of parenthesis and quotes with tab.

Usage:

you can enable tab-jump-out-mode to tab out of parenthesis, or just bind tab-jump-out to tab.

When working with auto complete engine like auto-complete and company, and yasnippet, you can set yas fallback behavior to tab jump out and do not enable tab jump out mode to ensure yasnippet run first.

### DIFF
--- a/recipes/scratches
+++ b/recipes/scratches
@@ -1,0 +1,1 @@
+(scratches :repo "cheunghy/scratches" :fetcher github)
--- a/recipes/tab-jump-out
+++ b/recipes/tab-jump-out
@@ -1,0 +1,1 @@
+(tab-jump-out :repo "cheunghy/tab-jump-out" :fetcher github)


cheunghy:
scratches is a package providing features for multiple savable scratches.

tab-jump-out
```diff
@@ -0,0 +1 @@
+(tab-jump-out :repo "cheunghy/tab-jump-out" :fetcher github)
```
syohex: `tab-jump-out` is in top directory. You should add recipe file in `recipes` directory.

cheunghy:
Thanks @syohex , I've moved the recipe to right place.

milkypostman:
in the future it is helpful to make each addition its own commit and PR.

is there any reason that scratches is a minor mode? it seems like this is just some helper functions.

The only other thing that I think @purcell may say is that there is no commentary and it would be good to fill that in for each package. This way something will show up in the help text if someone selects the package from the package list.

cheunghy:
@milkypostman minor mode toggles the availability of the helper functions. And if user don't want the default keybindings, he can just bind his keybindings and do not enable the minor mode.

purcell:
Re. `scratches`:

`"C-c s"` is reserved for user keybindings, so it shouldn't be a default prefix. See the [Key Binding Conventions](https://www.gnu.org/software/emacs/manual/html_node/elisp/Key-Binding-Conventions.html) info.

Re. `tab-jump-out`:

`defvar-local` requires Emacs 24.3, IIRC, or perhaps even 24.4. Please add a corresponding `;; Package-Requires` header, or use `defvar` + `make-variable-buffer-local`. Also `"Use tab to jump out of delimiter pairs"` would be a clearer description.

cheunghy:
@purcell Thanks. All these checks are done.

I changed key binding for scratches to `C-x M-s`, hopes it's not reserved. :D

Actually a lot of packages define keys on C-c map, it's not really user's reserved anymore, it seems like major-mode reserved and minor-mode careful bind. 

In my humble practice, C-x is emacs itself reserved, C-c is mode reserved, and I use C-z for my own keybindings. The reasons are:

1. default command for C-z is the same with C-x C-z, and C-x C-z feels consistent with C-x C-c.
2. C-z is just near C-x and C-c, make the three major prefix keys arranged same place on keyboard.

If package authors cannot use C-c prefix key, they will define keybindings in a strange unpredictable way.

IMHO, If somebody can work on the key binding conventions, he should revision it.

milkypostman:
@cheunghy i think you misunderstood what purcell said,

C-c + <single letter> is reserved for user bindings. You could use C-c C-s or C-c and any chorded keybinding. it's just that C-c plus a single letter should not be used by packages. C-x <anything> is actually worse.

cheunghy:
@milkypostman 
Thanks.
A lot of users use C-c C-s to toggle REPL and source file. And a lot of comint derived mode define this.
So I define it to be C-c C-r.

PR #3173, opened 2015-10-03
Added multi-compile

ReanGD:


purcell:
Thanks, but there's no `Commentary` section and no useful docstrings, so it's going to be hard for a user to figure out how to use this package: elisp libraries should be self-documenting. Can you please take some of the README contents and put them in the elisp file?

ReanGD:
Fixed.

But my English is poor, sorry (

purcell:
That's great, thanks. :-)

PR #3174, opened 2015-10-04
Split off EmacSQL backends into separate packages.

skeeto:
The `emacsql` package is broken into: `emacsql`, `emacsql-sqlite`, `emacsql-psql`, and `emacsql-mysql`. PostgreSQL users don't want to be bothered with a SQLite binary and SQLite users don't want pg.el. I've already made the appropriate changes in the EmacSQL repository.

See also: skeeto/emacsql/issues/15


purcell:
Makes sense, thanks.

PR #3175, opened 2015-10-04
Add go-playground recipe

grafov:
Add new recipe with go-playground mode for Go language. Under GPLv3.

### DIFF
--- a/recipes/go-playground
+++ b/recipes/go-playground
@@ -1,0 +1,1 @@
+(go-playground :repo "grafov/go-playground" :fetcher github)


milkypostman:
This looks good. Only thing that I saw was possibly that you don't need lexical binding?

@purcell are we still trying to sway people into avoiding lexical binding?

purcell:
Yes, use `;; Package-Requires: ((emacs "24"))` explicitly if `lexical-binding` is required, otherwise I'd strongly suggest removing that declaration.

grafov:
Thank for advice, I removed lexical-binding declaration from the source.

purcell:
Thanks - merged!

PR #3176, opened 2015-10-04
Add recipe for nameframe-perspective

john2x:
https://github.com/john2x/nameframe

Nameframe is a package to manage frames by name.

This package provides perspective.el integration for nameframe (#3170).

### DIFF
--- a/recipes/nameframe-perspective
+++ b/recipes/nameframe-perspective
@@ -1,0 +1,4 @@
+(nameframe-perspective
+ :fetcher github
+ :repo "john2x/nameframe"
+ :files ("nameframe-perspective.el"))


milkypostman:
would these two functions perform better as modes and using advice?

i don't have any qualms with the way they are now just asking. 

will wait for @purcell to chime in but looks ok to me.

john2x:
I based/copied my implementation on [`persp-projectile.el`](https://github.com/bbatsov/projectile/blob/master/persp-projectile.el).

To be honest, I'd prefer if users only have to install the one `nameframe` package and then just enable it with a command or something. (When does it make sense to have a minor mode?)

Would love to hear The Right Way to do this :)

milkypostman:
well separating to different packages is a decent thing to do so people don't get autoloads for the other package.

but I get what you mean. i was just looking at how you replace the function and was thinking that advising might be better. we'll see what @purcell says. he is the arbiter around here.

purcell:
The code uses `advice-add`, which only became available in Emacs 24.4 (see `nadvice.el`). You'd therefore need to declare a dependency on that Emacs version... however, using advice isn't idiomatic here, and I'd suggest the following instead:

- Create a `nameframe-make-frame-hooks` variable, and call `run-hooks` on it where appropriate inside `nameframe`, passing the frame as a parameter
- In `nameframe-perspective-init`, simply add a callback function to that hook variable: that callback function would simply be a renamed version of `nameframe-perspective--persp-switch-after-advice`.

For extra convenience, make a global minor mode which adds/removes that function to/from the hook when it is enabled/disabled. The user can then toggle this integration, and they can use the customize interface to enable the global mode at startup if they like.

Hope that makes sense!

john2x:
Thanks. I will look into this.

john2x:
Hello, I have made the changes to use hooks and a global minor mode to enable/disable nameframe-perspective integration ([diff](https://github.com/john2x/nameframe/commit/956fe013b3c9fa009e1dd2dba83d08bdc518c28c)).

PR #3179, opened 2015-10-05
add recipe `uimage'

lujun9972:


milkypostman:
Just some comments before merge,

- think this should be called `urlimage` as `uimage` seems a bit unclear.
- traditionally you should make some of these function private `prefixed with uimage--` where it makes sense. e.g., `uimage-no-images`, `uimage-display-images`.
- the code in general is a bit messy. it would be good to add a space before the `;;; Commentary` line.
    - The depth in the `uimage-mode-buffer` function is a bit deep.
- doesn't this require `image-file` all the time?

lujun9972:
1. I think the urlimage is too long to type. And also,because the functionality of uimage is similar with the iimage. so I think it will be better to have a similar name.
2. Actually,`uimage-no-images' and `uimage-display-images' are commands, not private functions
3. I have added a newline before ;;; Commentary line and reduced the depth of `uimage-mode-buffer' function.
4. The `image-file' now will be required all the time.

purcell:
Thanks. The default `:files` spec already includes `*.el`, so the spec here is unnecessary and should be removed from the recipe. Also, we want to build packages from development branches, not other branches which may or may not get updated regularly, so the `:branch` option should be removed from the recipe. To make "stable" releases, just tag commits with versions and they will be built in MELPA Stable.

lujun9972:
Thank you. I have removed these 2 options already.

purcell:
Thanks! Recipe added as a single commit in 346cb25, just to keep the commit history tidy.

PR #3183, opened 2015-10-07
Added recipe for evil-mc

gabesoft:
Recipe for evil-mc which provides the multiple cursors feature for evil-mode

Repo: https://github.com/gabesoft/evil-mc

### DIFF
--- a/recipes/evil-mc
+++ b/recipes/evil-mc
@@ -1,0 +1,1 @@
+(evil-mc :fetcher github :repo "gabesoft/evil-mc")


purcell:
Thanks. You should put some usage info in the `Commentary` section rather than sending people off to a README on the web -- Emacs is self-documenting, right, and should be usable offline. :-)  Plus, that info will then appear on the MELPA package page.

You might also like to add `;;;###autoload` cookies for the minor and global minor mode definitions.

gabesoft:
Thanks Steve, will do.

PR #3184, opened 2015-10-07
Add org-notify

groksteve:
Org-notify provides system notifications via notify-send for org agenda
deadlines which are active and due.

Original repository: [org-notify](https://github.com/groksteve/org-notify)

I am the package creator and maintainer and have tested the package build with melpa's current makefile.

### DIFF
--- a/recipes/org-alert
+++ b/recipes/org-alert
@@ -1,0 +1,1 @@
+(org-alert :fetcher github :repo "groksteve/org-alert")
--- a/recipes/org-notify
+++ b/recipes/org-notify
@@ -1,1 +1,0 @@
-(org-notify :fetcher github :repo "groksteve/org-notify")
--- a/recipes/org-notify
+++ b/recipes/org-notify
@@ -1,0 +1,1 @@
+(org-notify :fetcher github :repo "groksteve/org-notify")


purcell:
This is great, and there'd be nothing to stop me merging the PR right now, but can I first make a suggestion?

Instead of using `notify-send`, how about depending on and using the `alert` package, which will make your package work on all Emacs platforms, and would allow users to customise their notifications easily. See https://github.com/jwiegley/alert and let me know what you think. You might even consider renaming to `org-alert`.

purcell:
P.S. There can be multiple "done" states, so you should probably be determining them programmatically from `org`'s own settings. Same for `org-notify-todo-states`, really -- they could actually vary from one agenda file to another, and you should be able to get information from the org agenda list about whether the states of the items listed there are final or not.

purcell:
Another thought: in addition to items with deadlines, there are also items with scheduled times. It seems like `org-notify` should handle both.

groksteve:
Ok great, I wasn't aware of the `alert` package, it looks great so I'll bring it in now.

As for the TODO and DONE states, I'll have a look for a way to determine them from the agenda buffer but nothing jumped out at me on the first write.

Adding scheduled items is simple so I'll do that now too.

Thanks for the feedback, this is my first elisp package.

groksteve:
So I have made all of the above changes, renaming it to org-alert seemed like the right thing to do with the new dependency. New repo url [org-alert](https://github.com/groksteve/org-alert)

John Wiegley's alert package has been used to display notifications.
Users no longer need to specify their TODO and DONE states.
Scheduled items are notified now, not just deadlined ones.
Package doesn't clobber existing org-agenda buffers now.

purcell:
Great, thanks! Recipe added as a single commit in 2976b7f.

PR #3186, opened 2015-10-08
remove "snippets" default :files entry from haskell mode recipe

vyp:
Because neither the latest commit (on default master branch) or the
latest release has the "snippets" directory anymore.

---

I think maybe it was moved to https://github.com/AndreaCrotti/yasnippet-snippets/tree/master/haskell-mode?


purcell:
Yep, that's right - thanks a lot.

PR #3187, opened 2015-10-08
Add recipe for seeing-is-believing

jcinnamond:


purcell:
Thanks. A few little issues here:

- The .el file and package name need to match
- It's frowned upon to use `-mode` in the file/package/feature name for packages unless they provide a language major mode
- `C-c s` is [reserved for user customisations](https://www.gnu.org/software/emacs/manual/html_node/elisp/Key-Binding-Conventions.html), so you should use a different prefix
- Provide a `defgroup` in addition to your `defcustom`s so that the `seeing-is-believing` group has a docstring and is placed explicitly in the customization hierarchy

Hope that helps!

jcinnamond:
@purcell thanks for the detailed feedback. This is my first package and I appreciate you taking the time to help me out.

I've made the changes you suggested, so I think this is ready for another look.

purcell:
Yay, great, but I think `C->` is also deprecated as a minor mode prefix. You should usually pick `C-c SOMETHING` where SOMETHING has to follow the rather prescriptive rules for minor modes in the page I had linked to in a previous comment. For example, `rinari` uses `C-c ;`, so you might pick something like `C-c ,`.

jcinnamond:
Ah, I misunderstood that page. Finding emacs keybindings is one of the hard problems of computer science!

Before I push an update, would `C-c ?` work?

purcell:
> Before I push an update, would C-c ? work?

Yes, it looks like that would be fine. :-)

jcinnamond:
Excellent - I've pushed the change.

purcell:
Super, thanks - merged!

jcinnamond:
Hooray - thanks again for all your help with this.

purcell:
Happy to help.

PR #3188, opened 2015-10-08
Recipe: git-commit-insert-issue

vindarel:
Hi !

This package is to get a helm list of (github) issues when we type "Fixes #" in a commit message.

https://gitlab.com/emacs-stuff/git-commit-insert-issue (I'm the maintainer)

The recipe compiles and I can require it afterwards, but two notes:

- the list-package doesn't show my package description. I did put it in the first line and in a `;;; Summary: ` but this doesn't seem to be enough;
-  I rely on a package that isn't on melpa ([github-issues](https://github.com/inkel/github-issues.el)), so I included the source which I load with this code:

```
(let ((load-path (cons nil load-path)))
  (load-file "github-issues.el")) ;; not in MELPA, local copy.
```
It seems ok now since I can require it, but you may want to confirm.

Regards

### DIFF
--- a/recipes/git-commit-insert-issue
+++ b/recipes/git-commit-insert-issue
@@ -1,0 +1,1 @@
+(git-commit-insert-issue :fetcher gitlab :repo "emacs-stuff/git-commit-insert-issue")


milkypostman:
1. your header, `Author:`, `Keywords:` needs to use two `;`  and not three.  Enable `auto-fill-mode` then edit a new `.el` file and you can see the proper layout.

2. once you fix 1 you will likely need to change `package-requires` to,

        ((helm 0), (projectile 0), (s 0))

    I'm almost positive a version is required for this.

3. Licence should be spelled License or just removed.

4. You will need `;;;###autoload` functions in `git-commit-insert-issue.el` so that after installation *some* function is available to the user without needed to `require` the whole package.

5. all your functions should start with the `git-commit-insert-issue` prefix instead of having some that start with `issues` and some that start with `github-close` etc.

vindarel:
Oh thx, I was looking for `autoload`. 

That should be it: https://gitlab.com/emacs-stuff/git-commit-insert-issue/commit/63cd52d2098f2a9772093e131889a83726c2857a  

(I tried a bit with the `names` package but I'll stay simple and secure for now)

purcell:
No, the version numbers in the `Package-Requires` line need to be double-quoted. Please consider using something like `flycheck-package` to verify this for you, or at least `M-: (package-buffer-info)`.

vindarel:
`package-buffer-info` was very useful (I had tried `flycheck-package` with no result), you could add it into melpa's readme :)

hope it's ok with this commit: https://gitlab.com/emacs-stuff/git-commit-insert-issue/commit/640b514dc5d6911f1f50d67eb5a09f2e1d3d1801

purcell:
> I had tried flycheck-package with no result

`flycheck-package` currently has the weakness that it doesn't get enabled if the .el file starts off without any package-specific headers, ie. `;; Package-Version:` or `;; Package-Requires:`.

purcell:
If you try it again now and it doesn't work, please file an issue: https://github.com/purcell/flycheck-package

PR #3189, opened 2015-10-08
Add recipe for cargo

AtticHacker:
Hi!

https://github.com/AtticHacker/cargo.el
This package provides a few keybindings to manage the Cargo utility. Cargo is a package manager for Rust. If there's any extra info I need to provide or if I need to change anything in the repo let me know!

### DIFF
--- a/recipes/cargo
+++ b/recipes/cargo
@@ -1,0 +1,1 @@
+(cargo :repo "attichacker/cargo.el" :fetcher github)


milkypostman:
the only thing that I would do is rename it from `cargo-mode` to `cargo-minor-mode`. `yasnippet` does `yas-minor-mode` just to let you know that when you enable it you're only enabling the minor mode. i guess `evil-mode` is minor in a sense. I dunno.

I also think it's a bit weird that you have cl-defun in there just for the keyword argument. i get why you did it but I don't see it as a *huge* advantage. if you want to leave it thats fine but you could reduce the dependency.

AtticHacker:
Hi!
Thanks for the feedback, I've made some changes based on your comment. Here's the diff:
https://github.com/AtticHacker/cargo.el/commit/6661bb96bb284f243228f8e1fa585e84dc5d3d9e

If there's anything else let me know!

PR #3191, opened 2015-10-09
Add Bogdanp/ido-clever-match

Bogdanp:
An alternative matcher for ido.

https://github.com/Bogdanp/ido-clever-match

I am the maintainer.

### DIFF
--- a/recipes/ido-clever-match
+++ b/recipes/ido-clever-match
@@ -1,0 +1,1 @@
+(ido-clever-match :repo "Bogdanp/ido-clever-match" :fetcher github)


purcell:
Which `subr-x` functions are you using? It's possible you might need to depend on a specific newer version of Emacs.

Bogdanp:
@purcell I'm using `hash-table-keys` from `subr-x`

purcell:
In fact, since you're using the new advice system, you'll need to depend on Emacs 24.4 anyway.

Bogdanp:
I bumped the requirement to `24.4`. I didn't realize `advice-{add,remove}` were only introduced that recently. Thanks!

purcell:
In time I want to make `flycheck-package` detect and warn about things like that.

PR #3193, opened 2015-10-10
add recipe for cliphist

redguardtoo:
Read clipboard history from Parcellite on Linux and Flycut on OS X.

### DIFF
--- a/recipes/cliphist
+++ b/recipes/cliphist
@@ -1,0 +1,2 @@
+(cliphist :fetcher github
+                :repo "redguardtoo/cliphist")


purcell:
Thanks. Since you've provided a `-pkg.el` file with no `popup` package dependency in it, the `Package-Requires` header in `cliphist.el` will have no effect. Either list the dependency in the `-pkg.el` file, or delete that file because MELPA doesn't need it anyway. Looks okay otherwise.

redguardtoo:
Thanks for review, I removed the *-pkg.el

PR #3194, opened 2015-10-10
Add darktooth-b theme package

bryangarza:


purcell:
Thanks. You haven't provided any information about your package, but it looks like you've forked another theme and then just submitted a recipe with a different name -- that's not going to work, because the package name has to match up with the `.el` file name(s), and there's no `darktooth-b-theme.el`. Plus, just adding a `b` with no description of what's different isn't very compelling.

So what you should probably do is choose a new name for your theme, rename all the files and symbols accordingly, and submit a fresh recipe.

Issue #3192, opened 2015-10-09
idea: add "Next build in... " information on website


Silex:
Hello,

Currently I see on http://melpa.org that `Last build ended: 3 hours ago`.

It'd be nice if we had `Next build in: ~53 minutes` information somewhere, even if incorrect.

Simply store the last build time somewhere, and then do some math extrapolation to guess when the current build is going to finish.

Actually, the information could be like this:

```
Last build ended: $(now - last_build_end) hours ago
Current build ends at: $(current_build_start + (last_build_end - last_build_start))
Next build starts at: $(current_build_start + (last_build_end - last_build_start) + 30)
```

You get the idea :)

Silex:
Closing for the PR

purcell:
`[0-9]+` should always work, I guess.

PR #3195, opened 2015-10-10
Add build start information

Silex:
Following #3192.

This is not finished but is enough to start discussing it :)

I need to check wether `\d` is supported in sed (probably not), and then I need to add the extrapolations.

### DIFF
--- a/html/index.html
+++ b/html/index.html
@@ -48,6 +48,7 @@
   <![endif]-->
   <script src="//cdnjs.cloudflare.com/ajax/libs/mithril/0.2.0/mithril.min.js"></script>
   <script src="//cdnjs.cloudflare.com/ajax/libs/moment.js/2.8.4/moment.min.js"></script>
+  <script src="//cdnjs.cloudflare.com/ajax/libs/moment-duration-format/1.3.0/moment-duration-format.min.js"></script>
   <script src="//cdnjs.cloudflare.com/ajax/libs/lodash.js/2.4.1/lodash.min.js"></script>
   <script src="//cdnjs.cloudflare.com/ajax/libs/Cookies.js/1.1.0/cookies.min.js"></script>
   <script src="js/melpa.js"></script>
--- a/html/js/melpa.js
+++ b/html/js/melpa.js
@@ -444,15 +444,54 @@
 
   melpa.buildstatus = {};
   melpa.buildstatus.controller = function() {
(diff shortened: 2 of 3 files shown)


purcell:
`[0-9]+` should always work, I guess.

Silex:
@purcell: is `service/builder` the right file to add this info? The same logic seems duplicated in several other files.

milkypostman:
Yeah I think it's fine where you have it. I want to rereview this when I'm on a laptop but I think it looks OK and we can probably merge it in.

Silex:
A'ight, I'll summon you when it's all done.

Silex:
@milkypostman, @purcell: okay, I think I'm done. I didn't test, do you have an easy way to test this?

purcell:
One thing to beware of is that this will temporarily break the MELPA site because the newly-added info won't be available in the build status json until the first new build completes. So perhaps add a quick snippet in the JS to handle it. I'll probably tidy up the JS a bit afterwards anyway.

Silex:
Oh, good catch. I'll do that.

Silex:
Alright, the part to remove when `build-status.json` is correct is the `FIXME` if block

purcell:
Hey, thanks - your work here was really helpful. In the end I've currently committed some more minimal changes based on yours. I don't yet handle remembering the duration between builds, but I'll probably do that in due course. The main difference between your code and mine is that I'm trying to keep the display of the information more compact on the page. I also had to update the MELPA Stable build script to make sure that keeps working. Guess we'll find out in a couple of hours what (if anything) I've broken. :-)

milkypostman:
@Silex thank you so much for the help! i think you should see the fruits of your labor now on the site. I fixed some stupid things I had done on the server which made your change get delayed (require manual intervention) and future changes should resolve this issue.

purcell:
Yep. I plan to handle calculating durations so that we can display not just "Current build started: 2 hours ago", but "Current build started: 2 hours ago (last took 130 minutes" or something similar.

Silex:
:+1: 

Nice to see alternative implementation, I like the `update_json` idea. Glad the idea got through :)

purcell:
It's really helpful to be able to see at a glance more information about the build. I'm a bit surprised that the build time is currently 4-5 hours, actually.

milkypostman:
I am too. Will have to dig in further.

PR #3196, opened 2015-10-10
added empos

dimalik:
Locates academic papers online using pyopl.py
https://github.com/dimalik/empos
Author and maintainer

### DIFF
--- a/recipes/empos
+++ b/recipes/empos
@@ -1,0 +1,1 @@
+(empos :repo "dimalik/empos" :fetcher github)


purcell:
Thanks. Quick feedback:

- You should add a `defgroup` to place the custom items correctly in the customization hierarchy, and to give the group a description
- `pyopl-path` should be `empos-pyopl-path` -- see the [elisp coding conventions](https://www.gnu.org/software/emacs/manual/html_node/elisp/Coding-Conventions.html). Similarly, `visual-line-line-range` is not a valid symbol name in the context of this package.
- Using `(toggle-truncate-lines)` seems wrong. Either turn it on or off.
- The body of a `define-minor-mode` form will be called both when the mode is enabled and when it is enabled, so it should generally contain an `(if this-minor-mode ... ...)` form, where the "else" clause restores the previous state of the buffer

Hope that helps!

-Steve

dimalik:
hey thanks, that was indeed very helpful!
I remember i had that problem with the minor mode -hope it's fixed now (the rest were due to sloppiness I guess).
I've pushed a more recent commit, should I make another pull request?

best,

d

purcell:
Nope, that's fine - I'll merge this. A bit of extra feedback:

- You should really put the `;;;###autoload` above `empos-search` rather than `empos-mode`, because the former is what users are expected to call.
- It would be neater IMO to just call `(error ...)` when the minor mode is called in the wrong buffer

PR #3197, opened 2015-10-10
Add recipe for popup-keys

vermiculus:
At least starts work on QBobWatson/popup-keys#2

### DIFF
--- a/recipes/popup-keys
+++ b/recipes/popup-keys
@@ -1,0 +1,1 @@
+(popup-keys :repo "QBobWatson/popup-keys" :fetcher github)


purcell:
Looks good. The only thing which makes me sad is that the identifiers use the `popup-keys:` prefix instead of the proper `popup-keys-`. I wish everyone would follow the community's established [coding conventions](https://www.gnu.org/software/emacs/manual/html_node/elisp/Coding-Conventions.html) instead of inventing their own conventions or borrowing them from CL or Clojure. Maybe it'd be possible to change that?

vermiculus:
It may be. @QBobWatson, do you have thoughts on this? It needs to come from your repository; it's MELPA's policy (last I checked) to not accept forks except in special circumstances.

QBobWatson:
I'm glad someone else is finding popup-keys useful :)  There's actually a reason that I've made no effort to make the package easy to find.  Obviously popup-keys is a rewrite of magit's magit-key-mode.  After writing it in the summer of 2014, I contacted Jonas Bernoulli, a Magit maintainer, who made it clear that he was planning to update magit-key-mode himself and refactor it out as a separate package.  I have no idea if he's actually done so, but if he has or will, then I have no interest in maintaining a competing package.  Still I'd already made the git repository, and I saw no reason to take it down.

What exactly is the policy on forks?  If necessary I can give @vermiculus write access to the popup-keys repo; if he wants to change coding conventions in another branch, that's fine with me.

vermiculus:
Hmm, @tarsius – I know you've been very busy, but has this 'magit-key-mode' been in the back of your mind? It seems to have been removed (or renamed) in later versions of magit.

alezost:
`magit-key-mode` was replaced by `magit-popup` library.  It is a part of
Magit (since 2.1.0) and is available on MELPA.  And AFAIK there is a
plan to make a separate general-purpose `transient` library based on
`magit-popup`.

purcell:
Ah, yes, indeed: http://melpa.org/#/magit-popup

Does that mean we should skip adding this to MELPA? Seems so.

vermiculus:
Yeah, I'd agree. magit-popup is more likely to be active going forward and besides, I think it's time emacs as a community starts trying to have *fewer* ways of doing things :smile:

purcell:
:+1:

PR #3198, opened 2015-10-11
Add a recipe for  name-this-color.el

knl:
Hi,

I would like to add a recipe for my very first emacs package, name-this-color.

Thanks,
Nikola

### DIFF
--- a/recipes/name-this-color
+++ b/recipes/name-this-color
@@ -1,0 +1,1 @@
+(name-this-color :repo "knl/name-this-color.el" :fetcher github)


purcell:
Thanks! Quick feedback:

- You should consistently prefix all the symbols to match the feature name, so instead of `ntc-...` use `name-this-color-...`. See the [elisp coding conventions](https://www.gnu.org/software/emacs/manual/html_node/elisp/Coding-Conventions.html) for more on this.
- Similarly, `shade-this-color` should be `name-this-color-shade` or similar

Hope that helps!

-Steve

knl:
Hi Steve, thanks for the feedback. I went through the doc and it is a nice read. I'll switch to using `ntc-` as the prefix, since `name-this-color-` is way to long ("You should choose a short word to distinguish your program from other Lisp programs. ").

purcell:
You'd then have to rename the whole library too, so that everything matched. I'd suggest just using `name-this-color-` as the prefix. Bytes are cheap, and this is a small library.

knl:
Thanks for the pointers, they are quite valuable! I updated the API so that everything has a common prefix `ntc-`. However, as I like the name `name-this-color`, I'll keep that name of the project. The guidelines do not say anything about matching the name of the project/library with the prefix.

Is there something else I should do in order to get this recipe in MELPA?

purcell:
You're entitled to interpret the page as you like, but I've made the same suggestion twice now, and as a first-time package author you might want to consider whether or not it makes sense to ignore it. :-) Nonetheless, I'm merging this now in the interest of time.

PR #3201, opened 2015-10-12
Add dionysos recipe

nlamirault:
Signed-off-by: Nicolas Lamirault <nicolas.lamirault@gmail.com>


purcell:
You have not tested this locally, I think, because the recipe is invalid as described by the travis build failure. There are instructions for doing so in the README. Also, no `:files` spec is needed here: the default includes `"*.el"`, so you should remove the `:files` spec entirely.

nlamirault:
I update recipe.

purcell:
Thanks. Applied in a single commit in 890445e.

PR #3202, opened 2015-10-13
Add helm-flx

PythonNut:
This package adds `flx`-based sorting to `helm`.

* [PythonNut/**helm-flx**](https://github.com/PythonNut/helm-flx)

### DIFF
--- a/recipes/helm-flx
+++ b/recipes/helm-flx
@@ -1,0 +1,2 @@
+(helm-flx :repo "PythonNut/helm-flx"
+          :fetcher github)


purcell:
Thanks. The build failure here was spurious.

I was reading in the [elisp coding conventions](https://www.gnu.org/software/emacs/manual/html_node/elisp/Coding-Conventions.html) that we should avoid `eval-after-load` inside libraries, so you might like to see if there's a better way to handle that part of the code.

PR #3204, opened 2015-10-14
Add editorconfig-core recipe

10sr:
[EditorConfig](http://editorconfig.org/) core library written purely in Emacs Lisp.

URL: https://github.com/10sr/editorconfig-core-emacslisp

### DIFF
--- a/recipes/editorconfig-core
+++ b/recipes/editorconfig-core
@@ -1,0 +1,1 @@
+(editorconfig-core :repo "10sr/editorconfig-core-emacslisp" :fetcher github)


purcell:
Thanks. Package dependencies (`;; Package-Requires ...`) should be in the main `.el` file in order to be detected, so you'll need to move that line from the `-handle` file, and also add the missing dependency on `(cl-lib "0.5")`. Looks good otherwise.

10sr:
Thank you for your review. :point_up: I fixed them.

purcell:
Thanks - merged!

PR #3207, opened 2015-10-16
Add recipe for mu4e-alert

iqbalansari:
This adds recipe for `mu4e-alert`. I am the author of the package. The code lives in https://github.com/iqbalansari/mu4e-alert  

The package provides desktop notifications for unread emails in `mu4e` (`mu` to be more accurate). A dependency is declared on Emacs v24.1, since the package uses lexical binding. A dependency has not been declared on `mu4e` since it is not available on MELPA.

I tried building and installing the package locally using `package-build`.

Thanks

### DIFF
--- a/recipes/mu4e-alert
+++ b/recipes/mu4e-alert
@@ -1,0 +1,1 @@
+(mu4e-alert :fetcher github :repo "iqbalansari/mu4e-alert")


purcell:
Thanks. Mostly looks good. Your code shouldn't really [try](https://github.com/iqbalansari/mu4e-alert/blob/master/mu4e-alert.el#L91-L93) to intelligently choose a default notification style: just use `alert`'s default unless the user overrides it.

[Redefining missing core functions](https://github.com/iqbalansari/mu4e-alert/blob/master/mu4e-alert.el#L107-L126) is also something you should avoid. Refer to the [elisp coding conventions](https://www.gnu.org/software/emacs/manual/html_node/elisp/Coding-Conventions.html), which discuss this - search for `gnus-point-at-bol` on that page.

Hope that helps!

iqbalansari:
> Redefining missing core functions is also something you should avoid. Refer to the elisp coding conventions, which discuss this - search for gnus-point-at-bol on that page.

Thanks for the heads up, this has been fixed, I now use `s.el`

> Your code shouldn't really try to intelligently choose a default notification style: just use alert's default unless the user overrides it.

Since the package advertises itself as providing **desktop** notifications, I felt it would be better if it provided desktop notifications out of the box rather than having the user configure it. As such that setting affects only the notifications from `mu4e-alert`. I am okay with changing it if you still believe its not a good idea. As always thanks for your feedback

purcell:
> Since the package advertises itself as providing desktop notifications, I felt it would be better if it provided desktop notifications out of the box rather than having the user configure it. As such that setting affects only the notifications from mu4e-alert. I am okay with changing it if you still believe its not a good idea. As always thanks for your feedback

I really think it would be better to just use the `alert` default, and invite users to customize it. This way users will learn about `alert` for themselves, and you will not get bug reports about platform-specific issues. For example, it looks like your current code would not correctly detect the availability of the `terminal-notifier` helper on OS X -- that's just not something you should have to handle in your code! :-)

purcell:
Another fun fact: `(featurep 'notifications)` returns `t` on OS X, but then doesn't actually work... sigh. :-)

iqbalansari:
> For example, it looks like your current code would not correctly detect the availability of the terminal-notifier helper on OS X -- that's just not something you should have to handle in your code!

I was planning to deal with platform specific issues by adding something like "works on Linux (Ubuntu), pull requests for other platforms welcome" to the README but I see your point. I have updated the code to not try to be "clever". Thanks for you help and sorry for taking up your valuable time :smile:

purcell:
> Thanks for your help and sorry for taking up your valuable time :smile:

No problem. :-)  Merging now.

iqbalansari:
Thanks

PR #3209, opened 2015-10-17
Added django-manage-mode recipe

gopar:
Recipe for `django-manage-mode`. A minor mode for controlling `manage.py` which is what every django project uses.

https://github.com/gopar/django-manage-mode

I am the maintainer.

Let me know if I am missing anything, or have something out of place.

### DIFF
--- a/recipes/django-manage
+++ b/recipes/django-manage
@@ -1,0 +1,1 @@
+(django-manage :fetcher github :repo "gopar/django-manage")


purcell:
Thanks. A few points of feedback here:

- Name the file & package `django-manage`, not `django-manage-mode` -- the `-mode` suffix in package names is generally considered redundant unless the package is a language major mode
- Do not add `setup-django-manage-mode` to `python-mode-hook`: that is the user's job, and it particularly shouldn't be done via an autoload.
- Look at the built-in function `locate-dominating-file`: it will help you to tidy up `django-manage-root`. You don't need the regexp part of the code there, either.
- Please provide a `Commentary` section in the elisp headers to explain at a high level about how to use the package.
- Don't ever call `cd` or `cd-absolute`: instead, use `(let ...)` to set `default-directory` within a block of code. It will have the same effect, but won't cause strange side-effects
- `hydra-django-manage` should probably be called `django-manage-hydra`\
- Much of your code will be unsafe or broken if there are spaces in the current directory path name, e.g. the `compile` command. Look for opportunities to use `shell-quote-argument`
- Use `M-x checkdoc` to get some help writing compliant docstrings for your functions

Hope that helps!

-Steve

gopar:
Closing this pr and opening a new one with `django-manage` recipe. Also with updated code. Thanks for the feedback :)

PR #3211, opened 2015-10-17
Add magic-filetype

zonuexe:
https://github.com/zonuexe/magic-filetype.el

### DIFF
--- a/recipes/magic-filetype
+++ b/recipes/magic-filetype
@@ -1,0 +1,1 @@
+(magic-filetype :fetcher github :repo "zonuexe/magic-filetype.el")


purcell:
Thanks. All the top-level symbols should have the `magic-filetype-` prefix, so you should rename all the functions accordingly. Looks good otherwise.

zonuexe:
I followed your advice. https://github.com/zonuexe/magic-filetype.el/commit/cd0ba200f9c176557d1c20aefa9930a4ee88b015

But, `major-mode-of` is like DSL, it is shorter looks good.
So I had to define the alias, do you think?

purcell:
> I followed your advice. zonuexe/magic-filetype.el@cd0ba20

As per the conventions, the prefix should be `magic-filetype-`, not `magic-filetype:` -- it's not Common Lisp. :-)

> But, major-mode-of is like DSL, it is shorter looks good.
> So I had to define the alias, do you think?

Nope, still not valid. A good name for that function would be something like `magic-filetype-major-mode`. Put an alias in your personal config if you prefer it.

zonuexe:
Thank you for reply!

https://github.com/zonuexe/magic-filetype.el/commit/d97e61f9c9f586593080f6eba77b3742b5d70b49
https://github.com/zonuexe/magic-filetype.el/commit/338c260bbc23662078b014ccf580d3122df2acdc

purcell:
Super, thanks - looks great.

P.S. is `magic-filetype-reload-major-mode` just equivalent to calling `normal-mode`?

PR #3215, opened 2015-10-19
Add recipe for psc-ide emacs integration

rexim:
Recipe for Emacs integration with [psc-ide](https://github.com/kRITZCREEK/psc-ide)

### DIFF
--- a/recipes/psc-ide
+++ b/recipes/psc-ide
@@ -1,0 +1,4 @@
+(psc-ide :fetcher github
+         :repo "epost/psc-ide-emacs"
+         :files (:defaults
+                 (:exclude "purescript-extras.el")))


rexim:
cc @epost, @kRITZCREEK

purcell:
The recipe and package name here are not valid: they should match the .el file name, ie. `psc-ide`, not `psc-ide-emacs`.

Re. `psc-ide.el`:
- you should add a short `Commentary` section in the header so that the user will have some information about how to use the package
- don't put `)` on trailing lines - see the [standard elisp conventions](https://www.gnu.org/software/emacs/manual/html_node/elisp/Coding-Conventions.html), which are there to benefit the whole community

re. `psc-ide-protocol.el`:
- All the symbols must have a common package-specific prefix, e.g. `psc-ide-` -- defining very generic-looking symbols like `unwrap-result` and `filter-exact` is a big no-no. :-)

And for all the .el files, the `provide` form should be at the bottom of the file, not the top.

Hope that helps!

rexim:
@purcell thank you very much for your review! I'll try to fix all of these. :)

rexim:
@purcell all the remarks are fixed. Please take a look when you get a chance!

epost:
Thanks @rexim and @purcell!

purcell:
Nice work, thanks.

I'd suggest that you make sure each file will byte-compile in isolation: I suspect that https://github.com/epost/psc-ide-emacs/blob/master/psc-ide-backported.el would need a `(require 'cl-lib)` at the top to avoid warnings, for example. `flycheck` should tell you some of this stuff while you're working on the code, btw.

rexim:
@purcell thank you sir! Really appreciate your help! I'll check the isolated compilation of the files.

P.S. didn't know about flycheck, thanks! I certanly have to add this to my toolbox. :)

purcell:
There's also my `flycheck-package`, for when you're working on elisp packages. :-)

PR #3217, opened 2015-10-20
Add forecast.el

cadadr:
Forecast.el prepares weather reports from Forecast.io data.

### DIFF
--- a/recipes/forecast
+++ b/recipes/forecast
@@ -1,0 +1,1 @@
+(forecast :fetcher github :repo "cadadr/forecast.el")


purcell:
Looks good, thanks. I wonder if it might make sense to use `special-mode` as the parent mode rather than `fundamental-mode`?

cadadr:
Thanks for the merge. I'll look into special-mode as soon as I have time.

purcell:
> I'll look into special-mode as soon as I have time.

Yeah, it's basically for the things like `help-mode`: it gives you the `q` binding for free, and it's the standard parent mode in many cases dealing with temporary buffers unconnected to files.

cadadr:
Then I'll have to switch to it ASAP. Thanks.

PR #3218, opened 2015-10-21
Add recipe for magit-rockstar

tarsius:
https://github.com/tarsius/magit-rockstar/blob/master/magit-rockstar.el#L25

### DIFF
--- a/recipes/magit-rockstar
+++ b/recipes/magit-rockstar
@@ -1,0 +1,1 @@
+(magit-rockstar :fetcher github :repo "tarsius/magit-rockstar")


purcell:
Thanks - P.S. you've got two `;;; magit-rockstar.el ends soon` lines.

tarsius:
No, there is one "ends soon" and one "ends here" line :-)

That was once intended as a commentary on the convention of putting "eof" at eof, but then I started doing it for all my packages.

purcell:
Aha - missed that. :-)

PR #3219, opened 2015-10-21
Adding recipe for gregorio-mode

cajetanus:
Just adding a recipe for gregorio mode, available on github.

### DIFF
--- a/recipes/gregorio-mode
+++ b/recipes/gregorio-mode
@@ -1,2 +1,2 @@
 (gregorio-mode :fetcher github
-	       :repo "cajetanus/gregorio-mode")
+	       :repo "cajetanus/gregorio-mode.el")
--- a/recipes/gregorio-mode
+++ b/recipes/gregorio-mode
@@ -1,0 +1,2 @@
+(gregorio-mode :fetcher github
+	       :repo "cajetanus/gregorio-mode")


recipes/gregorio-mode
```diff
@@ -0,0 +1,2 @@
+(gregorio-mode :fetcher github
+	       :repo "cajetanus/gregorio-mode")
```
syohex: Typo ? `gregorio-mode` -> `gregorio-mode.el`.

cajetanus:
Sorry about that, forgot the .el at the end.

syohex:
Your package does not follow Emacs packaging convention.

Please see

Example
- https://github.com/milkypostman/melpa/blob/master/package-build.el#L1-L40 (header part)
- https://github.com/milkypostman/melpa/blob/master/package-build.el#L1427 (footer part)

purcell:
Yep. Additional feedback:

- "C-c u" is reserved for users (see the [keybinding conventions](https://www.gnu.org/software/emacs/manual/html_node/elisp/Key-Binding-Conventions.html)) so you can't use it here
- Consider adding `;;;###autoload` cookies for the major mode definition form and the form which extends `auto-mode-alist`.

When fixed, please open a fresh PR with a single recipe commit. Note that you can test it locally before submitting by following the directions in the README. :-)

PR #3220, opened 2015-10-21
Add go-rename

kostya-sh:
go-rename provides integration of the 'gorename' tool into Emacs.

Link to the package file in the official repository:
https://go.googlesource.com/tools/+/master/refactor/rename/go-rename.el

I am not the author of the package but use it quite a lot. I've also submitted
the following package.el compatibility changes for go-rename:
https://go-review.googlesource.com/#/c/15527/

### DIFF
--- a/recipes/go-rename
+++ b/recipes/go-rename
@@ -1,0 +1,4 @@
+(go-rename
+ :fetcher git
+ :url "https://go.googlesource.com/tools"
+ :files ("refactor/rename/go-rename.el"))


purcell:
Great, thanks. You might also like to submit a patch adding an `;;;###autoload` cookie above the `(defun go-rename (...` form, so that the command can be used without needing to explicitly `(require 'go-rename)` first.

kostya-sh:
Thank you for the suggestion. Will submit another patch soon.

PR #3224, opened 2015-10-22
Add ido-grid-mode

larkery:
ido-grid-mode is a package which displays ido prospects in the
minibuffer in a grid, a little like how zsh completions are displayed.

The package is at https://github.com/larkery/ido-grid-mode.el

I am the author and maintainer of ido-grid-mode.

### DIFF
--- a/recipes/ido-grid-mode
+++ b/recipes/ido-grid-mode
@@ -1,0 +1,1 @@
+(ido-grid-mode :fetcher github :repo "larkery/ido-grid-mode.el")


purcell:
Thanks, looks nice! A few small issues here:

- `nadvice.el` was only added in Emacs 24.4, so the package dependency needs to be on that Emacs version
- Consider using `cl-lib` instead of the deprecated `cl`. It'll probably just involve prefixing some function/macro invocations with `cl-`
- Try to use a consistent prefix for identifiers. `igm-` is much more likely to clash with another package and `ido-grid-`, and it's not worth the risk for the number of characters saved.
- It's conventional to avoid using the `-mode` suffix for packages unless they are language major modes, but I'm not going to push on this point.

Hope that helps!

larkery:
I have changed all except `-mode` :)

Thanks for the advice!

Silex:
Oh my god! Someone actually somewhat finishing what I started at https://github.com/creichert/ido-vertical-mode.el/pull/29

@larkery: 

``` elisp
(setq ido-grid-mode-max-columns 1)
(setq ido-grid-mode-max-rows 12))
```

That's *almost* the behavior I'd like... except hen going down the list I'd like to have the list scroll when you hit the bottom instead of it "jumping up" like it does now. Do you think it's easy to change? Here's a gif of what I mean: https://github.com/Silex/ido-vertical-mode.el/blob/master/screenshot-static.gif

larkery:
I think you might be able to wangle the rows and columns as you have done and then modify the functions bound to the keys in a hook of your own. If you have a look you'll see that at the moment when you 'overflow' it does a page down.

This is because with more than one column scrolling by one element at a time looks weird, and I don't want to have to layout columns you can't actually see so as to scroll all columns at once. With one column it looks OK though.

larkery:
```elisp
(defun ido-grid-mode-next-page () (interactive) 
   (call-interactively #'ido-next-match)
   (setq ido-grid-mode-offset (- ido-grid-mode-count 1))
```
might do the job for scroll down, and something analogous for scroll up.

larkery:
Just checked and
```elisp

(defun ido-grid-mode-next-page ()
  (interactive) 
  (call-interactively #'ido-next-match)
  (setq ido-grid-mode-offset (- ido-grid-mode-count 1)))

(defun ido-grid-mode-previous-page ()
  (interactive)
  (call-interactively #'ido-prev-match)
  (setq ido-grid-mode-offset 0))
```

Seems to do the trick. You could modify/advise `ido-grid-mode-move` instead if you want to keep the ability to page up and down in other circumstances

larkery:
My arrow doesn't move up and down either I'm afraid.

Silex:
@larkery: excellent! thank you for the example

Silex:
@larkery: having the arrow scroll down seems pretty simple to have given https://github.com/larkery/ido-grid-mode.el/blob/master/ido-grid-mode.el#L359

I'll try to make a PR with an option for it.

larkery:
OK - I'll look at putting in a different option for the scrolling behaviour.

larkery:
Just committed some changes to do the thing you would like:
```elisp
(setq ido-grid-mode-max-columns 1
       ido-grid-mode-prefix-scrolls t
       ido-grid-mode-scroll-down #'ido-grid-mode-next-row
       ido-grid-mode-scroll-up #'ido-grid-mode-previous-row)
```

Silex:
@larkery: fantastic! many thanks

purcell:
The `(cl-lib)` package dependency is malformatted and invalid, because the version number is mandatory. Normally you'd need `(cl-lib "0.5")`, but since you're already depending on emacs 24.4 you can just drop that dependency.

larkery:
I do apologise - I had somehow turned off flycheck for packages by mistake and didn't see it.

This is fixed now.

purcell:
No problem, thanks!

PR #3225, opened 2015-10-22
Add paper-theme.

cadadr:
Paper is a simple little theme.

### DIFF
--- a/recipes/paper-theme
+++ b/recipes/paper-theme
@@ -1,0 +1,1 @@
+(paper-theme :fetcher github :repo "cadadr/paper-theme")


purcell:
A few issues here:

- The first line is invalid. Needs `;;;` and `---`, not `;;` and `--`: use `flycheck-package` to check this, or `M-: (package-buffer-info)`
- The `Package-Requires` line is invalid: each entry is a `(name "number")` list, not a `(name . "number")` cons cell
- Don't set the font family on behalf of the user - this is inconsiderate, and users can always enable `buffer-face-mode` in org buffers to get sans-serif fonts
- You're not only loading all of `org` when the theme is loaded (that's bad!), but you're also setting `org-hide-leading-stars` at a global level, which a theme has no business doing. :-)

Hope that helps,
-Steve

cadadr:
Thanks for the feedback. My bad not cleaning it up before exporting from my repo, sorry. I've now pushed some commits that target the issues you point to. I may update if there are any other mistakes that I haven't noticed.

purcell:
:+1: Looks great now, thanks.

cadadr:
:+1:  Thanks

PR #3228, opened 2015-10-22
Add recipe for company-qml

cute-jumper:
company-qml is a company-backend for QML files. It is usable although its implementation is very naive at the moment.

https://github.com/cute-jumper/company-qml


purcell:
Thanks!

PR #3230, opened 2015-10-23
Limit to main file; ignore test files

thisirs:


purcell:
Can I suggest instead that you rename `test-state.el` to `state-test.el` or `state-tests.el`? We support that convention already, and exclude such files from packages by default.

thisirs:
Oh! I didn't know... Nice! Will do that ;)

PR #3233, opened 2015-10-24
Add unicode-troll-stopper

camsaul:
Hi,
This is a PR to add my [unicode-troll-stopper](https://github.com/camsaul/emacs-unicode-troll-stopper) package. `unicode-troll-stopper` is a simple minor mode that adds font-locking for Unicode homoglyphs like "Greek Question Mark" (`;`), which look like normal ASCII symbols but cause very frustrating syntax errors.

### DIFF
--- a/recipes/unicode-troll-stopper
+++ b/recipes/unicode-troll-stopper
@@ -1,0 +1,2 @@
+(unicode-troll-stopper :fetcher github
+                       :repo "camsaul/emacs-unicode-troll-stopper")


purcell:
This looks handy, but you don't need `dash` -- check out the built-in `regexp-opt` function. Also, the `.el` is missing in the filename on the first line. If the code needs `lexical-binding`, you should declare a dependency on `(emacs "24")`. I don't think it does, though, so you should probably remove the `lexical-binding: t` on the first line.

syohex:
`font-lock-flush` and `font-lock-ensure` are implemented only in development Emacs. Even Emacs 24.5 cannot use them. I suppose `font-lock-fontify-buffer` is available instead of them(however it causes byte-compile warning on develepment Emacs).

camsaul:
Hey guys, thanks for the feedback. I implemented the changes you suggested.

The docs for `font-lock-fontify-buffer` say it is intended for interactive use only so I switched it to 
```emacs-lisp
(font-lock-fontify-region (point-min) (point-max))
```

which works for me & doesn't cause the byte compiler to complain.

purcell:
Thanks!

PR #3234, opened 2015-10-24
Update recipe for emms-player-mpv-jp-radios

momomo5717:
Added a new directory to emms-player-mpv-jp-radios.

I might modify directories in future.
So now I think `"*/*.el"` could be better.

Thank you.

### DIFF
--- a/recipes/emms-player-mpv-jp-radios
+++ b/recipes/emms-player-mpv-jp-radios
@@ -1,11 +1,2 @@
 (emms-player-mpv-jp-radios :fetcher github :repo "momomo5717/emms-player-mpv-jp-radios"
-                           :files (:defaults
-                                   "agqr"
-                                   "anitama"
-                                   "hibiki"
-                                   "listen"
-                                   "onsen"
-                                   "radiko"
-                                   "radiru"
-                                   "saimaru"
-                                   "seaside"))
+                           :files (:defaults "*/*.el"))


milkypostman:
it's not terribly hard to update this and i'm not sure what we gain by the wildcard here. why do you think you would change the layout again?

momomo5717:
Thank you for the reply.

I sent similar PR several times.
So I think updating it many times would be trouble for you
and supposed the wildcard could be better.

Would it be better to specify directory names?
Sorry for my lack of knowledge.

Thank you.

milkypostman:
i guess my questioning was more in thinking about why the complexity and moving things etc. it just seems like this doesn't come up very much. why would you move the files to a subdirectory?

also, the problem is this, when you include files from subdirectories, they will get moved to the "root" directory of the package. so actually this will be incorrect.

purcell:
Yes, so in summary: the wildcard pattern equivalent to the old recipe would be `"*"`, not `"*/*.el"`, but then that would also match non-directories. Really the best solution here would be to either put all your plugin files or directories inside a single top-level `plugins` directory, or (better) to just keep all of your elisp at the top level of the repo.

momomo5717:
Thank you for the replies.

These subdirectories is used to be easy to distinguish radio station names at the repo.
So I think it is not a problem that the files move to the root directory of the package when it is installed from melpa.
and I didn't notice this until the PR.
Sorry if my interpretation is mistaken.

If it's inconvenient, I'd like to move the files to top-level at the repo.

Thank you.

purcell:
Yep, fine - merging now.

momomo5717:
Thank you for merging it!

Sorry for my insufficient explanation.

Thank you @milkypostman and @purcell for taking time on this and being patient with my poor English.

Issue #3235, opened 2015-10-24
melpa badges are not up to date


ptrv:
The melpa badges are not up to date. See package [ycmd](http://melpa.org/#/ycmd) or [flycheck](http://melpa.org/#/flycheck). The dates are not matching the ones from the title.

milkypostman:
looking into this.

milkypostman:
the pull code for images is broken for some reason

milkypostman:
```
$ gnutls-cli img.shields.io
Resolving 'img.shields.io'...
Connecting to '104.18.45.160:443'...
*** Fatal error: A TLS fatal alert has been received.
*** Received alert [40]: Handshake failed
*** Handshake has failed
GnuTLS error: A TLS fatal alert has been received.
```

something is wrong with modern gnutls on ubuntu with our imgshields code. i will keep digging when I get a chance. not sure why this is failing but it's only with version 24.3 of emacs.

rexim:
Also, we don't have a badge for [psc-ide](http://melpa.org/#/psc-ide)

http://melpa.org/packages/psc-ide-badge.svg returns 404

Could it be related to this issue? The package was added recently: https://github.com/milkypostman/melpa/pull/3215

cpaulik:
Could this be related to me getting an error when trying to build a new recipe?

This is what I get when I run `make recipes/live-py-mode`:

```shell
File: /home/cp/workspace/emacs/melpa/packages/live-py-mode-20150724.2151.entry
Contacting host: img.shields.io:80
gnutls.c: [0] (Emacs) fatal error: A TLS fatal alert has been received.
gnutls.c: [0] (Emacs) Received alert:  Internal error
gnutls.el: (err=[-12] A TLS fatal alert has been received.) boot: (:priority NORMAL :hostname img.shields.io :loglevel 0 :min-prime-bits 256 :trustfiles (/etc/ssl/certs/ca-certificates.crt) :crlfiles nil :keylist nil :verify-flags nil :verify-error nil :callbacks nil)
error in process filter: GnuTLS error: #<process img.shields.io<1>>, -12
Makefile:88: recipe for target 'recipes/live-py-mode' failed
```

This is on ArchLinux with emacs 24.5.1

milkypostman:
Same bug. I'm going to modify the script to use curl temporarily but
haven't had a free moment to do so.

purcell:
[...]

PR #3236, opened 2015-10-25
Added discover-point

netromdk:
* Added discover-point with which you can discover information about value at point or region in Emacs. If value is a number then binary/octal/decimal/hexadecimal/character values and conversions are shown. For strings each character is processed in the same way. Especially while programming I find myself converting between bin/oct/hex/dec/ascii/unicode often and this simple package makes it a bit more simple.
* Package can be found here: https://github.com/netromdk/discover-point
* I am the author.
* Used flycheck to check the package file.
* I tested with `make recipes/discover-point`,
* And installed successfully using `package-install-file`.

This is my first package so please do tell if something is not up to par.
Thanks!

### DIFF
--- a/recipes/describe-number
+++ b/recipes/describe-number
@@ -1,0 +1,1 @@
+(describe-number :repo "netromdk/describe-number" :fetcher github)
--- a/recipes/discover-point
+++ b/recipes/discover-point
@@ -1,1 +1,0 @@
-(discover-point :repo "netromdk/discover-point" :fetcher github)
--- a/recipes/discover-point
+++ b/recipes/discover-point
@@ -1,0 +1,1 @@
+(discover-point :repo "netromdk/discover-point" :fetcher github)


purcell:
Thanks! Have you seen the build-in `describe-char`? I don't see what additional functionality this package provides. The name `discover-point` doesn't really fit it well.

Quick feedback re. the code itself:

- If your code actually requires `lexical-binding`, you must declare a package dependency on `(emacs "24")`. If not, you should remove the `lexical-binding: t`
- `dp/` is a bad prefix to use. Not only does it not follow the [elisp conventions](https://www.gnu.org/software/emacs/manual/html_node/elisp/Coding-Conventions.html), but if everyone chose a 2-character symbol prefix, then there could only be 676 distinct prefixes, and there are already 2000+ packages in MELPA! :-)  So something looking like `discover-point--` would be the correct choice here.

Hope that helps - happy to discuss further.

netromdk:
Thanks for the feedback! Yes, I know `describe-char` and it's great but it only shows information for a single character, it doesn't interpret a number like "101" and "FE", for instance, and that was one of the main reason for making this package; to easily interpret/convert between representations.

I removed the lexical binding.

Hm maybe the name I chose wasn't the best and most suitable one. I will change it to be called `describe-number`, which is exactly what it is doing.

purcell:
Ah yes, great name!

netromdk:
Okay I changed the name everywhere now and tested recipe and install. Any chance you can reboot travis to forget the old name?

Thanks

purcell:
I just committed the updated version of the recipe directly in d5ed906 instead. Thanks!

P.S. You should use `describe-number--` as the private symbol prefix rather than `describe-number/` -- elisp is not Clojure, and the best thing for the community is if authors make it easy and consistent for everyone to work on each other's code.

netromdk:
Thank you! I made the change. :)

purcell:
Great. Looking forward to trying this out - I can see that it would be handy.

netromdk:
Do I have to do something specific for the badge to be created as well? I see that the package was included in the last build at http://melpa.org/#/describe-number but the badge was not created at http://melpa.org/packages/describe-number-badge.svg

When I do `make recipes/describe-number` locally then it does create the svg.

Thanks

purcell:
There's an issue with the badges right now: #3235. You haven't done anything wrong, and the badge will start working when we fix the other issue. Sorry for the disruption.

PR #3237, opened 2015-10-26
Adding recipe for gregorio-mode

cajetanus:
Modified gregorio-mode in accordance with MELPA guidelines. Changed C-u to C-+, put in header and footer text for MELPA

### DIFF
--- a/recipes/gregorio-mode
+++ b/recipes/gregorio-mode
@@ -1,2 +1,2 @@
 (gregorio-mode :fetcher github
-	       :repo "cajetanus/gregorio-mode")
+	       :repo "cajetanus/gregorio-mode.el")
--- a/recipes/gregorio-mode
+++ b/recipes/gregorio-mode
@@ -1,0 +1,2 @@
+(gregorio-mode :fetcher github
+	       :repo "cajetanus/gregorio-mode")


purcell:
Thanks. Quick feedback:

- `C-c +` and `C-c -` are discouraged for major modes, and `C-c f` is reserved for user customisation, so should not be used.
- `;;;###autoload` cookies would be a nice addition, at least for the `define-derived-mode`
- There's no need to call `use-local-map` or `run-hooks` in your mode declaration: the `define-derived-mode` macro generates code to do that automatically.
- Try not to mention specific keybindings in docstrings: you can use the `\\{` trick to mention commands instead, and their bindings will be automatically shown to the user.

Hope that helps! :-)

cajetanus:
Steve,

Many many thanks for the feedback! This is my first contribution to
MELPA so it is good to have such a friendly maintainer to help me
through the process!

I put a ;;;###autoload line before define-derived-mode, as per the docs.

Gregorio is undergoing a major revision change and this will have a
major effect on the use of gregorio-mode. I will modify all these
keystrokes in the commit that will accompany the release of the new
gregorio. There are quite a few users that will just be confused if I
were to modify them extensively now, but with the new release of
gregorio everything will be quite different, so I will also implement
all your corrections in the next release.

Adding the project to MELPA was at the request of one user, but I am
sure many would like to use MELPA to install gregorio-mode. When
gregorio becomes part of texlive I am sure there will be more users.

Thank you so much again for the feedback.

JJ +

-----Original Message-----
From: Steve Purcell
Sent: 10/26/2015 08:05:06 AM


Thanks. Quick feedback:

- `C-c +` and `C-c -` are discouraged for major modes, and `C-c f` is reserved for user customisation, so should not be used.
- `;;;###autoload` cookies would be a nice addition, at least for the `define-derived-mode`
- There's no need to call `use-local-map` or `run-hooks` in your mode declaration: the `define-derived-mode` macro generates code to do that automatically.
- Try not to mention specific keybindings in docstrings: you can use the `\\{` trick to mention commands instead, and their bindings will be automatically shown to the user.

Hope that helps! :-)

---

purcell:
Happy to help. Merged now.

PR #3238, opened 2015-10-26
Add Matrix.org client and API `mclient'

rrix:
Summary: `mclient' provides a simple chat client and protocol API to the
Matrix.org decentralized RPC system. Users can chat with other users or
use Matrix as a messaging layer for their projects.

Direct Link: https://fort.kickass.systems:10082/cgit/personal/rrix/pub/matrix.el.git

Association: It's my package :)

### DIFF
--- a/recipes/matrix-client
+++ b/recipes/matrix-client
@@ -1,0 +1,1 @@
+(matrix-client :url "git://fort.kickass.systems/personal/rrix/pub/matrix.el" :fetcher git)


purcell:
Hi. Thanks for this. Some feedback:

- `(json)` and `(request)` are invalid dependency specs (note to self: [here](http://fort.kickass.systems:10082/cgit/personal/rrix/pub/matrix.el.git/tree/matrix.el)): you need to include the version numbers. `flycheck-package` can help you get this right.
- You've put `Package-Requires` lines in multiple files, but only the one in the filename matching the package name will be used by MELPA, ie. `mclient.el`.
- `mclient.el` defines a command `matrix-client`, which is out of place because it lacks the `mclient-` prefix. How about just calling that command `mclient`?
- Including `matrix.el` inside this package introduces the possibility of clashes with other libraries called `matrix.el`: it's such a common name that this is particularly likely in this case. Perhaps `matrix-api.el` would be a safer and more descriptive choice? And if this is likely to be useful outside `mclient`, then it should perhaps be a separate package upon which `mclient` would depend.
- You're using `setq-local`, which is not available in Emacsen older than 24.3 (IIRC), so you should either use the classic `(set (make-local-variable ...) ...)` or declare a `Package-Requires` dependency on the corresponding minimum Emacs version.

Hope that helps!

rrix:
Thanks purcell; I'll bump this when I've updated the package. :)

rrix:
@purcell I bumped my package to 0.1.2, feel free to re-review at your leasure. For now I'm going to keep matrix-client and matrix-api bundled since they're both still pretty tightly intertwined and the API is nowhere near solid or frozen for 3rd party use; when it hits a point where I am comfortable maintaining an ABI, I'll push it in to its own package, I think.

purcell:
Ah, that looks great! Merging now.

milkypostman:
@rrix your side is down so builds are getting out of date. will it be coming back online?

rrix:
Aye, shoot, let me make a pull request to update the remote.

rrix:
Fixed in #4026

PR #3240, opened 2015-10-26
Add helm-fuzzier

EphramPerdition:
[helm-fuzzier](https://github.com/EphramPerdition/helm-fuzzier) improves Helm's fuzzy matching results. 
It works well in combination with `helm-flx`, or without it.

I am the author and maintainer of `helm-fuzzier`.

### DIFF
--- a/recipes/helm-fuzzier
+++ b/recipes/helm-fuzzier
@@ -1,0 +1,2 @@
+(helm-fuzzier :repo "EphramPerdition/helm-fuzzier"
+                 :fetcher github)


purcell:
Thanks! Quick feedback:

- You're missing a `defgroup`
- You should probably add a `;;;###autoload` cookie for the `define-minor-mode` form

Hope that helps!
-Steve

EphramPerdition:
Thanks, this is my first package. 
All fixed.

purcell:
Great, thanks!

PR #3242, opened 2015-10-27
Add ox-trac: Org Mode Export for Trac WikiMedia

JalapenoGremlin:
This package (ox-trac) provides a Org mode backend capable of generating WikiFormatting for Trac.

The repository for this package is: https://github.com/JalapenoGremlin/ox-trac

I am the original author and maintainer of this project. 

I have tested (C-c C-c) building the recipe and have installed it into my packages.

### DIFF
--- a/recipes/google-c-style
+++ b/recipes/google-c-style
@@ -1,3 +1,2 @@
-(google-c-style
- :url "http://google-styleguide.googlecode.com/svn/trunk/"
- :fetcher svn)
+(google-c-style :fetcher github
+                :repo "google/styleguide")
--- a/recipes/ox-trac
+++ b/recipes/ox-trac
@@ -1,0 +1,1 @@
+(ox-trac :fetcher github :repo "JalapenoGremlin/ox-trac")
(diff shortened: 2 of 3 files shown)


purcell:
Looks good, but you should aim for a consistent symbol prefix as per the [elisp conventions](https://www.gnu.org/software/emacs/manual/html_node/elisp/Coding-Conventions.html). Here you've got some symbols which are `org-export-trac-*`, some which are `org-trac-*`, and the package is called `ox-trac`. I'd suggest using `ox-trac` throughout.

JalapenoGremlin:
The naming conventions adheres closely to the org modules, 
ox-md, ox-html, ox-latex. 

Org is looking for a backend function of the form org-export-BACKEND-X. 
such as `org-html-export-as-html` which is defined in `ox-html.el`

purcell:
Ah, okay -- yes, I recall this arrangement now. Thanks for your patience: I just like to check that there's a reason for inconsistencies like this. Merging now.

JalapenoGremlin:
No problem. I only mentioned it cause I actually renamed some of my functions to match org mode. 
I think that the export engine was part of `org`, then `ox` came is as a `contrib` module, and then was integrated as a part of `org`. (so some non-standard naming came into being). 

I do appreciate all the work you do. Melpa is amazing and I am thankful.

PR #3245, opened 2015-10-27
Adding test-kitchen

jjasghar:
- Adding the test-kitchen for running test-kitchen inside emacs

### DIFF
--- a/recipes/test-kitchen
+++ b/recipes/test-kitchen
@@ -1,0 +1,1 @@
+(test-kitchen :fetcher github :repo "jjasghar/test-kitchen-el")


purcell:
Don't call this `test-kitchen-el` -- it's already in Emacs, so it's obviously elisp! Just use `test-kitchen.el` as the filename, `test-kitchen` as the feature and module name, and then everything will match the `test-kitchen-*` prefix you've (correctly) used for all the symbols in the code.

You can amend the recipe commit accordingly and force-push it when done to update the PR in place.

purcell:
Also: the `Package-Requires:` header in the file is syntactically invalid, so you should either remove it, or use `()` as its empty value.

jjasghar:
I think it's ready to go! :metal:

purcell:
Great, thanks.

PR #3247, opened 2015-10-28
merlin: use stable branch rather than master

def-lkb:
The current merlin recipe tracks the master branch.

Sometimes backward incompatible changes are pushed there and this will break merlin-mode for people using melpa.
We created the opam-stable branch to track the current stable version.

### DIFF
--- a/recipes/merlin
+++ b/recipes/merlin
@@ -1,1 +1,1 @@
-(merlin :fetcher github :repo "the-lambda-church/merlin" :files ("emacs/*.el"))
+(merlin :fetcher github :repo "the-lambda-church/merlin" :branch "beta" :files ("emacs/*.el"))


milkypostman:
If you want a stable release you should tag a release and direct people to
melpa stable (stable.melpa.org). Melpa is intended to track the master
branch and it is expected that sometimes things break.

We understand this sometimes causes developers issues but we have had this
discussion a number of times already and feel this is still the right
decision.

def-lkb:
Merlin is a software that can be used independently of emacs.

The problem here is that the same repository tracks the development of this software and of its emacs bindings.
Only developers of this software use the master branch, and they work directly with the elisp files from the git repository and not from the melpa one.

By tracking master, the melpa recipe targets an empty set of users: developers will use git directly, regular users will use the stable software, but then can't rely on melpa.

My impression is that the branch to track should be a "beta" one, where we include experimental features, excepting the ones that are going to break for anyone but the git users.
It is even better for us as this will bring useful feedback from users of our melpa package.

Does that sounds reasonable to you?

purcell:
> My impression is that the branch to track should be a "beta" one, where we include experimental features, excepting the ones that are going to break for anyone but the git users.
> It is even better for us as this will bring useful feedback from users of our melpa package.
> 
> Does that sounds reasonable to you?

Probably, yes. A primary concern is that you might one day stop using that branch as a consequence of changing your release process, then the MELPA package will get stale and nobody will notice. So we very much prefer to track the repo's default branch. You might even consider using the beta branch as the repo's default branch. But I'll leave that choice to you -- amend and force-push the commit in this PR if you prefer.

def-lkb:
Thanks for your answer.
I did not forget this PR, I have to think a bit more about our move, either going beta or leaving things as is... I will have more time next week.
Sorry for the delay.

def-lkb:
I updated the commit to point to beta branch.

Your concern that the MELPA package will get stale is valid, but it will be noticed: in our case, the elisp script is communicating with an helper process updated by other means.
This is why we get wrong reports currently. Pointing to the appropriate branch will not make the reports disappear, but it will make them right.

Please merge this PR.

purcell:
As far as I can see, the commit still points at the `opam-stable` branch. Did you forget to amend the recipe commit and force-push it.

def-lkb:
I amended without adding the file... Sorry for that!

purcell:
No probs, thanks - merged!

def-lkb:
Thank you for your quick response.

PR #3249, opened 2015-10-28
Add zone-nyan

wasamasa:
I love writing silly stuff.


purcell:
Yay, great! Maybe a couple of docstrings for the interactive functions...? :-)

wasamasa:
Thanks, I knew something was still missing.

PR #3250, opened 2015-10-30
Add grepint

kopoli:
This is a generic grep interface, which selects a grep-like program based on context. It uses helm to select candidates.

### DIFF
--- a/recipes/helm-grepint
+++ b/recipes/helm-grepint
@@ -1,0 +1,1 @@
+(helm-grepint :fetcher github :repo "kopoli/helm-grepint")


milkypostman:
Any way I could persuade you to call this helm-grepint? This would be more
inline with other helm packages.

kopoli:
Indeed that would be a more logical naming. I'll make the change.

kopoli:
Name of the repo, package and the contents of this commit have been changed.

purcell:
Yay, great - merged!

PR #3251, opened 2015-10-30
Add breaktime recipe.

joeheyming:


purcell:
Check out the built-in "type-break-mode" -- I think that makes your package redundant, but correct me if I'm wrong.

joeheyming:
yeah, its more robust.  mine just has a fancy animation inspired by M-x butterfly

PR #3252, opened 2015-10-30
adding caroline theme

xjackk:
adding Caroline, an Emacs theme.


purcell:
Pending https://github.com/xjackk/caroline-theme/pull/1

PR #3254, opened 2015-10-30
Add pastelmac and termbright themes

bmastenbrook:
A pair of themes, one for use with window systems and the other for use on plain terminals. See https://github.com/bmastenbrook/pastelmac-theme-el and https://github.com/bmastenbrook/termbright-theme-el for screenshots.


purcell:
The colon is missing after `Package-Requires` in each of these themes. They look fine otherwise, so just ping me when fixed.

bmastenbrook:
@purcell Oops. Fixed.

purcell:
Thanks!

PR #3256, opened 2015-10-31
Update recipe cal-china-x to point to github.

xwl:
i'm author of cal-china-x.el.  This package has been moved to github for long.  And the current version in melpla(which points to emacswiki) is pretty outdated.


purcell:
Great, thanks.

PR #3258, opened 2015-10-31
cal-china-x: Add new data file.

xwl:


purcell:
Thanks. Could you use `:defaults` instead of `"*.el"` please? Just amend and force-push the commit.

xwl:
Updated.

purcell:
Thanks!

PR #3260, opened 2015-11-01
Add recipe for BPR

ilya-babanov:
Hello!

This is a package for asynchronous process execution,  similar to `async-shell-command`.
Main purpose: run tests and builds in background, but with progress notifications.
More details in [repo](https://github.com/ilya-babanov/emacs-bpr/).


purcell:
Great, thanks.

PR #3262, opened 2015-11-01
Add commify recipe.

ddoherty03:
Have added a small package to toggle insertion of commas or other grouping characters to a number in the buffer.

https://github.com/ddoherty03/commify

I am the package author.

My first package, so beware!

### DIFF
--- a/recipes/commify
+++ b/recipes/commify
@@ -1,0 +1,3 @@
+(commify
+ :fetcher github
+ :repo "ddoherty03/commify")


purcell:
Thanks! The `-package.el` file is unnecessary here (and would have had to be called `-pkg.el` anyway) so you should delete it please. Also, `:commify` should be `'commify` [here](https://github.com/ddoherty03/commify/blob/master/commify.el#L73).

ddoherty03:
Wow.  That was fast.  Thanks.  Changes committed to 

https://github.com/ddoherty03/commify

purcell:
Great, thanks - merged!

ddoherty03:
That wasn't as painless as I feared.  Thanks.

PR #3264, opened 2015-11-03
fix(run-travis-ci): only attempt to build modified recipes if they exist

vyp:
i.e. They were not deleted.

Ref #3263 to see an example of the travis build failing due to this.

---

(Assuming the loop always stays in the same directory after building each recipe?)


purcell:
Great, thanks!

PR #3265, opened 2015-11-04
Added wilt recipe

abingham:
wilt.el provides `wilt-mode`, a mode which calculates WILT (a software complexity metric) for your code.


vyp:
The `:files` part is not needed as melpa chooses all *.el files in the root directory by default. (See the `:files` bullet point [here](https://github.com/milkypostman/melpa#recipe-format).)

abingham:
Sorry about that. Should be fixed now.

vyp:
No worries, that's cool. I can't merge this though, you need to wait for the official maintainers. In the meantime you should probably squash these commits into one. However, the official maintainers will also review the elisp of the package and suggest fixes/improvements if needed, so you may need to squash again after that too (or just force push).

abingham:
OK, I'll just wait for any further feedback before squashing things.

abingham:
I need to close and recreate this due to some unrelated changes in another recipe.

PR #3267, opened 2015-11-04
Adding noxml-fold recipe.

paddymcall:
* A brief summary of what the package does.

noXML-fold folds away markup in xml files (tags or whole elements),
similar to what tex-fold-mode does for (La)TeX files.

* A direct link to the package repository

https://github.com/paddymcall/noXML-fold

* Your association with the package

I wrote and maintain this package.

### DIFF
--- a/recipes/noxml-fold
+++ b/recipes/noxml-fold
@@ -1,0 +1,1 @@
+(noxml-fold :fetcher github :repo "paddymcall/noXML-fold")


milkypostman:
This package could use some love from the *checkdoc*.  You can alternatively install `flycheck` package and enable `flycheck-mode`.  But otherwise please run `checkdoc` to clean things up.

purcell:
Also, the symbol prefix should be `noxml` throughout, not `noXML` -- capitals are not idiomatic in elisp.

paddymcall:
Thank you for checking this. I've fixed things up in the current branch of https://github.com/paddymcall/noxml-fold.

purcell:
Looks good!

PR #3268, opened 2015-11-05
Make sure anchor cursors are pointers.

NicolasPetton:
Before:

melpa-current

After:

melpa-better

As you can see, huge improvement.

### DIFF
--- a/html/css/melpa.css
+++ b/html/css/melpa.css
@@ -34,6 +34,9 @@ pre code { /* Match highlight.js styles to bootstrap */
   font-size: 18px;
   line-height: 30px;
 }
+a {
+  cursor: pointer;
+}
 a[name] { /* Stop navbar from hiding anchors */
   padding-top: 60px;
   margin-top: -60px;


Silex:
It'd be more interesting to know why these links  _are not_ `pointer` by default, given it's normally the default.

NicolasPetton:
Philippe Vaucher <notifications@github.com> writes:

> It'd be more interesting to know why these links  _are not_ `pointer`
> by default, given it's normally the default.

That's because there's no href on them.

NicolasPetton:
Philippe Vaucher <notifications@github.com> writes:

> It'd be more interesting to know why these links  _are not_ `pointer`
> by default, given it's normally the default.

It's not the default on anchors, it's the default on anchors with an href.

Silex:
> It's not the default on anchors, it's the default on anchors with an href.

Oh, right... interesting.

purcell:
Thanks!

PR #3270, opened 2015-11-05
remove magit-tramp

tarsius:
This package broke a long time ago and nobody every complained about
that.  Also Yann who wrote this package never replied to my inquiries.

The repository is now at https://github.com/emacsattic/magit-tramp.


milkypostman:
thanks. is it worth pointing to the new repository or just drop it cause it's unmaintained?

tarsius:
Just drop it.

PR #3271, opened 2015-11-05
Add recipe for gmpl-mode

cute-jumper:
`gmpl-mode` is a major mode for editing GMPL(MathProg) files.

https://github.com/cute-jumper/gmpl-mode

### DIFF
--- a/recipes/gmpl-mode
+++ b/recipes/gmpl-mode
@@ -1,0 +1,2 @@
+(gmpl-mode :repo "cute-jumper/gmpl-mode"
+           :fetcher github)


purcell:
Thanks. `font-lock-defaults` and `indent-line-function` are always buffer-local, so you don't need to use `make-local-variable` with those. And maybe you'd like to use `prog-mode` as the parent mode instead of `fundamental-mode` -- if so, just also add `;; Package-Requires: ((emacs "24"))` near the `;; Keywords:` header.

cute-jumper:
Thank you, Steve! I really appreciate your suggestions and your work for Melpa!:+1:

purcell:
Happy to help.

PR #3273, opened 2015-11-06
Add doc/ivy.texi to the swiper recipe

abo-abo:
Will the info file be built automatically? I haven't tried this before.


milkypostman:
It should get built automatically. Do you want the resulting file to end up in the root of the package though?

abo-abo:
I guess. As long as it gets added to the Info path it doesn't matter.

purcell:
Should be fine then, thanks.

PR #3274, opened 2015-11-06
Add evil-magit

justbur:
This is a new set of vim-like key bindings for magit

cc @tarsius

### DIFF
--- a/recipes/evil-magit
+++ b/recipes/evil-magit
@@ -1,0 +1,1 @@
+(evil-magit :repo "justbur/evil-magit" :fetcher github)


tarsius:
As I said in https://github.com/magit/evil-magit/issues/1, my hope was that several implementations would be contributed, which could then be compared before we crown a winner. Based on https://github.com/magit/magit/issues/1968 I was hopeful, that this would happen. But it didn't and so it is only fair that the only candidate wins the race :-)

But I would like to move this out of the Magit organization. I created this repository as a organization repository to make it the "official place to forge vilified magit bindings", but now that there is an implementation, I don't want to give that an "official" blessing - not because I don't think it is good enough, but because my judgment would necessarily be rather superficial, coming from someone who is not familiar with Vim/Evil bindings. (From an Elisp perspective things look good.)

Also my experience with contributed extensions (`magit-{stgit,svn,topgit}`) wasn't so good in the past - the contributors disappeared after a while and I was stuck maintaining utilities I am not interested in using myself. And finally, since you wrote this, why not get all the praise for it?

(By the way, you might want to change the copyright and add author and maintainer header keywords. Personally I would also scrape the current history and start with a new "initial commit" at this point.)

You should have the necessary permissions to transfer the repository to your user account. Moving the repository also moved the issues. And Github redirects from the old location to the new.

justbur:
@tarsius not a problem at all. I'm happy to do this. I was surprised when
you gave me access in fact. Just one question for you quickly. Is it ok to
keep the name evil-magit?

tarsius:
Well, you showed initiative and also that you were taking other users' concerns into account, so I felt this was safe in your hands. And of course you can name this `evil-magit`. It's a good name, so why not use it?

justbur:
Great, thank you, and if you change your mind, I'd be happy to give ownership back

justbur:
This PR is now updated with the new repository

purcell:
Thanks folks - merged.

PR #3275, opened 2015-11-06
Add Plugins to Support Liquid Haskell

ranjitjhala:
I'd like to submit two recipes as one depend on each other. Thanks!

### DIFF
--- a/recipes/flycheck-liquidhs
+++ b/recipes/flycheck-liquidhs
@@ -1,4 +1,3 @@
 (flycheck-liquidhs
  :fetcher github
- :repo "ucsd-progsys/flycheck-liquidhs.el"
- :branch "master")
+ :repo "ucsd-progsys/flycheck-liquidhs.el")
--- a/recipes/flycheck-liquidhs
+++ b/recipes/flycheck-liquidhs
@@ -1,0 +1,4 @@
+(flycheck-liquidhs
+ :fetcher github
+ :repo "ucsd-progsys/flycheck-liquidhs.el"
+ :branch "master")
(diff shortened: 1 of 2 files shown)


purcell:
Thanks. Quick initial feedback:

1. Please omit `:branch "master"`
2. `liquid-types` does not match `liquid-tip`, but the two names *must* match in order to make a valid package.

purcell:
Ping!

ranjitjhala:
Thanks for the reminder !

ranjitjhala:
Hi, I think we've made the necessary changes. Is this good now? Many thanks!

purcell:
Thanks. I don't think there's any reason why `liquid-types` should depend on or refer to `flycheck` or `flycheck-liquidhs`. The `pos-tip` dependency also seems spurious, because it is required optionally (and in any case, don't forget that just because it is installed, it doesn't mean a user wants to use it.)

So I think you can trim the dependency list a little there!

Re. the recipes, they look good now, but I'll merge them when I have a few moments to squash the four recipe commits into one. If you feel like doing that yourself locally and force-pushing the branch back here, that'd speed the process along.

tarsius:
> Re. the recipes, they look good now, but I'll merge them when I have a few moments to squash the four recipe commits into one.

@purcell In cases like this, when you have already stated that you would merge, can I just do it?

purcell:
@tarsius Yep, though I'll do it in this case. Back in Feb the github PR merge couldn't squash, which is probably why I punted at the time.

tarsius:
I think the "merge by squashing" feature usually isn't the right thing to do, but for Melpa it is excellent.

purcell:
Yep, agreed. I lost count of the number of times I asked people to go back and tidy up their PRs before github added that feature.

ranjitjhala:
Thanks @purcell and @tarsius and @themattchan!

PR #3277, opened 2015-11-07
Add graphene-meta-theme

rdallasgray:
Adds the 'meta-theme' used by [graphene](https://github.com/rdallasgray/graphene), which I'm breaking out into its own project.


milkypostman:
merging but it should have `;;; graphene-meta-theme.el ends here` at the end. we add it though so got you covered.

PR #3279, opened 2015-11-07
Add emoji-fontset recipe

zonuexe:
https://github.com/zonuexe/emoji-fontset.el


purcell:
Thanks -- looks useful!

zonuexe:
Thank you!

PR #3280, opened 2015-11-07
Add x86-lookup recipe.

skeeto:
https://github.com/skeeto/x86-lookup

Quick jump from within Emacs to the relevant page of Intel's x86 instruction set manual.


purcell:
Thanks!

PR #3282, opened 2015-11-08
add emacs-jp/replace-colorthemes

mmckinst:
Upstream is OK with it being added: https://github.com/emacs-jp/replace-colorthemes/issues/10

### DIFF
--- a/recipes/color-theme-modern
+++ b/recipes/color-theme-modern
@@ -1,0 +1,3 @@
+(color-theme-modern
+ :repo "emacs-jp/replace-colorthemes"
+ :fetcher github)


purcell:
Yeah, not sure it's a great idea to add these _en masse_, and we couldn't do so yet anyway since there's no `replace-colorthemes.el` or `replace-colorthemes-pkg.el` from which MELPA could read overall package metadata such as the description. If we were to add all the themes in a single package, a different package name might be better, eg. `colorthemes-deftheme`, `colorthemes-modern`, so that it is listed alongside `colorthemes` in package lists.

I'll have to think about the trade-offs here.

/cc @syohex

mmckinst:
Its similar to base16-theme in that it provides a bunch of themes. A different name is not a problem.

milkypostman:
I would suggest we follow the base16 naming then. xxxxx-themes. But you will still need a xxxxx-themes-pkg.el file.

syohex:
Thanks for information. I have added `replace-colorthemes-pkg.el` at https://github.com/emacs-jp/replace-colorthemes/pull/15.

purcell:
@syohex a better name would be "colorthemes-modern", so that it is listed alongside the old "colorthemes" package. :-)

milkypostman:
@purcell this was my suggestion; when packages contain multiple themes ala base16-themes. but i am ok with this suggestion too. @syohex sorry for the confusion.

syohex:
> a better name would be "colorthemes-modern", so that it is listed alongside the old "colorthemes" package. :-)

Thanks for suggestion, I will change names.

syohex:
I have updated package name from `replace-colorthemes` to `color-theme-modern` at https://github.com/emacs-jp/replace-colorthemes/pull/16 .

@mmckinst  Could you update package name in recipe ?

mmckinst:
package name updated and commits squashed in to one. thanks for the help!

syohex:
:+1:

purcell:
Oh yay, nice work folks - thanks!

PR #3283, opened 2015-11-09
Add recipe for evil-indent-plus

TheBB:
This is a rebranded fork of `evil-indent-textobject`. That package has some serious problems which will probably never be fixed: https://github.com/cofi/evil-indent-textobject/issues/1. @cofi, who maintains that package, has no activity in the last year.

In Spacemacs we switched to using this fork, see https://github.com/syl20bnr/spacemacs/issues/2326 and https://github.com/syl20bnr/spacemacs/pull/2656.

Since we are trying to avoid using too many quelpa packages, it would be good to have this on MELPA or bundle it with Spacemacs. From the above issues, the general consensus is that this fixes a genuine problem that many people experience with the original package, so my first choice would be to have it in MELPA, so that non-Spacemacsers can benefit from it, too. I hope the name change and the extra features and bug fixes sufficiently distances this from the "fork" label.

Upstream: https://github.com/TheBB/evil-indent-plus

### DIFF
--- a/recipes/evil-indent-plus
+++ b/recipes/evil-indent-plus
@@ -1,0 +1,1 @@
+(evil-indent-plus :fetcher github :repo "TheBB/evil-indent-plus")


milkypostman:
If we do this it should contain the `:old-names` tag (see cask recipe) and point to the other recipe, which we should remove if it's an issue like this. 

It seems ok to me but for situations like this I like @purcell's input.

purcell:
FWIW, we're okay with identically-named forks, so long as the original repo is clearly unmaintained and the new maintainer is committed to working on the fork. In this situation it looks like there were simply differences of opinion re. the original code, and the wording of the pull request you linked to perhaps wasn't very diplomatic.

In any case, we can certainly add this and also leave the old package in place for now, I think.

TheBB:
For the record, that wasn't a pull request.

purcell:
Ah yes, my mistake. I suppose the author can't really be blamed for not following up on the feedback in that case.

Having had a quick look, [the autoloads](https://github.com/TheBB/evil-indent-plus/blob/master/evil-indent-plus.el#L171) are a bit bogus -- don't wrap them in an "eval-after-load", which is meant mainly for user configs.

syl20bnr:
I believe that the current version in MELPA of evil-indent-textobject is buggy and does not work as expected so maybe replacing is a better option instead of having two recipes.

TheBB:
In any case, I pushed some saner autoloads. That setup hails from the existing `evil-indent-textobject` package.

purcell:
Thanks!

PR #3286, opened 2015-11-10
add recipe for elog

lujun9972:
elog is a simple logging library for elisp. It is extended from logito.

### DIFF
--- a/recipes/elog
+++ b/recipes/elog
@@ -1,0 +1,1 @@
+(elog :fetcher github :repo "lujun9972/elog")


purcell:
Thanks. Quick feedback:

- Some docstrings would be helpful, I think
- No need to put "for Emacs" in the package description line -- it's redundant!

lujun9972:
I have added some doc-string and deleted the "for Emacs" from the package description line

purcell:
Thanks

PR #3287, opened 2015-11-10
Add `org-time-budgets` recipe

leoc:
Small but soon to be fancy.


purcell:
Thanks. This isn't quite a valid package yet. Please add a `provide` form, and also add the missing `.el` in the filename on the first line.

leoc:
Alright, sorry for that. Did the changes.

purcell:
Super, thanks.

PR #3288, opened 2015-11-11
Add some recipes

dochang:
- flymake-less
  - ri
  - cflow-mode
  - df-mode

### DIFF
--- a/recipes/cflow-mode
+++ b/recipes/cflow-mode
@@ -1,0 +1,3 @@
+(cflow-mode :url "git://git.sv.gnu.org/cflow.git"
+            :fetcher git
+            :files ("elisp/cflow-mode.el"))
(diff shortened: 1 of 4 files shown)


purcell:
Thanks. Generally we'd prefer separate PRs in cases like this, because each package needs to be reviewed individually for inclusion in MELPA.

* `flymake-less` is a bit rough, and I'd suggest `flycheck` instead, but I'm happy to add it if you're finding it helpful.
* `ri` and `df-mode` are both on Emacswiki, and I think (@milkypostman?) we had decided to avoid adding new Emacswiki-hosted packages.
* As for `cflow-mode.el`, it's not yet a valid package, because its header line says `cflow.el`, which doesn't match the actual filename. Otherwise it seems okay, so perhaps you could request a fix upstream?

milkypostman:
Lets not add additional emacswiki packages. Please ask the authors to use something with authentication.

@dochang we totally appreciate your effort to submit this PR but would you please submit each as separate PR because it's really hard to analyze these each.

PR #3290, opened 2015-11-11
Add fortune-cookie

andschwa:
A simple package which sets the `initial-scratch-message` to an output
from `fortune`, mimicking a shell login fortune. Also utilizes `cowsay`
if available, and is customizable.

https://github.com/andschwa/fortune-cookie

I am the author of this silly package.

### DIFF
--- a/recipes/fortune-cookie
+++ b/recipes/fortune-cookie
@@ -1,0 +1,2 @@
+(fortune-cookie :repo "andschwa/fortune-cookie" :fetcher github)
+


purcell:
Thanks. You know about the built-in `fortune` command too, right? Wonder if that might save needing to shell out to the `fortune` program.

andschwa:
I should add a note about why I didn't use `fortune.el` in the readme; it's lacking in a few ways. Shelling out myself ended up being much easier to deal with, especially when using `cowsay`.

Thanks for the merge!

PR #3291, opened 2015-11-11
Added therapy recipe.

abingham:


purcell:
Thanks. Sent you a quick fix in https://github.com/abingham/therapy/issues/3 tooo.

PR #3292, opened 2015-11-11
Added wilt recipe.

abingham:


purcell:
Thanks. Seems like there should also be an autoloaded command which returns/displays the WILT score, so that people can use it interactively or programmatically without enabling the minor mode. /cc @sixty-north

PR #3293, opened 2015-11-11
Add recipe for emojify

iqbalansari:
This package provides a minor mode (`emojify-mode`) which can display emojis in Emacs similar to how Github does.

The code lives at https://github.com/iqbalansari/emacs-emojify

Thanks


purcell:
Thanks!

PR #3294, opened 2015-11-11
Add recipe for region-state

xuchunyang:
Repo direct link: https://github.com/xuchunyang/region-state.el

### DIFF
--- a/recipes/region-state
+++ b/recipes/region-state
@@ -1,0 +1,1 @@
+(region-state :fetcher github :repo "xuchunyang/region-state.el")


purcell:
Thanks. Quick feedback:

- You're using `defvar-local` here, so you need to declare a package dependency on `(emacs "24.3")`
- Rather than only defining a global mode, consider defining a pair of modes: local and global. Users can then enable the minor mode in individual buffers ([example here](https://github.com/purcell/whitespace-cleanup-mode/blob/master/whitespace-cleanup-mode.el#L106))
- I think that [this part of the define-minor-mode docstring text](https://github.com/xuchunyang/region-state.el/blob/master/region-state.el#L167-L170) is generated automatically for you, so you don't need to include it.

xuchunyang:
Thanks for the review.

> You're using defvar-local here, so you need to declare a package dependency on (emacs "24.3")

I had already included the definition of `defvar-local` for <24,3 in https://github.com/xuchunyang/region-state.el/blob/6e0251e/region-state.el#L40-L51.

> Rather than only defining a global mode, consider defining a pair of modes: local and global. Users can then enable the minor mode in individual buffers (example here)

Good idea, but for now, as the only user of this package, local minor-mode is unneeded to me.

> I think that this part of the define-minor-mode docstring text is generated automatically for you, so you don't need to include it.

Yes, I did borrow it from elisp manual (info node: (elisp) Defining Minor Modes), while it still apples to my command.

purcell:
Thanks, merging now.

> I had already included the definition of defvar-local for <24,3

I'd recommend not doing that in this case: firstly, the elisp manual suggests not including backported core functions without aliasing them to a different name, and secondly, the cost of a few `(make-local-variable ...)` calls here doesn't really justify the extra code for the backported `defvar-local`.

PR #3295, opened 2015-11-11
Added recipe for highlight-indent-guides

DarthFennec:


purcell:
Thanks.

All the top-level symbols need to be prefixed with the package name, e.g. `highlight-symbol-guides-`, as per the [elisp conventions](https://www.gnu.org/s/emacs/manual/html_node/elisp/Coding-Conventions.html). Names like `get-guides` and `guide-line` are likely to clash with code elsewhere otherwise.

DarthFennec:
Oh, yeah whoops. Should be better now.

purcell:
Nice, thanks.

PR #3299, opened 2015-11-14
FIX: Exclude compile-shim.el; MOD: Formatting

cslux:
We recently added a compile-shim elisp file, for byte-compiling the elisp files in the make process, that's all it used for. Also I formatted the recipe a bit nicer.

Thanks,
Christian

### DIFF
--- a/recipes/rtags
+++ b/recipes/rtags
@@ -1,1 +1,4 @@
-(rtags :fetcher github :repo "Andersbakken/rtags" :files ("src/*.el"))
+(rtags :repo "Andersbakken/rtags"
+       :fetcher github
+       :files ("src/*.el"
+               (:exclude "src/compile-shim.el")))


purcell:
Would it be an option to put non-distributable helper elisp in a separate dir? That would save having to modify the recipe in cases like this.

cslux:
I already thought about that, I will discuss this with the rtags project
maintainer, I'm just a contributer.

As a side node, somebody else created this MELPA package and it is actually
pretty useless without the rtags-backend, so whoever installs this package
needs also to compile and install rtags, which includes the elisp files as
well. Right now it doesn't make much sense. I just read quickly over the
MELPA manual, I think it is possible to call make/cmake as well? If so, it
would be better to fetch the rtags project and compile everything and
install the binaries etc into e.g. the emacs.d directory.

If this is possible, please let me know it and I will do it the right way
:-).

Thank you, for your patience.

Christian

purcell:
> I just read quickly over the MELPA manual, I think it is possible to call make/cmake as well?

Nope, that's not possible, but we'd normally suggest that packages which depend on an executable at least report when it can't be found, and either provide the Emacs user with info about how to install it, or even help them to do that automatically.

purcell:
Sorry for letting the merging of this slip through the cracks.

cslux:
No problem, thanks for merging.

PR #3301, opened 2015-11-15
Add recipes/sproto-mode

m2q1n9:
[sproto](https://github.com/cloudwu/sproto) is another protocol library like google protocol buffers , but simple and fast. I write a sproto major mode to support syntax highlight.

https://github.com/m2q1n9/sproto-mode

### DIFF
--- a/recipes/sproto-mode
+++ b/recipes/sproto-mode
@@ -1,0 +1,1 @@
+(sproto-mode :repo "m2q1n9/sproto-mode" :fetcher github)


milkypostman:
One quick comment:

Standard style for something like this is to autoload the `sproto-mode` and then ask users to add the `(add-to-list 'auto-mode-alist '("\\.sproto\\'" . sproto-mode))` line to their `init.el` files.

We avoid adding packages that change behavior simply by installing because that could be quite confusing.

m2q1n9:
OK Fixed, see here:
https://github.com/m2q1n9/sproto-mode/commit/ae8057cc367254279df8124823915defee4c8e42

milkypostman:
had one more comment.

m2q1n9:
:)
https://github.com/m2q1n9/sproto-mode/commit/1753277d9f2163fb3bc58b983a9892831cf9874b

PR #3302, opened 2015-11-15
Add recipe for diredful

thamer:
This package provides a widget.el interface for setting faces and colors in dired mode.

https://github.com/thamer/diredful

Thanks.

### DIFF
--- a/recipes/diredful
+++ b/recipes/diredful
@@ -1,0 +1,1 @@
+(diredful :fetcher github :repo "thamer/diredful")


milkypostman:
Couple of comments,

- you shouldn't initialize `diredful` in the file. i.e., it shouldn't be automatically enabled just from loading it.
  
    best way to do this is to create a `diredful-enable` function that will load all the settings and turn it on.

- would `diredful` be better as a minor mode? it seems like maybe there is no way to disable it? If it can be toggled then defining it as a minor mode would be better.

thamer:
A minor mode sounds reasonable, so I made this change 9cb0af66a82ee979db5e86197352bf58fb04adb0 .

purcell:
You don't need `diredful` and `diredful-mode` now -- they both take an "enable" arg, so just provide the mode and scrap the `diredful` function. If desired, use `defalias` to alias `diredful` to `diredful-mode`. (Great package name, btw.)

thamer:
I think changing the name of the function would suffice, because it's also used elsewhere. I wouldn't want the LISP code to turn the minor mode on or off.

purcell:
:+1:

PR #3303, opened 2015-11-15
Add bind-map

justbur:
bind-map is a macro that generalizes the leader concept from vim for
emacs users. It is a simple way to bind a keymap to multiple prefixes
with the option to condition on the current major mode, minor mode, or
evil state. It should be an easy way to organize key bindings in
personal keymaps.

Repo at [emacs-bind-map](https://github.com/justbur/emacs-bind-map). I'm the author


purcell:
Thanks.

PR #3304, opened 2015-11-15
Include all .exe and .exe.config files

rneatherway:
There may soon be another .exe file from FSharp.Compiler.Service, and
fsautocomplete.exe.config has not been picked up by the current recipe,
which is affecting Windows installations.


rneatherway:
The Travis failure seems spurious, something about missing `.dir-locals.el`.

purcell:
> The Travis failure seems spurious, something about missing .dir-locals.el.

Indeed. Hopefully fixed in 149684b.

PR #3308, opened 2015-11-19
Added flycheck-css-colorguard recipe

Simplify:
flycheck-css-colorguard.el --- Flycheck extension that uses CSS Colorguard in order to detect similar colors in CSS file.

https://github.com/Simplify/flycheck-css-colorguard
I'm author. My first elisp package.


purcell:
Great, thanks.

PR #3309, opened 2015-11-19
add elogcat recipe

youngker:
I would like to add a recipe for my elogcat.el package.


purcell:
Thanks - sent you a few fixes in https://github.com/youngker/elogcat.el/pull/1.

purcell:
Great - merged! :-)

PR #3313, opened 2015-11-21
Remove go-play recipe

syohex:
go-play was integrated into go-mode. `go-play-buffer` and `go-play-region` are implemented in go-mode.el now.

- https://github.com/dominikh/go-play.el


purcell:
Thanks!

PR #3314, opened 2015-11-22
Add multi-line recipe

IvanMalison:
- Summary: multi-line everything from function invocations and definitions to array and map literals in a wide variety of languages
- Repository: https://github.com/IvanMalison/multi-line
- Association: Author/Maintainer


purcell:
Thanks.

PR #3316, opened 2015-11-22
Add yaml-tomato

RadekMolenda:
Yaml-tomato is a simple tool to copy or show the yaml path currently under cursor.

Github: https://github.com/RadekMolenda/yaml-tomato

I am the owner of the package. And this is an initial release

### DIFF
--- a/recipes/yaml-tomato
+++ b/recipes/yaml-tomato
@@ -1,0 +1,1 @@
+(yaml-tomato :repo "RadekMolenda/yaml-tomato" :fetcher github)


purcell:
Thanks. Quick feedback:

- You should use the prefix `yaml-tomato-` for all top-level variables, as per the [elisp coding conventions](https://www.gnu.org/s/emacs/manual/html_node/elisp/Coding-Conventions.html). Package-internal symbols can be `yaml-tomato--*`.
- Don't use `;;;###autoload` for `(require 's)`, but use `;;;###autoload` for each of the `interactive` commands
- Don't describe the installation method in the `;;; Commentary:` section -- use that section to briefly describe how to use the package once installed.

Otherwise, looks good. Hope that helps!

RadekMolenda:
It should be better now. Thanks for the feedback!

tomekw:
:+1:

purcell:
Yep, better, but `yaml-tomato/` should be `yaml-tomato-` -- this isn't Clojure, and we should put aside our personal preferences in order to fit in with the community's standard way of doing things. :-)

RadekMolenda:
Done this. Sorry about it :)

purcell:
Perfect, thanks!

PR #3319, opened 2015-11-22
Add scribble-mode

whilp:
`scribble-mode` is a major mode for editing documents written in Racket's scribble format. I'm merely an interested user; @emacs-pe hosts the upstream code here on Github.

Ref:

- Package repository: https://github.com/emacs-pe/scribble-mode

### DIFF
--- a/recipes/scribble-mode
+++ b/recipes/scribble-mode
@@ -1,0 +1,2 @@
+(scribble-mode :repo "emacs-pe/scribble-mode" :fetcher github
+             :files ("*.el"))


purcell:
Thanks. A few issues here though:

- There's no `scribble-mode.el` or `scribble-mode-pkg.el` -- what you really needed to do was call the package `scribble` in order to match the bundled file.
- `"*.el"` is included by default, and would normally be omitted here
- But `flycheck-scribble.el` has more specific dependencies, and should be packaged separately

So what you should do is create 2 recipes, `scribble` and `flycheck-scribble`, each of which points at the `scribble-mode` recipe and has a `:files` spec which names the specific `.el` file.

purcell:
I'll go ahead and cc @marsam here, too.

marsam:
Hi, sorry for the delay. Thanks for the recipe.

> There's no scribble-mode.el or scribble-mode-pkg.el

I've renamed the package to scribble-mode.el

> But flycheck-scribble.el has more specific dependencies, and should be packaged separately

I think `flycheck-scribble.el` is simple enough to be part of flycheck upstream, I'll make a pr later.

purcell:
> I think flycheck-scribble.el is simple enough to be part of flycheck upstream, I'll make a pr later.

Sure, so then this recipe should just say `:files ("scribble-mode.el")`. Can you amend the commit accordingly and force-push it back here please?

tarsius:
The upstream repository no longer contains any elisp files beside `scribble-mode.el`, so this can be merged now.

purcell:
Thanks!

PR #3320, opened 2015-11-23
New package hlint refactor mode

fatlazycat:
Allow hint suggestions ( haskell ) to be applied in place

https://github.com/mpickering/hlint-refactor-mode

Not author, but was asked to help see https://github.com/mpickering/hlint-refactor-mode/issues/4

### DIFF
--- a/recipes/hlint-refactor-mode
+++ b/recipes/hlint-refactor-mode
@@ -1,0 +1,1 @@
+(hlint-refactor-mode :fetcher github :repo "fatlazycat/hlint-refactor-mode")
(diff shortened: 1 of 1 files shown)


purcell:
This is great, and I'll use this myself, but it needs a few changes before adding it here:

- All the top-level symbols should be prefixed with the package name, as per the [elisp conventions](https://www.gnu.org/s/emacs/manual/html_node/elisp/Coding-Conventions.html)
- The `"-mode"` prefix is conventionally only used for packages which provide a major mode. A better name for this overall package and `.el` file, then, would be `hlint-refactor`.

/cc @mpickering

P.S. There's [some discussion](https://github.com/flycheck/flycheck/issues/530) about eventually adding "fixup" functionality to `flycheck`, and `hlint`'s refactoring support would be a perfect application of this.

cocreature:
@purcell Updated.

PR #3322, opened 2015-11-23
Add encourage-mode package.

halbtuerke:
This package is inspired by a Visual Studio plugin (https://github.com/Haacked/Encourage) which displays some encouragements when saving buffers.


purcell:
Fun! Thanks for this.

PR #3323, opened 2015-11-23
add recipe for code-library

lujun9972:


purcell:
Thanks - sent you a quick fix in https://github.com/lujun9972/code-library/pull/1

PR #3324, opened 2015-11-23
Add resize-window

AdrieanKhisbe:
Add recipe for @dpsutton [resize-window](https://github.com/dpsutton/resize-window)

### DIFF
--- a/recipes/resize-window
+++ b/recipes/resize-window
@@ -1,0 +1,1 @@
+(resize-window :fetcher github :repo "dpsutton/resize-window")


purcell:
Thanks, but `rw-` is not a good prefix -- if everyone chose a 2-character symbol prefix, we'd only have ~600 different ones. Bytes are cheap, so can we just replace `rw-` with `resize-window-` throughout please? /cc @dpsutton.

dpsutton:
done. thanks for the heads up.

AdrieanKhisbe:
you're welcome :)

purcell:
Awesome, thanks folks!

AdrieanKhisbe:
:)

PR #3325, opened 2015-11-24
Add org-tracktable recipe

tty-tourist:
[org-tracktable](https://github.com/tty-tourist/org-tracktable) is a package for tracking your writing progress in an org-table. It provides functions for inserting a table for tracking progress, updating that table, and getting status messages with your progress.

### DIFF
--- a/recipes/org-tracktable
+++ b/recipes/org-tracktable
@@ -1,0 +1,1 @@
+(org-tracktable :fetcher github :repo "tty-tourist/org-tracktable")


purcell:
Thanks! The `org-tt-` prefix should really just be `org-tracktable-`, both for clarity and to be consistent with the package name. Those extra few chars are cheap! :-)

tty-tourist:
Good idea :-)

Prefix changed to `org-tracktable` in package and README.

purcell:
Super, thanks!

PR #3327, opened 2015-11-25
Add easy-lentic

tumashu:


purcell:
Thanks. Quick feedback:

- The `\C-c j` keybinding is reserved for user customisations, so you should choose something different
- The `lentic-*` methods should be called `easy-lentic-*` to match the package name, right, or am I missing a technical reason for this naming?

tumashu:
lentic-* is for lentic package and can't rename....

purcell:
Okay, makes sense.

purcell:
Let me know when you've updated the keybindings, then I can go ahead and merge this.

tumashu:
```
The \C-c j keybinding is reserved for user customisations, so you should choose something different
```
I don't familar this field, suggest one?

purcell:
The elisp manual has [some guidelines](https://www.gnu.org/software/emacs/manual/html_node/elisp/Key-Binding-Conventions.html) - maybe `C-c .` or `C-c :`?

tumashu:
Fixed, use  'C-c .'

tumashu:
Fixed, use  'C-c .'

tumashu:
Fixed, use  'C-c .'

purcell:
Super, thanks.

PR #3328, opened 2015-11-27
flycheck-typescript-tslint added

Simplify:
Package: flycheck-typescript-tslint
URL: https://github.com/Simplify/flycheck-typescript-tslint
Description: Adds TSLint stylistic and programming errors to flycheck in typescript-mode
I'm author.

### DIFF
--- a/recipes/flycheck-typescript-tslint
+++ b/recipes/flycheck-typescript-tslint
@@ -1,0 +1,1 @@
+(flycheck-typescript-tslint :fetcher github :repo "Simplify/flycheck-typescript-tslint")


purcell:
Thanks -- `(typescript-mode)` isn't a valid package dependency, because the version is required. `flycheck-package` can help you with this.

Simplify:
Version number for typescript-mode added

purcell:
There's a missing double quote.

Simplify:
Fixed now

purcell:
Cool, thanks. For next time: `flycheck-package`. ;-)

Simplify:
I have it installed, but for some reason didn't get any error in buffer. Even C-c C-c in recipe installed package without version number without any problems. I still have it installed in GNU Emacs. As dependency is 'typescript-mode-0' listed.

purcell:
Hmm, interesting - thanks for the info. I'll take a look.

syohex:
https://www.gnu.org/software/emacs/manual/html_node/elisp/Library-Headers.html

Emacs document says version part is number as string.

> Its format is a list of lists. The car of each sub-list is the name of a package, as a symbol. The cadr of each sub-list is the minimum acceptable version number, as a string. For instance: 
> ;; Package-Requires: ((gnus "1.0") (bubbles "2.7.2"))

However implementation of parsing `Package-Requirements` treat version as `0` if version part is missing.

https://github.com/emacs-mirror/emacs/blob/7cc233e1e3da297882c006c1f07c628fbd4e94d5/lisp/emacs-lisp/package.el#L960-L962

purcell:
> However implementation of parsing Package-Requirements treat version as 0 if version part is missing.

If so, that's just a quirk of the implementation in recent Emacs versions, so people should *always* include the version number, because it is mandatory.

PR #3329, opened 2015-11-27
add recipt baidu-life

lujun9972:
I suppose that only Chinese people will use Baidu. So I write the doc-string in Chinese, I hope it's OK.

### DIFF
--- a/recipes/baidu-life
+++ b/recipes/baidu-life
@@ -1,0 +1,1 @@
+(baidu-life :fetcher github :repo "lujun9972/el-baidu-life")


purcell:
Yep, that's fine, but expanding the API key at compile time seems like a very strange choice -- defining the custom group and API key only `when-compile` is very unusual.

lujun9972:
yes,I agree with you. It did quite strange.
My purpose is to let user customize the API KEY, and at the sam time I think it will be better to let the API KEY become the argument of macro `baidu-life--with-api-key`
I don't know how to integrate them without compile error except use the `eval-when-compile`
 so any advise?

purcell:
Why does the API key have to be accessible at compile time anyway? Normally you would just access the API key variable at runtime.

lujun9972:
Oh. I am not familiar with the macro thing.
you are right, I will fix it, thank you.

PR #3330, opened 2015-11-27
Removing two packages 

cwittern:
The packages removed with this pull request are no longer needed.  I would like to remove them to avoid confusing users, who might install it by accident.
(I hope this is the right way to do this)
All the best,  Chris


purcell:
Yep, thanks.

PR #3331, opened 2015-11-28
a tool used to view html (in browser) generated by org-file once the org-file changes

kuangdash:
Or says, view html generated by org-mode immediately (real-time).

### DIFF
--- a/recipes/org-iv
+++ b/recipes/org-iv
@@ -1,7 +1,7 @@
 (org-iv :repo "kuangdash/org-iv"
         :fetcher github
-        :files ("*.el" ("default"
-                        "default/org-iv-front-file"
-                        "default/org-back-file"
-                        ("css" "default/css/*")
-                        ("js" "default/js/*"))))
+        :files (:defaults ("default"
+                           "default/org-iv-front-file"
+                           "default/org-back-file"
+                           ("css" "default/css/*")
+                           ("js" "default/js/*"))))
--- a/recipes/org-iv
+++ b/recipes/org-iv
@@ -1,0 +1,7 @@
+(org-iv :repo "kuangdash/org-iv"
+        :fetcher github
+        :files ("*.el" ("default"
+                        "default/org-iv-front-file"
+                        "default/org-back-file"
+                        ("css" "default/css/*")
+                        ("js" "default/js/*"))))


milkypostman:
please change the first element of the recipe to `:defaults`.

i'm not sold on the name. `org-immediate-view` isn't that long. normally really short abbreviations are discouraged. what about `org-impatient` since that's basically what you're hooking up here right?

kuangdash:
What's `:defaults`? I can't find it ……Any hints?

When it comes to the package name, I think that the shorter the better if there is no ambiguity. `M-x org-iv/` can get the command quickly while `org-impatient/` is too long to get other commands such as `manually-update` and `restart-view`. Anyway, from `ido` to `helm` , it seems to be difficult for us to revert the cleanness of emacs package names. Don't worry and let it go, buddy.

milkypostman:
I understand but that reminds me that Org-iv/ is not recommended for elisp.
I'd have to dig up the link but I you should change to org-iv- as a prefix.

kuangdash:
Here comes an experience I encountered before:

I had used a package named `org-ref` (git clone, not by melpa) which use `index` as a function name that conflicts with the function in another packages, then I fixed it by renaming `index` into `org-ref-index`. (in fact, I should rename it into `org-ref--index`)

So ... what if `org-mode` package choose `org-ref-index` as the name of one function in `org-mode`?

It's difficult to judge where `org-ref-index` comes from: `org` or `org-ref`?

I prefer `/` to `-` in order to distinguish package-name-part from function-name. I think it should be recommended, not be forbidden.

purcell:
Yes, that's why the [elisp conventions](https://www.gnu.org/software/emacs/manual/html_node/elisp/Coding-Conventions.html) say to use a short word (non-dashed) as the prefix, with `-` and `--` as the separator for public and private symbols respectively. Whatever your feelings about `/` vs `--`, the conventions are about the community working together to write code which everyone can understand and maintain, so it's worth putting aside personal preferences IMO.

vyp:
@kuangdash Btw @milkypostman meant change the `"*.el"` to `:defaults`, it is documented in the readme.

kuangdash:
@vyp I see it, thanks.

@purcell I prefer `/` to `-` in order to `distinguish package-name-part from function-name`. There should be a  symbol other than `-` to do this. `--` for the private one, I comply with it, but how about the public one?

lujun9972:
Personally, I prefer the `/` style, however, I still agree with @purcell .
I think it's useless to argue coding-style here, because the conventions is not decided by anyone here.

kuangdash:
Melpa is the largest of all elisp package sources, and the most convenient one. Thanks to @milkypostman @purcell and other maintainers.

@lujun9972 The problem here is not only about coding-style, but also about the prevention of name conflict between packages. What's more, I'm not the first one who use `/` to replace `-` in order to deal with the problem, and I only followed this practical convention not created a new one. 

As I said before, melpa is the largest of all elisp package sources, so with the highest probability of suffering from the name conflict problem. I won't change `org-iv/` into `org-iv-` except that anyone can give a more rational reason.

lujun9972:
@kuangdash en... I got your problem now. 'org-iv-' is indeed an easy confusing prefix.

purcell:
> What's more, I'm not the first one who use / to replace - in order to deal with the problem, and I only followed this practical convention not created a new one.

That's a very weak argument for ignoring community conventions, but I'm not going to dig in my heels here.

kuangdash:
@purcell sincerely speaking, the community conventions should be revised. Either `/` or `.`  or others should be added in order to distinguish package-name-part from public-function-name. 

And I don't think that elisp needs `namespace`.

purcell:
Yes, it's _possible_ the conventions should be revised -- I'm not sure. Perhaps this is worth bringing up on the emacs-devel mailing list?

PR #3332, opened 2015-11-28
Add ghq recipe

rcoedo:
This package is an emacs interface for the [ghq](https://github.com/motemen/ghq) command. It uses `async-shell-command` to execute the actions. It also offers integration with helm. I made this small package to ease project navigation in emacs.

The package can be found [here](https://github.com/rcoedo/emacs-ghq)

### DIFF
--- a/recipes/ghq
+++ b/recipes/ghq
@@ -1,0 +1,1 @@
+(ghq :repo "rcoedo/emacs-ghq" :fetcher github)


syohex:
What's different from [helm-ghq](https://github.com/masutaka/emacs-helm-ghq) ?

rcoedo:
Even though I am giving a helm helper to list the project, I aim to implement most of ghq's features like fetching repositories with asynchronous shell commands.

syohex:
> Even though I implement a helm helper to list the projects, I aim to implement other ghq's features like fetching repositories, which are completely unrelated to helm.

I see.

purcell:
Thanks. In that case, `helm` should be a soft dependency, ie. installing `ghq` shouldn't automatically install `helm` too.

rcoedo:
@purcell Makes sense, I removed the dependency.

purcell:
Great, thanks.

PR #3333, opened 2015-11-29
Add java-imports

dakrone:
This is utility functions for adding `import` statements for Java. It
includes persistent caching of the packages for a class name so they
don't have to be re-entered every time.

### DIFF
--- a/recipes/java-imports
+++ b/recipes/java-imports
@@ -1,0 +1,3 @@
+(java-imports
+ :fetcher github
+ :repo "dakrone/emacs-java-imports")


purcell:
Thanks. Quick feedback:

- You're missing a `(provide ...)` form
- Every top-level symbol should have the `java-imports-` prefix, so symbols like `current-line-text` are not allowed
- You should add the missing package dependency on `(cl-lib "0.5")`

Cheers,
-Steve

dakrone:
Hi @purcell, thanks for taking a look. I changed all of the methods to begin with `java-imports-`, added `(provide ...)` form, and removed the `(require 'cl-lib)` statement entirely.

purcell:
Awesome, thanks.

PR #3335, opened 2015-11-30
change org-grep's repository

tarsius:
When I created the Emacsorphanage I intended for it to be used when the maintainer of some package stops responding and/or no longer wants to maintain it. In the case of `org-grep` the situation sadly is different - François has died.

I have fixed a minor issue in `org-grep` and made that available from the orphanage. If users report other issues, then I intend to help them fix them and will integrate submitted fixes.


milkypostman:
Sad news.

PR #3339, opened 2015-11-30
remove non-package ph

tarsius:
@gromnitsky do you intend to fix that? Otherwise I think @purcell should remove this.

### DIFF
--- a/recipes/ph
+++ b/recipes/ph
@@ -1,3 +1,0 @@
-(ph :repo "gromnitsky/ph"
-	:fetcher github
-	:files ("*.el" "meta.json" "README" ("bin" "bin/ph-project-new")))


gromnitsky:
what should I fix exactly? it still works (at least on my machine). why did it suddenly become a 'non-package'?

tarsius:
It slipped though review, usually @purcell would have required that you make changes before adding it to Melpa. The library completely lacks a header. To get started see http://www.gnu.org/software/emacs/manual/html_node/elisp/Library-Headers.html. You should also cleanup the whitespace, I recommend that you use `untabify` on the complete buffer. [This](https://github.com/gromnitsky/ph/blob/master/ph.el#L485) looks especially messy.

purcell:
Yep, it's not ideal, but there's a valid `-pkg.el` file. Ideally all the `.el` files would be tidy and well-formed, and I'd strongly encourage @gromnitsky to address that, but we don't need to remove this from MELPA. A well-formed `ph.el` would allow us to extract its `Commentary` section to display on the MELPA package page for `ph`, for example.

tarsius:
@gromnitsky Are you considering to make the suggested changes?

PR #3341, opened 2015-11-30
change rcirc-color's upstream

tarsius:
~~Or maybe it should be removed altogether in favor of `rcirc-styles`. https://github.com/kensanata/elisp/issues/3#issuecomment-160671535~~


kensanata:
My mistake. https://github.com/aaron-em/rcirc-styles.el is a replacement for rcirc-controls, not rcirc-color.

tarsius:
I'll open a separate pr for that.

PR #3342, opened 2015-11-30
remove rcirc-controls in favor of rcirc-styles

tarsius:
https://github.com/milkypostman/melpa/pull/3341#issuecomment-160677050 cc @kensanata, @aaron-em


milkypostman:
is there any reason one would want rirc-controls still? should both packages exist?

aaron-em:
I can't see any reason why; they both do the same thing, can't be used simultaneously, and rcirc-controls has some pretty intractable issues which prompted me to write rcirc-styles to begin with.

milkypostman:
thanks for confirming

PR #3345, opened 2015-11-30
Add selectric-mode recipe

CestDiego:
Make your Emacs sound like a proper typewriter. Extremely useful if you
have a puny, silent, rubberish, non-clicky keyboard.

Added recipe with consent from
[author](https://github.com/rbanffy/selectric-mode/issues/3)

### DIFF
--- a/recipes/selectric-mode
+++ b/recipes/selectric-mode
@@ -1,0 +1,3 @@
+(selectric-mode :repo "rbanffy/selectric-mode"
+                :fetcher github
+                :files (:defaults "*.wav"))


purcell:
Thanks. The `:files` spec here should be `(:defaults "*.wav")`. Can you please amend and force-push the recipe commit accordingly? Looks good otherwise, though there should be an `;;;###autoload` for the minor mode, and it would probably good to limit the number of `afplay` processes which might run in parallel. (This mode is probably very bad for battery life!)

CestDiego:
Forced push, and made comment on recomendations to author.

purcell:
Excellent, thanks.

rbanffy:
Wow! I had no idea it would be that simple! I'll put a limit on the number of concurrent processes on the next iteration (develop branch already has a small improvement).

CestDiego:
:+1:

PR #3350, opened 2015-12-01
Get some packages from the Emacsorphanage

tarsius:
I have moved all packages that Melpa used to get from the Emacsmirror to the Emacsorphanage. That's the new "upstream" and Melpa should get these packages from there. The Emacsmirror also mirrors these new "upstreams". I might even cleanup some of these packages a bit.


purcell:
Awesome, thanks.

PR #3352, opened 2015-12-02
Add jb-misc-functions

vapniks:
This package contains miscellaneous functions used by some other packages that I wrote or maintain (including `org-readme.el`).
Here is a link to the repo: https://github.com/vapniks/jb-misc-functions

### DIFF
--- a/recipes/jb-misc-functions
+++ b/recipes/jb-misc-functions
@@ -1,0 +1,4 @@
+(jb-misc-functions
+ :repo "vapniks/jb-misc-functions"
+ :fetcher github
+)
--- a/recipes/org-readme
+++ b/recipes/org-readme
@@ -1,1 +1,4 @@
-(org-readme :fetcher github :repo "mattfidler/org-readme")
+(org-readme
+ :repo "mattfidler/org-readme"
+ :fetcher github
+)


purcell:
Thanks. I'm pretty much -1 on packages which just collect a few personal functions without an overriding theme or purpose. I don't see any of these functions (or similar code) used in `org-readme` currently, and I don't know that anybody would find and start using this package, so I'm not convinced it's worth splitting out yet. Plus I noticed a few specific issues:

- The `jb-` prefix isn't great: there are only 26x26 possible 2-character prefixes, and there are at least a few thousand elisp packages floating around already
- Needs a package dependency on `(cl-lib "0.5")`
- Dependencies on `dash` and `keymap-utils` should be on stable versions, not snapshot versions

vapniks:
Sorry, I mean that it is a dependency of `auto-document.el` which is a soft dependency of `org-readme.el`. I know that `auto-document.el` doesn't have a MELPA recipe yet, but that was next on my list. I also have another package `one-key.el` that I plan on putting on MELPA at some point that will also use functions in `jb-misc-functions.el`.
I envisioned that there will be other occasions where some miscellaneous function is used in more that one of my packages, and hence the `jb-misc-functions.el` library. What do you do in those situations?

I can change the prefix to `jbloggs-` if you prefer.

vapniks:
OK, I want to get this sorted ASAP so I'll follow your advice and put the functions back in the dependent libraries for now.

PR #3344, opened 2015-11-30
Update editorconfig recipe

10sr:
I'm a contributor of editorconfig.

Currently I'm working on to add core files from [editorconfig-core](https://github.com/10sr/editorconfig-core-emacslisp) and [editorconfig-fnmatch](https://github.com/10sr/editorconfig-fnmatch-el) to editorconfig repository (https://github.com/editorconfig/editorconfig-emacs/pull/46).
When :point_up: PR is merged, this repository will have four el files.
But I'm going to remain these three packages separated, and I submitted this PR so that `editorconfig` package won't contain these new files even after https://github.com/editorconfig/editorconfig-emacs/pull/46 is merged.

https://github.com/editorconfig/editorconfig-emacs

PR #3355, opened 2015-12-03
Update editorconfig-core and editorconfig-fnmatch upstream

10sr:
I'm a contributor of EditorConfig.

Now these two packages are included in official repository.
(I'll mark my repositories as deprecated later)

PS: Can you also merge #3344 if there is no problem?

http://github.com/editorconfig/editorconfig-emacs


purcell:
Done, thanks.

10sr:
Thanks!

PR #3358, opened 2015-12-06
Add fxrd-mode recipe

msherry:
This package allows for editing/validation of fixed width record files, such as NACHA or TSO6 files, commonly used for financial transactions.

fxrd-mode can be found at https://github.com/msherry/fxrd-mode -- I am the author.

### DIFF
--- a/recipes/fxrd-mode
+++ b/recipes/fxrd-mode
@@ -1,0 +1,2 @@
+(fxrd-mode :fetcher github :repo "msherry/fxrd-mode"
+           :files (:defaults "fxrd-modes/*el"))


milkypostman:
this is good but it feels a bit off since you have packaged multiple packages into one. to me the naming suggests that it could be used for any fixed-width files but really it's two modes packaged into one.

can the package be used independently? would someone have to create their own submode though?

i'm wondering if this should be something like `financial-modes` or something like that.

msherry:
@milkypostman Thanks for your comments! I agree, it does seem a little odd to have multiple modes in this one package. I also plan on adding more -- the ones I have immediate need for are American Express response files (CBNOT, EPRAW, GFSG, GSFN, and SPOMI), and Mastercard RM37/RM39.

As you can see, there are a lot of different filetypes. I'd prefer not to release a new package for each one, since they all have the same basic functionality and differ only in their layouts -- each new mode doesn't add anything significant at this point.

As for the name, your suggestion makes sense. However, it's not *only* the financial industry that uses this type of file -- it's just the one I'm most familiar with, due to my work at [Balanced Payments](https://github.com/balanced). [Here](https://www.washington.edu/admin/adminsystems/fastrans/fixed.html) is another example I found that's still in use, and comments on [this stackoverflow question](https://stackoverflow.com/questions/7666780/why-are-fixed-width-file-formats-still-in-use) list a few other industries that use them that are not financial in nature. This mode is general enough to be used by those files as well, and I plan to accept pull requests that add new specs/validations (which will shortly be split out into separate files to reduce clutter).

Given this plan, how should I proceed? Obviously between the two of us you're the packaging expert :smile:, so I'd love to get advice from you, if possible.

milkypostman:
@msherry @purcell will have better advice but mine is to put them all in one package called `fixed-width-record-modes` where you have a shared library, and each format has its own mode.

I can understand your purpose here so that is why I suggest it. the name is long and feels unique enough to avoid complications but this is one case where I would ask @purcell for his opinion first. so maybe give him a couple days to respond.

msherry:
@purcell Any further advice on this? I've been steadily improving this package and using it myself in my day-to-day work, and I'd love to be able to make it available to others who might find it useful, as well.

I'm not opposed to changing the name if that's what it takes -- if it's published as `fixed-width-record-modes`, will it still be able to pull from the github repo https://github.com/msherry/fxrd-mode, or will that require a lot of renaming?

purcell:
I'm actually okay with this, but there are 2 things we should fix: firstly, don't check an `autoloads` file into your repo, because `package.el` will generate this for you, and MELPA will try to bundle `*.el`. Just make sure the necessary `;;;###autoload` cookies are in the source code. Secondly, `:files (:defaults ("fxrd-modes" "fxrd-modes/*el"))` in the recipe would be better written as `:files (:defaults "fxrd-modes")`, which is equivalent. Perhaps you could amend the commit accordingly, and ideally squash the two recipe commits together and force-push them to update the PR?

msherry:
@purcell Thanks for the tips. I've made the suggested changes, and squashed the commits as requested. Please let me know if there's anything else you'd like to see here!

msherry:
I'll wait for the broken package to be fixed and then try this build again.

purcell:
Thanks, merged. BTW, your `defgroup` should be `fxrd`, not `FXRD`.

PR #3359, opened 2015-12-06
Add company-sourcekit

nathankot:
* Summary: completion for xcode swift projects using SourceKit
* Repo: https://github.com/nathankot/company-sourcekit
* Association: author

This package includes two files but one is dependent on the other so
feels okay to have them in the same package

### DIFF
--- a/recipes/company-sourcekit
+++ b/recipes/company-sourcekit
@@ -1,0 +1,4 @@
+(company-sourcekit
+  :repo "nathankot/company-sourcekit"
+  :fetcher github
+  :files ("company-sourcekit.el" "sourcekit.el"))
--- a/recipes/sourcekit
+++ b/recipes/sourcekit
@@ -1,4 +1,4 @@
-(company-sourcekit
+(sourcekit
   :repo "nathankot/company-sourcekit"
   :fetcher github
-  :files ("company-sourcekit.el"))
+  :files ("sourcekit.el"))


milkypostman:
is there any reason that someone would want the `sourcekit` package independent of the `company-sourcekit` package? in that case we would want to split them up but I'm ok with this for now as long as that is _not_ the case.

nathankot:
I've decided to split it up into two packages :) Please see updated pull request

purcell:
This is really the least possible text that would satisfy `package.el`, and it would be easily improved:

```el
;;; sourcekit.el --- library to interact with sourcekitdaemon -*- lexical-binding: t -*-
;; Package-Requires: ((emacs "24.3") (dash "2.12.1") (dash-functional "1.2.0"))

;;; Commentary:

;;; Code:
```

Try creating a fresh `sourcekit.el`, then running `M-x auto-insert` -- this will help you generate a more complete boilerplate. Paste the `Package-Requires` line and original code in there, and fill out the commentary a little bit, and things will be much better.

Same re. the headers in `company-sourcekit.el`, except that there you're also missing package dependencies on `cl-lib` (or `emacs` 24.3, as in `sourcekit.el`), and `sourcekit` itself.

nathankot:
Thanks for the help @purcell, I've pushed those changes, please take a look and let me know if it's all good!

purcell:
Looks great - merged!

PR #3360, opened 2015-12-06
Add recipe for mb-url

dochang:


milkypostman:
the recipe you have here is broken, please make the package name in the recipe match the name of the file.

milkypostman:
wait, sorry, the problem is in the `mb-url-pkg.el` file. the dependency shouldn't have quotes around the package name. 

please test the recipe in a local client then you can ping in chat here; and rerun the travis test again.

dochang:
Fixed!  @milkypostman  Thank you!

milkypostman:
the only think that I thought about here which @purcell has been pretty diligent about is naming. while `mb-url` seems ok, my concern is conflicts with another package that may be `mb` which has a variable like `mb-url` or something. But in that case I guess that other package should rename.

milkypostman:
merging but may need to revisit the name.

PR #3362, opened 2015-12-06
Per-recipe adjustments to version parsing

dunn:
This adds a new optional `:version-regexp` key to recipes; it's a regular expression where the capture group is the string you want sent to `version-to-list`.  I added one to the Erlang recipe as an example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# MELPA 
+# MELPA
 
 [![Build Status](https://travis-ci.org/milkypostman/melpa.png?branch=master)](https://travis-ci.org/milkypostman/melpa)
 
@@ -183,6 +183,7 @@ the following form (`[...]` denotes optional or conditional values),
  :fetcher [git|github|gitlab|bitbucket|bzr|hg|darcs|fossil|svn|cvs|wiki]
  [:url "<repo url>"]
  [:repo "github-gitlab-or-bitbucket-user/repo-name"]
+ [:version-regexp "<regexp>"]
  [:module "cvs-module"]
  [:files ("<file1>" ...)])
 ```
@@ -230,13 +231,19 @@ the `git`-based fetchers.
 specifies the branch of the git repo to use. This is like `:commit`, but
 it adds the "origin/" prefix automatically.
 
+- `:version-regexp` is a regular expression for extracting a
+  version-string from the repository tags.  Version-strings must be
+  parseable by Emacs' `version-to-list` , so for an unusual tag like
+  "OTP-18.1.5", we add `:version "[^0-9]*\\(.*\\)"` to strip the
+  "OTP-" prefix.
+
 - `:module`
 specifies the module of a CVS repository to check out.  Defaults to to
 `package-name`.  Only used with `:fetcher cvs`, and otherwise ignored.
 
 - `:files` optional property specifying the elisp and info files used to build the
 package. Automatically populated by matching all `.el`, `.info` and `dir` files in the
-root of the repository and the `doc` directory. Excludes all files in the root directory 
+root of the repository and the `doc` directory. Excludes all files in the root directory
 ending in `test.el` or `tests.el`. See the default value below,
 
         ("*.el" "*.el.in" "dir"
@@ -504,7 +511,7 @@ in your `package-archives` list.
              '("melpa-stable" . "https://stable.melpa.org/packages/"))
 ```
 
-An online list of available packages can be found at 
+An online list of available packages can be found at
 [https://stable.melpa.org](https://stable.melpa.org).
 
 ### Stable Version Generation
@@ -526,8 +533,3 @@ package.
   them. Any packages you already have installed from MELPA will never
   get "updated" to the stable version because of the way version
   numbering is handled.
-
-
-
-  
-
--- a/package-build.el
+++ b/package-build.el
@@ -106,9 +106,15 @@ Certain package names (e.g. \"@\") may not work properly with a BSD tar."
   :type '(file :must-match t))
 
 (defcustom package-build-write-melpa-badge-images nil
-  "When non-nil, write MELPA badge images alongside packages, for use on github pages etc."
+  "When non-nil, write MELPA badge images alongside packages, for use on GitHub pages etc."
   :group 'package-build
   :type 'boolean)
+
+(defcustom package-build-version-regexp "^[rRvV]?\\(.*\\)$"
+  "Default pattern for matching valid version-strings within repository tags.
+The string in the capture group should be parsed as valid by `version-to-list'."
+  :group 'package-build
+  :type 'string)
 
 ;;; Internal Variables
 
@@ -153,6 +159,15 @@ function for access to this function")
   "Remove trailing whitespace from `STR'."
   (replace-regexp-in-string "[ \t\n]*$" "" str))
 
+
+(defun package-build--valid-version (str &optional regexp)
+  "Apply to STR the REGEXP if defined, \
+then pass the string to `version-to-list' and return the result, \
+or nil if the version cannot be parsed."
+  (when (and regexp (string-match regexp str))
+    (setq str (match-string 1 str)))
+  (ignore-errors (version-to-list str)))
+
 (defun package-build--parse-time (str)
   "Parse STR as a time, and format as a YYYYMMDD.HHMM string."
   ;; We remove zero-padding the HH portion, as it is lost
@@ -166,55 +181,26 @@ function for access to this function")
     (concat (format-time-string "%Y%m%d." time)
             (format "%d" (or (string-to-number (format-time-string "%H%M" time)) 0)))))
 
-(defun package-build--string-match-all (regex str &rest groups)
-  "Find every match for `REGEX' within `STR'.
-Return a list containing the full match string and match for
-groups `GROUPS'.  The return list is of the form
-   ((FULL GROUP1 GROUP2 ...) ...)
-where FULL is the complete regexp match and
-GROUP1, GROUP2, ... are the regex groups specified by the
-`GROUPS' argument.  If `GROUPS' is nil then FULL and GROUP1 will
-be identical."
-  (let (result
-        (pos 0)
-        (groups (or groups '(0))))
-    (while (string-match regex str pos)
-      (push (cons (match-string 0 str) (mapcar
-                                        (lambda (group)
-                                          (match-string group str))
-                                        groups))
-            result)
-      (setq pos (match-end 0)))
-    result))
-
 (defun package-build--find-parse-time (regex &optional bound)
-  "Find REGEX in current buffer and format as a time version, optionally looking only as far as BOUND."
+  "Find REGEX in current buffer and format as a time-based version string, \
+optionally looking only as far back as BOUND."
   (package-build--parse-time (progn (re-search-backward regex bound)
-                        (match-string-no-properties 1))))
+                                    (match-string-no-properties 1))))
 
-(defun package-build--valid-version-string (str)
-  "Return true if STR is a valid version, otherwise return nil."
-  (ignore-errors (version-to-list str)))
-
-(defun package-build--find-tag-version-newest (regex &optional bound &rest additional-groups)
-  "Find the newest version matching REGEX after point, maybe stopping at BOUND.
-The first capture group 1 is examined, together with any ADDITIONAL-GROUPS."
-  (let* ((text (buffer-substring-no-properties
-                (or bound (point-min)) (point)))
-         (tags (cl-remove-if-not
-                (lambda (tag-version)
-                  (package-build--valid-version-string (cadr tag-version)))
-                (apply 'package-build--string-match-all regex text 1 additional-groups))))
-    (car (nreverse (sort tags (lambda (v1 v2)
-                                (version< (cadr v1) (cadr v2))))))))
-
-(defun package-build--find-parse-time-latest (regex &optional bound)
-  "Find the latest timestamp matching REGEX, optionally looking only as far as BOUND."
-  (let* ((text (buffer-substring-no-properties
-                (or bound (point-min)) (point)))
-         (times (mapcar 'package-build--parse-time
-                        (mapcar 'cadr (package-build--string-match-all regex text 1)))))
-    (car (nreverse (sort times 'string<)))))
+(defun package-build--find-version-newest (regex &optional bound)
+  "Find the newest version matching REGEX before point, optionally stopping at BOUND."
+  (let ((tags (split-string
+                (buffer-substring-no-properties
+                  (or bound (point-min)) (point)))))
+    (setq tags (cl-remove-if nil (mapcar
+                                  (lambda (tag)
+                                    (when (package-build--valid-version tag regex)
+                                      (list (package-build--valid-version tag regex) tag)))
+                                  tags)))
+    ;; Returns a list like ((0 1) ("v0.1")); the first element is used
+    ;; for comparison and for `package-version-join', and the second
+    ;; (the original tag) is used by git/hg/etc.
+    (car (nreverse (sort tags (lambda (v1 v2) (version-list-< (car v1) (car v2))))))))
 
 (defun package-build--run-process (dir command &rest args)
   "In DIR (or `default-directory' if unset) run COMMAND with ARGS.
@@ -248,8 +234,7 @@ In turn, this function uses the :fetcher option in the CONFIG to
 choose a source-specific fetcher function, which it calls with
 the same arguments.
 
-Returns a last-modification timestamp for the :files listed in
-CONFIG, if any, or `package-build-default-files-spec' otherwise."
+Returns the package version as a string."
   (let ((repo-type (plist-get config :fetcher)))
     (package-build--message "Fetcher: %s" (symbol-name repo-type))
     (unless (eq 'wiki repo-type)
@@ -421,7 +406,7 @@ A number as third arg means request confirmation if NEWNAME already exists."
           (package-build--run-process nil "svn" "checkout" repo dir)))
         (apply 'package-build--run-process dir "svn" "info"
                (package-build--expand-source-file-list dir config))
-        (or (package-build--find-parse-time-latest "Last Changed Date: \\([0-9]\\{4\\}-[0-9]\\{2\\}-[0-9]\\{2\\} [0-9]\\{2\\}:[0-9]\\{2\\}:[0-9]\\{2\\}\\( [+-][0-9]\\{4\\}\\)?\\)" bound)
+        (or (package-build--find-parse-time "Last Changed Date: \\([0-9]\\{4\\}-[0-9]\\{2\\}-[0-9]\\{2\\} [0-9]\\{2\\}:[0-9]\\{2\\}:[0-9]\\{2\\}\\( [+-][0-9]\\{4\\}\\)?\\)" bound)
             (error "No valid timestamps found!"))))))
 
 
@@ -440,7 +425,8 @@ Return a cons cell whose `car' is the root and whose `cdr' is the repository."
     (with-current-buffer (get-buffer-create "*package-build-checkout*")
       (let ((root (package-build--trim (plist-get config :url) ?/))
             (repo (or (plist-get config :module) (symbol-name name)))
-            (bound (goto-char (point-max))))
+            (bound (goto-char (point-max)))
+            latest)
         (cond
          ((and (file-exists-p (expand-file-name "CVS" dir))
                (equal (package-build--cvs-repo dir) (cons root repo)))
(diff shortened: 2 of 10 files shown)


package-build.el
```diff
@@ -166,55 +192,27 @@ function for access to this function")
...
-                                (version< (cadr v1) (cadr v2))))))))
-
-(defun package-build--find-parse-time-latest (regex &optional bound)
-  "Find the latest timestamp matching REGEX, optionally looking only as far as BOUND."
-  (let* ((text (buffer-substring-no-properties
-                (or bound (point-min)) (point)))
-         (times (mapcar 'package-build--parse-time
-                        (mapcar 'cadr (package-build--string-match-all regex text 1)))))
-    (car (nreverse (sort times 'string<)))))
+(defun package-build--find-version-newest (regex &optional bound)
```
dunn: This function replaces `package-build--find-tag-version-newest` and `package-build--find-parse-time-latest`.

dunn:
Tested by building all the A recipes in stable mode, and most of them in unstable; I didn't see any changes in versioning.

milkypostman:
this is amazing! Thanks! Will take a look at this in the next couple days.

README.md
```diff
@@ -183,6 +183,9 @@ the following form (`[...]` denotes optional or conditional values),
  :fetcher [git|github|gitlab|bitbucket|bzr|hg|darcs|fossil|svn|cvs|wiki]
  [:url "<repo url>"]
  [:repo "github-gitlab-or-bitbucket-user/repo-name"]
+ [:commit "<ref>"]
```
milkypostman: can you remove `:commit` and `:branch` here because they are kind of hacks that I actually wish we would remove. I don't want to advertise them. there are cases where we have made exceptions to this rule but in general we avoid these things like the plague. If there is ever a time that no recipe uses either I will remove the functionality.

package-build.el
```diff
@@ -501,17 +497,18 @@ Return a cons cell whose `car' is the root and whose `cdr' is the repository."
...
-                                        "^\\(?:v[.-]?\\)?\\([0-9]+[^ \t\n]*\\)$" bound)
-                                       (error
-                                        "No valid stable versions found for %s"
-                                        name)))))
-            ;; Using reset --hard here to comply with what's used for
-            ;; unstable, but maybe this should be a checkout?
-            (package-build--update-git-to-ref dir (concat "tags/" (car tag-version)))
-            (cadr tag-version))
+        (let* ( (bound (goto-char (point-max)))
+                (regexp (or (plist-get (package-build--config name) :version)
```
milkypostman: don't need `package-build--config` here because the `config` variable should already contain the config hash. 

thus you can use `(plist-get config :version)`.

package-build.el
```diff
@@ -769,7 +781,10 @@ Optionally PRETTY-PRINT the data."
               (lambda (elt)
                 (unless (symbolp (car elt))
                   (error "Invalid package name in dependency: %S" (car elt)))
-                (list (car elt) (version-to-list (cadr elt))))
+                (list
```
milkypostman: I don't actually think this line should use the new regexp. The regexp is for *our* packages but the dependencies shouldn't be effected by the regexp.

README.md
```diff
@@ -183,6 +183,9 @@ the following form (`[...]` denotes optional or conditional values),
...
+ [:branch "<branch-name>"]
+ [:version "<regexp>"]
```
milkypostman: i think maybe we should pick something like `:version-regexp` or `:versionregexp` as a name, because `:version` makes me fear we're going to get some PRs where people will put in `2.5.4` or something.

package-build.el
```diff
@@ -1080,9 +1094,10 @@ and a cl struct in Emacs HEAD.  This wrapper normalises the results."
 
     (package-build--message "\n;;; %s\n" file-name)
     (let* ((version (package-version-join
-                     (version-to-list
+                     (package-build--version-to-list
```
milkypostman: I don't think we need `package-build--version-to-list` here either. Because the *checkout* should return us a valid version and we should error if it doesn't. Because that means the regexp is broken.

package-build.el
```diff
@@ -166,55 +192,27 @@ function for access to this function")
...
-                        (mapcar 'cadr (package-build--string-match-all regex text 1)))))
-    (car (nreverse (sort times 'string<)))))
+(defun package-build--find-version-newest (regex &optional bound)
+  "Find the newest version matching REGEX after point, starting at BOUND or `point-min'."
+  (let ((tags (split-string
+                (buffer-substring-no-properties
+                  (or bound (point-min)) (point)))))
+    (setq tags (cl-mapcar
+                 (lambda (tag)
+                   (if (package-build--valid-version-string tag regex)
```
milkypostman: if -> when

package-build.el
```diff
@@ -153,6 +159,26 @@ function for access to this function")
...
+then pass the string to `version-to-list' and return the result."
+  (when regexp
+    (string-match regexp str)
+    (setq str (match-string 1 str)))
+  (version-to-list str))
+
+(defun package-build--valid-version-string (str regexp)
+  "Parse STR with REGEXP.
+Return the result of `package-build--version-to-list', or nil if invalid"
+  (ignore-errors (package-build--version-to-list str regexp)))
```
milkypostman: Since I don't hink we need `package-build--version-to-list` as an independent function, you can inline it here.

package-build.el
milkypostman: Still double checking the logic here to remember how I setup this time parsing business (it's been a while) but I think you can remove `package-build--find-parse-time` too.

milkypostman: ok, I see my confusion here. alright. so here is my curiosity; can we change this function so that it returns a tuple `('tag name' (0 1 2))` which is essentially `(<original string> <version parsed out of string>)`.

Because *then* we don't need to worry about re-parsing in the `package-build-package` function. it should also simplify things.

and it's not hard to change. Anything that uses this function (e.g., cvs and svn) checkouts, can just grab the `cadr`? What do you think?

i think this is close enough that if you don't want to change it up I am willing to patch in and fix up, but I'm more than happy to have your help  :)

package-build.el
milkypostman: OK, leaving the previous comment but I get why you are doing this here. I think we should drop it based on my comments above. The *checkout* functions should basically get the tag name and check that tag out, but should still return a valid version. I would prefer that the rest of code not ever need to worry about the version regexp.

milkypostman:
THANKS FOR THIS PR! I added a number of comments but I really wanted to say this is a nice patch and great work. Just needs a bit of polish I think. Sorry if any comments come off short or anything I appreciate your work.

package-build.el
dunn: Oh, right.

package-build.el
dunn: Right now I've got the checkouts (at least git and hg) returning the original version-string.  `package-build--find-version-newest` uses the version list internally to figure out which tag is newest, but then returns the original tag as a string so git and hg can actually check out that release.  But if SVN, darcs, etc., return the actual list, then I need to change this.

Should `package-build--find-version-newest` return something like `'((0 1) . "v0.1")`?  Then git/hg have access to the tag string for checkouts, but the checkout function itself could just return the `car`.

dunn: Oops, didn't refresh before posting.

README.md
dunn: Oh, yeah, OK. `:version-regexp` sounds good.

dunn:
No problem! Thank you for the review!

package-build.el
milkypostman: Yes exactly. It took me a second to figure out why we had the two functions to begin with and I'm still not sure I know but they're getting consolidated anyhow. I'm not sure how it'll work though with something like hg where you kinda need two regex. One to pull the tag and one to pull the version from the tag. However I think we can hardcore the "tag" regex and then use version-regexp to parse the version from the tag. ?

dunn:
OK, added the changes as a new commit to make reviewing easier, I can combine them when it's ready to merge.

Other commits:
- I made `tests/version-tests.el` to check all packages for version changes. This might not be useful in the future, so feel free to drop it.
- a few packages have tags like` v.20150303`, which now gets parsed as `(0 v.20150303)` without special handling.  We could instead change the default regexp to `"^[rRvV]?\\.?\\(.*\\)$"`, but that would cause `.5` to be parsed as `(5)` instead of `(0 5)`.  Is there a way to check for `[rRvV] or [rRvV]\\.` without using a capture group?
- fixed yatex's version parsing; previously the only valid tag was `v0`.

Five other packages will have different versions after this is merged:
- **cfengine-code-style** is now `(3 8 0)` instead of `(3 8 0 2 1)`, which is correct: https://github.com/cfengine/core/releases. I think it was previously picking up one of the beta releases.
- **ecb** is now `(2 -4 24)` instead of `(2 -3 24)`.  "Snapshot" has always been parsed as `-4` so I'm not sure how it was getting `-3`.
- **epl** is now `(0 7)` instead of `(71 4 8098)`.  I think this is right, and we should drop a few :+1:s into https://github.com/cask/epl/issues/30#issue-121963688.
- **roguel-ike** is now `(0 1)` instead of `(0 1 0)`.  Those are equivalent according to `version=`; but there's a tag for each so I think it's basically random which one gets picked.
- **tablist** doesn't have any stable versions: https://github.com/politza/tablist/issues/2; was there previously a `0.60` tag? How did the current stable package get created?

package-build.el
```diff
@@ -163,58 +184,31 @@ function for access to this function")
...
+  "Find the newest version matching REGEX after point, starting at BOUND or `point-min'."
+  (let ((tags (split-string
+                (buffer-substring-no-properties
+                  (or bound (point-min)) (point-max)))))
+    (setq tags (cl-mapcar
+                 (lambda (tag)
+                   (when (package-build--valid-version-string tag regex)
+                     ;; Returns a list like ((0 1) ("v0.1")); the
+                     ;; first element is used for determining the
+                     ;; newest version and the second (the original
```
dunn: Forgot to update the comment here; will do.

steckerhalter:
@milkypostman I saw you have made a comment about removing :commit and :branch from package-build. Since quelpa uses these and wants users to be able to use them in the future as well, do you think there is a way to keep them in some way?

milkypostman:
Yeah I don't think it's bad to keep them but I definitely don't want to
document them or endorse them for our recipes.

We have to comment a lot on cls using branch.

vyp:
@milkypostman Yep that's fine, I assume you are okay with mentioning these in the quelpa readme?

steckerhalter:
@milkypostman sounds good. thanks a bunch

milkypostman:
Yeah np mentioning in quelpa.

purcell:
Yeah, we should keep mentioning them in `package.el` docstrings if applicable, but not in the MELPA README.

package-build.el
```diff
@@ -589,17 +581,33 @@ Return a cons cell whose `car' is the root and whose `cdr' is the repository."
...
-                                        "^\\(?:v[.-]?\\)?\\([0-9]+[^ \t\n]*\\)[ \t]*[0-9]+:\\([[:xdigit:]]+\\)$"
-                                        bound
-                                        2)
-                                       (error
-                                        "No valid stable versions found for %s"
-                                        name)))))
-            (package-build--run-process dir "hg" "update" (nth 2 tag-version))
-            (cadr tag-version))
+        (let ( (bound (goto-char (point-max)))
+               (regexp (or (plist-get (package-build--config name) :version-regexp)
```
milkypostman: still think you can remove the `package-build--config` here. please use the `config` argument directly.

steckerhalter:
@dunn I tried the changes made in this PR. so `package-build-checkout` is now returning a list instead of a string like before, is that correct?

dunn:
@steckerhalter Yeah; does that break anything?

package-build.el
dunn: Woops, thought I fixed that.  Maybe got lost in a rebase.

dunn: Oh, I changed the one in `package-build--checkout-git` and forgot to do the same here.

steckerhalter:
@dunn kind off... in quelpa we use the version string to do some more things after the checkout. I tested with your changes and it failed so that's how I noticed. I think it shouldn't be a problem to change it on our side, we would just be glad to know when it is merged so that we can push out the change.

dunn:
Either way works for me; the version-as-list or the version string.

dunn:
Looks like I didn't test non-stable well enough:
```
 • Building recipe applescript-mode ...
emacs --no-site-file --batch -l package-build.el --eval "(let ((package-build-stable nil) (package-build-write-melpa-badge-images t) (package-build-archive-dir (expand-file-name \"./packages\" package-build--this-dir))) (package-build-archive 'applescript-mode))"

;;; applescript-mode

Fetcher: svn
Source: http://svn.osdn.jp/svnroot/macemacsjp/applescript-mode/trunk

Cloning http://svn.osdn.jp/svnroot/macemacsjp/applescript-mode/trunk to /Users/cat/Documents/clones/melpa/working/applescript-mode/
File: /Users/cat/Documents/clones/melpa/packages/applescript-mode-2015snapshot12snapshot15.entry
Built in 1.707s, finished at Tue Dec 15 11:27:54 2015
 ✓ Wrote 40 -rw-r--r--  1 cat  staff    16K Dec 15 11:27 ./packages/applescript-mode-2015snapshot12snapshot15.el
 8 -rw-r--r--  1 cat  staff   247B Dec 15 11:27 ./packages/applescript-mode-2015snapshot12snapshot15.entry
 8 -rw-r--r--  1 cat  staff   775B Dec 15 11:27 ./packages/applescript-mode-badge.svg
 8 -rw-r--r--  1 cat  staff    17B Dec 15 11:27 ./packages/applescript-mode-readme.txt
```

dunn:
New commit that I think fixes SVN and CVS, but all of the CVS repos were timing out for me so if someone else wants to check that'd be :100:

package-build.el
```diff
@@ -589,17 +582,33 @@ Return a cons cell whose `car' is the root and whose `cdr' is the repository."
...
-                                   (or (package-build--find-tag-version-newest
-                                        "^\\(?:v[.-]?\\)?\\([0-9]+[^ \t\n]*\\)[ \t]*[0-9]+:\\([[:xdigit:]]+\\)$"
-                                        bound
-                                        2)
-                                       (error
-                                        "No valid stable versions found for %s"
-                                        name)))))
-            (package-build--run-process dir "hg" "update" (nth 2 tag-version))
-            (cadr tag-version))
+        (let ( (bound (goto-char (point-max)))
```
milkypostman: What am I missing here: we set the `bound` variable by going to the max point which I assume will return `(point-max)`. But then below we call `goto-char` again?

steckerhalter:
@dunn then it would be simpler for us to keep the behavior as it was (with strings)

package-build.el
dunn: I think what's happening is 
- we set `bound` and point to `(point-max)`
- `((package-build--run-process dir "hg" "tags")` inserts the output of `hg tags` into the buffer before point.  At this stage point is again at `(point-max)`.
- then we need to move point back to `bound` in order to do the find-and-replace

milkypostman: Oh of course yes.

package-build.el
dunn: I _did_ update this comment, but I must have lost it in a rebase :see_no_evil:

package-build.el
dunn: It was my fault, I didn't fully understand how the buffer manipulation worked.

dunn:
New commit:
- `package-build-checkout` returns the version as string instead of the raw list
- I fixed some changes I made to docstrings, since I was confused about how buffer manipulation worked during the checkouts.  I saw that `bound` was before point but docstrings referred to searching "up to `bound`", and I thought it was a mistake.  I didn't realize it meant searching backwards.

milkypostman:
But searching backwards is ok I think right?

I will have a look tomorrow. Taking the night off from the keyboard.

dunn:
> But searching backwards is ok I think right?

Yep! So I restored that behavior.

dunn:
`blank-mode` is getting build with an older timestamp than is on MELPA:
```
Test persistent-versions condition:
    (ert-test-failed
     ((should
       (equal
        (version-to-list ...)
        (package-desc-version ...)))
      :form
      (equal
       (20130824 459)
       (20141029 816))
```
But the last-modified time in the headers is from 2013, not 2014:
```
HTTP/1.1 200 OK
Server: nginx/1.4.6 (Ubuntu)
Date: Thu, 17 Dec 2015 22:24:41 GMT
Content-Type: text/plain; charset=utf-8
Connection: keep-alive
Vary: Accept-Language
Last-Modified: Sat, 24 Aug 2013 11:59:25 GMT
X-EmacsWiki-Source: Apache/CGI
X-Proxy-Cache: HIT
```

package-build.el
```diff
@@ -1096,7 +1119,7 @@ and a cl struct in Emacs HEAD.  This wrapper normalises the results."
       (package-build--message "Built in %.3fs, finished at %s"
                   (time-to-seconds (time-since start-time))
                   (current-time-string))
-      file-name)))
+      (list file-name version))))
```
dunn: This change is only for the tests, so if those are rejected this should be too (which is why they're in the same commit).

package-build.el
```diff
@@ -463,11 +455,30 @@ Return a cons cell whose `car' is the root and whose `cdr' is the repository."
                             "-d" target-dir repo))))
         (apply 'package-build--run-process dir "cvs" "log"
                (package-build--expand-source-file-list dir config))
-        (or (package-build--find-parse-time-latest "date: \\([0-9]\\{4\\}-[0-9]\\{2\\}-[0-9]\\{2\\} [0-9]\\{2\\}:[0-9]\\{2\\}:[0-9]\\{2\\}\\( [+-][0-9]\\{4\\}\\)?\\)" bound)
-            (package-build--find-parse-time-latest "date: \\([0-9]\\{4\\}/[0-9]\\{2\\}/[0-9]\\{2\\} [0-9]\\{2\\}:[0-9]\\{2\\}:[0-9]\\{2\\}\\);" bound)
-            (error "No valid timestamps found!"))
-        ))))
 
+        ;; `cvs log` does not provide a way to view the previous N
```
dunn: I realized the `package-build--find-parse-time-latest` function was still being used by the CVS checkout function, so that's why I added this.

package-build.el
```diff
@@ -153,6 +159,21 @@ function for access to this function")
   "Remove trailing whitespace from `STR'."
   (replace-regexp-in-string "[ \t\n]*$" "" str))
 
+
+
+(defun package-build--config (package)
```
milkypostman: sorry i keep harping on this but this function should go away and the code should just live in the one place where it is used.

milkypostman:
added one last comment and I have one more thing to ask (even though I do feel slightly bad asking it given how much you have done): could you squash these commits into two or three such that they are,

- made the changes to package-build to support version regex
- updated recipes


Then when you are done with your final push, I will take and run this against all the recipes just to see that we are the same on MELPA as we are now.

Then I will take and do the same with MELPA stable.  I will spin up a clone of our servers and just run it there (with your patch of course) to confirm it's not broken anything.

milkypostman:
Also, if you are busy and all that I can do the squashing and reworking of the commits, it's just easier if I can hit merge here after I verify everything.

package-build.el
dunn: Oh no problem, I didn't realize it was down to one instance.

dunn:
The recipe changes shouldn't be one commit per recipe?

milkypostman:
they don't need to be. that's fine if you want to. mainly I just wanted to clean up all the commits to package-build. but you can collapse the recipe changes too if youw ant.

dunn:
If you're fine either way I'll leave the recipe commits separate.  What about the ERT tests; do you want those or should I drop them?

milkypostman:
i don't think so!  More tests the better!!!

dunn:
OK, squashed!

milkypostman:
Will get to this tomorrow testing. Will need some time don't have it
tonight.

purcell:
Thanks @dunn, and thanks @milkypostman for helping with all this -- I'm pretty slammed at the moment.

milkypostman:
@dunn looks good on my test server! i will submit first thing tomorrow morning. don't want to sleep through the night and have anything accidentally break.

milkypostman:
THANKS! WE MADE IT!!!

purcell:
Awesome work folks. Thanks so much for all your @dunn!

dunn:
No problem! Let me know if anything breaks down the line.

mdorman:
I notice, though, that epl on melpa-stable still has the wonky version number that's mentioned near the beginning of this discussion.  Do those (and perhaps the other packages that got recipe tweaks) need to be manually kicked to get them corrected?

dunn:
@mdorman yeah, looks like it. cc @milkypostman, the packages whose versions dropped are https://github.com/milkypostman/melpa/pull/3362#issuecomment-164361596

milkypostman:
@mdorman i don't think they are "wonky" just that they will have different numbers now, but @dunn they are correct right? or do we need to update the regex?

dunn:
The problem is that a few packages now have lower versions than previously, like EPL: http://stable.melpa.org/#/epl

And they aren't being made available since the build scripts think they're older.

milkypostman:
this I can fix. later tonight I will cause a complete rebuild. some
packages will be missing for a while but *shrug* its currently the only way
to make things right in the world. sorry I don't have a better answer but I
haven't spent much time in the stable world.

dunn:
That's fine, thanks so much!

milkypostman:
everything rebuilt. do the versions look as you expect?

dunn:
All perfect except http://stable.melpa.org/#/cfengine-code-style; when I build stable locally it's 3.8.0 which I think is what it's supposed to be: https://github.com/cfengine/core/releases

PR #3365, opened 2015-12-07
add recipe dmenu

lujun9972:


purcell:
Thanks, merged. You might consider switching from `cl` to the more modern `cl-lib`, and adding a `Package-Requires: ((cl-lib "0.5"))` header.

lujun9972:
I'll check it. thanks

PR #3366, opened 2015-12-08
org-babel support for restclient.el

alf:
Since restclient.el does not currently support org-mode and org-babel, I've written this extension. It's a trivial thing, but it's useful for those who need it.


purcell:
Thanks, looks good. Please amend the recipe commit to remove the `:files` spec - which is unnecessary here - and then force-push it back here.

alf:
Sorry I didn't get around to this. Too many things on my plate and it slipped my mind. Thanks for merging anyways. I'll keep the :files spec-thing in mind for next time.

Let me know if you still want me to do anything here.

PR #3367, opened 2015-12-09
Add drawille

sshbio:
Emacs lisp version of [drawille](https://github.com/asciimoo/drawille) library, which permit to display graphics using braille characters.

My version can convert matrices or strings to drawille representations:
```
######                                              
#     # #####    ##   #    # # #      #      ###### 
#     # #    #  #  #  #    # # #      #      #      
#     # #    # #    # #    # # #      #      #####  
#     # #####  ###### # ## # # #      #      #      
#     # #   #  #    # ##  ## # #      #      #      
######  #    # #    # #    # # ###### ###### ###### 
```

`M-x drawille-buffer RET`

```
⡏⠉⠉⡆⡖⠒⢢⢀⠔⠢⡀⡆⠀⢰⢰⢰⠀⠀⠀⡆⠀⠀⢰⣒⣒⠂
⠧⠤⠤⠃⠏⠉⠣⠸⠉⠉⠇⠗⠉⠺⠸⠸⠤⠤⠄⠧⠤⠤⠸⠤⠤⠄
```

### DIFF
--- a/recipes/drawille.el
+++ b/recipes/drawille.el
@@ -1,3 +1,1 @@
-(drawille
- :fetcher github
- :repo "sshbio/elisp-drawille")
+(drawille :repo "sshbio/drawille" :fetcher github)
--- a/recipes/drawille.el
+++ b/recipes/drawille.el
@@ -1,0 +1,3 @@
+(drawille
+ :fetcher github
+ :repo "sshbio/elsip-drawille")
--- a/recipes/drawille.el
+++ b/recipes/drawille.el
@@ -1,3 +1,3 @@
 (drawille
  :fetcher github
- :repo "sshbio/elsip-drawille")
+ :repo "sshbio/elisp-drawille")


recipes/drawille.el
```diff
@@ -0,0 +1,3 @@
+(drawille
+ :fetcher github
+ :repo "sshbio/elsip-drawille")
```
syohex: Typo `elsip-drawille` -> `sshbio/elisp-drawille`

syohex:
You should fix following part. I cannot send PR because git treat your package as binary file. I suppose git misjudges by ascii art part.

## package header at first line

#### Original

https://github.com/sshbio/drawille/blob/master/drawille.el#L1

#### Fixed version

```lisp
;;; drawille.el --- Drawille implementation in elisp
```

## cl.el

Because your package uses cl-lib macros/functions instead of cl.el such as `cl-dotimes`, `cl-loop`.

#### Original

https://github.com/sshbio/drawille/blob/master/drawille.el#L35

#### Fixed version

```
(require 'cl-lib)
```

And add package dependency at header part.

```
;; Package-Requires: ((cl-lib "0.5"))
```

sshbio:
Thank you much for these indications: This is my first package, and I did not already know how Melpa packages were built.

I did my best to make the package match these criteria.  Let see how the build occur...

About the repo name, any preference to name an elisp package (append ".el", "-el" prepend "emacs-", "lisp-"...)?

Thank you for your precious help!

purcell:
Looks good to me. We don't have any preferences re. repo names - the current one is fine.

sshbio:
Thank you @purcell!  Roger that.

recipes/drawille.el
sshbio: I read the syntax guide, then I have no excuse! :stuck_out_tongue:

wasamasa:
> Because your package uses cl-lib macros/functions instead of cl.el such as cl-dotimes, cl-loop.

This isn't quite correct.  `dotimes` is a built-in macro in `subr.el`.  `cl-dotimes` is a thing, but is only necessary if you need `cl-block` features, such as `cl-return` for breaking out of the loop.

sshbio:
Thank you @wasamasa.

@syohex I will use cl-loop in a next version, so I will add this dependency (`cl` or `cl-lib`) all the same.

PR #3369, opened 2015-12-10
Fix the filename for drawille package to unbreak build

msherry:
The build is currently broken due to a package name mismatch. This unbreaks it.


purcell:
My fault for missing that, sorry. Thanks for the fix.

PR #3371, opened 2015-12-11
Add minizinc-mode

m00nlight:
Add major mode for editing MiniZinc(http://www.minizinc.org) model file.


purcell:
Thanks. Please amend the recipe commit to remove the unnecessary `:files` spec, and then force-push it back here to update the PR.

m00nlight:
@purcell  Done with removing files fields in recipe

PR #3372, opened 2015-12-11
Modify xterm-keybinder recipe

yuutayamada:
Hi, I'm author of this package.
I added a configuration to use this package from urxvt.
(this package allows you to use unusable keybinds in terminal Emacs)


purcell:
Thanks.

PR #3373, opened 2015-12-11
Add org-themis recipe to melpa

zellio:
`org-themis` extends `org-mode` to handle multiple journal files to facilitate grouping tasks and notes under a package heading. It provides the `org-themis-minor-mode` which allows for creation, management, and manipulations of "projects". Projects in this context are a collection of files and directories.

The package can be found at https://github.com/zellio/org-themis

I am the author and maintainer of this package.

### DIFF
--- a/recipes/org-themis
+++ b/recipes/org-themis
@@ -1,0 +1,2 @@
+(org-themis :fetcher github
+            :repo "zellio/org-themis")


vyp:
You don't actually need the `:files` part in the recipe since this is done automatically. See the `:files` section in this part of the readme: https://github.com/milkypostman/melpa#recipe-format

zellio:
@vyp - thanks. I've correct that and updated the branch pointer to point to the "master" branch as I assume the github fetcher grabs the default branch from the repo.

purcell:
We want to track the default branch where development takes place, so it looks like you should remove the `:branch` option here. You can tag the code with version numbers to control what packages get built in MELPA Stable, while MELPA will always package your development versions. If possible, please squash the recipe commit(s) into a single commit and force-push it back here to help us keep our commit history tidy and atomic.

purcell:
P.S. A `Commentary` section is sorely needed here -- it will be displayed on the MELPA package page (and in Emacs when the user hits `?` in the package menU0, and without it there's no context for potential package users.

zellio:
@purcell - that makes more sense. Sorry for the added work, I'll get that cleaned up tomorrow.

zellio:
/pong - cleaned up the commit. Is now one and just the lines requested.

PR #3374, opened 2015-12-11
add recipes/golden-ratio-scroll-screen

jixiuf:
Scroll half screen down or up, and highlight current line before
or after scrolling


purcell:
Thanks - merged!

PR #3375, opened 2015-12-11
add recipes/dired-filetype-face

jixiuf:
Set different faces for different filetypes in dired

### DIFF
--- a/recipes/dired-filetype-face
+++ b/recipes/dired-filetype-face
@@ -1,0 +1,2 @@
+(dired-filetype-face :repo "jixiuf/dired-filetype-face" :fetcher github)
+


purcell:
Thanks. Needs a couple of fixes before merging:

Firstly, `i__d__f` isn't a valid symbol name: a unique package prefix should be included in all top-level symbols, as per the [elisp conventions](https://www.gnu.org/software/emacs/manual/html_node/elisp/Coding-Conventions.html).

Secondly, these autoloads are not permitted, because they will automatically load the package with dired whenever it is installed, not just when the user wants to use it:
```el
;; Append to mode hooks so ours are the last applied, overriding others.
;;;###autoload(add-hook 'dired-mode-hook 'dired-filetype-setup 'append)
(add-hook 'dired-mode-hook 'dired-filetype-setup 'append)
;;;###autoload(add-hook 'wdired-mode-hook 'dired-filetype-setup 'append)
(add-hook 'wdired-mode-hook 'dired-filetype-setup 'append)
```
You should either tell the user to `require` the file when necessary, or provide either an autoloaded "enable" command, or a global/local minor mode they can enable as desired.

jixiuf:
autoload cookie is removed . 
and i_d_f is renamed .

purcell:
Thanks

PR #3376, opened 2015-12-11
add bitbucket fetcher

mmckinst:
defaults to using `hg` since thats what bitbucket is mostly used for. bitbucket.com can also do git and the recipes using bitbucket.com+git (aria2, ert-junit, igv, jack-connect) still work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,9 +180,9 @@ the following form (`[...]` denotes optional or conditional values),
 
 ```lisp
 (<package-name>
- :fetcher [git|github|gitlab|bzr|hg|darcs|fossil|svn|cvs|wiki]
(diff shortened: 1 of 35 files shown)


purcell:
Fantastic work - thanks for this!

tarsius:
> recipes using bitbucket.com+git (aria2, ert-junit, igv, jack-connect) still work

I had difficulties understanding how cloning a git repository using hg could possibly work. Turns out the recipes of the mentioned packages were not changed, they still use the `git` fetcher. Mentioning this in case someone else gets confused about this too.

purcell:
Yeah, I missed that. I can see how that would be confusing. Noted in the README in 2e0405a.

PR #3377, opened 2015-12-12
Remove flycheck-processing

lunaryorn:
The Flycheck syntax checker for processing has become part of Flycheck itself, see flycheck/flycheck#812 and https://github.com/ptrv/processing2-emacs/issues/11.  /cc @ptrv


purcell:
Great, thanks.

Issue #2955, opened 2015-07-22
Evaluating a new version numbering scheme


purcell:
Requests have been made in #2944 and elsewhere for a version numbering scheme in which version numbers for MELPA packages are based on upstream version numbers. I summarised the barriers to making such a major change in a [lengthy comment there](https://github.com/milkypostman/melpa/issues/2944#issuecomment-122603115). This ticket reproduces the concrete plan I outlined there for evaluating a new version numbering scheme, updated with [feedback](https://github.com/milkypostman/melpa/issues/2944#issuecomment-122604242) from @dgutov.

---

I'd propose that someone should actually implement the corresponding changes in a branch, do a full build locally, and then audit the results for problems, such as uninstallable packages. If we still end up with a set of packages which is usable and sane, then switching over to this new scheme *might* be an option. If we determine that MELPA would no longer work well for MELPA users, or would be harder for us to maintain, then changing the official scheme would be out of the question.

As I see it, the logic for determining the snapshot version of a package should look like this:

- Determine a reference upstream version of the package:
  - Take the version in the `-pkg.el` file, if any, or
  - Take the version in the `Package-Version` or `Version` header of the main `.el` file, if any, or
  - Take the most recent numeric version tag for the repo, if any, or
  - Default to 0.0.0.0
- Increment the upstream version so the MELPA version will replace it:
  - For upstream versions without a suffix (`alpha`, `-cvs` etc), increment them by a micropoint, e.g. `0.8` => `0.8.0.1`
  - For upstream versions with a pre-release suffix (`alpha`, `pre` etc.), append the timestamp, i.e. turn `0.24-alpha2` into `0.24-alpha2.201506061234`
  - For snapshot upstream versions (`-cvs`, `-git`, `-snapshot`), strip the suffix and keep the number, so `0.24-cvs` => `0.24`
- Append a snapshot qualifier:
  - Determine the timestamp, using the existing method (this is very robust)
  - Use `-gitTIMESTAMP` or `-snapshotTIMESTAMP` as the version suffix

To audit these changes, we would then need to build all 2500+ packages locally, and review the package list using a recent `package.el` with only GNU ELPA and that local package set enabled: all packages should be listed as installable, and all the packages for which the versions defaulted to `0.0.0.1-...` should be reviewed.

[...]

PR #3378, opened 2015-12-12
w3m: fetch from GitHub

ttuegel:
The original anonymous CVS repository is very unreliable. Notably, the `shimbun` package from the same repository was already using `emacswirror/w3m`.

### DIFF
--- a/recipes/w3m
+++ b/recipes/w3m
@@ -1,4 +1,3 @@
-(w3m :url ":pserver:anonymous@cvs.namazu.org:/storage/cvsroot"
-     :module "emacs-w3m"
-     :fetcher cvs
+(w3m :repo "emacsmirror/w3m"
+     :fetcher github
      :files (:defaults "icons" (:exclude "octet.el" "mew-w3m.el" "w3m-xmas.el")))


purcell:
Hmm, we're moving away from using Emacsmirror here at the suggestion of its maintainer, @tarsius. Any thoughts on this, Jonas?

tarsius:
The repository on the Emacsmirror is no longer updated from the upstream cvs repository. But I don't think there is any development going on, so that doesn't matter.

I have moved the package to the [Emacsorphanage](https://github.com/emacsorphanage/w3m) in the hope that someone would volunteer to maintain it. But it's unlikely that this will happen. I've also added a little patch.

The Emacsmirror treats the repository on the Emacsorphanage as the "upstream". If someone makes minor changes and I merge that, then these changes will be mirrored after a delay (I don't import every day).

Never-the-less Melpa should use the repository in the orphanage instead of that on the mirror as the upstream, for both `w3m` and `shimbun`.

ttuegel:
> The repository on the Emacsmirror is no longer updated from the upstream cvs repository.

I don't think I've been able to update from upstream CVS in months, so that goes without saying :smiley_cat:

> Never-the-less Melpa should use the repository in the orphanage instead of that on the mirror as the upstream, for both w3m and shimbun.

I will prepare a new PR to do both. Thanks for the feedback!

lurdan:
Though upstream haven't been released regularly, development itself has been going on.
Please refer upstream...
cf. http://cvs.namazu.org/Development/emacs-w3m/ChangeLog?sortby=date&view=log

purcell:
Hmm, so should we be reverting #3381, so that `w3m` comes from CVS again?

lurdan:
I think so.
("cvs -d :pserver:anonymous@cvs.namazu.org:/storage/cvsroot co emacs-w3m" seems to work.)

tarsius:
Or we could try to convince the maintainers that it is time to switch to git (or hg).

Actually we should do that for all packages that don't use git or hg. Is it really worth supporting five additional version control systems for the following 16 packages?

```
| Package           | Type   |
|-------------------+--------|
| applescript-mode  | svn    |
| caml              | svn    |
| cg                | svn    |
| clang-format      | svn    |
| confluence        | svn    |
| darcsum           | darcs  |
| dic-lookup-w3m    | svn    |
| dsvn              | svn    |
| helm-ls-svn       | svn    |
| ruby-additional   | svn    |
| ruby-electric     | svn    |
| sml-modeline      | bzr    |
| swbuff            | cvs    |
| tex-smart-umlauts | darcs  |
| vc-fossil         | fossil |
| weblogger         | bzr    |
```

lurdan:
The problem I wish to solve is that melpa represents w3m as orphaned to its users.

I think that a) pointing upstream cvs is the first option, b) maintain mirror may be good, c) drop w3m recipe with unmaintained mirror may also be the choice.

purcell:
> Is it really worth supporting five additional version control systems for the following 16 packages?

`package-build` is a general-purpose thing. I don't mind having code in it for obscure VCSes, as long as - in practice - I don't have to spent much time thinking about them. And I currently don't have to. :-)

purcell:
Still not sure I know enough to decide the best course of action for most users here. I'm inclined to revert the w3m recipe but leave shimbun, though that feels a bit weird. I have no idea who is still responsible for development in CVS, but ideally someone would convince them to relocate.

lurdan:
This private mirror seems to be keeping up with upstream (for now).
https://github.com/ecbrown/emacs-w3m

@tsuchm @yamaoka
How do you think about these request (to migrate emacs-w3m development repository from cvs to github) ?

tsuchm:
Hello,

I am an original developer of emacs-w3m, and still continuously maintain CVS repository as our official development area.

Could you explain the reason why CVS repository is regarded as unreliable?

lurdan:
For a while, I've submitted PR which makes w3m (and shimbun) will be fetched from upstream correctly.

(In this case anything will be OK for me, as long as it works, though I personally prefer git over cvs.)

purcell:
Merged, thanks.

tarsius:
> Could you explain the reason why CVS repository is regarded as unreliable?

@tsuchm It's not unreliable as far as I remember. But most packages have now moved on to use git and I would suggest you do the same. Importing the cvs history into git isn't that hard and after you have done so some users might start to contribute in minor or not so minor ways.

tarsius:
@purcell Earlier you said:

> `package-build` is a general-purpose thing. I don't mind having code in it for obscure VCSes, as long as - in practice - I don't have to spent much time thinking about them. And I currently don't have to. :-)

We don't *have to*, but it would certainly be nice if `package-build.el` were improved. But the prospect of having to deal with 5 additional version control systems for the sake of 17 packages, makes at least myself much less likely to contribute. If it weren't for this and some accidental complexity in `package-build.el` regarding "version handling" and "checkout", then I would probably have already taken care of  #2955. I think the current version detection code has to be improved before implementing a new version scheme. That means dealing with those 5 vcs, and I just don't look forward to that. Long story short, supporting these little used vcs does have a real cost.

So I guess what I am looking for here, is a public statement that you support my effort to get the maintainers of the 17 packages that use something other than `git` or `hg` to switch. I am willing to help those maintainers quite a bit in the process. I think that is a better investment of my time than banging on fetcher specific code in `package-build.el`.

In a first step (which will be spread across several months), I would like to get maintainers of `cvs` and `svn`. Those vcs are obsolete(-ish) and they also require more code in `package-build.el` than the less-obsolete but still not often used vcs.

Actually, let's get concrete about removing `cvs` support. All that is required for that to be possible is to get @lurdan to move the `w3m` repository to `git`. That takes care of two out of four packages that still use `cvs`, `w3m` itself and `shimbun`. The other two `cvs` packages, `debian-bug` and `debian-changelog-mode` come from https://tracker.debian.org/pkg/emacs-goodies-el. They consist of individual libraries. The Emacsmirror does mirror those packages, but it does so by importing the [raw](https://alioth.debian.org/scm/viewvc.php/*checkout*/emacs-goodies-el/elisp/debian-el/debian-bug.el?root=pkg-goodies-el) [files](https://alioth.debian.org/scm/viewvc.php/*checkout*/emacs-goodies-el/elisp/dpkg-dev-el/debian-changelog-mode.el?root=pkg-goodies-el), just as if upstream didn't use a vcs at all. Melpa could get these packages from the Emacsmirror, and would not lose anything in the process. Of course after we have removed `cvs` support, we cannot add any new packages that still use `cvs`.

purcell:
I'd be happy to remove `cvs` support if the maintainers will buy into it, or if the emacsmirror will do the trick.

As for `svn`, I'm thinking we could easily just use `git-svn` for those packages in the immediate term.

As for the other VCSes that now appear to be statistical edge cases (fossil, darcs, cvs), I'd probably be fine to drop support for them.

tarsius:
@tsuchm I originally mistyped your name in my above message, https://github.com/melpa/melpa/pull/3378#issuecomment-292999432, and I am not sure if an edit triggers a notification.

So, would you be willing to consider making this move?

PR #3379, opened 2015-12-12
cdb: fetch from GitHub

ttuegel:
The anonymous CVS repository is very unreliable. The `ddskk` package from the same repository was already using GitHub.


purcell:
Thanks for this.

PR #3380, opened 2015-12-12
Remove several packages that conflict with built-in packages

tarsius:
See the commits for the reasoning in each case. Drop the package *or* the commit as you see fit :-)

/cc @sigma @byplayer @mooz @jrockway

### DIFF
--- a/recipes/cperl-mode
+++ b/recipes/cperl-mode
@@ -1,2 +1,0 @@
-(cperl-mode :repo "jrockway/cperl-mode" :fetcher github)
-
--- a/recipes/table
+++ b/recipes/table
@@ -1,1 +1,0 @@
-(table :fetcher github :repo "byplayer/table.el")
--- a/recipes/tabulated-list
+++ b/recipes/tabulated-list
@@ -1,2 +1,0 @@
-(tabulated-list :repo "sigma/tabulated-list.el" :fetcher github)
-


milkypostman:
will let this one breath a bit before commiting but seems fine.

jrockway:
I'm not really maintaining cperl-mode anymore, so I think it's fine to drop the package.

byplayer:
Original author never answer me.
Please drop table.


Thank you.

2015-12-13 6:12 GMT+08:00 Jonas Bernoulli <notifications@github.com>:

tarsius:
We've gotten an okay for `cperl-mode` and `table`. For `tabulated-list` and `shadow` we've gotten no reply, but it seems pretty clear to me that those have to be removed. Can we remove now?

byplayer:
OK, thank you for your following up.

Regards,
Yukio Goto

lurdan:
I've contacted the author of shadow.el (@mooz), and he acked to change its name soon.
So please wait a bit.

tarsius:
@lurdan Great. Thanks!

sigma:
that's fine with me (for tabulated-list). I guess nobody cares about emacs 23 anymore, and it's not like it'll be unavailable anymore (people can still find it on marmalade)

milkypostman:
we waiting on @mooz then?

tarsius:
> we waiting on @mooz then?

I've just rebased and removed the commit which removes `shadow`. So you can just merge this now.

The commit which removes `shadow` has to be recreated anyway, since it should also add the renamed package at the same time.

tarsius:
It has been a while...
Do you want me to send a new pr which removes `shadow`.
We can still add the renamed package back later, iff @mooz gets around to the renaming.

PR #3381, opened 2015-12-12
Fetch shimbun and w3m from emacsorphanage

ttuegel:
See #3378.


purcell:
Super, thanks. (And thanks @tarsius too!)

PR #3382, opened 2015-12-13
remove deprecated smart-operator

tarsius:
`electric-spacing` is the successor.
See https://github.com/xwl/smart-operator/issues/3.


purcell:
Thanks, makes sense.

PR #3386, opened 2015-12-14
add recipe for slack

yuya373:
Hi!

This is my own slack client package.
It allows you to using Slack from Emacs.

Github: https://github.com/yuya373/emacs-slack

Thanks!

### DIFF
--- a/recipes/slack
+++ b/recipes/slack
@@ -1,0 +1,1 @@
+(slack :fetcher github :repo "yuya373/emacs-slack")


purcell:
Thanks, looks good to me. You might consider removing the `slack-pkg.el` -- MELPA doesn't need it if there's a `Package-Requires` header in `slack.el`, so removing it would prevent some duplication.

PR #3388, opened 2015-12-14
New recipe for xquery-tool.

paddymcall:
xquery-tool (https://github.com/paddymcall/xquery-tool.el) is supposed to be a very simplistic interface to saxonb's xquery things. Mainly useful for running simple xpath queries.

I wrote this package.


purcell:
Super, thanks.

PR #3389, opened 2015-12-14
Add highlight-leading-spaces

mrBliss:


purcell:
Thanks. Mostly looks good, but you're missing a `provide` form. And don't use `emacs` as the parent group for your customisations: either leave it blank, or use a specific parent group.

mrBliss:
Thanks for the hints, I applied them. I was hoping that by using [flycheck-package] I would get everything right, but I just read that detecting a missing `provide` is still on the roadmap :smiley: 

[flycheck-package]: https://github.com/purcell/flycheck-package

purcell:
Yep, early days. :-)  Merging now...

PR #3390, opened 2015-12-14
Add Malabarba's missing packages

tarsius:
See https://github.com/milkypostman/melpa/issues/3383#issuecomment-164409053.

### DIFF
--- a/recipes/brightness
+++ b/recipes/brightness
@@ -1,0 +1,1 @@
+(brightness :fetcher github :repo "Malabarba/brightness.el")
--- a/recipes/bug-hunter
+++ b/recipes/bug-hunter
@@ -1,0 +1,1 @@
+(bug-hunter :fetcher github :repo "Malabarba/elisp-bug-hunter")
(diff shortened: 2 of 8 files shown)


vyp:
@tarsius the `lisp-assistant` name should be `lisa`.

tarsius:
@vyp Thanks for the notice. Already fixed.

Malabarba:
Thanks for the effort, but I don't want any of these to be merged. Some of these are not proper packages, they're just scripts. Others are packages that are not ready yet (and aren't properly formatted). 

The only repo there that is a proper package is the bug hunter, and there's no reason to add it to Melpa. It's already on Gelpa and it has no need for snapshot releases.

Malabarba:
Minor correction: spinner is also a proper package, but everything I said about the bug hunter applies to it too.

tarsius:
Alright. I'll close this then and add only `bug-hunter` and `spinner` to the Emacsmirror (allowing me and others to install them as submodules). Thanks for the information.

PR #3391, opened 2015-12-14
New recipe for org-ref. 

jkitchin:
This is a library for org-mode that enables citations, cross-references and bibliographies with export support for different backends. As far as I know this package builds correctly. I am the author and maintainer of the package.

### DIFF
--- a/recipes/org-ref
+++ b/recipes/org-ref
@@ -1,0 +1,5 @@
+(org-ref
+ :fetcher github
+ :repo "jkitchin/org-ref"
+ :branch "melpa"
+ :files ("*"))
--- a/recipes/org-ref
+++ b/recipes/org-ref
@@ -1,5 +1,4 @@
 (org-ref
  :fetcher github
  :repo "jkitchin/org-ref"
- :branch "melpa"
- :files ("*"))
+ :branch "melpa")


purcell:
Thanks. Quick feedback:

- Please remove the `:files` spec here: the default should be fine, and will include all the `.el` files except the tests. And `test-melpa.el` doesn't look like it should get distributed. My suggestion is that you put the distributable elisp and the "packaging" stuff in separate directories.
- Use `0` for the depended-upon version of `key-chord` rather than the snapshot version. It's safer for MELPA Stable users if `key-chord` ever gets tagged in future.
- `(org)` is an illegal dependency spec in all but bleeding-edge dev Emacs: the version number should therefore be considered mandatory. You probably want "8", or just depend on a recent Emacs version which has a decent `org` bundled.
- `defvar-local` is new in Emacs 24.3, so you'll need to depend on that version, or use the classic `defvar`/`make-variable-buffer-local` combination
- Requiring `helm` is a very heavy-weight dependency. You should really consider making it optional, since many people do not use or like `helm`, and if you can use something like `completing-read`, then that can get overridden by the user's completion framework of choice.
- It's not great to bundle a lot of unprefixed libraries, ie. `isbn.el`: these could end up clashing with separate packages in future. Better to either call them `org-ref-isbn.el`, or make them their own packages if they are truly reusable.

Hope that helps!

jkitchin:
- Please remove the :files spec here: the default should be fine, and will include all the .el files except the tests. And test-melpa.el doesn't look like it should get distributed. My suggestion is that you put the distributable elisp and the "packaging" stuff in separate directories.

I removed that.

- Use 0 for the depended-upon version of key-chord rather than the snapshot version. It's safer for MELPA Stable users if key-chord ever gets tagged in future.

done.

- (org) is an illegal dependency spec in all but bleeding-edge dev Emacs: the version number should therefore be considered mandatory. You probably want "8", or just depend on a recent Emacs version which has a decent org bundled.

I removed org and added an emacs dependency

- defvar-local is new in Emacs 24.3, so you'll need to depend on that version, or use the classic defvar/make-variable-buffer-local combination

I added an emacs dependency that covers this.

- Requiring helm is a very heavy-weight dependency. You should really consider making it optional, since many people do not use or like helm, and if you can use something like completing-read, then that can get overridden by the user's completion framework of choice.

Users who really dislike it can configure org-ref not to use it. I want the default to be helm.

- It's not great to bundle a lot of unprefixed libraries, ie. isbn.el: these could end up clashing with separate packages in future. Better to either call them org-ref-isbn.el, or make them their own packages if they are truly reusable.

I have taken these out. Thanks for the feedback!

purcell:
Your comment seems to have disappeared, and did you mean to close this PR?

Re. keeping `helm` as the default completion, if you were to make `org-ref` work without it installed, you'd then remove the formal package dependency on `helm` etc. so that `helm` wasn't installed for everyone.

jkitchin:
I did close the PR. At this time I don't have the time, or interest in
supporting a non-helm-dependent installation of org-ref. While it is
possible to run without helm, and then configure it to use helm, this
isn't what I want to make happen for new people who want to use org-ref.
I want them to be able to install one package, configured in the most
useful way I know of out of the box, which is with helm and helm-bibtex,
and then use it the way they can see it used in video tutorials I have
made. That is just my opinion, and there is more than one way to do.
People who don't want to use helm can customize org-ref to use reftex,
or ebib, or even roll their own selection function if they want, and
they can select alternatives to helm for the link menus. I don't see
installing helm as a great burden.

I thought it was ok to take out the other files, e.g. isbn.el, but then
I was not convinced. These all support scientific writing in ways that
are important and useful to me, and part of what I think of as org-ref.
I am considering renaming them to avoid other name-conflicts.

I am targeting new students, scientists and engineers who just need
something that will work with minimal effort for writing scientific
papers. In my opinion the current package provides that, with sensible
defaults (which use helm).

So, I did close the PR because I am not sure when I will make up my mind
about the additional packages, and I am not inclined to remove helm as a
dependency. I didn't want to leave an open-ended PR up.



Steve Purcell writes:

> Your comment seems to have disappeared, and did you mean to close this PR?
>
> Re. keeping `helm` as the default completion, if you were to make `org-ref` work without it installed, you'd then remove the formal package dependency on `helm` etc. so that `helm` wasn't installed for everyone.
>
> ---
> https://github.com/milkypostman/melpa/pull/3391#issuecomment-164997458

--
Professor John Kitchin
Doherty Hall A207F
Department of Chemical Engineering
Carnegie Mellon University
Pittsburgh, PA 15213
412-268-7803
@johnkitchin
http://kitchingroup.cheme.cmu.edu

purcell:
At the end of the day, the helm dependency and the bundling of the other files wouldn't be a barrier to merging this, but I try to take the time to give feedback on areas which aren't ideal. I prefer to check with authors that they are aware of trade-offs like that so that they can be adjusted (if it makes sense) before publishing, particularly if the recipes would look different in those cases.

So just let me know, because from my side it would be okay to merge this PR in its current state.

jkitchin:
Thanks, I appreciated the feedback. Your points were all good ones to consider, which I did, and it resulted in improvements to the package. Mission accomplished! I have addressed all of your points but removing helm as a dependency. I renamed many of the other libraries internally, and I think they belong where they are for now. Thanks for your work in maintaining a high standard in Melpa!

purcell:
Great, thanks -- I've committed a small variation on this recipe, targeting your default branch instead of `melpa`. We prefer to track the development version of code, because other branches will often get forgotten and end up stale. Note that any numeric version tags placed on a tree will be used to build a correspondingly-numbered version of the package in MELPA Stable.

Thanks again for your patience!

PR #3392, opened 2015-12-14
Add recipe for Clojars.

joshuamiller:
This adds a client for searching the Clojars repository for Clojure packages.


purcell:
Thanks. Looks good, and I'll merge it now, but you shouldn't hard-code the use of `ido-completing-read` -- just use `completing-read`, and the user's completion framework of choice (`ido` / `helm` / `icicles` etc) can override it as necessary.

PR #3393, opened 2015-12-15
Change diminish repository to the new official one

myrjola:
I have taken over maintainership of diminish after a private mail conversation with the original author Will Mengarini. The official repository is now https://github.com/myrjola/diminish.el. I have also notified Emacsmirror about this change https://github.com/emacsmirror/p/issues/67#issuecomment-164792676.


purcell:
Super, thanks.

PR #3394, opened 2015-12-15
remove web-server

tarsius:
See https://github.com/milkypostman/melpa/issues/3383#issuecomment-164880700.


purcell:
According to that comment, it's an actively-developed package, so I'm not sure we need to rush into removing it from MELPA. I would hope that the ELPA and MELPA Stable (if any) versions of this package would match up.

tarsius:
Either way is fine with me in this case.

PR #3395, opened 2015-12-15
remove minimap

tarsius:
This conflicts with the independent implementation in GNU Elpa.

### DIFF
--- a/recipes/minimap
+++ b/recipes/minimap
@@ -1,1 +1,0 @@
-(minimap :fetcher github :repo "dustinlacewell/emacs-minimap")


purcell:
This is probably the right decision, but it's a bit disappointing if the conclusion is that whenever someone names a library, simply putting it in GNU ELPA would allow it to push out a pre-existing library with the same name on MELPA. It would be nice if the ELPA maintainers looked out for this sort of thing, as we do.

/cc @dustinlacewell

tarsius:
The implementation in GNU Elpa is older than this one (2009 vs. 2011), so it's not a pre-existing library that is being pushed out of Melpa.

Silex:
Hum, I'm a user of this package... how would that removal affect me? Will the current minimap package be renamed?

tarsius:
That's up to @dustinlacewell. But if he doesn't take action, then his implementation will have to be removed from Melpa.

@Silex maybe you could give the GNU implementation a try and report on the differences. If you as a user of Dustin's implementation would end up preferring the GNU implementation, then that would make it easier to justify the removal of Dustin's.

Silex:
Hum, I just gave it a try and it's more or less equivalent... Dustin's work seems more recent, as given by the commentary:

```
;; Heavily based off minimap.el by David Engster however I
;; have refactored it a bit to support only a single global
;; minimap window. Usually this should be opened on startup
;; when there is a single window. Some defaults have also 
;; been changed.
```

@dustinlacewell version is basically a fork, with some bugs fixed and some behavior changes (you can see this by ediffing the two package) or look at the commit list https://github.com/dustinlacewell/emacs-minimap/commits/master.

In my experience, the dustinlacewell version is a bit prettier and bugs a little less... he should propose his changes upstream.

tarsius:
Could we please remove this now?

I agree that it would be nice if Dustin's changes made their way upstream, but for that he has to get active because he would have to agree to assign the copyright for his changes to the FSF.

milkypostman:
i guess i'm not sure I see the issue here. there is a gnu version (which people can install) or a melpa version (which seems to be an updated version of the gnu one). while there *is* a naming conflict, I think this is inline with our plan right? we had `python.el` package for many years before that got merged as the "official" python version.

tarsius:
Isn't there a policy against adding random forks? Granted we don't now *how* random this fork is. Maybe @dustinlacewell made an effort to get his changes merged but did not succeed. Or maybe he didn't want to go through the trouble or the initial changes were rejected for a good reason. If it is the latter then this would be a random fork. IMO Melpa shouldn't encourage that and therefore shouldn't distribute such forks.  But until he replies here, we cannot know.

But it's your call. Please feel free to close this.

milkypostman:
@tarsius yeah the problem is more of removing a recipe. it made it through the cracks and now I'm not really sure i believe we should remove it. but lets just do it an see if anyone complains.

tarsius:
We'll see :-/

PR #3396, opened 2015-12-15
recipe for private-diary

cacology:
I wrote this package as a simple way of maintaining two diary files and want to share it with colleagues. According to your instructions, I have checked it:

- it passes flycheck-package
- it creates a defcustom variable for the private diary which can be swapped with the non-private one with   a function in the package
- https://github.com/cacology/private-diary
- I wrote it and maintain it
- I altered it so it would work with Melpa

It builds and installs properly locally.

I welcome comments, suggestions, encouragement, or discouragement.

Respectfully submitted,
-James

### DIFF
--- a/recipes/private-diary
+++ b/recipes/private-diary
@@ -1,0 +1,1 @@
+(private-diary :fetcher github :repo "cacology/private-diary")


purcell:
Great, thanks.

PR #3397, opened 2015-12-15
Add C-c-combo Recipe

CestDiego:
This package will make announcer sounds if you are over certain WPM

### DIFF
--- a/recipes/c-c-combo
+++ b/recipes/c-c-combo
@@ -1,0 +1,3 @@
+(c-c-combo :repo "CestDiego/c-c-combo.el"
+           :fetcher github
+           :files (:defaults "*.wav"))


purcell:
Fun, but no capitals in package or symbol names please.

CestDiego:
I made the changes.

CestDiego:
@purcell I forgot about the capitals in packages rule.

purcell:
Thanks. You'll also need to prefix all top-level symbols with the package prefix, e.g. `get-announcer-file-paths` -> `c-c-combo--get-announcer-file-paths`

purcell:
(I think that's the only one, actually.)

CestDiego:
Woops I missed that one.

CestDiego:
Done @purcell

purcell:
Thanks.

PR #3398, opened 2015-12-16
Add package vdirel: Manipulate vdir repositories

DamienCassou:
- description: vdir is a standard storing calendars and contacts on a file system much like maildir is for emails. vdirel provides utility functions to manipulate vdir repositories from Emacs.

- link: https://github.com/DamienCassou/vdirel

- association: author

### DIFF
--- a/recipes/vdirel
+++ b/recipes/vdirel
@@ -1,0 +1,1 @@
+(vdirel :fetcher github :repo "DamienCassou/vdirel")


purcell:
`helm` is a very heavyweight dependency for this, which would stop many potential users from wanting to use it. Why not make that optional, and perhaps even just use `completing-read` -- the user's completion system of choice can then override that as desired.

DamienCassou:
Thanks for the piece of advice. I might decompose the package later. For now, I would like to focus on creation and modification of contacts, something I need now. Are you still willing to merge the PR as it is?

Regarding optional dependency, how would you specify that? By creating 2 separate packages, one `-core` and one `-helm`? Or by using `(with-eval-after-load "helm" ...)`?

DamienCassou:
@purcell Moreover, I don't think I can provide the basic insertion feature with `completing-read`: https://github.com/DamienCassou/vdirel#inserting-email-addresses

purcell:
> Regarding optional dependency, how would you specify that? By creating 2 separate packages, one -core and one -helm?

Yes, that would be ideal: `vdirel` and `helm-vdirel`.

> Or by using (with-eval-after-load "helm" ...)?

According to the docs, that's meant only for use in user config, and shouldn't normally be used in packages. (And, pedantically, `with-eval-after-load` is a recent Emacs addition, so the classic `eval-after-load` is a better choice for wider compatibility.)

I'll go ahead and merge this for now, but I'd definitely encourage splitting out a separate `helm-vdirel` package when you can.

DamienCassou:
Thanks

PR #3399, opened 2015-12-17
add ob-restclient.el

stardiviner:


stardiviner:
Here is the old pull request, but it does not update, I want to see ob-restclient in MELPA. So I added a new one.
https://github.com/milkypostman/melpa/pull/3366

purcell:
Thanks.

PR #3400, opened 2015-12-17
Repository moved to emacs-eclim organization.

nloyola:
The emacs-eclim organization is a GitHub organization and was created on
December 16, 2005. The previous project owner was [senny/emacs-eclim](https://github.com/senny/emacs-eclim).


purcell:
Thanks.

PR #3401, opened 2015-12-17
Add recipe for ac-html-angular.

osv:
Angular 1.5 completion data for EMACS `ac-html` and `company-web` modes


purcell:
Thanks. You should use `:defaults` rather than `"*.el"` in the `:files` pattern -- please can you amend the recipe and force-push it?

P.S. The code itself can be simplified slightly by using `push` rather than `setq` + `cons`.

osv:
amended and simplified :)

purcell:
Super, thanks.

PR #3403, opened 2015-12-18
Add johanvts/emacs-fireplace

hut8:
https://github.com/johanvts/emacs-fireplace

**What it does:**

displays a fireplace

**My relation to project**

I asked if I could make a package for it here: https://github.com/johanvts/emacs-fireplace/issues/1

Thanks.

### DIFF
--- a/recipes/fireplace
+++ b/recipes/fireplace
@@ -1,0 +1,1 @@
+(fireplace :fetcher github :repo "johanvts/emacs-fireplace")


AdrieanKhisbe:
:+1:

milkypostman:
haha, this is hilarious. but it's missing an autoload statement which means it will be worthless to add to MELPA at this point.

can you send a PR/file an issue with the original package to add the autoloads, then we can merge this.

I think all you need is to add,

```
;;;### autoload
```

before the `fireplace` function.

hut8:
Aww man, rookie mistake. I'm on it.

AdrieanKhisbe:
@milkypostman PR is pending https://github.com/johanvts/emacs-fireplace/pull/8 :)

hut8:
Heh. Beat me to it @AdrieanKhisbe

purcell:
This is lots of fun, so please bear with me while I get a little picky:

- The metadata looks like it's been hacked together just enough to work, but it's messy, with the `Commentary:` above the other headers etc. Please try using `M-x auto-insert` in a fresh `fireplace.el` and then use the more complete boilerplate it will generate for you.
- `fp-` is not a good prefix for private symbol names. Use `fireplace--`, as per the [elisp conventions](https://www.gnu.org/s/emacs/manual/html_node/elisp/Coding-Conventions.html).
- `C-s` is a keybinding [reserved for users](https://www.gnu.org/software/emacs/manual/html_node/elisp/Key-Binding-Conventions.html)

purcell:
Also: `smoke`, `flame`, `make-grid`, `gotoxy` etc. all need to be prefixed with `fireplace-` or `fireplace--`.

AdrieanKhisbe:
@purcell: your concern have now been addressed and PR has been merge.

I thinks it's time to put melpa on fire!! :D :fire:

purcell:
Thanks!

AdrieanKhisbe:
:fireworks:

PR #3404, opened 2015-12-18
Add vhdl-tools

csantosb:
Package to navigate a project containing vhdl sources, fork of `vhdl-goto-ref`, from where I add the functions I currently use.

### DIFF
--- a/recipes/vhdl-tools
+++ b/recipes/vhdl-tools
@@ -1,0 +1,1 @@
+(vhdl-tools :repo "csantosb/vhdl-tools" :fetcher github)


milkypostman:
when would you enable this but not as a major mode? in other words, what major mode would you enable this in?

this may take us a bit of time to review just due to the size but it seems pretty ok to me.

csantosb:
I guess you mean this ?

    (add-hook 'vhdl-mode-hook
        (lambda ()
            (vhdl-tools-mode 1)))

VHDL mode then.

purcell:
Quick feedback:

- Please use `:defaults` rather than `"*.el"` in the recipe: you can amend and force-push the commit to fix the pull request cleanly
- Needs to declare a package dependency on Emacs 24.3, ie. `(emacs "24.3")` in the `Package-Requires:` header
- Needs to declare a package dependency on the latest stable version of `outshine`


Looks good to me otherwise!

csantosb:
Thanks for the feedback. I think all should be ok now.

purcell:
Ah, no, sorry: I think initially you had `:files ("*.el" "*.sh")` in which case you'd have instead needed `:files (:defaults "*.sh")`, but if `:defaults` is adequate, then please just remove the `:files (...)` part altogether.

csantosb:
My fault: only `*.el` files are necessary. Better now ?

milkypostman:
@csantosb this is close but according to the travis failure, my guess is that your branch is out of date (merged instead of rebased) on our master branch. can you try fetching the latest from *us* and rebasing your change on that, then repushing?

csantosb:
Done, now my master branch is up to date with yours. But I am stilling getting a travis failure, any idea of what am I doing wrong ?

milkypostman:
looks like a cask problem actually maybe I misread it before. will see how a retry works.

PR #3405, opened 2015-12-19
popup-imenu package recipe

ancane:
Hello. I built this package to show imenu index in a popup window in response to this request https://github.com/ancane/scala-outline-popup/issues/7.
Lives in https://github.com/ancane/popup-imenu
Thanks.


fommil:
:+1:

PR #3407, opened 2015-12-19
Add recipe for rand-theme

gopar:
Summary: `rand-theme` randomly selects a theme from a predefined list. The list can be a white or black list. If white list is created then only select themes from that list, else select any theme that is not in the black list.

Link: [https://github.com/gopar/rand-theme](https://github.com/gopar/rand-theme)

Association: I am the creator and maintainer of this project.

Please let me know if there is anything I can fix. Thanks


purcell:
Thanks.

PR #3408, opened 2015-12-20
Add package smartwin

jerryxgh:
- description: smartwin uses a specifical window to show shell like
  buffers, the window will always at the bottom, and can shrink or
  enlarge automaticlly when move out or in it.

- link: https://github.com/jerryxgh/smartwin

### DIFF
--- a/recipes/smartwin
+++ b/recipes/smartwin
@@ -1,0 +1,1 @@
+(smartwin :fetcher github :repo "jerryxgh/smartwin")


purcell:
Whoah, this is quite an extreme approach, with all those `defadvice`. Seems like it would be a maintenance nightmare, and I feel like there must be something similar already: I know that ECB provides this functionality, for example.

Couple of little specific points, too: `(kbd "C-c s")` is reserved for end user customisation (see [the docs here](https://www.gnu.org/software/emacs/manual/html_node/elisp/Key-Binding-Conventions.html)), and since you're using `ad-activate` and `subr-x` you'll probably need to depend on at least Emacs 24.4.

jerryxgh:
Amazing that you know everything. I used ECB before, for me, it has problems to work with other plugins, too heavy and has a lot of things unneeded, but `compile-window` feature is very useful. My first idea is separating this from ECB, but it's too complicated, so I got smartwin. I use it everyday and fix problems continually. Days ago I thought maybe it can help others, so I came here.

Follow your opition, move `(kbd "C-c s")` out, leave key binding to user and update dependency to Emacs 24.4.

This is my first emacs plugin, maybe naive, but I wish it can help others.

Thanks

purcell:
Great, thanks.

PR #3413, opened 2015-12-22
Add projectile-sift

nlamirault:
Add projectile-sift and update recipe for sift.
Project homepage : https://github.com/nlamirault/sift.el

### DIFF
--- a/recipes/projectile-sift
+++ b/recipes/projectile-sift
@@ -1,0 +1,2 @@
+(projectile-sift :repo "nlamirault/sift.el" :fetcher github :files ("projectile-sift.el"))
+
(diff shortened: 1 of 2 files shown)


purcell:
The first element of the recipe for `projectile-shift` says `sift`, rather than `projectile-shift`, which is why the build failed: you should be able to just amend that commit and force push it back here to fix the PR.

bbatsov:
You should probably squash the first and the third commits together.

milkypostman:
^^^ yes please

nlamirault:
Done.

milkypostman:
This looks like a bad rebase. I'll just merge the recipe.

PR #3414, opened 2015-12-22
Adding html-frag-checker

Wilfred:
This nifty little module was written by Tobias to solve http://stackoverflow.com/questions/19363837/how-can-i-highlight-unmatched-html-tags-in-emacs

html_check_frag

### DIFF
--- a/recipes/html-check-frag
+++ b/recipes/html-check-frag
@@ -1,0 +1,1 @@
+(html-check-frag :repo "TobiasZawada/html-check-frag" :fetcher github)


purcell:
Thanks, but we're not adding new emacswiki packages these days: even the wiki maintainers have deprecated using it for code hosting. Is there an alternate, more authoritative source for this?

Wilfred:
I'm afraid not. I think the original user just used the wiki as a pastebin. What do you suggest?

purcell:
Option 1: send the maintainer a quick email to ask if they'd be willing to put the code on github or similar. They'll often comply, if the intent is to make the code available for packaging.
Backup plan, if no response: add the code to the [Emacs orphanage](https://github.com/emacsorphanage), which @tarsius would be able to enable.

tarsius:
@Wilfred, any success with Option 1?

Wilfred:
I've [added a comment on the SO question](http://stackoverflow.com/questions/19363837/how-can-i-highlight-unmatched-html-tags-in-emacs#comment57327795_19364696).

Wilfred:
Tobias has very kindly created a GitHub repo, so I've updated this PR to match.

milkypostman:
looks like a bad rebase on this. can you try again? i think that's the problem at least. the travis build seems to be doing a lot.

purcell:
Thanks. Pending https://github.com/TobiasZawada/html-check-frag/pull/2 and https://github.com/TobiasZawada/html-check-frag/pull/3.

TobiasZawada:
There is an issue with html-check-frag. Please see https://github.com/TobiasZawada/html-check-frag/issues/4. Anyone else who experienced this with emacs 25.0.50? I am still using Ubuntu 14.04 LTS at home. So testing at home is difficult.

PR #3415, opened 2015-12-22
add recipe for clean-buffers

lujun9972:


purcell:
Thanks. All the `cb-` symbols should be renamed to `clean-buffers-` -- `cb` is just too short, and the prefix should match the feature name whenever possible.

lujun9972:
DONE

purcell:
Thanks!

PR #3416, opened 2015-12-22
Add org-ref.org and citeproc dir to the recipe

jkitchin:
org-ref.org is a help file/manual for org-ref. The citeproc directory contains a citation processor for non-latex exports.


purcell:
Thanks. As noted in the README, `:defaults` is preferred to `".el"` unless there's a good reason, so perhaps you could amend the commit and force-push it here to fix the PR?

jkitchin:
Do you mean a recipe like this:
(org-ref :fetcher github :repo "jkitchin/org-ref" :files (:defaults "org-ref.org" "citeproc"))
to add the org-ref.org file, and the directory citeproc?

purcell:
Yep, exactly -- looks good now, thanks.

PR #3418, opened 2015-12-23
Add recipe for helm-img

l3msh0:
helm-img is an library for making helm-sources which return images as candidates.

https://github.com/l3msh0/helm-img

### DIFF
--- a/recipes/helm-img
+++ b/recipes/helm-img
@@ -1,0 +1,3 @@
+(helm-img
+ :repo "l3msh0/helm-img"
+ :fetcher github)


purcell:
Thanks. Sent you a fix in https://github.com/l3msh0/helm-img/pull/2.

purcell:
Also, `:files ("*.el")` is redundant and should be removed from the recipe -- all `.el` files are included by default.

purcell:
(Ideally, amend the recipe commit and force-push it here to cleanly update the PR.)

l3msh0:
Thank you for reviewing.
Merged your PR and amended the recipe commit.

PR #3419, opened 2015-12-23
Add recipe for helm-img-tiqav

l3msh0:
This pull request depends following one.
https://github.com/milkypostman/melpa/pull/3418

tiqav.com is joking image search engine for Japanese.
helm-img-tiqav execute search via tiqav.com and show results on helm buffer.

Project repository: https://github.com/l3msh0/helm-img-tiqav

### DIFF
--- a/recipes/helm-img-tiqav
+++ b/recipes/helm-img-tiqav
@@ -1,0 +1,3 @@
+(helm-img-tiqav
+ :repo "l3msh0/helm-img-tiqav"
+ :fetcher github)


purcell:
Needs the same code fix as in https://github.com/l3msh0/helm-img/pull/2, and the same recipe fix as in #3418.

purcell:
You should also escape the query param in `helm-img-tiqav-search` because otherwise queries containing `&`, `?` etc. will break the URL. There's a built-in elisp function for this, but I don't remember its name right now, sorry.

l3msh0:
Thanks!
I fixed the recipe and fixed escaping the query parameter by `url-hexify-string`.

purcell:
Great, thanks.

PR #3420, opened 2015-12-23
Add recipe for the super-save package

bbatsov:
A small package, containing Prelude's save-on-focus-lost logic.

### DIFF
--- a/recipes/super-save
+++ b/recipes/super-save
@@ -1,0 +1,1 @@
+(super-save :fetcher github :repo "bbatsov/super-save")


purcell:
Thanks. Couple of quick fixes are in order before merging, I think.

- `provide` form is missing
- The docstring `"Save the current buffer if `prelude-auto-save' is not nil."` needs updating. ;-)

Separately, I wonder if the various `defadvice` are needed when `focus-out-hook` is available?

bbatsov:
Ops, I'll address those right away. 

> Separately, I wonder if the various defadvice are needed when focus-out-hook is available?

Sure they are - it's triggered only if the frame itself loses focus (and is not available in older Emacsen).

bbatsov:
Done.

purcell:
> > Separately, I wonder if the various defadvice are needed when focus-out-hook is available?
> 
> Sure they are - it's triggered only if the frame itself loses focus (and is not available in older Emacsen).

Ah, I was thinking `focus-out-hook` was per window, not frame. The former would be more useful, and then you wouldn't need any `defadvice` at all, which was what I was referring to.

vyp:
@purcell Yes that would be pretty useful. Off topic, but I've been looking for some sort of window switch hook, do you know of any potential others that exist?

purcell:
@vyp Nope, sorry.

vyp:
Okay no worries. :+1:

PR #3421, opened 2015-12-23
Add go-playground-cli recipe

kosh04:
It is Go Playground web client tool, unlike go-playground recipe #3175.

go-playground-fmt.el is also implemented, but not need here.

### DIFF
--- a/recipes/go-playground-cli
+++ b/recipes/go-playground-cli
@@ -1,0 +1,3 @@
+(go-playground-cli :repo "kosh04/emacs-go-playground"
+                   :fetcher github
+                   :files ("go-playground-cli.el"))


milkypostman:
@kosh04 your commit to add the recipe contains a "merge" commit because you had an outdated fork from a previous PR>

please rebase your commit on top of `origin/master` of our repo and force push back here and ping when ready. thanks.

milkypostman:
does the "names" package work even when not compiling? it seems like you don't load that when loading from the file directly?

purcell:
@milkypostman The names macros are autoloaded IIRC, so they work without loading them explicitly.

@kosh04 I sent you a quick fix in https://github.com/kosh04/emacs-go-playground/pull/2.

kosh04:
@milkypostman update the commit.

milkypostman:
@kosh04 waiting on @purcell commit getting merged.

PR #3423, opened 2015-12-24
darcs: add support for stable versions

dunn:


package-build.el
```diff
@@ -325,19 +326,28 @@ A number as third arg means request confirmation if NEWNAME already exists."
...
-            (delete-directory dir t))
-          (package-build--princ-checkout repo dir)
-          (package-build--run-process nil "darcs" "get" repo dir)))
+  (let ((repo (plist-get config :url)))
+    (with-current-buffer (get-buffer-create "*package-build-checkout*")
+      (cond
+       ((and (file-exists-p (expand-file-name "_darcs" dir))
+             (string-equal (package-build--darcs-repo dir) repo))
+        (package-build--princ-exists dir)
+        (package-build--run-process dir "darcs" "pull" "--all"))
```
dunn: Maybe the behavior of darcs has changed, because on the version I tested with (2.10.2), `pull` is interactive by default.

package-build.el
```diff
@@ -190,8 +190,9 @@ optionally looking only as far back as BOUND."
 (defun package-build--find-version-newest (regex &optional bound)
   "Find the newest version matching REGEX before point, optionally stopping at BOUND."
   (let ((tags (split-string
-                (buffer-substring-no-properties
-                  (or bound (point-min)) (point)))))
+               (buffer-substring-no-properties
+                 (or bound (point-min)) (point))
+                "\n")))
```
dunn: I'm not sure if Git allows spaces in tags, but darcs does and the darcs repository itself has some wacky tags:
```
stable 2008-11-10
2.1.1rc2
2.1.1rc1
unstable before bytestring patches.
2.1.0
2.1.0pre3
2.1.0pre2
2.0.3pre1
this version works.
working version.
```
Changing the tag parser to split by newlines rather than by whitespace generally should prevent '2008-11-10' from being extracted as a valid tag.

package-build.el
```diff
@@ -516,7 +526,7 @@ Return a cons cell whose `car' is the root and whose `cdr' is the repository."
           ;; Using reset --hard here to comply with what's used for
           ;; unstable, but maybe this should be a checkout?
           (package-build--update-git-to-ref dir (concat "tags/" (cadr tag-version)))
-          ;; Return the version as a string
```
dunn: I copied some of the Git logic for darcs, and this comment confused me for a moment before I remembered that it wasn't referring to the original version-string.

package-build.el
milkypostman: this would explain why I have to reset the vc-darcs recipe so much.

milkypostman:
Thanks for all these PRs. Very very helpful.

dunn:
💝

PR #3424, opened 2015-12-24
Updated traad recipe.

abingham:
The traad server and emacs client code are now in separate repositories.
This change updates the recipe to refer to the new emacs client repo.


purcell:
Thanks. `traad-autocomplete` should be a separate package, with its own recipe. Please can you amend this so that there are 2 recipes, each one specifying its corresponding `.el` file using a `:files` spec?

abingham:
I've removed `traad-autocomplete` from the repository. It was a dead project that I intended to remove in any event.

purcell:
Super, thanks.

PR #3425, opened 2015-12-24
Add a recipe for crux

bbatsov:
crux is another package I've recently extracted from Prelude.

### DIFF
--- a/recipes/crux
+++ b/recipes/crux
@@ -1,0 +1,1 @@
+(crux :fetcher github :repo "bbatsov/crux")


purcell:
This looks great, but the `crux-default-keybindings` is a bit extreme -- most of those `C-c <letter>` combos are reserved for user customisation, and off-limits for packages. I'd say that part belongs in `prelude`.

purcell:
There are a couple of things here I can use myself to tidy up some local config bits-and-pieces. Guess we all end up sharing and stealing snippets. :-)

bbatsov:
> There are a couple of things here I can use myself to tidy up some local config bits-and-pieces.

Well, this is optional, so I don't think it's that extreme. I can easily introduce some global `crux-mode`, but that seems like an overkill to me. I'll extend the documentation at some point with more fine-grained setup instructions and I disclaimer about the nature of those keybindings.

> Guess we all end up sharing and stealing snippets. :-)

Indeed we do. :-)

purcell:
> Well, this is optional, so I don't think it's that extreme. I can easily introduce some global crux-mode, but that seems like an overkill to me. I'll extend the documentation at some point with more fine-grained setup instructions and I disclaimer about the nature of those keybindings.

Like I say, the keybinding config belongs in Prelude only. These are very handy commands and functions, but only Prelude users are going to want that exact set of keybindings, and neither a global mode nor disclaimers is the solution here. I feel like you've been influenced by commands like `windmove-default-keybindings`, but in that case the commands are all closely related, whereas `crux-setup-default-keybindings` is just a grab-bag of bindings you want in Prelude.

bbatsov:
OK, OK. I'm removing them right now. Although I still don't quite get what's wrong with having an **optional** way to setup everything in some manner, which may or may not be what you want. 

I'll just list all the commands in some table.

bbatsov:
Done.

purcell:
> Done.

Great - I sent you a fix for the table formatting. I don't see any code changes though.

bbatsov:
If people don't know about this function it's like it doesn't exist. Do you want me to delete it?

purcell:
Yes please.

bbatsov:
Done.

purcell:
Awesome - merging now. Looking forward to using some of these.

bbatsov:
> Looking forward to using some of these.

And contributing some more. ;-) I'd love to get as much useful commands here as possible. Long term, this package can be viewed as a staging ground for things to be added to Emacs itself eventually.

purcell:
Yeah, normally it's a bit of an anti-pattern to move lots of unrelated pieces of "personal" config into a package -- I've declined a few such packages previously, because they're usually not widely usable. In the case of `crux`, there are some genuinely useful functions and commands, but I'd still encourage looking for opportunities to group related functionality into distinct packages where possible.

PR #3428, opened 2015-12-24
vc-darcs: fix version detection

### LINKED ISSUE #2129: vc-darcs not migrating into melpa-stable
`vc-darcs` has been in MELPA for ages, and while it has version numbers and tags (it's currently at version 1.16), it is not available in MELPA-stable.  Note that `vc-darcs` is maintained in a `darcs` repository — perhaps that is the problem?

dunn:
Fixes #2129.

Should have caught this during #3423 :see_no_evil:

### DIFF
--- a/recipes/vc-darcs
+++ b/recipes/vc-darcs
@@ -1,1 +1,4 @@
-(vc-darcs :url "http://www.pps.univ-paris-diderot.fr/~jch/software/repos/vc-darcs/" :fetcher darcs)
+(vc-darcs
+ :url "http://www.pps.univ-paris-diderot.fr/~jch/software/repos/vc-darcs/"
+ :fetcher darcs
+ :version-regexp "vc-darcs-\\(.*\\)")


vyp:
btw, #3421 also has a similar ci build error. Not sure if related, but mentioning in case it is. (Perhaps somehow the result of your recent contributions?)

dunn:
Looks just like the VC software isn't being installed; darcs here and fossil in #3421. @milkypostman should Travis make sure to install them before running tests? I can't think what could have caused this to start showing up now, though.

milkypostman:
the reason it's failing is that the test tries to build only the latest recipes added/changed. Since we only have two darcs packages, we normally don't see these errors because we never try to build darcs packages.

milkypostman:
@dunn can you try rebasing this onto HEAD? then we can check if the changes I made to travis build fixes this?

dunn:
:ok_woman:

PR #3429, opened 2015-12-24
bzr: support stable versions

dunn:
I made changes to `package-build--find-version-newest` to support the bzr convention of writing tags like `REL1_4_5`.  The changes could only lead to trouble if someone had an odd tag like `0_1.2.2` that they expected to be parsed as `(0 -4 1 2 2)`; MELPA would now treat it as `(0 1 -4 2 -4 2)`.

An alternative approach would be to add another key to recipes, specifying the version-separator.

### DIFF
--- a/package-build.el
+++ b/package-build.el
@@ -193,11 +193,31 @@ optionally looking only as far back as BOUND."
                (buffer-substring-no-properties
                  (or bound (point-min)) (point))
                 "\n")))
-    (setq tags (cl-remove-if nil (mapcar
-                                  (lambda (tag)
-                                    (when (package-build--valid-version tag regex)
-                                      (list (package-build--valid-version tag regex) tag)))
-                                  tags)))
+    (setq tags (append
+                (mapcar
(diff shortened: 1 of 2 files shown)


milkypostman:
i'm going to commit this for now. it seems like the actual "right" way to do this is to let the regex actually pull out the version numbers. so you would write something like,

```
\([0-9]\)\(?:_\([0-9]\)\)*
```

which would produced some number of groups but which turn, `1_2_3` into the list `("1" "2" "3")` which we could cast to ints and it would be `(1 2 3)` and thus a version number.

dunn:
Yeah. For simplicity's sake I had `package-build--valid-version` only pay attention to the first capture group.  Looking again it might not be too hard to let it handle more; I'll try at least.

milkypostman:
Either way. It's not a big deal now but may be worth fixing for cases like
this and where people use weird versioning conventions. I'm happy to just
leave this as well.

PR #3432, opened 2015-12-26
Add recipe for dim

alezost:
Hello, would it be acceptable to add another package for setting
mode-line names of major and minor modes to MELPA?

### DIFF
--- a/recipes/dim
+++ b/recipes/dim
@@ -1,0 +1,1 @@
+(dim :repo "alezost/dim.el" :fetcher github)


milkypostman:
it would but I suspect that @purcell is going to add what it adds over things like `diminish` (what other ones exist?).

vyp:
@milkypostman [delight.el](http://www.emacswiki.org/emacs/delight.el)

purcell:
Fine with me, since there's a good rational for not using `diminish` or `delight` in the author's case.

alezost:
Thanks for merging!  I described my reasons for this package at https://github.com/alezost/dim.el#why

PR #3433, opened 2015-12-26
added kooten-theme

kootenpv:


milkypostman:
@purcell still adding themes here? Do you think we should spawn a separate repo for those? I guess it doesn't matter.

kootenpv:
@milkypostman  I was indeed confused; it seems weird to upload a theme into the package manager. Is there a better way to do it?

milkypostman:
oh no, it makes sense but just that lots of packages we have are themes. it
has been discussed before to maybe not make them part of the main package
repo but not many other choices.

purcell:
It's fine, but I sent a small fix in https://github.com/kootenpv/emacs-kooten-theme/pull/1.

kootenpv:
@milkypostman @purcell Can this still be merged after having merged purcell's changes?

purcell:
yup, done

PR #3005, opened 2015-08-09
stop pulling unauthenticated upstreams

glyph:
This will undoubtedly break many of these projects, but that's the point.  HTTPS is literally free, anyone can have it, it takes only a few minutes to set up, and while you wait you can use free HTTPS-enabled hosts like Github.

Issue #3004, opened 2015-08-09
start refusing to pull packages from unauthenticated upstreams


glyph:
A number of recipes retrieve code using either `http://` or `git://` transports; these should not be allowed.

My understanding is that typing `make` in the `melpa` repository is what does the actual building of packages, so it would be melpa doing the package signing when #1749 is implemented.  If my understanding is correct, I think it would be irresponsible to codesign a package that was just some code automatically scraped off of a wiki.

alphapapa:
Where do signed git tags/commits figure into this?  They're very easy to use.

Of course, if the GPG key is not fetched securely, and/or if it's not signed by a known trustworthy key, it's not technically authenticated (cf. Debian's key-signing parties).  But it would still be nice to have support for it.  And if MELPA stored the already-fetched GPG keys on its server, it could detect if a signing key changed, and it could refuse to build signed packages if they were signed by a different key, or unsigned packages if their previous releases were signed.

This would be a nice step in the right direction.  HTTPS/TLS is nice, but it's not a panacea, and it doesn't protect against something like GitHub being compromised.

glyph:
> Where do signed git tags/commits figure into this?

Git signatures rely on sha1, which has been more or less completely compromised by Google this year, so they don't provide a strong guarantee of authenticity either. Probably not worth bothering with.  (There's also the fact that commit-signing is not itself more meaningful than "this came from this repository" unless the signer verifies the contents of _every_ file, since the parent commit was not itself signed.)

> They're very easy to use.

Disagree. GPG's web of trust is so confusing that, as far as I know, there are no crypto experts who believe they can use it without serious preparation.  Unfortunately, the mode that makes the most sense for using GPG is when one is verifying a signature that can only have come from one key (like debian / apt signatures).

glyph:
> ... it doesn't protect against something like GitHub being compromised ...

Provider-independent security is always a good goal, but let's not try to build rocketships before we can even crawl.  This ticket is about _beginning_ to refuse to pull from package sources that are _entirely_ unauthenticated, and it's been open for almost 2 years.  Let's please have MELPA do that before getting off into the crypto-utopian weeds with peer-to-peer signatures.

alphapapa:
> Git signatures rely on sha1, which has been more or less completely compromised by Google this year, so they don't provide a strong guarantee of authenticity either.  Probably not worth bothering with.

I'm not sure I understand what you mean.  

[...]

PR #3436, opened 2015-12-26
Deprecate EmacsWiki

dunn:
Refs:
- #2342
- #3004
- #3005 

Is this premature?  If not, are there other places that need tweaking?


purcell:
Great - thanks for this.

AdrieanKhisbe:
:+1:

Issue #3306, opened 2015-11-19
[proposal] default to MELPA Stable


dunn:
I don't know if there are stats, but I'm guessing the vast majority of users are on MELPA Unstable rather than Stable.  I'd like to suggest that there should be some effort made to get more users on the Stable archive.

It seems like a logical step to take as a result of [the maturation of the ecosystem](http://www.emacswiki.org/emacs/2014-10-30) that Alex Schroeder cited when he told people to stop uploading code to EmacsWiki.  One of the things he thought users are now (or should be) demanding is _proper releases_.  Promoting proper—tagged, versioned—releases is good for users and developers; users get more stability (and maybe eventually the power to install specific versions) and developers are encouraged to think more about backwards compatibility and in general more \~user conscious developement\~, especially if we encourage semver.

But! I'm aware that Stable is underdeveloped and not actually used by the maintainers themselves (https://github.com/milkypostman/melpa/issues/3023#issuecomment-131473533, https://github.com/milkypostman/melpa/issues/2565#issuecomment-77275372), meaning that switching to Stable might provide a _worse_ user experience if it happened today.  But as things are now, there's very little motivation for contributors to improve Stable, or for package developers to tag stable releases (#1786).

So I think it'd be a really good idea to switch the default archive to Stable sometime in 2016.  Emacs 25 will be able to detect uninstallable packages (https://github.com/milkypostman/melpa/issues/1955#issuecomment-102679358), and once Stable supports the full range of fetchers (#2075) that would remove some of the major technical obstacles (unless there are others I'm not aware of).  I might take a stab at improving version detection (#2553) this weekend.  Of course the task would remain of convincing developers to get on board with regular versioned releases, but MELPA probably has enough clout to help drive these sorts of changes.

(Full disclosure: I'm the maintainer of the [Homebrew Emacs tap](https://github.com/Homebrew/homebrew-emacs) which does _not_ have that sort of clout; Homebrew is not very good at handling `HEAD` builds so this would help me there.)

vyp:
You mean like make "melpa" actually what "melpa stable" is now, and then make what "melpa" is now something like "melpa unstable"?

dunn:
Yeah, exactly.

purcell:
[...]

PR #3437, opened 2015-12-27
Better documentation for MELPA Stable

dunn:
I'm not wedded to any of the particular wording or organization choices here, but I think (as you know :see_no_evil: #3306) that Stable could be featured a bit more prominently.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ MELPA is a growing collection of `package.el`-compatible Emacs Lisp
 packages built automatically on our server from the upstream source
 code using simple recipes. (Think of it as a server-side version of
 [el-get](https://github.com/dimitri/el-get), or even
-[homebrew](https://github.com/Homebrew/homebrew).)
+[Homebrew](https://github.com/Homebrew/homebrew).)
 
 Packages are updated at intervals throughout the day.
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -51,17 +51,42 @@ Enable installation of packages from MELPA by adding an entry to
...
-If your Emacs has the variable `package-pinned-packages`, you can
-customize or modify that variable as needed. Otherwise, use the
-separate
-[package-filter.el](https://github.com/milkypostman/package-filter)
-package which we provide.
+* If your Emacs has the variable `package-pinned-packages` (available
+  in 24.4 and later), you can customize or modify that variable as
+  needed.
 
+* You can use the
```
vyp: Shouldn't this still be "Otherwise, you can use..." as before?

dunn: There's a third option (Stable) so that didn't seem appropriate, to me at least.

vyp: Ah I missed that this is a list! So yes it makes sense, sorry.

AdrieanKhisbe:
:+1:  :)

README.md
```diff
@@ -51,17 +51,42 @@ Enable installation of packages from MELPA by adding an entry to
...
+
+### MELPA Stable
+
+By default, MELPA provides the very latest revisions of packages.  If
+you prefer to only receive updates for tagged releases, use
+[https://stable.melpa.org](MELPA Stable) instead:
+
+```lisp
+(add-to-list 'package-archives
+             '("melpa" . "https://stable.melpa.org/packages/") t)
```
milkypostman: I think the name of this archive should be "melpa-stable"

README.md
```diff
@@ -148,6 +176,11 @@ through this process.
   recipe buffer: this will also prompt you to install the
   freshly-built package.
 
+  If the package has tagged releases, confirm that the correct version is
+  detected by running `STABLE=t make recipes/<NAME>`.  The version
```
milkypostman: Suggested revision,

> If the repo contains tags for releases, confirm the correct stable version is built by running ...

README.md
milkypostman: this would match with the html page as well.

PR #3438, opened 2015-12-27
Add dired-reuse recipe

crocket:
dired-single has been unmaintained for more than 2 years, and it hasn't moved to github and tagged a version.

dired-single is pretty much dead, and I think we need to start over with a new name.

I renamed it to dired-reuse. Mine has a version tagged, so it will be available on melpa-stable. Melpa packages are too unstable for my tasks.

### DIFF
--- a/recipes/dired-reuse
+++ b/recipes/dired-reuse
@@ -1,0 +1,1 @@
+(dired-reuse :fetcher github :repo "crocket/dired-reuse")


milkypostman:
should we remove `dired-single` then?

crocket:
I think it's ok to replace dired-single with dired-reuse since it seems abandoned.

purcell:
I can confirm that the author would be unreachable -- I unilaterally took over `whole-line-or-region` from him a few years ago. I wonder if there's much point renaming it, then. Maybe just call your fork `dired-single`, and we'll point the existing recipe at yours instead.

crocket:
Ok, I'll open a new pull request for `dired-single`.

PR #3439, opened 2015-12-27
Added CFF recipe

fourier:
A simple package to search C/C++ headers file by source file and vice versa in a typical different tree in git/svn repository. Allows to search using simple regexp/rules, like for a **file.cpp** file in **src** directory find the corresponding **file.h/hpp** file in *closest* **inc** directory.

### DIFF
--- a/recipes/cff
+++ b/recipes/cff
@@ -1,0 +1,1 @@
+(cff :fetcher github :repo "fourier/cff")


fourier:
Can't find why it is fail (no fails on my package). Can you please take a look?

fourier:
Any ideas why it seems to be broken? Cannot find anything related to cff.

purcell:
Nope, the build is a spurious failure, but you'll need the fix I sent you in https://github.com/fourier/cff/pull/1 so that we can merge.

fourier:
Updated, thanks.

purcell:
Super - merged!

PR #3441, opened 2015-12-28
Add git-command recipe

10sr:
Provide a way to invoke Git shell command using minibuffer.

https://github.com/10sr/git-command-el

### DIFF
--- a/recipes/git-command
+++ b/recipes/git-command
@@ -1,0 +1,1 @@
+(git-command :repo "10sr/git-command-el" :fetcher github)


purcell:
Thanks. The package dependencies are wrong here, though. Firstly, `ansi-color` is not an installable package, so you should instead depend on the corresponding Emacs version, e.g. `(emacs "24.2")`. However, I think `ansi-color` is available in all the Emacs versions which support `package.el`, so you should simply remove that dependency.

Also, you must declare dependencies on stable versions of other packages, not on snapshots, so please replace the version numbers of the dependency packages with the version numbers listed on https://stable.melpa.org.

10sr:
Thanks for your review!
I've fixed.

purcell:
Perfect - thanks.

PR #3442, opened 2015-12-28
Add elisp-lint recipe

AdrieanKhisbe:
Add a recipe for emacs linter from @nschum :)

https://github.com/nschum/elisp-lint


purcell:
Thanks

AdrieanKhisbe:
:)

PR #3445, opened 2015-12-28
Add Recipe for evil-textobj-column

noctuid:
[evil-textobj-column](https://github.com/noctuid/evil-textobj-column), written by myself, is a port of a vim plugin and provides text objects for acting on columns.


purcell:
Thanks!

PR #3447, opened 2015-12-29
Add Czech public holidays

hydandata:
This package adds Czech public holidays to calendar and org-diary (if enabled).

### DIFF
--- a/recipes/czech-holidays
+++ b/recipes/czech-holidays
@@ -1,0 +1,1 @@
+(czech-holidays :fetcher github :repo "hydandata/czech-holidays")


purcell:
This is fine, but don't `autoload` the `mapc` form -- packages shouldn't auto-enable themselves when installed. Just let the user call `czech-holidays-add` as necessary.

hydandata:
Thank you for pointing this out. I have applied the suggested change and clarified the description.

milkypostman:
does the var need to be autoloaded?

hydandata:
I was not sure about it, turns out no. Removed it. Thanks.

PR #3449, opened 2015-12-29
Added recipe for xml-quotes

ndw:
This package reads quotations from an XML file providing sequential and random access. It also supports using those quotations in mail signatures.

The package is mine, ndw/xml-quotes is the official repo.

I've tried to make it compatible, but this is my first attempt to add a package to MELPA so I'm prepared to learn I've done something wrong :-)

### DIFF
--- a/recipes/xml-quotes
+++ b/recipes/xml-quotes
@@ -1,0 +1,1 @@
+(xml-quotes :fetcher github :repo "ndw/xml-quotes")


milkypostman:
recipe looks good, couple things I've noticed,
- each function needs to be `xml-quotes-` as a prefix to match emacs community naming standards. `xml-quotes--` for private functions/variables.
- at least one function needs to have a `;;;###autoload` statement before it so emacs knows which functions to put in the autoloads file.
- most of your `defvar` should probably be `defcustom` so that they are customizable. and they should be in the `xml-quotes` customize group.

ndw:
Right. I updated the repo per your suggestions. Thanks!

purcell:
Thanks for this.

PR #3450, opened 2015-12-30
Add helm-ispell

syohex:
This package provides helm interface of ispell. This is useful for correcting misspellings.

Repository: https://github.com/syohex/emacs-helm-ispell


zobi8225:
:+1:

PR #3452, opened 2015-12-31
Redirect dired-single to crocket/dired-single

crocket:
It was originally opened as #3438. `dired-single` has been unmaintained for years. It is time to move the package to github and change the maintainer.


purcell:
Thanks!

PR #3455, opened 2015-12-31
Add recipe for key-leap

MartinRykfors:
Key-leap is a minor mode that populates the margin with a set of n-letter keywords for every visible line in the window. By calling an interactive command followed by the key of a line, the point is moved to that line.

It can be found at https://github.com/MartinRykfors/key-leap, where you can also se a .gif of it in action.

I am the author.


purcell:
Great, thanks.

PR #3456, opened 2016-01-01
Add hydandata-light-theme

hydandata:
A light color theme that has a bluish tint that is very easy on your eyes.

### DIFF
--- a/recipes/hydandata-light-theme
+++ b/recipes/hydandata-light-theme
@@ -1,0 +1,1 @@
+(hydandata-light-theme :fetcher github :repo "hydandata/hydandata-light-theme")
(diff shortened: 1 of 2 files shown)


purcell:
Hey, nice theme! I think you should avoid calling it "light-theme", though, because that name is too generic. Instead, choose some sort of prefix, e.g. `hydandata-light-theme`.

hydandata:
Thanks, have been using it for years :) I renamed as suggested.

purcell:
Thanks - squashed and committed as a single commit. P.S. the keyword "emacs" is redundant for an Emacs package, and you should probably remove it. ;-)

PR #3463, opened 2016-01-03
add recipe start-menu

lujun9972:
start-menu is a little tool that will add a "Start" menu in the menu bar 
You can use the Start menu to executing external program just like the start menu in windows.

### DIFF
--- a/recipes/start-menu
+++ b/recipes/start-menu
@@ -1,0 +1,1 @@
+(start-menu :fetcher github :repo "lujun9972/el-start-menu")


milkypostman:
some comments I see,

- there is not `;;;###autoload` functions defined. you should figure out how you want people to enable the menu. standard practice is to create a "setup" function that does this. you should remove https://github.com/lujun9972/el-start-menu/blob/master/start-menu.el#L296 because we should avoid automatically enable functionality from packagse.
- `start-menu-menu-conf` should probably be customizable as well as being a variable.
- where does `easy-define-menu` function come from?

lujun9972:
1. I have added start-menu-enable and start-menu-disable to enable or disable the menu
2. I have made `start-menu-menu-conf` customizble but I don't know how to write the `:type` property so I just leave it empty. Would you please help me fix it?
3. `easy-define-menu` is defined in easymenu.el which is part of emacs

purcell:
Thanks. I sent you a tiny fix to the package description in https://github.com/lujun9972/el-start-menu/pull/1.

I'm not sure how to write the `:type` for `start-menu-menu-conf`, sorry.

lujun9972:
@purcell  I got a probem. the new version of start-menu need another library called `config-parser`
I added `;; Package-Requires: ((cl-lib "0.5") config-parser)' to the package head 
But when I install `start-menu` from the melpa, the `config-parser` woun't be installed automatically 
I don't know what's the problem. Would you please help me?  Thank you.

purcell:
You would need `((cl-lib "0.5") (config-parser "0.1"))` or similar -- without the version number, the `Package-Requires` line isn't valid.

lujun9972:
oh,that's why. Thanks ;)

PR #3464, opened 2016-01-05
Add caseformat recipe

HKey:
Caseformat is a format based letter case converter.
It helps users to insert uppercase alphabetical characters without shift keys.

I'm the author of it.

Repository:
https://github.com/HKey/caseformat

### DIFF
--- a/recipes/caseformat
+++ b/recipes/caseformat
@@ -1,0 +1,1 @@
+(caseformat :repo "HKey/caseformat" :fetcher github)


purcell:
Thanks, merged.

PS. I *think* that
```el
(define-globalized-minor-mode global-caseformat-mode
  caseformat-mode
  (lambda () (caseformat-mode 1)))
```
can just be
```el
(define-globalized-minor-mode global-caseformat-mode
  caseformat-mode
  'caseformat-mode)
```

HKey:
Thank you for the merging and the advice.

Yes, it can be simplified.
And the quote is unnecessary, isn't it?

```emacs-lisp
(define-globalized-minor-mode global-caseformat-mode
  caseformat-mode
  caseformat-mode)
```
Sorry, I'm going to change it to be able to choose whether to enable the minor mode.
So that code will be changed to another.

PR #3465, opened 2016-01-05
Add recipe for tldr.el

kuanyui:
https://github.com/kuanyui/tldr.el

### DIFF
--- a/recipes/tldr
+++ b/recipes/tldr
@@ -1,0 +1,1 @@
+(tldr :fetcher github :repo "kuanyui/tldr.el")


milkypostman:
hey cool thing. here are comments,

- needs an `;;;###autoload` line above `tldr` function.
- `defvar` should really be `defcustom` so that these things can be customizable.
- do you need lexical binding for this? i think we are still suggesting that if you don't need it that you don't use it.
- thanks for adding the emacs dependency. Is there a reason to require such a new version of emacs? I don't have an issue I was just curious.

kuanyui:
>hey cool thing. here are comments,

Thank you for your comments because I still don't know a lot of things in Emacs Lisp.

>    needs an ;;;###autoload line above tldr function.

Now @syohex push a PR to fix it.

>    defvar should really be defcustom so that these things can be customizable.

Ok, now fixed.

>    do you need lexical binding for this? i think we are still suggesting that if you don't need it that you don't use it.

```elisp
(require 'autoinsert)
(auto-insert-mode t)
```

Emacs built-in `auto-insert-mode` insert that when I create a new `*.el` file.

>    thanks for adding the emacs dependency. Is there a reason to require such a new version of emacs? I don't have an issue I was just curious.

Because I have no elder Emacs version on my system to test...

syohex:
> thanks for adding the emacs dependency. Is there a reason to require such a new version of emacs? I don't have an issue I was just curious.

`with-temp-buffer-window` was introduced at Emacs 24.3. And I confirmed tldr.el works on Emacs 24.3. I updated minimum Emacs version to 24.3 at https://github.com/kuanyui/tldr.el/pull/1

milkypostman:
OK Cool, thanks :+1:

PR #3466, opened 2016-01-06
Add markdown-preview-eww

izuo:
This is markdown realtime preview by eww. 
more detail: https://github.com/niku/realtime-preview.el


AdrieanKhisbe:
LGTM

purcell:
Thanks for this. I think that the name "realtime-preview" is too general: perhaps "markdown-preview-eww" would be clearer? It would also fit better with https://github.com/ancane/markdown-preview-mode, which we're also adding to MELPA.

izuo:
Thanks, I agree to that.

izuo:
Oops

PR #3473, opened 2016-01-08
Updated request-deferred to use new fork

abingham:
This is a sibling PR of [the recent update to request](https://github.com/milkypostman/melpa/pull/3461). It seems smart to keep both the request and request-deferred recipes using the same source repo.


milkypostman:
Thanks

PR #3474, opened 2016-01-09
Added mexican-holidays recipe

shopClerk:
It is a very simple package with the Mexican holidays for the emacs calendar (and therefore, org-mode)


milkypostman:
tyvm

PR #3475, opened 2016-01-09
Add recipe for haxor-mode

krzysztof-magosa:
Hi,
Could you merge recipe for my haxor-mode?
Thank you.

### DIFF
--- a/recipes/haxor-mode
+++ b/recipes/haxor-mode
@@ -1,0 +1,1 @@
+(haxor-mode :repo "krzysztof-magosa/haxor-mode" :fetcher github)


milkypostman:
Just two comments,

1. i think the common practice for packages with modes it to have users add the mode to the `auto-mode-alist` manually or in their configs. It should not be automatically added by the package. This is easily done by doing,

        (eval-after "haxor-mode-autoloads" 
               (progn (add-to-list 'auto-mode-alist '("\\.hax\\'" . haxor-mode))

    The `(progn` may need a `'` before it, i can't remember. I wrote a macro many months ago to do this cleanly.

2. I think that some of our `defvar`s should be `defcustom`s. Notably the `haxor-tab-width` variable should be customizable. But some of the others could stay vars if they are not intended to be customize, but then they should also probably be prefixed with `haxor--` (two dashes).

thanks

purcell:
> i think the common practice for packages with modes it to have users add the mode to the auto-mode-alist manually or in their configs. It should not be automatically added by the package. This is easily done by doing,

In the case of major modes, and particularly when there are no pre-existing major modes for the same file types, it's actually okay for the mode to modify `auto-mode-alist` directly.

purcell:
I'll go ahead and merge this, and you can incorporate @milkypostman's feedback in due course.

krzysztof-magosa:
Thank you for the feedback, I'm Lisp beginner so every advice is invaluable.
I've applied changes and pushed as 0.2.0 version.
Thank you!

purcell:
Yay, great! Thanks for contributing.

PR #3478, opened 2016-01-10
Add Sisyphus as new package.

phillord:


milkypostman:
does this need `dash` as a dep?

phillord:
Strictly no -- dash is in the dependency graph via m-buffer, but it's a reasonable thing to add anyway (so I have!).

milkypostman:
https://github.com/phillord/sisyphus/blob/master/sisyphus.el#L64 ? :)

purcell:
> Strictly no -- dash is in the dependency graph via m-buffer, but it's a reasonable thing to add anyway (so I have!).

Fwiw, you should always declare dependencies locally, so that your code doesn't break unexpectedly e.g. if `m-buffer` stopped depending on `dash`.

phillord:

>> Strictly no -- dash is in the dependency graph via m-buffer, but it's a
> reasonable thing to add anyway (so I have!).
>
> Fwiw, you should always declare dependencies locally, so that your code
> doesn't break unexpectedly e.g. if `m-buffer` stopped depending on `dash`.


Yes, that's why I added it! In this case, I control m-buffer also, so
it's unlikely, but, yes, good practice.

Unfortunately, I can't think of a good way to test for this routinely.

Phil

purcell:
Eventually I'd like to have `flycheck-package` detect this: the guts of that library should get extracted into something like `package-lint.el` which could contain all sorts of checks that I often perform manually when reviewing packages and MELPA changes.

phillord:

> Eventually I'd like to have `flycheck-package` detect this: the guts of that
> library should get extracted into something like `package-lint.el` which could
> contain all sorts of checks that I often perform manually when reviewing
> packages and MELPA changes.


That functionality really needs to be in Emacs core, with flycheck
uncovering it.

It should not be too hard to implement either, at least not for symbols
in packages that are namespace clean.

Phil

PR #3481, opened 2016-01-11
Add markdown-preview-eww

izuo:
follow purcell's advice.

### DIFF
--- a/recipes/markdown-preview-eww
+++ b/recipes/markdown-preview-eww
@@ -1,0 +1,1 @@
+(markdown-preview-eww :repo "niku/realtime-preview.el" :fetcher github)


vyp:
cc @niku referencing #3466. The CI build error is: "Single file realtime-preview.el does not match package name markdown-preview-eww". So to fix I think you need to rename `realtime-preview.el` to `markdown-preview-eww.el` if you agree with @purcell's naming suggestion.

vyp:
Oh I didn't know you could do that, thanks @izuo, and sorry for the noise!

izuo:
@vyp , 
error
this is not work, I am confuse, maybe I need to read the document again.

vyp:
Oh, I thought you knew that would fix the CI error.

The error is because the file `realtime-preview.el` has to have the same name as the package, because I think that's just how emacs works. However, I'm not 100% sure, so wait for @purcell or someone more knowledgable to guide accordingly.

izuo:
@vyp , yes, I know that, but I find this is work when Multiple Packages in one Repository. such as:

no

So, I have some misunderstanding about it.

vyp:
Yes that example works because in the first case, `projectile` is the package name, and `projectile.el` is the file. In the second case, `helm-projectile` is the package name, and `helm-projectile.el` is the file, and the same concept for the third case.

Your first recipe submitted in this pull request did not come with any `:files` section:

``` elisp
(markdown-preview-eww :repo "niku/realtime-preview.el" :fetcher github)
```

When there's no `:files`, MELPA defaults to the following `:files` value:

``` elisp
("*.el" "*.el.in" "dir"
 "*.info" "*.texi" "*.texinfo"
```

As you can see, the first value there, `"*.el"`, takes all files ending in `.el` from the root of the repository. So for the https://github.com/niku/realtime-preview.el repository, it will take the file `realtime-preview.el` because that's the only file ending in `.el` there.

Hence why the CI build fails, because the package name `markdown-preview-eww`, is not the same as this chosen file `realtime-preview.el`. When you added `:files ("realtime-preview.el")` to the recipe, you didn't really make a difference to the recipe because the default value ends up being the same as this. So again, the same CI build error. Hope that's clear!

izuo:
@vyp , thanks a lot :+1:

niku:
Hi, I'm a author of the realtime-preview.el.

I agree that we should change package name to more specific one.
So, I plan to change following like:

https://github.com/niku/realtime-preview.el ->
https://github.com/niku/markdown-preview-eww

https://github.com/niku/realtime-preview.el/blob/master/realtime-preview.el ->
https://github.com/niku/markdown-preview-eww/blob/master/markdown-preview-eww.el

Will we solve this issue if the plan is applied?

vyp:
@niku Yes, I believe it should solve the issue. What I'm not sure of however, is if you really need to rename the file or not, or whether it's just a subjective stylistic recommendation by the MELPA people.

niku:
Thank you. I have changed it.

Would you check it and update pull-request @izuo?
Or should I create a new pull-request?

izuo:
@niku , thanks :+1: , I have updated it.

purcell:
Thanks all -- this is exactly the right result. Recipe added as a single commit in d9b3ad9.

@niku You should be careful of the case where `output-file-name` contains quotes. The function `shell-quote-argument` might help you... :-)

PR #3482, opened 2016-01-11
add yahoo-weather-mode

lujun9972:
Yahoo-weather-mode is a minor mode which displays weather information in mode-line. 

The weather information is coming from yahoo weather


purcell:
Nice, thanks.

Issue #3480, opened 2016-01-11
flycheck package ships development files


marsam:
Hi:

The [flycheck](https://github.com/flycheck/flycheck) package mistakenly adds `flycheck-ert.el` and `flycheck-buttercup.el` which are intented for development.

Should be enough to add `*-ert.el` and `*-buttercup.el` to `:exclude` in [`package-build-default-files-spec`](https://github.com/milkypostman/melpa/blob/10bf153/package-build.el#L139-L144)?

purcell:
Thanks, yes, that makes sense to me. Would you like to file a pull request?

lunaryorn:
These files were [deliberately part of Flycheck's package](https://github.com/flycheck/flycheck/blob/fe8b311cd90a7dc5cbfe1548104ca0b9f9212e9f/Cask#L6) to support unit tests in dependent packages.  They should never have been removed, and I would like to ask you to bring them back.  The removal has broken or will break test suites and Travis CI for many dependent packages.

I am surprised, disappointed, and also somewhat angry that changes to Flycheck's package—which is arguably an important one—were made without asking us.

lunaryorn:
I'd hope that in future we're asked before 3rd parties make changes to our package on our behalf, without even asking us for comment.

purcell:
Yep, that was unfortunate, sorry. We also can't always get everything right for all packages. Thanks for helping to correct this: I've reverted the related commit.

lunaryorn:
@purcell Thank you very much :)

purcell:
Should we just make flycheck depend on buttercup? These compilation errors at install time make flycheck look uncharacteristically bad:

```
Compiling file /Users/steve/.emacs.d/elpa/flycheck-20160216.138/flycheck-buttercup.el at Thu Feb 18 08:26:25 2016
Entering directory ‘/Users/steve/.emacs.d/elpa/flycheck-20160216.138/’
flycheck-buttercup.el:39:1:Error: Cannot open load file: No such file or directory, buttercup
```

(I could also imagine a "flycheck-dev" package...)

lunaryorn:
@purcell I don't know.  I tend to ignore these error but then again I'm spoiled by `org-plus-contrib` which runs amok on these warnings.

[...]

PR #3484, opened 2016-01-11
Update list of excluded files

marsam:
Add `*-ert.el` and `*-buttercup.el` to the list of excluded files. See: #3480

### DIFF
--- a/README.md
+++ b/README.md
@@ -283,7 +283,7 @@ ending in `test.el` or `tests.el`. See the default value below,
         ("*.el" "*.el.in" "dir"
          "*.info" "*.texi" "*.texinfo"
          "doc/dir" "doc/*.info" "doc/*.texi" "doc/*.texinfo"
-         (:exclude ".dir-locals.el" "test.el" "tests.el" "*-test.el" "*-tests.el"))
+         (:exclude ".dir-locals.el" "test.el" "tests.el" "*-test.el" "*-tests.el" "*-ert.el" "*-buttercup.el"))
 
     This option is necessary when there are multiple packages in the
 repository and thus the package should only be built from a subset of
--- a/package-build.el
+++ b/package-build.el
@@ -140,7 +140,7 @@ function for access to this function")
   '("*.el" "*.el.in" "dir"
     "*.info" "*.texi" "*.texinfo"
     "doc/dir" "doc/*.info" "doc/*.texi" "doc/*.texinfo"
-    (:exclude ".dir-locals.el" "test.el" "tests.el" "*-test.el" "*-tests.el"))
+    (:exclude ".dir-locals.el" "test.el" "tests.el" "*-test.el" "*-tests.el" "*-ert.el" "*-buttercup.el"))
   "Default value for :files attribute in recipes.")
 
 (defun package-build--message (format-string &rest args)


purcell:
I think there might be a mention of the defaults in the documentation, too -- would you mind checking, please, and also updating that?

marsam:
Updated :)

purcell:
That's great - thanks a lot!

lunaryorn:
Please revert this change.  These files were deliberately part of our package, and should never have been removed.

@purcell With all due respect, I'm considerably surprised that apparently any 3rd party can make breaking changes to Flycheck's package without our consent, let alone comment, and without making us aware.  I must admit that this has rocked my confidence in MELPA, and I'd prefer if this does not happen again.  I'd appreciate if any future changes to Flycheck's package did not occur without our explicit consent.

@marsam I appreciate your work and your effort, but I would have preferred if we had been asked for comment and consent first.

purcell:
@lunaryorn What do you think we should do?

milkypostman:
@lunaryorn i'm really sorry that this happened. we try our best to avoid causing developers more pain but inevitably sometimes these things happen. But I think i would rather *encourage* people to contribute and *guide* them when things go wrong---accepting and forgiving when mistakes happen. I think we can take away some knowledge and be more cautious about changes like this. It seemed fairly harmless. We try to put the onus of packaging on ourselves which takes some of the control out of developer's hands. 

at the same time I can easily see this as something either myself or @lunaryorn may have done and not even thought twice to consult you since it seemed externally harmless. Mainly because most of us open source devs have enough on our plate i know I personally appreciate when people take ownership and try to do the right thing.

but again, I understand that it would've been right to /cc you on one of the threads and i'm sure @marsam will do this in the future. it's etiquette, something you can see in #3510 that I follow, but something that one might not always "naturally* do.

with that said, I think that a consistent pattern of this should rock your confidence but one mistake is going to happen every once and a while.

sorry for the problems it caused.

purcell:
I think you meant "something either myself or @purcell may have done". And yes, I didn't think to check more than very superficially in this case. Totally agree with everything you wrote.

lunaryorn:
@milkypostman My confidence is shaken not so much my the issue itself—mistakes happen, no harm done :relaxed:.   But I'm surprised that this change was given only superficial review given that it actually changed the contents of the package and removed libraries :frowning:.  I had expected that we were at least mentioned when a 3rd party makes a change that removes files from Flycheck's package :pensive: 

I understand that we essentially have no control about the contents of our package, but to see the lack of control taking effect so directly still makes me feel uncomfortable :worried:

milkypostman:
I see what you mean. I think the main issue is that we forgot that packages
are sometimes used by cask or other tools/packages for testing. That's my
biggest takeaway. I'm sure that Purcell gave it a very deep evaluation to
make sure it wouldn't effect the most common use case: users loading the
package and using it in emacs.

Regardless it's painless for us to ping people in the future. No reason,
especially for someone who's created such an important package, we can't do
that.

lunaryorn:
@milkypostman Thank you very much :blush:

marsam:
I'm sorry, I should have informed it to the affected party, I'll do it in the future.

Issue #3487, opened 2016-01-12
magit recipe missing a file


prsteele:
I believe the recipe for magit is missing the file ``magit-autorevert.el``. As of now the elisp command ``magit-status`` fails, complaining about that file missing. If I manually add that file from https://github.com/magit/magit everything works correctly. I believe the recipe should be

```(magit :fetcher github
       :repo "magit/magit"
       :files ("lisp/magit-utils.el"
               "lisp/magit-section.el"
               "lisp/magit-git.el"
               "lisp/magit-mode.el"
               "lisp/magit-process.el"
               "lisp/magit-core.el"
               "lisp/magit-diff.el"
               "lisp/magit-wip.el"
               "lisp/magit-apply.el"
               "lisp/magit-log.el"
               "lisp/magit.el"
               "lisp/magit-sequence.el"
               "lisp/magit-commit.el"
               "lisp/magit-remote.el"
               "lisp/magit-bisect.el"
               "lisp/magit-stash.el"
               "lisp/magit-blame.el"
               "lisp/magit-ediff.el"
               "lisp/magit-extras.el"
               "lisp/magit-autorevert.el"
               "lisp/git-rebase.el"
               "Documentation/magit.texi"
               "Documentation/AUTHORS.md"
	       "COPYING"))
```

syohex:
Please see #3488

purcell:
Fixed as part of #3488. New package has not yet built, though.

PR #3488, opened 2016-01-13
Update Magit recipe

syohex:
- Add magit-autorevert.el

This is related to #3487, https://github.com/syl20bnr/spacemacs/issues/4585


purcell:
Thanks. Ideally we would use a combination of `"lisp/magit-*.el"` together with `:exclude` so that this doesn't happen again. Might you be able to prepare a further PR which does that?

syohex:
Will do.

PR #3489, opened 2016-01-13
Improve git clone speed

syohex:
'git clone' is too slow for repository which has long histories like
magit, helm etc. And git histories are not necessary if latest version
is used. So if there are no :commit and :branch parameters in recipe,
'--depth=1' option is set for fast cloning.

### DIFF
--- a/package-build.el
+++ b/package-build.el
@@ -534,7 +534,10 @@ Return a cons cell whose `car' is the root and whose `cdr' is the repository."
         (when (file-exists-p dir)
           (delete-directory dir t))
         (package-build--princ-checkout repo dir)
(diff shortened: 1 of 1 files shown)


purcell:
We need git histories because the latest commits are not always ones which affect the files included in the package. If an excluded "*-tests.el" file is updated in the most recent commit, that shouldn't change the timestamp of the package. And sometimes elisp files are included in very large repositories, e.g. erlang.

I expect this commit would stop that from working properly. Or am I misunderstanding what `--depth=1` does?

syohex:
>  If an excluded "*-tests.el" file is updated in the most recent commit, that shouldn't change the timestamp of the package.

Thanks for explanation. I missed this point.

PR #3490, opened 2016-01-13
Improve magit recipe by using glob pattern

syohex:
See
- https://github.com/milkypostman/melpa/pull/3488#issuecomment-171193897


purcell:
Thanks! You know, I'll just ping @tarsius here, so that he knows we're changing the recipe like this.

recipes/magit
```diff
@@ -1,26 +1,8 @@
...
-               "lisp/magit-sequence.el"
-               "lisp/magit-commit.el"
-               "lisp/magit-remote.el"
-               "lisp/magit-bisect.el"
-               "lisp/magit-stash.el"
-               "lisp/magit-blame.el"
-               "lisp/magit-ediff.el"
-               "lisp/magit-extras.el"
-               "lisp/magit-autorevert.el"
+       :files ("lisp/magit-*.el"
```
kouhin: @syohex @purcell I noticed that `magit.el` is not included here. Is it OK?
before: 
https://travis-ci.org/milkypostman/melpa/builds/101990888
after: 
https://travis-ci.org/milkypostman/melpa/builds/102039195

syohex: Thanks for information and very sorry. I have sent PR #3491.

tarsius:
Thanks. That's better.

martinp:
@Titan-C #3491 needs to be merged.

purcell:
Done, and I'm forcing an immediate build of the package too, since many people rely on it.

PR #3494, opened 2016-01-13
Add worse-defaults

hydandata:
Things in this package represent better (worse?) default settings for
many things that Emacs ships with.

This defaults are also somewhat opinionated, which might not be
acceptable for everyone, hence `worse` in the title.

##### What is included?

In short, a lot. This is definitely not one of those minimalist packages
that try to find the golden middle between sensible and practical, use
`better-defaults` if you prefer something like that.

A convenience function `goto-matching-paren` is also provided, it is bound
to `C-c {`.

### DIFF
--- a/recipes/worse-defaults
+++ b/recipes/worse-defaults
@@ -1,0 +1,2 @@
+(worse-defaults :fetcher github :repo "hydandata/worse-defaults")
+


purcell:
Hi! Thanks, but I'm going to decline this, because it's really a bunch of opinionated personal configuration. While that's okay, it's not a widely useful package, and I don't think it belongs in MELPA. 

Additionally, none of the functions or identifiers it introduces are prefixed with `worse-defaults-`, so it wouldn't (yet) be valid package code in any case.

(FWIW, I also don't think `better-defaults` is a good package.)

hydandata:
Thanks for the feedback! I mostly agree on it not being such a good idea, but did not see anybody else try, hope I did not miss some in search, and made a PR anyway. At least now something will come up in search :)

PR #3496, opened 2016-01-15
add org-rtm

pmiddend:
Description: Simple import/export from rememberthemilk to org-mode
Package repository: https://github.com/pmiddend/org-rtm

I'm the author (and maintainer, of course). I've tested the package in a sandbox.

### DIFF
--- a/recipes/org-rtm
+++ b/recipes/org-rtm
@@ -1,0 +1,2 @@
+(org-rtm :repo "pmiddend/org-rtm"
+	 :fetcher github)
--- a/recipes/rtm
+++ b/recipes/rtm
@@ -1,0 +1,1 @@
+(rtm :repo "pmiddend/emacs-rtm" :fetcher github)


purcell:
Thanks! I'm not super happy that this bundles `rtm.el`, which appears to be an upstream library which has been copied into the repo. Indeed, [simple-rtm](https://github.com/mbunkus/simple-rtm/tree/master/lisp) also bundles `rtm.el`. Which is the authoritative copy? We should make `rtm` a separate package so that both `org-rtm` and `simple-rtm` can depend on it.

pmiddend:
Absolutely agree with you. I modified the commit to include a package "rtm" and I will contact the author of simple-rtm so he can modify his package to depend on it.

purcell:
Yeah, if you maybe just submit a PR which adds a `Package-Requires` header to `simple-rtm` upstream, then I can take care of splitting the recipes.

pmiddend:
Okay, merge complete.

purcell:
Great!

PR #3498, opened 2016-01-16
Add flycheck-purescript recipe

marsam:
Hi:
I'd like to add a recipe for [flycheck-purescript](https://github.com/emacs-pe/flycheck-purescript).

Thanks in advance.

### DIFF
--- a/recipes/flycheck-purescript
+++ b/recipes/flycheck-purescript
@@ -1,0 +1,1 @@
+(flycheck-purescript :fetcher github :repo "emacs-pe/flycheck-purescript")


purcell:
Looks good, but you shouldn't [use autoloads to auto-enable the package](https://github.com/emacs-pe/flycheck-purescript/blob/master/flycheck-purescript.el#L185-L186). By all means provide an autoloaded command which users can use to do that, or do it non-autoloaded [like this](https://github.com/purcell/flycheck-ledger/blob/master/flycheck-ledger.el) -- you can always tell people to use `eval-after-load` to `(require 'flycheck-purescript)`. Note also that according to its docs, `eval-after-load` is meant for user configuration code, not for use inside libraries.

marsam:
Hi, thanks for your review, I've updated the package

purcell:
Great, but I think the setup is still a bit messy -- it's going to call `flycheck-purescript-configure` in every `flycheck` buffer, which isn't what you want. Why not just make `(flycheck-purescript-project-root)` cache its result in a local variable?

marsam:
> Great, but I think the setup is still a bit messy -- it's going to call flycheck-purescript-configure in every flycheck buffer, which isn't what you want.

I wasn't sure about that, but seems to be used in [flycheck-rust](https://github.com/flycheck/flycheck-rust/blob/e03c895/flycheck-rust.el#L33) to set `flycheck-rust-library-path`, and I think flycheck-cask and flycheck-haskell use the same method

> Why not just make (flycheck-purescript-project-root) cache its result in a local variable?

Could you expand please

purcell:
> I wasn't sure about that, but seems to be used in flycheck-rust to set flycheck-rust-library-path, and I think flycheck-cask and flycheck-haskell use the same method

Yep, those are suboptimal too IMO. :-)

Hey @lunaryorn, any idea what the reason for that was?

> > Why not just make (flycheck-purescript-project-root) cache its result in a local variable?

> Could you expand please

Sure:

```el
(defvar-local flycheck-purescript-project-root-cached nil)
(defun flycheck-purescript-project-root ()
  (setq flycheck-purescript-project-root-cached
        (or flycheck-purescript-project-root-cached
            ...(locate-dominating-file etc.)...)))
```

Then just call `flycheck-purescript-project-root` whenever you need the value.

lunaryorn:
@purcell I don't know; simplicity, mostly.  It's probably not as efficient as it could be, but it's simple and does no harm (some directory traversals give or take should make no difference).

I'm not sure, though, how caching in a local variable would make any difference here.  A local variable is, well, local, as such at some point the initial setup has to be done.

I think we're probably talking past each other.  What's the actual problem here?

purcell:
> I think we're probably talking past each other. What's the actual problem here?

Just seems far from ideal to do potentially expensive checker-specific setup in every flychecked buffer. Feels like we could come up with a better pattern.

> I'm not sure, though, how caching in a local variable would make any difference here. A local variable is, well, local, as such at some point the initial setup has to be done.

I was thinking this initial setup would ideally happen when the checker is run, ie. lazily, rather than eagerly before any checker is selected.

But in any case, I'll go ahead and merge this for now.

lunaryorn:
@purcell Flycheck Haskell and Flycheck Rust don't perform a full setup in every buffer.   They look for the presence of the corresponding project file, and exit early if that doesn't exist.  Looking for a project file is cheap, imho; essentially it's just an upwards directory traversal and bit of stat'ing directory entries.

We could surely come up with a better pattern, but the pain hasn't been large enough yet.  Actually there hasn't been any real pain yet for any user as far as I am aware.

A less eager setup would involve additional hooks in Flycheck, I think.  None of the existing Flycheck hooks serves the purpose of lazily initialising options, in my opinion.

purcell:
The directory traversal and `stat`-ing is only *relatively* cheap, I'd say.

I actually think more hooks in Flycheck would be a good thing. It's the standard Emacs way of providing extension points, after all. A hook per checker which runs whenever the checker is enabled in a buffer would do the trick here.

lunaryorn:
> The directory traversal and stat-ing is only relatively cheap, I'd say.

It appears to be cheap enough, though :smiley: 

> I actually think more hooks in Flycheck would be a good thing. It's the standard Emacs way of providing extension points, after all.

I don't disagree, but they weren't needed yet, and I think that there are more urgent tasks in Flycheck than refactoring a system that doesn't appear to be broken from my point of view.

> A hook per checker which runs whenever the checker is enabled in a buffer would do the trick here.

No, I don't think so.  The value of options can influence (via `:predicate`) whether a syntax checker is actually used or not.  Take a look at the `rust` checker.

In other words, we can't know whether we might use a syntax checker before we have taken a look at the project settings, i.e. the build file.  A hook that runs right before a syntax checker is actually used is too late.

We'd need a hook than runs after narrowing down the list of syntax checkers by the current major mode, but before `:predicate` runs.  But that's somewhat difficult currently, because the syntax checker selection is not a two phase process.  We look at everything at once, there's no place for such a hook currently.  We'd after to refactor the syntax checker selection deeply before we could introduce it.

At best we could currently add a hook than runs before the first syntax check, but that'd not be much better than what we have now, since at this point we'd still have to probe for all languages/project files.  It'd just happen a little later.

PR #3499, opened 2016-01-17
include elm.tags file with elm-mode

Bogdanp:
I added automatic TAGS generation based on a list of regular expressions to elm-mode this morning, I ended up using a separate file for the tags' regexps so that needs to be included with the package for the functionality to work.


purcell:
That's cool, but please use `:defaults` rather than just `"*.el"`, as noted in the README. It's more future-proof. You can amend and force-push the recipe commit in order to cleanly update the PR.

Bogdanp:
Done :)

purcell:
Great, thanks.

PR #3500, opened 2016-01-17
Add recipe for slirm

fbie:
Hi,

I'd like to add a recipe for Slirm, which helps users annotating individual entries in BibTeX files for systematic literature reviews. I am the author and will also maintain it. This is my first Emacs mode, so I'd be grateful for any suggestions for improvement. The GitHub address is http://github.com/fbie/slirm.

Thanks,
Florian

### DIFF
--- a/recipes/slirm
+++ b/recipes/slirm
@@ -1,0 +1,2 @@
+(slirm :fetcher github
+       :repo "fbie/slirm")


purcell:
Thanks! Quick comments:

- You should remove the `:files` spec from the recipe: where the default is adequate, we prefer that you don't override it unnecessarily.
- Only use a `;;;` for header lines in the opening comment section. For example, the copyright text should all be prefixed with `;;` instead. `;;;` has a special meaning. (`M-x auto-insert` will always insert "standard" boilerplate for a fresh file.)
- Consider adding `;;;###autoload` cookies for any commands which the user might want to use to launch slirm.
- You're using `subr-x` - depending on what functions in there you're relying on, you'll probably need to declare a package dependency on a recent Emacs version, e.g. `;; Package-Requires: ((emacs "24.4."))`

Hope that helps!

fbie:
Thanks for your feedback! I fixed what you suggested and the package installs and works fine on my system. Hope that everything is well now.

PR #3501, opened 2016-01-18
Add lines-away

siassaj:
Gives you awesome linum style effect except as lines distance away from point


syohex:
Could you tell me what different from [linum-relative](https://github.com/coldnew/linum-relative) ?

siassaj:
absolutely nothing, never knew linum relative existed. It even has a little -> instead of 0. I guess close the pull request :D

purcell:
Nice work though -- proves it was a good idea! :-)

PR #3504, opened 2016-01-18
new recipe 'zone-select'.

kawabata:
Current emacs 'play/zone.el' does not provide the features of selecting, previewing, adding or removing zone programs from 'zone-programs' variable.  This code provides such features.

### DIFF
--- a/recipes/zone-select
+++ b/recipes/zone-select
@@ -1,0 +1,2 @@
+(zone-select :repo "kawabata/zone-select"
+             :fetcher github)


purcell:
Super, thanks.

milkypostman:
i don't know a lot about *zone*. Is this the package that is part of Emacs?

i'm not opposed to adding this package but is there any merit in sending PR to upstream to make their vars customizable?

purcell:
Yes, it's bundled. And yes, this seems like an opportunity for a patch to Emacs' library itself.

PR #3505, opened 2016-01-18
add new receipe `zone-rainbow'.

kawabata:
This is a new zone program, like `zone-nyan' package.
Document is modified to reflect the addition of new package `zone-select' to add
this program to existing zone program lists.


purcell:
Thanks!

PR #3509, opened 2016-01-19
Add kaomoji.el

kuanyui:
https://github.com/kuanyui/kaomoji.el

### DIFF
--- a/recipes/kaomoji
+++ b/recipes/kaomoji
@@ -1,0 +1,1 @@
+(kaomoji :repo "kuanyui/kaomoji.el" :fetcher github)


kuanyui:
BTW, I want to ask that should I include `dict.sqlite3.xz` in the recipe for [moedict.el](https://github.com/kuanyui/moedict.el)?

purcell:
I feel like these should be called `helm-moedict` and `helm-kaomoji`, because they both only work with `helm`, which is a heavyweight dependency.

> BTW, I want to ask that should I include dict.sqlite3.xz in the recipe for moedict.el?

If it's necessary for the package to work, then yes: add `:files (:defaults "dict.sqlite3.xz")` to the recipe.

kuanyui:
Eh...I didn't add a "helm-" prefix because I feel it makes the package name ugly, just like a married name or so... (In fact -- Though it's my personal preference -- I hate the word "helm". If it is called its previous name "anything", I would very glad to call it "anything-kaomoji.el" without hesitate.)

> If it's necessary for the package to work, then yes: add :files (:defaults "dict.sqlite3.xz") to the recipe.

In fact, I've written some mechanism to download `dict.sqlite3.xz` automatically from my GitHub repository (but VERY slow when I test it.), I'm just worried if it will cost your server disk space & bandwidth too much.

purcell:
> Eh...I didn't add a helm- prefix because I feel it makes the package name ugly, and just like a married name or so... (In fact -- Though it's my personal preference -- I hate the word helm. If it is called its previous name anything.el, I would very glad to call it anything-kaomoji.el without hesitate.)
> 

Nah, let's just keep this name, then. But perhaps you'd like to mention in the package description line that it uses helm.

> In fact, I've written some mechanism to download dict.sqlite3.xz automatically from my GitHub repository (but VERY slow when I test it.), I'm just worried if it will cost your server disk space & bandwidth too much.

Ah yes, it's quite big, but I think that's not a huge problem for us, so let's try bundling it into the package for now.

kuanyui:
Got it. Thank you!!! :smile:

PR #3510, opened 2016-01-20
Set git repo to SourceForge.

dykstrom:
Hi!

I am the author of package log4j-mode which currently is fetched from emacsorphanage. Now, I have created a git repo for this package over at SourceForge. I have also added some new functionality which I would like published.

I have modified the recipe for log4j-mode so it looks like the onefor my other package, jtags. However, I have not managed to install a locally built package using this recipe. On the other hand, I cannot build and install jtags locally either, but it seems to work fine on your server. The error message I get is that the package is missing a Version or Package-version header though it clearly has one.

The git repo is found at: git://git.code.sf.net/p/log4j-mode/code.

I would appreciate any help on how to proceed. It would also be fine with me if you did the recipe change.

Best regards,

Johan Dykström

### DIFF
--- a/recipes/log4j-mode
+++ b/recipes/log4j-mode
@@ -1,1 +1,4 @@
-(log4j-mode :fetcher github :repo "emacsorphanage/log4j-mode")
+(log4j-mode
+ :fetcher git
+ :url "git://git.code.sf.net/p/log4j-mode/code"
+ :files ("src/*.el"))


milkypostman:
@tarsius FYI

tarsius:
@milkypostman Thanks for the notice.

tarsius:
@dykstrom Any particular reason why you are using sourceforge instead of something more modern and friendly like github, gitlab, or bitbucket?

dykstrom:
@tarsius It's just for historical reasons. When I started on log4j-mode a long time ago, the only service of this kind I knew was sourceforge. I have not done anything to log4j-mode for several years, and now when I did, it never occurred to me to move to github. If I start on something new I will probably not put it on sourceforge.

tarsius:
Thanks for the info

PR #3511, opened 2016-01-20
erc-hipchatify: add new recipe

seanfarley:
This package improves interaction with hipchat over bitlbee by using the hipchat api to download and display emoticons. Full link here: http://bitbucket.org/seanfarley/erc-hipchatify

I am the package author.

### DIFF
--- a/recipes/erc-hipchatify
+++ b/recipes/erc-hipchatify
@@ -1,0 +1,1 @@
+(erc-hipchatify :fetcher bitbucket :repo "seanfarley/erc-hipchatify")


purcell:
Thanks! A few little issues here:

- `setq-local` only came along in Emacs 24.3, while your dependency is on 24.2.
- `company` doesn't seem like a reasonable lurking dependency here: not everyone uses or wants `company`, so having it both auto-installed and auto-enabled would be surprising. Consider instead adding a function to `completion-at-point-functions` (which `company` can use), and/or splitting out a small package containing the `company` backend. Or maybe this whole thing is `company-erc-hipchatify`...?
- The `eval-after-load` form is not an appropriate thing to `autoload`, because it has the effect of auto-enabling the package once it is simply installed. Just like all the Emacs bundled libraries, it should need to be explicitly enabled. And `eval-after-load` is intended primarily for use in user configs. Just define that `erc` module at the top level and tell the user to use `(eval-after-load 'erc '(require 'erc-hipchatify))`.
- A quick run-through with `M-x checkdoc` would be worthwhile

Hope that helps!

seanfarley:
Thanks for the review! I'll work on addressing your comments and resend. Should I bump the version tag to something like `0.1.1`?

purcell:
The version doesn't really make much difference to MELPA, particularly prior to inclusion. I'll just leave this pull request open for now.

seanfarley:
Ok, I think I've addressed all the feedback. I've pushed this to my bitbucket repo. Anything else special I should do?

purcell:
Looks great, thanks - merged!

seanfarley:
Thanks! I feel so hip now!

PR #3515, opened 2016-01-21
remove shadow

tarsius:
See #3380.


purcell:
Thanks for this Jonas.

PR #3516, opened 2016-01-21
Add recipe for Contextual

akahl-lshift:
Contextual is a profile management system for Emacs that makes it possible to switch between groups of global variables easily. I am both the author and the maintainer of the package. I tried to follow all the instructions in the README as closely as possible including running a sandboxed instance of Emacs and installing the package inside it.

### DIFF
--- a/recipes/contextual
+++ b/recipes/contextual
@@ -1,0 +1,2 @@
+(contextual :repo "lshift-de/contextual" :fetcher github
+            :files ("contextual.el"))


purcell:
Thanks. You might not have guessed from the name, so perhaps you didn't know that the Emacs 24 theme system is designed for exactly this sort of thing? That's why `custom-theme-set-variables` exists, plus the ability to enable multiple themes concurrently. You're not the first to write a similar system, but I'd encourage you to spend some time investigating the built-in alternative before publishing this. I could easily imagine defining various contexts as themes, and then providing a quick way to switch between them.

akahl-lshift:
Hi @purcell, I actually didn't know that and learning about it is a little frustrating. I was looking for something like this for quite some time, and all I found was http://www.emacswiki.org/emacs/profiles.el, which is not in ELPA. Could you please elaborate on what other systems like this exist already?
Are you essentially saying you'd refactor Contextual to use Emacs' "theme" system under the hood, instead?

purcell:
> Hi @purcell, I actually didn't know that and learning about it is a little frustrating.

Yes, I have an unfair advantage because I'm a MELPA maintainer, so I see thousands of packages. :-)

> I was looking for something like this for quite some time, and all I found was http://www.emacswiki.org/emacs/profiles.el, which is not in ELPA. Could you please elaborate on what other systems like this exist already?

Yes, I think I might previously have declined to add `profiles.el` to MELPA, for the same reasons. Not sure about other similar systems.

> Are you essentially saying you'd refactor Contextual to use Emacs' "theme" system under the hood, instead?

Yes, or that if you did start using themes for this, Contextual might not even be necessary. In the meantime, though, I'll decline this PR, because if the theme system does the right thing in this case we should figure out how to make that more obvious to potential users rather than give them nonstandard alternatives. :-)

akahl-lshift:
Contextual adds the possibility to group "themes" and makes it possible to switch between them. Emacs themes allows multiselection, but that is not the same thing. Grouping is about having themes that are mutually exclusive - Contextual's README lists examples. 
Contextual also adds indication to the mode bar. I'm not aware Emacs themes does this. To sum up, out of the box, themes does not do what Contextual provides. Every user aware of themes would hack it and write there own Contextual-like system, eventually, because they won't find anything like this in ELPA. 
So if I refactor Contextual to use theme's internal functions, would you please reconsider adding it?

akahl-lshift:
@purcell I refactored Contextual on a branch to use the Emacs theme system underneath: https://github.com/lshift-de/contextual/tree/emacs-theme-refactor
It doesn't feel 100% natural and disabling a "theme" before enabling another one in the same context makes the whole screen flash since `disable-theme` resets all frame parameters whenever doing so, hence I commented it out. After performing the refactoring, I am now pretty sure Emacs' theme system was really designed for (visual) themes, only.
Please let me know whether you'd rather let the branch into MELPA or the original version after considering this.

purcell:
Hmm, let me have a think. I don't actually see those changes in that branch, though - perhaps you forgot to push your latest commits?

Seems like the screen flash is something that we can/should fix Emacs to avoid. I've noticed similar things in the past, and now with the likes of @Malabarba, @dgutov et al committing to Emacs itself, that sort of improvement should be more feasible than ever.


PS. Please use `contextual-` as the top-level symbol prefix, not `ctx-`.

Malabarba:
This flash is an indication that the frame display is being eagerly updated, whereas normally redisplay would only happen after all lisp code is done evaluating. I couldn't find any explicit calls to `redisplay` or `sit-for` in the related code, so this is either happening deeper into the C source or is some more complicated side-effect.
Either way, the flash is indeed very distracting, and I've filed a proper bug report so we may gather the knowledge of people who know a lot more about the display engine than me. :-)

akahl-lshift:
Oops. You are right, @purcell, I committed and pushed again, now. Later today I will swap out the prefix.
Thank you for reconsidering!

akahl-lshift:
Also swapped out the short ctx prefix, now.

purcell:
Thanks @akahl-lshift and @Malabarba. I think this has been fruitful. Let's see if we can follow up on the Emacs side of things to eliminate the cause of the flash, because I think that would be a great outcome.

In the meantime, I'd be fine with merging a recipe for either the code currently in the `emacs-theme-refactor` branch of `contextual`, or the code in `master` if you make the `ctx-` -> `contextual-` change there too.

You should also switch to a different prefix key, since `C-c c` is [reserved for end user customisation](https://www.gnu.org/software/emacs/manual/html_node/elisp/Key-Binding-Conventions.html).

akahl-lshift:
Great @purcell! I just merged and pushed everything to master. The keymap prefix is `s-c`, now and the documentation is adapted. The major version has also been increased to reflect the breaking changes. 

Is there any chance I'll have to remove the custom theme portion, again..?

Thanks for pointing out the prefix key clash, I wasn't aware of this.

purcell:
Cool, I've committed the recipe for this. Re. the custom theme portion, not sure what the best plan will be long-term. See how you get on: it'll be your choice.

PR #3518, opened 2016-01-22
Add recipe for mutant

p-lambert:
This package provides an interface for dealing with the [Mutant](https://github.com/mbj/mutant) testing tool, enabling one to easily launch it from a `.rb` file or even from a `dired` buffer. The generated output is nicely formatted and provides direct links to the errors reported by the mutations. I've tried to mimic the `rspec-mode` overall experience as much as possible.

**Link to the repository**: [mutant.el](https://github.com/p-lambert/mutant.el)

### DIFF
--- a/recipes/mutant
+++ b/recipes/mutant
@@ -1,0 +1,1 @@
+(mutant :fetcher github :repo "p-lambert/mutant.el")


purcell:
Thanks. You're auto-enabling this in `ruby-mode` etc. via autoloads, but this is not allowed: just like the packages in the Emacs standard library, packages should need to be explicitly enabled by the user. Maybe give the user a function they can add to their `ruby-mode-hook`, or perhaps just tell them to `(eval-after-load 'ruby-mode '(require 'mutant))`.

p-lambert:
Done, @purcell. Sorry about the redundant commits :grin:

purcell:
Thanks - merged!

PR #3519, opened 2016-01-22
Add recipe hamburg-theme

mswift42:
hamburg-theme is a color theme with a dark blue background and warm low
saturated colors.


purcell:
Please can you lower-case `'Hamburg` to `'hamburg`. The uppercase symbol is unidiomatic.

purcell:
(I also think it might lead to load failures on systems with case-sensitive filesystems, where `Hamburg-theme.el` would not be found to exist.)

mswift42:
Done, and thanks for the heads up.

PR #3521, opened 2016-01-22
Add recipe org-tfl

storax:
Hi, I would like to contribute my org-tfl package.

With org-tfl you can plan journeys with the [Transport for London API](https://api-portal.tfl.gov.uk/)
powered by helm and org mode.
See [org-tfl on github](https://github.com/storax/org-tfl).

### DIFF
--- a/recipes/org-tfl
+++ b/recipes/org-tfl
@@ -1,0 +1,2 @@
+(org-tfl :repo "storax/org-tfl" :fetcher github
+	 :files ("*.svg" "*.el"))
--- a/recipes/org-tfl
+++ b/recipes/org-tfl
@@ -1,2 +1,2 @@
 (org-tfl :repo "storax/org-tfl" :fetcher github
-	 :files ("*.svg" "*.el"))
+	 :files (:defaults "*.svg"))


purcell:
Nice! Re. the recipe, please use `:defaults` rather than `"*.el"`. The former includes the latter, but is preferred.

Re. the code itself, the Helm dependency feels heavy and incongruous. Many people don't get along with helm and would be surprised/dismayed that they (apparently) couldn't use this package without it. Can the helm sources be optional? How far could you get just using an explicit command-oriented workflow and `completing-read`? That way, the user's preferred completion framework will have an opportunity to helm them perform input.

storax:
Hi, thanks for the review. Your suggestions all seem reasonable. I hope i can manage to make helm optional and will try to use `completing-read`. I will get back as soon as I fixed it.

storax:
I implemented your suggestion and removed the helm dependency. It is now using `completing-read`. So far I've tested it with `icicles`, the default and `helm` and it is usable with all of them.

purcell:
Great! Sent you a further minor fix in https://github.com/storax/org-tfl/pull/3.

I tried this out locally in a recent Emacs on OS X, and the use of SVG stopped it from working for me: it seems like that format is not supported by all emacsen, so it might be something worth trying to work around.

storax:
Thank you so much for your work and the inclusion in melpa. I'll try a workaround or worst-case make icon support optional.

PR #3523, opened 2016-01-22
Revert "Update list of excluded files"

purcell:
Reverts milkypostman/melpa#3484


lunaryorn:
@purcell Thank you very much :relaxed:

PR #3525, opened 2016-01-23
Add recipe for evalator

seanirby:
I'm the author of [Evalator](https://github.com/seanirby/evalator) and I'd like to add it to MELPA.  

Evalator is a package for interactive REPL-like evaluation within emacs, plus more.  The idea is that you can try out code like a REPL but results are remembered for subsequent transformation.  You can also send results to elisp commands and functions, even if you used a language other than elisp to generate them.  Evalator uses helm to automatically display results in a helm buffer. 

Evalator comes with an elisp evaluation context, and that is the default, but it is designed to be extended with other context packages.  If I get this into MELPA I will also be submitting this [package](https://github.com/seanirby/evalator-context-cider) to add clojure support to evalator 

Regarding the name,  I suspect you'll want me to add the helm- prefix to it.  I'm certainly open to doing that, but I chose not because it's not used for narrowing a list, which is the typical helm use case.  Let me know what you think.

### DIFF
--- a/recipes/evalator
+++ b/recipes/evalator
@@ -1,0 +1,2 @@
+(evalator :fetcher github
+          :repo "seanirby/evalator")


purcell:
Thanks. Fine with the lack of helm prefix here, because this feels more like a new app built on helm than a helm plugin. Sent you a couple of little fixes.

The following code looked odd to me:

```el
    ;; Prevent history candidate from being selected
    (add-hook 'helm-after-update-hook (lambda ()
                                        (helm-next-line)))
```

I'd have expected `add-hook` to be called with the LOCAL flag here to avoid altering the global `helm-after-update-hook`, but actually given that `:helm-after-update-hook` is passed to `helm` a few lines later, I feel like the whole line is unnecessary: you could just pass a list containing the lambda to `helm`... But hey, I could be wrong.

seanirby:
Thanks so much for taking a look.  I will address all of your comments and pull requests ASAP.  

One of your pull request comments got me thinking about the helm dependency and I was wondering if you had an opinion on this.  

Ideally, I'd like this package to have no dependencies and to provide a built-in way of displaying results.  Kind of like how projectile uses ido by default, but allows support for helm, icicles, and others.  

The problem is I don't think I can use ido since you can't customize the way it generates results based on the user input(I think).  I also don't to write my own solution.  Is it alright if this package always had the dependency on helm-core, but supported different display methods through extension package like ```icicles-evalator``` or something?  If so, would you be alright adding this to MELPA before I add this extension functionality?

purcell:
It'd be fairly easy to provide a built-in way of displaying results. I think that at such a time as you have that, or want to support multiple front-ends, you can split out a `helm-evalator`, `icicles-evalator` etc., and just have the core library provide the support functions they'd need.

seanirby:
Cool,  I will look into doing that.

Regarding your comment about the update-hook,  the code is correct as is, but I really should've renamed the local that I defined.   I just fixed that:
https://github.com/seanirby/evalator/commit/27ccac4fdc9b2c439d2eff09bff8dce52ccff5ee

purcell:
Oh, right - I didn't even notice it was a locally-bound var. :-)

seanirby:
Your PRs have been merged.  Let me know if you need me to do anything else.

seanirby:
ping / @purcell

purcell:
Just one more quick thing: the default `:files` pattern includes `"*.el"`, so you should avoid overriding it in this recipe. If you amend the recipe commit to remove it, then force-push it here, that will cleanly update the PR. Ping me when done and one of us will merge it. :-)

seanirby:
Fixed.  

Once again, thanks for your time and the review.  I think what you and @milkypostman are doing here is great. :)

purcell:
Thanks Sean! Merged.

PR #3526, opened 2016-01-24
add optional pretty-print option for archive-contents

emacs18:
Currently archive-contents file created is a text file with one line.
I normally like to have this file pretty printed so that it is much easier to read.
Also it is much easier to examine the changes from version to version.

### DIFF
--- a/package-build.el
+++ b/package-build.el
@@ -1419,13 +1419,14 @@ Returns the archive entry for the package."
   (interactive)
   (setq package-build--recipe-alist-initialized nil))
 
-(defun package-build-dump-archive-contents (&optional file-name)
+(defun package-build-dump-archive-contents (&optional file-name pretty-print)
   "Dump the list of built packages to FILE-NAME.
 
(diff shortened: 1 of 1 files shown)


purcell:
I think we'd be okay with just pretty-printing this by default, eh @milkypostman, rather than adding an arg? There's no total size concern because it'll be served gzipped anyway. I suppose if it took minutes to pretty-print the full MELPA `archive-contents` that'd be a pain, but it probably doesn't...

emacs18:
My archive-contentts file has 288 packages.  Evaluating
(package-build-dump-archive-contents) with and without pretty printing takes 0.2
and 2.1 seconds respectively. This is on virtual linux running within win7 as
host OS. My laptop is about two years old.

If the additional time is not a problem for you, I am quite ok with it myself.
Please let me know if it is so that I can submit modified pull request that
pretty prints always.

purcell:
Hmm, so that could easily mean 20-30s for our full package list of 2900 packages, which I think is too long. In that case, I'll go ahead and just merge this.

PR #3527, opened 2016-01-24
Add recipe for mocha-mode.

scottaj:
This is a package for running mocha tests with NodeJS.

This is the main repository, of which I am the owner/maintainer: https://github.com/scottaj/mocha-mode

### DIFF
--- a/recipes/mocha
+++ b/recipes/mocha
@@ -1,0 +1,3 @@
+(mocha
+ :repo "scottaj/mocha.el"
+ :fetcher github)
--- a/recipes/mocha-mode
+++ b/recipes/mocha-mode
@@ -1,3 +1,0 @@
-(mocha-mode
- :repo "scottaj/mocha-mode"
- :fetcher github)
--- a/recipes/mocha-mode
+++ b/recipes/mocha-mode
@@ -1,0 +1,3 @@
+(mocha-mode
+ :repo "scottaj/mocha-mode"
+ :fetcher github)


purcell:
Thanks. Few little issues here:

- The library is called `mocha-mode`, but there's no mode provided. In any case, unless you're providing a major mode the `-mode` suffix should be dropped.
- `run-mocha` should be `mocha-run` -- put the package prefix first
- The `defvar`s should be `defcustom`s, within an associated `defgroup`
- `;;;###autoload` cookies for the initial entry point commands would be advisable
- The `mocha-test-at-point` command makes promises in its docstring which I don't think it can keep, e.g. running the test at point.

It feels like a bit more work is necessary here before sharing this with the wider community. :-)

scottaj:
Hello, thanks for the feedback. I made the following changes:

1. Updated the file to be `mocha.el`, also updated the source repository and recipe here to reflect that.
1. Renamed `run-mocha` to `mocha-run`
1. Turned most of the `defvar`s into `defcustom`s One only makes sense to configure locally per project, so I left that one a var.
1. I was a little confused about this one. after reading the emacs docs and looking at what some other packages were doing I added the autoloads to `mocha-command` and `mocha-run`.
1. Removed `mocha-test-at-point` for now, hopefully I can add that soon.

purcell:
Cool, thanks. Sent you a small fix in https://github.com/scottaj/mocha.el/issues/9.

You shouldn't really be changing the global `compilation-regexp*` vars here: since you're running the compilation yourself, you have the opportunity to set exactly the regexps you want locally to the mocha compilation buffer, which also has the benefit of making the highlighting faster (because it's not looking through all the possible error regexps).

For an example of how to do this, check out [ruby-compilation.el](https://github.com/eschulte/rinari/blob/master/util/ruby-compilation.el).

I'll go ahead and merge this in the meantime, but I'd definitely encourage you to investigate the above.

Cheers!

scottaj:
Awesome, thanks for the tip. I will update that soon for sure. Thank you for all the great feedback.

PR #3530, opened 2016-01-25
add tar-util.el and exclude skk-lookup.el

yutopia:
I am one of maintainer of ddskk. https://github.com/skk-dev/ddskk
new file tar-utile.el was added.
and skk-lookup.el is excluded at this version of ddskk.


purcell:
Thanks!

PR #3531, opened 2016-01-25
Add a recipe for chapel-mode.

russel:
This mode file has been taken from the Chapel repository into this separate repository so that it can be put into MELPA and can be develop in a separate lifecycle to the Chapel compiler and runtime system code itself.

### DIFF
--- a/recipes/chapel-mode
+++ b/recipes/chapel-mode
@@ -1,0 +1,1 @@
+(chapel-mode :fetcher github :repo "russel/Emacs-Chapel-Mode")


purcell:
Thanks. You might consider using `define-derived-mode` to define `chapel-mode`, since it takes care of generating a lot of the boilerplate currently in that command.

russel:
Thanks for the reminder. I am just picking this up and the original is quite old. Shortly I will tag 0.8 to get something into stable as well as having this in "rolling". Thanks for MELPA, it is a superb way of working with Emacs.

purcell:
Happy to hear MELPA is helpful to you. Cheers!

PR #3532, opened 2016-01-25
add dumb-jump

jacktasia:
Howdy,

Just wanting to add my experimental jump to definition package: https://github.com/jacktasia/dumb-jump

### DIFF
--- a/recipes/dumb-jump
+++ b/recipes/dumb-jump
@@ -1,0 +1,1 @@
+(dumb-jump :repo "jacktasia/dumb-jump" :fetcher github)


purcell:
Thanks. Quick feedback:

- I think `dumb-jump-get-project-root` could be simplified a lot by using `locate-dominating-file` in the mode where it is passed a function
- Please wrap your comments, and don't use Markdown
- For `dumb-jump-back`, investigate whether `pop-tag-mark` and `xref-push-marker-stack` will save you some work (example [here](https://github.com/purcell/elisp-slime-nav/blob/master/elisp-slime-nav.el#L85-L88))
- Don't do this:
```el
(if (s-blank? cmd)
       nil
      (dumb-jump-parse-grep-response rawresults cur-file line-num))
```
do this instead:
```el
(unless (s-blank? cmd)
  (dumb-jump-parse-grep-response rawresults cur-file line-num))
```
- There's already `compilation-regexp-alist` etc. for parsing grep results, so you're repeating work here, and I feel like the code could be dramatically shrunk. Consider running the search in a compilation buffer (see function `compilation-start`, for example) and just relying on `compilation-mode`'s parsing. Then if somebody wants to use `ag` or `pt` instead of `grep` (which they should), things should Just Work™.
- I'm wary of all those separate `grep` options: it feels like it would make the whole library closely tied to `grep`. Better to define a whole command template at once IMO.

Hope that helps!

jacktasia:
Thanks for the great feedback!

> I think `dumb-jump-get-project-root` could be simplified a lot by using `locate-dominating-file` in the mode where it is passed a function

Thanks. I did not know about `locate-dominating-file` and I've made changes to use that

> Please wrap your comments, and don't use Markdown

I think I've addressed all of these instances.

> For `dumb-jump-back`, investigate whether `pop-tag-mark` and `xref-push-marker-stack` will save you some work. Is `xref.el` going to be in emacs 25?

I did not know about the marker system. Really good to know, but it doesn't look ilke in this case it would save much work.

> Don't do this (...)

Fixed.

> There's already `compilation-regexp-alist` etc. for parsing grep results,(...)

I spent some time looking into this and I could not figure out how to use `compilation-mode` to parse grep results. I am extremely interested in this if it really will allow me to abstract easily away the search program being used...do you have any pointers on where to look? I've only seen references to `compilation-error-regexp-alist` and couldn't see any usages with `grep` or the like.

Thanks again!

purcell:
> Is xref.el going to be in emacs 25?
> 
> I did not know about the marker system. Really good to know, but it doesn't look ilke in this case it would save much work.

Yes, and the code I linked to shows how to easily do the same thing even without xref.el. At first glance, it looks like it would save a fair bit of code.

> I spent some time looking into this and I could not figure out how to use compilation-mode to parse grep results. I am extremely interested in this if it really will allow me to abstract easily away the search program being used...do you have any pointers on where to look? I've only seen references to compilation-error-regexp-alist and couldn't see any usages with grep or the like.

Check out `grep-mode`, which is built on top of `compilation-mode`. In your position I'd do something like the following: use `compilation-run` to fire off a `grep` command and put the resulting buffer into `grep-mode`. Rename or hide the buffer if necessary, then use `compilation-filter-hook` to get a callback after results have been written into the buffer (this may happen several times for long result sets). The `grep-mode`/compilation machinery will automatically find grep results, so you can just use functions like `compilation-next-error` to jump through results in the file, or there'll be lower-level functions in `compile.el` if necessary.

jacktasia:
> Yes, and the code I linked to shows how to easily do the same thing even without xref.el. At first glance, it looks like it would save a fair bit of code.

Yeah, it makes total sense. Although I have concerns that things would not be isolated then. I could be wrong, but it _seems_ like a user could manually push to the marker stack, and then they would get unexpected results when they use jump back. For instance, a user jumps multiple times and then at some point they see something else they want to go back to when they're done going down the rabbit hole they are in so they manually push to the marker stack. Ideally, they would be able to jump all the way back and then still use their manual markers. I also want a record of the files and line numbers which I would lose out on.

> Check out `grep-mode`, which is built on top of `compilation-mode` ...

I've been messing around with this, but I've been unable to get much of an advantage. Managing the multiple calls to `compilation-filter-hook` doesn't seem to be trivial. I couldn't find much usage of this beyond colorizing the compilation buffer. My current parsing of the grep response was extremely easy to write and for me to understand. I can't say the same for this route. I think there's absolutely a theoretical gain with this, but I've been unable to realize this in practice. Ideally, I'll come back to this if/when there is support for other search programs. (`ag` etc.).

purcell:
Re. the marker stack, that's standard behaviour across many packages which use that technique, so I don't think it would confuse users. Markers can be inspected, so you can get a filename by looking at the marker's buffer, for example.

Anyhow, as far as getting this merged is concerned, I noticed there's a function `message-prin1` which is missing the namespace prefix. Happy to merge this once that's fixed up so you can see where this code ends up taking you. :-)

P.S. I really wouldn't suggest loading all of `org` just for `org-combine-plists` -- you should probably just reproduce that little function locally.

jacktasia:
> Re. the marker stack, that's standard behaviour across many packages which use that technique, so I don't think it would confuse users. Markers can be inspected, so you can get a filename by looking at the marker's buffer, for example.

I know I can get the filename from the marker, but as far as I can tell I can't get the line number from it.

> Anyhow, as far as getting this merged is concerned, I noticed there's a function message-prin1 which is missing the namespace prefix. 

Fixed this.

> I really wouldn't suggest loading all of org just for org-combine-plists -- you should probably just reproduce that little function locally.

Great point. I've pulled this out.

Thanks again for all your great feedback!

purcell:
Great, merged, thanks. Package should build in the next few hours. :-)

> know I can get the filename from the marker, but as far as I can tell I can't get the line number from it.

For this, I think you'd just temporarily jump to the marker inside an elisp function with `save-excursion`, and look at the line number.

PR #3533, opened 2016-01-26
Add ‘char-menu’ package

mrkkrp:
Create your own menu for fast insertion of Unicode charaters.

Here is the repo: https://github.com/mrkkrp/char-menu


purcell:
Great, thanks.

PR #3535, opened 2016-01-27
Add metafmt

lvillani:
[metafmt](https://github.com/lvillani/metafmt) is a small front-end to many beautifier programs. This pull
request adds its Emacs integration to MELPA so that files are formatted
on save.

### DIFF
--- a/recipes/metafmt
+++ b/recipes/metafmt
@@ -1,0 +1,2 @@
+(metafmt :repo "lvillani/metafmt"
+         :fetcher github)


purcell:
Looks good, thanks. Sent you a quick fix in https://github.com/lvillani/metafmt/pull/1.

A colleague has been working on a [similar tool in Haskell](https://github.com/hjwylde/omnifmt) but we haven't written an emacs front end yet. :-)

PR #3537, opened 2016-01-27
Add asy-mode

PythonNut:
This package provides a major mode for the [Asymptote](https://github.com/vectorgraphics/asymptote) vector graphics language. I contacted the current maintainer, John Bowman, to get this ball rolling. I am not officially part of the project, but I am working to get this into MELPA (see [my PR](https://github.com/vectorgraphics/asymptote/pull/8)).  

Hopefully, I haven't made any mistakes. :)

### DIFF
--- a/recipes/asy-mode
+++ b/recipes/asy-mode
@@ -1,0 +1,3 @@
+(asy-mode :fetcher github
+          :repo "vectorgraphics/asymptote"
+          :files ("base/asy-mode.el"))


purcell:
Sorry, merged this prematurely, so I've reverted it in #3543.

I don't think the upstream code is ready for packaging yet:

- There's no consistent naming of top-level symbols: we've got "lasy-*", "asy*",  "pdf-view-command" etc. These should at least *all* be consistently `asy-*`/`lasy-*`
- Setting `mode-name` at the top level is completely bogus.
- Some of the code [added to `asy-mode-hook`](https://github.com/vectorgraphics/asymptote/blob/master/base/asy-mode.el#L621-L630) is a mixture of redundant and inappropriate (e.g. force-disabling `flyspell`), and if it's necessary then it should all be in the mode definition anyway, not a hook function which could be removed.
- The `two-mode-mode` dependency and how that's detected at load time etc. is weird. Should it be a package dependency?
- `eval-after-load` in libraries is generally a code smell (see the docs for that function), and there's a lot of that here
- Should the `lasy-*` stuff all be in a separate file?
- Rebinding F4 isn't a good plan, since that's a built-in binding commonly used when working with macros
- Trailing comment line should have `;;;`, not just `;;`

Hope that helps! Happy to comment further if you have questions.

PythonNut:
CC: @johncbowman

PythonNut:
@purcell Thanks a bunch. I realize there's a lot of work to be done here, and I don't expect it to be finished immediately. 

-----------------------

> There's no consistent naming of top-level symbols: we've got "lasy-", "asy", "pdf-view-command" etc. These should at least all be consistently asy-*/lasy-*

This can probably be fixed without too much trouble.

> Setting `mode-name` at the top level is completely bogus.

Pretty much. Shouldn't be too hard to fix.

> Some of the code added to `asy-mode-hook` is a mixture of redundant and inappropriate (e.g. force-disabling `flyspell`), and if it's necessary then it should all be in the mode definition anyway, not a hook function which could be removed.

Just delete the whole hook, maybe? It looks like code was once part of someone's config. 

> The `two-mode-mode` dependency and how that's detected at load time etc. is weird. Should it be a package dependency?

Not sure how to handle this.

> Should the lasy-* stuff all be in a separate file?

Maybe even a separate package? Then I can defer working with the `two-mode-mode` stuff. Also, there are other solutions to having multiple major modes in a single buffer that render this non-critical.

> `eval-after-load` in libraries is generally a code smell (see the docs for that function), and there's a lot of that here

Two actual instances: (1) [here](https://github.com/vectorgraphics/asymptote/blob/master/base/asy-mode.el#L645) and (2) [here](https://github.com/vectorgraphics/asymptote/blob/master/base/asy-mode.el#L997). Both could probably be moved into the future `lasy` package. (Which, given that I am <code>la<b>z</b>y</code>, may not happen for a long time, if ever.)

> Rebinding F4 isn't a good plan, since that's a built-in binding commonly used when working with macros

<kbd>C-c</kbd> <kbd>g</kbd> perhaps?

> Trailing comment line should have `;;;`, not just `;;`

An easy fix.

purcell:
> @purcell Thanks a bunch. I realize there's a lot of work to be done here, and I don't expect it to be finished immediately.
> 
> There's no consistent naming of top-level symbols: we've got "lasy-", "asy", "pdf-view-command" etc. These should at least all be consistently asy-/lasy-
> This can probably be fixed without too much trouble.
> 
> Setting mode-name at the top level is completely bogus.
> Pretty much. Shouldn't be too hard to fix.
> 
> Some of the code added to asy-mode-hook is a mixture of redundant and inappropriate (e.g. force-disabling flyspell), and if it's necessary then it should all be in the mode definition anyway, not a hook function which could be removed.
> Just delete the whole hook, maybe? It looks like code was once part of someone's config.
> 
> The two-mode-mode dependency and how that's detected at load time etc. is weird. Should it be a package dependency?
> Not sure how to handle this.
> 
> > Should the lasy-* stuff all be in a separate file?
> Maybe even a separate package? Then I can defer working with the two-mode-mode stuff. Also, there are other solutions to having multiple major modes in a single buffer that render this non-critical.
> 
> eval-after-load in libraries is generally a code smell (see the docs for that function), and there's a lot of that here
> Two actual instances: (1) here and (2) here. Both could probably be moved into the future lasy package. (Which, given that I am lazy, may not happen for a long time, if ever.)
> 
> Rebinding F4 isn't a good plan, since that's a built-in binding commonly used when working with macros
> C-c g perhaps?
> 
> Trailing comment line should have ;;;, not just ;;
> An easy fix.

purcell:
> @purcell Thanks a bunch. I realize there's a lot of work to be done here, and I don't expect it to be finished immediately.

:+1:

> Just delete the whole hook, maybe? It looks like code was once part of someone's config.

Yep, not clear. Could be.

> > The two-mode-mode dependency and how that's detected at load time etc. is weird. Should it be a package dependency?
> Not sure how to handle this.

Normally I'd say to make sure there's a package for "two-mode-mode", and just add a package dependency, but this code is also magically redefining code from "two-mode-mode", which is... suboptimal. :-)  I guess the intent of all this needs to be picked apart to figure out what really makes sense.

> > Should the lasy-* stuff all be in a separate file?
> Maybe even a separate package? Then I can defer working with the two-mode-mode stuff. Also, there are other solutions to having multiple major modes in a single buffer that render this non-critical.

Yes, a separate package would be good, I reckon.

> eval-after-load in libraries is generally a code smell (see the docs for that function), and there's a lot of that here
> Two actual instances: (1) here and (2) here. Both could probably be moved into the future lasy package. (Which, given that I am lazy, may not happen for a long time, if ever.)
> 
> Rebinding F4 isn't a good plan, since that's a built-in binding commonly used when working with macros
> C-c g perhaps?

Hmm, not that either. There are some [rules of thumb about this in the elisp manual](http://www.gnu.org/software/emacs/manual/html_node/elisp/Key-Binding-Conventions.html).

> Trailing comment line should have ;;;, not just ;;
> An easy fix.

Great. Well, let me know how you'd like to proceed, and ping me if you have any questions.

PythonNut:
@purcell I'm going to wait for the Asymptote maintainer to give me the green light. Obviously some of these changes have the potential to break people's configs. I'll only work on this if he agrees that this is the path forward.

johncbowman:
Hi Emacs friends,
Yes, you have the green light to bring asy-mode.el up to date, in line with
current standards. I can do some limited testing.
It would be nice to keep it backwards compatible where possible, but there
are far fewer users of asy-mode.el than
asy users (or emacs users, for that matter). So if there are things that
need fixing I would rather do it now than wait.

I haven't used lasy-mode for a while but it can be quite useful. Philippe
Ivalid (pivaldi) is on github if you want to contact him as well.

All the best,

-- John

purcell:
Thanks @johncbowman! /cc @pivaldi.

PythonNut:
Thanks! I'll work on this as soon as I have a free momen.

pivaldi:
Hello,
I do not have the opportunity to work on asy-mode.el now.
It was my first package for Emacs, the code may be improved easily.

PR #3538, opened 2016-01-27
new recipe `zone-sl'.

kawabata:
zone-sl.el is a new `play/zone.el' program which zone out emacs with steam locomotives.


purcell:
Fun, thanks!

PR #3544, opened 2016-01-29
Create org-auto-export

llcc:
Automatically and asynchronously export org-file to pdf by ox-latex

### DIFF
--- a/recipes/org-auto-export
+++ b/recipes/org-auto-export
@@ -1,0 +1,2 @@
+(org-auto-export :repo "llcc/org-auto-export"
+                 :fetcher github)


purcell:
Thanks. A few issues here:

- This shouldn't be a _global_ minor mode
- It seems a waste to hard-code the use of `org-latex-export-to-pdf`, and it doesn't fit with the general name of the package. Why not provide a `defcustom` for the export function so it can be overridden?
- Might it be better to just use `after-save-hook`, rather than periodically saving and exporting the buffer? It's generally bad practice to save the user's file without asking.
- Sent you a small fix in https://github.com/llcc/org-auto-export/pull/2

Hope that helps!

llcc:
Thanks for you advice. I will think about the export mechanism again!!

tarsius:
@llcc Here's a friendly ping for you.

tarsius:
If you would still like to get this added to Melpa, then please address the mentioned issues and reopen this pull request.

PR #3545, opened 2016-01-29
add recipe:blog-admin

CodeFalling:
`blog-admin` is a program for writing blog in emacs. See https://github.com/CodeFalling/blog-admin

### DIFF
--- a/recipes/blog-admin
+++ b/recipes/blog-admin
@@ -1,0 +1,3 @@
+(blog-admin
+ :fetcher github
+ :repo "codefalling/blog-admin")


purcell:
Thanks. Mostly looks good. Some small tips:

- Consider adding `;;;###autoload` cookies for the main startup commands like `blog-admin-start`
- Use `define-derived-mode` to declare `blog-admin-mode` -- this will auto-generate most of the code you currently have in that command
- Don't try to define keybindings based on `(blog-admin-backend-get-backend)` when the file loads -- what if the user changes the backend after loading this file?

Hope that helps!

CodeFalling:
Thanks for your tips,It should be better now!

purcell:
Great, thanks!

PR #3549, opened 2016-01-30
Add recipe for evalator-context-cider package

seanirby:
Hello again,

I'd like to add [evalator-context-cider](https://github.com/seanirby/evalator-context-cider) to MELPA.

This package adds Clojure support to [evalator](https://github.com/seanirby/evalator)

If you want to manually test this then the newest version of evalator is required, which may or may not be built in MELPA yet.

### DIFF
--- a/recipes/evalator-clojure
+++ b/recipes/evalator-clojure
@@ -1,0 +1,3 @@
+(evalator-clojure :fetcher github
+                  :repo "seanirby/evalator-clojure"
+                  :files (:defaults "*.clj"))


purcell:
Thanks! Please can you use `:defaults` rather than `"*.el"`? The former includes the latter, and is preferred when possible. You would ideally amend and force-push the recipe commit to cleanly update the PR.

Apart from that, it all looks good, but you should add info in the `Commentary` section of the header about how to enable/use the package.

seanirby:
Thanks! I will make those changes ASAP and ping you here when it's all ready to go.

seanirby:
*ping* @purcell 

I've addressed your issues.  I also renamed the package and made it depend on the parent evalator package because I needed certain things to autoload correctly.

purcell:
Super, thanks Sean.

PR #3550, opened 2016-01-31
Add recipe for xml+

bddean:
xml+ is a library to query xml/html nodes. 

It's at https://github.com/bddean/xml-plus


purcell:
Thanks!

PR #3551, opened 2016-02-01
Add recipe for dired-atool

HKey:
Dired-atool is an utility to pack/unpack files with [atool](http://www.nongnu.org/atool/) on dired.
This provides some command to call atool.

I'm the author of it, the repository is https://github.com/HKey/dired-atool.


purcell:
Thanks!

HKey:
Thank you for the merging!

PR #3552, opened 2016-02-01
remove otter-mode

tarsius:
The library has been [removed from the upstream repository](https://github.com/scvalex/script-fu/commit/712b2e3ed9ad56570cde7d824ab5a7b11874772d)

There is also [an older `otter-mode`](https://github.com/schaueho/otter-mode). Maybe that should be added to Melpa instead. @schaueho wrote it two decades ago and hasn't really touched it since then.

### DIFF
--- a/recipes/otter-mode
+++ b/recipes/otter-mode
@@ -1,1 +1,0 @@
-(otter-mode :fetcher github :repo "scvalex/script-fu" :files ("otter-mode.el"))


purcell:
Yes, we should probably have just switched the recipe to point at the older version.

schaueho:
JFTR, I have no opinion on whether the ancient otter-mode I wrote or some other should be available on melpa. The Otter theorem prover itself hasn't had any update since 2004, so I can't imagine there are still that many users of it.

purcell:
Thanks @schaueho. From our point of view it's free to publish your code as a package, so we might as well do it. We'll be able to see from the download count whether or not it's getting used.

PR #3553, opened 2016-02-02
org-random-todo

unhammer:
Summary: Shows a random TODO from your org-agenda-files (or other list)
every so often, using message (and optionally notifications-notify).

Link: https://github.com/unhammer/org-random-todo

Association: Author.

### DIFF
--- a/recipes/org-random-todo
+++ b/recipes/org-random-todo
@@ -1,0 +1,3 @@
+(org-random-todo :fetcher github
+                 :repo "unhammer/org-random-todo"
+                 :files ("org-random-todo.el"))


purcell:
Thanks! Couple of suggestions:

- Rather than only ever checking `org-random-todo-cache-idletime` at load time, and then enabling reminders forever, maybe provide a global minor mode which can be toggled on or off at any point
- Just because `notifications` is available doesn't mean the user wants that to be loaded or used in preference to `message`. Why not use the excellent `alert` package so that your code can just call `alert`, and the user can specify how they would like to be alerted?

Looks good otherwise!

purcell:
(See https://github.com/jwiegley/alert for the latter)

unhammer:
Thanks for the suggestions; both should be in there now :-)

By the way, is it possible to require version 1.2 of alert even though melpa gives me something like 20151123.759?

purcell:
Yes, and in fact you *should* depend on "1.2" so that if you tag your code with a version number to make it available on MELPA Stable then all of its dependencies would be installable from there without resorting to MELPA.

The dependency version number is mandatory, BTW, so you'll have to add it to the current `Package-Requires` line.

unhammer:
OK, should be fixed now :-) 

(The Travis failure seems to be during some apt command, is it expected?)

purcell:
Travis failure seems spurious. Package looks good now, though - merging. Thanks!

PR #3554, opened 2016-02-03
Emacs' own autocorrect-mode

jbranso:
This package just abbrev-mode to store in some 130 abbreviations.  For example it expands "teh" into "the".

I'm not sure why web-mode got in the commit list, but that won't be a problem will it?  I just deleted a line.


syohex:
You need to rebase and squash commits. And I recommend you to use topic branch instead of master branch to send PR.

See also
- https://yangsu.github.io/pull-request-tutorial/
- https://help.github.com/articles/using-pull-requests/
- https://www.atlassian.com/git/tutorials/rewriting-history
- https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History

syohex:
And your package does not follow emacs packaging convention. Please see Emacs packaging documentation.

purcell:
Have you looked at the built-in `flyspell` library? It provides the option `flyspell-auto-correct-word`, and I feel like that would be the idiomatic way to solve this problem.

jbranso:
That's true.  I'll change the package to add an advice before/after the space is pressed.  If the last word was composed of only letters, then I'll call flyspell-auto-correct-word.  Thanks for the tip.

purcell:
I assumed `flyspell-auto-correct-word`

jbranso:
That doesn't seem to be the case.  The variable "flyspell-auto-correct-word" is not documented. bummer...  I've set it "flyspell-auto-correct-word" to t and to 1.  But both times "teh" did not autocorrect to "the".  But when I typed "teh" and called flyspell-auto-correct-word, "teh" changed to "the".  

So I could write my own package that adds an advice to call the function "flyspell-auto-correct-word" after the user presses the space key, but before emacs inserts the space char.  

Or I could just hack flyspell to do what I want to do, and try to push that change upstream.

PR #3555, opened 2016-02-03
Add recipe for search-files

dandavison:
Tested by following the instructions in the melpa README:

- `make clean` && `make recipes/search-files`
- `EMACS=$(which emacs) make sandbox`
- `(package-list-packages)` => select "search-files" and use buttons to install

https://github.com/dandavison/emacs-search-files

### DIFF
--- a/recipes/search-files
+++ b/recipes/search-files
@@ -1,0 +1,2 @@
+(search-files :repo "dandavison/emacs-search-files"
+              :fetcher github)


purcell:
Thanks! I see a few issues here:

- It's far from ideal to use a very general name ("search-files") for a package which only works if those files are in a git repo
- You're relying on the `projectile` package, where the function `vc-git-root` would give you the same functionality for free
- You've hard-coded your personal preference to fall back from `git grep` to `ag`, even if that program is not installed. It seems to me like users would be better served by just calling `vc-git-grep` or `ag-project` (in the `ag.el` package) directly as desired.
- There are a couple of special case "definition" regexps hardcoded for python and emacs lisp, but these will never work as well as solutions tailored for those languages, e.g. `elisp-slime-nav`.
- The package metadata/boilerplate is missing copyright, licence and commentary info

Hope that helps!

dandavison:
@purcell thanks for the code review! Those are all very helpful suggestions. To be honest though, I didn't know I was opening myself up for code review on the project itself! This was just something I put together quickly -- my main aim was to familiarize myself with how to create and publish an emacs package. I see now that, quite rightly, you have some standards here! And I agree that my project doesn't meet them yet. Let me try to fix that and get back to you.

purcell:
Hey, free code review! :-)

FWIW, the packaging side of things was fine; I'd suggest enabling `auto-insert-mode`, then when you create fresh elisp files Emacs will guide you through inserting the standard boilerplate. Or call `M-x auto-insert` directly.

dandavison:
> It seems to me like users would be better served by just calling vc-git-grep 

I started to act on this suggestion and thus remembered that my decision not to use commands like `vc-git-grep` was deliberate: it (and related commands) run the shell command asynchronously. This makes it much more awkward for the initiating code to interact with the results buffer. I suspect it made sense when those commands were first designed due to slow search times, but with a modern laptop `git grep` is returning results for me in 0.2s - 0.3s  in a codebase of 5.9 million lines. So I believe that I actually want the main thread to be blocked while waiting for results.

PR #3557, opened 2016-02-03
Add recipe for evil-mu4e

JorisE:
Evil-mu4e sets up keybindings for mu4e that make sense for Evil/VIM
users.

### DIFF
--- a/recipes/evil-mu4e
+++ b/recipes/evil-mu4e
@@ -1,0 +1,1 @@
+(evil-mu4e :repo "JorisE/evil-mu4e" :fetcher github)


purcell:
Thanks. Couple of little issues:
- `((mu4e))` is not a valid dependency. The version number is mandatory.
- You'll also need to declare a package dependency on `evil`
- Don't autoload `evil-mu4e-init` *and* call it at the top level. If someone calls the command via the autoload, it will be invoked twice. Just remove the top level code which calls it
- Please add some basic setup info in the `Commentary` section instead of referring readers to a separate README file to which they won't have access locally. Elisp should ideally be self-documenting! :-)

Hope that helps!

purcell:
I'd also suggest running `M-x checkdoc`, which will help you tidy up your docstrings.

JorisE:
Thanks a lot for the comments. The dependency issues have been fixed. I think the autoload issue has been fixed as well.

purcell:
Nice: merging, thanks!

PR #3558, opened 2016-02-04
Add recipe for cabledolphin

legoscia:
[Cabledolphin](https://github.com/legoscia/cabledolphin) is a tool for capturing network traffic from and to Emacs Lisp programs. I created it to be able to analyse encrypted traffic — with Wireshark, all you see is the encrypted data, but since Cabledolphin captures data on the Emacs Lisp level, it is able to see the cleartext.


purcell:
Looks great, thanks!

PR #3560, opened 2016-02-04
Add latex-unicode-math-mode

Christoph-D:
[latex-unicode-math-mode](https://github.com/Christoph-D/latex-unicode-math-mode) is a minor mode that allows the user to enter Unicode symbols in LaTeX math environments. For example, it replaces `\phi` with `φ` and `==` with `≡`.

Compared to the `TeX` input method it has more convenient keybindings and a .sty file to compile the resulting .tex file with pdflatex

### DIFF
--- a/recipes/latex-unicode-math-mode
+++ b/recipes/latex-unicode-math-mode
@@ -1,0 +1,4 @@
+(latex-unicode-math-mode
+ :repo "Christoph-D/latex-unicode-math-mode"
+ :fetcher github
+ :files (:defaults "*.sty"))


vyp:
I know it isn't required, but did you mean to release it without a license? I would have opened an issue on the repository, but have disabled issues for it.

Christoph-D:
No, I forgot.  I release it under the MIT license.

vyp:
:sparkles: :smile:

purcell:
Thanks. Please use `:defaults` instead of naming the single `.el` file in the `:files` list, ie. `:files (:defaults "*.sty")`. You can amend and force-push the recipe commit to cleanly update the PR. Other than that, looks good. I'd suggest adding `;;;###autoload` cookies for each of the minor mode definitions.

Christoph-D:
I changed it to `(:defaults "*.sty")` as you requested.  The autoload cookies are a good idea, I added those, too.

purcell:
Super, thanks.

PR #3561, opened 2016-02-05
add omtose-phellack-theme

franksn:
[Omtose-phellack](https://github.com/franksn/omtose-phellack-theme)  is just another dark bluish theme for emacs. For both the gui and the terminal (WIP). 

I am the author.

### DIFF
--- a/recipes/omtose-phellack-theme
+++ b/recipes/omtose-phellack-theme
@@ -1,0 +1,1 @@
+(omtose-phellack-theme :fetcher github :repo "franksn/omtose-phellack-theme")


purcell:
Thanks. I sent you a few fixes in https://github.com/franksn/omtose-phellack-theme/pull/1, but there's a more general issue with the theme which you should probably look into.

You might not know that the return value of `display-graphic-p` can vary from frame to frame, so it's not appropriate to check it just at the time the theme is loaded and then build the theme accordingly. And indeed, the theme/face system allows for this.

Ultimately, each face definition should have two clauses, one for non-graphic display, and the other for graphic displays. So instead of having `((t (... attributes ...)))` you would have something like

```
(((class color) (min-colors 257)) ... attrs ...))
 (t  ... attrs ...))
```

(I don't remember the syntax off-hand, but you can refer to the [relevant part of the manual](https://www.gnu.org/software/emacs/manual/html_node/elisp/Defining-Faces.html).)

Hope that helps!

franksn:
done sir, thanks a lot.

purcell:
Thanks. The result is no different, though, so maybe I wasn't explaining clearly enough.

If you have multiple variants of the faces for different displays, the way to handle it is to put all the variants in the face definitions, so that Emacs will choose the right ones according to the frame on which the face is later displayed.

So after loading the theme in an emacs instance, the correct colours will be displayed both in graphical and non-graphical frames connected to that instance. You'll be able to see that's not the case here by launching a graphical Emacs frame, loading the theme, and then connecting to the same Emacs in a terminal using `emacsclient -t`.

I also think the new packaging is not an improvement -- much better to keep it all in `omtose-phellack-theme.el` as before IMO.

franksn:
Ah i misunderstood.... okay got it, will do so

purcell:
Cool. I think you'll be happy with the result. P.S. a screenshot in your repo would be a nice touch. :-)

franksn:
Alright i think it's done, and i added a few scrots too :)

purcell:
Yep, great, and I'll go ahead and merge this now.

However, note that the `.pkg.el` is unnecessary and can/should be deleted from the repo (and it should have been `-pkg.el` anyway). In its current form it will just cause byte-compilation errors when people install the package, because it will get byte-compiled, not treated as a package descriptor.

Plus, I'd still strongly suggest just putting everything in `omtose-phellack-theme.el`, as it was previously -- much neater, since there's no reason for the current separation.

franksn:
Thanks, and well... I intend to make another variant of it (with bright colors and light background perhaps) thus the separation. :) As to that .pkg.el, i'll fix it right away

Thanks again.

PR #3562, opened 2016-02-05
Add recipe for company-shell.

Alexander-Miller:
https://github.com/Alexander-Miller/company-shell

Company-shell is a backend for company mode. It provides completions for binaries found on your $PATH and fish shell's functions, using man pages as doc strings. I am the author.

### DIFF
--- a/recipes/company-shell
+++ b/recipes/company-shell
@@ -1,0 +1,1 @@
+(company-shell :repo "Alexander-Miller/company-shell" :fetcher github :files ("company-shell.el"))


purcell:
Thanks. Few things I noticed here:

- The `Package-Requires` header is invalid currently, because it's missing the version number strings -- those are mandatory. Use the ones on GNU ELPA or MELPA Stable rather than the snapshot versions on MELPA. `flycheck-package` can help detect this sort of issue, btw.
- You might consider just using `exec-path` rather than splitting `$PATH` yourself. The two should match up in any well-configured Emacs, and the `exec-path-from-shell` helps users with this. (And `path-separator` is better than `":"` in any case.)
- Directories on the path might contain spaces, which I'm pretty sure would break the code which says `(format "%s --help" arg)`.
- If I have `fish` installed on my system, but it's not my `$SHELL`, I wouldn't want its functions to be included in my completions. Probably the fish completions should only take place in `fish-mode`.
- I think `string-empty-p` arrived only in Emacs 24.4, so you'll need to add `(emacs "24.4")` to the list of package dependencies.

Since the fish thing seems like a special case, I'm wondering if there should be separate `company-executables` and `company-fish-functions` packages, so that the two don't get muddled...? Seems like it would be cleaner. Fish users would then just install both.

Alexander-Miller:
@syohex was kind enough to supply a pull request for the `Package-Requires` and `string-empty-p` issues. They ought to be fixed now.

> You might consider just using exec-path

Will do.

> Directories on the path might contain spaces, which I'm pretty sure would break the code which says (format "%s --help" arg)

The `arg` variable used here will only ever be just the binary's name. A binary's path is only ever used to propertize it for company's annotation. I've just tried adding a directory a space to my PATH and the annotation was correct, so spaces shouldn't be an issue.

> I'm wondering if there should be separate company-executables and company-fish-functions packages

The package is barely 100 LOC as it is - splitting it in two would really be overkill, especially since I'd have to duplicate a lot of code for propertyzing, meta- and doc-strings and backend functions. Of course I could create some kind of common framework for both, but that, again, is complete overkill. You do have a point that users may sometimes want to exclude fish functions, and I think that feature would best be served by introducing a simple boolean that decides whether to also append fish's functions to the list of completions.

purcell:
Or have one package supply two separate company backends. Then company will append the fish functions to the completions list for you when the fish backend is enabled.

Alexander-Miller:
Sounds good. I'll try and get it done tomorrow.

Alexander-Miller:
Changes are up. `company-shell` and `company-fish-shell` are now strictly separated.

purcell:
Nice, seems a bit cleaner! Merging now.

P.S. `(when (null company-shell--cache) ...)` should probably just be `(unless company-shell--cache ...).

PR #3563, opened 2016-02-05
get jabber from github

tarsius:
Jabber is available from both Sourceforge and Github, and he HEAD is
identical in both repositories.  http://emacs-jabber.sourceforge.net
doesn't say which is the official repository so lets use Github.

@legoscia is this a good course of action?

### DIFF
--- a/recipes/jabber
+++ b/recipes/jabber
@@ -1,4 +1,4 @@
 (jabber
- :url "git://git.code.sf.net/p/emacs-jabber/git"
- :fetcher git
+ :fetcher github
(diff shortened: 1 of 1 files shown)


legoscia:
Either should work. In principle the Sourceforge repository has a higher bus factor, as the other maintainers have write access to it as well, but I keep the Github one up to date as well.

purcell:
Thanks @legoscia, I think we should keep it pointing at Sourceforge for now, then.

tarsius:
Thanks @legoscia.

PR #3564, opened 2016-02-05
Local declarative programming with Org elements.

tj64:
org-dp (https://github.com/tj64/org-dp) is meant for local programming,
i.e. for situations where the complete parse tree of an Org-mode file is
not needed, but the focus is on the Org element at point.

It leaves parsing and interpreting to the Org-mode parser framework and
creates, updates and deletes Org elements by acting on their uniform
internal (p-list) representations rather than on their diverse textual
representations.

I'm the library's only author and maintainer.

### DIFF
--- a/recipes/org-dp
+++ b/recipes/org-dp
@@ -1,0 +1,5 @@
+(org-dp
+ :fetcher github
+ :url "https://github.com/tj64/org-dp"
+ :repo "tj64/org-dp"
+ :files ("org-dp.el"))
--- a/recipes/org-dp
+++ b/recipes/org-dp
@@ -1,5 +1,1 @@
-(org-dp
- :fetcher github
- :url "https://github.com/tj64/org-dp"
- :repo "tj64/org-dp"
- :files ("org-dp.el" "org-dp-lib.el"))
+(org-dp :fetcher github :repo "tj64/org-dp")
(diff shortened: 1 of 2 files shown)


purcell:
Hi! Thanks for this.

I don't understand the packaging intent here -- why two packages? Feels like both files should be part of a single `org-dp` package. Otherwise you'll need to add a `Package-Requires` header to `org-dp-lib` so that it depends on `org-dp`.

In any case, you'll need a `Package-Requires` header in order to declare a dependency on `cl-lib` 0.5.

tj64:
Hi, I thought optional files should have their own recipe, but I agree, its much better to put both files in one recipe. I added the cl-lib dependency and fixed the related code (by merging a friendly pull request).
Cheers

purcell:
> I thought optional files should have their own recipe,

Yeah, my rule of thumb for that is that if the optional files have different dependencies, then package them separately. So if a library comes with a `helm` plugin, I'll usually suggest that the author should split it out into a separate package.

Can I suggest you make an `experimental` subdir in your repo for the non-distributable file, then the recipe can simply be:

```el
(org-dp :fetcher github :repo "tj64/org-dp")
```

since `*.el` are included by default.

tj64:
Done, thanks for the tip.

purcell:
Great, just committed the simplified recipe in a single commit. :-)

PR #3565, opened 2016-02-06
remove midje-test-mode

tarsius:
`midje-test-mode.el` requires `clojure-test-mode` which
was removed from `clojure-mode` more than a year ago.

https://github.com/clojure-emacs/clojure-mode/commit/c7f762976d4147d36b89133c5d9b93accfc9e898


purcell:
Yep, makes sense, thanks.

PR #3566, opened 2016-02-06
add dumb jump

jacktasia:
Hi,

Just trying to get dumb-jump merged again https://github.com/jacktasia/dumb-jump

The tests failed on the last attempt, which I think was due to an external error.


purcell:
I'll just reopen #3532, where all the comments are, otherwise things will get fragmented.

purcell:
(If I need to investigate build failures, I can easily do it there.)

PR #3568, opened 2016-02-06
Add recipe for replace pairs

davidshepherd7:
Adds functions to replace/query replace related pairs of things, e.g.

    Foo(bar[X])

to

    Foo(bar<X>)

with one command.

Package repo is here: https://github.com/davidshepherd7/replace-pairs

I'm the author.

### DIFF
--- a/recipes/replace-pairs
+++ b/recipes/replace-pairs
@@ -1,0 +1,2 @@
+(replace-pairs :repo "davidshepherd7/replace-pairs"
+               :fetcher github)


purcell:
Thanks, looks good. Think you could use some of the `electric-pair` machinery to tidy it up -- I think it already knows about things like `{}` vs `()` vs `do ... end`. You might also want to add `;;;###autoload` cookies for the main interactive commands a user might call.

davidshepherd7:
Good idea, looks like `electric-pair` does it by reading the syntax table though, so I'll have to figure out how that works first :)

Thanks, completely forgot about autoloads.

purcell:
Yep, syntax table should always know about the relevant pairs in the current buffer. Generally speaking, using regexps is going to give you all sorts of problems when dealing with text like `(")")` -- best to lean on the common emacs functions which lean on the syntax table etc.

PR #3569, opened 2016-02-06
move package-build.el to lisp/ subdirectory

tarsius:
This allows me to mirror the `package-build` package on the Emacsmirror
using a fetcher build around `git subtree`.  I would like to do that
because I already mirror several packages which depend on it.  I do not
mirror complete repositories when they contain much much more than just
the elisp, except when I can mirror just a subtree using `git subtree`.

Of course the history extracted using `git subtree` will lack the old
history, but that's better than not mirroring the package at all.

I have tested building packages locally, but if it is possible to test this before putting it into production, then that would probably be a good idea. At the very least you have to review this carefully, as my understanding of any of the scripts and makefiles that make use of this library is very limited.

Thanks for considering!

Ps: The "re-announcement" of the Emacsmirror can happen any time now! :-)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,7 @@ ifeq ($(NEED_CL-LIB), t)
 	EMACS_COMMAND := $(EMACS_COMMAND) --eval "(package-initialize)"
 endif
 
-EVAL := $(EMACS_COMMAND) --no-site-file --batch -l package-build.el --eval
+EVAL := $(EMACS_COMMAND) --no-site-file --batch -l lisp/package-build.el --eval
 
 TIMEOUT := $(shell which timeout && echo "-k 60 600")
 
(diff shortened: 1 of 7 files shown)


vyp:
May I take this opportunity to ask, (@milkypostman, @purcell) what about putting package-build in a separate repository altogether? My logic is, one repository for recipes, and another one for package-build library. Either way it's your preference, just suggesting.

tarsius:
Actually I would like that even better and volunteer to do the `git filter-branch` to preserve the history.

I noticed the travis error and will look into that.

milkypostman:
I think we should move this to the MELPA organization while we're doing
this. I propose the following order,

. Move to MELPA org
. Pull out package-build to its own repo.

For the latter I don't think there is any reason to filter history. The
second repo can be a branch of the original.

purcell:
+1. Though I think if it's not too hard, it'd be worth starting the new repo afresh with `filter-branch`. A fork would carry the full history of everything.

tarsius:
The history can be extracted using:

```
$ git branch package-build master
$ git filter-branch --index-filter '
git rm --cached "*" > /dev/null
git reset $GIT_COMMIT -- package-build.el > /dev/null
' --commit-filter '
test $# = 1 && test -z "$(git ls-tree $1)" && skip_commit "$1" && exit
args="$@"
tree="$1"
shift
while test -n "$1"
do
    shift
    test "$tree" = $(git rev-parse "$1^{tree}") && map "$1" && exit
    shift
done
git commit-tree $args
' package-build
```

I used my own `--commit-filter` because `--prune-empty` fails miserably at pruning empty commits.

I can upload that somewhere, but you might just as well run these commands again. I got this tip:

```
$ git log -1 package-build
commit 2e8a15b0ad52fe9771ba7243aca838d6774858ed
Author: Steve Purcell <steve@sanityinc.com>
Date:   Fri Jan 15 21:24:34 2016 +1300

    Fix indentation
```

As a sanity check I ran the following commands. As expected they did not produce any output.

```
$ git log --format=%s master -- package-build.el > a
$ git log --format=%s package-build > b           
$ diff -u a b
```

tarsius:
Superseded by  #3570.

PR #3571, opened 2016-02-08
remove save-sexp

tarsius:
I stopped maintaining this package a long
time ago but forgot to remove it from Melpa.

### DIFF
--- a/recipes/save-sexp
+++ b/recipes/save-sexp
@@ -1,1 +1,0 @@
-(save-sexp :repo "tarsius/save-sexp" :fetcher github)


purcell:
Might it not have a few users? 144 downloads probably doesn't correspond to more than a few experimental installations, but I'm wondering if there's any harm in keeping such things. If we removed all unmaintained packages from MELPA I'm not sure the result would be better for users.

tarsius:
> If we removed all unmaintained packages from MELPA I'm not sure the result would be better for users.

I think the result would be worse if we did that. But that's not what I am asking for, I want this one package removed. As you surely have noticed, I am currently and once more working on fixing errors in the Emacsmirror and Melpa by comparing the two sets of packages. There are huge lists of issues but they are quickly getting smaller. I do not have the illusion that I can ever get to "issue zero", but I would rather not have one of my own packages appear in the lists of inconsistencies. That would just undermine my motivation to work on the more serious issues.

tarsius:
Oh, and if someone does complain, then I would move it to the orphanage, and mirror from there again and also add to Melpa again.

purcell:
:+1:

PR #3574, opened 2016-02-08
Fix upstreams of various packages

tarsius:
I checked that the new upstreams are more up-to-date than the old.

While doing so I also ran into some packages were it is not obvious whether we should stick with the old upstream or not.

* `desktop-registry` and `mode-icons` by @ryuslash are available from github as well from repositories on his own server. All repositories appear to be up-to-date. **Edit:** changed recipes as per reply below.

* `jabber` by @legoscia are available from github as well as from sourceforge. The homepage on sourceforge only mentions the repository on sourceforge. Both repositories appear to be up-to-date. **Edit**: use sourceforge as per #3563.

I would prefer getting the packages from the github repositories, as that would allow users to click on a link on Melpa's webpage to go to the web view of the repository. (And also because that's already what I do on the Emacsmirror.) @ryuslash @legoscia what do you think?

* `goto-chg` was modified [on github]( https://github.com/martinp26/goto-chg/commit/2440191e1ebe14a6fac5278bd9dcb2fd0be9ff59) by @martinp26 and on the wiki by David Andersson, the original author. I have gone back to mirroring the version from the wiki. @martinp26 I suggest you contact David and suggest that he merges your changes and makes the result available as a git or hg repository.

### DIFF
--- a/recipes/desktop-registry
+++ b/recipes/desktop-registry
@@ -1,1 +1,1 @@
-(desktop-registry :fetcher git :url "git://ryuslash.org/desktop-registry.git")
+(desktop-registry :fetcher github :repo "ryuslash/desktop-registry")
--- a/recipes/dummy-h-mode
+++ b/recipes/dummy-h-mode
@@ -1,1 +1,1 @@
-(dummy-h-mode :fetcher wiki)
+(dummy-h-mode :fetcher github :repo "yascentur/dummy-h-mode-el")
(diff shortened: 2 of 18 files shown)


ryuslash:
Hi,

> * `desktop-registry` and `mode-icons` by @ryuslash are available from github as well from repositories on his own server. All repositories appear to be up-to-date.

I've been meaning to ask to change the location of my packages, so feel free to change their upstreams to my github projects, or let me know that I should.

tarsius:
Thanks @ryuslash.

purcell:
Merged now, thanks.

tarsius:
Thanks for the review!

PR #3575, opened 2016-02-08
update :repo of many github packages

tarsius:
Before making these changes I did double check using the Github Api.

### DIFF
--- a/recipes/4clojure
+++ b/recipes/4clojure
@@ -1,1 +1,1 @@
-(4clojure :fetcher github :repo "joshuarh/4clojure.el")
+(4clojure :fetcher github :repo "losingkeys/4clojure.el")
--- a/recipes/ac-math
+++ b/recipes/ac-math
@@ -1,1 +1,1 @@
-(ac-math :repo "vitoshka/ac-math" :fetcher github :files ("ac-math.el"))
+(ac-math :repo "vspinu/ac-math" :fetcher github :files ("ac-math.el"))
--- a/recipes/angry-police-captain
+++ b/recipes/angry-police-captain
@@ -1,2 +1,2 @@
 (angry-police-captain :fetcher github
-                      :repo "rolando2424/angry-police-captain-el")
+                      :repo "rolpereira/angry-police-captain-el")
--- a/recipes/ansi
+++ b/recipes/ansi
@@ -1,2 +1,2 @@
-(ansi :repo "rejeep/ansi" :fetcher github)
+(ansi :repo "rejeep/ansi.el" :fetcher github)
 
--- a/recipes/ant
+++ b/recipes/ant
@@ -1,1 +1,1 @@
-(ant :fetcher github :repo "apgwoz/ant-el")
+(ant :fetcher github :repo "apg/ant-el")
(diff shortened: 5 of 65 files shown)


recipes/editorconfig-core
```diff
@@ -1,3 +1,3 @@
-(editorconfig-core :repo "editorconfig/editorconfig-emacs"
+(editorconfig-core :repo "10sr/editorconfig-core-emacslisp"
```
10sr: I moved `editorconfig-core` to editorconfig/editorconfig-emacs so no need to change this :innocent: 
`10sr/editorconfig-core-emacslisp` is not maintained any more.

recipes/editorconfig-fnmatch
```diff
@@ -1,3 +1,3 @@
-(editorconfig-fnmatch :repo "editorconfig/editorconfig-emacs"
+(editorconfig-fnmatch :repo "10sr/editorconfig-fnmatch-el"
```
10sr: Same as :point_up:

recipes/editorconfig-core
tarsius: Thanks. I have reverted these changes and will fix this on the Emacsmirror instead.

tarsius:
I have now reviewed all cases where both the repository tracked by the Emacsmirror and the one tracked by Melpa still exist. In most cases it actually was Melpa which was correct. I have dropped the changes which "fixed" the respective recipes and will instead fix the Emacsmirror.

tarsius:
The following packages are still available from both places and I think the one tracked by the Emacsmirror is the correct one:

* `google` by @hober was forked by @steckerhalter and @hober then merged the additional commit.
* `plantuml-mode` by @zwz was forked by @wildsould who then made some changes in 2013. In 2014 and 2015 @zwz made changes too. History diverged.
* `ponylang-mode` @purcell's repo was forked by @SeanTAllen, which was then forked by @abingham 
* `qml-mode` by @xwl was forked by @coldnew. Both repos are at the same commit.

Undecided:

* `point-stack` by @mattharrison was forked by @dgutov. History diverged.
* `pomodoro` two independent implementations by @baudtack and @vderyagin. @baudtack's version was updated more recently.
* `ponylang-mode` @purcell's repo was forked by @SeanTAllen, which was then forked by @abingham. The latter two are at the same commit, the former lags behind 40 commits.

SeanTAllen:
@tarsius I'm now the official maintainer of ponylang-mode. My repo should be considered the canonical one.

tarsius:
Thanks @SeanTAllen!

dgutov:
> point-stack by @mattharrison was forked by @dgutov. History diverge

 I'm really not sure which fork should be used. My fork is not just a cleanup or upgrade. It implements different, though arguably better, logic for storing the stack of positions.

purcell:
Thanks for this @tarsius. Huge job to review it all though, sorry, will take me a while. So far I've noticed that `qml-mode` and `plantuml-mode` changes feel a little spurious and undesirable.

purcell:
Applied, with the exception of the qml-mode change. Thanks!

tarsius:
Thanks for the review!

The Emacsmirror now gets `qml-mode` from the same repository as Melpa.

PR #3576, opened 2016-02-09
Add recipe for `helm-describe-modes`.

xiongtx:
This is a package to replace `describe-mode` with a Helm interface, actions, and more.

### DIFF
--- a/recipes/helm-describe-modes
+++ b/recipes/helm-describe-modes
@@ -1,0 +1,2 @@
+(helm-describe-modes :fetcher github
+		     :repo "emacs-helm/helm-describe-modes")


purcell:
Thanks. The `helm-def-source--*` macros should be renamed so that they have a namespace prefix consistent with the rest of the package. Looks good otherwise.

xiongtx:
Done. Now I know why people name their packages with 3-7 letter names!

purcell:
:-)  Yep, the [elisp manual says](https://www.gnu.org/software/emacs/manual/html_node/elisp/Coding-Conventions.html):

```
You should choose a short word to distinguish your program from other Lisp programs. The names of all global symbols in your program, that is the names of variables, constants, and functions, should begin with that chosen prefix.
```

There's also the "nameless" package, which hides long name prefixes. :-)

PR #3582, opened 2016-02-10
Add auto-virtualenv

marcwebbie:
Automatically activate python virtualenvs on emacs based on project name or `.python-version` file.

repo: https://github.com/marcwebbie/auto-virtualenv

### DIFF
--- a/recipes/auto-virtualenv
+++ b/recipes/auto-virtualenv
@@ -1,0 +1,1 @@
+(auto-virtualenv :fetcher github :repo "marcwebbie/auto-virtualenv")


purcell:
Thanks. I've merged this, but I'd say the code is trying to be too clever, and in particular, hardcoding all those possible "project root" indicator filenames isn't a good plan. If someone adds a `Makefile` in a subdir, why would they expect that to break `auto-virtualenv`? Consider simply using `vc-root` or similar, or use `projectile-root` if (and only if) it's `fboundp`. The approach taken by [pyenv-auto-mode](https://github.com/ssbb/pyenv-mode-auto/blob/master/pyenv-mode-auto.el), for example, is much more predictable, even if simplistic.

PR #3584, opened 2016-02-10
added ess-view recipes

GioBo:
Dear MELPA Maintainers,

  ess-view is a simple package which lets users of ess (Emacs Speaks Statistics) inspect R dataframes through an external spreadsheet software (this is useful for medium/big-size objects which do not display well inside the terminal): with a simple keybinding users can digit the name of the dataframe and ess-view will take care of converting it to a temporary csv file and open it via a spreadsheet; there's also the chance of saving any change done to the spreadsheet file and load back the results into the original R dataframe.

The original repository of ess-view is found at https://github.com/GioBo/ess-view.

### DIFF
--- a/recipes/ess-view
+++ b/recipes/ess-view
@@ -1,0 +1,1 @@
+(ess-view :fetcher github :repo "GioBo/ess-view")


purcell:
Thanks. A few issues here:

- You have a package dependency on `ess`, but not on the other packages you're using, i.e. `s`, `f`.
- The formatting is very random and inconsistent here, with odd whitespace and `(` / `)` on separate lines.
- Don't use `_` in function names -- use `-`: this is lisp! :-)
- As noted in the [elisp conventions](https://www.gnu.org/s/emacs/manual/html_node/elisp/Coding-Conventions.html), you need to prefix all of your top-level symbols with the package name, ie. `ess-view-random-string`, not `random_string`.
- Where you build up commands containing file names, use `shell-quote-argument` so that spaces or other characters in the file names will not break the command (or system!)
- Run `M-x checkdoc` and fix its suggestions, so that the docstrings are complete and correctly formatted.
- "A small package" is not a helpful part of the package description, and you should probably remove it
- Consider adding a more complete `Commentary` section so that users have more information about what they can expect from the package, and how to use it

Hope that helps!

GioBo:
Dear Steve, 

Thank you very much for your detailed review of the package (this is my first attempt at writing an elisp package and I apologize for all the mistakes that were present in the code).
I've just pushed a new version to [GioBo/ess-view](https://github.com/GioBo/ess-view) which includes the following changes:

* I added the needed dependencies in `Package-Requires` (I searched for the the version number of `ess-site` and `ess-inf` but I was not able to find them in their source codes),

* I've used [lispy](https://github.com/abo-abo/lispy/blob/master/lispy.el) to format the code (I was not aware of this package until a few hours ago),

* I converted ` _ ` to `-` (yes...it's more 'lispy' now),

*  I added the `ess-view-` prefix to all the top-level symbols,

* the only file name used in a string-command is being created by `make-temp-file` thus I think `shell-quote-argument` is not needed in this case,

* I checked all doctrings using `checkdoc`,

* I agree...and changed the short package description,

* I added a more complete description of the package and of its usage in the `Commentary`.


Thank you again for your comments and the time you spent checking the code.

GioBo:
Sorry..having a look at other pull request I've just realized that the version number for dependencies is mandatory...I'll try to fix this as soon as possible.

G.

GioBo:
I fixed the issue with version number of dependencies and I modified the package removing some redundancies.

purcell:
Cool, looks good, except that `ess-inf` and `ess-site` are not individually-installable packages, so they should not be listed in `Package-Requires`: they are just files contained in the `ess` package.

GioBo:
Whoops..evidently I was confusing "features" and "packages". Thanks again for the help provided.
G.

purcell:
Merged, thanks.

PR #3585, opened 2016-02-11
add recipe ob-go

stardiviner:


purcell:
Thanks. You should probably drop the package dependency on go-mode -- it's not strictly needed or even used by the code, so it's arguably just a recommended add-on.

pope:
The dependency has been dropped.

PR #3586, opened 2016-02-11
Add select-themes recipe

jasonm23:
Simple interactive theme selector that disables other custom themes before loading the selected theme or "-- Default --"

### DIFF
--- a/recipes/select-themes
+++ b/recipes/select-themes
@@ -1,0 +1,3 @@
+(select-themes
+    :fetcher github
+    :repo "jasonm23/emacs-select-themes")


syohex:
You need not to specify `:files` attribute. It is enough to use default pattern. Default pattern is as below. Please see `:files` section of [here](https://github.com/melpa/melpa#recipe-format) in more detail.

```
("*.el" "*.el.in" "dir"
 "*.info" "*.texi" "*.texinfo"
```

jasonm23:
Sure, but it's not a problem to include it.  See `use-package`

syohex:
However you need to update MELPA recipe if you add some files in that repository in the future. (While you need not to update MELPA recipe if files you add are matched against default pattern.)

jasonm23:
That's ok, I know how it works.  Please note, this package won't have additional files.

In the unlikely occasion that it does, I will update or remove the `:files` keyword as appropriate.

Thanks for raising the flag anyway, someone else may not know how it works.

purcell:
Yep, when the default is adequate, we would prefer that it is not overridden. I'll make that clearer in the README.

jasonm23:
I'm only reluctant to fix it now because I'm away from a machine for 24-48+ hours, phone only.

I can probably ssh onto something and do it if it's a requiremebtZ

jasonm23:
Done via the magic of tiny fonts and unix shells :)

jasonm23:
Travis errors don't seem to be related to **select-themes**

jasonm23:
closing and reopening to force travis to rerun

PR #3589, opened 2016-02-12
Update move-text

jasonm23:
Don't pull from the emacswiki - new repo added

### DIFF
--- a/recipes/move-text
+++ b/recipes/move-text
@@ -1,1 +1,1 @@
-(move-text :fetcher wiki)
+(move-text :fetcher github :repo "emacsfodder/move-text")


purcell:
Yay, thanks.

Wilfred:
Argh, I patched the version on the wiki just after this changed. :)

jasonm23:
You're the reason why this had to happen.  Your fix broke some people and there was no oversight.

jasonm23:
We will need to nail down which versions broke

jasonm23:
Would you open an issue on the repo itself, and we will get it fixed up :+1:

jasonm23:
(In the meantime we can make the old version stable and the latest on wiki, edge.)

Wilfred:
Sorry about that. I've opened https://github.com/emacsfodder/move-text/issues/2 so we can fix this properly.

PR #3593, opened 2016-02-12
use github fetcher for markdown-mode recipe

vyp:
Ref <https://github.com/jrblevin/markdown-mode/issues/84>.

---

Assuming @jrblevin is okay with this of course.


jrblevin:
Thanks, @vyp. Yes, this is fine with me.

purcell:
super, thanks

PR #3594, opened 2016-02-13
release nix-sandbox.el on melpa

svenkeidel:
Hello,

I wrote an emacs package that allows to work with sandboxes in NixOS. I want to release it on MELPA because people asked for it. I used this package every day on work for the last few months and it works. @purcell, let me know what you think.

Best,
Sven

### DIFF
--- a/recipes/nix-sandbox
+++ b/recipes/nix-sandbox
@@ -1,0 +1,3 @@
+(nix-sandbox :fetcher github
+             :repo "travisbhartwell/nix-emacs"
+             :files ("nix-sandbox.el"))


purcell:
Thanks. Few things I noticed:

- Maybe set a non-nil parent group for the `nix` custom group: this way you can place it logically in the overall customisation hierarchy.
- Can you guarantee `sandbox` won't have spaces? If not, building a command with `concat` is going to bite you one day. `shell-quote-argument` would be a useful and safe precaution there.
- Autoloading `nix-exec-path-map` feels like it should be unnecessary, particularly since it's not a documented var for end-user customisation
- Please don't define unprefixed top-level symbols like `map-nil`
- Consider `M-x checkdoc`, and fixing the recommendations. Your users will thank you!

Hope that helps. :-)

svenkeidel:
Thank you very much for the code review @purcell. I will improve the code and submit again.

svenkeidel:
I incorporated the suggested improvements in https://github.com/travisbhartwell/nix-emacs/commit/ebd66d5b894c2f27b694cdbc851320556f3fbc95

@purcell, please let me know if there is anything else I can improve.

purcell:
Looks good, thanks -- merging now!

PR #3595, opened 2016-02-14
Add recipe for sly-repl-ansi-color

noctuid:
[sly-repl-ansi-color](https://github.com/PuercoPop/sly-repl-ansi-color) adds ansi color support to the sly mrepl. I've now tested it successfully in the sandbox.

### DIFF
--- a/recipes/sly-repl-ansi-color
+++ b/recipes/sly-repl-ansi-color
@@ -1,0 +1,3 @@
+(sly-repl-ansi-color
+ :fetcher github
+ :repo "PuercoPop/sly-repl-ansi-color")


purcell:
Thanks. Please also declare a dependency on `(cl-lib "0.5")`.

syohex:
`sly` requires emacs-24.3 or higher versions. cl-lib.el was bundled since Emacs 24.3.

purcell:
True, but I think it's good practice to not rely on implicit transitive dependencies.

noctuid:
@syohex Thank you! Sorry, the dependency on cl-lib slipped by me.

PR #3596, opened 2016-02-14
Add helm-purpose-recipe

bmag:
Summary: helm-purpose provides some [helm](https://github.com/emacs-helm/helm/) commands and sources for [window-purpose](https://github.com/bmag/emacs-purpose).
Link: https://github.com/bmag/helm-purpose/
Association: I'm the author of helm-purpose and window-purpose

The `helm-purpose.el` file is small, but I didn't want to add it to window-purpose because I didn't want to force a helm dependency on non-helm users. Since there are already lots of `helm-*` packages, I assume it's ok.

Thanks for your work on melpa

### DIFF
--- a/recipes/helm-purpose
+++ b/recipes/helm-purpose
@@ -1,0 +1,1 @@
+(helm-purpose :repo "bmag/helm-purpose" :fetcher github)


purcell:
Thanks, merged, but I'm wondering if the following autoload causes `window-purpose` to eagerly loaded when the autoload files are evaluated at Emacs startup time, because that wouldn't be ideal:

```el
;;;###autoload
(defalias 'helm-purpose-mini-ignore-purpose
  (without-purpose-command #'helm-mini)
  "Same as `helm-mini', but disable window-purpose while this command executes.")
```

bmag:
Thanks, I'll check it

PR #3597, opened 2016-02-14
add org-clock-convenience

dfeich:
This package introduces a number of convenience functions for time
tracking in Org-mode. Allows modifying clocked times directly from the
log lines in the agenda buffer.

Thanks,
Derek

### DIFF
--- a/recipes/org-clock-convenience
+++ b/recipes/org-clock-convenience
@@ -1,0 +1,3 @@
+(org-clock-convenience
+ :fetcher github
+ :repo "dfeich/org-clock-convenience")


purcell:
Thanks. I'd strongly suggest just using `org-clock-convenience-` for the symbol prefix in this code to match the containing package name. The extra bytes are essentially free these days!

You probably also need to declare dependencies on `cl-lib` and `org` via `;; Package-Requires: ((cl-lib "0.5") (org "8"))`, and/or maybe even something like `(emacs "24.3")`.

dfeich:
Thanks for the tips. I had fought a bit with myself, originally when I started the coding, to choose such a long prefix org-clock-convenience-* since it fills the lines awfully fast. But I completely agree that the clarity and consistency are much improved. So, I adapted the prefix and I also added the Package-Requirements.

Cheers,
Derek

purcell:
Super, looks great!

PR #3598, opened 2016-02-14
Add recipe for silkworm-theme.

mswift42:


purcell:
Looks good, but maybe there's scope for a better package description than "Theme"? ;-)

mswift42:
You are right, just the word 'Theme' does not do  justice to the sheer brilliance and magnificence of this work of art

Thanks for noticing :)

purcell:
:laughing:

PR #3599, opened 2016-02-14
VHDL-capf package added.

sh-ow:
This package provides a completion-at-point-function for vhdl-mode. This shall improve auto-completion in vhdl-mode because with the help of this function any completion framework can be used, like company-mode (with all its benefits like fuzzy matching auto-completion).
The package's repo is the following: https://github.com/sh-ow/vhdl-capf

I (sh-ow) am the creator and maintainer of this package.

### DIFF
--- a/recipes/vhdl-capf
+++ b/recipes/vhdl-capf
@@ -1,0 +1,1 @@
+(vhdl-capf :fetcher github :repo "sh-ow/vhdl-capf")


purcell:
Thanks. Sent you a quick fix, but you should also rename all top-level symbols to ensure they have the prefix `vhdl-capf-`, as per the [elisp coding conventions](https://www.gnu.org/software/emacs/manual/html_node/elisp/Coding-Conventions.html).

Note also that `(if (cond ...) (form) nil)` should just be `(when (cond) (form))`. Similarly, `(when (eq didchanges nil) ...)` should be `(unless didchanges ...)` etc., and `(if (null x) nil y)` is `(when x y)`. Some of those stylistic changes would definitely aid readability.

And you might find `M-x checkdoc` helpful for making sure your docstrings are well-formed.

Hope that helps!

sh-ow:
Thanks for your feedback, Steve!

I've fixed this readability issues, hopefully I haven't overseen any.

purcell:
Great, thanks, but you still need to prefix `flatten` etc. with `vhdl-capf-`.

sh-ow:
I'm sorry for this inadvertence. Should be fixed now...

purcell:
Nice, thanks.

PR #3600, opened 2016-02-14
Add recipe for dark-mint-theme

shaunvxc:
I've created this theme over time and think that it is at the point where others might like to use it too. 

I am the primary maintainer, the the repo is here: https://github.com/shaunvxc/dark-mint-theme

### DIFF
--- a/recipes/dark-mint-theme
+++ b/recipes/dark-mint-theme
@@ -1,0 +1,1 @@
+(dark-mint-theme :repo "shaunvxc/dark-mint-theme" :fetcher github)


purcell:
Thanks. Note that your commented-out code for setting frame transparency is not at all the right way to handle things within the theme system. Happy to discuss why if that's something you want to bring back. :-)

shaunvxc:
Happy with no transparency for now, but definitely interested in learning the correct way to handle that

purcell:
Essentially, I think there _is_ no correct way to handle transparency via theming! You can use `custom-theme-set-variables` to specify how a theme should affect variables at the point it is enabled, but modifying `default-frame-alist` via that mechanism that would override any other frame params.

The key issue with your previous/current approach is that it will only take effect when the theme is first loaded, but not each time it is _applied_: the transparency code is therefore only loosely associated with the theme, which isn't what you want.

P.S. There's a separate [seethru package](http://melpa.org/#/seethru) which might interest you, since it provides easy commands for adjusting the current frame transparency.

shaunvxc:
Thanks so much this was quite helpful!

PR #3601, opened 2016-02-15
add recipe emacsist-view

lujun9972:


purcell:
Thanks. Merging, but I also sent you a fix in https://github.com/lujun9972/emacsist-view/pull/2.

PR #3602, opened 2016-02-16
Refactoring :files attribute of syohex's packages

syohex:
- Remove needless :files attribute
- Use :defaults as possible


purcell:
Super, thanks.

PR #3603, opened 2016-02-16
Add cheerilee recipe

Vannil:
`cheerilee` is a package to create simple graphical applications in ELisp.
It provides functions to create visual elements, and some basic interactions.
I'm the author of the package, which can be found here: https://github.com/Vannil/cheerilee.el

Thank you for your time.


purcell:
Thanks!

PR #3604, opened 2016-02-16
Fix MELPA Stable link

jschwab:
The URL and description were switched.


purcell:
Thanks for catching that1

purcell:
Thought I'd clicked "merge" at the time... doh!

PR #3608, opened 2016-02-18
Add recipe for click-mode

bmalehorn:
`click-mode` is a major mode for editting configuration files for the Click Modular Router Project ([homepage](http://read.cs.ucla.edu/click/click)). I'm the author of the package, which you can get at https://github.com/bmalehorn/click-mode.

I've tried installing it in a sandboxed Emacs and it works fine. If you need to test it, open up [bmalehorn/click-mode/example.click](https://raw.githubusercontent.com/bmalehorn/click-mode/master/example.click) and verify it highlights properly. It should look something like this:

Screenshot

Let me know if there's anything I need to change to get it merged in. It's not worth writing a package unless it gets into melpa!

### DIFF
--- a/recipes/click-mode
+++ b/recipes/click-mode
@@ -1,0 +1,1 @@
+(click-mode :fetcher github :repo "bmalehorn/click-mode")


purcell:
Thanks. Sent you some minor fixes in https://github.com/bmalehorn/click-mode/pull/1. You might also want to consider using `prog-mode` rather than `fundamental-mode` as the parent mode -- it would be a no-brainer, but would require Emacs 24, so you'd also want to add a `;; Package-Requires: ((emacs "24"))` header if you do that.

bmalehorn:
Alright, I've merged your recommended changes. I also [inherit from `prog-mode` now](https://github.com/bmalehorn/click-mode/commit/a36292659efe3b1e9b5ae6f8cffcb1e3bf7b52fb); Emacs is >= 24 everywhere I use it.

purcell:
Yay, thanks.

PR #3611, opened 2016-02-20
Ensure SCM tools are available under travis

### LINKED ISSUE #3592: Ensure all SCM tools are available under travis
See build failure in #3590 due to `darcs` not being installed there.

marsam:
Fixes #3592 
Also I took the liberty from evm to flycheck's [emacs-travis](https://github.com/flycheck/emacs-travis/) which is more easy to use

I know which packages are necessary because long ago [I used travis-ci to build melpa](https://github.com/emacs-pe/melpa/blob/master/.travis.yml)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,11 @@
 language: ruby
-
+# XXX: Requires sudo because travis-ci apt-package-whitelist doesn't contain the
+# following packages: cvs darcs fossil texinfo
+sudo: required
 before_install:
+  - sudo apt-get -yq update
+  - sudo apt-get -yq install bzr git mercurial subversion cvs darcs fossil texinfo
+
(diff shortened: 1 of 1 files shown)


purcell:
Great, thanks. What do you make of the build error though? (I'm going to re-start it in case it was a transient problem.)

marsam:
Seems to a problem with the CFLAGS used in emacs-travis, I'm filling an issue

purcell:
Cool, thanks.

purcell:
Given there's no activity on https://github.com/flycheck/emacs-travis/issues/1, would it make sense to cut this back to just installing the extra packages, then file a separate PR later to switch to `emacs-travis`?

marsam:
I've updated the pull request to only cover the installation of packages.

purcell:
Yay, thanks.

PR #3612, opened 2016-02-20
Add eink-theme recipe

maio:
E Ink color theme

Recipe for https://github.com/maio/eink-emacs

I'm maintainer.

### DIFF
--- a/recipes/eink-theme
+++ b/recipes/eink-theme
@@ -1,0 +1,1 @@
+(eink-theme :fetcher github :repo "maio/eink-emacs")


purcell:
The build would only have told you about the name in recipe, so it's not even as useful as me yet, to be honest. :-)

maio:
Ok, it seems that I fixed everything you mentioned. And now it installs correctly (well at least on my machine). :)

purcell:
Cool, thanks - merging now. Sent you a further little fix in https://github.com/maio/eink-emacs/pull/1.

PR #3613, opened 2016-02-21
Add slime-docker recipe.

daewok:
This package provides an easy way to launch Lisp processes in Docker containers and then connect to them using SLIME.

I created this package and it is hosted at https://github.com/daewok/slime-docker


purcell:
Thanks. You might like to add an `;;;###autoload` cookie for the `slime-docker` command.

daewok:
Thanks, I'll do that!

PR #3614, opened 2016-02-21
Add hexo.el

kuanyui:
[hexo.el](https://github.com/kuanyui/hexo.el) is a frontend of [hexo](https://github.com/hexojs/hexo)  for Emacs.

### DIFF
--- a/recipes/hexo
+++ b/recipes/hexo
@@ -1,0 +1,1 @@
+(hexo :fetcher github :repo "kuanyui/hexo.el")


purcell:
Thanks. Some quick feedback about the code:

- Don't use `:` or `/` in function names: just use `-`
- Consider what might happen if commands fail, e.g. instead of `hexo clean;hexo generate;hexo deploy` you might consider `hexo clean && hexo generate && hexo deploy`
- All `interactive` commands should have docstrings. Try `M-x checkdoc`.
- Rather than `(if cond (progn ...))`, just use `(when cond ...)`. Similarly `(if cond ... nil)` is the same as `(when cond ...)`.
- Don't ever use `cd`: instead, bind `default-directory` while running commands: this has the same effect.
- It's weird to constantly be checking whether commands have been executed in a buffer which is in `hexo-mode` -- seems like unnecessary boilerplate: just bind those commands in that mode's keymap, and don't worry.
- `hexo-find-root-dir` can probably just be simplified to a call to the built-in `locate-dominating-file` function, which I think does what you want here (and the calls to `file-truename` are unusual, and therefore probably unnecessary)

Hope that helps!

kuanyui:
Thanks for your code reviewing :bow: 

>    Don't use `:` or `/` in function names: just use `-`

That's on purpose. I use different separators to hint user. Commands with `/` cannot be called outside `hexo-mode`. When M-x, they will be mixed in the same completion list and hard to read.

 >   Consider what might happen if commands fail, e.g. instead of hexo clean;hexo generate;hexo deploy you might consider hexo clean && hexo generate && hexo deploy

Got it, thank you!

>    All interactive commands should have docstrings. Try M-x checkdoc.

Thanks, I'm working on this now.

>    Rather than (if cond (progn ...)), just use (when cond ...). Similarly (if cond ... nil) is the same as (when cond ...).
>    Don't ever use cd: instead, bind default-directory while running commands: this has the same effect.

I use cd on purpose, which make user can use `C-x C-f` (or Magit ...etc) in `hexo-mode` with the repo as `default-directory`

Another thing, I tried to get rid of an unnecessary `cd` in macro `hexo-repo-only` like this:

```lisp
(defmacro hexo-repo-only (&rest body)
  `(let ((dir (or hexo-root-dir (hexo-find-root-dir))))
     (if dir
         (let ((default-directory dir)) ,@body)
       (message "Please run his command under a Hexo repo directory."))))
```

But 

```lisp
(hexo-repo-only
      (hexo-start-process-shell-command "hexo clean ; hexo generate ; hexo server --debug --drafts;pwd"))
```
`pwd` becomes wrong(not under the correct repo path). Any ideas about this? 

>    It's weird to constantly be checking whether commands have been executed in a buffer which is in hexo-mode -- seems like unnecessary boilerplate: just bind those commands in that mode's keymap, and don't worry.

No,  because myself is exactly use M-x to execute the commands like `hexo-move-article`, `hexo-new`, I don't want the commands which can only be used in `hexo-mode` pollute my M-x commands completion candidates list.

Or maybe I rename the hexo-mode-only commands with double dashes **`hexo--`** prefix?

>    `hexo-find-root-dir` can probably just be simplified to a call to the built-in `locate-dominating-file` function, which I think does what you want here (and the calls to `file-truename` are unusual, and therefore probably unnecessary)


Thanks, I didn't know `locate-dominating-file` before. I'm rewriting this function now.
I use `file-truename` as the row ID in `tabulated-list`. `expand-file-name` is probably enough...?

purcell:
> That's on purpose. I use different separators to hint user. Commands with / cannot be called outside hexo-mode. When M-x, they will be mixed in the same completion list and hard to read.

I'm not sure this is obvious or idiomatic, or a good reason to stray from the official elisp conventions. If every package introduces its own conventions based on the author's personal tastes, that's not beneficial to users. :-)

> I use cd on purpose, which make user can use C-x C-f (or Magit ...etc) in hexo-mode with the repo as default-directory

Oh, okay, if that's only happening in `hexo-mode` then no problem.

> No, because myself is exactly use M-x to execute the commands like hexo-move-article, hexo-new, I don't want the commands which can only be used in hexo-mode pollute my M-x commands completion candidates list.

No other package does this, so this is a very unusual thing to do. Even if you think it's unreasonable, I'd discourage this.

> Or maybe I rename the hexo-mode-only commands with double dashes hexo-- prefix.

The `hexo--` prefix would indicate a package-private symbol.

There's a bunch of stuff in the official [elisp coding conventions](https://www.gnu.org/software/emacs/manual/html_node/elisp/Coding-Conventions.html) which relate to this, and I'd suggest following them whenever possible because that's better for the community.

> .
> I use file-truename as the row ID in tabulated-list. expand-file-name is probably enough...?

Yes, I think so.

kuanyui:
> I'm not sure this is obvious or idiomatic, or a good reason to stray from the official elisp conventions. If every package introduces its own conventions based on the author's personal tastes, that's not beneficial to users. :-)
> The hexo-- prefix would indicate a package-private symbol.

Thanks for your patience&explanations :bow: , I got it. Now all interactive commands that only available in hexo-mode have a `hexo-command-` prefix.

purcell:
Ah, that's much neater! Merging now, thanks. :-)

kuanyui:
Thank you :bow: !

PR #3615, opened 2016-02-22
Added elm-yasnippets recipe

abingham:
This is just a collection of yasnippets for elm-mode.


purcell:
Thanks. Please use `:defaults` instead of `"*.el"`: as noted in the README, it's marginally preferable. The `-pkg.el` file is redundant and can/should be deleted from the repo.

abingham:
Sorry about that...I was just copying someone else's recipe. Everything should be fixed now.

purcell:
Cool, thanks.

PR #3619, opened 2016-02-22
Add evil-multiedit recipe

hlissner:
Description: A multiple cursor implementation that integrates `iedit-mode` into `evil-mode`, bringing Sublime Text's (or Atom's) multiple cursor feature to Emacs.

A little hackish, and is 24.5+ only (because I use the new advising library), but I'll iron out some of those kinks over the next week or two. As I'm relatively new to elisp, any advice would be appreciated!

+ [Repository](https://github.com/hlissner/evil-multiedit)
+ I am the author

### DIFF
--- a/recipes/evil-multiedit
+++ b/recipes/evil-multiedit
@@ -1,0 +1,1 @@
+(evil-multiedit :repo "hlissner/evil-multiedit" :fetcher github)


purcell:
Thanks, looks good, but you should add `(emacs "24.5")` to the `Package-Requires` list if that version of Emacs is currently required. (I *think* 24.4 also has `nadvice.el` though...)

hlissner:
You're right! In that case do you think a `(emacs "24.4")` dependency is still necessary?

purcell:
Yes, you should add that dependency. :-)

hlissner:
Done!

purcell:
Cool, thanks

PR #3620, opened 2016-02-23
Add sudo-edit package to recipes.

nflath:


recipes/sudo-edit
```diff
@@ -0,0 +1,3 @@
+(mv-shell
```
syohex: Wrong package name. `mv-shell` -> `sudo-edit`

recipes/sudo-edit
```diff
@@ -0,0 +1,3 @@
...
+ :url "https://github.com/nflath/sudo-edit.git"
```
syohex: To use `:fetcher github` and `:repo` attribute is better than `:fetcher git` + `:url` attribute.

purcell:
Thanks. Looks good, but please don't do this:

```el
(global-set-key (kbd "C-c C-r") 'sudo-edit-current-file)
```

The user should choose any global keybindings. Maybe suggest the above in the `Commentary` section of the file.

If you want to make the code more robust, you might consider handling the case where `buffer-file-name` is nil. And maybe `(concat "/sudo:root@localhost:" x)` should be extracted as a little helper function to avoid repeating the magic string.

When I commit this I'd squash the two commits to keep our history clean. If you have a moment to that, it would make my life easier.

nflath:
I've fixed the keybinding and squashed the commit.

purcell:
I don't see any change re. the keybinding. Did you maybe forget to push changes?

nflath:
Yep.  Pushed now.

purcell:
Yay, thanks.

PR #3621, opened 2016-02-23
Using :defaults in elixir- and ember-yasnippets.

abingham:
I had copied these recipes when making elm-yasnippets, and in doing I
learned that I should be using :defaults in the :files specifications.
So I figured I'd do a good deed and update these recipes.


purcell:
Yay, that's great - thanks!

PR #3622, opened 2016-02-23
Add yoshi-theme recipe

ryuslash:
*Description*: It is a theme that I started on in 2012 and still use today.

- I am the author/maintainer.
- [repository](https://github.com/ryuslash/yoshi-theme)

### DIFF
--- a/recipes/yoshi-theme
+++ b/recipes/yoshi-theme
@@ -1,0 +1,1 @@
+(yoshi-theme :fetcher github :repo "ryuslash/yoshi-theme")


purcell:
Looks good - just pending a fix (https://github.com/ryuslash/yoshi-theme/pull/1), which is probably due to you copying code from elsewhere. If so, where was that? Perhaps we can get the same fix made upstream.

ryuslash:
I don't think I copied it from anywhere, this is (or was) the way to enable file-local minor-modes. I'll go and accept your pull request now. Thanks!

purcell:
Yay, thanks.

PR #3623, opened 2016-02-23
Add harvest.el recipe

kostajh:
Adds the [harvest.el](https://github.com/kostajh/harvest.el) package, of which I'm the author.

### DIFF
--- a/recipes/harvest
+++ b/recipes/harvest
@@ -1,0 +1,1 @@
+(harvest :repo "kostajh/harvest.el" :fetcher github)


kostajh:
Hmm, the tests didn't run -- apologies if I did something wrong!

purcell:
Thanks. Quick feedback:

- Version number is mandatory in dependencies, so `(swiper)` is invalid.
- Use `(let ...)` rather than `setq` to introduce local vars like `request-url`
- Don't indent subsequent lines of multi-line docstrings
- Don't autoload changes to org-clock hooks: just because somebody has your package installed doesn't mean they want it enabled, just like with all the optional packages in the Emacs standard library.
- Please add a short `Commentary` section to the comment boilerplate describing the package and basic setup/usage

Hope that helps!

kostajh:
Thanks for the feedback, very much appreciated! I'll work on making those changes.

Steve Purcell writes:

> Thanks. Quick feedback:
>
> - Version number is mandatory in dependencies, so `(swiper)` is invalid.
> - Use `(let ...)` rather than `setq` to introduce local vars like `request-url`
> - Don't indent subsequent lines of multi-line docstrings
> - Don't autoload changes to org-clock hooks: just because somebody has your package installed doesn't mean they want it enabled, just like with all the optional packages in the Emacs standard library.
> - Please add a short `Commentary` section to the comment boilerplate describing the package and basic setup/usage
>
> Hope that helps!
>
> ---
> https://github.com/melpa/melpa/pull/3623#issuecomment-188072771


-- 
@kostajh | kosta@kostaharlan.net

kostajh:
@purcell thanks so much for your review & suggestions. I've made some updates, please let me know what you think!

purcell:
Looking good! Couple of little things I picked up still:

- `alist-get` should be called `harvest-alist-get`
- Consider using the built-in `user-emacs-directory` instead of hard-coding `~/.emacs.d`
- You check for `"~/.emacs.d/.harvest/auth.el"` and mention `"~/.emacs.d/harvest/auth.el"` (missing dot) in a message: maybe use a variable so that they always match up?

kostajh:
OK, I've addressed those issues. I really appreciate your code review, it's been helpful!

purcell:
Great, thanks.

PR #3624, opened 2016-02-24
Add recipe for general.el

noctuid:
[This package](https://github.com/noctuid/general.el) provides a single function to make key definition more convenient for both evil and non-evil users. It is written by myself, and I have tested it in the sandbox.

### DIFF
--- a/recipes/general
+++ b/recipes/general
@@ -1,0 +1,3 @@
+(general
+ :fetcher github
+ :repo "noctuid/general.el")


purcell:
Thanks. Feels like `general--delay` should be a macro. That `eval` is extremely fishy. :-)

noctuid:
`general--delay` comes straight from evil. I think I originally tried changing it to a macro but got stuck somewhere. Thanks for merging.

purcell:
Eurgh, okay then! :-)

noctuid:
It is pretty ugly; I'll fix it. Thanks for the suggestion.

PR #3628, opened 2016-02-27
get zeitgeist from new upstream

tarsius:
The old page on launchpad says: "Zeitgeist Data-Sources hosts its code
at git://anongit.freedesktop.org/zeitgeist/zeitgeist-datasources."


purcell:
Cool, thanks.

PR #3629, opened 2016-02-27
add auto-pause

lujun9972:
auto-pause is a library for creating auto-pause-process which will be paused when Emacs idle for specific time and be resumed when emacs become busy again


purcell:
Looks good. Pending https://github.com/lujun9972/auto-pause/issues/3.

lujun9972:
thanks

PR #3632, opened 2016-02-28
Add gams-mode recipe

ShiroTakeda:
This is a major mode for editing GAMS (General Algebraic Modeling System) files.

The package repository: https://github.com/ShiroTakeda/gams-mode
I (Shiro Takeda) am an author and maintainer of this package.

### DIFF
--- a/recipes/gams-mode
+++ b/recipes/gams-mode
@@ -1,0 +1,5 @@
+(gams-mode
+ :repo "ShiroTakeda/gams-mode"
+ :fetcher github
+ :files ("gams-mode.el")
+)


purcell:
Thanks. Quick feedback:

- You shouldn't define unprefixed versions of `line-beginning-position` and `line-end-position`: instead, prefix them as you did with `gams-replace-regexp-in-string`, and as the [elisp conventions](https://www.gnu.org/software/emacs/manual/html_node/elisp/Coding-Conventions.html) suggest.
- `substitute-all-key-definition` should also be prefixed with `gams-` or `gams--`. (Also: does the `define-key` / `[remap]` trick do the same thing as this function?)
- Avoid `:` and `*` in identifier names -- it's not idiomatic, as per the elisp conventions above. Instead, use a meaningful common prefix, e.g. `gams*kill-processor` could be `gams-process-kill-processor`.

Hope that helps! :-)

ShiroTakeda:
Thank you for helpful comments and suggestions.

I want to fix my code according to your comments and Emacs lisp conventions.
After I fix the code, I will let you know.

ShiroTakeda:
Hi

I have fixed my code (gams-mode.el) according to your advices.

I quitted defining `line-beginning-position` and `line-end-position` by myself and use the original `line-beginning-position` and `line-end-position`.

I attached prefix `gams-` to all functions and variables defined in gams-mode.el.

I quitted using symbol `*` and `:` in the name of functions and variables.

purcell:
Great! Thanks - merged. :-)

PR #3633, opened 2016-02-28
Modify emojify recipe to include all files from data directory

iqbalansari:
The above changes are required to support some new features in the package.

Thanks


purcell:
Better than

```el
("*.el" ("data" "data/*") ("images" "images/*"))
```

would be simply

```el
(:defaults "data" "images")
```

Please could you amend the recipe commit accordingly and force-push it here?

iqbalansari:
I have made the change you suggested and forced pushed.

purcell:
Yay, thanks.

PR #3634, opened 2016-02-28
JIRA Backend for Org Export Engine

stig:
This module plugs into the regular Org Export Engine and transforms Org files to JIRA markup for pasting into JIRA tickets & comments.

Its homepage is here: https://github.com/stig/ox-jira.el

I am its author. It certainly doesn't support all JIRA / Org elements, but it supports enough that I find it useful, and others might too.

### DIFF
--- a/recipes/ox-jira
+++ b/recipes/ox-jira
@@ -1,0 +1,3 @@
+(ox-jira :repo "stig/ox-jira.el"
+         :fetcher github
+         :files ("ox-jira.el"))


purcell:
Thanks. Can you please remove the `:files` spec, which is redundant here because the defaults work fine. You can amend the recipe commit and force-push it here to update the PR cleanly.

Re. the code itself, there's already an [org-jira](https://github.com/baohaojun/org-jira) in MELPA, so to avoid clashes I'd suggest sticking with a symbol prefix which matches your package name, ie. rename your `org-jira-*` symbols to `ox-jira-*`.

stig:
Thanks!

I'll do those changes!

Stig

stig:
I've fixed the prefixes. If I remove the `:files` section it looks like the `tangle-org-files.el` file is included in the distribution, which I don't want. It's just there to ensure files are tangled before running tests. (The _actual_ source is in `ox-jira.org`.

purcell:
Oh, okay, cool - merging now. Thanks!

PR #3636, opened 2016-02-29
Add ace-mc recipe

mm--:
This adds the package [ace-mc](https://github.com/mm--/ace-mc), which makes it easy to add and remove multiple-cursor mode cursors using ace-jump.

### DIFF
--- a/recipes/ace-mc
+++ b/recipes/ace-mc
@@ -1,0 +1,1 @@
+(ace-mc :repo "mm--/ace-mc" :fetcher github)


purcell:
Thanks - looks handy! Please use the `ace-mc-` symbol prefix rather than `ace-mc/`. The latter is not idiomatic, and it's good to follow the [elisp conventions](https://www.gnu.org/s/emacs/manual/html_node/elisp/Coding-Conventions.html) for the benefit of the wider community.

mm--:
Sure thing. I've just changed it now.

purcell:
Yay, great. Merged!

PR #3643, opened 2016-03-01
add recipe ob-sly.

stardiviner:
https://github.com/stardiviner/ob-sly

### DIFF
--- a/recipes/ob-sly
+++ b/recipes/ob-sly
@@ -1,0 +1,1 @@
+(ob-sly :fetcher github :repo "stardiviner/ob-sly")


purcell:
It feels like you should be declaring a package dependency on `sly` and simply `require`-ing it. `declare-function` is an unnecessary shortcut, and this package won't work without `sly` installed anyway.

stardiviner:
@purcell I check out MELPA  README, but have not found any `:depend` or similar keywords. Where should I declare this dependency?

stardiviner:
I updated my `ob-sly`. added `(require 'sly)` and commented out that `declare-function` line.

purcell:
Add a metadata header like this:

```
;; Package-Requires: ((sly "2.14"))
```

:-)

purcell:
(i.e. in your ob-sly.el file, next to the `;; Author:` header.)

stardiviner:
Updated, check it out. Thanks for reviewing.

purcell:
Super, thanks.

PR #3646, opened 2016-03-01
Added recipe for zenity-color-picker

Soft:
Package for inserting and adjusting colors using Zenity. https://bitbucket.org/Soft/zenity-color-picker.el

### DIFF
--- a/recipes/zenity-color-picker
+++ b/recipes/zenity-color-picker
@@ -1,0 +1,1 @@
+(zenity-color-picker :fetcher git :url "https://bitbucket.org/Soft/zenity-color-picker.el.git")


purcell:
Thanks for this.

You should really choose a longer symbol prefix than `zc`: there are only 26x26 (676) possible two-character prefixes, but almost 3,000 packages on MELPA. I understand that `zenity-color-picker-` is a bit too long, but `zenity-pc-` or _maybe_ `zpc-` would be better.

Looks good otherwise. :-)

Soft:
Thanks for the feedback. I changed the project to use the longer prefix you suggested.

purcell:
Super, thanks!

PR #3651, opened 2016-03-03
grails-minor-mode recipe

lifeisfoo:
I'm the author of this minor mode that allows an easy navigation of grails projects. 
This recipe isn't related to the existing grails-mode (this is a brand new minor-mode).

I've built it locally and everything seems to be ok:

```
 melpa git:(master) ✗ make recipes/grails-minor-mode
 • Building recipe grails-minor-mode ...
emacs --no-site-file --batch -l package-build.el --eval "(let ((package-build-stable nil) (package-build-write-melpa-badge-images t) (package-build-archive-dir (expand-file-name \"./packages\" package-build--this-dir))) (package-build-archive 'grails-minor-mode))"

;;; grails-minor-mode

Fetcher: github
Source: lifeisfoo/grails-minor-mode

Cloning https://github.com/lifeisfoo/grails-minor-mode.git to /Users/alessandro/Dev/emacs/melpa/working/grails-minor-mode/
Note: grails-minor-mode :files spec is equivalent to the default.
Saving file /Users/alessandro/Dev/emacs/melpa/packages/grails-minor-mode-20160303.1724.el...
Wrote /Users/alessandro/Dev/emacs/melpa/packages/grails-minor-mode-20160303.1724.el
Wrote /Users/alessandro/Dev/emacs/melpa/packages/grails-minor-mode-readme.txt
File: /Users/alessandro/Dev/emacs/melpa/packages/grails-minor-mode-20160303.1724.entry
Built in 0.889s, finished at Thu Mar  3 17:41:54 2016
 ✓ Wrote 16 -rw-r--r--  1 alessandro  staff   7,6K  3 Mar 17:41 ./packages/grails-minor-mode-20160303.1724.el
 8 -rw-r--r--  1 alessandro  staff   161B  3 Mar 17:41 ./packages/grails-minor-mode-20160303.1724.entry
 8 -rw-r--r--  1 alessandro  staff   750B  3 Mar 17:41 ./packages/grails-minor-mode-badge.svg
 8 -rw-r--r--  1 alessandro  staff   1,1K  3 Mar 17:41 ./packages/grails-minor-mode-readme.txt 
 Sleeping for 0 ...
sleep 0
```

### DIFF
--- a/recipes/grails
+++ b/recipes/grails
@@ -1,0 +1,1 @@
+(grails :fetcher github :repo "lifeisfoo/emacs-grails" :branch "master")


purcell:
Thanks. I'd be interested to know how this differs from the existing `grails-mode` -- it would arguably be better for users if you simply contributed any missing functionality to that package.

Quick feedback on the current code:

- It's really not idiomatic to use `-minor-mode` in the name of a package, and such a name would only cause people to be confused about the existing `grails-mode`
- You'll need to prefix all top-level symbols with the package name to avoid possible name classes, as per the [elisp conventions](https://www.gnu.org/software/emacs/manual/html_node/elisp/Coding-Conventions.html), so `extract-name` should be `grails-minor-mode-extract-name` etc.
- The `:files` spec is redundant here, and should therefore be removed from the recipe. Can you amend the recipe commit and force-push it please?
- Keybindings like `\C-c-d` are [reserved for user customisation](http://www.gnu.org/software/emacs/manual/html_node/elisp/Key-Binding-Conventions.html), so you mustn't override them in your minor mode.

Hope that helps!

lifeisfoo:
Hi Paul, first I want to thank you for your advice: this is my first emacs packages, so I'm still learning how to do it.


- Honestly, the existing mode is an [empty](https://github.com/Groovy-Emacs-Modes/groovy-emacs-modes/blob/master/grails-mode.el#L86) and [almost inactive](https://github.com/Groovy-Emacs-Modes/groovy-emacs-modes/commits/master/grails-mode.el) mode: it defines four commands, but only one (listing project files) is implemented, and [it has been around for years](https://github.com/Groovy-Emacs-Modes/groovy-emacs-modes/blob/master/grails-mode.el#L3) without changing. My minor mode provides commands to quickly open domain/controllers/services just typing their names, or automatically grabbing the name from the current context (buffer-file-name).  I'm planning to add more useful features, but that's another story.
- I'm going to rename it with something like `grails-fast-mode`, just to differentiate it better from the existing `grails-mode` and avoid the `minor` word.
- I'll fix the top level symbol as you suggest (thank you for the link).
- I've added the `:files` spec to the recipe since I'm planning to add a test file to the repository. Since I'm not doing it now, I can't add the `:exclude` spec to the recipe now. If it is a problem, I'll remove it.
- I've already read the key bindings convention and noticed that _"sequences consisting of C-c followed by any other punctuation character are allocated for minor modes"_. Why the sequence `C-c` `-` `d` is wrong? From what I read, `-` (minus sign) isn't a reserved punctuation character after `C-c`. 

Let me know what you think.

purcell:
> Honestly, the existing mode is an empty and almost inactive mode: it defines four commands, but only one (listing project files) is implemented, and it has been around for years without changing. My minor mode provides commands to quickly open domain/controllers/services just typing their names, or automatically grabbing the name from the current context (buffer-file-name). I'm planning to add more useful features, but that's another story.

:+1:

> I'm going to rename it with something like grails-fast-mode, just to differentiate it better from the existing grails-mode and avoid the minor word.

Or perhaps just `grails.el`...


> I'll fix the top level symbol as you suggest (thank you for the link).
> I've added the :files spec to the recipe since I'm planning to add a test file to the repository. Since I'm not doing it now, I can't add the :exclude spec to the recipe now. If it is a problem, I'll remove it.

Just call the file `*-tests.el` or `*-test.el` and it will automatically be excluded by the default `:files` spec.

> I've already read the key bindings convention and noticed that "sequences consisting of C-c followed by any other punctuation character are allocated for minor modes". Why the sequence C-c - d is wrong? From what I read, - (minus sign) isn't a reserved punctuation character after C-c.

Oh, right, then the sequence is just a little confusing as written, IMO. Maybe `(kbd "C-c - d")` would be clearer. :-)

lifeisfoo:
Hi Paul, I think I've done everything now.
I've also added the `:branch` spec to the recipe since I use multiple branches during development.

purcell:
Who is this mysterious Paul person? Not me! :-)

I've applied this without the redundant `:branch` spec in be01962. We prefer to build from your default development branch, which indeed appears to be `master` in this case, so `:branch` is unnecessary. P.S. You can place a numeric version tag on your repository at any point, and we'll build a corresponding package in MELPA Stable.

lifeisfoo:
Sorry Steve :fearful:,
this is truly embarrassing. Probably I've read this name elsewhere on github while writing the first message: damned multitasking.

Thank you again for your time and tips.

purcell:
Not a problem. :-)

PR #3652, opened 2016-03-04
Add tabulated-list-ext package

Silex:
Hello!

Please wait a bit before merging.

Things to discuss:

* Wether I should prefix everything with `tabulated-list-ext` or is it fine to extend some obvious features with `tabulated-list` only?
* ????
* Profit!

This idea happend thanks to @dougm because of https://github.com/Silex/docker.el/issues/13

### DIFF
--- a/recipes/tle
+++ b/recipes/tle
@@ -1,0 +1,1 @@
+(tle :fetcher github :repo "Silex/tabulated-list-extensions")


purcell:
> Wether I should prefix everything with tabulated-list-ext or is it fine to extend some obvious features with tabulated-list only?

Use the explicit longer prefix.

Also, I think perhaps this could be a minor mode which adds functionality to tabulated-list-mode? It looks like it provides generally-useful extensions, so providing a derived major mode makes it less convenient.

Silex:
@purcell: interesting approach about the minor mode... I'll think about it and refactor.

Anyhow I'll use the longer prefix and ping you when I'm done.

Silex:
@purcell: I was thinking of renaming the mode to `tle` (tabulated-list-extensions)

Is it ok to name interactive functions `tle/mark-all` or is that for private use only? should I prefer `tle-mark-all` ?

purcell:
> @purcell: I was thinking of renaming the mode to tle (tabulated-list-extensions)

Sure, that would make sense.

> Is it ok to name interactive functions tle/mark-all or is that for private use only? should I prefer tle-mark-all ?

Please prefer `tle-mark-all`. You should avoid `/` -- this is occasionally seen in elisp written by Clojure enthusiasts, but the [elisp conventions](https://www.gnu.org/s/emacs/manual/html_node/elisp/Coding-Conventions.html) prescribe `-` for public symbols and `--` for private symbols.

Silex:
Alright, thanks for your input! /me starts hacking

Silex:
@purcell: alright, I think you can merge :-)

https://github.com/Silex/tabulated-list-extensions

purcell:
Ah, I thought from the comments earlier that your plan was to rename the whole thing to `tle.el`?

Silex:
hum, hum. hum!

Yeah, maybe it doesn't make sense to partially rename it. I kinda liked that the "fullname" was "tabulated-list-extensions" but then commands were shortened to "tle-foo".

Like `multiple-cursors` does, but then it's the one that uses the weird `mc/foo` notation.

If you look at https://github.com/Silex/tabulated-list-extensions/blob/master/tabulated-list-extensions.el, right now only the `provide` and the group name is called `tabulated-list-extensions`, all the rest is named `tle-`.

What do you advice?

Silex:
@purcell: ping :)

purcell:
The current state of the code is okay, and I'd be happy to merge it. Ideal is, I think, when the symbol prefix matches the package name, so that even if someone creates a `tle.el` package, there's no chance of name clashes. A compromise might be `tabulated-list-ext`. Just let me know what you prefer.

Silex:
Hum, that makes sense... but then if you do `package-install`, the name `tle` is really unappealing, while the name `tabulated-list-extensions` is totally guessable about what it does.

I'll think for a while and decide between `tle` or `tabulated-list-extensions`.

Silex:
@purcell: okay, I renamed in the repository https://github.com/Silex/tabulated-list-extensions/

Now I wonder what happens with the recipe here `(tabulated-list-extensions :fetcher github :repo "Silex/tabulated-list-extensions")`

Should I rename here too? I kinda like the idea of the "displayed name" being "tabulated-list-extensions" when package-installing :)

purcell:
Yes, the package name has to match the contained filename.

Silex:
@purcell: alright, you can merge it IMHO

purcell:
Awesome, done - thanks!

Silex:
Thanks! :+1:

Silex:
@purcell: damnit, someone made https://github.com/politza/tablist

It's basically exactly what I did except with 300% more features and just overall better.

It's very likely that I'll delete tle and use tablist for my projects, is there any specific procedure for it? just delete the recipe and make a PR?

purcell:
> @purcell: damnit, someone made https://github.com/politza/tablist
> 
> It's basically exactly what I did except with 300% more features and just overall better.

That means you had a good idea, so don't be sad! ;-)

> It's very likely that I'll delete tle and use tablist for my projects, is there any specific procedure for it? just delete the recipe and make a PR?

I can just revert the recipe commit here.

PR #3653, opened 2016-03-04
Add exwm-x

tumashu:
exwm-x is an extension of exwm

### DIFF
--- a/recipes/exwm-x
+++ b/recipes/exwm-x
@@ -1,0 +1,1 @@
+(exwm-x :fetcher github :repo "tumashu/exwm-x")


purcell:
Thanks.

- Please amend the commit to remove the `:files` spec, which is unnecessary here.
- The `exwm-x-example.el` file should probably not be included in the package, because it will not byte-compile without `use-package`. Perhaps move it to an "examples" subdirectory?

tumashu:
exam-x-example.el is useful, if i put it ot examples subdir,  we should deal with load-path,  
what about:
```
;; Local Variables:
;; no-byte-compile: t
;; End:

```

tumashu:
```
Please amend the commit to remove the :files spec, which is unnecessary here.
```
This has some other problem: how to deal with .dir-local.el file and how to remove snapshot dir?

purcell:
> what about:
> 
> ;; Local Variables:
> ;; no-byte-compile: t
> ;; End:

Yep, fine.

> This has some other problem: how to deal with .dir-local.el file and how to remove snapshot dir?

Neither of those would be included by the default pattern. Check out our README to see the defaults.

tumashu:
Fixed, check again.

purcell:
Super, thanks.

PR #3654, opened 2016-03-04
Play Mode

bomgar:
Adds support for scala play. Currently the routes file is supported.

### DIFF
--- a/recipes/play-routes-mode
+++ b/recipes/play-routes-mode
@@ -1,0 +1,1 @@
+(play-routes-mode :fetcher github :repo "brocode/play-routes-mode")


syohex:
Your package does not follow Emacs packaging convention. Please read following links.
- https://www.gnu.org/software/emacs/manual/html_node/elisp/Coding-Conventions.html

bomgar:
I made some adjustments. Can you please check again?

syohex:
- All variables/functions should start with package prefix, in this case `play-` or `play-mode` for avoiding name conflict. Emacs has only one namespace.
- Use `defvar` instead of `setq` for declaring global variables

bomgar:
I made the changes you suggested. Thanks for your patience.

purcell:
Cool, sent you some fixes in https://github.com/brocode/play-mode/pull/1.

I kinda feel like this package should simply be called `play-routes-mode`.

bomgar:
I renamed the package and the repository. Thank you for your help. I'm new to elisp and trying to learn these things.

purcell:
Awesome, thanks.

PR #3656, opened 2016-03-05
Add ob-sagemath

stakemori:
[ob-sagemath](https://github.com/stakemori/ob-sagemath) provides org-babel support for [SageMath](http://www.sagemath.org/). I am the author of this package.


purcell:
Thansk

PR #3657, opened 2016-03-05
Added projectile-direnv package

christianromney:
This package reads an .envrc file in a projectile project root and sets exported environment variables accordingly. Kindly see [the repository](https://github.com/christianromney/projectile-direnv) for more details/

I am the package maintainer.

Signed-off-by: Christian Romney <cromney@pointslope.com>

### DIFF
--- a/recipes/projectile-direnv
+++ b/recipes/projectile-direnv
@@ -1,0 +1,2 @@
+(projectile-direnv :repo "christianromney/projectile-direnv" :fetcher github)
+


purcell:
Thanks, looks good, but I'm concerned that suggesting putting this into `projectile-mode-hook` is a really bad idea! Specifically, if you're working on multiple projects then as you open files in each project, the environment variables - which are set globally in EMACS - will flip back and forth between those for each project. This will confuse users, so I think it'd be better just to tell them how to manually "activate" the vars for a specific project.

christianromney:
Thanks @purcell. I will definitely disclaim that in the README. Can you think of any way to sanely achieve per-project vars that don't clash? I plan on introducing unload functionality soon, but since the emacs environment is global across buffers, this isn't going to be as clean as I would like. Perhaps keeping track of things in a hashtable keyed by project could work...but your feedback is very much welcome.

purcell:
No - there's no way to avoid clashes since the environment variables are global. They're even shared across multiple frames, so if you have a terminal frame connected to a GUI Emacs, both will share the same env vars!

Rather than come up with a clever solution, I'd probably just prominently describe the behaviour to avoid surprises.

christianromney:
Deal. Thanks for the feedback.

PR #3659, opened 2016-03-05
Add wttrin

bcbcarl:
`wttrin` provides the weather information from [wttr.in](http://wttr.in/).

### DIFF
--- a/recipes/wttrin
+++ b/recipes/wttrin
@@ -1,0 +1,2 @@
+(wttrin :fetcher github
+        :repo "bcbcarl/emacs-wttrin")


purcell:
Thanks. Quick feedback:

- Please depend on the [stable version of `xterm-color`](http://stable.melpa.org/#/xterm-color), not the snapshot version
- `erase-buffer` is preferable to `(delete-region (point-min) (point-max))`
- Don't call `ido-completing-read` specifically: not everyone likes `ido`. Instead, just use `completing-read`, and then the user's preferred completion package can override it (e.g. with `ido-everywhere`, `helm` etc.).

Hope that helps!

bcbcarl:
Fixed on 2461961a6c2b3a72bcabdd543eacb03a501ce0bf. Thank you!

purcell:
Super, thanks!

PR #3662, opened 2016-03-06
Add recipe for rats, a minor mode for running Go tests inside Emacs.

ane:
```
         ,_---~~~~~----._         
  _,,_,*^____      _____``*g*\"*, 
 / __/ /'     ^.  /      \ ^@q   f 
[  @f | @))    |  | @))   l  0 _/            SQUEAK!
 \`/   \~____ / __ \_____/    \       -----´
  |           _l__l_           I   
  }          [______]           I            Rats is a minor mode for running Go tests conveniently
  ]            | | |            |            from inside Emacs, so that you don't have to mess
  ]             ~ ~             |            around with a terminal
   |                           |   
    |                         |   
```

### DIFF
--- a/recipes/rats
+++ b/recipes/rats
@@ -1,0 +1,1 @@
+(rats :fetcher github :repo "ane/rats.el")
--- a/recipes/squeak
+++ b/recipes/squeak
@@ -1,1 +1,0 @@
-(squeak :fetcher github :repo "ane/squeak.el")
--- a/recipes/squeak
+++ b/recipes/squeak
@@ -1,0 +1,1 @@
+(squeak :fetcher github :repo "ane/squeak.el")


purcell:
Thanks. You're defining a bunch of top-level symbols which don't have the `squeak-` prefix, so you should rename those as per the [elisp conventions](https://www.gnu.org/software/emacs/manual/html_node/elisp/Coding-Conventions.html).

purcell:
Also, don't use `loop`. Instead, depend on `(cl-lib "0.5")`, then `(require 'cl-lib)` and use `cl-loop` instead.

ane:
Those top level symbols, you mean e.g. `go-executable-name`, `pass-regexp...` and so on?

ane:
There, should be fixed now, here's the commit https://github.com/ane/squeak.el/commit/25075cd2415756db538d4b3d2aa56062e4253238

purcell:
Super, thanks. Sorry for the delay getting back to you.

PR #3663, opened 2016-03-06
added package for persistent overlays

mneilly:
persistent-overlays is a minor mode that stores overlays between Emacs sessions. By default it stores only overlays with the invisible property. This is useful when using other minor modes such as hideshow or outline so that your hidden text/code is remembered between sessions. It has various custom variables that control auto load, auto merge, storage directory etc.

https://github.com/mneilly/Emacs-Persistent-Overlays

The package is clean per flycheck-package, I have tested that the package builds and can be installed with package-install-file.

### DIFF
--- a/recipes/persistent-overlays
+++ b/recipes/persistent-overlays
@@ -1,0 +1,6 @@
+(persistent-overlays
+ :fetcher github
+ :repo "mneilly/Emacs-Persistent-Overlays"
+ :files ("persistent-overlays.el"
+	 "README.md"
+	 "LICENSE"))


purcell:
Thanks, looks handy. We have a `pov-mode` in MELPA, which (reasonably) uses the `pov-` prefix, so I think you should use `persistent-overlays-` here so that your symbol prefix is unique.

mneilly:
Yeah, that makes sense. I've done the renaming and pushed to my master branch. Thanks!

purcell:
Super, thanks!

PR #3664, opened 2016-03-07
Add recipe for evil-extra-operator

Dewdrops:
This extension adds some useful operators to evil. See [evil-extra-operator](https://github.com/Dewdrops/evil-extra-operator).


purcell:
Thanks, looks good.

PR #3665, opened 2016-03-07
Add exwm and xelb

tumashu:
exwm's author don't use melpa, so i add a pull request to these package, Note, xelb is a dependance of exwm.

### DIFF
--- a/recipes/exwm
+++ b/recipes/exwm
@@ -1,0 +1,1 @@
+(exwm :fetcher github :repo "ch11ng/exwm")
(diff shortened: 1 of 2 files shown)


tumashu:
more detail: https://github.com/ch11ng/exwm/issues/139#issuecomment-193106515

ch11ng:
Please close this PR because both packages are already in GNU ELPA, and I prefer distributing tested versions only.

purcell:
Thanks @ch11ng - yep, we generally avoid adding packages to MELPA if they're already in ELPA, unless the author wishes otherwise.

PR #3666, opened 2016-03-09
Add recipe for decide-mode

CestDiego:
Random decisions for emacs, with functions and a minor mode. Roll
various types of dice etc.

I'm not the author of this package, Got consent in
https://github.com/lifelike/decide-mode/issues/1

### DIFF
--- a/recipes/decide
+++ b/recipes/decide
@@ -1,0 +1,1 @@
+(decide :fetcher github :repo "lifelike/decide-mode")


purcell:
Thanks. Sent @lifelike a small fix in https://github.com/lifelike/decide-mode/pull/3

purcell:
@CestDiego I didn't notice immediately, but this package should be called `decide`, not `decide-mode`, so that it matches the contained filename. Can you amend and force-push the recipe commit please?

CestDiego:
done :)

purcell:
Thanks!

PR #3668, opened 2016-03-09
Add punpun-theme

wasamasa:
I made [a new theme](https://github.com/wasamasa/punpun-theme).

### DIFF
--- a/recipes/punpun-theme
+++ b/recipes/punpun-theme
@@ -1,0 +1,1 @@
+(punpun-theme :fetcher github :repo "wasamasa/punpun-theme")


purcell:
Thanks! These should be two themes, not a single theme which can be reconfigured: your toggle mechanism defeats the customisation system.

wasamasa:
Damnit, you found me out.  I found customize' way of allowing for theme variants pretty clunky, so that's why I went for a single-file approach.

wasamasa:
I've converted it to a multi-file package with two separate themes now.

purcell:
:-)  Looks good. I'd suggest you rename `punpun-theme.el` to `punpun-theme-common.el`, though, otherwise I expect it will show up as a potentially-loadable theme due to its filename, but then produce an error when selected.

wasamasa:
Ugh, Emacs.  I expected `provide-theme` to do that, but apparently that's just done *after* loading the theme.  Fixed.

purcell:
:+1:

wasamasa:
Now that you mention it, how exactly does MELPA figure out the description text for a multi-file package? Judging by the code it checks whether there's a `my-package.el` file and extracts its commentary with `lm-commentary`, but that would fail on my package as `punpun-theme.el` is no more...

purcell:
Correct. There's currently no fallback if there isn't a "main" .el file. It's really only themes like this where it makes much sense to have a separate `-pkg.el` file IMO, and I'm not personally too bothered.

wasamasa:
So the only reason there is a description is because back when I had a `punpun-theme.el`, the description was extracted and cached?

purcell:
I guess so, yes. The way package archives work, there's only one readme file in the repository for each package, and not each package version: when you press '?' in the package menu, it looks for `PACKAGE_ARCHIVE/thepackage-readme.txt`. We've generated that file previously for punpun-theme, and I guess it's still sitting there in the repository.

PR #3669, opened 2016-03-10
add smblog (mode to view samba log files)

aaptel:


purcell:
Thanks. You'll need to add light and dark variants for all your face definitions, or (better) just `:inherit` from some standard faces so that they will immediately look good with any installed theme.

aaptel:
On Sat, Mar 12, 2016 at 1:28 AM, Steve Purcell <notifications@github.com>
wrote:

> Thanks. You'll need to add light and dark variants for all your face
> definitions, or (better) just :inherit from some standard faces so that
> they will immediately look good with any installed theme.
>

I have tested and the basic face look good on both. I have to tweak the
hilight faces more, yes.

purcell:
That's missing the point. Looking good for you is not equivalent to being a good choice for everyone. There will always be themes with which your colours don't look good, and hard-coding colours when not completely necessary is actively hostile to theming, so the advice I offered re. `:inherit` is simply best practice.

aaptel:
Ok. Which faces should I pick from? I don't want something too generic.

purcell:
Generally you'd choose a selection of standard faces like `font-lock-builtin-face`, `font-lock-keyword-face` etc., plus `success`, `warning`, `error`, `highlight`, `outline-1`, `outline-2` etc.

PR #3671, opened 2016-03-10
recipe for itasca pkg. : major modes for Itasca software datafiles.

jkfurtney:
Thanks.

### DIFF
--- a/recipes/itasca
+++ b/recipes/itasca
@@ -1,0 +1,3 @@
+(itasca :repo "jkfurtney/itasca-emacs"
+        :fetcher github
+        :files ("itasca.el" "ac-dict" "snippets"))


purcell:
Thanks. A few little issues here:

- You use `cl-lib`, but that's not available in all Emacsen >= 24. Either declare a dependency on `(cl-lib "0.5")` or depend on a newer Emacs version
- You use `ert`, which is also not available in all 24.x Emacsen.
- The `provide` form should be at the end of the file.
- The long strings of space-separated symbol names should probably just be `'(lisp lists of symbols)`: `symbol-name` will give you a string version of each if you need it

Hope that helps.

jkfurtney:
Thank you for your comments Steve. I have addressed the issues.

* The dependency on `cl-lib` has been removed.
* Emacs 24.3 or greater is required.
* The `provide` has been moved to the end of the file.
* The long strings have been changed into lists of symbols.

purcell:
Nice, thanks.

PR #3674, opened 2016-03-11
Add magit-p4 package.

lexa:
This plug-in is a glue between git-p4 and Magit.

url: https://github.com/qoocku/magit-p4

I'm a contributor.


purcell:
Thanks!

PR #3675, opened 2016-03-12
Add robots-txt-mode recipe

zonuexe:
[robots-txt-mode](https://github.com/zonuexe/robots-txt-mode) is major-mode for editing [Robots exclusion standard](https://en.wikipedia.org/wiki/Robots_exclusion_standard) file.


purcell:
Nice, thanks.

PR #3676, opened 2016-03-13
Add command queue recipe

Yuki-Inoue:
Command queue is a simple asynchronous shell command execution package.
You publish shell commands, they are put into queue.
The queue would be processed each by each.

I'm author.


purcell:
Thanks. Pending https://github.com/Yuki-Inoue/command-queue/pull/1.

Yuki-Inoue:
@purcell Thanks, I've merged your MR.

purcell:
Thanks!

PR #3677, opened 2016-03-15
danneskjold-theme

rails-to-cosmos:
Hello there! Please, checkout my high-contrast theme.

### DIFF
--- a/recipes/danneskjold-theme
+++ b/recipes/danneskjold-theme
@@ -1,0 +1,1 @@
+(danneskjold-theme :fetcher github :repo "rails-to-cosmos/danneskjold-theme")


purcell:
Thanks. The big `setq-default` [block of code](https://github.com/rails-to-cosmos/danneskjold-theme/blob/master/danneskjold-theme.el#L15-L33) there has no place in a theme -- that should be in your personal init file. Same with the `custom-set-faces` modification of the `'default` face, and in any case you should never specify fonts as part of a theme. The rest is fine though. :-)

rails-to-cosmos:
Thanks for your reply, I fixed it.

purcell:
Great, thanks!

PR #3679, opened 2016-03-17
add notify package

tkhoa2711:
add recipe for my library `notify.el`

### DIFF
--- a/recipes/notify
+++ b/recipes/notify
@@ -1,0 +1,1 @@
+(notify :repo "tkhoa2711/notify.el" :fetcher github)


purcell:
Thanks. I'd suggest that we don't merge this, though: the `alert` package does the same thing much more comprehensively, and publishing this particular library with such a general name will just confuse newbies who are looking for a notification library.

tkhoa2711:
That should be fine. I took a look at the `alert` library, it looks pretty decent, honestly.. In fact I was just trying to see how to add a package to melpa :)

purcell:
Cool - thanks for taking the time to put this PR together. :-)

PR #3680, opened 2016-03-17
Add switch-theme recipe

pierre-lecocq:
Hello,

Juste a new recipe for https://github.com/pierre-lecocq/switch-theme

Thanks in advance

### DIFF
--- a/recipes/switch-theme
+++ b/recipes/switch-theme
@@ -1,0 +1,3 @@
+(switch-theme
+ :repo "pierre-lecocq/switch-theme"
+ :fetcher github)


purcell:
Does this differ significantly from https://github.com/toroidal-code/cycle-themes.el ? If not, we should skip adding this.

pierre-lecocq:
Thanks for your comment, Steve!

switch-theme differs from cycle-theme on these points:

- Have a (simple and working) cache system to load the last theme in use before exiting Emacs
- Can have a theme specific configuration via a lambda associated to a given theme (executed when the theme is loaded)

If you think these two points are not significant, therefore, no problem to close this PR!

Regards,
Pierre

purcell:
Thanks, makes sense. I feel like you're reinventing the customisation system to some extent, though: if you automatically save `custom-enabled-themes` when you switch the theme, then Emacs will take care of restoring it for you upon startup. 

It then shouldn't be necessary to save an index offset because you can just find the currently-enabled theme in the list of themes.

When the theme list is empty, I'd consider just defaulting it to the full list of available themes. (And the list should be a list of symbols, not strings, IMO.)

A more general concern is that it's perfectly valid to have multiple themes enabled simultaneously: your lambda case could actually be handled in many cases by an additional enabled theme which contains only variable definitions -- themes don't have to be purely visual.

If you want me to go ahead then I can merge this, but I'd suggest digging into some of those options.

pierre-lecocq:
No Steve, I trust your experience and I will dig in the options you gave me. I was not aware of these.

Even though Melpa reached the 3k recipes (congrats, mates!!), it might not be a good decision to grow that number at the expense of the packages quality.
So I will close this PR and work on the clues you gave me.

Thanks again for your feedback and your wonderful work in general on Melpa !

purcell:
:+1: Thanks for your support - appreciated!

PR #3681, opened 2016-03-17
Add org-bookmark-heading package

alphapapa:
This package provides Emacs bookmark support for org-mode.  You can bookmark headings in org-mode files and jump to them using standard Emacs bookmark commands.  Also supports Helm.

http://github.com/alphapapa/org-bookmark-heading

I am the author.

MELPA and Stable build tests pass, and the package installs and works correctly in the sandbox.

Thanks for your work on MELPA!

### DIFF
--- a/recipes/org-bookmark-heading
+++ b/recipes/org-bookmark-heading
@@ -1,0 +1,1 @@
+(org-bookmark-heading :fetcher github :repo "alphapapa/org-bookmark-heading")


purcell:
Thanks. Few little issues here:

- It would be better if the symbol name prefix matched the package, so perhaps this should be `org-bookmark.el`?
- Using `setq-local` means you must declare a package dependency on at least Emacs 24.3.
- Same with `with-eval-after-load`, though that might actually be Emacs 24.4. But in any case, you should avoid `eval-after-load` in packages, because it's meant for end-user configuration according to the [elisp coding conventions](https://www.gnu.org/software/emacs/manual/html_node/elisp/Coding-Conventions.html). And don't autoload this -- see the next point.
- Don't autoload modifications to `org-mode-hook`: your package should not auto-enable itself. Note that packages in the Emacs standard library do not auto-enable themselves, and third-party packages should work the same way -- give the user a way to enable the functionality.
- `(when (not ...) ..)` is more clearly written as `(unless ...)`

Hope that helps!

alphapapa:
Hi, thanks for your feedback.  This is my first Emacs package.  I went through all the docs carefully and tested extensively, but it seems like I always miss something.

1.  About the name, as I mentioned in the description, there is already a package in `org-mode/contrib` called `org-bookmark.el`.  This is unfortunate, because it provides completely different functionality, not what I would expect from the name.  So I gave it this name to avoid conflicts while being a bit more descriptive.
2.  Thanks, I didn't know that.  I'll add a dependency on Emacs 24.3.
3.  I tried several different ways of doing it, but `with-eval-after-load` seemed like the best way I could find.  That block of code shouldn't run unless the user has `helm` installed, and not until after `helm-bookmark` is loaded.  I thought that I shouldn't `require` it, because my package shouldn't cause `helm` to load automatically.  But it needs to load as soon as `helm-bookmark` is loaded, otherwise the actions won't be available in `helm-filtered-bookmarks`.
4.  This is kind of a tough one.  You see, my package is basically adding missing functionality to org-mode.  I am mystified as to why org doesn't have built-in support for bookmarking headings in org files, but it doesn't, so I put this together to integrate as natively as possible with Emacs' own bookmark support and with org-mode.  So, for example, when a user bookmarks an org heading, the `org-bookmark-make-record` function in this package is stored with the bookmark record in `bookmark-alist`.  So if this package isn't loaded yet, and a user tries to access a bookmark to an org heading made with this package, he will get a `void-function` error.
   
    I've thought about submitting it to org-mode itself, and I may do that, because it really does seem like missing functionality, but I thought it would be good to get it out on MELPA first.

    Anyway, if you insist, I'll remove the autoloads and tell users to `require` the package in their init file.  I thought I would let people avoid that step.  I may be mistaken, but I'm pretty sure I have packages installed that autoload themselves, and for a tiny package like this one that integrates into Org and Helm, it seemed sensible.
5.  Sometimes I forget about `unless`.  Thanks.  :)

So, thanks again for your feedback.  Let me know what you think about these items and I'll fix up the rest.

alphapapa:
Is there anything else I need to do?

purcell:
>     About the name, as I mentioned in the description, there is already a package in org-mode/contrib called org-bookmark.el. This is unfortunate, because it provides completely different functionality, not what I would expect from the name. So I gave it this name to avoid conflicts while being a bit more descriptive.
> 
>     Thanks, I didn't know that. I just noticed setq-mode-local, and it looks like the more correct way to do it. Does it also require a dependency? Can you point me to the information about what versions these functions appear in? I can't find it in the Emacs manual, and a quick googling only seems to bring up a bug filed against another package mentioning it.

>     According to mode-local.el (which is part of Emacs), it was first copyrighted in 2004, so I'm guessing that the Emacs version isn't an issue. So I went ahead and changed it to use that.

I'd probably just use the classic `(set (make-local-variable 'var) value)`. `setq-mode-local` looks a bit more magical.

>     I tried several different ways of doing it, but with-eval-after-load seemed like the best way I could find. That block of code shouldn't run unless the user has helm installed, and not until after helm-bookmark is loaded. I thought that I shouldn't require it, because my package shouldn't cause helm to load automatically. But it needs to load as soon as helm-bookmark is loaded, otherwise the actions won't be available in helm-filtered-bookmarks.

Yeah, I guess the point is that you shouldn't normall be shipping code which may or may not get conditionally loaded later, depending on what packages are available: it typically won't byte-compile cleanly when a user doesn't have that package. The right solution is separate packages for optional extras which integrate with other subsystems like company, helm, yasnippet etc. But hey, I'm not going to get picky here.

>     I added a dependency on Emacs 24.4 by adding this line in the el file. Is that the right way to add a dependency for Emacs? I couldn't find it in the docs anywhere, and grepping through my elpa directory only found two packages declaring an Emacs version, and both were in their *-readme.txt file rather than their el files.
> 
> ;; Package-Requires: ((emacs "24.4"))

Yes, that's perfect.


>     It seems like a good idea to autoload the package, because bookmarks made with it will not be accessible if it's not loaded (see bookmark-make-record-function). It's really strange that org-mode doesn't have built-in support for bookmarking, so it seems like this package is adding missing functionality. But I don't want to violate users' expectations, so I removed the autoloads.
> 
>     Sometimes I forget about unless. Thanks. :)
> 
> So, thanks again for your feedback. Let me know if you think I should change anything else. When it passes inspection, I'll tag it for Stable as well.

Basically looks okay, so I'll merge and leave any further tinkering up to you. Thanks for your patience.

alphapapa:
Thanks for all your help.

PR #3683, opened 2016-03-20
Add smart-comment recipe

paldepind:
This is my first Emacs package so I apologize if anything is wrong.

[smart-comment](https://github.com/paldepind/smart-comment) is documented here.

### DIFF
--- a/recipes/smart-comment
+++ b/recipes/smart-comment
@@ -1,0 +1,2 @@
+(smart-comment :repo "paldepind/smart-comment"
+               :fetcher github)


purcell:
Thanks, great job. Maybe fire up `M-x checkdoc` and fix the suggestions there -- I'm only mentioning it because otherwise it all looks good, and I figure you might be interested to know about checkdoc. :-)  Merging now...

paldepind:
Thank you for merging and for the tip on `checkdoc`. I'm currently trying it out!

PR #3684, opened 2016-03-20
Add scratch-message package

thisirs:
Here is a package to insert and change quotes in the scratch buffer!

Thanks!

### DIFF
--- a/recipes/scratch-message
+++ b/recipes/scratch-message
@@ -1,0 +1,1 @@
+(scratch-message :fetcher github :repo "thisirs/scratch-message")


purcell:
Thanks. `scratch-message-toggle-activate` looks just like a mode function -- I think it's likely you should define a minor mode instead! :-) Looks good otherwise.

thisirs:
Thanks for the review! I thought it was overkill to define a minor mode but
it actually requires less code :) I just committed the change.

2016-03-22 8:46 GMT+01:00 Steve Purcell <notifications@github.com>:

purcell:
Nice!

PR #3689, opened 2016-03-21
Fix MELPA Stable & Contributing links in README

maio:


README.md
```diff
@@ -90,7 +90,7 @@ you prefer to only receive updates for tagged releases, use
 
 ## Contributing
 
-See the [CONTRIBUTING.md][] document.
```
milkypostman: im not sure I like this change as I would like to point out that the file name is part of the repo and is called CONTRIBUTING.md
 I'll commit this without this change. Thanks for pointing out the problem with the other link.

maio: `[CONTRIBUTING.md][]` is invalid Markdown. It renders as:

image

So if you want to keep it, then it should be `[CONTRIBUTING.md](CONTRIBUTING.md)`

milkypostman: whoops, works in some flavors.

I submitted 1ca9af77af79dfcee3811e6ffba60e53859113c2 to clean things up.

milkypostman: and b94357e35825c25816b71cdc3ce0fe161cef0ac0

milkypostman:
i think i fixed everything let me know if I missed something. thanks for the pull request!

PR #3690, opened 2016-03-21
Add turkish-mode package

emres:
This package provides on-the-fly conversion to Turkish characters (such as `ç`, `ö`, `ü`, `ğ`, etc.) for Emacs users who don't have a keyboard with Turkish letters, or who don't prefer to type Turkish letters when typing Turkish words. Users can also select a region, buffer, etc. to convert to Turkish characters. 

https://github.com/emres/turkish-mode

I am the maintainer.

### DIFF
--- a/recipes/turkish
+++ b/recipes/turkish
@@ -1,0 +1,1 @@
+(turkish :fetcher github :repo "emres/turkish-mode")


emres:
Is the `The Travis CI build could not complete due to ...` error above related to the "GitHub status problem of [18:18 CET We're investigating a high number of errors.](https://status.github.com/messages)"?

purcell:
Thanks. The travis failure has cleared up, so don't worry about that.

Specific feedback:
- `;;;` has a specific meaning -- it introduces headers. Please can you format the boilerplate at the top of the file more conventionally. Use `M-x auto-insert` in a fresh file to see what it would look like. Please put the version variable after all the package description comments (if you *must* have one, which I'd advise against).
- Don't bind `C-m` on the user's behalf -- that's one of the keystrokes reserved for user customisation. See the [elisp keybinding guidelines](https://www.gnu.org/software/emacs/manual/html_node/elisp/Key-Binding-Conventions.html).
- The mandatory `(provide 'turkish)` and `;;; turkish.el ends here` lines are missing from the end of the file
- You provide a `turkish-mode` which can be toggled (great!) but that means you shouldn't be changing the user's preferred coding systems unconditionally. Ideally the mode would work without those lines, or the coding systems would be set locally by the minor mode when it is enabled (ie. set the local version of `buffer-file-coding-system` etc.), and then unset when it is disabled.

Hope that helps!

purcell:
Thanks, merged - sent you a small fix in https://github.com/emres/turkish-mode/pull/2.

PR #3692, opened 2016-03-24
Add support for pycoverage.el

mattharrison:
I have tested this locally and it installs on my Mac. Would love to have it in melpa. Cheers!

### DIFF
--- a/recipes/pycoverage
+++ b/recipes/pycoverage
@@ -1,2 +1,2 @@
 (pycoverage :fetcher github :repo "mattharrison/pycoverage.el"
-            :files (files "pycoverage.el" "cov2emacs/bin/cov2emacs" "cov2emacs/cov2emacslib/__init__.py" "cov2emacs/cov2emacslib/meta.py"))
+            :files ("pycoverage.el" "cov2emacs/bin/cov2emacs" "cov2emacs/cov2emacslib/__init__.py" "cov2emacs/cov2emacslib/meta.py"))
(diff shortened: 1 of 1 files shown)


purcell:
Thanks. Please remove the `-pkg.el` file: it overrides the package metadata in `pycoverage.el`, and therefore masks the Emacs 24.3 dependency declared there. Cask also does not need it, so delete the `-pkg.el` file, remove the `(package "pycoverage" ...` line in `Cask` and just keep the `(package-file ...)` line.

The `Cask` file suggests that you'll also need a `:files` entry in the recipe in order to bundle extra files, so perhaps `:files (:defaults STUFF HERE)`, where "STUFF HERE" is extra entries to put the `.py` files and binary in the right places inside the assembled package. See our README for more about that.

Re. the code itself, a couple of comments:
- `after-save-hook` is global by default, so modifying it in a local minor mode won't do what you want. Use the `LOCAL` flag for `add-hook`.
- Seems inappropriate to abuse `linum-mode` for displaying the coverage status: I'd expect overlays or fringe indicators instead
- Prefer `(unless ....)` to `(if (not ...) ...)`. Similarly, when there's no "else" form in an `if`, use `when` instead.

Hope that helps!

mattharrison:
@purcell thanks for the advice. I think I've covered all of it, except for linun vs fringe. I based my mode on another mode that also used linum mode for coverage. I've yet to hear complaints, but will do some investigation.

Is dropping linum a requirement for getting this merged?

purcell:
Nope, merging now in any case, but definitely dig into less hacky alternatives. :-)

PR #3695, opened 2016-03-24
Add goose-theme recipe.

thwg:
[goose-theme](https://github.com/thwg/goose-theme) is an Emacs color theme, which I maintain.


purcell:
Thanks, looks good. Please can you amend the recipe commit to remove the unnecessary `:files` spec, then force-push it here to update the pull request?

thwg:
Okay, thanks.  I've update the PR.

purcell:
Awesome, thanks!

PR #3696, opened 2016-03-24
Add biblio-core

cpitclaudel:
This is the base package for a collection of bibliographic reference
browsing packages; it contains the browsing UI and the search
infrastructure.

I broke this package into small subpackages so that people can install just the sources that they are interested in (in the style of ac or company)

I'm the maintainer :)

### DIFF
--- a/recipes/biblio-core
+++ b/recipes/biblio-core
@@ -1,0 +1,4 @@
+(biblio-core
+ :repo "cpitclaudel/biblio.el"
+ :fetcher github
+ :files ("biblio-core.el"))


purcell:
Thanks, I understand the intent, but I think this packaging is not ideal. The usual reason for splitting packages is not to minimise the amount of code installed, but to minimise dependencies on separate possibly-unwanted packages.

As an example, if I write `foo-mode.el` and provide a `foo-mode-flycheck.el`, then the latter should be in a separate package because it will depend on `flycheck`, and not everyone who wants `foo-mode.el` will want to have `flycheck` installed too.

In your case, it looks like all of these `biblio*` files would form a package which depended only on Emacs, so there's no benefit to splitting them out.

cpitclaudel:
> Thanks, I understand the intent, but I think this packaging is not ideal. The usual reason for splitting packages is not to minimise the amount of code installed, but to minimise dependencies on separate possibly-unwanted packages.

Thanks Steve. I'm not sure I agree: in particular, it's convenient to allow other packages to depend on just parts of this. For example, it could be useful for a package to have a way to resolve DOIs (which biblio-doi provides). Having to download support for a bunch of other APIs would make it harder to depend on this. Similarly, it would be nice to be able to just install one engine instead of many.

Similarly, other packages such as gscholar-bibtex could benefit of the presentation layer (in biblio-core), without requiring their users to download support for five new APIs (gscholar already has support for DBLP, too, so it would add an extra level of confusion).

That being said, what would you think of having just two packages, biblio-core and biblio? (though maybe separating biblio-doi out would be nice too). This would answer teh concerns that I reaised, while making it easy to add new bibliographic sources to biblio.

You call :)

cpitclaudel:
I wonder what the right way to write the definitions would be in that case though. Maybe subdirectories of the original repo? (so that new files are automatically picked). I think it's true that the current packaging scheme will make it annoying for me to add new sources (I'll need a new package every time)

cpitclaudel:
So, I'll either make two or three packages (core + all backends + maybe doi), or just one with everything. Let me know :)

cpitclaudel:
Duh, `:exclude`, silly me.

purcell:
> That being said, what would you think of having just two packages,
> biblio-core and biblio? (though maybe separating biblio-doi out would
> be nice too). This would answer teh concerns that I reaised, while
> making it easy to add new bibliographic sources to biblio.

Yes, either of those splits (2 / 3 packages) sounds like a good
compromise. Would probably suggest the 2-package split for now.

cpitclaudel:
Thanks; update incoming :)

cpitclaudel:
Ok. IIUC this PR doesn't need to be updated; I'll update the biblio one, and close the other ones :)

PR #3697, opened 2016-03-24
Add biblio-doi

cpitclaudel:
Import bibliographic references by DOI

I'm the maintainer :)


purcell:
See #3696

PR #3698, opened 2016-03-24
Add biblio-arxiv

cpitclaudel:
Browse and import references from arXiv (plugs into biblio-core)

I'm the maintainer :)


purcell:
See #3696

PR #3699, opened 2016-03-24
Add biblio-crossref

cpitclaudel:
Browse and import bibliographic references from CrossRef

I'm the maintainer :)


purcell:
See #

PR #3701, opened 2016-03-24
Add biblio-dissemin

cpitclaudel:
Check open access status of bibliographic references

I'm the maintainer :)


purcell:
See #3696

PR #3702, opened 2016-03-24
Add biblio

cpitclaudel:
Browse and import bibliographic references from a variety of sources

This is a meta-package; I expect most people will use this one, but those who don't want all bibliographic sources don't need to install it, and can install just the relevant packages. 

I'm the maintainer :)

### DIFF
--- a/recipes/biblio
+++ b/recipes/biblio
@@ -1,0 +1,4 @@
+(biblio
+ :repo "cpitclaudel/biblio.el"
+ :fetcher github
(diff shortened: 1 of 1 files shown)


purcell:
See #3696

cpitclaudel:
Ok, PR updated to reduce this to two packages :)

cpitclaudel:
I'd love a quick check to make sure that I got that :excludes thing right, but hopefully I did :)

purcell:
Yay, thanks!

PR #3705, opened 2016-03-24
Add indentation to LFE mode

ahjones:
LFE mode now has a separate source file for indentation.


purcell:
Thanks!

yurrriq:
I think this needs [`lfe-start.el`](https://github.com/rvirding/lfe/blob/develop/emacs/lfe-start.el) too, which maybe obviate the need for rvirding/lfe#210.

ahjones:
I didn't add `lfe-start.el` because https://github.com/rvirding/lfe/pull/50 explicitly excluded it.

yurrriq:
Ah, thanks. I'm not to familiar with how MELPA packages work.

PR #3706, opened 2016-03-24
add recipe for chee

eikek:
Hello, 

I'd like to ask for inclusion of this package called “chee”. It is an interface to a command line tool for managing/working with photo collections (which is also called chee). I'm the author of both. The tool is here https://github.com/eikek/chee and the emacs sources are behind the `emacs` folder.

Regards
Eike


purcell:
Great, thanks!

PR #3708, opened 2016-03-26
flycheck-cython recipe

lbolla:
Flycheck interface to Cython.
Code available at https://github.com/lbolla/emacs-flycheck-cython
Author: Lorenzo Bolla <lbolla@gmail.com>

PR #3709, opened 2016-03-26
flycheck-elixir recipe

lbolla:
Flycheck interface for Elixir.
Code available at https://github.com/lbolla/emacs-flycheck-elixir
Author: Lorenzo Bolla <lbolla@gmail.com>

### DIFF
--- a/recipes/flycheck-elixir
+++ b/recipes/flycheck-elixir
@@ -1,0 +1,1 @@
+(flycheck-elixir :fetcher github :repo "lbolla/emacs-flycheck-elixir")


purcell:
Same feedback as for #3708.

lbolla:
Done, too. Cheers.

purcell:
Thanks!

cpitclaudel:
Hey @lbolla,

Thanks for this package! Flycheck used to have native support for Elixir, but we had to disable it because the elixir checker actually ran all the code that it checked; see https://github.com/flycheck/flycheck/issues/630. There's also an open Flycheck issue about restoring support for elixir at https://github.com/flycheck/flycheck/issues/877.

Does your code suffer from the same problem? If so, it could be useful to mention it; if not, maybe we'd be very interested in merging a pull request :)

/cc @lunaryorn

lunaryorn:
@cpitclaudel The definition appears to be the same that we used to have. 

@lbolla I'd appreciate if you could extend the README of your package to include a warning that the syntax checker will execute the buffer.  We could add your package to our list of extensions then.

lbolla:
Indeed, my checker definition suffers the same problem. I've updated the README.

lbolla:
What do you think about having another Elixir checker (`flycheck-elixir-safe`?) using `elixir -e 'r = System.argv |> List.first |> File.read! |> Code.string_to_quoted; if elem(r, 0) == :error do IO.inspect(elem(r,1)); end' -- filename.ex`. People can then use the "safe" on untrusted code and both on trusted one.

cpitclaudel:
@lbolla We don't usually include such programs in flycheck proper, but this would be a good candidate for a separate package, or even for the same package, as a second option :)

PR #3712, opened 2016-03-27
Fix outshine recipe after splitting sources into 2 files.

tj64:


milkypostman:
could you please squash all these commits into one and then repush? it's a bit confusing that there are 5 commits but they only change recipes.

tj64:
Can you just ignore/delete this pull request and instead pull the new one I just made?
It's contains just the one intended commit.

PR #3713, opened 2016-03-27
Adding one-time-pad-encrypt package

garvinguan:


garvinguan:
I fixed the file name conflict. How can I restart the build?

purcell:
A few small issues here:

- The package description on the first line of the .el file is blank
- The filename on the first line of the .el file does not match the actual filename
- The `(provide 'one-time-pad-encrypt)` form is missing after the other code
- The mandatory `;;; one-time-pad-encrypt.el ends here` trailing comment is missing
- You're using camelCase for variable names, when you should be using lisp-case.

Hope that helps!

garvinguan:
Those comments definitely helped. I've just added the above changes. Please let me know how it looks!

Thanks

purcell:
Thanks, sent you a further couple of little fixes in https://github.com/garvinguan/emacs-one-time-pad/pull/2

garvinguan:
I merged the fixes. Great tips, thanks!

purcell:
Great, thanks.

PR #3714, opened 2016-03-27
Fix outshine recipe after splitting source file.

tj64:
Sorry for the confusion, don't know how these old commits entered into the pull request, they should have been pulled already. I think I messed up a trying to squash, so I deleted the melpa fork and started from scratch, now the pull request looks like intended.

### DIFF
--- a/recipes/outshine
+++ b/recipes/outshine
@@ -1,1 +1,1 @@
-(outshine :fetcher github :repo "tj64/outshine" :files ("*.el"))
+(outshine :fetcher github :repo "tj64/outshine")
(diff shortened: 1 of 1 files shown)


purcell:
Please just remove the `:files` spec -- the default includes `*.el` anyway. You can amend and force-push the recipe commit to update the PR cleanly.

tj64:
Squash, amend, force-push ..puh, I hope to use git/magit more frequently again in the future, right now I'm a bit lost here. Sorry for the hassle.

purcell:
Not to worry, thanks anyway -- definitely worth looking into those git tricks though. :-)

PR #3716, opened 2016-03-29
Add helm-org-rifle

alphapapa:
This package lets you "rifle" through Org buffers, quickly displaying results in a Helm buffer.

http://github.com/alphapapa/helm-org-rifle

I am the author.

The build test passes without warnings, and the package installs and works correctly in the sandbox.

Thanks!

### DIFF
--- a/recipes/helm-org-rifle
+++ b/recipes/helm-org-rifle
@@ -1,0 +1,1 @@
+(helm-org-rifle :fetcher github :repo "alphapapa/helm-org-rifle")


purcell:
Thanks!

While I understand the novelty of the "this is my rifle" text, you might consider at least providing a docstring for `helm-org-rifle` which actually describes what the command does, in the Emacs tradition. ;-)

alphapapa:
Haha, good point, I'll have to figure out something... :)  Thanks!

PR #3721, opened 2016-03-31
hfst-mode

unhammer:
Summary: syntax highlighting and minor convenience function for editing
Helsinki Finite State Transducer source files

Link: https://github.com/unhammer/hfst-mode

Association: Author.

### DIFF
--- a/recipes/hfst-mode
+++ b/recipes/hfst-mode
@@ -1,0 +1,1 @@
+(hfst-mode :fetcher github :repo "unhammer/hfst-mode")


purcell:
Thanks. The filename doesn't match the package name, hence the build failure. I suggest you rename `hfst.el` to `hfst-mode.el`, and update the header/footer comments to match. Apart from that, looks okay, though you should remove or change the keybinding, because `C-c <letter>` is reserved for end-user customisation: see the [keybinding conventions](https://www.gnu.org/software/emacs/manual/html_node/elisp/Key-Binding-Conventions.html). Hope that helps!

unhammer:
renamed and changed keybinding, does it look OK? (I don't think I can request a travis rebuild myself.)

purcell:
Looks good, thanks. You might consider using `define-derived-mode` instead of just providing the `hfst-mode` command: that way, you'll save a lot of the boilerplate currently in that command. :)

PR #3722, opened 2016-03-31
weechat-osx-notify

Kungi:
A weechat.el notifier based on terminal-notifier for OSX.

### DIFF
--- a/recipes/weechat-osx-notify
+++ b/recipes/weechat-osx-notify
@@ -1,0 +1,2 @@
+(weechat :fetcher github
+         :repo "Kungi/weechat-osx-notify)


purcell:
Some little issues here:

- Missing quote in the recipe -- I guess you didn't test the recipe locally. ;-)
- Use `shell-quote-argument` if you want to safely escape parameters to programs

But most importantly, weechat is cross-platform, and terminal-notifier is not, so I'd strongly suggest you simply use the excellent [alert package](https://github.com/jwiegley/alert), which allows the user to get notifications however they prefer, including via terminal-notifier.

Kungi:
Thanks for the review.

No I did not test the recipe locally :-(.
Using the alert package sounds like a good Idea. I will rewrite the code and rename it to weechat-alert. Then I will open another pull request.

purcell:
Yay, great!

PR #3723, opened 2016-04-01
Adding a recipe for flycheck-pkg-config

Wilfred:
Flycheck defines a `flycheck-clang-include-path' variable that it
searches for headers when checking C/C++ code.

This package provides a convenient way of adding libraries to that list,
using pkg-config and completion.


purcell:
Thanks!

PR #3726, opened 2016-04-02
Add blockdiag-mode recipe

xcezx:
[blockdiag-mode](https://github.com/xcezx/blockdiag-mode) is major mode for editing [blockdiag](http://blockdiag.com/en/blockdiag/index.html) files.

### DIFF
--- a/recipes/blockdiag-mode
+++ b/recipes/blockdiag-mode
@@ -1,0 +1,1 @@
+(blockdiag-mode :fetcher github :repo "xcezx/blockdiag-mode")


purcell:
Thanks, looks good. Since you're using smie, you should probably declare a dependency on the Emacs minor version which introduced it -- perhaps `(emacs "24.3")`?

purcell:
Also, don't reconfigure the `blockdiag-mode` syntax table every time the mode is enabled -- just do it once at the top level, after the `define-derived-mode` form.

xcezx:
Thank you for a review.
I want make modifications in reference to advice.

PR #3730, opened 2016-04-03
Create redtick

ferfebles:
Little pomodoro timer inside the mode-line.

### DIFF
--- a/recipes/redtick
+++ b/recipes/redtick
@@ -1,0 +1,1 @@
+(redtick :repo "ferfebles/redtick" :fetcher github)


purcell:
I guess you didn't actually try to build this locally per the instructions. :-) We won't automatically find the elisp inside your `src/lib` subdirectory, so you'll either need to flatten that redundant subdir (recommended), or provide an explicit `:files` spec in the recipe anyway. You'd need something like `:files ("src/lib/*")`, given the current structure.

Re. the code itself, some quick feedback:

- The word "emacs" in the package description is redundant: all elisp is for emacs!
- Consider adding an `;;;###autoload` cookie comment for the `redtick-start` command
- You use a variable called `redtick-enabled`, which you check in a number of places -- this is a sign to me that this functionality could be more neatly implemented as a global minor mode.

Hope that helps!

ferfebles:
Sorry :-(

I sent this pull request late at night, and didn't read all the instructions. I will read them carefully, and try to solve the problem with the subdir.

I was thinking too that this could be a minor mode, but I didn't feel confident I could do it. In fact, I don't like too much how the code is arranged. I'm looking for some information about elisp best practices.

And thanks for the code feedback. This is my first elisp code, and I really appreciate your comments.

ferfebles:
Implemented as a minor-mode, and added a ;;;###autoload for the redtick-mode command. I didn't add ;;;###autoload for the redtick-start command, because It has no use without redtick-mode enabled. I removed the redundant emacs references.

I moved the file redtick.el to the package root folder. 

The redtick-mode has no lighter, because I want the 'tick' always at the end of the mode-line.

I've tested it locally and installed without problems, using package-build-current-recipe.

Brief summary.
This package shows a red tick in the end of the mode-line. When you click on it, it shows a yellow one char bar, that during 25 minutes becomes smaller and reddish, until disappears. Then the process starts again but only for 5 minutes and with a greenish colour. At the end of the 30 minutes, the red tick is shown again. It can be restarted at any time, and the elapsed time is shown as a popup when the mouse is over the tick.

Maintainer: me :-)

I hope that everything works ok after reading the manual. And again, thanks for your comments.

purcell:
Code looks good now. (BTW, I'd probably autoload `redtick-start`, and have it enable the mode if it isn't already enabled: seems like that would be a good experience for the user. Maybe also just call that command `redtick`...)

Did you forget to push your recipe change, though? The version pushed here still doesn't work because of the `src` subdir.

ferfebles:
I've changed the code autoloading `redtick-start`, and renamed it as `redtick`. I've tested the current recipe with the redtick repository, and everything seems ok.

I think that the recipe doesn't need to be changed, as I moved the `redtick.el` file to the root folder of the repository. The failed Travis CI build date is two days ago, when the `redtick.el` file were inside the `src/lib` folder.

There is any way of running again the Travis CI build, or do I need to 'change' the recipe (perhaps only the order) in order to rerun the build?

purcell:
Yeah, looks good to me. Merging now - thanks!

ferfebles:
Funny to see my package when I run package-list-packages... :-)
Thanks!!!

purcell:
Step 3: profit. ;-)

PR #3733, opened 2016-04-03
Add recipe for howm

Yuki-Inoue:
Official HP: http://howm.osdn.jp/

"Howm is a note-taking tool on Emacs. It is similar to emacs-wiki.el; you can enjoy hyperlinks and full-text search easily. It is not similar to emacs-wiki.el; it can be combined with any format."

Repo: git://git.osdn.jp/gitroot/howm/howm.git

Association with Package: I'm fan user of this package.

### DIFF
--- a/recipes/howm
+++ b/recipes/howm
@@ -1,7 +1,5 @@
 (howm  :fetcher git
        :url "git://git.osdn.jp/gitroot/howm/howm.git"
        :files
-       ("*.el" "dir"
-        "*.info" "*.texi" "*.texinfo"
-        "doc/dir" "doc/*.info" "doc/*.texi" "doc/*.texinfo"
-        (:exclude ".dir-locals.el" "test.el" "tests.el" "*-test.el" "*-tests.el")))
+       (:defaults
+        (:exclude "*.el.in")))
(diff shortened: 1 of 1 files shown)


purcell:
Thanks. I think the `:files` spec here is equivalent to `(:defaults (:exclude "*.el.in"))` -- please could you confirm that, and update the recipe if so?

purcell:
What Emacs version does `howm` require? I see that it assumes `cl-lib`, but there's no dependency declaration for it. Ideal would be (at least) to add a `;; Package-Requires: ((cl-lib "0.5"))` metadata header comment to `howm.el`.

Yuki-Inoue:
The recipe is fixed as you suggested. For dependecy, I'll contact author and ask if it could be fixed.

Yuki-Inoue:
@purcell I don't mind to ask the author for request to add that package-require line, but one question:

isn't cl-lib a built-in library part of emacs?
While I looked at my system to see why howm has been working, I've notieced it being the built-in library.

purcell:
> isn't cl-lib a built-in library part of emacs?

Only if you have a sufficiently Emacs 24.3 or greater, so there needs to be a dependency on either Emacs 24.3 _or_ `cl-lib` 0.5 (which is installable separately even in older emacs versions). I'd usually suggest the latter.

Yuki-Inoue:
@purcell The official git repo is updated. Could you check?

Yuki-Inoue:
Bump.

purcell:
Great, thanks!

PR #3734, opened 2016-04-03
Add clear-text recipe

xuchunyang:
This is a minor mode which make user use only common words: if uncommon word is inserted, delete it  immediately.

Link: https://github.com/xuchunyang/clear-text.el


purcell:
Thanks!

PR #3735, opened 2016-04-03
Add swap-regions recipe

xuchunyang:
Provide a command to swap two regions of text.

Link: https://github.com/xuchunyang/swap-regions.el

### DIFF
--- a/recipes/swap-regions
+++ b/recipes/swap-regions
@@ -1,0 +1,2 @@
+(swap-regions :fetcher github
+              :repo "xuchunyang/swap-regions.el")


purcell:
You are autoloading the `(de)activate-mark-hook` changes for this, which is inappropriate: it should be possible for a user to install this package but not enable it. Instead, provide a global minor mode which installs those hooks when it is enabled, and removes them when disabled: you can then autoload the minor mode definition, and the user can enable the minor mode if they want this functionality.

xuchunyang:
Thanks for reviewing. I have fixed the issue you described.

purcell:
Nice, looks great! Thanks.

PR #3737, opened 2016-04-03
Adding a recipe for cask-mode.

Wilfred:
cask-mode is a straightforward major mode for editing Cask files.

### DIFF
--- a/recipes/cask-mode
+++ b/recipes/cask-mode
@@ -1,0 +1,3 @@
+(cask-mode
+ :repo "Wilfred/cask-mode"
+ :fetcher github)


purcell:
Thanks. Quick notes:

- Due to use of `setq-local`, you'll need to formally depend on Emacs 24.3 or greater
- In any case, `prog-mode` requires Emacs 24
- `font-lock-defaults` is always buffer-local, so `setq` would do fine there
- Cask uses `package-build.el`, so do you also need to highlight `:github`, `:gitlab` etc?

Wilfred:
OK, those issues should all now be resolved.

In the case of fetchers, I initially consulted the [Cask DSL docs](http://cask.readthedocs.org/en/latest/guide/dsl.html). After a brief skim of the package-build.el code, I've expanded my list:

```lisp
(or ":github" ":gitlab" "bitbucket" "wiki"
              ":git" ":bzr" ":hg" ":darcs" ":fossil" ":svn" ":cvs")
```

Is this all of them? I couldn't see a canonical list of all legal fetchers.

lunaryorn:
@Wilfred Thank you for this mode :heart_eyes: :+1:

purcell:
> In the case of fetchers, I initially consulted the Cask DSL docs and used the values listed there. After a brief skim of the package-build.el code, I've expanded my list:

I think that's probably all of them. You're right that there isn't a definitive list anywhere... I should probably add one. :-)

PR #3738, opened 2016-04-04
Create otama recipe

yoshinari-nomura:


purcell:
Thanks. I also sent you a fix in https://github.com/yoshinari-nomura/otama/pull/1, which I would recommend merging.

yoshinari-nomura:
Thanks. I've merged your PR.

PR #3739, opened 2016-04-04
exclude the ert tests from find-file-in-project

redguardtoo:


purcell:
This is unnecessary, because the `tests` subdir wouldn't have been matched by the default `:files` pattern anyway, so there's no need to exclude it.

redguardtoo:
Ok, thanks for the head up

PR #3741, opened 2016-04-05
Update julia-mode for repo move

ihnorton:
The julia-mode file now lives at https://github.com/JuliaLang/julia-emacs

cc @Wilfred @yuyichao

### DIFF
--- a/recipes/julia-mode
+++ b/recipes/julia-mode
@@ -1,4 +1,4 @@
 (julia-mode
- :repo "JuliaLang/julia"
+ :repo "JuliaLang/julia-emacs"
  :fetcher github
- :files ("contrib/julia-mode.el"))
+ :files ("julia-mode.el"))


recipes/julia-mode
```diff
@@ -1,4 +1,4 @@
 (julia-mode
- :repo "JuliaLang/julia"
+ :repo "JuliaLang/julia-emacs"
  :fetcher github
- :files ("contrib/julia-mode.el"))
+ :files ("julia-mode.el"))
```
Wilfred: This line is unnecessary now I think. The test file will be ignore, and there are no other .el files in the repo.

ihnorton: I don't actually use emacs :grimacing: so I can't test...

(wanted to reply to an on -users with "MELPA package available", but noticed the url was outdated)

purcell: > This line is unnecessary now I think. The test file will be ignore, and there are no other .el files in the repo.

Yes, that's correct: the `:files` spec can and should be omitted in this case.

milkypostman:
ping if we need to commit or something.

yuyichao:
Seems that this is already fixed by https://github.com/melpa/melpa/commit/8522d197cb1b2c139959e7189765001c5ee7e61a

tarsius:
Indeed.

PR #3742, opened 2016-04-05
Add cat-mode recipe

ad1217:
An emacs minor mode that helps manage buffers by assigning them cats (short for categories`).
https://github.com/ad1217/emacs-cat-mode

### DIFF
--- a/recipes/cat-mode
+++ b/recipes/cat-mode
@@ -1,0 +1,1 @@
+(cat-mode :fetcher github :repo "ad1217/emacs-cat-mode")


purcell:
Thanks. Various issues here which need attention.

- Copyright header is blank
- `(ibuffer)` is not a valid dependency spec, because the version number is mandatory. In any case, `ibuffer` is built in, so you don't need this
- All the top-level symbol names should begin `cat-`. So `buffer-cat` and `kill-cat` etc. are not good names, according to the [elisp conventions](https://www.gnu.org/s/emacs/manual/html_node/elisp/Coding-Conventions.html)
- `with-eval-after-load` only arrived in Emacs 24.4, so that's going to be an issue. In practice, you should remove it. You probably want to `(eval-when-compile (require 'ibuffer))` and then put at least some of the code in `cat-init-ibuffer` at the top level. Note that the elisp conventions warn against using `eval-after-load` in anything but end-user configuration code.
- Docstrings are quite lacking: give `M-x checkdoc` a try :-)
- This last one is trivial, but `catagories` on the first line should be `categories`.

Hope that helps!

tarsius:
Please reopen if you still want to get this into Melpa.

PR #3743, opened 2016-04-05
search all words separately

jacktasia:
Currently search only finds exact matches. This PR attempts to improve search results by separately searching each word in the search term. So while every word in a search term is still required, the order no longer matters. 

For example, before and after:

image


purcell:
Nice, thanks!

PR #3746, opened 2016-04-06
add recipe for ob-coffee

pandeiro:
The ob-coffee package adds coffeescript support for org-babel.

(I'm not the author, just a user.)

### DIFF
--- a/recipes/ob-coffee
+++ b/recipes/ob-coffee
@@ -1,1 +1,3 @@
-(ob-coffee :fetcher github :repo "zweifisch/ob-coffee")
+(ob-coffee
+ :fetcher github :repo "zweifisch/ob-coffee"
+ :files (:defaults "repl.js"))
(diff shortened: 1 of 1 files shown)


purcell:
Thanks! This recipe won't include `repl.js`, which is probably a mistake. You might want to add `:files (:defaults "repl.js")`. Don't forget to test it out locally, as per the README.

pandeiro:
Sorry about that Steve, and thanks for checking my work.

I've added the snippet you mentioned and tested locally and it's working.

One doubt I'm left with is whether this package should actually be depending on `(org "8")`, if that is really necessary. It is a large installation and I don't understand whether this will override the user's current (or default) org-mode installation as a side-effect of installing this package. /cc @zweifisch

purcell:
> One doubt I'm left with is whether this package should actually be depending on (org "8"), if that is really necessary. It is a large installation and I don't understand whether this will override the user's current (or default) org-mode installation as a side-effect of installing this package.

That's a trade-off, and there are some specific issues around org as a built-in package in certain 24.x Emacs versions, but generally speaking we should always put correct package dependencies in our packages rather than second-guess how people might have installed the other other code we need. In this case, users will always be able to get a satisfactory package from the orgmode ELPA repository if necessary.

PR #3748, opened 2016-04-06
Add spark recipe

alvinfrancis:
This provides functionality for creating text sparklines in emacs-lisp.  It is a port of the common lisp package [cl-spark](https://github.com/tkych/cl-spark).
Link: [spark](https://github.com/alvinfrancis/spark)
Association: Maintainer

### DIFF
--- a/recipes/spark
+++ b/recipes/spark
@@ -1,0 +1,1 @@
+(spark :fetcher github :repo "alvinfrancis/spark")


purcell:
Thanks. You should remove the `X-Original-Version` header. Whatever you put in your `Version:` or `Package-Version:` header will get rewritten to `X-Original-Version` during our build process.

alvinfrancis:
Okay. I've removed the `X-Original-Version` header. Thanks too.

tarsius:
This depends on `cl-format` which, as far as I can tell, is only available [from Marmalade](https://marmalade-repo.org/packages/cl-format).

@politza, is the repository available somewhere? If not, then please publish it somewhere so that this package can be added to Melpa.

dakrone:
+1 to what @tarsius said, I'd like to use this but it's uninstallable without using the Marmalade repo (which would then make the package I want to use it in require the Marmalade repo). I wasn't able to find a repository for the cl-format code though :(

alvinfrancis:
I didn't realise that `cl-format` wasn't in melpa.  I'll see about rewriting the parts that require it so as to drop that dependency.  But I think it would be better if `cl-format` could also be included in melpa.  Common Lisp `format` is just so good.

purcell:
Happy to add `cl-format` if it lives in an official VC repo somewhere.

dakrone:
Related issue: https://github.com/alvinfrancis/spark/issues/1 looks like cl-format doesn't have a repo yet, but hopefully will soon :)

PR #3749, opened 2016-04-06
Add helm-restclient.el and remove json-reformat.el to restclient recipe

IvanMalison:
@pashky

### DIFF
--- a/recipes/restclient
+++ b/recipes/restclient
@@ -1,3 +1,3 @@
 (restclient :repo "pashky/restclient.el"
             :fetcher github
-            :files ("json-reformat.el" "restclient.el"))
+            :files ("restclient.el" "helm-restclient.el"))


AlexCharlton:
I was just about to open the same PR 😄

purcell:
This isn't what we really need yet, sorry. `helm-restclient` should be a separate package, with a declared dependency on `helm` and `restclient`, otherwise the whole thing would need to depend on `helm` and therefore force all users to have it installed -- not a good outcome. And the [reference](https://github.com/pashky/restclient.el/blob/master/restclient.el#L543-L544) to `helm-restclient` from `restclient.el` should go away.

purcell:
If you can wrangle the corresponding changes upstream, you can push the (now) 2 recipes here to update the PR.

IvanMalison:
@purcell Agreed about the long term fix, but I think that this is better than the alternative, which is that restclient fails to load for all helm users.

helm-restclient.el is only loaded if the user actually already has helm so I don't think it would harm them.

EDIT: I suppose I am making adding a recipe out to be a bigger deal than it is

pashky:
Haha. As I expected, it's easier to get quality help by breaking the world and waiting rather than asking :)

Anyway, what I was trying to achieve here is a zero-configuration soft dependency, so those who have helm installed will get support automatically, but others won't. I realise now it's possible only at the cost of no byte compilation if I put the whole contents of helm-restclient under eval-after-load (I don't understand how with-eval-after-load works yet).

Packaging helm part separately will still require users to install it knowingly, which is not optimal. At the very least I'd need to rename it to restclient-helm or it will get unnoticed by those discovering from package list.

pashky:
@purcell Is it ok to keep reference to helm part under ignore-errors? Or how do I trigger auto-loading of it otherwise when restclient-mode is called?

purcell:
> anyway, what I was trying to achieve here is a zero-configuration soft dependency, so those who have helm installed will get support automatically, but others won't. I realise now it's possible only at the cost of no byte compilation if I put the whole contents of helm-restclient under eval-after-load (I don't understand how with-eval-after-load works yet).

Yep, exactly.

> Packaging helm part separately will still require users to install it knowingly, which is not optimal. At the very least I'd need to rename it to restclient-helm or it will get unnoticed by those discovering from package list.

Yes, that renaming might make sense. I feel like helm users probably often look at all the available `helm-*` packages anyway and install ones which look interesting. :-)

> Is it ok to keep reference to helm part under ignore-errors? Or how do I trigger auto-loading of it otherwise when restclient-mode is called?

In general, you simply shouldn't make that automatic, to be honest. You have hundreds of libraries bundled with Emacs which do not auto-enable themselves: external packages should work the same way, and simple be available for use. The user should instead be easily able to enable them. So consider adding an autoloaded `helm-restclient-setup` command which the user can add to their config.

PS, big fan of `restclient`, thanks for writing it!

pashky:
> In general, you simply shouldn't make that automatic, to be honest. You have hundreds of libraries bundled with Emacs which do not auto-enable themselves: external packages should work the same way, and simple be available for use. The user should instead be easily able to enable them. So consider adding an autoloaded helm-restclient-setup command which the user can add to their config.

Ok, point taken. Let's see if current approach will cause any other problems and introduce setup call then. I kind of like to keep restclient tiny and not requiring any configuration or complex dependencies set up (that's why it's still based on url.el, for example), yet it must provide rich enough functionality to beat Postman for most use cases :)

PR #3753, opened 2016-04-07
Created javascript web labels for LibreJS.

aplaice:
Web labels for:
mithril.js
moment.js
lodash.js
cookies.js
melpa.js

The added break tag in the footer is ugly html, so it might be necessary to change the structure/formatting.

### DIFF
--- a/html/index.html
+++ b/html/index.html
@@ -40,7 +40,11 @@
     </noscript>
   </main>
   <footer class="container">
-    <small><a href="https://github.com/melpa/melpa/blob/master/html/js/melpa.js">Source code for this page &rarr;</a></small>
+    <small>
+      <a href="https://github.com/melpa/melpa/blob/master/html/js/melpa.js">Source code for this page &rarr;</a>
+      <br/>
+      <a href="jslicense.html" data-jslicense="1">JavaScript license information</a>
+    </small>
   </footer>
   <!--[if lt IE 9]>
     <script src="//cdnjs.cloudflare.com/ajax/libs/es5-shim/2.3.0/es5-shim.min.js"></script>
--- a/html/jslicense.html
+++ b/html/jslicense.html
@@ -1,0 +1,72 @@
+<!DOCTYPE html>
+<html lang="en-US">
+  <head>
+    <meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
+    <title>MELPA</title>
+    <meta name="description" lang="en" content="The largest and most up-to-date repository of Emacs packages."/>
+    <meta http-equiv="X-UA-Compatible" content="IE=edge"/>
+    <link rel="shortcut icon" href="favicon.ico" />
+    <link href="/updates.rss" rel="alternate" title="MELPA updates" type="application/rss+xml" />
+    <!--[if lt IE 9]><script src="//cdnjs.cloudflare.com/ajax/libs/html5shiv/3.7/html5shiv.min.js" type="text/javascript"></script><![endif]-->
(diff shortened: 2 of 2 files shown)


html/jslicense
```diff
@@ -0,0 +1,72 @@
...
+          <td><a href="//cdnjs.cloudflare.com/ajax/libs/moment.js/2.8.4/moment.js">moment.js</a></td>
+	</tr>
+	<tr>
+          <td><a href="//cdnjs.cloudflare.com/ajax/libs/lodash.js/2.4.1/lodash.min.js">lodash.min.js</a></td>
+          <td><a href="http://www.jclark.com/xml/copying.txt">Expat</a></td>
+          <td><a href="//cdnjs.cloudflare.com/ajax/libs/lodash.js/2.4.1/lodash.js">lodash.js</a></td>
+	</tr>
+	<tr>
+          <td><a href="//cdnjs.cloudflare.com/ajax/libs/Cookies.js/1.1.0/cookies.min.js">cookies.min.js</a></td>
+          <td><a href="http://www.jclark.com/xml/copying.txt">Unlicense</a></td>
```
mattiasb: Should this link to [Unlicense](http://unlicense.org/UNLICENSE) instead?

aplaice:
Oops! Yes, it should. 
I'll fix it.

aplaice:
Unfortunately, it seems that 
a) LibreJS parses only the URL in the table, ignoring the actual contents (e.g. the name of the license)
b) The Unlicense (or more precisely the link to the license) is not recognised by LibreJS as "approved" and indeed it's not mentioned at: http://www.gnu.org/licenses/javascript-labels.html.

This means that the "javascript web labels" solution currently won't be able to work...

artlogic:
I've asked about unlicense here: http://lists.gnu.org/archive/html/help-librejs/2016-04/msg00000.html

I'm hoping there's a response soon.

artlogic:
I can confirm that this works on my system (except for the unrecognized license. However, I would suggest changing `jslicense` to `jslicense.html` so MIME types can be detected properly.

purcell:
Sorry, this one passed me by. Ready to be merged? Looks good to me.

artlogic:
@purcell - it looks like @aplaice has put in a request (separate from my own) to get unlicense added to librejs: http://lists.gnu.org/archive/html/bug-librejs/2016-04/msg00000.html

It looks like we're still waiting on the librejs folks to respond to either of our requests. It's only been a few days since @aplaice's request, so we may want to give it a bit more time.

Because unlicense is currently unrecognized, this pull request doesn't fix the problem. It does get us closer. I'd still suggest changing `jslicense` to `jslicense.html` so MIME types can be detected properly.

aplaice:
Basically the situation is exactly as described by @artlogic — since LibreJS still does not seem to recognise the Unlicense, implementing these changes to MELPA's website wouldn't benefit anyone, in any way. (In this case, none of the alternative methods of labelling "free" javascript would work either, unless and until LibreJS starts recognising the Unlicense.)

Regarding `jslicense.html`, I've already changed the name of the file and the link to it, as you (@artlogic) suggested.

purcell:
Awesome, thanks -- I'll hold off on merging until you give me the go-ahead. :-) Really appreciate the time you've spent on this.

artlogic:
@purcell I'm thinking it might be best to merge this patch and update the website. At that point it will be up to LibreJS to add Unlicense, and it will be clear we're waiting on them.

purcell:
Done, thanks.

PR #3755, opened 2016-04-08
Add composable.el recipe

paldepind:
My second Emacs package [composable.el](https://github.com/paldepind/composable.el).

Thank you for the great work you do on MELPA.

### DIFF
--- a/recipes/composable
+++ b/recipes/composable
@@ -1,0 +1,2 @@
+(composable :repo "paldepind/composable.el"
+            :fetcher github)


purcell:
Interesting, thanks. You probably want to add some `;;;###autoload` cookie comments.

paldepind:
Yes. But I am slightly confused about `;;;###autoload`. Should I add it to _all_ `defun`s or only some? I've looked at other packages and I can't seem to find a clear pattern.

purcell:
Just the symbols that the user might reasonably want to use without having to require the library, ie. usually those ones which kick off the functionality in the package. I'm guessing that might just be `composable-mode` here.

paldepind:
Ok. Thank you. I might understand it now. `require` is for explicitly loading an entire file. But with autoload users can implicitly load a file just by calling an entry point function that will then trigger the loading?

purcell:
Exactly, yes.

paldepind:
Great. I somehow couldn't figure that out from the manual. Thank you for taking the time to explain :smile:

PR #3758, opened 2016-04-08
Add ten-hundred-mode

aaron-em:
It makes you only use the ten hundred most usual English words (except words that start with big letters, those are okay).

### DIFF
--- a/recipes/ten-hundred-mode
+++ b/recipes/ten-hundred-mode
@@ -1,0 +1,3 @@
+(ten-hundred-mode :fetcher github
+                  :repo "aaron-em/ten-hundred-mode.el"
+                  :files (:defaults "lib"))


purcell:
Thanks. Quick feedback:

- Please use `:defaults` rather than `"*.el"`, as noted in the README
- You may not depend on `cl-lib` 1.0 -- use `0.5` instead, which is installable even in Emacsen which do not already have it bundled. The contents are identical.
- You're using `subr-x` -- you might need to depend on Emacs 24.3 or greater.
- Instead of `(if (not cond) (progn ...))`, just use `(unless cond ...)`.
- The trailing comment should have `;;;`, not `;;`.
- Please use `M-x checkdoc` and fix the suggestions.
- You should add an `;;;###autoload` cookie comment for the `define-minor-mode` form.

Hope that helps!

aaron-em:
That helps a lot, thanks! (Thanks in particular for the mention of `checkdoc`, which had somehow escaped my notice before now. That's really awesome!)

Feedback implemented and squashed onto the prior commit - I noticed I'd hosed the authorship information, so wanted to straighten that out, too. The `subr-x` dependency has been removed; I was only using `string-join` from it, anyway, and it's easy enough to replace that with `mapconcat #'identity`.

Thanks again! I'm still pretty new to writing Emacs libraries that are meant for use by people other than myself, and I really appreciate your taking the time to clue me in on how to do a better job of it.

purcell:
Great, thanks!

PR #3759, opened 2016-04-09
Add recipie for liso-theme

caisah:
Dark theme inspired by TangoTango.

### DIFF
--- a/recipes/liso-theme
+++ b/recipes/liso-theme
@@ -1,0 +1,1 @@
+(liso-theme :fetcher github :repo "caisah/liso-theme")


milkypostman:
only think @purcell will have to comment on is the part that changes ansi colors.

https://github.com/caisah/liso-theme/blob/master/liso-theme.el#L309

purcell:
> only think @purcell will have to comment on is the part that changes ansi colors.

Yes, those and the top-level `setq-default` forms are invalid for a theme. You want those variables applied when the theme is enabled, not just the first time it's loaded. So use `custom-theme-set-variables` instead.

caisah:
Thanks! I've made the changes, I hope it's ok now.

purcell:
Super, thanks.

Issue #3760, opened 2016-04-09
make sandbox results in an error


Yuki-Inoue:
I cloned the latest source code, and kicked `make sandbox` as mentioned in CONTRIBUTING.md. The result was an error of:

```
$ EMACS=`which emacs` make sandbox
 • Updating packages/archive-contents ...
emacs --eval "(package-initialize)" --no-site-file --batch -l package-build.el --eval '(package-build-dump-archive-contents)'
Symbol's function definition is void: package-initialize
make: *** [packages/archive-contents] Error 255
$
```

Am I missing something?

syohex:
Which emacs version did you use ?

```
% `which emacs` --version
```

Yuki-Inoue:
OK, I understand the reason.

```
$ `which emacs` --version
GNU Emacs 22.1.1
Copyright (C) 2007 Free Software Foundation, Inc.
GNU Emacs comes with ABSOLUTELY NO WARRANTY.
You may redistribute copies of Emacs
under the terms of the GNU General Public License.
For more information about these matters, see the file named COPYING.
```

I stumped on this because `emacs --version` showed `24.5.1`, but then I realized I had an aliase setting on my `.bash_profile`.
```
alias emacs='/Applications/Emacs.app/Contents/MacOS/Emacs -nw'
```

Thanks and sorry for taking your time.

Yuki-Inoue:
Reopening because I noticed the problem still exist even after I choose (what I think is ) the correct emacs.

```
$ /opt/homebrew-cask/Caskroom/emacs/24.5-1/Emacs.app/Contents/MacOS/Emacs --version
GNU Emacs 24.5.1
Copyright (C) 2015 Free Software Foundation, Inc.
GNU Emacs comes with ABSOLUTELY NO WARRANTY.
You may redistribute copies of Emacs
under the terms of the GNU General Public License.
For more information about these matters, see the file named COPYING.
$ EMACS=/opt/homebrew-cask/Caskroom/emacs/24.5-1/Emacs.app/Contents/MacOS/Emacs make sandbox
 • Updating packages/archive-contents ...
emacs --eval "(package-initialize)" --no-site-file --batch -l package-build.el --eval '(package-build-dump-archive-contents)'
Symbol's function definition is void: package-initialize
make: *** [packages/archive-contents] Error 255
```

Any idea...?

syohex:
You should use `EMACS_COMMAND` instead of `EMACS` Please see Makefile.

[...]

PR #3761, opened 2016-04-09
Fix make sandbox document

Yuki-Inoue:
Replace the EMACS variable with EMACS_COMMAND, because that's the
variable used in the Makefile.

Refer: #3760


purcell:
Oh, sorry, I didn't this pull request before I applied the same fix myself directly. Thanks anyway!

Yuki-Inoue:
It's OK, thanks for the doc update anyway.

PR #3763, opened 2016-04-10
add recipe ob-sly

stardiviner:


purcell:
Great -- go ahead and reverse the file & symbol naming changes in the repository, then ping me here and I'll merge this PR.

stardiviner:
@purcell It's ok now.

purcell:
Not quite -- `ob-sly.el` still says `ob-lisp` in it.

purcell:
(On the first line, that is. And `org-babel-execute:lisp` etc. should be `org-babel-execute:sly` to avoid clashing with the other `ob-lisp`.)

stardiviner:
@purcell I think `org-babel-execute:lisp` can't be modified. If I change `org-babel-execute:lisp` to `org-babel-execute:sly`, Then I have to write Org-mode babel src block from `lisp` into `sly`.

```org
#+BEGIN_SRC sly
...
#+END_SRC
```
This is not a good solution.

And I think user still can choose to use orginal Org-mode default `slime` by setting option `org-babel-lisp-eval-fn`.
So let this extension `ob-sly` to override org-mode's orginal `ob-lisp` is a solution. I don't know whether it is the appropriate in this case.

PR #3764, opened 2016-04-10
add recipe org-protocol-capture-html

stardiviner:
https://github.com/alphapapa/org-protocol-capture-html


alphapapa:
I appreciate stardiviner's putting this together, but the reason I haven't done this myself yet is that I'm not ready to.  There are several things to configure outside of Emacs, and I want to figure out the best way to present the info about it for MELPA users.

purcell:
Cool, thanks. When the time comes, feel free to ping me here and I can just reopen and merge the PR.

PR #3767, opened 2016-04-11
add recipe:vue-mode

CodeFalling:
[`vue-mode`](https://github.com/CodeFalling/vue-mode) is major-mode for [`vue.js`](https://github.com/vuejs/vue)

### DIFF
--- a/recipes/vue-mode
+++ b/recipes/vue-mode
@@ -1,0 +1,3 @@
+(vue-mode
+ :fetcher github
+ :repo "CodeFalling/vue-mode")


purcell:
Thanks. I've merged this, but I have a couple of suggestions:

- You should probably depend on `web-mode`, because it seems like most (all?) users of this package would need to have it installed
- Rather than add `mmm-mode-on` to `vue-mode-hook`, why not just call it in the `define-minor-mode` form? It seems mandatory, and hooks are generally just for extension.

PR #3768, opened 2016-04-11
add recipe: micgoline

yzprofile:


purcell:
Thanks. Quick feedback:

- The `-pkg.el` file is unnecessary here: best to remove it from the repo
- An `-autoloads.el` file will be generated automatically when the package is installed, so you should remove this one from the repo too
- All the top-level symbols in `micgoline.el` should begin with the name prefix `micgoline-`, as per the [elisp conventions](https://www.gnu.org/s/emacs/manual/html_node/elisp/Coding-Conventions.html).

yzprofile:
Fixed, thanks for these suggestions.

purcell:
Great, thanks.

PR #3771, opened 2016-04-12
Add recipe for helm-notmuch

xuchunyang:
Summary: Helm interface for searching emails in notmuch
Homepage: https://github.com/xuchunyang/helm-notmuch


purcell:
Nice, thanks.

PR #3773, opened 2016-04-12
Added recipe for ivy-bibtex.

tmalsburg:
ivy-bibtex is like helm-bibtex but for the ivy completion framework.  It can be used to perform bibliographic searches in BibTeX libraries.

### DIFF
--- a/recipes/ivy-bibtex
+++ b/recipes/ivy-bibtex
@@ -1,0 +1,1 @@
+(ivy-bibtex :fetcher github :repo "tmalsburg/helm-bibtex" :files ("ivy-bibtex.el" "bibtex-completion.el"))


tmalsburg:
Is there anything holding us back from merging this?  I'm happy to make changes if necessary.  Thank you.

purcell:
It looks like both `ivy-bibtex` and `helm-bibtex` use `bibtex-completion`: it wouldn't be appropriate to bundle that file into both packages. What's your plan?

purcell:
(Sorry for the slow reply, by the way. This one fell off my radar.)

tmalsburg:
I bundled bibtex-completion.el with both ivy-bibtex and helm-bibtex for two reasons: 1.) These packages are alternatives and I doubt that anyone would install both. 2.) I think no one would ever want to install bibtex-completion alone and I didn't want to clutter MELPA with an additional package when there are so few possible uses.

However, I think you are right and it may be cleaner to factor bibtex-completion out into a separate package.  If you favor this solution after reading my reasoning above, I'll go ahead an make a recipe for bibtex-completion.el  Once that's merged, I can modify the recipes for helm-bibtex and ivy-bibtex.

Thank you very much, Steve.  The emacs ecosystem is benefiting a lot from MELPA.

purcell:
Makes sense. I think that since they all live in the same repo, the bundling is unlikely to cause any problems. The right thing to do would probably be to split out the common file as a package, but I'll merge this recipe as-is, and leave you to decide when/if to add that separate package.

PR #3775, opened 2016-04-13
add clomacs recipe

kostafey:
Add [clomacs](https://github.com/clojure-emacs/clomacs) recipe please.


syohex:
I suppose you should add `:files` attribute like `:files ("src/elisp/*.el")` to the recipe file.

purcell:
Super, thanks.

PR #3776, opened 2016-04-13
Add euslisp-mode

iory:
euslisp-mode is a major mode for editing Euslisp code(https://github.com/euslisp/EusLisp).


purcell:
Thanks!

iory:
Thanks!

PR #3779, opened 2016-04-15
Add niceify-info

aaron-em:
I use Info manuals a lot and I thought they could be rendered more usefully. This library does that. It seems like not a lot of people use Info manuals, but those who do might find this useful as well.


purcell:
Great, thanks.

PR #3781, opened 2016-04-16
url has changed

dlwe:


purcell:
Thanks.

dlwe:
On 2016/04/17 8:10, Steve Purcell wrote:
> Merged #3781 <https://github.com/melpa/melpa/pull/3781>.
> 
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/melpa/melpa/pull/3781#event-630121996>
> 

Hello.
I cannot update the package since version dic-lookup-w3m-20151209.1548.
I deleted it and ran package-list-packages. Then it shows an older
version dic-lookup-w3m 20140513.1741.
Do you know why?

milkypostman:
Probably a bad push or changing the url changed the last commit to be
earlier date.

dlwe:
I couldn't update the package since before I pushed the recipe. I have no idea how to fix this...

purcell:
Fixed now, manually on the server: http://melpa.org/#/dic-lookup-w3m, "20160427.1001".

dlwe:
Tahnk you very much.

dlwe:
Hello.
It seems w3m package has the same problem.
I have the version 20151013.327 and cannot update.
On the server http://melpa.org/#/w3m, it appears as 20121224.1747.
Maybe some other packages too?

purcell:
> It seems w3m package has the same problem. I have the version 20151013.327 and cannot update. On the server http://melpa.org/#/w3m, it appears as 20121224.1747. Maybe some other packages too?

No, that package comes from https://github.com/emacsorphanage/w3m, and none of the commits there are as recent as 2015, so the 2012* version looks correct to me. In Dec 2015 we switched the source for w3m from wiki to the above, so probably we were building a 2015* version until that point. Now that we've cleared out all the old builds, the actual version is lower.

It's not a big problem, IMO, because if someone makes commits in the new location, it will result in a 2016* version.

dlwe:
On 2016/04/30 11:48, Steve Purcell wrote:
> Fixed now, manually on the server: http://melpa.org/#/dic-lookup-w3m,
> "20160427.1001".
> 

Can you update again?
It will be 20160526.

purcell:
> Can you update again? It will be 20160526.

When I get some time I'll investigate why these updates aren't happening automatically. I can't keep manually updating to fix this, sorry.

milkypostman:
@dlwe @purcell think I found the bug. working on a fix.

milkypostman:
OK, well did a quick fix and I think it should work now but we'll see. probably broke other things :)

PR #3783, opened 2016-04-18
Added hack-mode

beefsack:
Here is a major mode for [HHVM's Hack](http://hacklang.org/).  This mode is from the [HHVM source](https://github.com/facebook/hhvm/tree/master/hphp/hack/editor-plugins/emacs), moved to a standalone repository which I maintain.

Repository: https://github.com/beefsack/hack-mode

### DIFF
--- a/recipes/hack-mode
+++ b/recipes/hack-mode
@@ -1,0 +1,1 @@
+(hack-mode :fetcher github :repo "beefsack/hack-mode")


purcell:
Maintaining a fork is a bit of an anti-pattern, and likely to lead to the two getting confusingly out of sync over time. I feel like it would be better to either submit any fixes back to the original repo, or lobby to have the HHVM folks appoint you as maintainer of the mode so that they could remove it from their repo. Have you considered or pursued either of those options?

beefsack:
Hi @purcell, yeah that's valid, will close this and see if there's a better way of handling it than forking.

PR #3784, opened 2016-04-18
Add docker-api recipe

Silex:
Hello,

This package https://github.com/Silex/docker-api.el is a generic interface to https://docs.docker.com/engine/reference/api/docker_remote_api, it's meant to be later used by https://github.com/Silex/docker.el as the "implementation" for it.

@purcell: before you merge, can you enlighten me a little bit about the options I have for dependencies/versioning when I have multiple files?

For example, in `docker.el` I use a pkg file like https://github.com/Silex/docker.el/blob/master/docker-pkg.el. I use this because I copy-pasted from another package... but could I also remove this file and put all the dependencies/version in `docker.el` but not the other files? What is the "best" way to deal with this? With the pkg file I had to specify `:url` otherwise it didn't link back to the repo, is that also a problem if I put the dependencies inside `docker.el` ?

Thanks in advance!

### DIFF
--- a/recipes/docker-api
+++ b/recipes/docker-api
@@ -1,0 +1,1 @@
+(docker-api :fetcher github :repo "Silex/docker-api.el")


purcell:
MELPA doesn't need the `-pkg.el` file at all: you can just put all the package metadata in the main `.el` file (ie. the one matching the package name), and we'll get it from there. (Our handling of extended metadata like `url` is patchy in both cases, and something we need to work on over time.)

In any case, as the Travis build pointed out, the current recipe is invalid because the recipe file is named `docker-api` while the name inside the package form is `docker` -- the two must match. You can test the package building process locally using `C-c C-c`, or as described in the README. :-)

Silex:
Ok, I'll get rid of the `-pkg` files then. Thanks! I'll ping you when you can merge this.

Silex:
Ok you can merge this. Thanks!

Silex:
Alright, you can merge this :-)

purcell:
Thanks!

PR #3785, opened 2016-04-18
Add erc-twitch recipe

vibhavp:


purcell:
Thanks. The trailing comment line should read `;;; erc-twitch.el ends here`, not `;;; emotes.el ends here`. Looks good otherwise.

vibhavp:
Fixed.

purcell:
Thanks1

PR #3786, opened 2016-04-18
Added sign recipes

mola-T:
[https://github.com/mola-T/sign](https://github.com/mola-T/sign)

I am the author.

The package is to provide hook like features.
It can achieve something like hook with arguments and and delayed hook call.

### DIFF
--- a/recipes/sign
+++ b/recipes/sign
@@ -1,0 +1,1 @@
+(sign :fetcher github :repo "mola-T/sign")


purcell:
Thanks. The name is quite confusing, I'd say: "sign" is not a natural abbreviation of "signal", so I think the package should be called "signal". Also, please provide some information in the `Commentary` section rather than simply referring people to the github page -- the `Commentary` text appears on the MELPA package package, and also in the `M-x list-packages` menu when a user presses `?`. Few people will discover this package if it isn't clearly described. :-)

mola-T:
Thanks for you advice, I added `commentary`.

For the package name, I can say that I use signal originally. However, `signal` itself is `Emacs` built-in function. During developing the package, I find that there will be `font-lock` conflict. Moreover, I am afraid people will be confused with the `Emacs` built-in `signal` which is related to `error`

mola-T:
Any response?

purcell:
How about "signals", then? That would be clearer. Then you'd have `defsignal`, `signals-connect`, `signals-disconnect` etc. Also, `emitB` would be `signals-emitb`.

mola-T:
I think "signal" is better than "signals" XD
I will make a new pull request. You can close this one.

purcell:
But signal

PR #3790, opened 2016-04-19
added flymake-solidity

kootenpv:
flymake config for solidity (.sol) files


recipes/flymake-solidity
```diff
@@ -0,0 +1 @@
+(flymake-solidity :repo "kootenpv/flymake-solidity.el" :fetcher github)
```
syohex: Repository name is `kootenvpv/flymake-solidity`, `.el` is not necessary.

kootenpv:
Is it now a valid submission?

purcell:
Yes, great - thanks!

PR #3791, opened 2016-04-19
Add recipe for shift-number

alezost:
Hello, would it make sense to add a simple package to
increase/decrease the number at point?


purcell:
Yep, nice. Did you see https://www.emacswiki.org/emacs/IncrementNumber ? Looks like https://github.com/cofi/evil-numbers has some functionality you might like to review and possibly include.

alezost:
Oh, I didn't know about that.  Thanks for merging and for the pointers!

PR #3793, opened 2016-04-19
add xref-js2 recipe

NicolasPetton:
### Description

xref-js2 is a package for navigating to definitions or references in JavaScript projects in Emacs.
It adds an xref backend for JavaScript files.

### Repository
https://github.com/NicolasPetton/xref-js2

### Association with the package
Author

### DIFF
--- a/recipes/xref-js2
+++ b/recipes/xref-js2
@@ -1,0 +1,1 @@
+(xref-js2 :fetcher github :repo "NicolasPetton/xref-js2")


purcell:
`projectile` feels like quite a heavyweight hidden dependency for this. Since the package requires `ag`, have you considered using `ag.el`, which already has some functionality for running `ag` in the current project?

purcell:
Also, I'm a bit freaked out by the Emacs 25 dependency :-)

Can this not work with a 24.x Emacs, with appropriate dependencies for `seq` etc.? Just wondering.

NicolasPetton:
@purcell this requires `xref`, which is new in Emacs 25.
Regarding `projectile` and `ag.el`, AFAIK, `ag.el` uses compilation-mode, which I do not want, but I removed the dependency to `projectile` anyway.

purcell:
:+1: thanks

PR #3794, opened 2016-04-20
Add Kotlin mode from Quantumman/kotlin-mode

jasonm23:


syohex:
This is duplicated with #3727

jasonm23:
Thank you

gregghz:
see: https://github.com/melpa/melpa/pull/3973

PR #3797, opened 2016-04-21
Restore request repositories to original ones

syohex:
I got `tkf/request` permission from @tkf. And I have applied all patches which are applied at @abingham forked repository.

See
- https://github.com/tkf/emacs-request/pull/15#issuecomment-211036531


abingham:
FWIW, I support this change. When we created the fork, the plan was always to switch it back to @tkf's repo when that became viable.

purcell:
Super, thanks.

PR #3798, opened 2016-04-21
Add recipe for embrace

cute-jumper:
Embrace.el: Add/Change/Delete pairs based on `expand-region', similar to `evil-surround'.

See https://github.com/cute-jumper/embrace.el

### DIFF
--- a/recipes/embrace
+++ b/recipes/embrace
@@ -1,0 +1,2 @@
+(embrace :repo "cute-jumper/embrace.el"
+         :fetcher github)


purcell:
Thanks - I'll merge this, but it's really not ideal that a lot of the code refers to `evil-surround` functions. Ideally you'd separate that stuff out into an `evil-embrace` package (which is also a great name!) so that it will byte-compile cleanly for everyone.

cute-jumper:
Thank you for merging this. `Evil-embrace` sounds good! I'll do it.

PR #3799, opened 2016-04-21
Add ejc-sql recipe.

kostafey:
Please add recipe for [ejc-sql](https://github.com/kostafey/ejc-sql).
It's multiple files project, so it may cause some problems....

### DIFF
--- a/recipes/ejc-sql
+++ b/recipes/ejc-sql
@@ -1,3 +1,3 @@
 (ejc-sql :repo "kostafey/ejc-sql"
-         :files ("*.el" "project.clj" "src")
+         :files (:defaults "project.clj" "src")
          :fetcher github)
--- a/recipes/ejc-sql
+++ b/recipes/ejc-sql
@@ -1,0 +1,3 @@
+(ejc-sql :repo "kostafey/ejc-sql"
+         :files ("ejc-sql-pkg.el" "project.clj" "src")
+         :fetcher github)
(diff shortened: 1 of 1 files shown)


purcell:
Have you tried building this locally and testing it, as described in the README? I expect that it won't work, because the elisp will all end up in a "src/elisp" subdir of the installed package, and will therefore not be on the `load-path`. You probably want to put all the elisp in the root of the project.

kostafey:
My idea was to make the stucture common to Clojure projetcs. So, can I put some `load.el` file in the root dir with `add-to-list` call. Other solution you mentioned, is to move all Elisp files to root. Is there ather way? Furthermore, it's the first (and only one at this moument) practice of [clomacs](https://github.com/clojure-emacs/clomacs) usage, so can you give an advice to choose the standard way to build mixed Elisp-Clojure Emacs extensions?

purcell:
Don't use `load.el`, because that name is likely to clash with other files elsewhere on `load-path`. Just put all the elisp in the project root. There's not really any reason to avoid that IMO. I don't really have any other advice to offer on the subject though. :-/

kostafey:
Ok, done!

kostafey:
Is it ok to get `src` directory as is in recipe?
`:files ("*.el" "project.clj" "src")`

purcell:
Yep. If you wouldn't mind, please replace `"*.el"` with `:defaults` -- it's a little more general, and equivalent in this case.

kostafey:
Done.

purcell:
Super, thanks!

kostafey:
Thank you!

PR #3800, opened 2016-04-22
EGO: a static site generator based on Emacs, Git and Org-mode

kuangdash:


purcell:
Quick feedback:

- `(ido "")` in the `-pkg.el` file is an invalid dependency for two reasons: 1) the version number is mandatory, and 2) `ido` is built-in, and not installable separately. So you should remove that dependency.
- `(cl-lib "1.0")` should be `(cl-lib "0.5")`, because the former version is not installable separately in older Emacs versions
- Symbols should have the form `ego-...`, not `ego/...`, according to the [elisp conventions](https://www.gnu.org/s/emacs/manual/html_node/elisp/Coding-Conventions.html) -- this is not Clojure or Common Lisp, and the best thing for the community is if we all follow the conventions.
- `string-suffix-p` is relatively new, I think, so you might need to declare a dependency on `(emacs "24.3")` or similar

kuangdash:
Thank you for the comment,  I have modified the contents of ego-pkg.el .

But when it comes to the point 3: " ego-..., not ego/...," … I‘m afraid of the conflict between “ego” and other packages with “ego-” prefix. (though there is none, temporarily)

purcell:
> But when it comes to the point 3: " ego-..., not ego/...," … I‘m afraid of the conflict between “ego” and other packages with “ego-” prefix. (though there is none, temporarily)

Well, `ego/` doesn't make that much less likely, and some of the `ego--` functions might still clash, so that's not really an argument for '/'.

Essentially, if you publish a package called `ego`, then you "own" the `ego-` prefix -- if someone else has clashing names, then that's their problem, not yours.

kuangdash:
That elisp convention is strange …

 It says “Separate the prefix from the rest of the name with a hyphen, ‘-’. This practice helps avoid name conflicts” . That practice were created in order to avoid name conflicts while there is no rational explanation, and I think that it has no effect on resolving name conflicts.

purcell:
Yes. That's the convention, so let's stick with it. :-) Choosing new different conventions isn't a concrete solution to anything. Once you've published `ego.el`, the `"ego-"` prefix is legitimately "yours", and you can use it without worrying about possible future name clashes.

kuangdash:
Considering the 'ego-git', 'ego-util' and etc, I have changed 'ego/' into 'ego-'…the related part of elisp convention is still controversial, after all.

purcell:
Great, thanks!

PR #3801, opened 2016-04-22
Add recipe for pathify

alezost:
Hello, I'm not sure whether this package may be useful for anyone else.
Its only purpose is to make symlinks in `~/bin` (or another directory).

If it doesn't look appropriate for MELPA, please don't hesitate to
reject it.  Thanks.

https://github.com/alezost/pathify.el

### DIFF
--- a/recipes/pathify
+++ b/recipes/pathify
@@ -1,0 +1,1 @@
+(pathify :repo "alezost/pathify.el" :fetcher github)


purcell:
Looks good, thanks.

The following code isn't really ideal:

```el
(defcustom pathify-directory
  (let ((dir (expand-file-name "~/bin")))
    (if (file-directory-p dir)
        dir
      (message "WARNING: set `pathify-directory' variable.")
      temporary-file-directory))
  "Directory where symlinks are created."
  :type 'string
  :group 'pathify)
```
because it's perfectly valid for people to load this file before loading their custom variable settings, and those people will see the warning.

So just set the default to `~/bin` and then throw an error when the commands are called if the directory doesn't exist.

alezost:
Ouch, indeed.  I didn't think about it.  Thanks for pointing at
this and for merging!

PR #3802, opened 2016-04-23
Add signal recipe

mola-T:
https://github.com/mola-T/signal

I am the author.

The package is to provide hook like features.
It can achieve something like hook with arguments and and delayed hook call.

Thank you.

### DIFF
--- a/recipes/signal
+++ b/recipes/signal
@@ -1,0 +1,1 @@
+(signal :repo "Mola-T/signal" :fetcher github)


purcell:
Thanks. I've merged this, but `signal-emitB` should be `signal-emitb` (lower-case identifiers only, please), and "advance hook" should be "advanced hook". :-)

mola-T:
All fixed. Thanks.

PR #3803, opened 2016-04-23
Add fifo recipe

mola-T:
https://github.com/mola-T/fifo

I am the author.

The package contains an abstract class to provide first in first out functions for inherited class.

Thank you.

### DIFF
--- a/recipes/fifo
+++ b/recipes/fifo
@@ -1,0 +1,1 @@
+(fifo :fetcher github :repo "mola-T/fifo")
--- a/recipes/fifo
+++ b/recipes/fifo
@@ -1,1 +1,0 @@
-(fifo :fetcher github :repo "mola-T/fifo")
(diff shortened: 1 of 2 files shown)


purcell:
It's unusual in elisp for authors to use eieio classes: anyone wanting to use the `fifo` functionality would need to create a subclass, and very few people would want to do that. So using such a generic name for this package isn't a good choice, and I'd suggest renaming to `fifo-class.el`. (P.S. For an example of a more idiomatic elisp fifo implementation, see https://github.com/commercialhaskell/stack-ide/blob/master/stack-mode/stack-fifo.el)

mola-T:
I will change the package name to `fifo-class`.

mola-T:
You can have a look. Thank you.

purcell:
Nice, thanks.

PR #3805, opened 2016-04-23
Add thread recipe

mola-T:
https://github.com/mola-T/thread

I am the author.

It is a multithreading library.

Need these two recipes to be approved first.
https://github.com/melpa/melpa/pull/3803
https://github.com/melpa/melpa/pull/3802

Thank you.

### DIFF
--- a/recipes/thread
+++ b/recipes/thread
@@ -1,3 +1,0 @@
-(thread :repo "Mola-T/thread"
-        :fetcher github
-        :files ("*.el" (:exclude "thread-pkg.el")))
--- a/recipes/thread
+++ b/recipes/thread
@@ -1,0 +1,3 @@
+(thread :repo "Mola-T/thread"
+        :fetcher github
+        :files ("*.el" (:exclude "thread-pkg.el")))
--- a/recipes/timp
+++ b/recipes/timp
@@ -1,3 +1,2 @@
 (timp :repo "Mola-T/timp"
-      :fetcher github
-      :files ("*.el"))
+      :fetcher github)
--- a/recipes/timp
+++ b/recipes/timp
@@ -1,0 +1,3 @@
+(timp :repo "Mola-T/timp"
+      :fetcher github
+      :files ("*.el"))


wasamasa:
@mola-T This is a pretty misleading name for something not giving you threads at all.  Have you considered naming it differently, like mp.el?  Also, how is this different from [async.el](https://github.com/jwiegley/emacs-async/)?

mola-T:
I am opened to any advice on the name.
My package is different from [async.el](https://github.com/jwiegley/emacs-async/) that threads can be persist and jobs send to thread can be separated and accumulated.
Persist implies you can use timer in the thread and it is also capable to hold a database.

purcell:
Quite a few issues here, sorry:

- The name is too general, particularly for a new and unproven package. It also seems likely that `thread-*` identifiers might later appear in Emacs. So a different name would be good.
- For `advice-add`, you need to depend on at least Emacs 24.4
- `thread.packet.getQuitWarn` etc. do not conform to the [elisp conventions](https://www.gnu.org/s/emacs/manual/html_node/elisp/Coding-Conventions.html)
- Random indentation and empty lines
- Don't mention "Emacs" in the package description -- all elisp is for Emacs! :-)
- The package dependencies in `thread.el` don't match those in `thread-pkg.el`. If you must provide the latter file, then put the info there.
- Please include some usage info in docstrings and the `Commentary` section of the main `.el` file: that info will appear on the MELPA package page and in the `list-packages` menu. Emacs code should be self-documenting, so directing people to a README on github isn't ideal.

Hope that helps!

mola-T:
Since making change on this package is not easy, I would like to ask for your advice first.

1. Do you think `timp` is a good name which stand for "thread in multi process"
2. `Emacs 24.4` will be appended to the package requirement.
3. I hope I can keep the naming style of the package. I have two reasons. First, the package is very OO in nature. Most of them are objects. I think the `dot` can reflect the relation of the object. For example, under `thread.packet.getQuitWarn`, `thread.package` is an object `getQuitWarn` is the method of the object. At the same time, OO may not be a convention in Emacs. The second reason, the package is a library. The end users hardly see any of these functions. There won't be no confusion do to end users. However, I appreciate and respect you as you pay great effort to test and examine packages. I expressed my hope to keep this style. I am also happy to have you holding the final word whether I must change the style. Please reply me whether you need me to change. I will do it without hesitate.
4. Random indentation and empty lines will be fixed.
5. Don't mention "Emacs" in the package description < will be fixed.
6. About `thread-pkg.el`, I thought I need to provide this in multi-files package when I read the Emacs Lisp Manual. If it is not, I will remove it.
7. `Commentary` will be enriched.

Thanks for you patient and effort.
I hope I can get response on point 1 and 3.

purcell:
Re. 1: yes, `timp` would be a reasonable choice.
Re. 3: the [`eieio` docs](https://www.gnu.org/software/emacs/manual//html_mono/eieio.html#Naming-Conventions) say that you should follow the regular elisp conventions in this case, so I really think that's the best choice if you expect anyone to ever use this package.

mola-T:
Please have a look ^^

mola-T:
Any progress?

purcell:
Sorry for the delay. Yes, looks good, but could you remove the `:files` spec from the recipe please? The default already includes "*.el", so you don't have to override it.

mola-T:
Thanks Done^^

purcell:
Thanks!

PR #3806, opened 2016-04-23
Add recipe for emms-status

alezost:
Hello, I was not satisfied with the existing `emms-mode-line`
look, so I wrote an alternative.  Would it be acceptable to add
it to MELPA?

https://github.com/alezost/emms-status.el


alezost:
Thanks for  merging!

PR #3807, opened 2016-04-23
Add recipe for helm-systemd

Lompik:
Helm's interface to control systemd units (on linux).

I created of the package.
https://github.com/lompik/helm-systemd

I tested using `emacs -Q` and `package-install-file`. Not sure where to put autoloads magics. Only one function for the user.

### DIFF
--- a/recipes/helm-systemd
+++ b/recipes/helm-systemd
@@ -1,0 +1,3 @@
+(helm-systemd :repo "lompik/helm-systemd"
+                 :fetcher github
+                 :files ("helm-systemd.el"))


purcell:
Thanks. Quick feedback:

- For with-editor, you should depend on 2.5.0, not the snapshot version (see http://stable.melpa.org/#/with-editor)
- If you're depending on Emacs 24.4, I'd strongly suggest you go ahead and use `cl-lib` instead of the long-deprecated `cl` library
- Use a consistent namespace prefix for all of your public symbols, please, as per the [elisp conventions](https://www.gnu.org/s/emacs/manual/html_node/elisp/Coding-Conventions.html). Names like `sysd-*`, `concatspace` and `get-sd-canditates` are not valid

Lompik:
Thanks ! All applied.

purcell:
Great, thanks.

PR #3808, opened 2016-04-23
ycmd - Don't install third-party folder

ptrv:
Since the original repo of `emacs-request` is updatead again (#3797) `ycmd` is using the upstream version of request and we don't need to install the bundled version.


ptrv:
Upstream `request` causes problems in `ycmd`. I need to look into it. I'll reopen the PR when the issues are fixed

ptrv:
Issues with upstream `request` package have been fixed and we use it now instead of our own `ycmd-request` version.

purcell:
Thanks.

PR #3810, opened 2016-04-24
add recipe sendto

lujun9972:


purcell:
Nice, thanks - looks useful. Consider going over it with `M-x checkdoc` and fixing the recommendations. :-)

PR #3813, opened 2016-04-26
Add selected recipe

Kungsgeten:
`selected.el` is a package which enables a keymap when the region is active. https://github.com/Kungsgeten/selected.el

### DIFF
--- a/recipes/selected
+++ b/recipes/selected
@@ -1,0 +1,1 @@
+(selected :repo "Kungsgeten/selected.el" :fetcher github)
--- a/recipes/steam
+++ b/recipes/steam
@@ -1,0 +1,1 @@
+(steam :repo "Kungsgeten/steam.el" :fetcher github)


Kungsgeten:
Sorry, I added another recipe and did not realize that it would merge with this pull request :/

purcell:
So do you want both to be reviewed? If not, you can update or replace this PR.

Kungsgeten:
Yes please, I would like both reviewed, but I read that I was supposed to submit a separate pull request for each package I wanted to submit.

However `steam.el` is also a package I've created. It is a way to launch Steam games from Emacs. It also provides features to insert the user's Steam library in an org-mode file.

purcell:
`selected` is a really nice idea. Quick feedback:

- I wonder if the very generic name is going to cause some confusion, and that something like "active-region-mode" might be clearer
- The minor mode `selected-region-active` should be called `selected-region-active-mode`.

Quick feedback on `steam.el`:

- You should drop the ` from Emacs` from the package description line, since it's redundant: all elisp is for doing stuff from Emacs :-)
- You might have to hex-encode certain usernames for this to be safe: `(format "http://steamcommunity.com/id/%s/games?tab=all&xml=1" steam-username)`
- And you should probably use shell-quote-argument here just to be safe: `(shell-command (format "explorer steam://rungameid/%s" id))`
- It would be worth adding `;;;###autoload` cookies, at least for `steam-launch`

Kungsgeten:
The reason for excluding `-mode` from `selected-region-active` was that the command wasn't supposed to be run by the user. However, I've added `-mode` now. The name of the package could be changed (but I guess I have to submit another recipe and change the repo name etc. I'm not very used to working with git). 

There's a package already on Melpa named [region-bindings-mode](https://github.com/fgallina/region-bindings-mode), so after learning this I was hesitant to  submit `selected.el`. However someone told me that `region-bindings-mode` hadn't been updated in a long time and that it had problems in Emacs 25. `selected.el` also provides major-mode specific bindings.

I've commited your suggested changes to `steam.el`.

purcell:
Nah, that's okay -- I'll merge both of these now as they are.

PR #3814, opened 2016-04-26
Add the bin and resources folder to slime-docker.

daewok:
bin/ currently contains a script to make it easier to forward SSH Agent
into containers on computers using docker-machine.

resources/ currently contains .json files describing seccomp profiles
for Docker, including one needed to get rid of some SBCL warnings.


purcell:
Thanks. `:files` here can simply be `(:defaults "bin" "resources")`. Can you update the PR accordingly please?

daewok:
Done! Thanks for the pointer.

purcell:
Thanks

PR #3820, opened 2016-04-27
Add tawny-mode

phillord:
Tawny provides richer support for the tawny-owl Clojure library.

I am the author of tawny-mode.el and tawny-owl.

### DIFF
--- a/recipes/tawny-mode
+++ b/recipes/tawny-mode
@@ -1,0 +1,3 @@
+(tawny-mode
+ :fetcher github :repo "phillord/tawny-owl"
+ :files ("emacs/tawny-mode.el"))


purcell:
Thanks! Quick feedback:

- You may not bind keys like `(kbd "C-c s s")`, since `C-c <letter>` is reserved for user customisation. See the [keybinding conventions here](https://www.gnu.org/software/emacs/manual/html_node/elisp/Key-Binding-Conventions.html).
- `if-let` is probably newer to Emacs than the minimum version which `cider` requires, so you might need to depend explicitly on a recent Emacs version.
- `M-x checkdoc` ;-)
- `tawny-mode-check-for-nrepl-buffer` is clunky, and so is having to check its return value in multiple places: it would be more idiomatic to simply `error` or `user-error` when there's no `nrepl` connection

phillord:
@purcell You're a harsh critic:-)

purcell:
Yay, thanks!

> @purcell You're a harsh critic:-)

Guilty, sorry. :-)

PR #3822, opened 2016-04-28
Add recipe for evil-lispy

sp3ctum:
Hi guys! I'd like to request the addition of [evil-lispy](https://github.com/sp3ctum/evil-lispy) to the archive.

It's a minor mode whose purpose is to make using [lispy](https://github.com/abo-abo/lispy) convenient for evil-mode users. To this end, evil-lispy emphasizes the use of evil-mode bindings and modal editing as the main navigation mode, and for more advanced editing provides the full command set provided by lispy in a toggleable minor mode (an evil state to be precise).

As for the project history, I've picked up development left off from @bcarrell in my own fork. With regards to this I'd like to call for discussion. While I know the guidelines state forks in general are not accepted.
However, I've tried to [contact](https://github.com/sp3ctum/evil-lispy/issues/10) the original author a week ago, and received no response (@bcarrell if you see this, please let me know what you think!). Please let me know what you think about all of this :smile:

### DIFF
--- a/recipes/evil-lispy
+++ b/recipes/evil-lispy
@@ -1,4 +1,3 @@
 (evil-lispy :repo "sp3ctum/evil-lispy"
             :fetcher github
-            :branch "master"
-            :files ("*.el"))
+            :branch "master")
--- a/recipes/evil-lispy
+++ b/recipes/evil-lispy
@@ -1,0 +1,4 @@
+(evil-lispy :repo "sp3ctum/evil-lispy"
+            :fetcher github
+            :branch "master"
+            :files ("*.el"))


bcarrell:
No objections here.

purcell:
Thanks. Mostly looks great. Quick feedback:

- Please remove the redundant `:files` spec from the recipe - per the README, we prefer to leave it unspecified when the default value suffices, as it does here.
- The package dependencies are malformed: `;; Package-Requires: (lispy evil)` -- version numbers are mandatory, so this should be something like `;; Package-Requires: ((lispy "0.26") (evil "1.2.12"))`
- You should probably avoid binding `C-1` and `C-2` -- those are standard emacs bindings for prefix arguments, and you shouldn't break them for people.

Cheers!

sp3ctum:
:+1: 
- `:files` specification has been removed. Really good idea.
- version numbers have been added
- `C-1` and `C-2` are actually present to be familiar to lispy. See the [lispy readme features section](https://github.com/abo-abo/lispy#features), namely:

```
C-1 shows documentation in an overlay
C-2 shows arguments in an overlay
```

Do you think it's okay?

purcell:
> C-1 shows documentation in an overlay
> C-2 shows arguments in an overlay
> 
> Do you think it's okay?

Well, it's not a blocker, but perhaps something to mention to @abo-abo.

PR #3824, opened 2016-04-29
Added shen-elisp.

deech:


deech:
Github link: http://github.com/deech/shen-elisp

Description: An implementation of the Shen programming language in Elisp.

PR #3825, opened 2016-04-29
Added shen-elisp recipe.

deech:
http://github.com/deech/shen-elisp

An implementation of the Shen programming language in Elisp.

I am the package author.

### DIFF
--- a/recipes/shen-elisp
+++ b/recipes/shen-elisp
@@ -1,0 +1,4 @@
+(shen-elisp
+ :repo "deech/shen-elisp"
+ :fetcher github
+ :files ("shen*.el"))


purcell:
Hi! Change of heart? Not ready?

deech:
It wasn't. It is now. Very much appreciate the bump!

purcell:
Thanks - looks very cool!

Quick questions/comments:

- Do you intend to bundle `install.el`?
- Technically `:defaults` is preferred to `"*.el"` in recipes, since it also includes other files you might add later, e.g. info manuals
- We normally wouldn't bundle `.org` files into packages because they're not at all discoverable after installation. And `shen-elisp.org` looks like the tangled source for the `.el` files, so that really isn't needed afaict.
- At least for public commands, you should follow the [elisp coding conventions](https://www.gnu.org/s/emacs/manual/html_node/elisp/Coding-Conventions.html) rather than CL/Clojure style, so `shen-repl` rather than `shen/repl`.
- Snippets like `(require 'shen "shen.el")` are just workarounds for broken installations, and might subvert loading of byte-compiled files. Just use `(require 'shen)`

deech:
Thanks for your feedback!
- No the install.el isn't doing anything.
- I'll remove the org files
- In fact the KLambda/*.kl should probably be removed as well.
- Ah, I didn't know about the naming conventions. Unfortunately I doubled down on the 'shen/' prefix early on for all the written and generated Elisp so I'd rather not change that now.
- I'll remove the filenames from the `require` forms.
Thanks for your feedback!

purcell:
Great thanks - this misses the `-pkg.el` file though. Would you mind adding that into the `:files` list please? `"shen*.el"` might do the trick.

deech:
Sorry for all the errors. This is now fixed and intermediate commits have been squashed. Thanks again!

purcell:
Awesome, thanks.

PR #3826, opened 2016-04-29
Add elisp-format recipe

Yuki-Inoue:
https://www.emacswiki.org/emacs/elisp-format.el

I'm an user of this package.


purcell:
Thanks.

PR #3827, opened 2016-04-29
Add recipe for yankpad

Kungsgeten:
[Yankpad](https://github.com/Kungsgeten/yankpad) is a way to insert text snippets from an org-mode file. The snippets may use [Yasnippet](https://github.com/capitaomorte/yasnippet) syntax, if the user has Yasnippet installed, but it is not a requirement. I'm the creator and maintainer.

I get a warning from flycheck: `yas-expand-snippet is not known to be defined`. Perhaps I've missed something about how to require packages without actually having them as a dependency.

### DIFF
--- a/recipes/steam
+++ b/recipes/steam
@@ -1,0 +1,1 @@
+(steam :repo "Kungsgeten/steam.el" :fetcher github)
--- a/recipes/selected
+++ b/recipes/selected
@@ -1,0 +1,1 @@
+(selected :repo "Kungsgeten/selected.el" :fetcher github)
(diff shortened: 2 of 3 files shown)


purcell:
Thanks, looks good, but the `steam` and `selected` recipes are also in this PR in its current form, so I can't easily merge the `yankpad` recipe. Can you fix that up?

recipes/steam
```diff
@@ -0,0 +1 @@
+(steam :repo "Kungsgeten/steam.el" :fetcher github)
```
Kungsgeten: M

PR #3828, opened 2016-04-29
elpy: Remove release branch

jorgenschaefer:
Elpy development has stabilized sufficiently so that average users can (mostly) safely use the master branch without causing unnecessary workload to the maintainers. Also, MELPA Stable allows users to pick a stable version if they so choose.


purcell:
Super, thanks Jorgen!

PR #3829, opened 2016-04-30
Update the location for the geben recipe source

ahungry:
I am not the original geben author, but I will now be the author/maintainer of future work on the repository.

Geben is already in the MELPA repository, as our issue discussed, just updating to a branch that will not be obsolete when Google Code disappears.


purcell:
As per the travis failure, you'll just need to update the `geben` recipe so it doesn't describe itself as `geben-helm-projectile`. ;-)

ahungry:
Oh wow, that's what I get doing it after 2am at night.

Thanks!

> As per the travis failure, you'll just need to update the `geben` recipe so it doesn't describe itself as `geben-helm-projectile`. ;-)
>
> ---
> https://github.com/melpa/melpa/pull/3829#issuecomment-216005361

-- 
Matthew Carter (m@ahungry.com)
http://ahungry.com

purcell:
Awesome, thanks.

PR #3831, opened 2016-05-01
Add moonscript recipe

lassik:
Hi,

Here's a recipe that adds a major mode for the MoonScript language (a Lua preprocessor similar to CoffeeScript)

BR,
Lassi

### DIFF
--- a/recipes/moonscript
+++ b/recipes/moonscript
@@ -1,0 +1,3 @@
+(moonscript
+ :fetcher github
+ :repo "k2052/moonscript-mode")


lassik:
Ping @k2052

purcell:
Why exclude `moonscriptrepl-mode.el` from this? I'd be inclined to omit the `:files` spec here and just allow the default `*.el` to apply.

Re. the main code:

- A better way to disable `electric-indent-mode` is to set `electric-indent-inhibit` to mark the mode as incompatible with `electric-indent`.
- `font-lock-defaults` is always buffer-local, so `setq` is fine: no `make-local-variable` is necessary
- If you're happy to depend on Emacs 24, `prog-mode` would be a much better parent mode than `fundamental-mode`.

lassik:
Thank you for these suggestions @purcell! I fixed the issues and sent pull requests to @k2052.

The REPL mode was excluded because it is extremely rudimentary (even the REPL itself isn't an official MoonScript REPL, but a little hack posted on a wiki). Now that you asked about it I cleaned it up a bit too, in an attempt to make it consistent with the main mode, so I guess we might as well include it. The original plan was to add it in later once it has matured, or to ship it as a separate package.

lassik:
Fixes now live at https://github.com/k2052/moonscript-mode

I also modified this PR (amend + force-push) to use the default `:files` regexp so the REPL mode is now in.

purcell:
Super, thanks!

PR #3832, opened 2016-05-01
starting to phase out scala-mode2

fommil:
`scala-mode2` has been the de facto standard `scala-mode` for over 3 years now and it is confusing some people.

This is starting the process of phasing out `scala-mode2` and moving to `scala-mode`. Each step needs to wait for a melpa dist.

1. make melpa's `scala-mode2` build off a branch (this PR)
2. swap melpa's definition of `scala-mode` to (what is currently) `scala-mode2`
3. change `scala-mode2` to be a zero content package that depends on `scala-mode`
4. change `ensime` to depend on `scala-mode`, tell everybody to drop their direct dependency on `scala-mode2`
5. should we delete the `scala-mode2` recipe? I would hope it will be preserved in melpa `stable` but be deleted in the dev / unstable repo. We could do the same for `scala-outline-popup` (which is now split into functionality in ensime and the awesome `popup-imenu`).

@reactormonk would you consider moving your `ob-scala` into `scala-mode` as a file with optional dependencies? We do this in `ensime`, the functionality simply wouldn't be available if `ob` is not installed, but `scala-mode` wouldn't depend on `ob`.

FYI (no actions needed, I'm on it :smile:): @hvesalai @milkypostman @bradwright

### DIFF
--- a/recipes/scala-mode2
+++ b/recipes/scala-mode2
@@ -1,3 +1,4 @@
 (scala-mode2
  :repo "ensime/emacs-scala-mode"
+ :branch "scala-mode2"
  :fetcher github)


hvesalai:
I approve of this message

purcell:
> should we delete the scala-mode2 recipe? I would hope it will be preserved in melpa stable but be deleted in the dev / unstable repo.

If you delete the recipe, the package will disappear from MELPA Stable. So you might consider a plan which preserves the `scala-mode2` repo, at least for a while.

purcell:
(In general, we'd advise just getting as directly as possible to the final packaging state, and removing obsolete recipes and packages: we really can't rely on people receiving and installing any incremental updates we might introduce to ease their transition.)

fommil:
ok, we'll leave `scala-mode2` alive as a dummy package for a while. We're ok for this to be merged when you're ready. I'll send more in phases.

purcell:
Super, thanks.

fommil:
ok, it is done! I staged a change to `scala-mode2` so that it just depends on `scala-mode`.

@purcell how can we make a final **stable** release of `scala-mode2` which just delegates to `scala-mode`? It lives on a different branch but on the same repo as `scala-mode`, so I can't tag it.

Is there any chance you could just delete `scala-mode2`?

purcell:
> @purcell how can we make a final stable release of scala-mode2 which just delegates to scala-mode? It lives on a different branch but on the same repo as scala-mode, so I can't tag it.

What do you mean by "delegates"? You mean a "final" scala-mode2 which will tell people to install scala-mode instead? Maybe you need to create a fork and a separate recipe? It would also imply that scala-mode2 would remain in the main MELPA repo. Feels like a lot of hassle to me.

fommil:
yeah, deleting is much simpler. PR to do just that.

purcell:
Awesome, thanks for methodically working through all these changes.

fommil:
no problem! It was confusing the hell out of new users :smile:

PR #3834, opened 2016-05-02
Add projectile-hanami

avdgaag:
[Projectile Hanami](https://github.com/avdgaag/projectile-hanami) is an Emacs minor mode, based on Projectile, for navigating projects using the Hanami web framework. With Projectile Hanami, you can:

* navigate through entities, repositories, actions, views and templates are
  different apps in your project;
* run `hanami server`
* run `hanami console`
* run Rake tasks

Projectile Hanami is based on Projectile Rails, but is not a complete port of all its functionality. I am the author of the package.

### DIFF
--- a/recipes/projectile-hanami
+++ b/recipes/projectile-hanami
@@ -1,0 +1,2 @@
+(projectile-hanami :repo "avdgaag/projectile-hanami"
+                   :fetcher github)


purcell:
Thanks, looks useful.

Quick feedback:
-  "C-c e" is [reserved for end-user customisation](https://www.gnu.org/software/emacs/manual/html_node/elisp/Key-Binding-Conventions.html), so you shouldn't use it as a default in your code.
- Check out the `regexp-opt` function for cleaning up the `(?:create\\|identical\\|...` regexps.
- You should add a `defgroup` form to define the customisation group and place it helpfully in the customisation hierarchy.

avdgaag:
Thanks for the feedback!

> "C-c e" is reserved for end-user customisation, so you shouldn't use it as a default in your code.

I’m following the example set by Projectile (`C-c p`), Projectile Rails (`C-c r`) and Alchemist (`C-c a`), so I reckon using something other than C-c [letter] will be _more_ confusing. I prefer to leave it as is.

> Check out the regexp-opt function for cleaning up the (?:create\\|identical\\|... regexps.

I’m not bothered by a regular expression literal. Emacs regex syntax is, well, unfortunate, but using a list of strings instead, and transforming that to get to what’s essentially a simple (if slightly long) regex seems like cleverness for cleverness’ sake to me. I prefer to leave this as is (again, much like Projectile Rails already does).

> You should add a defgroup form to define the customisation group and place it helpfully in the customisation hierarchy.

Good point, added it!

purcell:
> I’m not bothered by a regular expression literal.

Fair enough.

> I’m following the example set by Projectile (C-c p), Projectile Rails (C-c r) and Alchemist (C-c a), so I reckon using something other than C-c [letter] will be more confusing. I prefer to leave it as is.

They're wrong too, and I think it's not great for the community to ignore the guidelines that have been designed so that everyone follows the same principles. I'd encourage you to use a different default prefix, and then if users want to bind their "C-c e" to your keymap, then that's their choice.

I maintain Rinari, and that uses `C-c ;` (and `C-c '` !) by default, which are reserved for minor modes like this. Why not use the same for `hanami`? I don't think anyone would use `hanami` *and* `rinari`?

avdgaag:
> They're wrong too, and I think it's not great for the community to ignore the guidelines that have been designed so that everyone follows the same principles. I'd encourage you to use a different default prefix, and then if users want to bind their "C-c e" to your keymap, then that's their choice.

Alright, let’s do that, then! To me, it feels the real convention is to do actually use `C-c [letter]` for this kind of thing, but it’s not really that important to me. `C-c ;` it is!

purcell:
(Related: I just filed https://github.com/tonini/alchemist.el/issues/229. Hopefully we can clear up some of the confusion upstream, because it's natural to look at high-profile codebases to see how they do things, and that makes it even more important for them to do the right thing.)

purcell:
Awesome, merged - thanks! I'm going to have to try this out myself. :-)

avdgaag:
Great, thanks!

PR #3835, opened 2016-05-02
Add paganini-theme

onurtemizkan:
Hi, 
Here is [paganini](https://github.com/onurtemizkan/paganini), a new dark theme for Emacs :smile:

### DIFF
--- a/recipes/paganini-theme
+++ b/recipes/paganini-theme
@@ -1,0 +1,1 @@
+(paganini-theme :repo "onurtemizkan/paganini" :fetcher github)


purcell:
Thanks. Please remove the `:files` spec from the recipe: it's not required in the common case where you want to include all the `.el` files in the repo.

Also, this unsafe local variable setting (which you've probably copied from a theme elsewhere) should be removed:

```el
;; eval: (when (fboundp 'rainbow-mode) (rainbow-mode +1))
```

If you like using `rainbow-mode`, you can get the same effect with a [snippet like this](https://github.com/purcell/emacs.d/blob/master/lisp/init-lisp.el#L257-L262) in your emacs config.

onurtemizkan:
Thanks for the feedback, I'm just removing `rainbow-mode` configuration for now. :smile:

purcell:
Great, thanks!

PR #3837, opened 2016-05-02
systemd: add *.txt files to recipe

holomorph:
i'm storing byte-compile-time data in text files now, as it is much less painful to manage that way


purcell:
Thanks!

PR #3838, opened 2016-05-02
Add Emacs interface for rope, Python refactoring library.

mcepl:


purcell:
This recipe won't work at all, because it won't include any files. Please test it locally.

There will need to be a `:files` spec to include the necessary files, and also some elisp in the repo to provide at least the minimal package metadata. This might be a `ropemacs-pkg.el` package descriptor containing the corresponding `define-package` form, and presumably also declaring a dependency on `pymacs`.

purcell:
Closing because no response.

PR #3840, opened 2016-05-03
Add recipe `helm-qiita`

masutaka:
I want to add [`helm-qiita.el`](https://github.com/masutaka/emacs-helm-qiita). This is a helm interface for [Qiita](https://qiita.com/).

Qiita is a Japanese web service like a little http://stackoverflow.com/ .

Thanks :sparkles:

### DIFF
--- a/recipes/helm-qiita
+++ b/recipes/helm-qiita
@@ -1,0 +1,1 @@
+(helm-qiita :repo "masutaka/emacs-helm-qiita" :fetcher github)


purcell:
Thanks. Quick feedback:
- Please depend on the latest stable version of helm, not the snapshot version
- The `helm-qiita:` symbol prefix should be `helm-qiita-` in order to conform to the [elisp conventions](https://www.gnu.org/s/emacs/manual/html_node/elisp/Coding-Conventions.html)

masutaka:
@purcell 

Thanks for your review. I've fixed.
https://github.com/masutaka/emacs-helm-qiita/blob/master/helm-qiita.el

purcell:
Perfect, thanks! Merged. :-)

masutaka:
Thanks a lot!

PR #3842, opened 2016-05-04
add recipe ob-php

stardiviner:
https://github.com/stardiviner/ob-php

### DIFF
--- a/recipes/ob-php
+++ b/recipes/ob-php
@@ -1,0 +1,1 @@
+(ob-php :fetcher github :repo "stardiviner/ob-php")


stardiviner:
Oops, the package `s` always want to be dead sometimes.

purcell:
MELPA was a bit broken at the time of the travis build, so don't worry about that. :-)

The package description needs to be fixed, because it says: "Execute Redis queries within org-mode blocks."

stardiviner:
@purcell ok, fixed, sorry for my reckless.

purcell:
Great, thanks.

PR #3847, opened 2016-05-04
Rename emms-status to emms-state

alezost:
As pointed at https://github.com/alezost/emms-state.el/issues/1
there is another package that provides `emms-status` feature, so
I renamed this one `emms-state`.

Thanks and sorry for worries :-)


purcell:
Thanks

PR #3849, opened 2016-05-04
evil-lispy: don't specify

tarsius:
The specified "master" branch is the "HEAD branch".  It also is
the only branch.  So it is not necessary to explicitly specify it.


purcell:
Thanks

tarsius:
Hm, the commit message summary seems a bit incomplete ;-)

PR #3850, opened 2016-05-04
get elisp-format directly from the Emacswiki

tarsius:
The repository on the Emacsmirror is created from the file from
the Emacswiki.  Packages from the Emacswiki are only infrequently
updated on the Emacsmirror, so it is better if Melpa gets those
packages directly from the Emacswiki.


purcell:
Thanks Jonas.

This PR is superseded by #3870, in which a new maintainer is now maintaining a github-based fork of this package.

tarsius:
Better still ;-)

PR #3851, opened 2016-05-04
Add coverlay

kytwb:
Test coverage overlay for Emacs

### DIFF
--- a/recipes/coverlay
+++ b/recipes/coverlay
@@ -1,1 +1,1 @@
-(coverlay :repo "FossiFoo/coverlay.el" :fetcher github)
+(coverlay :repo "twada/coverlay.el" :fetcher github)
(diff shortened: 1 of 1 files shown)


FossiFoo:
nice. i meant to do this for a while.

purcell:
Thanks. I sent a small fix in https://github.com/twada/coverlay.el/pull/8, but the library also uses the `C-c l` keybinding, which is [reserved for user customisation](https://www.gnu.org/software/emacs/manual/html_node/elisp/Key-Binding-Conventions.html). Would you mind suggesting a new binding to the maintainer?

kytwb:
@purcell the keybindings have been updated in this merged pull-request https://github.com/twada/coverlay.el/pull/9

purcell:
Super, thanks!

PR #3852, opened 2016-05-04
Add ycmd-eldoc

ptrv:


purcell:
`ycmd-eldoc` doesn't have any further package dependencies, and so it should be part of the `ycmd` package, not a separate package. Maybe adjust the `:files` in the `ycmd` spec instead.

ptrv:
Ok, I see. I'll adjust the `ycmd` recipe

PR #3857, opened 2016-05-05
Add new recipe for my library

eric-hansen:
**A brief summary of what the package does.**
Provides a way to natively display notifications on a Mac system (i.e.: doesn't use a third party program or library)

**A direct link to the package repository.**
https://github.com/eric-hansen/mac-notifier

**Your association with the package (e.g., are you the maintainer? have you contributed? do you just like the package a lot?).**
Author and maintainer

### DIFF
--- a/recipes/mac-notifier
+++ b/recipes/mac-notifier
@@ -1,0 +1,1 @@
+(mac-notifier :fetcher github :repo "eric-hansen/mac-notifier")


purcell:
Thanks. This is a nice alternative method to the `terminal-notifier` approach, but the best thing for users IMO would be to incorporate this into the [alert package](https://github.com/jwiegley/alert), which is increasingly the standard way to handle alerts. `alert` has sensible backends for various platforms, so that users don't need to start installing or coding against system-specific packages. So I'd suggest you send a PR upstream adding an `osascript` variant that a wider user base would then have access to.

I know this is only a small package, but I've seen various little notification libraries come and go over the years, each of which has generally been platform-specific and full-featured, and at a certain point it's really hard for users to know which of all the similarly-named packages is going to give them what they want.

eric-hansen:
If you want this closed feel free.  I submitted a PR to the package you suggested with my changes and he accepted it so it does kind of render this PR moot now.  Thanks for the words, one day I hope to get a MELPA package in though. :D

purcell:
> If you want this closed feel free. I submitted a PR to the package you suggested with my changes and he accepted it so it does kind of render this PR moot now. Thanks for the words, one day I hope to get a MELPA package in though. :D

I think that's already a really helpful contribution, particularly if `alert` is eventually merged into Emacs, so don't worry too much. I look forward to your next submission. :-)

PR #3858, opened 2016-05-05
add recipe for flyspell-correct

d12frosted:
I hope I didn't miss anything from `CONTRIBUTING.md`.

### DIFF
--- a/recipes/flyspell-correct
+++ b/recipes/flyspell-correct
@@ -1,0 +1,1 @@
+(flyspell-correct :repo "d12frosted/flyspell-correct" :fetcher github)


purcell:
Thanks. Quick feedback:
- Don't start your package description with "package for" (and you can't line-wrap it either)
- Please declare a dependency on flycheck using a `Package-Requires` header
- Consider using `cl-lib` instead of `cl`. If you're only using it for `lexical-let`, you can simply declare `lexical-binding` for the file, then the lambda will bind lexically anyway.
- If you're going to ship code which uses `ivy`, depend on `ivy` too (and require it, at least at compile time)
- Don't use `/` or `//` to mark the prefix of symbol names - please follow the [elisp conventions](https://www.gnu.org/s/emacs/manual/html_node/elisp/Coding-Conventions.html) and use `-` or `--`
- You might be interested in https://github.com/flycheck/flycheck/issues/530

d12frosted:
@purcell wow, thank you very much for your feedback, didn't expect it :) 

> Don't start your package description with "package for" (and you can't line-wrap it either)

Done.

> Please declare a dependency on flycheck using a Package-Requires header

Am I missing something or this package doesn't depend on `flyckeck`, but on `flyspell`, which is AFAIK part of Emacs?

> Consider using `cl-lib` instead of `cl`. If you're only using it for `lexical-let`, you can simply declare `lexical-binding` for the file, then the lambda will bind lexically anyway.

What do you mean? Are you talking about doing following steps?

1. Add ` ;;; -*- lexical-binding: t -*-` to `flyspell-correct.el`.
2. Converting `lexical-let` to regular `let`?

If this is what you meant - done :) 

> Maybe use `declare-function` to declare the ivy and helm functions.

Done. Just can't figure out why `&key` is not working in arguments list. So I've changed it to `&optional`. Also, what is best way to avoid warning about reference to a free variable `helm-pattern` in this case? Using `defvar` doesn't sound right. 

> In general, though, it's bad practice to include code which uses functions from package you're not formally depending on or require-ing.

Yes, I understand. This package has 'dynamic' dependencies :) 

> Don't use `/` or `//` to mark the prefix of symbol names - please follow the elisp conventions and use `-` or `--`

Done. And thank you for the link - I'll take a deeper look.

purcell:
> Am I missing something or this package doesn't depend on flyckeck, but on flyspell, which is AFAIK part of Emacs?

Sorry, that's me being stupid in the middele of a bunch of stuff.

> If this is what you meant - done :)

Yes, exactly.

Merging now - thanks!

purcell:
No idea why I read this as `flycheck-correct`... must have been wishful thinking, I guess, hence the link I posted to that flycheck issue. Apologies for the confusion.

d12frosted:
The funny thing, when I was creating recipe I named it `flycheck-correct` by accident. Then I corrected it (without `flyspell-correct`, that's a shame) :) 

One again, thank you for your review :) And thanks for merging!

PR #3859, opened 2016-05-05
Add darkokai-theme recipe.

sjrmanning:
Recipe for the theme found [here](https://github.com/sjrmanning/darkokai).


purcell:
Thanks, looks good. Sent you a suggestion in https://github.com/sjrmanning/darkokai/pull/1.

PR #3861, opened 2016-05-05
add ksp-cfg-mode recipe

lashtear:
KSP-cfg-mode is a major-mode for editing the .CFG files of Kerbal Space Program and ModuleManager.  It includes syntax highlighting, intelligent indentation, and context-sensitive help, all of which are customizable.

The authoritative repository is on github, at:  https://github.com/lashtear/ksp-cfg-mode

I am the original author of this package.

### DIFF
--- a/recipes/ksp-cfg-mode
+++ b/recipes/ksp-cfg-mode
@@ -1,0 +1,3 @@
+(ksp-cfg-mode
+ :fetcher github
+ :repo "lashtear/ksp-cfg-mode")


purcell:
Thanks, that's a nice polished piece of work. Quick feedback:

- `font-lock-defaults` is always buffer-local, so `setq` will suffice
- There should be no need to call `(font-lock-fontify-buffer)` explicitly. In fact, you shouldn't technically assume that font-lock-mode will be enabled.
- If the only features of Emacs 24.3 you need is `cl-lib` and `setq-local`, then you might consider instead depending on `cl-lib` 0.5, which works even in older emacsen, and the classic `(set (make-local-variable 'name) value)` pattern.

I'll go ahead and merge the recipe, since none of the above are blockers, but they're worth a look.

lashtear:
Thanks!  I'll take a look and see about making those improvements.

PR #3862, opened 2016-05-06
dix-evil

unhammer:
Summary: evil integration for dix (dix is a minor mode under nxml for
editing Apertium (machine translation) XML dictionaries and rules files)

Link: https://github.com/unhammer/dix

Association: Author.

(This file is rather trivial, but putting it in the main dix package
would make that depend on Evil, which most users probably wouldn't
want.)


purcell:
Make sense - thanks.

PR #3864, opened 2016-05-06
Recipe: Add dired-quick-sort.

xuhdev:
- A brief summary of what the package does.

Persistent quick sorting of Dired buffers in various ways with hydra.

- A direct link to the package repository.

https://gitlab.com/xuhdev/dired-quick-sort

- Your association with the package (e.g., are you the maintainer? have you contributed? do you just like the package a lot?).

I'm the author

### DIFF
--- a/recipes/dired-quick-sort
+++ b/recipes/dired-quick-sort
@@ -1,0 +1,3 @@
+(dired-quick-sort
+ :repo "xuhdev/dired-quick-sort"
+ :fetcher gitlab)


purcell:
Thanks.

While it's okay (and recommended) to put `lexical-binding: t` on the first line, you should move the other settings there to a `Local Variables` block:

```
;; Local Variables:
;; fill-column: 80
;; coding: utf-8
;; indent-tabs-mode: nil
;; sentence-end-double-space: t
;; End:
```

xuhdev:
Thanks for your suggestion! Now fixed.

PR #3866, opened 2016-05-07
Add spu recipes

mola-T:
Hello,

SPU stands for Silent Package Upgrader.

https://github.com/mola-T/SPU

There are numerous package for upgrading packages. However, I have not met one that can suppress the freezing during `Contacting Host: elpa.gnus.org:80` and `Contacting Host: melpa.org:433`. The process takes about 5 to 10 seconds. I always believe from the view of the end user, the ui freezing for 1 second is long, 2 senconds is too long, longer than 3 seconds is unacceptable. So I write this package.

### DIFF
--- a/recipes/spu
+++ b/recipes/spu
@@ -1,0 +1,1 @@
+(spu :fetcher github :repo "mola-T/spu")


purcell:
Thanks.

PR #3867, opened 2016-05-07
Remove flycheck-typescript-tslint package

Simplify:
flycheck-typescript-tslint is now part of Flycheck package.


purcell:
Super, thanks.

PR #3869, opened 2016-05-08
Update import-js recipe with new repo

kevinkehl:
ImportJS has moved to its own organization. Additionally, editor plugins
for the (now node-js based) library have been broken out into separate
repos, see:
https://github.com/Galooshi/import-js/commit/b33acfffaf0c119a9b2f35abdabda5fbb059d693

Future versions should track the new repo at
https://github.com/Galooshi/emacs-import-js


purcell:
Thanks.

PR #3870, opened 2016-05-08
Change source repo of elisp-format

Yuki-Inoue:
Change the source repository of elisp-format for following 2 reasons.

1. I think it is more desirable to maintain the code in a github repo instead of (as an mirror of) emacswiki.
2. Want to make modification to the original source code (namely, add autoloads magic comments).

The original code was released at EmacsWiki with GPL, so I becoming a maintainer of this code for MELPA is OK, I think.


purcell:
Sounds good to me - thank you!

PR #3871, opened 2016-05-08
added 'quiet' package

zzkt:


purcell:
Thanks! Nice idea. Quick feedback:

- The `(provide ...)` form should come *after* the rest of the code, not before it.
- You might like to add `;;;###autoload` cookie comments for the two main commands
- I wonder if `disconnect` might be a more descriptive alternate name for this package?

zzkt:
i've moved the provide (thanks for pointing it out) and added a couple of autoloads...

purcell:
Thanks.

PR #3872, opened 2016-05-08
Add ox-bibtex-chinese recipe

tumashu:


purcell:
Thanks. You should change the "\C-cb" because that is [reserved for end-user customisation](https://www.gnu.org/software/emacs/manual/html_node/elisp/Key-Binding-Conventions.html). Otherwise looks good - merging now.

tumashu:
Fixed, thank you

At 2016-05-10 03:27:39, "Steve Purcell" <notifications@github.com> wrote:


Thanks. You should change the "\C-cb" because that is reserved for end-user customisation. Otherwise looks good - merging now.

PR #3876, opened 2016-05-09
Add reek recipe

hanmoi-choi:
I have add minor mode for a simple interface of Ruby Gem Reek, https://github.com/troessner/reek

### DIFF
--- a/recipes/reek
+++ b/recipes/reek
@@ -1,0 +1,1 @@
+(reek :fetcher github :repo "hanmoi-choi/reek-emacs")


purcell:
Thanks. Quick feedback:
- You should change the `C-c C-e` prefix, because that is reserved for major modes: see https://www.gnu.org/software/emacs/manual/html_node/elisp/Key-Binding-Conventions.html
- The tramp filename handling looks unusual to me. Is it really necessary? Normally tramp just does the right thing without needing any code.
- Please remove "Emacs" from the package description: it's redundant because all elisp packages are for Emacs. Maybe use "An interface for the Reek ruby linter" or similar.

Also, did you know that `flycheck` has support for `reek`?

hanmoi-choi:
Hi Purcell,

Thank you for your kind and meaningful comment.

Indeed, I thought that it would be great if `flycheck` support reek but I could not find any information from its github page. Could you share info or link about that?

I will fix other things.

purcell:
> Indeed, I thought that it would be great if flycheck support reek but I could not find any information from its github page. Could you share info or link about that?

There's an in-progress pull request here: https://github.com/flycheck/flycheck/pull/886

tarsius:
@hanmoi-choi what's the current status of this? Do you still want to get this into Melpa?

tarsius:
Okay, thanks.

PR #3877, opened 2016-05-09
add recipe emoticon 

TsubasaWp:
Hi, I want to add recipe emoticon from https://github.com/TsubasaWp/Emoticon, thx!

### DIFF
--- a/recipes/emoticon
+++ b/recipes/emoticon
@@ -1,0 +1,2 @@
+(emoticon :repo "TsubasaWp/Emoticon"
+            :fetcher github)


purcell:
Sorry, that's not a well-formed package: you don't have any of the necessary header or footer boilerplate or metadata. Also, since this is for `helm`, it should normally be called `helm-emoticon`. But actually, I feel like there must already be a helm source for emoji... Worth having a look around, I think.

TsubasaWp:
Thanks for the comment :)

PR #3878, opened 2016-05-09
Add ox-bibtex-chinese recipe

tumashu:
This pull request is real ox-bibtex-chinese,

### DIFF
--- a/recipes/org-chinese-utils
+++ b/recipes/org-chinese-utils
@@ -1,0 +1,1 @@
+(org-chinese-utils :fetcher github :repo "tumashu/org-chinese-utils")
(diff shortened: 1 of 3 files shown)


purcell:
Is `org-plus-contrib` required to use `ox-bibtex`? Normally it wouldn't be good practice to depend on `org-plus-contrib` because we try to make sure all MELPA packages will work with only MELPA and GNU ELPA configured as package sources, and `org-plus-contrib` comes from the Org package repository.

tumashu:
Fixed, thanks very much. https://github.com/tumashu/ox-bibtex-chinese/commit/405fee92e41468a361ac5beff4a8d2dc4c895b73

purcell:
Thanks.

PR #3881, opened 2016-05-10
Add say-what-im-doing recipe

Benaiah:
This package makes emacs say the commands you invoke out loud, using text-to-speech.

### DIFF
--- a/recipes/say-what-im-doing
+++ b/recipes/say-what-im-doing
@@ -1,0 +1,1 @@
+(say-what-im-doing :repo "benaiah/say-what-im-doing" :fetcher github)


purcell:
Thanks. You need to put a package description after the `---` on the first line, and you should also add an `;;;###autoload` cookie comment for the `define-minor-mode` form.

purcell:
(Also, running `M-x checkdoc` and fixing the errors reported would be a very good idea.)

Benaiah:
Ok, I've fixed the issues you brought up and corrected the checkdoc errors (there's technically one checkdoc error left, but it's because it thinks the phrase command-line in a docstring is a symbol, when it's just referring to the command line).

purcell:
Great, thanks.

PR #3882, opened 2016-05-11
Add flymd recipes

mola-T:
https://github.com/mola-T/flymd

flymd is a on the fly markdown preview package.

### DIFF
--- a/recipes/flymd
+++ b/recipes/flymd
@@ -1,3 +1,3 @@
 (flymd :fetcher github
        :repo "mola-T/flymd"
-       :files ("flymd.el" "flymd.html"))
+       :defaults (:defaults "flymd.html"))
--- a/recipes/flymd
+++ b/recipes/flymd
@@ -1,3 +1,3 @@
 (flymd :fetcher github
        :repo "mola-T/flymd"
-       :defaults (:defaults "flymd.html"))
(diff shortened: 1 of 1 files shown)


milkypostman:
please change files to `(:defaults "flymd.html")`

how does this compare with markdown-mode? it seems like with markdown mode all you need is an on-save hook. also there is some timer function builtin to emacs I think where you can hookup a function call to a timer.

it seems like this might be better as a minor mode?

purcell:
> it seems like this might be better as a minor mode?

+1

mola-T:
@milkypostman 

1. How does this compare with markdown-mode?

   I love the `markdown-mode`. But it cannot fulfill my need. The way you suggested is not the best for me. New browser windows/tabs will be created upon function calls or the browser page needs to be refresh constantly. 

In fact, it is more close to https://github.com/ancane/markdown-preview-mode. But I also have problem using this.....

2. I don't implement this as a minor mode because it is overhead and not user-friendly.
 
   For example, if user want to bring up a live preview, they do `flymd-mode` and a browser page is brought up. If the user close the browser page accidentally, they can't call `flymd-mode` again to open the browse page. It will lead to another function like `flymd-open-browser` which I don't prefer.

purcell:
Thanks.

PR #3883, opened 2016-05-11
for store data as json file 

xcwen:


purcell:
Thanks!

The packaging for this really isn't optimal: it's a big pile of code which requires both `auto-complete` and `company`, while users will only ever want to have one of those installed. So this should be (at least) 2 separate packages, each of which depends on a specific completion framework. (The name pattern `ac-*` is really just shorthand for `auto-complete-*`, so anything intended for use with `company` shouldn't be named `ac-*`.)

xcwen:
ac-php-company.el  rename company-php.el

and remove commpay from Package-Requires?!

purcell:
> ac-php-company.el rename company-php.el
> 
> and remove commpay from Package-Requires?!

Yes: `company-php` should become a separate package, with its own `Package-Requires` header which lists `company` as a dependency.

xcwen:
ok , i remove company.

(ac-php :repo "xcwen/ac-php" :fetcher github :files ( "ac-*.el"  "phpctags" "*.json"))

Dragnucs:
@xcweb I update my packages this morning and completion stoped, because `ac-php` is no longer available. What should I do?

purcell:
We need a fresh recipe for `company-php`...

xcwen:
ok ..

etu:
@xcwen you're gonna create a recipie for `company-php` or should someone else?

PR #3884, opened 2016-05-12
add nlinum-relative

CodeFalling:
https://github.com/CodeFalling/nlinum-relative is relative line number besed on nlinum-mode


purcell:
Great, thanks.

CodeFalling:
Sorry, but why MELPA seems not build this packages with `linum.el` in it? See https://github.com/CodeFalling/nlinum-relative/issues/4

purcell:
I've answered on that issue. Thanks.

PR #3885, opened 2016-05-12
Add yankpad recipe.

Kungsgeten:
I made a pull request earlier for this, but I couldn't figure out how to remove unnecessary files from the pull request, so I made a new one. @purcell handed the last one.

Anyway, [yankpad](https://github.com/Kungsgeten/yankpad) is a way to paste text snippets from an org-mode file. It can use [yasnippet](https://github.com/capitaomorte/yasnippet) and/or [projectile](https://github.com/bbatsov/projectile), but neither are a requirement.


purcell:
Thanks.

PR #3887, opened 2016-05-12
Recipe for Helm integration with phpunit.el

eric-hansen:
**A brief summary of what the package does.**
Lets the user choose which function to run a test on via PHPUnit by selecting the function in a Helm buffer.

**A direct link to the package repository.**
http://github.com/eric-hansen/phpunit-helm

**Your association with the package (e.g., are you the maintainer? have you contributed? do you just like the package a lot?).**
I contributed to the original phpunit.el package which caused the author to want the Helm integration removed out of my contrib.  So, I created this instead.

*Relevant communications with the upstream package maintainer (e.g., package.el compatibility changes that you have submitted).*
PR where this all started from: https://github.com/nlamirault/phpunit.el/pull/16


milkypostman:
how hard to rename to helm-phpunit? because i think that would fit better with our existing packages. while there are some with a 'helm' postfix, I think the other way is better.

purcell:
Agreed.

eric-hansen:
Easy pleasy, done! :)

purcell:
Thanks!

- Best to depend on the latest stable version of helm, not "0": http://stable.melpa.org/#/helm
- As per the travis failure, the recipe filename must be `helm-phpunit`, not `helm-phpunit.el`
- The function `phpunit-selected-test` should be renamed to `helm-phpunit-selected-test`, so that you have a consistent namespace for top-level symbols.

eric-hansen:
Made the changes; sorry for the filename, was still waking up hah.

milkypostman:
ty

PR #3890, opened 2016-05-13
Add aws-ec2 package

Yuki-Inoue:
Emacs interface to AWS, only supported ec2 for now.
I'm author.


purcell:
Thanks. Sent you a small fix in https://github.com/Yuki-Inoue/aws.el/pull/2.

PR #3891, opened 2016-05-13
Added 'resources' folder

ferfebles:
Added resources folder to the package, to store sounds used when the pomodoro is running.

Recipe tested with 'package-build-current-recipe': package is downloaded and installed correctly.


milkypostman:
thanks

PR #3892, opened 2016-05-14
added viking-mode

TLINDEN:
viking-mode is a small minor mode which allows to quickly kill (or delete) things at point. Repeat the key multiple times to delete more and more things (word, line, paragraph, buffer).

Repo is: https://github.com/TLINDEN/viking-mode

I am the creator and maintainer, reachable via email (contained in package header).

### DIFF
--- a/recipes/viking-mode
+++ b/recipes/viking-mode
@@ -1,0 +1,2 @@
+(viking-mode :repo "tlinden/viking-mode" :fetcher github)
+


milkypostman:
internal variables should be prefixed with `viking--` instead of `vk/`. the former is emacs lisp convention and the latter is more of a clojure thing afaik.

you shouldn't need `turn-on-viking-mode` because `viking-mode` will do this fine.

TLINDEN:
@milkypostman: just that I understand it right: you suggest to rename internal ```(defun vk/``` to ```(defun viking--``` and the public ones to ```(defun viking-``` (one dash)?

purcell:
Yep, that's right, as per the [official elisp conventions](https://www.gnu.org/software/emacs/manual/html_node/elisp/Coding-Conventions.html). :-)

TLINDEN:
Ok, I renamed them. Better now?

purcell:
Great, thanks. You should add a `defgroup` form, by the way, to explicitly declare that customization group and place it in the overall customization hierarchy.

TLINDEN:
Thanks for merging. I added the '''defgroup''' form, but there's still a problem though: If I Do M-x customize-group viking<TAB> then emacs shows 2 choices: viking and viking-mode. If I select the former, I see this:
```
Viking group:  Group definition missing. 
       State : visible group members are all at standard values.
   
Show Value Viking Global Mode 
   Non-nil if Viking-Global mode is enabled. More
```
Could you help me fixing this as well? That would be great,


Thanks,
Tom

purcell:
Maybe you have an old group definition hanging around in your Emacs runtime for `viking`. What if you restart?

TLINDEN:
Yes, multiple times since that post :)

However, I found the culprit:
```
;; just in case someone wants to use it globally
(define-global-minor-mode viking-global-mode
  viking-mode
  (lambda () (viking-mode t)
    ))
```
This definition didn't contain a ```:group```  keyword, therefore a default group has been generated, it seems. So, this fixes this:
```
(define-global-minor-mode viking-global-mode
  viking-mode
  (lambda () (viking-mode t))
  :group 'viking-mode
  )
```


best,
Tom

purcell:
Try using `define-globalized-minor-mode` instead, by the way. I think it will allow you to remove the lambda.

PR #3894, opened 2016-05-15
Add judge-indent package

yascentur:
Please add a new package: judge-indent.


purcell:
Nice, thanks. You should add an `;;;###autoload` cookie for the mode definition form.

yascentur:
Thanks. Autoload cookie is added.

PR #3896, opened 2016-05-15
Added modern-cpp-font-lock

ludwigpacifici:


purcell:
Thanks. Quick feedback:
- All the top-level symbols should use the package prefix, so `string-length<` should be `modern-c++--string-length<` or similar, as per the [elisp conventions](https://www.gnu.org/s/emacs/manual/html_node/elisp/Coding-Conventions.html)
- The `clear memory` [section](https://github.com/ludwigpacifici/modern-cpp-font-lock/blob/master/modern-cpp-font-lock.el#L238-L243) has no reasonable purpose, and is unidiomatic: either don't worry about that little bit of memory, or do things the lisp way and `(let ...)` those variables across the scope in which you use them, so that they will be garbage collected.

ludwigpacifici:
Hello @purcell, thanks for the review.

 * Done with commit `900eee7416ed1c32c6f0329f7bffd395d5fe363e`.
 * Done the lisp way with commit `d5f12b1a5ffd55466d79953a790381c27ef7db35`. For information, the `clear memory` is suggested in this tutorial [here](http://ergoemacs.org/emacs/elisp_syntax_coloring.html).

Tell me if you have other remarks.

purcell:
Nice, thanks.

PR #3897, opened 2016-05-16
Add fastdef recipe

redguardtoo:
fastdef insert terminology from google top search results.

### DIFF
--- a/recipes/fastdef
+++ b/recipes/fastdef
@@ -1,0 +1,1 @@
+(fastdef :fetcher github :repo "redguardtoo/fastdef")


purcell:
Thanks. Some quick thoughts:
- Please depend on `ivy`, not `swiper`: see https://melpa.org/#/ivy
- The `:files` spec shouldn't be necessary here. Either delete the `-pkg.el` file, or include them both in the package, by removing the `:files` spec.
- `fastdef` isn't a very obvious name. Why not something like `google-define` or `search-define`?

redguardtoo:
thanks for head up. I update fastdef and recipe.
I'd rather keep the original name because I have plan to make fastdef have more ways to get the definition of a term.

purcell:
Okay. But now you need to delete the `-pkg.el` file: it's not needed by MELPA, but when present it overrides the metadata in the `fastdef.el` file, which means your package dependencies will disappear.

redguardtoo:
I deleted the *-pkg.el

purcell:
Great, thanks.

PR #3898, opened 2016-05-16
Add ivy-hydra, modify ivy

abo-abo:
Fixes https://github.com/abo-abo/swiper/issues/464

### DIFF
--- a/recipes/ivy
+++ b/recipes/ivy
@@ -1,5 +1,5 @@
 (ivy :repo "abo-abo/swiper"
      :fetcher github
      :files (:defaults
-             (:exclude "swiper.el" "counsel.el")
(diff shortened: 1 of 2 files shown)


purcell:
Thanks. The first element of the recipe form for `ivy-hyrda` should be `ivy-hydra`, but it's currently `ivy` -- that's the reason for the build failure. Please could you update that?

purcell:
<kbd>C-c C-c</kbd> lets you test these recipes locally, btw, which we recommend.

abo-abo:
> C-c C-c

Thanks, a very nice feature. I've force-pushed on the PR, should be correct now.

purcell:
Super, thank you.

PR #3899, opened 2016-05-17
Add cursor-in-brackets recipe

yascentur:
Please add a new recipe of cursor-in-brackets.

### DIFF
--- a/recipes/cursor-in-brackets
+++ b/recipes/cursor-in-brackets
@@ -1,0 +1,1 @@
+(cursor-in-brackets :fetcher github :repo "yascentur/cursor-in-brackets-el")


purcell:
Thanks. You can use `define-globalized-minor-mode` instead of `define-global-minor-mode` to avoid the lambda there.

yascentur:
Thanks for merging and reviewing.
According to describe-function in Emacs 24.5, define-global-minor-mode is an alias for \`define-globalized-minor-mode' in \`easy-mmode.el'.
It seems that the lambda cannot be avoided by using define-globalized-minor-mode instead of define-global-minor-mode.

purcell:
Oh, really? Okay. In any case, as of Emacs 24,
```el
(lambda ()
    (cursor-in-brackets-mode t))
```
is equivalent to `'cursor-in-brackets-mode`, so you only need the former if you want to support Emacs < 24.

yascentur:
OK. It is fixed.

PR #3900, opened 2016-05-17
ghost-mode recipe

javaguirre:
ghost-mode is a package to manage posts from your [Ghost](https://ghost.org) blog platform using its [Rest API](https://api.ghost.org/).

The package repository is in [github](https://github.com/javaguirre/ghost-mode).

Do I need to specify anything else?

Thank you! :-)

### DIFF
--- a/recipes/ghost-mode
+++ b/recipes/ghost-mode
@@ -1,0 +1,3 @@
+(ghost-mode :repo "javaguirre/ghost-mode"
+            :fetcher github
+            :files ("ghost-mode.el"))


purcell:
Thanks. Quick feedback:
- This doesn't provide a mode, so please call it something like `ghost` or `ghost-blog`, rather than `ghost-mode`.
- The advice around `url-http-handle-authentication` will not only affect requests made to the ghost API, so that is not good practice
- Consider changing some of the `defvar` to `defcustom`.
- Don't specify `:files` in the recipe please -- the default includes `"*.el"`, which is fine here.

Hope that helps!

javaguirre:
Thank you @purcell! I'll do that! :-)

javaguirre:
I'm gonna create a new PR with the new package name and your suggested changes, I'll close this one.

PR #3901, opened 2016-05-17
add immortal-scratch recipe

jpkotta:
This is a simple global minor mode that respawns the scratch buffer when it's killed, similar to just clearing the buffer.


purcell:
Handy, thanks. Surprised there's not a convenient built-in function for creating the scratch buffer though.

PR #3903, opened 2016-05-17
add highlight-fixme recipe

jpkotta:
A minor mode to highlight and search words like "FIXME" and "TODO".


purcell:
How does this differ from [hl-todo](https://github.com/tarsius/hl-todo)? I'd rather not add highly similar packages, because it's confusing for users.

jpkotta:
I didn't know about `hl-todo`; this is not much different.  I'd recommend not adding it to melpa.

purcell:
:+1: thanks

PR #3904, opened 2016-05-17
add recipe for highlight-operators

jpkotta:
A minor mode for syntax highlighting common operators like '+' and '&'.


purcell:
Thanks. I might be wrong, but IIRC `(called-interactively-p 'any)` doesn't work in older Emacsen, so you might need to add a `Package-Requires` dependency on a certain minimum Emacs version.

PR #3906, opened 2016-05-18
scala-mode2 is dead, long live scala-mode

fommil:
this deletes some packages that are now no longer used.

### DIFF
--- a/recipes/scala-mode2
+++ b/recipes/scala-mode2
@@ -1,4 +1,0 @@
-(scala-mode2
- :repo "ensime/emacs-scala-mode"
- :branch "scala-mode2"
- :fetcher github)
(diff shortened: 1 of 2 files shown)


purcell:
Yay!

tarsius:
@fommil Can you explain in more detail how scala-outline-popup "is no longer used", so that I can determine whether I should move it from the Emacsmirror to the Emacsattic or whether I should remove it completely.

fommil:
@tarsius @ancane abstracted it to work for arbitrary modes, so no longer `scala-outline-popup` but now https://github.com/ancane/popup-imenu

tarsius:
Thanks! In that case I'll remove the old package.

PR #3907, opened 2016-05-18
add elisp-fmt, CLI wrapper around elisp-format

fommil:
as discussed in https://github.com/cask/cask/issues/353 I don't know what needs to be done to get the `bin` directory using `Cask` so just going straight to publish.

1. Do I need to add the `:files` section here? My `Cask` already declares it.
2. Should I really add `*-test.el` to the exclude if I'm planning on adding tests later?


fommil:
interesting, it is looking for https://gitlab.com/fommil/elisp-fmt.el.git, I'll update my config

PR #3909, opened 2016-05-18
Add recipe for restclient-test

simenheg:
[restclient-test.el](https://github.com/simenheg/restclient-test.el) is a tool for turning [restclient.el](https://github.com/pashky/restclient.el) documents into interactive test suites.

I'm the package creator, and the repository is here: https://github.com/simenheg/restclient-test.el.

I had to provide `:files` because `*-test.el` is excluded by default.


purcell:
Nice, thanks!

PR #3910, opened 2016-05-18
Add recipe for annotate-depth

netromdk:
annotate-depth.el is a minor mode for showing if contents, mostly code, is too deep indentation
wise, which would hint at the need of creating sub functions.

I'm the package creator.


purcell:
Thanks.
- There are a bunch of unrelated commits in this PR which need to be cleaned up before I could conveniently apply it
- Don't shim `defvar-local` if it's missing. That's bad practice, as outlined in the [elisp conventions](https://www.gnu.org/software/emacs/manual/html_node/elisp/Coding-Conventions.html). Easier in this case just to use the classic `(make-variable-buffer-local 'name)` with `defvar` if you're not keen to declare a dependency on Emacs 24.3.

netromdk:
Sorry, I merged with upstream/master to make my fork up-to-date with melpa. If I re-fork then I orphan the other commit I have made to melpa before.

Okay, I thought it was okay since irony, magic and flycheck do it, too. I will change it.

Thanks

purcell:
> Okay, I thought it was okay since irony, magic and flycheck do it, too.

They're setting a bad example. :-)

netromdk:
Hehe yes, they are then.

I updated so it uses `(make-variable-buffer-local ..)` instead.

netromdk:
I will do another PR that is clean.

PR #3912, opened 2016-05-19
Change go-impl repository

syohex:
Please see
- https://github.com/dominikh/go-impl.el/issues/2

My version implements all features of original version and in addition, implements package and interface completion feature.

CC: @dominikh


dominikh:
:+1:

purcell:
Thanks folks.

PR #3913, opened 2016-05-19
Add recipe for 'datetime' package.

doublep:
Datetime is a library for parsing, formatting, matching and recoding timestamps and date-time format strings. It is in _alpha_ stage, i.e. not generally useful for other people. However, it will be used by 'logview' package (also mine, already in MELPA) to improve automatic log format detection. The library is designed to be generally useful, but currently most of functionality is missing.

* https://github.com/doublep/datetime

* I'm the author of the library.

### DIFF
--- a/recipes/datetime
+++ b/recipes/datetime
@@ -1,0 +1,2 @@
+(datetime :repo "doublep/datetime"
+          :fetcher github)


purcell:
Super, thanks. Looks promising.

PR #3915, opened 2016-05-19
add recipe for scrooge

cosmicexplorer:
This is just a copy/paste of the mode provided by [thrift](melpa.org/#/thrift) already; I would normally declare it as a derived mode, but I don't think you can just delete entries from the syntax table as a derived mode, and that's required for this to work; also thrift-mode is < 200 lines anyway. Currently it just highlights lines starting with `#@namespace` nicely (instead of just making them comments like thrift mode currently does).

### DIFF
--- a/recipes/scrooge
+++ b/recipes/scrooge
@@ -1,0 +1,1 @@
+(scrooge :fetcher github :repo "cosmicexplorer/emacs-scrooge")


purcell:
Don't see why you can't change the syntax table in a derived mode. It would certainly be best to avoid duplicating the `thrift` code.

cosmicexplorer:
That's true. I wasn't familiar enough with syntax tables. I've fixed that and made it derive from `thrift-mode` and simplified the highlighting a bit so now it's less than half its size. I'm not sure if the `Package-Requires` section is done correctly since `thrift` doesn't give a specific version number except for the `Package-Version` field.

purcell:
Thanks. Just pending https://github.com/cosmicexplorer/emacs-scrooge/pull/1 now -- https://stable.melpa.org/#/thrift was the place to look for the version. :-)

cosmicexplorer:
Thanks for the link -- I'll use that next time.

purcell:
Super, thanks.

PR #3917, opened 2016-05-20
Added recipe for annotate-depth.

netromdk:
[annotate-depth.el](https://github.com/netromdk/annotate-depth/blob/master/annotate-depth.el) is a minor mode for showing if contents, mostly code, is too deep indentation wise, which would hint at the need of creating sub functions.

I'm the package creator.


purcell:
Thanks for reworking and resubmitting the PR.

PR #3918, opened 2016-05-20
Add markdown-mac-link recipe

xuchunyang:
Summary: Insert Markdown links to items selected in various Mac apps
Homepage: https://github.com/xuchunyang/markdown-mac-link

### DIFF
--- a/recipes/markdown-mac-link
+++ b/recipes/markdown-mac-link
@@ -1,0 +1,1 @@
+(markdown-mac-link :fetcher github :repo "xuchunyang/markdown-mac-link")


syohex:
https://github.com/xuchunyang/markdown-mac-link/blob/f9ba1d6a8fafaf2b45f77295e59d5df63fb35b07/markdown-mac-link.el#L166-L169

`read-multiple-choice` is very new function. Only development Emacs users can use it(it is not defined on even pretest version)

xuchunyang:
> read-multiple-choice is very new function. Only development Emacs users can use it(it is not defined on even pretest version)

Fixed

purcell:
Nice, thanks.

I feel like there should be a general package like `mac-grab-link` which this uses, because that could be useful in various ways. :-)

PR #3919, opened 2016-05-20
Add datetime-format recipe

zonuexe:
- Datetime formatting helper library
- https://github.com/zonuexe/emacs-datetime
- I am author


purcell:
Thanks. We don't include documentation unless it is in GNU Info format, so please remove the `:files` spec.

zonuexe:
@purcell Thanks. Fixed it.

purcell:
Thanks

PR #3920, opened 2016-05-20
Add timecop.el

zonuexe:
- Datetime testing helper library
- https://github.com/zonuexe/emacs-datetime
- I am author


purcell:
Thanks - same feedback re. the `:files` spec as for #3919.

purcell:
:+1:

PR #3921, opened 2016-05-21
Delete python3-info

milkypostman:
python3-info will not build on the server because the build times out. the package is just too big.

@dvhansen any chance there is a better way to do this? it's compiling the info files that is just too much. if they were prebuilt in the repo it might work, otherwise i think we should just remove the recipe.


purcell:
Good call.

PR #3922, opened 2016-05-21
add go recipe files path

stardiviner:


purcell:
I think this should actually just be `"back-ends"` rather than `"back-ends/*.el"`: the `back-ends` directory should be included as-is in the package, not flattened.

stardiviner:
@purcell right. So I should use `:files ("back-ends" "stone.wav")` ?

purcell:
Not quite: I'm suggesting `:files (:defaults "back-ends" "stone.wav")`

stardiviner:
@purcell I updated my PR. Please check it out. Thanks

purcell:
Thanks!

PR #3924, opened 2016-05-24
Add counsel-dash

nathankot:
[counsel-dash](https://github.com/nathankot/counsel-dash) is an ivy-mode frontend for Dash docsets :)

### DIFF
--- a/recipes/counsel-dash
+++ b/recipes/counsel-dash
@@ -1,0 +1,4 @@
+(helm-dash
+  :repo "nathankot/counsel-dash"
+  :fetcher github
+  :files ("counsel-dash.el"))
--- a/recipes/counsel-dash
+++ b/recipes/counsel-dash
@@ -1,4 +1,3 @@
-(helm-dash
+(counsel-dash
   :repo "nathankot/counsel-dash"
-  :fetcher github
-  :files ("counsel-dash.el"))
+  :fetcher github)


nathankot:
Sorry for the opening/closing, been pinging Travis to re-rerun the build.
There was a missing colon in the `Package-Requires` section, fixed by https://github.com/nathankot/counsel-dash/commit/5ea829311690d7963651e850ff5365f6deccf423 - however re-running the Travis build doesn't seem to pick the change up.

Unless something else is wrong that I am unaware of :(

recipes/counsel-dash
```diff
@@ -0,0 +1,4 @@
+(helm-dash
```
syohex: `helm-dash` -> `counsel-dash`

recipes/counsel-dash
```diff
@@ -0,0 +1,4 @@
...
+  :repo "nathankot/counsel-dash"
+  :fetcher github
+  :files ("counsel-dash.el"))
```
syohex: You need not specify `:files` attribute. It is enough to use default value.

recipes/counsel-dash
nathankot: ahh what was I thinking xD

purcell:
The README has instructions for testing the recipe locally before submitting, BTW... :-) (tl;dr - `C-c C-c`)

purcell:
Merged - thanks. Seems less than optimal to depend on `helm-dash` for an `ivy` helper... perhaps you could push upstream for the core of `helm-dash` to be extracted as a `helm`-independent library...?

nathankot:
😂  thanks @purcell 

I've suggested this on areina/helm-dash#119, just waiting for feedback from the helm-dash maintainers - but I'll be happy to do it if they indicate they will accept such a change :)

purcell:
Super, thanks for following up on that.

PR #3926, opened 2016-05-24
Add mini-header-line

ksjogo:
Creator and maintainer
https://github.com/ksjogo/mini-header-line

### DIFF
--- a/recipes/mini-header-line
+++ b/recipes/mini-header-line
@@ -1,0 +1,1 @@
+(labburn-theme :repo "ksjogo/labburn-theme" :fetcher github)


purcell:
Thanks. You didn't test this locally following the instructions: the build failure shows the recipe says `labburn-theme`, so that needs fixing. :-)

Further notes:
- You can use `bound-and-true-p` to simplify `(and (boundp 'flycheck-mode) flycheck-mode)`
- The hard-coded list of modes makes this look like personal configuration, not a package ready for general use.
- Don't use `defadvice` inside a function
- `setq-default` is only available in Emacs 24.3 and greater, so you'd need to declare a dependency on the minimum compatible Emacs version
- The focus hooks are even newer
- Don't use `:` as a symbol specifier: instead, follow the [elisp conventions](https://www.gnu.org/s/emacs/manual/html_node/elisp/Coding-Conventions.html).
- Better to declare a face than an individual background colour.
- It's not great to have lots of references to specific modes or behaviours, like `flycheck` or `js2-mode`: you should instead make this package neutral and customisable.

Overall, I think this isn't ready for wider use yet, so I'm going to close this PR, but feel free to resubmit at a later date.

ksjogo:
Oh, yeah, that is an error.
Thanks a lot for the feedback, I will improve the package. Just recently got into the spirit to move my rather large configuration into more manageable packages, so miss some experience there, will report back.

purcell:
Feel free to ping me if you have any questions.

ksjogo:
Just added customization options and fixed the elisp conventions: https://github.com/ksjogo/mini-header-line/commit/55b07f76034a7846d7a4b29289286f8063367c16
Regarding the direct checks for other modes: yes, that was totally unneeded, now just checking if a file is associated with it.
Direct integrated error functionality is quite helpful imo. js2 and flycheck are my mains, do you think another one should be added by default?

purcell:
Thanks -- looking much better! Sent you a small fix. Having `:background "#292929"` without a corresponding `:foreground` is a bit of a code smell, because it guarantees that the results will look bad with many themes. Perhaps just use `bold` as the default?

Also, face-remap is an unusual way to do this: why not just propertise the buffer's `header-line-format` with that additional face if the current buffer is active? That way, you can just define `mini-header-line-active` to `:inherit header-line`, which would feel more natural. (Maybe that's simply not possible, though, due to how the header-line works...)

Using `global-mini-header-line-mode-hook` to do all the work isn't quite right: better to just define the minor mode with `:global t` and put all that conditional code in the body of the mode definition.

I feel like the advice on `other-window` etc. should be redundant: if your goal is to replace the modeline, why not advise something like `force-mode-line-update`, which many libraries call directly. Seems like it might work more reliably.

purcell:
(Sorry for the slow reply, have had a crazy week.)

ksjogo:
Thank you very much for the help.
It is bold now as default.

I am unsure about propertising the header-line-format. If I just use propertize and the custom face, it will overwrite all the other faces present in that string.

Mode configuration is simplified now.

PR #3929, opened 2016-05-25
add recipe for counsel-osx-app

d12frosted:
Simple as that 😸

### DIFF
--- a/recipes/counsel-osx-app
+++ b/recipes/counsel-osx-app
@@ -1,0 +1,1 @@
+(counsel-osx-app :repo "d12frosted/counsel-osx-app" :fetcher github)


purcell:
Thanks! The prefix for public symbols should really be `counsel-osx-app-` throughout, not `counsel-apps-`, to match the package name and avoid possible future clashes with other packages.

purcell:
Also, you're using `user-error`, which is unavailable in some of the Emacs versions which `ivy` supports, so you should explicitly depend on the minimum necessary Emacs version via your `Package-Requires` clause.

purcell:
Bonus note:
```
(if (bound-and-true-p file)
    (format "open '%s' -a '%s'" file app)
  (format "open '%s'" app))
```
will break if filenames have quotes in them. Just use `(format "open %s -a %s" (shell-quote-argument file) (shell-quote-argument app))` instead.

d12frosted:
> Thanks! The prefix for public symbols should really be counsel-osx-app- throughout, not counsel-apps-, to match the package name and avoid possible future clashes with other packages.

Wow. I clearly forgot to do that after I renamed the package. Thanks for catching this 😊 

> Also, you're using user-error, which is unavailable in some of the Emacs versions which ivy supports, so you should explicitly depend on the minimum necessary Emacs version via your Package-Requires clause.

Ah, I thought that it's enough to require `ivy`, so proper version of Emacs will be required by cross dependency. But it's not a big deal.

> Bonus note:

Ah, I didn't know. Thanks for that. It's time to update some other functions I have in my `init.el` 😸

purcell:
> Ah, I thought that it's enough to require ivy, so proper version of Emacs will be required by cross dependency. But it's not a big deal.

Ivy needs Emacs 24.1, but I *think* `user-error` was added a little later. I might be wrong, though. Good to be explicit with dependencies in any case.

xuchunyang:
> Ivy needs Emacs 24.1, but I think user-error was added a little later

`user-error` was added in Emacs 24.3: https://github.com/emacs-mirror/emacs/blob/1ee91bf89176251f6e399c8436dca0248cdd6f6b/etc/NEWS.24#L2207

d12frosted:
@xuchunyang thanks for finding the proper version number 💯 

Updated my package.

d12frosted:
Ping 😸

purcell:
Thanks. Sorry for the slow response. Merged now. :-)

d12frosted:
Thanks!

PR #3930, opened 2016-05-26
add recipe for blak3mill3r/vmd-mode

blak3mill3r:


purcell:
Thanks. Sorry for the slow response. Quick feedback:
- `defvar-local` requires Emacs 24.3
- Don't use `/` in symbol prefixes - follow the [elisp conventions instead](https://www.gnu.org/s/emacs/manual/html_node/elisp/Coding-Conventions.html) and use `-`
- Use `defcustom` instead of `defvar` for settings the user is likely to modify, e.g. the binary path
- Remove `for Emacs` from the package description please: all elisp packages are for Emacs :-)

blak3mill3r:
Hey, thanks for the tips!

https://github.com/blak3mill3r/vmd-mode/compare/e645cd6e9bfccfa4da985f0e67244fd94f6bbe3d...master

I changed the `/`s to `-`s, used `defcustom`, and fixed the package description.

I added a note about the 24.3 requirement (is that what you had in mind?).

purcell:
Almost: see https://github.com/blak3mill3r/vmd-mode/pull/5 <https://github.com/blak3mill3r/vmd-mode/pull/5> and https://github.com/blak3mill3r/vmd-mode/pull/6 <https://github.com/blak3mill3r/vmd-mode/pull/6> :-)

purcell:
Thanks!

PR #3931, opened 2016-05-26
Ghost-blog recipe

javaguirre:
ghost-mode is a package to manage posts from your [Ghost](https://ghost.org) blog platform using its [Rest API](https://api.ghost.org/).

The package repository is in [github](https://github.com/javaguirre/ghost-blog-emacs).

Do I need to specify anything else?

Thank you! :-)

### DIFF
--- a/recipes/ghost-blog
+++ b/recipes/ghost-blog
@@ -1,0 +1,2 @@
+(ghost-blog :repo "javaguirre/ghost-blog-emacs"
+            :fetcher github)


purcell:
Thanks! Sorry for the slow response - merged now. :-)

You might consider depending on `(cl-lib "0.5")` and using that instead of the now-defunct `cl`: usually it just means using the `cl-` prefix for all the symbols from `cl`.

javaguirre:
Thank you!

I will make the change you suggested regarding`cl`. :-)

PR #3932, opened 2016-05-26
added recipe function-follow

aasoliz:
Function-follow finds a method call's definition within the current buffer, open buffers, or files. 

The github [repo](https://github.com/aasoliz/Function_Follow).

I am both the creater and maintainer of the package.

### DIFF
--- a/recipes/function-follow
+++ b/recipes/function-follow
@@ -1,0 +1,3 @@
+(function-follow
+ :fetcher github
+ :repo "aasoliz/Function_Follow")


purcell:
Thanks. Sorry for the slow reply. Quick feedback:
- `ff/` is not an acceptable symbol prefix, sorry. Use `function-follow-` as per the [elisp conventions](https://www.gnu.org/s/emacs/manual/html_node/elisp/Coding-Conventions.html)
- Don't provide unprefixed features like `find-files`: either make it `function-follow-find-files`, or just put those 2 functions in the main `.el` file
- We'd suggest deleting the `-pkg.el` file, and just putting the package dependencies in `function-follow.el` as a `;; Package-Requires: ...` header
- In Emacs 24.5 and up, there's a new `xref` package, which is the neatest way to integrate code like this into a standard workflow. Look at `xref-js2`, for example, which does something similar to your package but for javascript. And my `elisp-slime-nav` package does the same thing for elisp.

Hope that helps.

aasoliz:
Thanks for the feedback. I'll check out the other packages.

tarsius:
@aasoliz Here's a friendly ping for you.

tarsius:
If you would still like to get this added to Melpa, then please address the mentioned issues and reopen this pull request.

PR #3935, opened 2016-05-28
Add epm recipe

xuchunyang:
Summary: A CLI wrapper for package.el to do some Emacs package management from command-line
Homepage: https://github.com/xuchunyang/epm


purcell:
Nice, thanks!

PR #3936, opened 2016-05-28
Add emacs-trr in recipes

iory:


purcell:
Thanks. The `TRR:` prefix should be `trr-`, as per the [elisp conventions](https://www.gnu.org/s/emacs/manual/html_node/elisp/Coding-Conventions.html) - should be a straightforward search/replace for @kawabata. :-)

Looks good otherwise.

iory:
@purcell 
replaced TRR: -> trr-

purcell:
Super, thanks!

iory:
Thanks, too!

PR #3937, opened 2016-05-28
Add Tiny Menu

aaronbieber:
Tiny Menu (`tiny-menu.el`) is a lightweight, easily configurable menu framework for creating single-key minibuffer menus whose entries can run any Emacs function.

### DIFF
--- a/recipes/tiny-menu
+++ b/recipes/tiny-menu
@@ -1,0 +1,3 @@
+(tiny-menu
+ :fetcher github
+ :repo "aaronbieber/tiny-menu.el")


purcell:
Thanks. Quick comments:
- `air-menu-of-menus` doesn't have the `tiny-menu-` prefix
- Thanks to `lexical-binding`, I feel like `tiny-menu-run-item` can probably be a function rather than a macro

aaronbieber:
As always, Steve, your thoroughness is inspiring. I completely overlooked that prefix from prior to tidying this for release.

I haven't had a lot of experience with lexical binding, so I'll experiment with a function solution and then ping this thread if that's alright.

aaronbieber:
Setting `lexical-binding` did the trick; I had actually tried returning a lambda previously, which of course didn't work because lexical binding was not enabled. In this simple case I feel like the macro solution is equally legible but since lexical binding is the way of the future this is a good change.

Are we good to go?

purcell:
Yup, super.

Re. the lexical-binding thing, since it's generally good practice to avoid macros whenever a function will do instead, I think that was the best choice. :-)

PR #3941, opened 2016-05-30
Add helm-smex

ptrv:


ptrv:
Repo: https://github.com/ptrv/helm-smex

purcell:
Nice! `smex` is one of the reasons I don't use `helm`... :-)

ptrv:
Yeah! I missed `smex` too since I switched to helm a couple of month ago. It is so simple and efficient.

PR #3944, opened 2016-05-30
get transpose-frame from wiki

tarsius:
Getting this package from the nu-mode repository (see #3757) was a
mistake because that isn't the upstream repository, it just bundled
this third-party library.  And now the nu-mode repository no longer
bundles transpose-frame.el.


purcell:
Good catch, thanks

PR #3945, opened 2016-05-31
Add related minor mode

julien-montmartin:
Hello,

I would like to add a package of mine : related minor mode, which helps you to navigate across similarly named buffers.

Comments are welcome :)

### DIFF
--- a/recipes/related
+++ b/recipes/related
@@ -1,0 +1,1 @@
+(related :fetcher git :url "https://bitbucket.org/lyude/related")


purcell:
Thanks. Quick feedback:
- MELPA has a "bitbucket" fetcher, which would be preferable to "git" in this recipe
- C-t and C-S-t are awful default bindings, because they shadow everyday built-in commands. Instead, follow the [emacs keybinding conventsions](https://www.gnu.org/software/emacs/manual/html_node/elisp/Key-Binding-Conventions.html) for minor modes.
- When using `cl-lib`, you need to add a `;; Package-Requires: ((cl-lib "0.5"))` metadata line
- Don't use `setq` in `let` to set multiple values which depend on each other. Instead, use `let*`.

Most importantly, though, I think that the built-in `find-file.el` library does what this library tries to do - and more - in a more general way, so it seems redundant to add this very specific library to MELPA, particularly with a general name like `related`.

julien-montmartin:
Hi Purcell,

Thanks for your comments, especially let* :)

- Changed key bindings to "C-x <up>" and "C-x <down>", which are reported "unused" by emacs -q
- Added proper Package-Requires
- Removed setq in let 
 
I know of find-file. I use it a lot. But it works on files, and rely on search paths. When files are close to one another, it's great. But when files are scattered across the file system, it's not so great. Related was started as a work around this limitation. Moreover, related "groups" more stuff together. For example you could not cycle between "foo_v1.c", "foo_v2.c" and "foo_v3.c" with find-file. With related it works, with nothing to configure.

> MELPA has a "bitbucket" fetcher, which would be preferable to "git" in this recipe

Could not get it to work yesterday evening. Looking at the doc, it says :

> Note: bitbucket assumes hg: git repos hosted on bitbucket should use the git fetcher.

My bitbucket repo use git... Are you sure I should use the git fetcher ?

Thanks

julien-montmartin:
To further elaborate on my previous comment, find-file, as it name suggests is about finding files, whereas related is about grouping buffers. Find-file uses search paths and associations lists of known extensions, whereas related uses no configuration and try to guess how to group buffers. In simple scenarios, both find-file and related can be used to toggle between foo.h and foo.c, but that’s really all there is in common. In the end find-file and related could be used together, the first for searching and opening (or creating) files, and the second to group them.

purcell:
You're right about the `bitbucket` fetcher. And thanks for the explanation about differences between this and find-file.

`find-file` has a nice way to specify mappings for related files. I suspect you could add a similar way to customise the functionality for finding "related" buffers, and that that might make the library more useful for many users.

Circular lists of values in Emacs are typically called "rings". Have you seen all the `ring-*` functions? I feel like you could use them to simplify the code.

I also feel like it would be useful to provide a function which allows you to switch to any related buffer directly, e.g. `related-switch-buffer`: you'd just use `completing-read` to directly select a value from the related buffer list.

julien-montmartin:
Hi Steve,

`related-switch-buffer` on top of `completing-read` is a nice idea, easily done !

In the future, I might allow one to configure the "grouping heuristic" (a big name right now), but it's not ready yet. It won't work the same as find-file though. I would like to keep the "guessing" approach. I don't want to enumerate associations based on extensions (or whatever).

I know of rings, but it's not exactly what I need. I could use them, but I think it would not make the code simpler, just make it look a bit different. It may change in the future.

purcell:
Cool, thanks.

PR #3946, opened 2016-05-31
Add smbc plugin recipie

sakshamsharma:
Plugin to view [SMBC](http://smbc-comics.com) web-comic inside Emacs
[URL for plugin](https://github.com/sakshamsharma/melpa)

### DIFF
--- a/recipes/emacs-smbc
+++ b/recipes/emacs-smbc
@@ -1,1 +1,0 @@
-(emacs-smbc :repo "sakshamsharma/emacs-smbc" :fetcher github)
--- a/recipes/emacs-smbc
+++ b/recipes/emacs-smbc
@@ -1,0 +1,1 @@
+(emacs-smbc :repo "sakshamsharma/emacs-smbc" :fetcher github)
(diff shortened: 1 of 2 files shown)


purcell:
Thanks. Quick feedback:

- This should simply be called `smbc`. "emacs-" is very redundant in an elisp library.
- All top-level symbols should be named `smbc-*`, to conform with the [elisp conventions](https://www.gnu.org/s/emacs/manual/html_node/elisp/Coding-Conventions.html)
- `with-help-window` isn't appropriate here. It doesn't make sense for the resulting buffer to be in `help-mode`. Instead, create the buffer, then use a function like `display-buffer`, and consider putting the buffer into `special-mode`, which still gives you the handy `q` keybinding.

sakshamsharma:
Hi! Thanks for the tips! I've acted upon all those suggestions, could you please have a look again? [URL](https://github.com/sakshamsharma/emacs-smbc).
Thanks!

purcell:
Looks great, but you'll still need to rename and edit the MELPA recipe so it's for `smbc` and not `emacs-smbc` please.

sakshamsharma:
I've updated it now. Didn't want to force push in a merge request, so I think a squash would be needed. 
Thanks!

purcell:
Thanks, merged. (Force-push would have been okay, btw.)

PR #3950, opened 2016-06-01
Revert 3378 master

lurdan:
cf. #3378


purcell:
Yep, thanks for this.

PR #3951, opened 2016-06-01
term-manager

IvanMalison:
Summary: Maintain term-mode buffer based on some peice of emacs context
Repository: https://github.com/IvanMalison/term-manager
Association: Author/Maintainer

PR #3952, opened 2016-06-01
term-projectile

IvanMalison:
Depends on #3951

Summary: Maintain term-mode buffer based on current-projectile project
Repository: https://github.com/IvanMalison/term-manager


purcell:
`((term-manager))` isn't a valid dependency spec. The version number (in quotes) is mandatory. You should also depend on the latest stable projectile version, right?

IvanMalison:
Should be fixed up now. Thanks!

purcell:
Great - thanks!

PR #3953, opened 2016-06-02
New recipe: bnfc-mode

jmitchell:
BNFC is useful for defining context-free grammars as labelled-BNFs.
These files may be compiled into parser implementations in several
programming languages, pygments syntax highlighting, and
documentation.

Currently the mode provides basic font-locking support. I am the
author and current maintainer.

Repo: https://github.com/jmitchell/bnfc-mode

### DIFF
--- a/recipes/bnfc
+++ b/recipes/bnfc
@@ -1,0 +1,3 @@
+(bnfc
+  :repo "jmitchell/bnfc-mode"
+  :fetcher github)


purcell:
Thanks! Quick feedback:
- Should have an `;;;###autoload` for the mode declaration and the `auto-mode-alist` modification
- `prog-mode` requires Emacs 24, and `setq-local` requires Emacs 24.3, so you should add `;; Package-Requires: ((emacs "24.3"))`

purcell:
Oh, and you shouldn't call `(font-lock-fontify-buffer)` directly. Don't forget that - in theory at least - users can have `font-lock-mode` disabled.

jmitchell:
Thanks for the feedback, @purcell. I've made the suggested changes and retested it using `make sandbox`.

purcell:
Great, thanks!

PR #3955, opened 2016-06-02
Add basic-c-compile recipe

nick96:
Pretty simple package but has been quite useful to me, hoping to put it out there so it can be useful to others.
Nick


purcell:
Thanks! Quick feedback:
- All top-level symbols need to be prefixed with the namespace, e.g. `basic-c-compile-...`, and `basic-c-compile--...` for internal functions: see the [elisp conventions](https://www.gnu.org/s/emacs/manual/html_node/elisp/Coding-Conventions.html)
- Interpolation with `%s` will explode when file paths have spaces in them: use `shell-quote-argument` to make the interpolated values safe
- Don't use `buffer-name` to get the filename: when using extras like `uniquify`, the two might not match each other
- Consider an `;;;###autoload` cookie for your main `interactive` command definition(s)

Hope that helps!

nick96:
Just want to clarify points 3 and 4.
Will `buffer-file-name` have the same issues as `buffer-name`?
What is the `;;;###autoload` cookie for? I had a look at the docs and am not really sure. It would just go above each interactive function right?
Thanks for the feedback!

purcell:
> Will buffer-file-name have the same issues as buffer-name?

No. `buffer-file-name` always returns the actual (full) file path of the file associated with the buffer - if any.

> What is the ;;;###autoload cookie for? I had a look at the docs and am not really sure. It would just go above each interactive function right?

It lets the user invoke a command without needing to add `(require 'some-library)` to their startup files first. So yes, you'd typically add one to each interactive function that the user might use as a first step in using your library.

nick96:
Disregard commit `dd42cce`. I managed to push Emacs autosave files.

purcell:
Sorry for delay: had to process manually because there are spurious changes in this PR.

I have a local copy of the recipe which I'll commit once https://github.com/nick96/basic-c-compile/pull/2 is merged.

purcell:
Thanks, added now.

PR #3956, opened 2016-06-02
Add synquid-mode

cpitclaudel:
Synquid is a programming language with a focus on synthesis. Flycheck is used to provide feedback during development.

### DIFF
--- a/recipes/synquid
+++ b/recipes/synquid
@@ -1,0 +1,1 @@
+(synquid :repo "cpitclaudel/synquid-emacs" :fetcher github)


purcell:
Thanks!

You'll need to declare a dependency on Emacs 24.3, due to use of setq-local. (BTW, font-lock-defaults is always buffer-local, so setq is fine there.)

It's bad practice to make a major mode dependent on flycheck, so you should package that part separately. Similarly, I feel like enabling prettify-symbols-mode is something a user should decide, not a major mode author.

purcell:
Also: don't enable `font-lock-mode`. Old-fashioned though this idea is, it should be possible to use a major mode without font-lock enabled.

cpitclaudel:
Thanks @purcell :) I fixed the font-lock thing and added the Emacs dependency. I agree it's better to specify it explicitly than to depend on the Emacs 24.3 dependency exported by Flycheck.

Re Flycheck and prettify-symbols-mode: I don't think removing them would be a good idea. Synquid is a synthesis language with a strong mathematical component, and the mode relies crucially on Flycheck to interact with the synthesizer. IOW this mode is mostly useless without Flycheck; live feedback is the core feature. The same goes for `prettify-symbols-mode`: this is how the language is supposed to look, so disabling it wouldn't be the best.

Hope this makes sense :)

purcell:
If the package provides a full environment rather than just a major mode, perhaps it would make sense to call it `synquid`?

cpitclaudel:
Not sure: that's what the repo of the actual language is called, so it could be confusing. Additionally, wouldn't that go against the convention of having a package called `language-mode` for a given `language`? `racket-mode`, `julia-mode`, `haskell-mode` and `js2-mode` all have much more than a basic major mode.

purcell:
There's never any confusion between an emacs library for a thing and the thing itself: a `-mode` isn't there to distinguish the two, and the repo for the former can always be called "thing-emacs". I've frequently suggested dropping redundant `-mode` suffixes for this reason.

I realise I'm being somewhat pedantic here (hey, that's kinda my job w.r.t. MELPA), but there's an expectation that `blah-mode` provides just the basic major mode support for `blah` and has minimal dependencies, such that it doesn't assume anything significant about the user's workflow. None of those major modes you listed have dependencies anywhere near as heavyweight as `flycheck`.

In this sense, `synquid-mode` is a complete `synquid` environment & opinionated workflow, not just a major mode, and the convention is to not suffix such things, so I think `synquid` would be a good name for it.

If `flycheck` were to have a significant competitor `notflycheck` that many users preferred, it would really suck if `notflycheck` users were forced to install and use `flycheck` simply in order to have Emacs support for a particular language -- that's exactly what a modular packaging system allows us to avoid. But if there were a `flycheck`-based integrated environment for that language, that would be a different decision for the user.

Having done some digging I see that `nim-mode`, `dart-mode` and `z3-mode` all currently depend on `flycheck` - all three packages could (and should) have their optional `flycheck` support moved to separate packages.

What do you think?

cpitclaudel:
Thanks for the thoughtful reply, and sorry for the delay. As always, I think your comments are spot-on :) Thus:

* I've renamed the repo into `synquid-emacs`
* I've renamed the package into `synquid`
* I've updated the PR accordingly

Thanks again for your help!

purcell:
Great, thanks!

cpitclaudel:
Yay, thanks so much :)

PR #3957, opened 2016-06-02
prettify-symbols for Greek and move darcula-theme to gitlab

fommil:
since the FSF gave gitlab an A rating, I'm thinking of moving all my personal emacs projects to gitlab.

- https://gitlab.com/fommil/emacs-darcula-theme
- https://gitlab.com/fommil/emacs-prettify-greek

### DIFF
--- a/recipes/darcula-theme
+++ b/recipes/darcula-theme
@@ -1,2 +1,2 @@
-(darcula-theme :fetcher github
-               :repo "fommil/darcula-theme-emacs")
+(darcula-theme :fetcher gitlab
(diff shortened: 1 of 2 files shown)


purcell:
Thanks. Maybe provide a simple autoloaded function which can easily add the `prettify-greek` symbols to the prettify config?

fommil:
Yeah I'll have a think about how to do that. The reason I didnt s because most major modes make this optional so a user would always explicitly list their symbols. Adding an update/mutate function would then split it into two steps.

PR #3958, opened 2016-06-03
Add flycheck mix syntax checker

tomekowal:
### Package summary
This package adds syntax checking for Elixir programming language using a tool shipped with Elixir called `mix`. Apart from regular flycheck syntax checker definition it also does two additional things:

* searches for project root, because `mix` works only with Elixir projects and not single files
* defines safety variable `flycheck-mix-enable-checking` which is by default `nil`

Elixir uses macros during compilation so it can run arbitrary code when compiling a project. The variable makes sure that user is able to enable checking only for trusted projects.

### Repository
https://github.com/tomekowal/flycheck-mix

### My association with the package.
I am the maintainer.

### DIFF
--- a/recipes/flycheck-mix
+++ b/recipes/flycheck-mix
@@ -1,0 +1,1 @@
+(flycheck-mix :repo "tomekowal/flycheck-mix" :fetcher github)


purcell:
Thanks! This seems to be integrating with flycheck in a non-standard way, since it requires changing `flycheck-check-syntax-automatically` from the default. The `flycheck-mix-enable-checking` variable also seems like a hack, since the usual way to handle this would be to remove the check from `flycheck-checkers` if you don't want it: other flycheck checkers don't have similar variables. Further, don't use `%s` to interpolate file paths, in case they include quotes or whitespace: generally you should be escaping them first with `shell-quote-argument` or similar. Hope that helps!

tomekowal:
Thank you! This helps a lot. I'll try to improve the package according to the comments. I just need to make sure that I understand them correctly:

* wrt `flycheck-check-syntax-automatically`: mix only works on entire projects which is a good thing, because elixirc reports false positives when it doesn't know the context. This means that the syntax check only makes sense after file save. This might be non standard way, but it is the only way I see that allows accurate syntax checking. Is there any obvious way to improve that?
* wrt `flycheck-mix-enable-checking`: I can remove the variable. What is the standard way to add `elixir-mix` to `flycheck-checkers` on a per project basis? Can I somehow do it using directory variables?
* wrt `%s`: I understand I can still use it, but I firstly need to use `shell-quote-argument`?

Thanks again for your time. This is my first attempt on writing a package so all the feedback on making things more standard is golden! :)

purcell:
> wrt flycheck-check-syntax-automatically: mix only works on entire projects which is a good thing, because elixirc reports false positives when it doesn't know the context. This means that the syntax check only makes sense after file save. This might be non standard way, but it is the only way I see that allows accurate syntax checking. Is there any obvious way to improve that?

You know, let's invite @tonini over to this conversation, because I know he'll have some good info to add. :-)

> wrt flycheck-mix-enable-checking: I can remove the variable. What is the standard way to add elixir-mix to flycheck-checkers on a per project basis? Can I somehow do it using directory variables? [EDIT: After reading those two threads: flycheck/flycheck#630 and flycheck/flycheck#894 I believe that the package should not be concerned with this at all and the user should enable flycheck manually. Do you agree?]

Normally you'd just add your new checker to `flycheck-checkers` globally. And you can also define `:predicate` for the checker if there are situations in which you can determine that your checker would not be able to run properly. (Presence of the executable is already checked for automatically.)

> wrt %s: I understand I can still use it, but I firstly need to use shell-quote-argument?

Yes, that's usually the way to do it.

tomekowal:
I applied two of the discussed changes:

* deleted the `flycheck-mix-enable-checking` - now it is up to the user to disable it
* quoted directory with `shell-quote-argument`
* I left the `flycheck-check-syntax-automatically` unchanged.

purcell:
Great, merged. :-)

tonini:
Morning :)

@tomekowal Sorry for my delay. :-)

I just noticed you're using `locate-dominating-file` for locating the root of a mix project. You need to be aware that hex packages also including a `mix.exs` file and that could lead to some issue if you jumping around for example with the ability of alchemist and looking for definitions of code which is served by dependency packages.

There is also the case of umbrella project which we currently discuss here https://github.com/tonini/alchemist.el/issues/185

PR #3959, opened 2016-06-03
Add intero recipe (new package)

chrisdone:
Intero is a new Haskell package. It depends on flycheck and company, and the two built-in packages cl-lib and comint. haskell-mode is also necessary for the flycheck mode. Test on Emacs 24.5.1. More information here http://commercialhaskell.github.io/intero But I'm waiting until the package is in MELPA before adding "how to get" to the page.

### DIFF
--- a/recipes/intero
+++ b/recipes/intero
@@ -1,0 +1,3 @@
+(intero :repo "commercialhaskell/intero" 
+        :fetcher github
+        :files ("elisp/intero.el"))


purcell:
Yay, great, nice to see this getting pushed out to a wider user base!

I noticed that the way you're doing `intero-goto-definition` won't allow `M-,` to work intuitively -- if you add something like the following just before jumping to the new location, then all will be well:

```
      (if (fboundp 'xref-push-marker-stack) ;; Emacs 25
          (xref-push-marker-stack)
        (with-no-warnings
          (ring-insert find-tag-marker-ring (point-marker))))
```

In Emacs 25, `M-,` is bound to `xref-pop-marker-stack` by default, and in lower versions you'd bind `M-,` to `pop-tag-mark`.

PR #3960, opened 2016-06-04
add recipe emacs-profiles

myTerminal:
This package can be used to declare multiple configuration sets as profiles and name them to be able to load them as and when required. These declared configuration profiles can also be loaded at Emacs startup.

The package repository can be found at [https://github.com/myTerminal/emacs-profiles](https://github.com/myTerminal/emacs-profiles).

I am the creator and maintainer of this package.

---
This change is [Reviewable](https://reviewable.io/reviews/melpa/melpa/3960)

### DIFF
--- a/recipes/emacs-profiles
+++ b/recipes/emacs-profiles
@@ -1,0 +1,1 @@
+(emacs-profiles :repo "myTerminal/emacs-profiles" :fetcher github)


purcell:
Thanks!

This is really just a global minor mode with a numbered menu for user-defined functions: I'm not sure this provides anything new or widely useful, and packages like this have been submitted before and declined. Specifically, the Emacs theme system is designed to cover some of these needs, such that you can specify both faces and other settings -- multiple themes can be enabled simultaneously.

To define profiles for `emacs-profiles`, it's necessary to write elisp, and if people are going to do that, it would be much better for them to define a global minor mode for each profile, e.g. `office-mode`, `home-mode` and/or themes for those profiles.

So I'm going to decline this, sorry.

PR #3961, opened 2016-06-05
New recipe: circe-notifications

eqyiel:
This is something similar to `erc-notifications` in Emacs core, but for
Circe users.

I only know of about a dozen users but it might be helpful for them to
have it packaged in MELPA.

Hope I've done this correctly!

Repo: https://github.com/eqyiel/circe-notifications

### DIFF
--- a/recipes/circe-notifications
+++ b/recipes/circe-notifications
@@ -1,0 +1,1 @@
+(circe-notifications :fetcher github :repo "eqyiel/circe-notifications")


purcell:
Makes sense to add this, but the `dbus`/`growlnotify` etc. code makes me sad, because there's an amazing library called [`alert`](https://github.com/jwiegley/alert) which hides all that stuff away and gives you notifications on any platform. Since you're already depending on other libraries, why not change your code slightly to use `alert`?

You're using `advice-add` etc., too, so you must depend on Emacs 24.4, since `nadvice.el` was unavailable in earlier versions.

And if you assume a very recent Emacs, there is native support for focus events, which might help you to drop the `xdotool`/`xprop` code.

eqyiel:
Thanks for your feedback, I didn't think I'd be ever get a PR reviewed by @purcell!

I've had a go at implementing your suggestions.

Apart from [this issue](https://github.com/jwiegley/alert/issues/30) `alert` is a lot nicer.  I'm using [this workaround](https://github.com/eqyiel/circe-notifications/blob/master/circe-notifications.el#L206-L210) for the time being.

```elisp
;; If alert-user-configuration is nil, the :style keyword is ignored.
;; Workaround for now is just to add a dummy rule that does nothing.
;; https://github.com/jwiegley/alert/issues/30
(if (eq alert-user-configuration nil)
    (alert-add-rule :continue t))
```

purcell:
Yay, nice!

PR #3963, opened 2016-06-05
Add no-littering

tarsius:
https://github.com/tarsius/no-littering


purcell:
Thanks - you're missing a `(cl-lib "0.5")` package dependency though. :-)

tarsius:
Ups. I've fixed that now.

PR #3965, opened 2016-06-06
add recipe for 'ifind-mode-plus'

djangoliv:


purcell:
Thanks. A few issues here:

- This file won't byte-compile cleanly because it references `workspace-dir`. Don't make this code rely on the user setting a variable first: that's really not the emacs way to do things. Instead, make it a `defcustom` with a reasonable default value and allow the user to override.
- The package description on the first line cannot be wrapped: you should shorten it, e.g. to "Search like isearch across files in a workspace"
- Re. `ifind-excluded-files-list` and `ifind-excluded-dirs-list`, there are already multiple lists like this (e.g. `grep-find-ignored-files` and `grep-find-ignored-directories`), and code for building `find` commands: try to re-use those rather than reinvent the wheel.
- Use `define-minor-mode` to define `ifind-mode` if possible. The code in `ifind-exit` would then live in the mode definition form. Modifying `overriding-terminal-local-map` looks wrong to me.
- You're missing the mandatory `provide` form.
- Since the library is called `ifind-mode-plus`, the `ifind-*` symbols like `ifind-abort` should instead be called `ifind-mode-plus-*`.

Overall this feels like it's a mixed bag of project-switching logic and grep-find, and I'm not sure that it's going to be widely useful in its current form when there are more polished alternatives like `projectile` and `ag` etc., not to mention `helm`. Note that `ag.el`, in particular, handles searching in projects really nicely, without requiring the user to "switch" from one to another.

So I'm going to decline this package for now, and I'd encourage you to explore alternatives. If - after a while - you still feel like this package brings something new to the emacs ecosystem, feel free to resubmit.

PR #3967, opened 2016-06-06
Add copyit recipe

zonuexe:
* Yanking functions (file path, file contents...)
* https://github.com/zonuexe/copyit
* I am author


purcell:
Thanks. `copyit-pandoc.el` has different package dependencies: each of the two `.el` files in this repo should have its own recipe.

zonuexe:
@purcell ok, fixed it.

purcell:
Super, thanks.

PR #3968, opened 2016-06-06
Add pandoc recipe

zonuexe:
* [Pandoc](http://pandoc.org/) wrapper
* https://github.com/zonuexe/copyit
* I am author


purcell:
Great, thanks.

PR #3969, opened 2016-06-07
Add new swagger-to-org recipe

ahungry:
I am the author, the package can be found here:

https://github.com/ahungry/swagger-to-org

The purpose is to parse a swagger.json file (OpenAPI format, interchangeably known as Swagger) and produce an org-mode buffer/file export (the main JSON objects/topics broken down into appropriate org-mode headings, with the object structures in org-mode source blocks).

I didn't see anything like this in MELPA at the moment (or even outside of emacs for creating static documentation from swagger.json files), so I created it tonight.

Let me know what you think.


purcell:
Thanks. You probably want to add an `;;;###autoload` for the interactive command.

ahungry:
Oh good call!  Adding in latest commit right now.

PR #3970, opened 2016-06-07
* Add Buffer Layers recipe

swflint:
I'd like to add my library to Melpa.  It provides a fairly simple buffer management system.

### DIFF
--- a/recipes/buffer-layers
+++ b/recipes/buffer-layers
@@ -1,0 +1,1 @@
+(buffer-layers :repo "swflint/buffer-layers" :fetcher github)


purcell:
Thanks. Quick feedback:
- Rather than altering the user's global map, provide a global minor mode which they can enable or disable: it will have its own keymap
- Consider using `cl-lib` instead of the obsolete `cl`. That mostly just involves declaring a `;; Package-Requires: ((cl-lib "0.5"))` dependency, and prefixing all the `cl` symbols with `cl-`, e.g. `cl-remove-if`.
- The macro looks a bit excessive IMO: I think there's a simpler implementation lurking in there, possibly by wrapping a use of `cl-defstruct` to define information which layer functions can operate on.

swflint:
>>>>> Steve Purcell writes:

    SP> Thanks. Quick feedback:

    SP> * Rather than altering the user's global map, provide a global
    SP> minor mode which they can enable or disable: it will have its
    SP> own keymap

I'll work on that one right away.

    SP> * Consider using cl-lib instead of the obsolete cl. That mostly
    SP> just involves declaring a ;; Package-Requires: ((cl-lib "0.5"))
    SP> dependency, and prefixing all the cl symbols with cl-, e.g.
    SP> cl-remove-if.

Done.  I wasn't aware of the difference, but hey, I don't do too much
elisp programming.

    SP> * The macro looks a bit excessive IMO: I think there's a simpler
    SP> implementation lurking in there, possibly by wrapping a use of
    SP> cl-defstruct to define information which layer functions can
    SP> operate on.

Yeah, there probably is, and I might try to change that later on, but I
wrote it that way for easy debugging.


Thanks for the tips!

Sam

--
Samuel W. Flint
4096R/266596F4
      (9477 D23E 389E 40C5 2F10  DE19 68E5 318E 2665 96F4)
(λs.s s) λs.s s

swflint:
Is there anything else I can do to get this merged?

purcell:
Yay! Little more feedback:
- You should use the `buffer-layers-` prefix throughout for public symbols, as per the [elisp conventions](https://www.gnu.org/s/emacs/manual/html_node/elisp/Coding-Conventions.html). (`define-buffer-layer` would be the only exception.)
- You're going to need one or both of a decent `Commentary` and a thorough docstring for `define-buffer-layer` if anybody else is likely to use this, which should be the main point of adding it to MELPA. :-)

swflint:
Done!

--
Samuel W. Flint
4096R/266596F4
      (9477 D23E 389E 40C5 2F10  DE19 68E5 318E 2665 96F4)
(λs.s s) λs.s s

purcell:
Looks good. Merging now.

One last thing I noticed was that `(define-key ctl-x-map (kbd "L") buffer-layers-map)` will produce a completely uncustomizable binding. You can just bind `(kbd "C-x L")` in `buffer-layers-mode-map` at the top level: I *think* `define-minor-mode` will implicitly declare and enable that keymap for you.

PR #3973, opened 2016-06-09
Adds kotlin-mode

gregghz:
RE: https://github.com/melpa/melpa/pull/3727 & https://github.com/melpa/melpa/pull/3794 since the maintainer of the original package has abandoned it.

### DIFF
--- a/recipes/kotlin-mode
+++ b/recipes/kotlin-mode
@@ -1,0 +1,2 @@
+(kotlin-mode :repo "gregghz/kotlin-mode"
+             :fetcher github)


purcell:
Thanks. I'd suggest you don't `(define-key map (kbd "<tab>") 'c-indent-line-or-region)`, because `TAB` should normally be bound to `indent-for-tab-command`, which can intelligently trigger completion. Probably you should simply set `indent-line-function` appropriately and leave `TAB` unchanged.

jasonm23:
Copying myself in on this.

purcell:
@jasonm23 - you can just click in the sidebar under `Notifications`, I think. :-)

jasonm23:
Hey Steve, I was just adding myself to notify @gregghz

We should have linked a different fork than Quantumman's so we can maintain it after their disappearance.

purcell:
> We should have linked a different fork than Quantumman's so we can maintain it after their disappearance.

Not sure what you mean: this PR links to @gregghz's fork.

jasonm23:
It means I was wrong :)

PR #3974, opened 2016-06-09
Add nixops-tramp recipe

tathougies:
https://github.com/tathougies/nixops-tramp.el

### DIFF
--- a/recipes/nixops-tramp
+++ b/recipes/nixops-tramp
@@ -1,0 +1,1 @@
+(nixops-tramp :fetcher github :repo "tathougies/nixops-tramp.el")


purcell:
Thanks. I wouldn't suggest depending on and calling `exec-path-from-shell` from your library: it's not made for that, and is likely to cause problems or confusion on some people's systems. Instead, use the documentation to tell people what env vars they will need, and perhaps suggest that they can use `exec-path-from-shell` to help them set them.

Looks good otherwise. :-)

tarsius:
@tathougies it seems you don't have to do much for this to be included in Melpa. Do you still intend to do so or should this be closed?

tarsius:
I just had a look at the upstream repository. The suggested changes haven't been implemented and op does not appear to be particularly interested in getting this into Melpa. Also it has been a while and so I am closing this. @tathougies feel free to come back here and finish the process of getting this into Melpa.

PR #3975, opened 2016-06-10
Add helm-cider recipe

xiongtx:
Request to add a new package, [Helm CIDER](https://github.com/clojure-emacs/helm-cider).


purcell:
Thanks. Please remove the `:files` spec: it is redundant here.

xiongtx:
Removed the `:files` spec.

purcell:
Thanks!

PR #3976, opened 2016-06-11
Add tern-context-coloring recipe

jacksonrayhamilton:
This package allows for [tern](https://github.com/ternjs/tern) with the [tern-context-coloring](https://github.com/jacksonrayhamilton/tern-context-coloring) Node.js plugin to function as a back-end for [context coloring](https://github.com/jacksonrayhamilton/context-coloring) in Emacs.

Package repository: https://github.com/jacksonrayhamilton/tern-context-coloring

I am the maintainer.

To verify that this works, you can use `make sandbox` and `package-install` `tern-context-coloring`. Create a `.tern-config` file with the contents `{"plugins":{"context-coloring":{"charOffset":1}}}` in your home directory. Have Node.js installed (tested with version 6.2,1), `tern` and `tern-context-coloring` installed via npm (`npm i -g tern`, `git clone https://github.com/jacksonrayhamilton/tern-context-coloring.git && cd tern-context-coloring && npm link` [I will publish this package very soon]). Then visit a JavaScript buffer, do `tern-mode`, then do `context-coloring-mode`, and the functions and variables in the buffer should be colored according to their scope depth.

Or if that's too much of a hassle, just check if the functions got autoloaded. :) Tested this all myself and it worked.

### DIFF
--- a/recipes/tern-context-coloring
+++ b/recipes/tern-context-coloring
@@ -1,0 +1,3 @@
+(tern-context-coloring :fetcher github
+                       :repo "jacksonrayhamilton/tern-context-coloring"
+                       :files ("emacs/tern-context-coloring.el"))


purcell:
Thanks. The degree of autoloading you have is excessive, because it should be possible for a user to have a package installed without it being enabled: downloaded packages should behave like the libraries bundled with Emacs in this regard.

So instead of the slightly hacky autoloading of modifications to the context coloring hashtable, you should do something like provide an autoloaded initialisation function which the user can add to `tern-mode-hook`, or `context-coloring-mode-hook`.

Looks good otherwise! :-)

jacksonrayhamilton:
The goal there was for simplicity of configuration; I'm trying to imitate Flycheck mode's approach, where all the user needs to do is `(add-hook 'my-mode 'context-coloring-mode)`, and then context-coloring can just automatically figure out what strategy to use.

The ratio of autoloaded to non-autoloaded code is also a bit misleading; I thought it'd be acceptable given the amount that gets autoloaded (only a few functions; even though that's half the package, it's not much, probably not as much as some other packages autoload).

My original thought was that a little extra ugliness under the hood was worth the LOC reduction in user init files. Do you still disagree? If so, I'll rework the API to utilize hooks instead.

purcell:
> My original thought was that a little extra ugliness under the hood was worth the LOC reduction in user init files. Do you still disagree? If so, I'll rework the API to utilize hooks instead.

Yes, but you can provide a single autoloaded function (`tern-context-coloring-enable` or similar) which does the necessary setup for the user, so that the necessary init file changes are trivial. (The problem is with the package auto-enabling, not with autoloaded code in general.)

jacksonrayhamilton:
Okay, I'll rework this a little bit. Thanks for your input.

jacksonrayhamilton:
Alright, updated the package so that it's manually enabled:

```elisp
(eval-after-load 'context-coloring
  '(tern-context-coloring-setup))
```

(The name `tern-context-coloring-setup` is meant to be similar to another Tern plugin's setup function, `tern-ac-setup`.)

purcell:
:+1: thanks!

PR #3977, opened 2016-06-11
Add grab-mac-link and remove markdown-mac-link

xuchunyang:
Repo: https://github.com/xuchunyang/grab-mac-link.el
Description: Grab link from Mac Apps and insert it into Emacs

markdown-mac-link is removed by the way, becuase grab-mac-link makes it
unneeded.

### DIFF
--- a/recipes/grab-mac-link
+++ b/recipes/grab-mac-link
@@ -1,0 +1,1 @@
+(grab-mac-link :fetcher github :repo "xuchunyang/grab-mac-link.el")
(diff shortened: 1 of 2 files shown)


purcell:
Nice. Sorry for the delay. I can imagine wanting to programmatically grab links, so I think perhaps `grab-mac-link` should take 2 args, and then only when called interactively should it present the `[c]hrome [s]afari [f]irefox [F]inder [m]ail:` prompts.

xuchunyang:
Great idea! I will implement it soon. Thank you.

PR #3978, opened 2016-06-11
fix apel and litecoin-ticker

aborn:


purcell:
Yay, thanks for that.

PR #3979, opened 2016-06-11
Add new yapfify recipe.

JorisE:
Yapfify can be used to (automatically) call the python formatter [YAPF](https://github.com/google/yapf) on a buffer.

### DIFF
--- a/recipes/yapfify
+++ b/recipes/yapfify
@@ -1,0 +1,1 @@
+(yapfify :repo "JorisE/yapfify" :fetcher github)


purcell:
Thanks. Quick feedback:
- `call-yapf-bin` should be `yapfify-call-bin` or `yapfify-exec`
- Instead of the messy toggling code, consider defining a simple `yapfify-on-save-mode` minor mode which manages modifying the hooks. The result would be much neater IMO.
- Re. your `Wow, isn't this weird` comment, I *think* you can use `print` or similar within the `with-help-window` block to display text there, since it binds the standard output stream to the newly-created buffer.
- Looks like the let binding `(file (buffer-file-name))` is unused

purcell:
Another tip: `M-x checkdoc` :-)

JorisE:
Thanks for the feedback!

Especially the minor-mode. That really saves a lot of hassle :p

purcell:
Yay, happy that helped.

PR #3983, opened 2016-06-13
mini-header-line

ksjogo:
add (the improved) mini-header-line

### DIFF
--- a/recipes/mini-header-line
+++ b/recipes/mini-header-line
@@ -1,0 +1,1 @@
+(mini-header-line :fetcher github :repo "ksjogo/mini-header-line")


purcell:
Nice, thanks.

I feel like the use of `ad-activate` and `ad-deactivate` will potentially disable advice that other packages may have added to `select-window`. Best to explicitly avoid this.

ksjogo:
Oh, yes, it also disabled the evil advice on select window. Just changed that: https://github.com/ksjogo/mini-header-line/commit/1480c578a1f4c77365744d2487ae868a36d49d2a

PR #3986, opened 2016-06-14
add f3 recipe

cosmicexplorer:
`f3` is a package which uses helm and find to find and operate on files quickly. The repository is located [here](https://github.com/cosmicexplorer/f3). I am the package maintainer.

### DIFF
--- a/recipes/f3
+++ b/recipes/f3
@@ -1,0 +1,1 @@
+(f3 :fetcher github :repo "cosmicexplorer/f3")


purcell:
Thanks. Quick feedback:
- You'll need to depend on Emacs 24.4, for `subr-x`
- You should set a parent group for the `defgroup`, probably `'files`
- Please don't simply refer people to the github README in the `Commentary` section: there should be basic usage info and docstrings in the file itself, since Emacs should be self-documenting as far as possible.

cosmicexplorer:
I only used a single function from `subr-x`, so that has been removed. I set the parent group to `find-file`; I'll add commentary to the file presently.

cosmicexplorer:
Ok, I've written a short script to generate a Commentary section from the readme. I'll add more docstrings later today.

purcell:
Thanks.

PR #3987, opened 2016-06-14
Add unicocde-escape recipe

kosh04:
https://github.com/kosh04/unicode-escape.el

I think it is easy to use, but export functions `unicode-unescape` and `unicode-unescape-region`
might have deviate from the rule of package namespace.

I'll be modify these codes if necessary.


purcell:
> export functions unicode-unescape and unicode-unescape-region
> might have deviate from the rule of package namespace.

Thanks for pointing that out. I think we can let that pass in this case... :-)

PR #3988, opened 2016-06-15
Create org-ref-ivy

jkitchin:
This is an org-ref variant that does not depend on helm, and uses ivy instead.

### DIFF
--- a/recipes/org-ref
+++ b/recipes/org-ref
@@ -1,1 +1,4 @@
-(org-ref :fetcher github :repo "jkitchin/org-ref" :files (:defaults "org-ref.org" "org-ref.bib" "citeproc"))
+(org-ref-helm :fetcher github
+              :repo "jkitchin/org-ref"
+              :files ("org-ref-helm*.el"))
+
--- a/recipes/org-ref
+++ b/recipes/org-ref
@@ -1,4 +1,4 @@
-(org-ref-helm :fetcher github
-              :repo "jkitchin/org-ref"
-              :files ("org-ref-helm*.el"))
+(org-ref :fetcher github
+	 :repo "jkitchin/org-ref"
+	 :files ("org-ref-helm*.el"))
 
--- a/recipes/org-ref-core
+++ b/recipes/org-ref-core
@@ -1,0 +1,6 @@
+(org-ref-core :fetcher github
+	      :repo "jkitchin/org-ref"
+	      :files (:defaults
+		      "org-ref.org" "org-ref.bib" "citeproc"
+		      (:exclude "org-ref-ivy*.el" "org-ref-helm*.el")))
+
--- a/recipes/org-ref-ivy
+++ b/recipes/org-ref-ivy
@@ -1,0 +1,4 @@
+(org-ref-ivy :fetcher github :repo "jkitchin/org-ref"
+	     :files (:defaults
+		     "org-ref.org" "org-ref.bib" "citeproc"
+		     :exclude "*helm*"))
--- a/recipes/org-ref-ivy
+++ b/recipes/org-ref-ivy
@@ -1,4 +1,4 @@
 (org-ref-ivy :fetcher github :repo "jkitchin/org-ref"
 	     :files (:defaults
 		     "org-ref.org" "org-ref.bib" "citeproc"
-		     :exclude "*helm*"))
+		     (:exclude "*helm*")))
--- a/recipes/org-ref-ivy
+++ b/recipes/org-ref-ivy
@@ -1,4 +1,3 @@
-(org-ref-ivy :fetcher github :repo "jkitchin/org-ref"
-	     :files (:defaults
-		     "org-ref.org" "org-ref.bib" "citeproc"
-		     (:exclude "*helm*")))
+(org-ref-ivy :fetcher github
+             :repo "jkitchin/org-ref"
+             :files ("org-ref-ivy*.el"))


ericdanan:
I think you need parenthesis around `:exclude "*helm*"`.

jkitchin:
Good catch. I pushed a commit to fix that.

purcell:
I presume that `org-ref.org` is simply the documentation for `org-ref`? We wouldn't normally include that in the package, unless there's an easy way for the user to access it: they'd never normally find it, tucked away in the package installation dir.

purcell:
Also, this recipe is for a package called `org-ref-ivy`, but `:defaults` will include all `.el` files, which I think is probably not what you wanted.

jkitchin:
There is an easy way to access the help file via a command.

On Friday, June 17, 2016, Steve Purcell <notifications@github.com> wrote:

> I presume that org-ref.org is simply the documentation for org-ref? We
> wouldn't normally include that in the package, unless there's an easy way
> for the user to access it: they'd never normally find it, tucked away in
> the package installation dir.
>
> —
> Reply to this email directly, view it on GitHub
> <https://github.com/melpa/melpa/pull/3988#issuecomment-226906318>, or mute
> the thread
> <https://github.com/notifications/unsubscribe/ABiRVgEpu3g5zk63re0DhKoJi3PxKnJDks5qMzH8gaJpZM4I2Py->
> .
>


-- 
John

-----------------------------------
Professor John Kitchin
Doherty Hall A207F
Department of Chemical Engineering
Carnegie Mellon University
Pittsburgh, PA 15213
412-268-7803
@johnkitchin
http://kitchingroup.cheme.cmu.edu

jkitchin:
I am pretty sure it is what I want. We just want to exclude the helm
dependency of Org-ref.

On Friday, June 17, 2016, Steve Purcell <notifications@github.com> wrote:

> Also, this recipe is for a package called org-ref-ivy, but :defaults will
> include all .el files, which I think is probably not what you wanted.
>
> —
> Reply to this email directly, view it on GitHub
> <https://github.com/melpa/melpa/pull/3988#issuecomment-226906450>, or mute
> the thread
> <https://github.com/notifications/unsubscribe/ABiRVvBp5u_urTuWsqAUKLYHYXt4CypNks5qMzJTgaJpZM4I2Py->
> .
>


-- 
John

-----------------------------------
Professor John Kitchin
Doherty Hall A207F
Department of Chemical Engineering
Carnegie Mellon University
Pittsburgh, PA 15213
412-268-7803
@johnkitchin
http://kitchingroup.cheme.cmu.edu

purcell:
But this package will include `org-ref.el`, which says it depends on `helm`... I don't really get the strategy here. Both `org-ref` and `org-ref-ivy` would end up bundling all the core `org-ref` code. That doesn't seem right to me. I'd expect there to be a core package, then separate add-ons for `helm` and `ivy` support.

jkitchin:
Hm. I thought it would work this way. I am traveling until next wed. I will
have to take a look at it then. Thanks.

On Friday, June 17, 2016, Steve Purcell <notifications@github.com> wrote:

> But this package will include org-ref.el, which says it depends on helm...
> I don't really get the strategy here. Both org-ref and org-ref-ivy would
> end up bundling all the core org-ref code. That doesn't seem right to me.
> I'd expect there to be a core package, then separate add-ons for helm and
> ivy support.
>
> —
> Reply to this email directly, view it on GitHub
> <https://github.com/melpa/melpa/pull/3988#issuecomment-226909376>, or mute
> the thread
> <https://github.com/notifications/unsubscribe/ABiRVvO1_sUfIfpbUtKDedaX79kFHsLhks5qMzlGgaJpZM4I2Py->
> .
>


-- 
John

-----------------------------------
Professor John Kitchin
Doherty Hall A207F
Department of Chemical Engineering
Carnegie Mellon University
Pittsburgh, PA 15213
412-268-7803
@johnkitchin
http://kitchingroup.cheme.cmu.edu

jkitchin:
Ok, I think I have straightened this out. There is now an org-ref-core which is a backend independent core lib for org-ref.  org-ref-ivy will depend on that, and not include org-ref.el or any of the *helm*.el files, which depend on helm. It should otherwise include all the other el files, and the org-ref.org file (which is accessible via M-x org-ref-help.

org-ref will still depend on helm.

purcell:
Hmmm. This still looks off to me. I'd expect this PR to result in three recipes that look something like this (untested):

```el
(org-ref :fetcher github
         :repo "jkitchin/org-ref"
         :files (:defaults
                 "org-ref.org" "org-ref.bib" "citeproc"
                 (:exclude "org-ref-ivy*.el" "org-ref-helm*.el")))
```
```el
(org-ref-ivy :fetcher github
             :repo "jkitchin/org-ref"
             :files ("org-ref-ivy*.el"))
```
```el
(org-ref-helm :fetcher github
              :repo "jkitchin/org-ref"
              :files ("org-ref-helm*.el"))
```

jkitchin:
The PR does actually do what it should, which is create an org-ref-ivy package that doesn't depend on helm (I tested it, and from what I can see it does that).  Your suggestion is an alternative approach, but I haven't done that so far because I want users to install only one package, and I don't want to rename org-ref to org-ref-helm.

purcell:
It's easy enough to say "just install either org-ref-helm or org-ref-ivy depending on your preferred completion framework," and have package dependencies do the right thing. We shouldn't be distributing two different packages with any of the same contents. Effectively the docs + the org-ref core code should be the org-ref package (or "org-ref-core" if you prefer), and that would be declared as a package dependency by `org-ref-helm` or `org-ref-ivy`.

ericdanan:
It seems to me that @purcell raises two distinct issues:

1. Organizing files into packages "efficiently" (creating a core package so as not to have `org-ref` and `org-ref-ivy` containing common files)
2. Naming packages "consistently" (renaming `org-ref` to `org-ref-helm`, symmetrically to `org-ref-ivy`)

I may be wrong, but my impression is that @jkitchin is mostly concerned about the latter issue, because he wants users who don't know what a completion framework is (including some who already use `org-ref`) to be able to simply install (or update) `org-ref` and have everything working right away, with the `helm` backend (the `ivy` backend was developed more recently).

In that case perhaps a solution could be to separate a core package and two backend-specific addons as @purcell suggests, but name the core package `org-ref-core` and the `helm` addon `org-ref` instead of `org-ref-helm`? It seems to me that the user experience of installing `org-ref` would then remain unchanged, and the former issue would be solved. And if it is essential to solve the latter issue as well, perhaps the `helm` addon could be named `org-ref-helm` but `org-ref` could then be defined as an "alias"  for `org-ref-helm` in some way? (I don't know if this is possible, perhaps one could create an "empty" `org-ref` package having `org-ref-helm` as dependency?)

To be sure, let me make clear that I am not in charge of the `org-ref` repository in any way, I was simply involved in the discussions there about creating an `org-ref-ivy` package. So thiese are merely suggestions. If I can help in any way with whatever ends up being implemented, I will be happy to do it.

ericdanan:
It seems that the checks fail because the recipe file name must match the package name (currently, recipe file `org-ref` for package `org-ref-helm`)

jkitchin:
I have created the three recipes that I think are correct now. One is org-ref-core which is the main code used, and can be used with a simple reftex backend. org-ref is the helm powered backend, and depends on org-ref-core. I want to keep it the same name org-ref for backward compatibility. and org-ref-ivy which depends on org-ref-core and uses ivy for the completion backend.

The only issue is I currently have the org-ref-core dependency on org-ref and org-ref-ivy commented out since it causes my Travis CI tests to fail since the package doesn't exist yet. If you let me know if/when this will be applied, I will put the dependencies in and live with the short failed test time.

Thanks,

jkitchin:
I am declaring bankruptcy on this. It doesn't look like this separation is actually working too well, and it works fine to have one package. I am closing this until I am more confident in splitting it.

PR #3990, opened 2016-06-16
github-search

IvanMalison:
Summary: Clone github repositories by searching github using gh.el
Repository: https://github.com/IvanMalison/github-search

This depends on https://github.com/sigma/gh.el/pull/70 but I thought I would put this here so that it can be merged once that is accepted.

### DIFF
--- a/recipes/github-search
+++ b/recipes/github-search
@@ -1,0 +1,1 @@
+(github-search :repo "IvanMalison/github-search" :fetcher github)


purcell:
How does this compare with https://github.com/dgtized/github-clone.el ?

IvanMalison:
> How does this compare with https://github.com/dgtized/github-clone.el ?

It offers search through the github api.

With github-clone, you need to know the exact name of the user and the exact name of the repository.

With some minor changes you could use github-search in concert with github-clone.

I did file an issue there to add this functionality to github-clone (https://github.com/dgtized/github-clone.el/issues/18) but I ultimately decided against it since I wasn't really convinced the rest of github clone ends up being that useful, and I prefer to use magit to clone.

purcell:
Cool, thanks. I would actually love to switch from `github-clone`. The `magit` dependency here is a little heavyweight, I feel, but I can't imagine that anyone *wouldn't* be using `magit` if they're working with git, so :shrug: :-)

purcell:
This actually doesn't work for me, because there's no `gh-search` feature in `gh` or `github-search`, so the `(require 'gh-search)` form explodes.

IvanMalison:
@purcell yeah i mentioned in the request that

This depends on sigma/gh.el#70 but I thought I would put this here so that it can be merged once that is accepted.

purcell:
Oh, I see. Sorry, my bad -- I was a bit premature here. I'll wait for now rather than revert the commit and re-open this.

PR #3991, opened 2016-06-16
add typoscript-mode

ksjogo:
modernized ts-mode taken from https://www.emacswiki.org/emacs/ts-mode.el
more development will follow


purcell:
Thanks. When updating the code, maybe consider using `prog-mode` as the parent mode, and depending on Emacs 24 or greater. :-)

PR #3993, opened 2016-06-20
new recipe - obvious parentheses

bigos:
I thought i might be useful to be able to tweak colours of rainbow-delimiters irrespective of colours decided upon you by theme. My suggestion adds easily discernible colours which might make Lisp editing easier.

### DIFF
--- a/recipes/obvious-parentheses
+++ b/recipes/obvious-parentheses
@@ -1,0 +1,1 @@
+(obvious-parentheses :repo "bigos/obvious-parentheses" :fetcher github)


purcell:
Thanks! Quick feedback:

- The currently-missing version number for the `rainbow-delimiters` package dependency is mandatory
- All top-level symbols should be prefixed with `obvious-parentheses-`, as per the [elisp conventions](https://www.gnu.org/s/emacs/manual/html_node/elisp/Coding-Conventions.html)
- You've got `(require 'rainbow-delimiters)` in the command as well as at the top level: you don't need the former.
- It seems very unwise to override `highlight`, which is a core face most complete themes will already define

A further stylistic note: `(let (rgb) (setq rgb EXPR) ...)` should simply be `(let ((rgb EXPR)) ...)`

But more importantly, the code assumes that the colour of the default face in the frame in which it initially executes is the correct one to start from, but of course colour definitions can vary according to the current frame's capabilities, and by using `t` in the colour class of the face definitions, you're setting the new colours for every frame type.

So if you ran this code in a terminal frame, then it would unconditionally override all `rainbow-delimiters` faces for every possible frame type using the colours found in that frame.

This is okay if it suits your local config, but I don't think it's a general solution, and probably not really a widely usable package, even with the fixes at top. For now, then, I'm going to decline the package, since I think it's not ready for prime-time.

If you do work on it further, you might consider using your logic to define a theme which only provides definitions for these few faces: you can load multiple themes at one time, so you could use that technique to reversibly layer these face redefinitions on top of any other theme.

Hope that all helps!

bigos:
Thank you for detailed explanation. I'm not that brilliant with Emacs themes so I didn't even think about solving the problem this way. but your idea about themes might solve another problem I have about slime output colourization, so eventually I might kill two birds with one stone if I learn how to create my own theme.

Fanael:
> Have you ever thought that subtle coloring of Rainbow Delimiters is not enough?

Uh, no? I'm the maintainer and pretty much the sole author of rainbow-delimiters, and I don't even know *what the default colors are*.

That's how parens are colored here: delims Hardly subtle, isn't it?

Really, just use a better theme, or override the colors in your init.el.

bigos:
I wanted to try some easier to distribute way of overriding colors. I can see that creating a theme is the cleanest solution. Also I disagree with your colors. I'm not sure if I am color blind, but some of your colors are hard to tell from another. The innermost red clearly stands out, but the next two light bronzes are difficult to tell apart. I tried to find out what distances between hues work best. It's hard to agree with  theme creators choice. Beauty is in the eye of beholder. I think we should seek better way of indicating bracket depth. It helps not only with Lisp, but can be handy in other languages too.

Fanael:
> Also I disagree with your colors. I'm not sure if I am color blind, but some of your colors are hard to tell from another. The innermost red clearly stands out, but the next two light bronzes are difficult to tell apart.

That's because they are. It's some third-party theme I use so I don't have to maintain my own, but it's less than stellar in some regards, like this one.

At any rate, the point is that the rainbow-delimiters defaults aren't very important precisely because we have themes, so it's easy to override any and all face properties.

> I tried to find out what distances between hues work best.

That's an interesting problem, actually. Thanks to `rainbow-delimiters-pick-face-function`, it's possible play with that without touching the faces themselves. For example, picking 6 equidistant hues from CIE L\*a\*b\* color space with some random constants:
```emacs-lisp
(defun some-rainbow-delimiters-pick-face (depth match _loc)
  (cond
   ((<= depth 0) 'rainbow-delimiters-unmatched-face)
   ((not match) 'rainbow-delimiters-mismatched-face)
   (t
    (let* ((angle (- 1.7 (* 2.0 float-pi (/ (% depth 6) 6.0))))
           (a (* 50.0 (cos angle)))
           (b (* 50.0 (sin angle)))
           (color (mapcar (lambda (val) (min 1.0 (max 0.0 val)))
                          (color-lab-to-srgb 66.0 a b))))
      (list :foreground (apply #'color-rgb-to-hex color))))))

(setq-local rainbow-delimiters-pick-face-function #'some-rainbow-delimiters-pick-face)
```

bigos:
the problem is equidistant do not work for me. I need some king of
logarithmic distances. For example red and orange have close hue, but are
easily discernible. on blue/violet end similar hue distance is not
noticeable. Also same luminosity colors do not look the same yellow with
same luminosity appear to be brighter than red of violet.

perhaps I should prepare some screenshots or app where you could change
values and see how colors appear.

Fanael:
> the problem is equidistant do not work for me. I need some king of logarithmic distances. For example red and orange have close hue, but are easily discernible. on blue/violet end similar hue distance is not noticeable. Also same luminosity colors do not look the same yellow with same luminosity appear to be brighter than red of violet.

Are you using HSV or HSL? Sounds like it. Consider using CIE L\*a\*b\* instead, because its very point is that it's not linear and uniform changes of components result in uniform changes of *perceived* color.

bigos:
it's not so simple, depending on value of L the shape of available perceivable colors changes.

https://en.wikipedia.org/wiki/Lab_color_space#/media/File:Lab_color_space.png

I don't know yet if the formulas on the wikipedia page could be used for that.

Fanael:
> depending on value of L the shape of available perceivable colors changes

These images show that the shape of colors representable *in sRGB* changes. Unfortunately, sRGB is a) what most of common hardware uses and b) widely known to badly suck.

bigos:
each luminosity could be approximated as rectangle/ triangle. using color values for the corners, or at equal fraction between the corners could lead to a solution. I need to read and think about it.

bigos:
https://en.wikipedia.org/wiki/Primary_color#Additive_primaries  

it's a triangle, due to limits of the rgb display many theoretcally possile and dscernable shades of green are not available.


another interesting reading, lots of useful information
https://dot-color.com/category/color-science/

bigos:
https://dotcolordotcom.files.wordpress.com/2012/08/cie-1976-vs-1931-ipad-2x.png

Above link explains why I was able to use nearby hues for  for colors from red to yellow-green. On the CIE-1976 gamut shows that displays are fairly accurate representing color subtleties and the sRGB triangle edge lies fairly close to humanly perceptible colors in the red - yellow green range. That also explains why I had such problems with greens and blues. So I'm not color blind after all.

That also means that finding some equidistant CIE-Lab color values that could be converted to sRGB and still appear at uniform color distance on computer displays is a problem. From yellow-green to blue gamut line is far away from the sRGB triangle edge. That means that colors on that edge would appear less saturated, hence problem with discerning colors. 

Thank you for pointing me towards CIE-Lab and making me read that stuff. I think we are back to square one. Experimenting with colors near the edges of sRGB triangle (most saturated ones) could still give reasonable color sets for delimiters for various backgrounds, but I don't think you could just write an elisp function for that.  It is interesting problem and with more people involved we could come up with something interesting.

PR #3994, opened 2016-06-20
Add chinese-pyim-basedict recipe

tumashu:
This is the default dict of chinese-pyim


purcell:
Thanks. The `(featurep 'chinese-pyim)` test seems strange to me. Shouldn't this package depend on `chinese-pyim`?

tumashu:
No, chinese-pyim will depend chinese-pyim-basedict

purcell:
Okay, cool.

PR #3996, opened 2016-06-20
Add nash-mode recipe

tiago4orion:
Nash-mode is a major mode for [Nash](https://github.com/NeowayLabs/nash) shell scripts. It adds simple fontification and integration with nashfmt for automatic apply of nash coding style.

I'm the maintainer of both projects.

### DIFF
--- a/recipes/nash-mode
+++ b/recipes/nash-mode
@@ -1,0 +1,3 @@
+(nash-mode
+ :repo "tiago4orion/nash-mode.el"
+ :fetcher github)


purcell:
Thanks. Quick feedback:
- You need to declare a package dependency on Emacs 24.3, because you're using `setq-local`
- The `nashfmt-` prefix is inconsistent with `nash-`: you should use either `nash-fmt-` or `nash-nashfmt-` instead.
- Your instruction to use `(add-hook 'before-save-hook 'nashfmt-before-save)` isn't ideal: why change a global hook like that, then only do something if the buffer is in `nash-mode`. Instead, tell people to do something like: `(add-hook 'nash-mode 'nash-nashfmt-enable-on-save)`, where `nash-nashfmt-enable-on-save` would be a function which calls `(add-hook 'before-save-hook 'nash-nashfmt nil t)`, ie. uses the local version of `before-save-hook`.

Hope that helps!

purcell:
Great, thanks.

PR #3997, opened 2016-06-22
Add recipe for lispyville

noctuid:
[lispyville](https://github.com/noctuid/lispyville) is a package written by myself that provides methods for better integrating evil with lispy. It is similar to evil-cleverparens in that it provides "safe" operators (i.e. they won't unbalance parentheses) as well as other commands/motions, but it uses lispy instead of smartparens.

### DIFF
--- a/recipes/lispyville
+++ b/recipes/lispyville
@@ -1,0 +1,1 @@
+(lispyville :fetcher github :repo "noctuid/lispyville")


purcell:
You've got a spurious `elpa.el` in that repository which will get bundled into the package. Can you maybe remove the file, or move it to a subdir?

Also, can you depend on non-zero stable versions of `lispy` and `evil` please? Check on MELPA Stable to find these, e.g. https://stable.melpa.org/#/lispy

noctuid:
I moved it into a subdirectory.

> Also, can you depend on non-zero stable versions of lispy and evil please? 

I can't depend on a stable version for lispy (my package won't work with the latest stable version), but I added a version number for evil.

purcell:
> I can't depend on a stable version for lispy (my package won't work with the latest stable version), but I added a version number for evil.

Cool. You should therefore probably depend on the next planned stable version of lispy, because `0` says that the current stable version will work fine.

noctuid:
Thanks, I will.

bpstahlman:
@noctuid Can you elaborate on the statement, "I can't depend on a stable version for lispy (my package won't work with the latest stable version)". I'm having trouble, documented [here](https://github.com/noctuid/lispyville/issues/15), and I'm wondering whether it could have something to do with this.

Where can I find documentation on lispyville installation and configuration? I assumed I wouldn't need to do anything but load lispyville in a lispy-mode hook. Do I also need to install a specific version of lispy, or is that handled automatically by lispyville package dependencies?

PR #3998, opened 2016-06-22
Add simple-bookmarks recipe

jtkDvlp:
A smart enhancement for emacs providing a simple but convenient interface to your recently and often used bookmarks / functioncalls.

https://github.com/jtkDvlp/simple-bookmarks/

### DIFF
--- a/recipes/simple-bookmarks
+++ b/recipes/simple-bookmarks
@@ -1,0 +1,2 @@
+(simple-bookmarks :fetcher github
+                  :repo "jtkDvlp/simple-bookmarks")


purcell:
Thanks. Quick feedback:
- Don't use `sb/` as a prefix. There are only 26x26 two-character prefixes available, which is fewer than there are packages on MELPA alone. Just use `simple-bookmarks-`, as per the [elisp conventions](https://www.gnu.org/s/emacs/manual/html_node/elisp/Coding-Conventions.html). (Same applies to `sbf/`, `sbi/` etc -- just use the same prefix throughout.)
- `Emacs bookmark / functioncall manager` can just be `Bookmark / functioncall manager`, since all emacs packages are for Emacs. Same goes for `for emacs` in the commentary. :-)
- Use `defcustom` to declare the `simple-bookmarks` custom group, and to place it logically within the standard customisation hierarchy

Hope that helps!

jtkDvlp:
Thanks for feedback.

> Don't use sb/ as a prefix. There are only 26x26 two-character prefixes available, which is fewer than there are packages on MELPA alone. Just use simple-bookmarks-, as per the elisp conventions. (Same applies to sbf/, sbi/ etc -- just use the same prefix throughout.)

Got it

> Emacs bookmark / functioncall manager can just be Bookmark / functioncall manager, since all emacs packages are for Emacs. Same goes for for emacs in the commentary. :-)

Also got it XD

> Use defcustom to declare the simple-bookmarks custom group, and to place it logically within the standard customisation hierarchy

Don´t got it. I already used the `defcustom` but only for one customizable var all the other vars should not be customized. Isn´t it right this way?

purcell:
> Don´t got it. I already used the defcustom but only for one customizable var all the other vars should not be customized. Isn´t it right this way?

Sorry, my mistake: I meant to write `defgroup`, not `defcustom`!

jtkDvlp:
Ok, thanks. Now there is a group in v1.6.

purcell:
Great - merged! :-)

PR #3999, opened 2016-06-23
xcode-mode package.

nicklanasa:


purcell:
Thanks! Quick feedback:
- "for emacs" in the package description is redundant and should be removed. All elisp packages are for Emacs! :-)
- `(setq compilation-scroll-output t)` is inappropriate, as it rudely changes a global default for users without asking them -- keep that in your personal config
- Create a prefix map for `C-c C-x` to avoid repeating that everywhere. But in any case, `C-c C-letter` is [reserved for major modes](https://www.gnu.org/software/emacs/manual/html_node/elisp/Key-Binding-Conventions.html), so you should change to a different prefix
- I'd suggest bundling all the little `.el` files into `xcode-mode.el`: the current fine-grained separation really isn't idiomatic

Hope that helps!

nicklanasa:
@purcell https://github.com/nicklanasa/xcode-mode/pull/18

purcell:
Nice, thanks!

PR #4001, opened 2016-06-23
Add python-test recipe

marsam:
Hi:
I'd like to add the `python-test` package

Thanks in advance

### DIFF
--- a/recipes/python-test
+++ b/recipes/python-test
@@ -1,0 +1,4 @@
+(python-test
+ :fetcher github
+ :repo "emacs-pe/python-test.el"
+ :files ("python-test.el"))


purcell:
Thanks. As author of the `unittest.py` module many years ago, I approve of this package! :-)

The only suggestion I'd make would be to use `:inherit 'error` / `:inherit 'info` for the face definitions rather than define new colours, otherwise those faces will automatically look awful in every theme.

marsam:
> Thanks. As author of the unittest.py module many years ago, I approve of this package! :-)

Cool! I did not know it :+1: . 
Thanks, I've updated the package with your suggestions.

PR #4003, opened 2016-06-24
add v2ex-mode

aborn:
v2ex-mode for visiting v2ex.com.

### DIFF
--- a/v2ex-mode
+++ b/v2ex-mode
@@ -1,0 +1,1 @@
+(v2ex-mode :fetcher github :repo "aborn/v2ex-mode")


xuchunyang:
Repo link: https://github.com/aborn/v2ex-mode

xuchunyang:
Sent your a small fix. Pending https://github.com/aborn/v2ex-mode/pull/5

aborn:
Thanks @xuchunyang ! Merged!

purcell:
Thanks. Quick feedback:
- Please use `v2ex--` for private functions instead of `v2ex/`, as per the [elisp conventions](https://www.gnu.org/s/emacs/manual/html_node/elisp/Coding-Conventions.html).
- You can remove `in emacs` from the package description: it's redundant.
- Consider using `special-mode` as the parent mode. It will give you the `q` functionality for free, and it is a good convention to follow.

purcell:
Also, you should use `defcustom` to declare the `v2ex` custom group and specify its position in the standard customisation hierarchy.

aborn:
v2ex-mode.el code updated! Thanks @purcell advice.

purcell:
Thanks. I think you can/should remove the `v2ex-quit` command and binding now too.

aborn:
Yeah, I have removed the v2ex-quit command in latest code. Thanks again!

PR #4004, opened 2016-06-24
Add evil-colemak-basics recipe

wbolster:


purcell:
Looks good, but please depend on the latest stable version of evil, not `0`. And `(lambda () (evil-colemak-basics-mode t))` in the globalised mode definition can just be `'evil-colemak-basics-mode`.

wbolster:
thanks for the review! this was my first attempt at writing some emacs lisp. :)

i'll add a version number for the dependency. the other suggestion does not work correctly, so i did not apply it...

purcell:
> the other suggestion does not work correctly, so i did not apply it...

Odd. Well, okay... Merging now. :-)

PR #4005, opened 2016-06-24
Added https://github.com/K0HAX/git-load recipe.

K0HAX:
Git-load is a package that allows easily loading packages from arbitrary git repositories.

### DIFF
--- a/recipes/git-load
+++ b/recipes/git-load
@@ -1,0 +1,1 @@
+(git-load :repo "k0hax/git-load" :fetcher github)


purcell:
Thanks. Quick feedback:
- This isn't a well-formed package: you're missing a `provide` form, and the standard trailing comment line.
- Please depend on the latest stable version of the `git` package, not 0. See https://stable.melpa.org/#/git
- You should probably `expand-file-name` instead of `concat` to build file paths
- `M-x checkdoc` will help you clean up your docstrings

Overall, though, it's quite simplistic, and doesn't cover basic needs like generating `autoload` files, so I would prefer not to merge this: most users will be better served by either `el-get`, `quelpa` or MELPA itself, and this solution - while it may be great for you - is likely to cause confusion.

PR #4006, opened 2016-06-24
Add Niconama Comment Viewer recipe

NOBUTOKA:
This package is comment viewer of Niconico Live Broadcast (<http://live.nicovideo.jp/>).
User of this package can show comment of broadcast and submit comment to broadcast via emacs.

I am the auther of this package.

https://github.com/NOBUTOKA/niconama.el

### DIFF
--- a/recipes/niconama
+++ b/recipes/niconama
@@ -1,0 +1,1 @@
+(niconama :fetcher github :repo "NOBUTOKA/niconama.el")


purcell:
Thanks. Quick feedback:
- Don't use `caddr`, `search` etc. from `cl`: `cl-lib` has replaced `cl`, so use `cl-caddr`, `cl-search` etc. instead
- Names like `pick-string-from-list` are not valid, because you need to have a namespace prefix. Instead, use `niconama-` throughout, as per the [elisp conventions](https://www.gnu.org/s/emacs/manual/html_node/elisp/Coding-Conventions.html).
- You need to add a `(provide 'niconama)` form and a `;;; niconama.el ends here` comment at the end of the file.

Hope that helps!

NOBUTOKA:
Thanks for quick reply.
I have fixed these problem in [ea58ffc375595032f81d0525978aed9c0dc08420](https://github.com/NOBUTOKA/niconama.el/commit/ea58ffc375595032f81d0525978aed9c0dc08420).
Please confirm.

purcell:
Nice! Thanks. :-)

NOBUTOKA:
So thank you for accepting.
Are there something to do after this?
And should I let you know when I update this package majorly?
I'm so sorry to trouble you but I'm newbie in MELPA.

purcell:
No, there's nothing you need to do. MELPA will build the package automatically every few hours, and will always use the latest version of the code in your github repository. If you apply a numeric version tag, the package will also appear in MELPA Stable (stable.melpa.org). :)

NOBUTOKA:
I understand it.
Thanks a lot.

PR #4007, opened 2016-06-25
Add recipes/emacs-leuven

fniessen:
Emacs Leuven is a (big) set of better presets for Emacs.

https://github.com/fniessen/emacs-leuven

Best regards!

### DIFF
--- a/recipes/emacs-leuven
+++ b/recipes/emacs-leuven
@@ -1,0 +1,3 @@
+(emacs-leuven
+ :fetcher github
+ :repo "fniessen/emacs-leuven")


purcell:
Hi - nice config!

However, complete emacs configs really aren't a good fit for `package.el` and distribution via MELPA, sorry, so I'm going to decline this.

fniessen:
Hello. I know this is a meta-package, and I wanted to use MELPA so that newbies could simply insert one line in their .emacs and that was it. That'd help them a lot.

But, OK, I respect your decision.

Thanks for MELPA!

PR #4009, opened 2016-06-26
Adding a recipe for pyimport

Wilfred:
Pyimport is a package I've written that automates the process of adding or removing imports in Python scripts.

See the [GIFs in the readme](https://github.com/Wilfred/pyimport) for more details.


purcell:
Great, thanks.

PR #4010, opened 2016-06-26
Add palette-emacs-theme recipe

aslpavel:
Color palette based Emacs theme. Can be easily extended to support more color palettes.
Comes with gruvbox-{dark,light} included. Please read README.md for details.

### DIFF
--- a/recipes/palette-theme
+++ b/recipes/palette-theme
@@ -1,3 +1,4 @@
-(palette :repo "aslpavel/palette-emacs-theme"
-         :fetcher github
-         :files ("palette-theme.el"))
+(palette-theme 
+ :repo "aslpavel/palette-emacs-theme"
+ :fetcher github
+ :files ("palette-theme.el"))
--- a/recipes/palette-theme
+++ b/recipes/palette-theme
@@ -1,4 +1,3 @@
 (palette-theme 
  :repo "aslpavel/palette-emacs-theme"
- :fetcher github
- :files ("palette-theme.el"))
+ :fetcher github)
--- a/recipes/palette-theme
+++ b/recipes/palette-theme
@@ -1,0 +1,3 @@
+(palette :repo "aslpavel/palette-emacs-theme"
+         :fetcher github
+         :files ("palette-theme.el"))


purcell:
Thanks. Quick feedback:
- `presset` should be `preset` throughout :-)

purcell:
Thanks. Couple of small pieces of feedback:

- `presset` should be `preset` throughout :-)
- `(if cond (progn ...))` can more simply be written `(when cond ...)`
- The `:files` spec in the recipe is unnecessary and should be removed, please

This is an interesting approach you've taken. Overall, I think you shouldn't be trying to modify a single theme by cycling it through variants -- all sorts of things can go wrong with that, and you're working against the theme system. But you have all the machinery you need to elegantly generate multiple themes, and so it would be neater to have a `defpalette-theme` macro which can generate an entire palette theme variant as a distinct theme. Why not simply do that instead? You'd then ship several files, e.g. `palette-gruvbox-dark-theme`, and the user would have a nicer and more idiomatic experience. (I do something similar with my `tomorrow` themes, BTW.)

aslpavel:
Thank you for review. Fixes:
  - `preset` misspelling
  - `(if ... (cond` -> `(when`
  - `:files` removed

I might update this to macro later. But this (updating theme in place) approach seem to work. I often switch between light/dark themes. And this way you don't have to load different themes. I just load theme once `(load-theme 'palette t)` and cycle over configured palettes with keybinding for  `palette-preset-cycle`

aslpavel:
ping_cat_tail

purcell:
> I often switch between light/dark themes. And this way you don't have to load different themes.
But that's what the theme system is for!

> I just load theme once (load-theme 'palette t) and cycle over configured palettes with keybinding for palette-preset-cycle

When you use the theme system as intended, then you can use general-purpose packages like [cycle-themes](https://github.com/toroidal-code/cycle-themes.el) to do exactly that instead of reinventing the functionality in your package. :-)

milkypostman:
@aslpavell i can understand if you have put a lot of effort into this theme/package and want it to be accessible through MELPA but I tend to agree with @purcell that it is kinda building a theming system on top of an existing theming system. if you look at something like [base16-theme](http://melpa.org/#/base16-theme) you can see where someone generates a bunch of themes based off a palette and includes the actual themes in the package -- they are separate themes conforming to the emacs theme standard -- which I believe is what should be done here.

the theme system has been designed to try to standardize on a model which is consistent between theme developers. if every theme developer were to include their own specialized functions for tweaking their theme it would be chaos.

PR #4012, opened 2016-06-28
Add elpa-clone recipe

dochang:


purcell:
Thanks. What's the plan for this?

I'm wary of distributing this in its current form because our server will get hammered if people use it. Would you mind introducing a non-zero default sleep interval between successive fetches, so that there is some sort of basic rate limiting?

(Note also that a package repository usually contains `somepackage-readme.txt` files for each package too. There may also be GPG signatures.)

xuchunyang:
BTW, there is an [ELPA mirror](https://mirrors.tuna.tsinghua.edu.cn/help/elpa/) provided by Tsinghua University TUNA Association, so if you want clone, for example, melpa, then use `rsync://mirrors.tuna.tsinghua.edu.cn/elpa/melpa/`. rsync can save resources heavily.

purcell:
Yes, we could also possibly set up an rsync service.

dochang:
@purcell What about 5 seconds as minimal download interval? Is it enough?

dochang:
Currently it lacks some features (signature, readme, etc).  I will add them in the future.

I wrote this package because there is no standard method for mirroring ELPA archive.  Most archives don't support sync services like rsync and there is only HTTP which supported by all ELPA archives.

dochang:
@purcell A variable named `elpa-clone-download-interval` has been added.  Default value is 5 seconds.

purcell:
Super, thanks! Separately I've opened #4020, because I think we should provide rsync access at some point.

PR #4014, opened 2016-07-01
avy-flycheck

magicdirac:
Jump to and fix syntax errors using flycheck with avy interface

### DIFF
--- a/recipes/avy-flycheck
+++ b/recipes/avy-flycheck
@@ -1,0 +1,2 @@
+(ace-link :repo "magicdirac/avy-flycheck"
+          :fetcher github)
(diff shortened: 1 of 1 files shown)


purcell:
Thanks. Quick feedback:
- Your `Package-Requires` line is broken. Since you're a flycheck user, grab `flycheck-package` and it will help you fix it!
- You should also declare a dependency on `cl-lib "0.5"`

magicdirac:
I think `Package-Requires` has been fixed by this commit. Thanks.

purcell:
Yes, great - thanks!

PR #4015, opened 2016-07-01
Ox clip

jkitchin:
This is a new package to allow cross-platform formatted copy from org-mode.


jkitchin:
sorry! this has too many commits in it.

PR #4017, opened 2016-07-01
Add flycheck-bashate

alexmurray:


purcell:
Thanks. I've merged this for now, but you might consider submitting a PR to flycheck itself in order to add it. If you do so successfully, we can then remove this package again.

PR #4019, opened 2016-07-01
unison

unhammer:
Summary: simple wrapper for Unison file sync program
(see http://www.cis.upenn.edu/~bcpierce/unison/ )

Link: https://github.com/unhammer/unison.el

Association: Author.

(or should it be named "unison-sync" in case someone wants the rather nice name "unison" for anything else?)

### DIFF
--- a/recipes/unison
+++ b/recipes/unison
@@ -1,0 +1,1 @@
+(unison :fetcher github :repo "unhammer/unison.el")


purcell:
Thanks. Looks good, and the name is fine as-is. You're presumably using the now-deprecated `cl` library primarily for `lexical-let`, but you're targeting Emacs 24, so why not just declare `lexical-binding: t` and get rid of the `cl` dependency?

purcell:
Also, I think you should allow the `unison` command to prompt for arguments, and use the customised args as the default. Most users will just try running that command, then it will break messily due to the dummy arguments.

unhammer:
@purcell thanks, set lexical-binding: t now. I'm not very used to that yet, so have been using it sparingly so far, but I guess it's the new "standard" way of doing non-customisation elisp :) I set `unison-args` to nil, which will make unison print Usage, and made the error on that note what variable needs setting.

PR #4021, opened 2016-07-02
Add switch-buffer-functions recipe

10sr:
This package provides a hook variable <del>`post-command-current-buffer-changed-functions`</del> `switch-buffer-functions`, which will be run when the current buffer has been changed after each interactive command.

I'm the author of this pckage.

<del>https://github.com/10sr/post-command-current-buffer-changed-functions-el</del>
https://github.com/10sr/switch-buffer-functions-el

### DIFF
--- a/recipes/switch-buffer-functions
+++ b/recipes/switch-buffer-functions
@@ -1,0 +1,2 @@
+(switch-buffer-functions :fetcher github
+                         :repo "10sr/switch-buffer-functions-el")


purcell:
Thanks. The package name is longer than the package description! I think we can do better. How about calling this "switch-buffer-hook" or similar?

10sr:
Thanks! I've changed the name to `switch-buffer-functions`.

purcell:
Nice!

10sr:
Thank you!

PR #4022, opened 2016-07-02
Add npm-mode.

mojochao:
This change adds a recipe providing the 'npm-mode' package.

### DIFF
--- a/recipes/npm-mode
+++ b/recipes/npm-mode
@@ -1,0 +1,1 @@
+(npm-mode :fetcher github :repo "mojochao/npm-mode")


purcell:
Thanks. Quick feedback:
- Don't use `/` as the prefix separator: please use `-`, as per the [elisp conventions](https://www.gnu.org/s/emacs/manual/html_node/elisp/Coding-Conventions.html)
- Calling `ansi-term` is a strange and heavyweight choice. Why not use a lower-level primitive such as `compilation-start`? It's still possible to colorise output that contains ANSI escape sequences.
- Don't assume people want to use ido: call `completing-read`, not `ido-completing-read`. Then `helm` and `ido` can intercept those calls.
- It's not obvious to me why Emacs 24.4 is needed for the code: I feel like you could just depend on Emacs 24 and `cl-lib "0.5"`

Hope that helps!

mojochao:
Thanks for the pointers Steve.  Per you points:

* I went ahead and moved to elisp namespace naming conventions
* I did not change out the ansi-term use, but will check out compilation-start Real Soon Now (tm)
* I switched to using more general 'completing-read', and lo and behold it works with helm!
* Switched to emacs-23.1 dependency (when 'json package was added).  Why would I need to depend on cl-lib 0.5?

Thanks again for the pointers.  I learn something new every day. :-)

mojochao:
Oops.  I meant to comment not 'close and comment'! reopen as needed.

purcell:
Cool, but for historical reasons, you'll actually need to omit the Emacs 23 dependency altogether: the only versions of `package.el` which support Emacs 23 don't understand built-in packages or pseudopackages like `emacs`.

purcell:
Also, `npm-mode--find-file` can more simply be written by using the `locate-dominating-file` function.

mojochao:
Thanks for the tip!  It did make the code a lot tidier.  One other question, how long will it take to have this pull request pulled into master?  Just wondering when I can expect to see it up on melpa.

purcell:
> One other question, how long will it take to have this pull request pulled into master?

Can take a while these days now that I've got a busy day job, but just for you... CLICK!

PR #4023, opened 2016-07-03
Add recipe for parsec.el

cute-jumper:
This is a parser combinator library for Emacs Lisp, which is similar to Haskell's Parsec.

https://github.com/cute-jumper/parsec.el

### DIFF
--- a/recipes/parsec
+++ b/recipes/parsec
@@ -1,0 +1,2 @@
+(parsec :repo "cute-jumper/parsec.el"
+        :fetcher github)


purcell:
Nice, thanks. You should declare package dependencies on `(emacs "24")` (for `lexical-binding` support) and `(cl-lib "0.5")`.

cute-jumper:
Thanks. The dependencies have been added.

purcell:
Thanks!

PR #4024, opened 2016-07-04
Add vdiff recipe

justbur:
This is a package I'm working on to (roughly) emulate vimdiff.

It's usable right now, but still a work in progress.

### DIFF
--- a/recipes/vdiff
+++ b/recipes/vdiff
@@ -1,0 +1,1 @@
+(vdiff :repo "justbur/emacs-vdiff" :fetcher github)


purcell:
Does this fulfill a similar use case to `smerge-mode`?

Silex:
@purcell: just curious, how does one use `smerge-mode`? I'm a heavy `ediff` user, and to me `smerge-mode` doens't look very straightfoward to use...

That said, it made me discover `emerge-files`, which is an interesting alternative to ediff.

justbur:
@purcell not quite. smerge appears to be for merge conflicts. vdiff is an alternative to ediff for diffing and editing two arbitrary files or buffers. I wanted something like vimdiff in emacs and wasn't aware of an existing tool that worked the same way. You can get a sense of what it is from the [screenshot](https://github.com/justbur/emacs-vdiff#screenshot). The main features are folding of unchanged lines, commands for jumping between changes and moving changes between buffers, and syncing of scrolling the buffers.

purcell:
Thanks, merged. Sent you a small fix in https://github.com/justbur/emacs-vdiff/pull/2.

PR #4025, opened 2016-07-05
Font-man recipe

shackra:
I fixed the definition of the package for melpa, so it shouldn't fail this time.

I also updated my package to work better, thanks to `face-remap` and `switch-buffer-functions`, so this package is fully functional AFAICT

### DIFF
--- a/recipes/font-man
+++ b/recipes/font-man
@@ -1,1 +1,1 @@
-(font-man :repo "shackra/font-man" :fetcher bitbucket)
+(font-man :url "https://bitbucket.org/shackra/font-man.git" :fetcher git)
--- a/recipes/font-man
+++ b/recipes/font-man
@@ -1,0 +1,1 @@
+(font-man :repo "shackra/font-man" :fetcher bitbucket)


shackra:
In case someone is wondering what this package does:

font-man

purcell:
I really want to be encouraging here because I'd be delighted if someone wrote a better version of my `default-text-scale` package.

This approach really doesn't seem like a reliable or efficient one, though, and I'm wary of encouraging its distribution. It's hooking every single command, because you're trying to keep all windows in sync with each other manually, rather than using (say) `face-override-spec` to modify the default face in one place, which Emacs would then use everywhere. 

(Re. the code itself: use `(dolist ...)` instead of all the `setf` fiddling; and "temporally" should absolutely be "temporarily" instead.)

shackra:
I'll check the code and use dolist instead. Looping a list of buffers get a significant overhead on Emacs performance? Because I make the code to scale the text only if it wasn't scaled before and only if it is currently visible.

El 6 de julio de 2016 21:39:16 GMT-06:00, Steve Purcell <notifications@github.com> escribió:
>I really want to be encouraging here because I'd be delighted if
>someone wrote a better version of my `default-text-scale` package.
>
>This approach really doesn't seem like a reliable or efficient one,
>though, and I'm wary of encouraging its distribution. It's hooking
>every single command, because you're trying to keep all windows in sync
>with each other manually, rather than using (say) `face-override-spec`
>to modify the default face in one place, which Emacs would then use
>everywhere. 
>
>(Re. the code itself: use `(dolist ...)` instead of all the `setf`
>fiddling; and "temporally" should absolutely be "temporarily" instead.)
>
>---
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/melpa/melpa/pull/4025#issuecomment-230970926

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

tarsius:
@purcell Should we add a new label for things like #4025,  #4055, and #4071.

purcell:
@tarsius "awaiting-upstream" is my crappy label for that, though I don't apply it consistently

tarsius:
@shackra do you intend to address the concerns raised by @purcell or should we close this?

shackra:
I was thinking on doing benchmarks to prove that isn't a big deal performance-wise. 

In any case, I will send a new pull request when ready. 

El 8 de septiembre de 2016 3:52:46 GMT-06:00, Jonas Bernoulli <notifications@github.com> escribió:
>@shackra do you intend to address the concerns raised by @purcell or
>should we close this?
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/melpa/melpa/pull/4025#issuecomment-245549349

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

tarsius:
Okay. Thanks for the heads up.

PR #4028, opened 2016-07-05
Add recipe for org2elcomment

cute-jumper:
This package is mainly used to convert an org-mode file to the comments of an Elisp source code. It is a very simple package so I'm not sure whether it is worth being on melpa.

https://github.com/cute-jumper/org2elcomment

### DIFF
--- a/recipes/org2elcomment
+++ b/recipes/org2elcomment
@@ -1,0 +1,2 @@
+(org2elcomment :repo "cute-jumper/org2elcomment"
+               :fetcher github)


purcell:
Yeah, it's okay to include it in MELPA, but you'll need to declare a package dependency on whichever Emacs version introduced `pulse.el`.

cute-jumper:
I have no idea when it is introduced so I just silently require the `pulse` by `(require 'pulse nil t)` and use `featurep` to check in the runtime. Is it OK to do this?

purcell:
Yeah, should be okay. I checked, and `pulse.el` has been there since (at least) Emacs 24.1, so you can also consider just depending on `(emacs "24")`.

PR #4029, opened 2016-07-05
Added recipe for pushover.el

swflint:


swflint:
Is there a way to trigger the CI to re-run?

milkypostman:
@swflint is that your api key in the file or an invalid one? you should probably make that blank by default.

- variables shouldn't start with `*`

- `pushover-send` probably needs an `;;;### autoload` comment before it.

swflint:
It is mine, and if I start getting lots of people using it, I'll update it, but as of now, it's an already set-up api key, including image, name and description.  If you prefer I remove it, I  will.

Fixed both.

alezost:
(I am not a man to decide but) @swflint, I think this shouldn't be done
like this: `pushover-api-key` must not be a constant(!) and its default
value should be nil (or an empty string maybe).  Ideally this variable
(and `pushover-user-key`) should be defined with `defcustom` and the
package documentation should mention that a user needs to set these
variables.

swflint:
Feel free to submit a pull request.  I don't use customize, so writing a `defcustom` might be better suited to someone else.

alezost:
Feel free to submit a pull request.

I'm not going to use "pushover.el".  I just wanted to point that it's
not really ready.

    ... writing a defcustom might be better suited to someone else

This sounds like you made the package only for yourself.  If I was
responsible for MELPA I would reject this package in the current state.

Sorry if what I wrote sounds rude, I just don't like when the things are
not done properly, so when I saw this PR, I couldn't pass by :-)

swflint:
The feel free to submit a PR is because as I don't use customize, I'm not entirely sure what would be entirely necessary.  Someone who does use it would be better at writing a `defcustom`.

Whether or not things are done properly isn't always cut and dry.  I tend to think that customize is a bad thing, thus I don't use it.  Lack of familiarity with a utility makes it hard to write to support it.

vyp:
`defcustom` uses `defvar` as a subroutine, so it works for those that do not use `M-x customize` too. So I think it would be idiomatic? (And obviously just consult the help for more, it's very similar to `defvar`.) Btw I also think `M-x customize` is a bad thing! :)

swflint:
Oh, I'm aware that it uses `defvar` under the hood, just since I don't use it, I'm not inclined to write a `defcustom`, with the whole categorization stuff.

vyp:
Ah I see, my bad. No worries!

swflint:
No problem at all.

purcell:
Great, thanks. See https://github.com/swflint/pushover.el/pull/3 for defcustoms etc.

I'm fine with leaving the API key in - that's your risk, if anyone's. :-)

PR #4030, opened 2016-07-06
Added ssh-deploy recipe

cjohansson:
ssh-deploy enables optional automatic uploads and manual uploads, downloads and diffs via key-paired authorized password-less ssh-connections globally or on a per directory basis via DirectoryVariables. It is useful when working with local files that are deployed to remote hosts.

It uses local scp for transfers (upload/download) and tramp and ediff for diffs. I am the maintainer and creator of the package. I have tested the package with the tests on the https://github.com/melpa/melpa/blob/master/CONTRIBUTING.md page.

https://github.com/cjohansson/emacs-ssh-deploy

### DIFF
--- a/recipes/ssh-deploy
+++ b/recipes/ssh-deploy
@@ -1,0 +1,1 @@
+(ssh-deploy :repo "cjohansson/emacs-ssh-deploy" :fetcher github)


milkypostman:
Here is a prescreening of notes but @purcell can give the final decision on whether this is complex enough.

- you should make an 'ssh-deploy-mode' (minor mode) that enables keybindings rather than installing them by default. or you should describe in your instructions how to just setup the keybindings manually. There is a function in `use-package` that makes it easy to define the keybindings along with the package. 

    in general packages should avoid defining any keybindings themselves. at worst they should provide a key-map as part of a minor mode that can be edited by the user or a 'setup' function that defines a set of default keybindings.

- let people define the after-save hook themselves. possibly provide a function to enable or disable this functionality on a particular buffer.

- you need to make at least one of the functions autoloader by prefixing with `;;;### autoload`. likely the three `-handler` functions should have autoload.

hope that helps.

purcell:
Yep, I agree with all that feedback. Also:
- Camel-case is out in elisp, sorry, so `localRoot` -> `local-root` etc.
- leave blank lines between top-level forms
- You should use a `defgroup` to declare the `ssh-deploy` custom group and to place it appropriately in the appropriate top-level group

You might also check out [firestarter](https://github.com/wasamasa/firestarter), which can be a nice general alternative to rolling your own "do-blah-on-save" functionality.

cjohansson:
Alright thanks, I will fix those items and do another commit soon.

cjohansson:
Hello again, I have done the requested changes now, and I have pushed them to my repo. I also did the contribution tests again just to verify that it didn't broke. 
This is the first time I make a pull request so have patience with me if I have misunderstood you in any way :)

* Removed CamelCase
* Removed key-bindings
* Removed save-hook
* Added blank lines between top-level forms (I'm not sure what top-level forms is but I added blank lines between function definitions and variable definitions)
* Added defgroup
* Added autoloader comment for the handler functions

Regards
Christian

cjohansson:
I corrected the group for the group-definition and replaced double lines of white-space with single lines. Do you mind taking another look?

milkypostman:
thanks!

PR #4033, opened 2016-07-08
omtose-phellack recipe fix

franksn:
recipe fix.


purcell:
This will *only* include the elisp in the `variants` subdir. Is that right? I doubt it.

franksn:
it is correct.:)

purcell:
I applied that too hastily, and then had to revert it. Essentially, you must have a file matching the name of the package/recipe so that the metadata can be located. In your case, since you don't want to bundle `omtose-phellack-theme.el` into the package, you might like to place an `omtose-phellack-theme-pkg.el` in the `variants` subdir: it would contain a `define-package` form.

franksn:
done, thanks a lot.

purcell:
Thanks - I’ve reapplied the changes in this PR, so they’ll take effect the next time the package builds.

PR #4035, opened 2016-07-09
Add recipe for contrast-color-picker

yuutayamada:
A small utility package that finds contrast color against specified color using CIEDE2000 algorithm, which is builtin function in color.el


purcell:
That's cool, but a "color picker" is usually a GUI dialog for choosing a colour, so I find the naming a bit confusing. Maybe just `contrast-color`?

yuutayamada:
> but a "color picker" is usually a GUI dialog for choosing a colour,
Thank you for your advice. I'll rename to `contrast-color` and send another PR.

PR #4036, opened 2016-07-09
package-build's :files is the same as :defaults

vyp:
Feel free to reject this for whatever reason you may have!


milkypostman:
Can't think of a reason. Guess I'll have to accept!

vyp:
Haha no I meant if there was a reason for it in the first place, and if it was still valid (e.g. maybe it's there for CI purposes or something). But all good! 👍👍

PR #4037, opened 2016-07-10
Add recipe for contrast-color

yuutayamada:
I renamed contrast-color-picker (#4035) to this recipe as purcell suggested.


purcell:
Great - thanks! This looks really useful.

PR #4038, opened 2016-07-10
added recipe xah-css-mode

xahlee:
major mode for CSS. Alternative to emacs builtin one. Major diff is that only exact CSS words are colored. (For example, class names are not colored.) And keyword completion for all CSS words.

A direct link to the package repository:
https://github.com/xahlee/xah-css-mode

I'm the author.

Thanks Steve Purcell.

### DIFF
--- a/recipes/xah-css-mode
+++ b/recipes/xah-css-mode
@@ -1,0 +1,1 @@
+(xah-css-mode :repo "xahlee/xah-css-mode" :fetcher github)


purcell:
Thanks, looks interesting!

You'll need to use `Package-Requires` to declare a dependency on `(emacs "24.3")` due to the use of `setq-local`. (`color.el` was available in Emacs 24.1, btw.)

You should use `define-derived-mode` instead of just creating a `xah-css-mode` command: it'll save you a lot of boilerplate (setting mode name, killing local variables etc), and will also generate things like the keymap and hook definitions. As a parent mode, use `prog-mode`.

It's weird to modify the global `xah-css-key-map` while starting the mode in a specific buffer: instead, do that at the top level.

xahlee:
Thanks Steve for spending time looking into!

I fixed all your recommendations. Thanks again. Let me know if there's still issues.

purcell:
Thanks. I sent you a little fix in https://github.com/xahlee/xah-css-mode/pull/1.

xahlee:
fixed. Thank again for the help.

PR #4040, opened 2016-07-11
Buffer Layers is now Buffer Sets

swflint:
I'm renaming my packages to Buffer Sets, and would like to have that fixed in Melpa.


purcell:
Sure. While we're at it. I'd suggest you rename the repo as well, then modify this PR to refer to the new name.

swflint:
Renamed and updated (squashed commit).

purcell:
Super, thanks.

PR #4041, opened 2016-07-11
add emoji-recall

lujun9972:


recipes/emoji-recall
```diff
@@ -0,0 +1 @@
+(emoji-recall :fetcher github :repo "lujun9972/emoji-recall.el" :files ("emoji-recall.el" "emoji-cheat-sheet/*.png"))
```
milkypostman: please change `"emoji-recall.el"` to `:defaults`

milkypostman:
i think that only `emoji-recall-game-start` needs to be autoloaded.

if you are going to use lexical-binding (do you really need to?) then you need to add a `package-requires: ...` for emacs 24.

i *think* the format is, `Package-Requires: ((emacs 24))`

vyp:
@milkypostman I think 24 needs to be in double quotes, but yes.

milkypostman:
thank you, my review skills are rusty. or nonexistent. one of those

lujun9972:
Yes, we do need lexical-binding. so I have added the package requires and changed "emoji-recall.el" to :defaults

PR #4044, opened 2016-07-13
Add recipe for dfmt

qsimpleq:


milkypostman:
i think this package would fit better in the [d-mode](https://github.com/Emacs-D-Mode-Maintainers/Emacs-D-Mode) package. Have you tried reaching out to the maintainers? it would be better as an on-save hook i think.

for example, go-mode has an integration with gofmt. i am ok with a merge but i think it's worth exploring that option first.

other comments,

- `dfmt-region-or-buffer ` and `dfmt-setup-keys` (*maybe* `dfmt-file`) seem to be the only functions that need to be autoloaded.
- `dfmt-region-or-buffer` probably doesn't need an `arg` since it's not used.
- why are the aliases defined for each function?

qsimpleq:
Thanks.

> i think this package would fit better in the d-mode package. Have you tried reaching out to the maintainers? it would be better as an on-save hook i think.

Good. I will try to discuss it with the authors of d-mode. While waiting for.

dfmt-region-or-buffer - my fault.

qsimpleq:
[https://github.com/Emacs-D-Mode-Maintainers/Emacs-D-Mode/issues/73](https://github.com/Emacs-D-Mode-Maintainers/Emacs-D-Mode/issues/73)

The authors are silent 20 days. I think I waited enough and we can issue a separate package dfmt, temporarly.

purcell:
Thanks. Sent you a couple of little fixes in https://github.com/qsimpleq/elisp-dfmt/pull/5.

PR #4045, opened 2016-07-13
update repository location for helm-unicode

bomgar:
Hi,

as discussed here https://github.com/shosti/helm-unicode/pull/3 I'm taking over the maintenance for helm-unicode. I hope this is the right way to do this.


shosti:
:thumsup:

milkypostman:
thanks!

PR #4046, opened 2016-07-14
Adding a recipe for refine

Wilfred:
Refine is a package I've written that makes editing lists easy. I think it's ship-shape and ready for MELPA.

[See the GitHub page for more information](https://github.com/Wilfred/refine).

### DIFF
--- a/recipes/refine
+++ b/recipes/refine
@@ -1,0 +1,1 @@
+(refine :repo "Wilfred/refine" :fetcher github)


purcell:
Thanks!

You'll need to depend on Emacs 24.3 due to your use of `defvar-local` and `setq-local` (`user-error` is pretty new too). Also, `cl-prettyprint` is not in `cl`: you probably want to depend on `cl-lib "0.5"` and import that library instead of the deprecated `cl`.

Wilfred:
OK, those issues should now be sorted. Do let me know if I've missed anything silly :)

purcell:
Looks good, thanks!

PR #4047, opened 2016-07-16
add discourse api

lujun9972:


purcell:
Thanks! Please depend on the latest stable version of `s` (1.11.0), not a snapshot version. And I'd also recommend running `M-x checkdoc` and fixing the suggestions.

lujun9972:
Thank you for reviewing.
I have modified the dependency and have been mending some doc-strings, is it OK now?
Since my English is not good, any fix is welcome.

purcell:
Thanks!

PR #4049, opened 2016-07-17
Add helm-hunks recipe

torgeir:
A helm interface for jumping around and staging git hunks

https://github.com/torgeir/helm-hunks.el

### DIFF
--- a/recipes/helm-hunks
+++ b/recipes/helm-hunks
@@ -1,0 +1,1 @@
+(helm-hunks :fetcher github :repo "torgeir/helm-hunks.el")


purcell:
Thanks - looks handy!

Quick feedback:
- Please depend on a non-zero version of `helm`
- `subr-x` requires Emacs 24.4 (or maybe 24.3, don't recall) -- you need to declare a package dependency on that Emacs version too
- Don't require `cl-macs` for `cl-loop`, use `cl-lib`. (When targeting older Emacsen, `cl-lib` in GNU ELPA provides `cl-loop`, and there is no `cl-macs.el`.)
- You should add an `;;;###autoload` marker for the `helm-hunks` command
- `M-x checkdoc`
- Just because `'git-gutter+-refresh` is `fboundp` doesn't mean `git-gutter` is enabled. Might want to check this still works if the mode is disabled.
- `helm-hunks-refresh-hook` should perhaps be a `defcustom`

purcell:
(Also, you shouldn't be binding user-reserved key sequences like `C-c o`. Check out the (keybinding conventions section of the manual)[https://www.gnu.org/software/emacs/manual/html_node/elisp/Key-Binding-Conventions.html] to see the rules for this.)

torgeir:
> Quick feedback

Thanks! Fixed upstream!

> (Also, you shouldn't be binding user-reserved key sequences like C-c o and C-c p. Check out the keybinding conventions section of the manual to see the rules for this.)

I have changed `C-c p` to `C-c C-p`. 

`C-c o` "other window" (and `C-c C-o` "other frame") are helm "convention", so many other helm plugins use the same keys for this functionality (e.g. [helm-buffers](https://github.com/emacs-helm/helm/blob/master/helm-buffers.el#L157-L158), [helm-projectile](https://github.com/bbatsov/helm-projectile/blob/master/helm-projectile.el#L507), [helm-ag](https://github.com/syohex/emacs-helm-ag/blob/master/helm-ag.el#L422)). Also, these are also bound only when the helm-hunks buffer is open.

purcell:
> Thanks! Fixed upstream!

Nice! Looks good.

> C-c o "other window" (and C-c C-o "other frame") are helm "convention", so many other helm plugins use the same keys for this functionality (e.g. helm-buffers, helm-projectile, helm-ag). Also, these are also bound only when the helm-hunks buffer is open

Oh, that's a shame, because technically that's still bad practice, and what happens then is that the bad practice spreads! I'll merge anyway though. :-)

torgeir:
Alright, thanks! I know, seems we're in too late 😞 

What's a good way to expose such bindings, a var to opt in that is checked before binding to them?

PR #4050, opened 2016-07-19
Add xwidgete recipe

tuhdo:
The repo is here: https://github.com/tuhdo/xwidgete

### DIFF
--- a/recipes/xwidgete
+++ b/recipes/xwidgete
@@ -1,0 +1,1 @@
+(xwidgete :fetcher github :repo "tuhdo/xwidgete")


purcell:
Thanks, but `xwidgete.el` isn't formatted with the standard elisp boilerplate, so it isn't a valid package in its current form. In particular, you need the standard `;;; xwidgete.el --- Your short description here` first line, and a trailing `;;; xwidgete.el ends here` trailing line.

tuhdo:
Done.

vyp:
@tuhdo I think it needs three dashes like `;;; xwidgete.el ---` in the header line, instead of two like `;;; xwidgete.el --`. Not sure though.

tuhdo:
Oh yes it's true 3 dashes are needed. Fixed. Thanks.

purcell:
Thanks, merged. You should also add a `;;; Commentary:` section providing an overview of how to use the package. :-)

PR #4052, opened 2016-07-19
Added evil-mc-extras recipe

gabesoft:
Added ``evil-mc-extras`` to allow for extended functionality for [evil-mc](https://github.com/gabesoft/evil-mc)


purcell:
Nicely packaged, thanks. Merging now.

I suggest you don't provide the turn-on and turn-off functions, though. It's really not idiomatic these days. Just provide the minor mode, and declare an autoload for that.

PR #4053, opened 2016-07-20
Adds system-packages

jabranham:
Wrapper functions for various package managers. Support for several package managers for Arch, Fedora, Debian/Ubuntu, and Mac.


purcell:
Nice, thanks. Please run `M-x checkdoc` over it to help you format the docstrings more idiomatically. You should also add `;;;###autoload` cookie comments for the main interactive entry points.

PR #4054, opened 2016-07-20
widgetjs, minor mode for the widgetjs js library

DamienCassou:
repository: https://github.com/foretagsplatsen/emacs-js

association: I need the package and the maintainer (Nicolas Petton)
asked me to package it on melpa:
https://github.com/foretagsplatsen/emacs-js/issues/1. widgetjs is a
sub-project of emacs-js which will get a separate PR to melpa.

communication: I got write access to the repository and did a few
changes to permit packaging on melpa:
a1e91bdcecdea50c80f5ff87f7a4f7a2c249713e (and the following)

PR #4055, opened 2016-07-20
emacs-js, auto setup of a JavaScript environment

DamienCassou:
- repository: https://github.com/foretagsplatsen/emacs-js
- association: I need the package and the maintainer (Nicolas Petton) asked me to package it on melpa: https://github.com/foretagsplatsen/emacs-js/issues/1
- communication: I got write access to the repository and did a few changes to permit packaging on melpa: a1e91bdcecdea50c80f5ff87f7a4f7a2c249713e (and the following)

emacs-js depends on widgetjs which got a separate PR: https://github.com/melpa/melpa/pull/4054

### DIFF
--- a/recipes/emacs-js
+++ b/recipes/emacs-js
@@ -1,0 +1,1 @@
+(emacs-js :fetcher github :repo "foretagsplatsen/emacs-js" :files ("emacs-js.el"))


purcell:
Thanks. I'm not really keen to merge this submission in its current form. It's a grab-bag of quite personal configuration, rather than a true reusable package. There are no consistent namespace prefixes, it uses key bindings reserved for end users (e.g. "C-c m"), and it generally can't be turned on/off or easily customised. As the comment says,

> ;; Some of of the settings below are specific to Företagsplatsen, but most of it
> ;; should be generic and reusable.

PR #4056, opened 2016-07-20
Add lyrics recipe

marsam:
Hi:
I'd like to add the recipe for [lyrics.el](https://github.com/emacs-pe/lyrics.el)

Thanks in advance.


purcell:
Nice! Thanks.

PR #4058, opened 2016-07-21
Include css for documentation

phillord:
m-buffer is documented directly within Emacs with HTML generated from
the source code. The CSS file defines the style for this HTML.


purcell:
Thanks!

PR #4060, opened 2016-07-22
Add recipe for coati

st4ll1:


purcell:
Thanks. Quick feedback:

- Don't use `;` for full-line comments, use `;;`
- Keywords are empty. You should at least have `external`, to match the `defgroup` parent group
- The `emacs-coati-` prefix should be `coati-` ("emacs" is redundant for identifiers in elisp anyway, right?)
- `X for Emacs` isn't a great package description. "for Emacs" is redundant. How about something like "Communicate with Coati"
- You're using the new `subr-x` library, probably for `string-suffix-p`. This means you need to include a package dependency header, e.g. `;; Package-Requires: ((emacs "24.4"))`
- `(if (null x) (progn y))` is the same as `(unless x y)`
- Don't use `""` as the mode lighter: use something like `" coati"`, and then users can use something like `diminish` if they want to hide it.
- `(message "%s" (concat "Could not process the message from coati" string))` can just be `(message "Could not process the message from coati: %s" string)`
- `M-x checkdoc` will help you make your docstrings well-formed.

Hope that helps!

st4ll1:
Thanks for the feedback. 
It the first time i wrote a emacs plugin.
Your feedback helped a lot.

purcell:
Looks great! Nice work.

PR #4061, opened 2016-07-22
Add php-scratch recipe

mallt:
- php-scratch is a scratch buffer to interactively evaluate php code
- repository url: https://github.com/mallt/php-scratch

### DIFF
--- a/recipes/php-scratch
+++ b/recipes/php-scratch
@@ -1,0 +1,2 @@
+(php-scratch :fetcher github
+             :repo "mallt/php-scratch")


purcell:
Thanks. Quick feedback:
- You're going to need to declare package dependencies for `s` and `php-mode`. -
- Since you're using `user-error`, you'll need to depend on Emacs 24.4 (I think). 
- `local-set-key` is usually a warning sign. Why not define a derived mode of php-mode so that you have a proper key map to modify?

Hope that helps!

Fanael:
> Since you're using `user-error`, you'll need to depend on Emacs 24.4 (I think).

Emacs 24.3.

purcell:
> Emacs 24.3.

Thanks, wasn't sure offhand!

Fanael:
As I said before, `C-h n` is your friend in this sort of situations.

mallt:
Thanks a lot for the feedback! 
I've added the package dependencies and moved the keybindings to a derived mode.

purcell:
Nice! Thanks.

PR #4062, opened 2016-07-23
add recipe kiwix.el

stardiviner:


purcell:
Thanks. Quick feedback:
- Remove "for Emacs" from the package description line: it's redundant
- Please add the standard metadata headers, and a `Commentary:` section illustrating basic usage.
- `'custom-group` is not a valid parent group for the `defgroup`
- The `defcustom`s are missing `:type` specifiers
- If `(getenv "HOME")` returns a path with spaces, the command will be invalid: don't use concat without also using `shell-quote-argument`
- Probably `kiwix-server-command` etc. should be `defcustom`s.
- You use `setq-local` (and `cl-lib`), so you need to declare a package dependency on `(emacs "24.3")`
- The `org`-related setup code looks quite ad-hoc and hard-coded. I feel like it would be better to omit it.

Overall the code is not yet really polished enough to be useful to other people, I think. :-)

stardiviner:
@purcell Thanks for your tips, I modified my code following your suggestions.
Review it. I will modify it. I think the code is better now.

purcell:
Thanks. Sent you a further fix in https://github.com/stardiviner/kiwix.el/pull/1. The `org` code still *really* doesn't belong in there IMO. Maybe include that code in the `Commentary` instead so that users can add it to their configs and match it to their local kiwix settings. In the meantime, I'll go ahead and merge this.

stardiviner:
@purcell I see. I will move it to `Commentary`.

PR #4063, opened 2016-07-23
Add keyswap-mode recipe

hardenedapple:
Add keyswap-mode recipe.

I wrote this package.

https://github.com/hardenedapple/keyswap-mode.el

This mode allows swapping a set of keys on a per-major-mode basis.

e.g. In programming languages it's often useful to switch the numeric keys and their shifted counterparts so the more frequent keys are easier to press.

### DIFF
--- a/recipes/keyswap-mode
+++ b/recipes/keyswap-mode
@@ -1,0 +1,1 @@
+(keyswap-mode :repo "hardenedapple/keyswap-mode.el" :fetcher github)


purcell:
Thanks. Quick feedback:
- Since this provides a minor mode rather than a major mode, this would more conventionally be called `keyswap.el` rather than `keyswap-mode.el`. This would also make your use of the `keyswap-` prefix more consistent with the package name.
- The `lexical-let` from the deprecated `cl` package is unnecessary because you're targeting Emacs 24+ anyway: just declare `lexical-binding` on the first line and you should be good to go.

Otherwise looks good, and handy!

hardenedapple:
Thanks for the pointers -- I've made the changes, but have changed the branch name, commit message, and URL to follow the 'keyswap.el' rather than 'keyswap-mode.el' convention.

Hence I'm closing this pull request and opening up another with the matching names.

PR #4068, opened 2016-07-24
Add `org-commentary` recipe

smaximov:
(**EDIT**: the package name was changed from `org-doc`  to `org-commentary`).

General info:

* Summary: generate or update conventional [library headers][headers] using Org mode files.
* URL: https://github.com/smaximov/org-commentary.
* My association with the package: author/maintainer.

[headers]: https://www.gnu.org/software/emacs/manual/html_node/elisp/Library-Headers.html

### DIFF
--- a/recipes/org-commentary
+++ b/recipes/org-commentary
@@ -1,0 +1,3 @@
+(org-commentary :repo "smaximov/org-commentary"
+                :fetcher github
+                :files (:defaults "bin"))
--- a/recipes/org-doc
+++ b/recipes/org-doc
@@ -1,3 +1,0 @@
-(org-doc :repo "smaximov/org-doc"
-         :fetcher github
-         :files (:defaults "bin"))
--- a/recipes/org-doc
+++ b/recipes/org-doc
@@ -1,0 +1,3 @@
+(org-doc :repo "smaximov/org-doc"
+         :fetcher github
+         :files (:defaults "bin"))


purcell:
Thanks. Some feedback:
- Will this work without the script in `bin`? If not, then this package won't work unless people have `cask` installed and available on their exec-path, which is problematic for the majority of potential users. I'd suggest finding a way to remove that dependency: a number of packages do similar things without needing `cask`.
- `org-doc:` is not a valid prefix: use `org-doc-` throughout instead, as per the [elisp conventions](https://www.gnu.org/s/emacs/manual/html_node/elisp/Coding-Conventions.html).

I've seen a couple of other packages which do the same as this one: have you tried them? How does this compare?

smaximov:
Thank you for the feedback!

> Will this work without the script in `bin`? If not, then this package won't work unless people have `cask` installed and available on their exec-path, which is problematic for the majority of potential users. I'd suggest finding a way to remove that dependency: a number of packages do similar things without needing `cask`.

Valid point. The other way to invoke `org-page` (provided it is already installed via MELPA) from the shell is to run `package-initialize` before loading the dedicated elisp file:

``` bash
$ emacs -Q --batch --eval '(package-initialize)' -l org-doc-cli -f org-doc -- \
    [OPTION]... ORG-FILE ELISP-FILE
```

I have added this information to README ([link](https://github.com/smaximov/org-doc#not-using-cask)).

> `org-doc`: is not a valid prefix

Fixed.

> I've seen a couple of other packages which do the same as this one: have you tried them?

I've found only one similar package &mdash; [org2elcomment](https://github.com/cute-jumper/org2elcomment). Some differences:

- `org-doc` provides CLI in addition to the interactive command.
- `org-doc` can update the `Change Log` (aka `History`) section in addition to `Commentary`.
- `org2elcomment` requires the `Commentary` section to be terminated by `Code`, so it is impossible, for example, to have a `Change Log` section after `Commentary`. `org-doc` fixes that.
- `org-doc` contains changes to the defaults export process (see [org-doc-export.el](https://github.com/smaximov/org-doc/blob/22ba897148f1e61d8ac2fece11e08ac8e2b64a22/org-doc-export.el#L20) for details):
    * `ascii` export back-end, which is used in `org2elcomment` by default, [outputs](https://github.com/cute-jumper/org2elcomment/blob/0e2733f2eede71abc4586909aaf49e4deaf2df1e/org2elcomment.el#L24) an extra block containing project and author names. I find this information redundant, since package headers already have it. Unfortunately, this block is hard-coded into `ascii` backend, so `org-doc` provides a custom backend which excludes that information from export.
    * `org-doc` provides a fix to handle export of custom drawers for Org mode versions prior to 8.3.

purcell:
> Valid point. The other way to invoke org-page (provided it is already installed via MELPA) from the shell is to run package-initialize before loading the dedicated elisp file:

Yes: that would be better, but still not ideal. This is all a compromise because you'd like to provide a separate CLI program, but I feel like that is not as valuable as you think. After all, it will be hidden in an obscure directory under `~/.emacs.d/elpa/...`, so you can't easily use it from the command line. I don't think people will use it, unless perhaps they install `org-doc` *without* `package.el` so that it is always in the same directory.

Re. your other points, thanks for the explanations. I can see how this is distinct from `org2elcomment` etc.

I'd basically be happy to merge this recipe now, but one slight hesitation I have is that `org-doc` is and *extremely* general name, given that `org-mode` is used so extensively for all manner of documentation. I feel like the potential for confusion and name clashes is high. Might it make sense to rename to `org-docstring`?

smaximov:
> After all, it will be hidden in an obscure directory under ~/.emacs.d/elpa/..., so you can't easily use it from the command line.

`(package-initialize)` will take care of adding that directory (in fact, all directories inside `~/.emacs.d/elpa/`) to the load path. You can check this by comparing the output of these two commands:

``` bash
$ emacs -Q --batch --eval '(message "%S" load-path)'
```

``` bash
$ emacs -Q --batch --eval '(package-initialize)' --eval '(message "%S" load-path)'
```

So both ` $ emacs -Q --batch --eval '(package-initialize)' ...` and `$ cask emacs ...` will work identically. The latter variant is for those (myself included) who use Cask for development and don't want to install `org-doc` "globally", e.g. inside `~/.emacs.d/elpa`. As the added benefit, it saves some typing.

> unless perhaps they install `org-doc` _without_ `package.el` so that it is always in the same directory.

This would work if `org-doc` has no dependencies. But `org-doc` depends on `dash.el`, so the user has to install it through `package.el` and add its path to the load path or install it to the same directory.

> `org-doc` is and extremely general name

Well, I'm not happy with that name myself, but that's the best I've come with. Personally, I don't like `org-docstring` because it's rather misleading. `docstring`s have  well defined [meaning](https://www.emacswiki.org/emacs/DocString) in the context of Emacs, and my package has nothing to do with them.

How about you mark this PR as pending until I come up with a better name? Of course, if you have some more suggestions, you're welcome :)

purcell:
> (package-initialize) will take care of adding that directory (in fact, all directories inside of ~/.emacs.d/elpa/) to the load path. You can check this by comparing the output of these two commands:

Yes, but my point was that nobody will easily be able to find the *script* in order to run it from outside Emacs. So I wonder if you need it at all.


> Well, I'm not happy with that name myself, but that's the best I've come with. Personally, I don't like org-docstring because it's rather misleading. docstrings have well defined meaning in the context of Emacs, and my package has nothing to do with them.
> 
> How about you mark this PR as pending until I come up with a better name? Of course, if you have some more suggestions, you're welcome :)

Maybe `org-commentary`?

smaximov:
> Yes, but my point was that nobody will easily be able to find the script in order to run it from outside Emacs

The script is intended to use only with Cask, not to be invoked directly.

> Maybe `org-commentary`?

Thanks, I think I will use this one.

smaximov:
@purcell, I have renamed the package and have changed the recipe, so I believe that it's ready to be merged now.

purcell:
Awesome, thanks!

PR #4069, opened 2016-07-25
add helm-dired-history

jixiuf:


purcell:
Thanks, looks good. Please also declare a dependency on `(cl-lib "0.5")`, since you're using that directly.

jixiuf:
> declare a dependency on (cl-lib "0.5")


done

purcell:
Great, thanks!

PR #4070, opened 2016-07-25
update ivy-erlang-complete for multiple files

s-kostyaev:
For now I think that parse erlang source code with escript is more correct than with elisp + findutils. So I add one escript for parsing erlang records with types and ignoring comments. Maybe I add more parser scripts later.


purcell:
Thanks.
- The `:files` pattern herre can simply be written `(:defaults "bin")`.
- Github is reporting conflicts in this PR. Not sure why - could you take a look please?

s-kostyaev:
Thanks for comment. Fixed.

purcell:
Great, thanks. Sorry about the slow turnaround.

PR #4071, opened 2016-07-25
Add bool-flip recipe

michaeljb:
**Bool-flip** provides a function to flip the boolean symbol under point (e.g., replace "true" with "false" and vice versa), and a function to add a "boolean pair".

I wrote this package.

https://github.com/michaeljb/bool-flip

### DIFF
--- a/recipes/bool-flip
+++ b/recipes/bool-flip
@@ -1,0 +1,1 @@
+(bool-flip :repo "michaeljb/bool-flip" :fetcher github)


purcell:
Thanks - useful idea!

The implementation feels somewhat overly complex and not very idiomatic to me, though. The hash arrangement keyed by major mode is basically just an ad-hoc re-implementation of buffer-local variables. Instead, you should just have a `defcustom` list of pairs, mark it as safe for buffer-local overriding, and then people can modify it as needed in their major mode hooks.

But in any case, why make this mode specific at all? It sees like in all the pair cases you've added by default, you could simply look at the symbol-at-point and unambiguously find its counterpart.

Hope that helps!

tarsius:
@michaeljb do you intend to address the raised concerns and get this into Melpa, or should we close this?

Wilfred:
I'm very interested in this package, I'm happy to send a PR to get it shipshape.

tarsius:
Great, thanks!

Wilfred:
@purcell I've taken a quick look at the code, and I can't see how to make your proposal work.

`bool-flip-pair` has a problem with `t`: in emacs-lisp-mode, we want to toggle that with `nil`, but in Scheme we want to toggle `t` with `f`. Now strictly speaking, a boolean literal in Scheme is `#t` and `#f`, but you occasionally see `t` and `f` with SQL.

I'm also not sure that it makes sense as buffer-local variables. This configuration should apply for all buffers in a major mode, whereas buffer-local variables are set per-buffer. There's precedent for this design: smartparens does the same thing (see the `sp-pairs` variable).

Ultimately, this package works and it's a good citizen (not trampling global settings/keybindings). It's also <100 LOC and AFAICS this is the author's first experience of contributing. How much refactoring do we want to demand from newbies?

purcell:
> bool-flip-pair has a problem with t: in emacs-lisp-mode, we want to toggle that with nil, but in Scheme we want to toggle t with f. Now strictly speaking, a boolean literal in Scheme is #t and #f, but you occasionally see t and f with SQL.

Maybe cycling between 3 values would actually be nice...? But yes, I see what you mean.

> I'm also not sure that it makes sense as buffer-local variables. This configuration should apply for all buffers in a major mode, whereas buffer-local variables are set per-buffer.

Yes, so you add a function which sets the pairs locally to the hook for your major mode.

> There's precedent for this design: smartparens does the same thing (see the sp-pairs variable).

Smartparens' customisation is absolutely not a role model here, I'm afraid. It's very un-emacsy to configure a mode by calling functions like `sp-local-pair` which rummage around in complex data structures if the same thing can be accomplished using simple lists which are then amenable to saving with the customization system, or overriding locally.

So here, `bool-flip-pair` is less idiomatic and flexible than a `bool-flip-pairs-alist` or similar.

> Ultimately, this package works and it's a good citizen (not trampling global settings/keybindings). 
> It's also <100 LOC and AFAICS this is the author's first experience of contributing. How much refactoring do we want to demand from newbies?

I agree, but I guess what I'm looking for is a maintainer who is responsive to comments, since we're kicking off a relationship of distributing their package. This early stage is a great chance to ensure the API of a new package is idiomatic and widely useful in time for early adopters to have a great experience with it.

purcell:
(And fwiw, I am interested in using this package too!)

tarsius:
> I agree, but I guess what I'm looking for is a maintainer who is responsive to comments, since we're kicking off a relationship of distributing their package.

I second this, which isn't to say that this package cannot be made fit for Melpa. @michaeljb, one can usually get some mentoring here, which is a great learning experience. Are you interested in that?

michael-brandt-cu:
First off, apologies for not giving this thread any attention, I've just been busy with other things in my free time lately.

I would be interested in improving this package, and receiving some mentoring in Emacs Lisp. Mostly I've learned just enough to customize my Emacs and maybe tweak snippets from others' setups.

> I agree, but I guess what I'm looking for is a maintainer who is responsive to comments, since we're kicking off a relationship of distributing their package. 

I'll be more responsive now.

> But in any case, why make this mode specific at all? 

What @Wiflred said.

> Maybe cycling between 3 values would actually be nice...?

I disagree. It's true that cycling between 3 wouldn't be *that* much worse, but I'd be much happier with just 2.

> So here, `bool-flip-pair` is less idiomatic and flexible than a `bool-flip-pairs-alist` or similar.

So, something like this for the base case?

```elisp
(setq bool-flip-pairs-alist
      '((T . F)
        (F . T)))
```

One of the reasons for `bool-flip-pair` is to avoid writing the pair twice; I'd rather do `(bool-flip-pari "y" "n")` than `(append bool-flip-pairs-alist '((y . n) (n . y)))`. Would a helper function like `bool-flip-pair` be acceptable if it's just adding to the alist, or would it just always better to write the pair both ways?

I don't have much experience with buffer-local variables or mode hooks, so I could use some pointers there as well.

Thanks.

michaeljb:
First off, apologies for not giving this thread any attention, I've just been busy with other things in my free time lately.

I would be interested in improving this package, and receiving some mentoring in Emacs Lisp. Mostly I've learned just enough to customize my Emacs and maybe tweak snippets from others' setups.

> I agree, but I guess what I'm looking for is a maintainer who is responsive to comments, since we're kicking off a relationship of distributing their package. 

I'll be more responsive now.

> But in any case, why make this mode specific at all? 

What @Wiflred said.

> Maybe cycling between 3 values would actually be nice...?

I disagree. It's true that cycling between 3 wouldn't be *that* much worse, but I'd be much happier with just 2.

> So here, `bool-flip-pair` is less idiomatic and flexible than a `bool-flip-pairs-alist` or similar.

So, something like this for the base case?

```elisp
(setq bool-flip-pairs-alist
      '((T . F)
        (F . T)))
```

One of the reasons for `bool-flip-pair` is to avoid writing the pair twice; I'd rather do `(bool-flip-pari "y" "n")` than `(append bool-flip-pairs-alist '((y . n) (n . y)))`. Would a helper function like `bool-flip-pair` be acceptable if it's just adding to the alist, or would it just always better to write the pair both ways?

I don't have much experience with buffer-local variables or mode hooks, so I could use some pointers there as well.

Thanks.

purcell:
> So, something like this for the base case?
> 
> (setq bool-flip-pairs-alist
>       '((T . F)
>         (F . T)))

Yes, exactly! :-)

> One of the reasons for bool-flip-pair is to avoid writing the pair twice; I'd rather do (bool-flip-pair "y" "n") than (append bool-flip-pairs-alist '((y . n) (n . y))). Would a helper function like bool-flip-pair be acceptable if it's just adding to the alist, or would it just always better to write the pair both ways?

Technically, it depends what you do with the helper: if you declare `bool-flip-pairs-alist` as a `defcustom` (as you should), then in some cases users will have customised that var before `bool-flip` is loaded. This means you would have to avoid using `bool-flip-pair` at the top-level of your code to define the default pairs, because it would munge the user's preferences.

For this reason, I would personally suggest avoiding the helper and just keeping things explicit and transparent.

michaeljb:
I'm having trouble converting the text at the point to a key that works with the alist. Using `(thing-at-point 'symbol)` worked fine with my initial hash setup but I just keep getting `nil` from `alist-get`.

I tried experimenting in the scratch buffer with this function:

```emacs-lisp
(defun message-my-alist-value ()
  (interactive)
  (let (my-alist my-var to-message)
    (setq my-alist '((T . F) (F . T)))

    (setq my-var (thing-at-point 'symbol))
    (message my-var)

    (setq to-message (alist-get my-var my-alist))
    (message to-message)))
```

I executed `message-my-alist-value` with `T` at point, and `(message my-var)` adds `T` to the Messages buffer, but `(message to-message)` doesn't add anything to the Messages buffer.

Outside of this function, I could store a symbol `T` in a variable, and use that variable with `alist-get`. `thing-at-point` returns a string, and so I tried using the result of `make-symbol` as the key but that still didn't work.

Haven't had any luck searching for the answer, hoping someone here can help!

purcell:
Easy. This modified version will help you see what's going on:
```
(defun message-my-alist-value ()
  (interactive)
  (let (my-alist my-var to-message)
    (setq my-alist '((T . F) (F . T)))

    (setq my-var (thing-at-point 'symbol))
    (message "Symbol is: %S" my-var)

    (setq to-message (alist-get my-var my-alist))
    (message "Retrieved from alist: %S" to-message)))
```

In short, `thing-at-point` returns a string, while your alist keys are symbol. Maybe use strings in the alist? Alternative would be to use `intern` or `intern-soft` to convert the strings to symbols.

michael-brandt-cu:
Thanks for the tip, I'll take another look this weekend.

michaeljb:
Got it working in https://github.com/michaeljb/bool-flip/pull/1, just needed the functions `intern` and `symbol-name`.

I couldn't find a function like `stringp` for an `alist`, but `listp` returns `t` for objects of `alist` type, so I thought it would work as the `:safe` argument for `defcustom`.

tarsius:
Try this:

```lisp
(require 'cl-lib)

(defcustom bool-flip-alist
  '(("T"    . "F")
    ("t"    . "f")
    ("TRUE" . "FALSE")
    ("True" . "False")
    ("true" . "false")
    ("Y"    . "N")
    ("y"    . "n")
    ("YES"  . "NO")
    ("Yes"  . "No")
    ("yes"  . "yes")
    ("1"    . "0"))
  "List of values flipped by `bool-flip-do-flip'."
  :group 'bool-flip
  :safe 'listp)

;;;###autoload
(defun bool-flip-do-flip ()
  "Replace the symbol at point with its boolean opposite."
  (interactive)
  (let ((old (thing-at-point 'symbol)))
    (if old
	(let ((new (or (cdr (assoc  old bool-flip-alist))
		       (car (rassoc old bool-flip-alist)))))
	  (when new
	    (cl-destructuring-bind (beg . end)
		(bounds-of-thing-at-point 'symbol)
	      (let ((insert-after (= (point) beg)))
		(delete-region beg end)
		(insert new)
		(when insert-after
		  (goto-char beg))))))
      (user-error "Nothing to flip here"))))
```

michaeljb:
Thanks @tarsius!

I didn't know that `or` returned the first non-nil expression, instead of just returning `t`. `cl-destructuring-bind` is also new to me.

I added your changes to the PR I linked earlier.

michaeljb:
I merged michaeljb/bool-flip#1, if there's any more suggested changes I figure I can make them directly on master until this is actually released.

tarsius:
Looks good to me now ;-)

PR #4073, opened 2016-07-25
add evil-visual-replace recipe

troyp:
Pull Request Summary:
-------------------------------
* __package summary:__ Provides versions of the `query-replace` <kbd>(M-%)</kbd> and `replace-regexp` <kbd>(C-M-%)</kbd> commands which work correctly for evil-mode visual blocks (rectangular regions).
* __URL:__ https://github.com/troyp/evil-visual-replace

### DIFF
--- a/recipes/evil-visual-replace
+++ b/recipes/evil-visual-replace
@@ -1,0 +1,1 @@
+(evil-visual-replace :fetcher github :repo "troyp/evil-visual-replace")


purcell:
Thanks - looks good, but could you please use the prefix `evil-visual-replace-` instead of `evil-virep-` so that it matches the package name? It's always safer to ensure the two match.

troyp:
Okay, I've done that.

Do you really mean the prefix, though? I know I used the wrong feature name in the `provide` call. It should have matched the package name, but didn't. The prefix-name, though... - it's common for packages to have a shorter prefix name (yasnippet, dash, evil-nerd-commenter, ...). If your functions are going to be called a lot from lisp, as in a library, you want a short prefix, but may want a longer (meaningful) package name.

Anyway, for this package there's no need for a short prefix, so I've changed both the feature name and prefix to match the package name.

purcell:
Thanks, merged.

I think it's good to default to matching the prefix with the package name in almost every case, because community conventions help us all work more smoothly together.

`yas/` is definitely not a good prefix either: `yasnippet-` wouldn't have been ridiculously long there. `evilnc` breaks the rule, but then `evilnc` would arguably have been a better package name. Similarly, `evil-virep` would have been a decent package name in this case.

Don't get me started on `mc/` for `multiple-cursors`... if we could only go back in time and tidy that up...!

PR #4074, opened 2016-07-27
Added eloud screen reader package to recipes

smythp:
Please let me know if I need to take any other steps to add the package to melpa. Thank you!

### DIFF
--- a/recipes/eloud
+++ b/recipes/eloud
@@ -1,4 +1,4 @@
 (eloud
  :fetcher github
- url "https://github.com/smythp/eloud"
- repo "smythp/eloud")
+ :url "https://github.com/smythp/eloud"
+ :repo "smythp/eloud")
--- a/recipes/eloud
+++ b/recipes/eloud
@@ -1,0 +1,4 @@
+(eloud
+ :fetcher github
+ url "https://github.com/smythp/eloud"
+ repo "smythp/eloud")
--- a/recipes/eloud
+++ b/recipes/eloud
@@ -1,4 +1,3 @@
 (eloud
  :fetcher github
- :url "https://github.com/smythp/eloud"
  :repo "smythp/eloud")


troyp:
I think your recipe's wrong. `:url` and `:repo` need to be keywords, ie. they should start with a colon.

smythp:
Added the colons to the keyword arguments. Thank you!

smaximov:
It may sound like nitpicking, but the `:url` keyword is redundant. Quoting README:

> In the case of the `github` fetcher, use `:repo` instead of `:url`; the git URL will then be deduced.

smythp:
No, that's okay. Removed the redundant keyword argument. Actually saw that but thought I'd err on the side of verbosity.

purcell:
Thanks, looks good. Some feedback:
- *All* top-level symbols should have the `eloud-` prefix, as per the [elisp conventions](https://www.gnu.org/s/emacs/manual/html_node/elisp/Coding-Conventions.html), so that means identifiers like `hook-map` and `hyphen-start-p` are out
- Your `defcustom`s should have a `:type` and a parent group, declared with `defgroup`.
- Please run `M-x checkdoc` and follow its feedback about your docstrings.
- Instead of the deprecated `cl` library, declare a package dependency on `cl-lib` (version 0.5) and use that instead.
- Whenever you find yourself declaring a `-toggle` function, what you really want is a minor mode instead!
- You're using `advice-add` etc, but `nadvice` was only added in Emacs 24.4, so you'll need to declare a package dependency on that Emacs version
- Are you sure you need to use advice, and not just `post-command-hook`? The latter would be much cleaner if viable.

Hope that helps!

P.S. In general, `flycheck` with `flycheck-package` would have provided some of the above feedback. :-)

smythp:
Thank you for taking the time to give me this detailed feedback. This is my first Emacs package, so this has been a great help. Didn't realize that Flycheck was the secret weapon for adhering to elisp style. :)

I have:

- Removed redundant "for Emacs" in description
- Prefixed all functions and variables with "eloud-"
- Added appropriate :type and :group keyword args to my customize variables
- Only used 'cl library in two places, so refactored to remove it as a dependency
- Removed toggle function and put toggle logic in the body of the minor mode definition
- Added dependency for Emacs 24.4 in header
- Fixed all errors in checkdoc and flycheck

I can't be absolutely sure that advice is the way to go. However, I did originally try to implement this package with hooks instead of advice, and couldn't get any traction. Hooks really aren't meant for this kind of wide-ranging overhaul, their timing is arbitrary based on when they were inserted and the general after-command hooks don't have any easily accessible way of referring to the last command. Advice is messy, but it's powerful enough to work around timing issues, since you can add speech logic either before or after calling the advised function, or some combination of both. Advice does add whole classes of bugs, and it's been a pain to use, so I would avoid it and use hooks if I could. I may refactor to use more hooks for cases like switching buffers, since that might work more universally than what I have now. 

Thanks again for being so constructive in your response. The code is a lot tidier and better documented for your input, and now I know which tools to use to clean up my next package! :)

purcell:
Yay, nice work!

> Hooks really aren't meant for this kind of wide-ranging overhaul, their timing is arbitrary based on when they were inserted and the general after-command hooks don't have any easily accessible way of referring to the last command.

While your `pre-command-hook` or `post-command-hook` function is running, I believe you can refer to the variables `last-command` and `this-command`. One big advantage of hooks is that you can add functions to the *buffer-local* version of a hook, and thus allow your minor mode to be turned on only in specific buffers.

In any case, I'll go ahead and merge now, but definitely dig a bit more into those hooks, because I think the result would be a lot cleaner if you can get it to work.

smythp:
Darn! this-command was what I was looking for all along. I knew about last-command, but couldn't find that one. Well, fooey.

Yes, it may actually be worth it to refactor to use hooks. I may need to let the package sit for awhile, though, since it will have to be a fairly complete rewrite. Thanks again for the advice and the feedback. In OS and the internet in general, one occasionally encounters people that are either competent or good-natured. It's a real pleasure to run into someone who is both... :)

purcell:
Happy to help. :-)

PR #4075, opened 2016-07-28
Added a recipe for ros-helm

davidlandry93:
This package interfaces the ROS middleware with helm.

It can be found [here](https://www.github.com/davidlandry93/ros-helm).

This is my first contribution and I am available for corrections.

Thanks!

### DIFF
--- a/recipes/helm-ros
+++ b/recipes/helm-ros
@@ -1,0 +1,1 @@
+(helm-ros :fetcher github :repo "davidlandry93/helm-ros")
--- a/recipes/ros-helm
+++ b/recipes/ros-helm
@@ -1,1 +1,0 @@
-(ros-helm :fetcher github :repo "davidlandry93/ros-helm")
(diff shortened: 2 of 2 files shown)


purcell:
Thanks! Quick feedback:
- Please depend explicitly on `(cl-lib "0.5")`
- `helm-ros-` is probably more idiomatic, based on existing helm package names
- The prefix here should be `ros-helm-` throughout, not `ros-`, `ros-helm/` or `ros-helm//` (you can use `ros-helm--` for private identifiers)
- Don't `global-set-key` from this package. Instead, provide a minor mode which users can toggle on and off in order to get those bindings.
- `(add-to-list 'auto-mode-alist '("\\.launch\\'" . nxml-mode))` seems completely unrelated and should be removed

Hope that helps!

davidlandry93:
Thanks for the comments. I implemented your suggestions. I renamed the repository to helm-ros.

purcell:
Thanks. `ros-process-mode-map` and `ros-process-mode-hook` are apparently unused and can/should be deleted.

PR #4076, opened 2016-07-29
Add recipe for suggest.el

Wilfred:
suggest.el is a new package I've written that suggests functions that return the value you want.

suggest_el

[GitHub repo](https://github.com/Wilfred/suggest.el).

### DIFF
--- a/recipes/suggest
+++ b/recipes/suggest
@@ -1,0 +1,1 @@
+(suggest :repo "Wilfred/suggest.el" :fetcher github)


purcell:
Nice, thanks. Merging now, but can I please strongly suggest (no pun intended) you consider never suggesting function from the deprecated `cl` library. Each function there has an equivalent in `cl-lib`, which would be a more responsible suggestion. :-)

Wilfred:
Good point. I've updated the code as suggested (to `cl-first` etc) :)

PR #4077, opened 2016-07-29
add recipe for mode-on-region

miketz:
I'd like to add my package "mode-on-region" to melpa.  
It's for working with code in multi-language files. You select a region of text. The region is opened in a new buffer with a dedicated language mode. Edits are copied back to the original buffer.

The github repo is at https://github.com/miketz/mor

I'm the only author.

### DIFF
--- a/recipes/mode-on-region
+++ b/recipes/mode-on-region
@@ -1,0 +1,1 @@
+(mode-on-region :repo "miketz/mor" :fetcher github)


smaximov:
IMHO, it's better not to use <kbd>C-x C-c</kbd> as the prefix for keybindings, since that combination stands for `save-buffers-kill-terminal` in the vanilla GNU/Emacs.

Fanael:
Isn't this package basically the same as [edit-indirect](https://github.com/Fanael/edit-indirect)?

purcell:
> Isn't this package basically the same as edit-indirect?

Yes, it appears to be, in which case I would prefer not to add this new package.

PR #4078, opened 2016-07-29
Add recipe for cubicaltt

david-christiansen:
This recipe builds the Emacs major mode for cubical, an implementation of cubical type theory developed at Chalmers.

I am neither the primary developer of cubicaltt nor this mode, but I did get permission from the developer to add it to MELPA. I'm also willing to update the recipe should it prove necessary, but this has never happened with the other recipes that I have on MELPA. I have also contributed a couple of features to the mode.

This was discussed at https://github.com/mortberg/cubicaltt/pull/45 , and the package headers were added at https://github.com/mortberg/cubicaltt/pull/46 .

CC @mortberg

### DIFF
--- a/recipes/cubicaltt
+++ b/recipes/cubicaltt
@@ -1,0 +1,4 @@
+(cubicaltt
+ :fetcher github
+ :repo "mortberg/cubicaltt"
+ :files ("cubicaltt.el"))


purcell:
Thanks. Quick feedback:
- `:files` is redundant here, right?
- Most of the code in `cubicaltt.el` uses the prefix `ctt-`, but the two names should really match. We should rename one or the other.
- The mode name string `"ctt mode"` should be `"ctt"`, and then the redundant `(setq mode-name "ctt")` form can be deleted
- Code like `(concat ":cd " dir "\n")` will break if paths contain spaces. Always use `shell-quote-argument` or something similar.
- `*identifier*` is not idiomatic in elisp. Just `identifier` is preferred.
- `(require 'newcomment)` is redundant, since `comment-dwim` is autoloaded
- Relatedly, there's no point rebinding and overriding `comment-dwim`: the major mode can just set `comment-start` and `comment-end` locally instead, then `comment-dwim` will Just Work™ automatically

Hope that helps!

david-christiansen:
Hi @purcell, thanks!

I'll go implement most of those suggestions.

WRT `:files`, I added that because the mode is in a repo that is primarily dedicated to the compiler, and I don't want it to slurp in any extra files that e.g. end with `.info`. Is that not the right way?

WRT `:cd`, the underlying compiler actually doesn't want quotes, and filenames with spaces work without quoting.

I'll send a PR to `cubical` soon, then comment here.

purcell:
> WRT :files, I added that because the mode is in a repo that is primarily dedicated to the compiler, and I don't want it to slurp in any extra files that e.g. end with .info. Is that not the right way?

Okay, that sounds reasonable. :-)

> WRT :cd, the underlying compiler actually doesn't want quotes, and filenames with spaces work without quoting.

Great - just as long as somebody has considered that case!

david-christiansen:
@purcell : I've improved the comment about `:cd` and implemented the other suggestions at https://github.com/mortberg/cubicaltt/pull/50 .  Thanks!

purcell:
Thanks. Sorry for the slow turnaround.

david-christiansen:
I actually didn't expect this to get merged until the PR over at cubicaltt
with the Elisp updates got merged.

PR #4079, opened 2016-07-30
add flyspell-correct-ido recipe

d12frosted:
Simple as that 😸 

CC @jpkotta (via https://github.com/d12frosted/flyspell-correct/pull/25)

### DIFF
--- a/recipes/flyspell-correct-ido
+++ b/recipes/flyspell-correct-ido
@@ -1,0 +1,3 @@
+(flyspell-correct-ido :repo "d12frosted/flyspell-correct"
+                      :fetcher github
+                      :files ("flyspell-correct-ido.el"))


purcell:
Thanks. Every Emacs has `ido`, so it doesn't make sense to distribute this separately from the core of `flyspell-correct`: its dependencies are no different. Maybe instead submit a PR to add this file to the `flyspell-correct` recipe?

purcell:
Bonus: you can then make `flyspell-correct` do something useful out of the box, by making the IDO interface the default. In fact, there should be a default in any case which simply calls "completing-read" -- most completion packages try to override that function anyway.

d12frosted:
Thanks for the review, @purcell.

Everything you said makes perfect sense. 💯

PR #4080, opened 2016-07-30
Add recipe for all-the-icons

domtronn:
https://github.com/domtronn/all-the-icons.el

Hi there, this recipe adds a helper package for using Icon Fonts within Emacs which can be properly propertized and formatted like any other text, for things like mode lines, file trees, tabs etc.

I've added in the `ttf` font files as part of the repo, and linked through to their respective licenses, all of which allow for distribution and re-use.

I am the maintainer of the package.

### DIFF
--- a/recipes/all-the-icons
+++ b/recipes/all-the-icons
@@ -1,4 +1,4 @@
 (all-the-icons
- :repo "domtronn/all-the-icons"
+ :repo "domtronn/all-the-icons.el"
  :fetcher github
  :files ("all-the-icons.el" ("data" "data/*.el")))
--- a/recipes/all-the-icons
+++ b/recipes/all-the-icons
@@ -1,0 +1,4 @@
+(all-the-icons
+ :repo "domtronn/all-the-icons"
+ :fetcher github
+ :files ("all-the-icons.el" ("data" "data/*.el")))
--- a/recipes/all-the-icons
+++ b/recipes/all-the-icons
@@ -1,4 +1,4 @@
 (all-the-icons
  :repo "domtronn/all-the-icons.el"
  :fetcher github
- :files ("all-the-icons.el" ("data" "data/*.el")))
+ :files (:defaults "data"))


purcell:
Thanks! Quick feedback:
- The `:files` spec can more concisely be written `(:defaults "data")`
- There's no `Commentary` describing how to use this code. We pull package descriptions from this section, and once installed, it's where curious users would look to learn more.
- Please depend on the stable versions of packages, not snapshots
- Why would this package depend on `flycheck`? That seems bogus.
- Don't use the abbreviated prefix `ati` or `/`: use `all-the-icons-` (or `all-the-icons--` for private symbols), as per the elisp conventions
- `s/geneate/generate/` ;-)
- Hard-coding all the colour stuff is quite weird -- how would people theme this? The icons come from fonts, so can be styled with faces, which can then be themed. I'd suggest separating the colour aspect from the icon selection, because it's easy for users to colourise icons.
- Running `(locate-dominating-file dir ".gitmodules")` twice successively is redundant and likely to be slow
- The command `"cat %s | grep submodule | awk '{print $2}' | grep %s"` is clunky & non-portable: just write a quick elisp function which opens the file and searches for the text you want.

Hope that helps!

domtronn:
Awesome, thanks for the feedback! :) I've addressed all but one of the above, but have a couple of questions...

> Don't use the abbreviated prefix `ati` or `/`: use `all-the-icons-` (or `all-the-icons--` for private symbols), as per the elisp conventions

I've renamed all of these to `all-the-icons` and used dashes to distinguish private symbols. But I was just wondering if you can use a shortened prefix _(i.e. `ati`)_ for some functions, like `all-the-icons-icon-for-buffer -> ati-icon-for-buffer`? Just to give users a short hand?

> Hard-coding all the colour stuff is quite weird -- how would people theme this? 

I agree. However, one of my main motivations to create this package was to mimic the [File Icons for Atom](atom.io/packages/file-icons) package, and I wanted to maintain the colour variants between two files which share the same icon - _.e.g_ between `.haml` & `.html` or JavaScript and NodeJS files - and I was unsure how best to allow people to alter the mappings between a file match and these colours. Felt easier to just not give them the choice :P

As a minimum, I would want to convert all those colour variables into faces? 
I could also look at generating the dark and light variants of each colour?

The icon macros, _(e.g. `all-the-icons-octicon`)_ can take a colour hex string, I was thinking I could add an optional colour argument to the icon for buffer function which overrides the one in the alist. I could then expand that to allow you to pass in a face so people could call,

```elisp
(all-the-icons-octicon "file-binary" 'font-lock-symbol)
;; or
(all-the-icons-icon-for-buffer "#345465")
(all-the-icons-icon-for-buffer 'font-lock-symbol)
``` 

I'll have a bit of a play, but any suggestions/thoughts would be greatly appreciated! Thanks for your help and time :)

purcell:
Sorry for the slow reply here...

> But I was just wondering if you can use a shortened prefix (i.e. ati) for some functions, like all-the-icons-icon-for-buffer -> ati-icon-for-buffer? Just to give users a short hand?

No, that's not good style, but you could maybe just rename the package to "allicons"?

> As a minimum, I would want to convert all those colour variables into faces?

Yes, I think so. To make an icon you would then just propertise it with the font family and the corresponding face. Your snippet which looks like

```el
`(:family ,family :height ,height :foreground ,col)
```
would then look more like
```
`(:family ,family :height ,height :inherit ,other-face)
```
where `other-face` was the name of one of the defined faces, with foreground colours for both light and dark backgrounds.

This would allow people to then customise those faces in themes.

> I could also look at generating the dark and light variants of each colour?

Each face can include light and dark variants, so you get that for free. :-)

domtronn:
Fantastic thanks! I've updated it this morning and broken all of the hard coded colours into a separated file _(`all-the-icons-faces.el)_` containing faces with light and dark variants, so the colours should now be themeable. :)

I've also updated the icon function calls to take the icon name followed by a plist, so that people can apply faces to it directly, _e.g._ `(all-the-icons-octicon "icon-name" :face 'some-face)`

purcell:
Great, thanks! Merging this now. Couple of final pieces of feedback:
- I'd be slightly concerned about `define-icon` clashing with another package at some point
- Probably `'appearance` rather than `'tools` should be the parent customisation group
- The submodule checking feels pretty clunky and out-of-place here. Pretty sure you can use something in `vc-git` to do this more easily.

PR #4081, opened 2016-07-30
Add gg-find-def recipe

mallt:
- gg-find-def is a function to find the definition of a symbol in a git repository using git grep
- repository url: https://github.com/mallt/gg-find-def

### DIFF
--- a/recipes/gg-find-def
+++ b/recipes/gg-find-def
@@ -1,0 +1,2 @@
+(gg-find-def :fetcher github
+             :repo "mallt/gg-find-def")


purcell:
Thanks. This is interesting. It's also very similar to @jacktasia's [dumb-jump](https://github.com/jacktasia/dumb-jump/blob/master/dumb-jump.el) so I'm hesitant to merge this because it would arguably be better for everyone if you simply joined forces. You're using `git-grep`, and dumb-jump uses `ag`, but otherwise the goals look the same.

You should also both check our `xref-js2`. If you're comfortable targeting an even newer Emacs version, you might consider building on top of `xref`, because that's essentially what you're both trying to do. (See also my `elisp-slime-nav`, which works both with `xref` and the older `find-tag-marker-ring` scheme.)

mallt:
That makes sense, I'll close this PR. Thanks a lot for your insights!

PR #4083, opened 2016-07-31
add ido interface file to flyspell-correct recipe

d12frosted:
Following #4079.


purcell:
Great - thanks!

d12frosted:
@purcell Thanks for the merge! 

Cheers

PR #4084, opened 2016-07-31
Add composer recipe

zonuexe:
- PHP [Composer](https://getcomposer.org/) interface
- https://github.com/zonuexe/composer.el
- I am author


purcell:
Thanks. You should probably add at least one `;;;###autoload`... :-)

zonuexe:
@purcell Thanks!

PR #4085, opened 2016-07-31
Add recipe for nicolaspetton/jade

NicolasPetton:
Jade connects to a browser tab or nodejs process and provides many features for JavaScript development, including a REPL (with auto completion) & object inspection, an inspector, with history and navigation, and even a stepping Debugger, similar to edebug, or cider.

- [x] I am the author of the package
- [x] Package builds and installs properly
- [x] Package fetched from official source (on [github](https://github.com/nicolaspetton/jade))

### DIFF
--- a/recipes/jade
+++ b/recipes/jade
@@ -1,0 +1,1 @@
+(jade :fetcher github :repo "NicolasPetton/jade")


purcell:
Nice! Thanks. :-)

Issue #3905, opened 2016-05-18
Getting an Emacs wiki package on melpa stable


DamienCassou:
Is there any way to get an Emacs-wiki-hosted package into melpa stable?

purcell:
Nope, sorry. It'd have to be moved to a proper SCM with tag support.

PR #4086, opened 2016-07-31
move flymake-cursor source from wiki to github

ryoung786:
MELPA currently sources ```flymake-cursor``` from the emacswiki, which as I understand it makes it impossible to tag and get into MELPA Stable.  I'm trying to transition over to using just MELPA Stable for my packages, and I happen to use this package quite a bit (it's great!).  My hope is that this pull request gets merged, and then since the ```emacs-flymake-cursor``` [repo](https://github.com/illusori/emacs-flymake-cursor) already has a tag (1.0.2) it'll automatically get included in the next melpa stable build.

I hope that makes sense - I'm not the author/maintainer of the ```emacs-flymake-cursor``` github repo, so I'm not sure if this is the correct way forward.  I searched around previous issues and found https://github.com/melpa/melpa/issues/3905

Also maybe worth noting, this github repo is the one that marmalade uses to provide the package.

### DIFF
--- a/recipes/flymake-cursor
+++ b/recipes/flymake-cursor
@@ -1,2 +1,2 @@
-(flymake-cursor :fetcher wiki)
+(flymake-cursor :repo "illusori/emacs-flymake-cursor" :fetcher github)
 


purcell:
Thanks - this makes sense, and I will merge it, though the EmacsWiki version was updated more recently, so this PR will cause the MELPA version number to drop. I'm also a bit concerned that the @illusori repo is unmaintained, since there are 2 pull requests which have not been responded to: https://github.com/illusori/emacs-flymake-cursor/pulls

(As a more general point, definitely consider using `flycheck` instead of `flymake`. I switched ages ago and haven't looked back.)

illusori:
My forks of Flymake and related stuff are up for adoption with my blessing if anyone wants them, they're unmaintained and likely to remain so - I've got no free time for open-source work at the moment and have moved over to Flycheck for my day job.

purcell:
Thanks Sam. So @ryoung786, would you like to fork that repo and take over maintainership?

ryoung786:
Thanks to both of you.  Somehow I totally missed the boat on flycheck.  I spent some time today switching over from flymake and it does everything I need.

I'm happy to fork and maintain flymake-cursor if that'd be helpful (even though I probably won't be using it anymore).  

Thanks @purcell and @illusori

purcell:
Thanks. I put a bunch of work into `flymake` extensions too, then encouraged lunaryorn in the early stages of Flycheck because I saw a lot of potential for it to be much better. Which it now is! It's a good time to be using Emacs.

lunaryorn:
@purcell Thanks for your support and your help. It's been a pleasure to work with you, and your support and your kind words definitely kept me going. :blush:

lunaryorn:
@illusori Thank you for this information.  

We have a [comprehensive comparison of Flymake and Flycheck](http://www.flycheck.org/en/latest/user/flycheck-versus-flymake.html) in our documentation which references your fork at a couple of places.  Now that you've stopped maintaining your fork and considering that the last commit is almost three years ago, would you mind if we removed all mentions of the fork from this document?

illusori:
@lunaryorn No objections at all. Flycheck does things the way I wanted to change Flymake to do them, if I hadn't had to maintain backwards compatibility, so as far as I'm concerned Flymake is a dead-end solution. My fork is there if anyone is stuck using it, but as you say, it's been three years, it should be enough time for people to move off it and accept the march of progress. :D

PR #4088, opened 2016-08-01
Floobits requires its own patched highlight.el

rnkn:
On Fri, 1 Jul 2016, at 03:20 AM, Geoff Greer wrote:
> Floobits bundles its own patched highlight.el. It should work fine as long as MELPA hasn’t stripped that file from the Floobits package.

### DIFF
--- a/recipes/floobits
+++ b/recipes/floobits
@@ -1,4 +1,4 @@
 (floobits
  :fetcher github
  :repo "Floobits/floobits-emacs"
- :files ("floo" "floobits.el" "floobits.py"))
+ :files ("floo" "floobits.el" "highlight.el" "floobits.py"))


milkypostman:
the problem is that floobits depends on *highlight* package. so it seems weird that it would depend on it and the then include its own.

also it seems primed to cause conflict and mismatch problems.

can we figure out a way to get the main *highlight* package updated? or maybe a way to patch on top of the package instead of a same-name fork?

rnkn:
Come to think of it, if Floobits require a version of highlight specifically patched for Floobits, it should use a different namespace prefix.

Given that I'm a fairly disinterested party I'm closing this and letting Floobits sort it out with Drew Adams.

PR #4089, opened 2016-08-01
add inferior-spim-mode

hiddenlotus:
https://github.com/hiddenlotus/inferior-spim

Hi, it's an inferior mode for spim.

### DIFF
--- a/recipes/inferior-spim
+++ b/recipes/inferior-spim
@@ -1,0 +1,1 @@
+(inferior-spim :repo "hiddenlotus/inferior-spim" :fetcher github)


purcell:
Thanks. Quick feedback:
```
(defgroup inferior-spim nil
  "Run a spim process in a buffer."
:group 'inferior-spim)
```
has an invalid `:group`, which denotes the _parent_ custom group.

```
(use-local-map inferior-spim-mode-map)
```
is redundant -- `define-derived-mode` generates this form for you.

Apart from that, the symbol prefixes are a bit inconsistent, between `run-spim`, `inferior-spim-*` and `spim-*`. As per the [elisp conventions](https://www.gnu.org/s/emacs/manual/html_node/elisp/Coding-Conventions.html), you should use a single prefix throughout, probably `inferior-spim-`.

hiddenlotus:
Hi, I changed the group from `inforier-spim` to `asm`, and all symbol with a prefix `inferior-spim`.
Thanks for your feedback.

purcell:
Thanks. I'll go ahead and merge this, but you should almost certainly also fix the code which looks like `(format "load \"%s\"\n" filename)`, because certain filenames will break it (e.g. if they contain `"`!). Safer to do something like this: `(format "load %s\n" (shell-quote-argument filename))`

PR #4090, opened 2016-08-01
add www-synonyms receipe

spebern:


purcell:
Thanks. Quick feedback:
- "a simple package" is not an adequate package description
- There are various misspellings, e.g. "spansih", "russioan"
- The prefix `www-synonyms/` should be `www-synonyms--`, as per the [elisp conventions](https://www.gnu.org/s/emacs/manual/html_node/elisp/Coding-Conventions.html)
- `helm` is a large and opinionated dependency to require to allow selection of a string. Just use `completing-read`, and packages like `helm` and `ido-ubiquitous` will override its behaviour when the user installs and enables them.

Hope that helps!

spebern:
Thanks for your suggestions and sorry for the inconvenience of those typos. I should have run a spellchecker earlier.

purcell:
Super, thanks!

PR #4091, opened 2016-08-02
point flymake-cursor repo to my github fork

ryoung786:
As discussed in https://github.com/melpa/melpa/pull/4086

If this is not necessary (as I've only added a warning to the repo readme to advise switching to flycheck), feel free to reject the pull request.


illusori:
FWIW, I've spun off my version into an organisation and given Ryan ownership, should prevent the need to keep following people's personal repos around in the future.

purcell:
Cool, thanks - Ryan, please ping me when you've updated this PR to point at the new organisation's version, since that is presumably the plan now.

ryoung786:
@purcell all set

purcell:
Great, thanks! Sorry for the slow turnaround.

PR #4092, opened 2016-08-02
add pollen-mode company-pollen recipe

lijunsong:


purcell:
Thanks. Quick feedback:
- Please add a dependency on `(cl-lib "0.5")`
- `defvar-local` was introduced in Emacs 24.3, so your dependency on Emacs 24.1 is insufficient (`read-only-mode` is also relatively new, and would not work in 24.1)
- `C-c '` is a binding [reserved for end users](https://www.gnu.org/software/emacs/manual/html_node/elisp/Key-Binding-Conventions.html), so you must not use it
- The `(put 'pollen-minor-mode-on 'permanent-local-hook t)` line looks unusual to me: should you be defining a global minor mode instead?

Hope that helps.

lijunsong:
Thank you!

I totally missed the first two you've mentioned. I'll fix the code.

For `C-c '` keybindings, I expect that some (or most) of the pollen users who also use emacs are familiar with org-mode, so `C-c '` can definitely minimize their overhead to remember keystrokes, and improve user experience. This is the first time I try to do packaging, please let me know if following convention is a must here (so org-mode is an exception of this convention?). I can definitely make it `C-c <`.

Using global minor mode is what I try to avoid. For the symbol property, it helps `pollen-minor-mode-on` function  run **locally** only in buffers of pollen files, or buffers where users have turned on this minor mode. I think this is better than global minor mode, which runs every time any major mode is changed, even in unrelated buffers.

Thanks again for review!

lijunsong:
Update pushed.

- Dependencies updated
- The key-binding has switched to `C-c C-c`.

purcell:
Great - thanks!

PR #4095, opened 2016-08-03
add org-arc-todo

yitang:
add org-arc-todo package. https://github.com/yitang/org-arc-todo

### DIFF
--- a/recipes/org-arc-todo
+++ b/recipes/org-arc-todo
@@ -1,0 +1,1 @@
+(org-arc-todo :fetcher github :repo "yitang/org-arc-todo")


purcell:
Thanks. The build failed because the recipe file name does not match its contents. Can you fix it please?

Mostly looks good otherwise, but `(shell-command (concat "arc todo " content))` makes me nervous, because `content` could contain anything! Consider using `shell-quote-argument` for that. :-)

tarsius:
@yitang Do you still want to get this into Melpa and plan to address the mentioned issues? Or should we close this?

tarsius:
I am closing this now. @yitang this can still be added to Melpa, but you need to make the requested change and generally reply to your requests.

PR #4096, opened 2016-08-03
add dired-explorer

jidaikobo-shibata:
Hi,

Please add my elisp to melpa.
dired-explorer is minor-mode that provides Explorer like select file at dired.

thanks.

### DIFF
--- a/recipes/dired-explorer
+++ b/recipes/dired-explorer
@@ -1,1 +1,1 @@
-(dired-explorer :repo "jidaikobo-shibata/dired-explorer" :fetcher github :files ("*.el"))
+(dired-explorer :repo "jidaikobo-shibata/dired-explorer" :fetcher github)
--- a/recipes/dired-explorer
+++ b/recipes/dired-explorer
@@ -1,0 +1,1 @@
+(dired-explorer :repo "jidaikobo-shibata/dired-explorer" :fetcher github :files ("*.el"))


purcell:
Thanks. Quick feedback:
- The `*.el` :files spec is redundant because the defaults would have the same effect, so please remove it
- You need to declare a package dependency on `(cl-lib "0.5")`
- `" Expr"` is a confusing mode lighter. How about `" Expl"` instead?
- `(put 'dired-find-alternate-file 'disabled nil)` seems out of place
- The following seems inappropriate - just tell the user to add `dired-explorer-mode` to their `dired-mode-hook`
```
(add-hook 'dired-mode-hook
          (lambda ()
            (define-key dired-mode-map ":" (lambda () (interactive) (dired-explorer-mode t)))
(dired-explorer-mode t)))
```
- You have `(declare-function string-trim "string-trim" (x))`, but there is no `"string-trim"` library. In my recent Emacs version, `string-trim` is in the new library `subr-x`.

Hope that helps!

jidaikobo-shibata:
Hi purcell,

I am surprised for your obliging advice.

> - The `*.el` :files spec is redundant because the defaults would have the same effect, so please remove it
I removed it.

> - You need to declare a package dependency on `(cl-lib "0.5")`
I added.

> - `" Expr"` is a confusing mode lighter. How about `" Expl"` instead?
I adopted your suggestion.

> - `(put 'dired-find-alternate-file 'disabled nil)` seems out of place
I removed them.

> - The following seems inappropriate - just tell the user to add `dired-explorer-mode` to their `dired-mode-hook`
I removed and add a instruction.

> - You have `(declare-function string-trim "string-trim" (x))`, but there is no `"string-trim"` library. In my recent Emacs version, `string-trim` is in the new library `subr-x`.
Oh, I didn't know that. I replaced function.

Thank you very much again!
I love Emacs and Emacs surrounding persons :-)

purcell:
Great - thanks!

jidaikobo-shibata:
Thank you for merge!

PR #4097, opened 2016-08-03
Add mysql-to-org-mode recipe

mallt:
- Minor mode to output the results of mysql queries to org tables
- https://github.com/mallt/mysql-to-org-mode

### DIFF
--- a/recipes/mysql-to-org
+++ b/recipes/mysql-to-org
@@ -1,0 +1,2 @@
+(mysql-to-org :fetcher github
+              :repo "mallt/mysql-to-org-mode")


purcell:
Thanks, this looks useful. Quick feedback:
- It's a minor mode, so drop the `-mode` suffix from the library/package name please. Just call it `mysql-to-org`
- `C-c m` is [reserved for end users](https://www.gnu.org/software/emacs/manual/html_node/elisp/Key-Binding-Conventions.html), so you should choose a different prefix
- In `mysql-to-org-scratch`, maybe set the `sql-product` var while calling `sql-mode`?
- You're including `expand-region` just for one use of `(er/mark-inside-quotes)` -- there's probably an easier way to do that, e.g. using `syntax-ppss` to check whether you're in a string, and then jumping up to the enclosing sexp. (See [this code](https://github.com/purcell/emacs.d/blob/master/lisp/init-editing-utils.el#L204-L211) for example.)

mallt:
Thanks a lot for the feedback! I've update the repo to include the suggestions.

purcell:
Thanks. I'll go ahead and merge this, but I also notice you've added a company dependency, which is quite opinionated and not going to suit many users. A more general approach is to provide a `completion-at-point` function: company will then use it to gather candidates, thus making everyone happy! :-)

mallt:
Thanks a lot @purcell! I will change the company dependency to completion-at-point.

PR #4099, opened 2016-08-04
Update ledger-mode recipe to reflect its new home.

jaccarmac:
See ledger/ledger@15d18d664f0e9c5e454bf4927f7d0e0bca02b0c2 and the
recent mailing list discussion on ledger-cli.

Since old elisp files were removed from the next branch of
ledger/ledger, the old recipe does not work at all any more.


purcell:
Awesome, thanks!

PR #4101, opened 2016-08-05
Ls

troyp:
* __package summary:__ list processing utilities focusing on concision and readabllity.
* __URL:__ https://github.com/troyp/ls.el
* __notes:__ This and __fn.el__ were going to be one package but I decided it made more sense to split them up. There's still a lot I want to add to these, but I want to put them on melpa now, since I'm working on another package and I want to be able require these.

### DIFF
--- a/recipes/fn
+++ b/recipes/fn
@@ -1,0 +1,2 @@
+(fn :fetcher github :repo "troyp/fn")
+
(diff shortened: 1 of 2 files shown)


troyp:
I don't know what's going on with the Travis checks, but I've somehow put both recipes in this pull request, when I only wanted one, so I'll redo it.

PR #4104, opened 2016-08-05
add fn.el recipe

troyp:
- **package summary:** functional programming utilities focusing on concision and readabllity.
- **URL:** https://github.com/troyp/fn.el
- **submitter relationship to package:** author
- **notes:** This and **ls.el** were going to be one package but I decided it made more sense to split them up. There's still a lot I want to add to these, but I want to put them on melpa now, since I'm working on another package and I want to be able require these.

### DIFF
--- a/recipes/fn
+++ b/recipes/fn
@@ -1,0 +1,1 @@
+(fn :fetcher github :repo "troyp/fn.el" :files ("fn.el"))
--- a/recipes/fn
+++ b/recipes/fn
@@ -1,1 +1,1 @@
-(fn :fetcher github :repo "troyp/fn.el" :files ("fn.el"))
+(fn :fetcher github :repo "troyp/fn.el")


purcell:
Thanks. Quick feedback:
- You shouldn't provide `:files` here because the default will do the right thing.
- Don't require `cl-macs`, which might not be available in all the Emacs versions you're targeting. Instead, you should be requiring `cl-lib`, and declaring a package dependency on `(cl-lib "0.5")`
- The `fn` macro will break in certain edge cases: it will shadow the name `args` in the body, so if a user relied on lexically binding to a local with that name, the results would be unexpected. Not sure you can easily fix this, but you could at least document it.
- You don't need to include `dash` or `dash-functional` here at all, afaict: you're only using `-map` and `-lambda`. If that's the case, you could lighten the dependencies.

troyp:
Thanks for the feedback.

It seems I do need `:files`. When I tried without it, the Travis build was failing, and it turns out it can't handle repo with the ".el" in the name. When I looked at similarly named libraries, this was how they solved it.

Re: `cl-lib`: Thanks, I've done that.

Re: `fn`. Yes, you're right. The initial implementation was really naive (it was only temporary). It also caused compiler warnings since it bound unused variables. I've rewritten it to remove both of those issues (and added a test for variable capture, although the test file's not actually part of the package).

Re: dash dependencies: Actually, it currently uses `-map`, `-lambda`, `-flatten`, `-intersection` and `-elem-index`. There are 9 uses in total (not counting the test file, examples, etc). A lot of those are in the serious implementation of `fn`, which you hadn't seen since I hadn't pushed it yet. Two of the uses are `-map` which could replaced with the slightly uglier `mapcar`, so 7 really. So far. I could remove them if it was really important, but they do actually help keep the code a lot more concise. (The new `fn` would be quite an effort to wade through without them.) I guess I could remove the 'dash-functional dependency pretty easily by just removing the `-lambda`. But the other reason I haven't tried to avoid them is because I consider them to be absolute minimal requirements for the kind of code `fn.el` is meant to facilitate. `fn.el` is intended to build on top of them (which I mention in the README). Anyway, I could remove them if it's necessary.

purcell:
Great, thanks!

Sorry - I must have missed those further `dash` function references in review, and of course you're quite right to keep the dependency in that case!

Re. the `:files` spec, whatever the Travis build said, it's correct to omit `:files` here. If there's an error in the Travis build as a result - which I've never seen before due to file naming - then I'll fix that separately. You should find that the recipe without `:files` builds and installs correctly with the test method noted in the README.

troyp:
Okay, I've removed `:files`.

You're right, it looks like it's working. I have no idea what happened before.

troyp:
fn's API has been bugging me, so I've decided to streamline it while there's still time (since it will be difficult to remove things if it's on melpa). I've removed a couple of functions and shifted all the arithmetic factories (which are a lot more specialized and cluttered up the API) to a file in a subdirectory (where they should be out of the package).

I realise this leaves fn.el a rather minimal library with only a handful of core macros, but I think it's more coherent this way and the conservative choice is to leave out anything I'm unsure of. I have a few more functions I'll probably add soon, but I'm still considering.

Anyway, tell me if this is a problem.

tarsius:
I am very interested in something like `fn` and I would actually like it to be named `fn` as you intend to do. I think it is not only okay but favorable if the package providing that macro did not include anything else. These other libraries you are mentioning should be moved to a completely new repository.

I also think that `fn` should be rock solid before being added to Melpa.

I was hopping that [`smart-lambda`](https://github.com/abo-abo/short-lambda) would become a reality, but that relies on a patch to the C part of Emacs and that was rejected. But maybe the lisp parts can serve as inspiration.

> Not sure you can easily fix this, but you could at least document it.

I would go further; you should try to address all issues, not document limitations. There should also be extensive tests and the result should be reviewed by some knowledgeable people before the first release.

I'm not trying to discourage you. On the contrary, I would love `fn` and think others would to. But for it to be a success it has to be very good and that will take a lot of work. But if you do put in that work, then you would also get the deserved exposure - e.g. I would use it in Magit and that would result in many downloads.

tarsius:
Just found this again: http://oremacs.com/2015/01/22/clojure-sugar-for-elisp/

troyp:
@tarsius: You have a point. I did actually consider that. The `fn` macro is likely to be used more than the rest put together. I'm open to moving everything else into another repo and keeping `fn` on its own. Having a macro-only library (whether or not it's just `fn`) would also allow it be used with `eval-when-compile` without affecting the namespace at runtime. (Well, I guess you could do that anyway, as long you only used the macros.)

I hadn't seen that oremacs post. Fuco mentioned in the comments that the `%n` syntax was going to be implemented in dash, but that doesn't seem to have happened. It would be nice if it did.

Thanks for pointing out `short-lambda`. Even though the `#()` patch was rejected, the `short-lambda` macro is still useful if given a short alias. It's actually equivalent to `fn` but with a different syntax for the anaphoric parameters (or will be once I get around to adding a "rest" arg to `fn`).

The issue purcell mentioned (and another I was aware of) has been fixed. There are no problems with this implementation that I know of, but let me know if you find any. One change I do intend to make is to add an anaphoric "rest" parameter (probably `<...>`).

I'll add more extensive unit tests when I get a chance. At the moment the unit tests are basically just the examples from the documentation. Let me know if you have any suggestions of edge cases or patterns of usage that should be tested.

What do you think of the interface? Do you think `<n>` are suitable? This clojure `%n` syntax is an alternative, although it feels a bit line-noisy to me. I find `<n>` highly readable/scannable, personally.

tarsius:
@troyp I'll try to get back to this soonish.

tarsius:
> I'll try to get back to this soonish.

Ah well. I have to admit I have lost interest a bit, I really want to use `#(foo ...)`  - `(fn (foo ...))` doesn't quite cut it for me. So I don't really care about the details much anymore. Since the issues raised by @purcell have been fixed, he should probably merge this.

> What do you think of the interface? Do you think `<n>` are suitable? This clojure `%n` syntax is an alternative, although it feels a bit line-noisy to me. I find `<n>` highly readable/scannable, personally.

I like `%n` much better than `<n>`, but that's just me.

purcell:
Thanks @tarsius and @troyp!

PR #4105, opened 2016-08-05
add ls.el recipe

troyp:
* __package summary:__ list processing utilities focusing on concision and readabllity.
* __URL:__ https://github.com/troyp/ls.el
* __notes:__ This and __fn.el__ were going to be one package but I decided it made more sense to split them up. There's still a lot I want to add to these, but I want to put them on melpa now, since I'm working on another package and I want to be able require these.

### DIFF
--- a/recipes/ls
+++ b/recipes/ls
@@ -1,0 +1,1 @@
+(ls :fetcher github :repo "troyp/ls.el" :files ("ls.el"))


troyp:
Re `ls-zero-when`: This isn't all that specific, really. Use cases for this actually come up quite a lot. Maybe not in everybody's code, but in mine and in plenty of other peoples' code I've seen. The use case for this is when you have a couple of lists in correspondence (either you've produced two lists in the same process - like an unzipping a alist - or you've started with one list and generated another by processing it). You want to recombine those two lists, but first you need to filter out some entries in one. One way to handle this is to "zero out" the unwanted elements with an identity/null value (which for Lisp is usually going to be `nil`). There are other similar scenarios... basically this is for "removing" elements from a list without changing the indexing.

Well, I disagree about `cl-loop` (obviously, which is why I wanted sugar for it), but `ls-range` isn't really meant to be the main range function. I was planning to add a normal range function (possibly just a short alias or wrapper for `number-sequence`) for that purpose, but hadn't decided on the best name (`ls-seq` maybe) or interface yet.

I don't think these libraries are quirky. They're aimed at a specific style of programming, and but most of the stuff in them is mostly common in other language APIs. Even with `ls-zero-when` and related functions, IIRC, there are common idioms in R and J for doing something analogous. Then are things like predicates for testing for proper lists, which aren't that common, but just seem like really basic stuff that should be available. (As to whether they're idiomatic, well TBH my sense of typical Emacs Lisp is that it's really verbose and hard read. These libraries are definitely aimed at more concise programming.)

That said, I can understand why they may seem that way (well, `ls.el`, anyway). `ls.el` is quite "patchy". It's pretty hit-and-miss what's included and what's been left out for now. I split out the list functions from `fn.el`, but I haven't given a lot of thought yet to the list API, so the functions in it right now are just the ones I was confident about. Everything else, I put off to decide on (and implement) later. The only reason I put this here now was so I could use it in other packages, but it's probably a mistake. It's not really a coherent library yet.

Forget `ls.el` for now. I'll resubmit it at some point when it's more complete (unless I find an alternative). If you have further concerns with `fn.el` let me know on its PR.

PR #4106, opened 2016-08-06
add ciel.el recipe

cs14095:
Hi, please add ciel.

A command that is clone of "ci" in vim.
https://github.com/cs14095/ciel.el

Thank you.

### DIFF
--- a/recipes/ciel
+++ b/recipes/ciel
@@ -1,0 +1,1 @@
+(ciel :repo "cs14095/ciel.el" :fetcher github)


purcell:
Thanks. Quick feedback:
- `C-c i` and `C-c o` are [reserved for end users](https://www.gnu.org/software/emacs/manual/html_node/elisp/Key-Binding-Conventions.html), so you must not use them in your minor mode. But that's easy to solve: you don't need a global minor mode anyway - just tell users to bind those two keys globally.
- The `%name` convention is unusual, and I would advise against it
- I think `(unless (null %region)` is equivalent to `(when %region`
- `""` isn't a great docstring ;-)
- You seem to have reinvented prefix maps: why not just create individual commands for each subcommands, and bind them into a prefix map?

Hope that helps!
-

cs14095:
Thank you for response.
So I should

 - remove `(define-minor-mode ` and `(defvar ciel-mode-map `
 - add `(global-set-key "\C-ci" 'ciel-ci)`

then is this ok?

purcell:
Yes, that's enough to merge this recipe.

But I'd also suggest you investigate creating individual commands for `c i w` etc, because they might be useful individually. You can use a function or macro to eliminate duplicated code, and then put all the commands in a prefix keymap. :-)

cs14095:
Hi, I changed you pointed out.
Please look over https://github.com/cs14095/ciel.el/blob/master/ciel.el.

purcell:
Thanks!

PR #4109, opened 2016-08-07
add leanote recipe

aborn:
A minor mode writing markdown leanote
https://github.com/aborn/leanote-emacs

### DIFF
--- a/recipes/leanote
+++ b/recipes/leanote
@@ -1,0 +1,1 @@
+(leanote :fetcher github :repo "aborn/leanote-emacs")


purcell:
Thanks. Quick feedback:
- `C-c u` etc are [reserved for end user customisation](https://www.gnu.org/software/emacs/manual/html_node/elisp/Key-Binding-Conventions.html) so you must not use them in your minor mode
- `spaceline` seems like a strange UI dependency which users wouldn't expect to be needed for writing content.
- `helm` is a similarly heavyweight dependency which also seems inappropriate here. `swiper` too. Perhaps those should be soft-dependencies?
- `M-x checkdoc` will help you fix up your docstrings so they are formatted conventionally
- `subr-x` is very new, so you should probably declare a package dependency on something like `(emacs "24.4")`.

Hope that helps!
-

aborn:
Hi, @purcell 
* Changed the key binding with prefix C-x C-l 
* Using spaceline as optional, user can active it by (leanote-spaceline-status) if necessary.
* Helm and swiper are also optional, and not as heavyweight dependency.
* Docstrings are formatted conventionally.
* Last, add (emacs "24.4") dependency.  
Thanks your code review work. Best wishes!  
https://github.com/aborn/leanote-emacs/blob/master/leanote.el

purcell:
Great, thanks. `leanote-find` still uses `ivy-read`, though, which indicates the package won't really work without `ivy`, so you'd need a hard dependency for that.

Why not just make it `leanote-find` use `completion-read` by default, which `helm` will override? Then you don't need to mention `helm` or `ivy` -- in practice, you should never be trying to figure out what completion method the user prefers, and your code would be a lot simpler in this case.

aborn:
Yep, completing-read is an elegant method for note finding. I have use it and remove swiper & helm. Thanks @purcell again. 
https://github.com/aborn/leanote-emacs/blob/master/leanote.el

purcell:
Great, thanks. P.S. you can now replace `leanote--open-note` and `leanote-find` with the following:

```el
(defun leanote-find (note)
  "Find note by title with completing-read."
  (interactive
   (list (completing-read "search note by title: "
                          (leanote-get-all-notes-from-cache))))
  (unless (file-exists-p note)
    (error "No such file"))
  (find-file note))
```

aborn:
Modified!  Thanks.

PR #4110, opened 2016-08-07
Add ghub and glab recipes

tarsius:


vyp:
Just in case you forgot or something, tarsius/glab doesn't exist yet! :wink:

syohex:
`tarsius/glab` is missed. https://github.com/tarsius/glab returns 404 status code. And `ghub` can be used only Emacs 25 or higher version because it uses `pcase-lambda`. But it does not specify minimum Emacs version.

tarsius:
Now they do (`glab` had the same issue).

purcell:
Yay, thanks.

PR #4112, opened 2016-08-08
Add coq commentor package

Ailrun:
Add new package for comment supporting for [coq](https://coq.inria.fr/) proof assistant.

### DIFF
--- a/recipes/coq-commenter
+++ b/recipes/coq-commenter
@@ -1,0 +1,3 @@
+(coq-commentor
+ :repo "ailrun/coq-commenter"
+ :fetcher github)


purcell:
Thanks! The build failed because the name in the recipe is mis-spelt: `coq-commentor`. Can you fix that please?

There are various issues with the code, though:
- Unprefixed top-level names like `between-pos` are not valid. Use `coq-commenter-` for every top-level symbol, e.g. `between-pos` should be `coq-commenter-between-pos`.
- `C-;` and `C-'` are [reserved for end-user customisation](https://www.gnu.org/software/emacs/manual/html_node/elisp/Key-Binding-Conventions.html) so you must not use them in your code
- All the code you have which uses `global-set-key` and `custom-initialize-set` should not be doing that. Provide a global minor mode if you want certain keybindings available everywhere, but you probably just want a local minor mode instead: read about `define-minor-mode`.


Hope that helps!

Ailrun:
Thank you for kind feedback! I'll fix them

PR #4120, opened 2016-08-11
add reverse-im package

a13:
Adds translation mappings for input-methods other than default english one, so all keybindings are usable while non-default system layout is active.

https://github.com/a13/reverse-im.el

### DIFF
--- a/recipes/reverse-im
+++ b/recipes/reverse-im
@@ -1,0 +1,1 @@
+(reverse-im :repo "a13/reverse-im.el" :fetcher github)


purcell:
Thanks. Looks good, but all the top-level symbols in the file should have the prefix `reverse-im-`, as per the [elisp conventions](https://www.gnu.org/s/emacs/manual/html_node/elisp/Coding-Conventions.html).

a13:
Thanks! 
Fixed(I hope so) in 92c30e412496e48540e97e780c520be877e8abef

purcell:
Thanks!

Issue #3570, opened 2016-02-07
Migrating MELPA to organization and separating packages.


milkypostman:
Two things we want to do,
1. move the MELPA repo to the MELPA organization.
2. separate recipes from the `package-build` package.

What will we do with the recipes? Will they live in the root of the repo? I think this is what we need so that  we can checkout the `package-build.el` package (which includes the scripts) then the recipe repo will be checked out to `./recipes`.

The only other downside is that we need to get the Travis system setup for the recipes repo. That will need to checkout package-build I guess.

For the first step I think I will just move the repo over.

milkypostman:
1edd038

milkypostman:
welcome to the new org, we will have to rename these repos at some point I think, but I don't have time to do the split right now, it'll take a bit more time.

purcell:
I don't think there's any need to split things up further, or move things around inside the repo. Let's just extract `package-build.el` to the root of a separate repo using Jonas' script, and switch `recipes/package-build` to point at that instead. We'd then need to ensure that the build scripts install the latest version of `package-build` when starting up.

purcell:
There'll be some sort of update necessary for travis too, right?

purcell:
Actually, Travis seems to still be working.

tarsius:
> Let's just extract package-build.el to the root of a separate repo using Jonas' script, and switch recipes/package-build to point at that instead.

Is there a reason why you cannot just do so now? ;-)

> We'd then need to ensure that the build scripts install the latest version of package-build when starting up.

I guess something could go wrong here, but if it did then that would be noticed very quickly and would probably be easy to fix, right?

This should be fairly easy to do for you wouldn't it? I could give it a try to, but that would take longer and come with a greater risk of getting it wrong, since I am not familiar with the build scripts.

tarsius:
We've done that.

Issue #4113, opened 2016-08-08
Use a new Emacs on the build server


purcell:
See #4087. Probably need to use a new Ubuntu snapshot, or maybe an Emacs-specific PPA.

tarsius:
What version do you plan to use?

I am asking because I started cleaning up `package-build.el` and would like to use the latest (well upcoming) release, 25.1. I don't think it's necessary for `package-build` to support older releases as I would assume that everyone who is interested in something like that would also use the latest Emacs release (or the development version).

vyp:
> I don't think it's necessary for package-build to support older releases as I would assume that everyone who is interested in something like that would also use the latest Emacs release (or the development version).

By older are you saying 24.5 (current stable release according to https://www.gnu.org/software/emacs/)? Because quelpa uses package-build and non-developers (users) use it.

purcell:
At this point I'd probably plan to support maybe 24.4 and up.

milkypostman:
i think we should probably to update to 16.04? instead of upgrading in
place I will probably bring up a new instance then switch over when I get a
chance?

purcell:
Yep, sounds good.

lunaryorn:
@tarsius I should probably weight in on behalf of Cask to say that we use `package-build` in Cask, and are definitely not ready to drop support for Emacs 24.  Should you bump package-build.el to Emacs 25 only we'd have to fork.

I should also like to note out that I had hoped to see research about the actual users of a library before making assumption about what Emacs versions it's being used with.  I came across this issue accidentally; had I not seen it we'd likely have run into troubles with Cask following your cleaning up of package-build.

tarsius:
Okay everyone, I'll stick to backward compatible cleanup for now ;-)

lunaryorn:
@tarsius Thanks :blush:

purcell:
Now that we have a separate repo, we can probably put in place some regression tests and a travis build across supported Emacs versions.

tarsius:
@milkypostman Could you please update to 25.3 now? Don't know about the server but at least on Travis Melpa still uses 24.5.

Maybe Melpa could just use the binaries created by @npostavs. We use that successfully [for Magit](https://github.com/magit/magit/blob/master/.travis.yml) and I don't thing he has any plans to stop providing these binaries.

npostavs:
[...]

Issue #3818, opened 2016-04-27
d3js visualization


Lompik:
Hi

In melpa.js I saw

    // TODO D3 visualisation for deps

I've actually made one https://github.com/Lompik/emacs-melpa-network based on the json served on melpa. An automatic build is doable with nodejs but I don't have this yet. 

Basically, I'm looking to see if that TODO is still pending/current and if I help complete it in some way.

Regards.

purcell:
That's very neat! I hadn't really thought of trying to display all the packages on a single chart -- what I had in mind was to be able to show a tree of all directly and indirectly dependent packages for any specific package, so on the [page for `s`](http://melpa.org/#/s), for example, I'd love to see a graph showing not just the "Needed by" list of packages, but also the packages which depend on those.

Lompik:
You mean like :
https://fab974.org/blog/melpa-d3js/spackagedeps.svg

the `s` package is probably one of the most filled in term of dependency with 230 packages.

purcell:
Haven't forgotten about this - just very busy, and finding it hard to tackle bigger things. One of the things I'm wondering is how to integrate the d3 diagrams with the rest of the mithril.js code.

Lompik:
I'm not an expert but it seems we can can call methods on DOM element via the config attribute .. and have the d3 in the mehods. https://lhorie.github.io/mithril/integration.html and https://lhorie.github.io/mithril/mithril.html#the-config-attribute.

AFAIU you want to have the d3 code inside https://github.com/melpa/melpa/blob/master/html/js/melpa.js#L108 and process the json data from there, on the client side. I was more thinking about generating the 3000+ svgs in the `Makefile`. 
I guess the question I have for you is: how much interactiveness are you planning for the graphs ? 

There are plenty of discrepencies in json data (wrong depencies names, version, etc) that comes from the providers. Maybe tring to fix could be a first step.

purcell:
> AFAIU you want to have the d3 code inside https://github.com/melpa/melpa/blob/master/html/js/melpa.js#L108 and process the json data from there, on the client side. I was more thinking about generating the 3000+ svgs in the Makefile.

Yeah, I guess so, because the json data is already available in the client, and we don't pre-process much data on the server. Issue #3728 relates to switching to generating most of the website content on the server, but that's non-trivial esssentially due to cache invalidation. :-)

> I guess the question I have for you is: how much interactiveness are you planning for the graphs ?

Probably just links through to the corresponding MELPA package pages.

[...]

PR #4121, opened 2016-08-12
Move package-build.el to a seperate repository and merge it as a subtree

tarsius:
In #3569 and #3570 we talked about moving `package-build.el` to a separate repository, without going into the details of how it can then be made available to the other build tools again.

I think it is best if we simply add the `package-build` repository as a subtree of the `melpa` repository. That way we don't have to worry about potential issues when fetching the latest version and we also explicitly track which revision was used at any given time.

I have created the filtered history as described at https://github.com/melpa/melpa/pull/3569#issuecomment-181034310 and pushed the result to https://github.com/tarsius/package-build.

If you decide to use this, then you should fork my `package-build` repository to `melpa/package-build`. Once you have done that I will delete my repository, making the latter the "root of the network".

Everyone who works on `package-build.el` and Melpa should then clone the `package-build` repository and then add that clone as a remote inside the Melpa repository:

```sh
cd /path/to/melpa/..
git clone git@github.com:melpa/package-build.git
cd melpa
git remote add package-build ../package-build
git fetch package-build
```

After making changes in the `package-build` repository, they can be imported into the Melpa repository using `make pull-package-build`.


purcell:
Done. Thanks so much for putting this PR together. Let's see what breaks!

purcell:
I guess we'll leave issues for package-build here for now, and only gradually move over to maintaining package-build separately at its new home.

purcell:
Build's failing with an apparent load-path issue and I'm not sure why...

tarsius:
I'll look into it.

tarsius:
I forgot to adjust the path in the `package-build` recipe. Fixed in #4123. But that doesn't seem related.

The error must be in https://github.com/melpa/melpa/blob/8511dbe322a17e8fdf2bb433def776756ce0e9ed/run-travis-ci.sh#L13. But your abb42a067451fbf0720ce2dd571b2c787c30b528 should address that. I haven't gotten any further than that yet.

tarsius:
Uh, wait... that fix is broken... having high hopes for https://travis-ci.org/melpa/melpa/builds/151783986.

tarsius:
Okay, now on to the bigger issue: you should not have squashed. `git subtree` needs the merge commits that it creates.

I have rerolled the changes since you squashed my commits. I have cherry-picked the various recipe commits and since history had to be rewritten anyway I have integrated the various path adjustment fixup commits into "Move package-build.el to a separate repository".

The result is in `tarsius/master`. I think you should hard reset `master` to `tarsius/master` as soon as possible and then force push. If any contributors are having issue due to the reset, then I can help with that. But I don't think we really have a choice here.

tarsius:
Pending https://travis-ci.org/tarsius/melpa/builds/151791773

purcell:
> I think you should hard reset master to tarsius/master as soon as possible and then force push.

Done. That makes sense. Thanks for figuring this out. The squash was the default when merging via Github, since we typically squash recipe PRs... sigh.

tarsius:
You are welcome!

purcell:
If it's okay with @milkypostman I'd quite like to add you as a MELPA committer, not that I'd expect you to be actively working on this project, but because I have a lot of faith in your judgement and attention to detail.

tarsius:
I would very much appreciate that, thanks.

milkypostman:
Sounds fine with me.

milkypostman:
is there any issue with just separating these packages eventually? I feel like we could just check the recipes out into their own directory and just have the scripts update the root and the packages?

both submodules and subtrees seem to be annoying, 

```
git clone package-build
cd package-build
git clone recipes
```

i'm fine with this for now too because I do appreciate your work on this, I just assumed the scripts would go with package build.

i'm also OK if we have a `toolchain` repo that only contains everything besides the recipes.

so 

```
git clone toolchain
cd toolchain
git clone package-build
git clone recipes
```

thoughts? also I don't really care about bloated history that much but filtering is fine too.

milkypostman:
oh, and we could just rename *this* repo to `recipes` and then move all the recipes to the root of the repo. but just ideas.

purcell:
Cool. I've added @tarsius to the MELPA org, and as a collaborator on the 2 repos. Yay!

Regarding what should live where, I'd ultimately like to get `package-build.el` out of the MELPA repo entirely, such that the MELPA repo would bootstrap itself at build time by installing the MELPA package of `package-build`, or directly grabbing it from git. We'll need to move recipe format docs to the package-build repo ultimately. Recipes themselves could live in a separate repo. And some of `package-build.el` is specific to MELPA, e.g. the default directory paths, but also the `shields.io` badge image stuff, so we should move that to a separate `melpa.el` or similar in this repo IMO. I'm not in a super hurry to do any of that.

tarsius:
> Cool. I've added @tarsius to the MELPA org, and as a collaborator on the 2 repos. Yay!

Weeeeeee!

> not that I'd expect you to be actively working on this project

I've started labeling issues, because I think it helps a lot to see what areas need work and how much. (Then, depending on your mood, you can concentrate on the areas that need the most work *or* one where you actually have a change to "get done" in a usable time frame, for example.)

* I would like to add a `build` label distinct from the existing `website`. The former would be used for things like #4113 and the latter for #3818.

* I have alread added a `version` label for all the discussions about version schemes. This subject keeps coming up, usually without proper review of past discussions. Labeling might help here. And I *still* intend to work on that eventually myself. Not having to hunt down the relevant discussions yet once more when I have another look, might help motivate me to do so.

tarsius:
@purcell, @milkypostman I have made my Melpa membership public. You should probably do the same, which you can do at https://github.com/orgs/melpa/people.

PR #4122, opened 2016-08-12
add del-line recipe

zhuanqizhirou:
del-line: delete line/word directly , without kill ring.


purcell:
Thanks, but that's not a well-formed package:

- None of the headers are formatted correctly for a package, and there's no `Commentary` section
- The `provide` form is missing
- The `pw-` prefix for all top-level symbols should be `del-line-`
- You shouldn't be binding keys on the user's behalf using `global-set-key`

tarsius:
@zhuanqizhirou Do you still want to get this into Melpa and plan to address the mentioned issues? Or should we close this?

PR #4126, opened 2016-08-12
Add recipe for `gforth.el` from smtlaissezfaire

mrhmouse:


purcell:
Thanks. Please ignore that the commit history in this PR has ended up messed up -- it's the result of some fixes on our end.

In any case, I've taken a look. The `gforth.el` file itself isn't a valid package yet, for various reasons, and would therefore require some fixing up:
- The `provide`-d feature name doesn't match the filename
- The symbol prefix (`forth-`) should also match the feature name, ideally
- The code for shimming `when`, `unless`, `font-lock-warning-face` etc is inappropriate these days

In general I think the file could do with a clean-up and modernisation, because there are further points such as

- It's better to use `define-derived-mode` instead of hand-rolling all the major-mode machinery the old-fashioned way
- Using `;` for full-line comments is incorrect
- Helpers like `forth-filter` can be replaced with `cl-filter` these days
- `;; workaround for bug in `reindent-then-newline-and-indent'` is likely obsolete

Not sure how you want to proceed - maybe file an upstream PR?

mrhmouse:
Thanks for taking a look! I'm not sure about an upstream PR, because the actual upstream is the GForth distribution. I may hack away at `gforth.el` for a bit, though, and open another request here when I've got something workable.

PR #4127, opened 2016-08-12
Add ditz-mode recipe

tarsius:


purcell:
Thanks. I submitted an upstream fix too - https://bitbucket.org/zondo/ditz-mode/pull-requests/1/minor-fixes-for-melpa-packaging/diff

PR #4129, opened 2016-08-12
Add a recipe for read-aloud

gromnitsky:


purcell:
Thanks, added. You might want to add "say", which is the standard tts utility on OS X:

```
say [-v voice] [-r rate] [-o outfile [audio format options] | -n name:port | -a device] [-f file | string ...]
```

PR #4130, opened 2016-08-12
Add distel-completion recipe

sebastiw:
**Summary:** Erlang/Distel completion backend for both auto-complete-mode and company-mode.
**Url:** https://github.com/sebastiw/distel-completion
**Submitter relationship to package:** author


purcell:
Sorry, these should be 2 or 3 separate packages, with distinct package dependencies on `company` and `auto-complete`, otherwise the user must have both completion packages installed. The common core could be called `distel-completion`.

PR #4132, opened 2016-08-12
Add evil-text-object-python recipe

wbolster:
this pr adds a recipe for `evil-text-object-python`, available from:

https://github.com/wbolster/evil-text-object-python

...of which i'm the author.

local `package-install` works fine.


purcell:
Thanks, added.

PR #4117, opened 2016-08-10
rename recipe emacs-eclim to eclim for use-package

stardiviner:
https://github.com/senny/emacs-eclim/issues/277

PR #4133, opened 2016-08-13
Add other eclim recipes

stardiviner:


purcell:
Pending senny/emacs-eclim#281. This replaces #4117.

PR #4134, opened 2016-08-13
new repo and update branch

hexmode:


purcell:
Re. `holiday-pascha-etc`:
- `provide` form is missing
- `emacs` in the package description is redundant, and should be removed
- Needs a brief `;;; Commentary:` section to describe what the package is and how to use it

The other change looks good. I dislike it when developers hide all their ongoing work in a non-default branch... it rather misses the point of github, which is full of developers rather than end-users.

hexmode:
I'm not sure what you mean by hiding on-going work.  I had the idea for this a year ago, but only now sat down to do it.  All this is just from this mornings work. (But you're probably talking about something I'm completely missing...)

I'll work on the issues you pointed out tomorrow.

purcell:
I'm referring to the `next` branch of `terranpro/magit-gerrit`, not to anything you've done. :-)

tarsius:
`magit-gerrit` has a `next` branch because `magit` did the same for a while. But since `magit`'s `next` branch was merged more than a year ago, it's about time the same happens for `magit-gerrit` now.

/cc @terranpro

terranpro:
D'oh - been out of the loop for awhile, and things change so fast!
Thanks for the ping, will merge up and bid farewell to the `next` branch.

hexmode:
could you re-check now?

purcell:
Looks good, thanks.

PR #4135, opened 2016-08-13
First check-in of new salesforce-utils recipe

grimnebulin:
Hi--

This is a very short package of Salesforce utility functions, all related to converting a 15-character Salesforce ID into an 18-character ID-with-checksum.

The source repository is here:

https://github.com/grimnebulin/emacs-salesforce

I am the creator/maintainer.

### DIFF
--- a/recipes/salesforce-utils
+++ b/recipes/salesforce-utils
@@ -1,0 +1,1 @@
+(salesforce-utils :fetcher github :repo "grimnebulin/emacs-salesforce")


syohex:
- There is no `provide`(ex `(provide 'salesforce-utils)` ) so users cannot load it by `(require 'salesforce-utils)`.
- It uses cl-lib functions but it does not load cl-lib and does not have a package-dependency header. (Load cl-lib.el instead of cl.el and remove `eval-when-compile` because it uses not only macros but also functions)
- Load `thingatpt` package for using `word-at-point` which is not autoloaded function

grimnebulin:
OK, I've updated my source repo per your comments.  Please have another look.

purcell:
Thanks! Pending https://github.com/grimnebulin/emacs-salesforce/pull/1.

purcell:
Great, thanks.

PR #4136, opened 2016-08-13
Added haxe-imports recipe

accidentalrebel:
Code for dealing with Haxe imports

This package is based on java-imports by Mathew Lee Hinman found here http://www.github.com/dakrone/emacs-java-imports.

This package adds the import-haxe-class function, which will prompt for the class at point, then the package, then add the required import statement for the class and cache a "class-name -> package" relationship for any future importing of the class.

URL: http://www.github.com/dakrone/emacs-haxe-imports


accidentalrebel:
Pull request approved. Thanks!

PR #4137, opened 2016-08-14
Eclim recipes

stardiviner:
- ac-eclim
- company-eclim


purcell:
Hi - these recipes aren't valid, sorry. The names in the recipes don't match the recipe file names. Just leave things for now, and when https://github.com/emacs-eclim/emacs-eclim/pull/1 is merged, I'll fix up all the recipes as necessary.

PR #4139, opened 2016-08-14
Add jpop recipe

tarsius:


purcell:
I think this recipe should bundle the python script...?

PR #4146, opened 2016-08-15
Add recipe for auto-complete-distel and company-distel

sebastiw:
**Summary:** Erlang/Distel completion backend for auto-complete-mode and company-mode.
**Url:** https://github.com/sebastiw/distel-completion
**Submitter relationship to package:** author

### DIFF
--- a/recipes/auto-complete-distel
+++ b/recipes/auto-complete-distel
@@ -1,4 +1,4 @@
 (auto-complete-distel
  :fetcher github
  :repo   "sebastiw/distel-completion"
- :files ("auto-complete-distel.el" "distel-completion-lib.el"))
+ :files ("auto-complete-distel.el"))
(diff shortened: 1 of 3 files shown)


purcell:
Thanks, but this way we end up with the same file in both packages: the `distel-completion` part should be a package which both the others depend on. (As a bonus, it could also provide a completion-at-point function which would be usable even without `auto-complete` or `company`.) Note that there are still a number unprefixed symbols in `distel-completion-lib` (e.g. `erl-get-function`, `try-erl-desc-cache`) which should all be renamed to start `distel-completion-`.

sebastiw:
Hi I prefixed all variables and functions and added the library to a separate package.

purcell:
Awesome, thanks!

PR #4147, opened 2016-08-15
Recipe ob spice

stardiviner:


purcell:
Thanks. `ob-spice` looks good. Misc feedback re. `spice-mode`:
- Don't use `(defvar ... nil)` followed immediately by `(setq ...)` - just use `defvar`, or `defconst`. The current code is very unidiomatic.
- The `(autoload 'spice-mode "spice-mode" "Spice  Editing Mode" t)` form has no effect or benefit, since it will only be evaluated when the file has already been loaded. Replace that line with `;;;###autoload`
- Your `spice-mode` function definition is extremely old-fashioned, and contains a lot of unnecessary boilerplate. Replace it with `define-derived-mode`, and then you can delete all [these lines](https://github.com/stardiviner/spice-mode/blob/c808eb84dacb25e157ecbec646803061e665a4c0/spice-mode.el#L78-L83) and [this line](https://github.com/stardiviner/spice-mode/blob/c808eb84dacb25e157ecbec646803061e665a4c0/spice-mode.el#L92)
- Many of the functions are called `fume-*` -- these should be renamed to `spice-*` or `spice-mode-*`

Hope that helps!

stardiviner:
@purcell Thanks a lot, I refactored the code. Check it out.

purcell:
Yep, great. Just pending https://github.com/stardiviner/spice-mode/pull/1 now.

stardiviner:
@purcell get merged.

tarsius:
Thanks!

PR #4148, opened 2016-08-15
Add kys

hoverwinter:
Hello, I'm the maintainer of kys. KYS manage emacs sessions with tags and store them on disk, you can resume the work environment at any time, for detail information please go to [kys](https://github.com/hoverwinter/kys).

### DIFF
--- a/recipes/kys
+++ b/recipes/kys
@@ -1,0 +1,1 @@
+(kys :fetcher github :repo "hoverwinter/kys")


Silex:
Hi! Not related to your PR but concerning your package, I was looking for something like this since a long time. Can you elaborate a little bit the difference with `desktop-save-mode` and maybe others ?

Full list of similar packages at https://www.emacswiki.org/emacs/SessionManagement

purcell:
Yep, looks cool, but I also wonder if you could simply be using `desktop.el` to save/restore the sessions...?

hoverwinter:
Yeah, thank you for the suggestions. One big difference is that kys manage session by giving them multiple tags, and a directory can have multiple sessions. Also some functionality such as locking session is under developing, so now they may seem similar. Perhaps I should issue the request a bit later.

tarsius:
> Perhaps I should issue the request a bit later.

Do you think this is ready now?

tarsius:
I am closing this now due to long silence. @hoverwinter please let as know once things are ready.

PR #4149, opened 2016-08-15
indie-buffers

anler:
indie-buffers package recipe, a package that combines indirect buffers and narrowing.

### DIFF
--- a/recipes/indie-buffers
+++ b/recipes/indie-buffers
@@ -1,0 +1,1 @@
+(indie-buffers :repo "anler/indie-buffers" :fetcher github)


purcell:
Thanks. Does this provide anything over and above the established https://github.com/Fanael/edit-indirect package? If not, we strongly prefer not having multiple very similar packages.

anler:
😱 sorry, didn't know about that one! Don't merge it since what I have now is pretty much a subset of that package,  if mine evolve in a different direction I'll create a new PR. Thanks

purcell:
I guess that proves it was a good idea! :-)

Fanael:
And I guess that proves edit-indirect could do with better publicity.

purcell:
Yeah, we've seen two reimplementations of it recently, and I hate to see people waste coding time. A nicer README might help people take the package more seriously? And maybe I'll add it to my own config, because the more-downloaded packages are often discovered more easily. Better keyword search on melpa.org might help too.

P.S. Having looked at the code just now, you're binding `"C-c '"`, but you really shouldn't because that's reserved for end user customisation.

Fanael:
> A nicer README might help people take the package more seriously?

Sure, everything will be nicer than no readme at all.

> Having looked at the code just now, you're binding `"C-c '"`, but you really shouldn't because that's reserved for end user customisation.

We've had this discussion before: https://github.com/melpa/melpa/pull/2198#issuecomment-64381776. It's the same key binding `org-edit-src-code` uses, so even thought it's less than stellar, it's staying.

PR #4150, opened 2016-08-15
Added recipe for digit-groups

adamsmd:
Hi, I recently developed the `digit-groups` package and would like to release it on MELPA.

# Brief summary of what the package does.

The `digit-groups` Emacs package makes it easier to read large numbers by
highlighting digits at selected place-value positions (e.g., thousands place,
millions place, billions place, etc.).

For example, in the text `9876543210.123456789`, the default configuration
formats in bold the 3, 6, and 9 before the decimal (`.`) because they are in
the thousands, millions, and billions positions as well as the 3, 6, and 9
after the decimal (`.`) because they are in the thousandths, millionths, and
billionths positions.

# Direct link to the package repository.

https://bitbucket.org/adamsmd/digit-groups

# My association with the package

I am the developer of this package

# Relevant communications with the upstream package maintainer

None.

### DIFF
--- a/recipes/digit-groups
+++ b/recipes/digit-groups
@@ -1,0 +1,1 @@
+(digit-groups :fetcher bitbucket :repo "adamsmd/digit-groups")


purcell:
Thanks. Quick feedback:
- The package description line is _way_ too long! Keep it snappy, and use the `Commentary` section to provide more info.
- In the `Commentary`, don't only refer people to the custom group docs: this text will be used on the MELPA package page and when pressing "?" in the emacs package list, at which point the user will not have the package installed.
- Autoloading a private function (`digit-groups--set-mode-hooks`) is a code smell, and indeed you're trying to be too clever with those mode hooks. Hooks are for the user to customize, so get rid of `digit-groups-mode-hooks`. What you really want here is a local minor mode `digit-groups-mode` which the user can add to `prog-mode-hook` or similar, and also be able to toggle on/off conveniently. And optionally you can also provide a globalized version of that minor mode to turn it on everywhere. Both of those mode definitions would generally be marked as autoloaded.

Hope that helps!

adamsmd:
Thank you for the feedback.  (As you can guess, I am not an experienced elisp programmer.)  I have made the suggested changes.

purcell:
Looks great now! Just one thing: you're using `defvar-local`, which only arrived in Emacs 24.3. You should therefore either add `(emacs "24.3")` to the `Package-Requires` list, or replace the `defvar-local` with a regular `defvar` plus a subsequent call to the classic `make-variable-buffer-local` function.

adamsmd:
Done.

purcell:
Awesome, thanks.

PR #4153, opened 2016-08-16
Add emit

IvanMalison:
Summary: A collection of functions and macros that are generally useful for the configuration of emacs
Repository: https://github.com/IvanMalison/emit

### DIFF
--- a/recipes/emit
+++ b/recipes/emit
@@ -1,0 +1,1 @@
+(emit :repo "IvanMalison/emit" :fetcher github)


purcell:
Hi! This looks neat, but I can't really tell from the package description, Commentary or even the README what anyone would do with it. Without that, I'm not sure anyone would be likely to use it... :-)

Also, code like `(emit-prefix-selector imalison:multi-line` is really a `defun` of some sort, so a name beginning `def` would make things clearer.

syohex:
It has some issue(error and warnings) at installing. I sent PR(https://github.com/IvanMalison/emit/pull/1), please see it.

IvanMalison:
@syohex Thanks! merged!

@purcell I'm hoping that this package can become a home for miscellaneous useful functions that don't have a home elsewhere, something like @bbatsov's [crux](https://github.com/bbatsov/crux), but without a focus on interactive functions. I actually intend to put a whole bunch more stuff in this package, but I haven't quite gotten around to it yet. The functions that are included aren't really representative of what this is supposed to be (except for prefix-selector), so I think it might be useful if I give some examples of the types of thing that I meant to be included in this package.

- join-paths - sort of like python's os.path.join -- takes variadic arguments and appends them together to build a path in a platform agnostic way. The implementation would be something like:

``` emacs-lisp
(defun emit-join-paths (&rest paths)
  (substring (mapconcat 'file-name-as-directory paths nil) 0 -1))
```
- flatten-imenu-indexes would provide a general way to take a nested (multiple menu) imenu index and make it flat
see https://github.com/IvanMalison/dotfiles#flatten-imenu-indexes for an implementation

- fill-or-unfill-paragraph

implementation would probably be taken from http://endlessparentheses.com/fill-and-unfill-paragraphs-with-a-single-key.html

- shell-command-on-region-replace

- copy-buffer-file-name

- copy-buffer-file-path 

I'm not particularly invested in having THIS package be the general emacs config package, but I definitely think that something like it should exist. I'm sure that every one of these functions has been implemented many times over by various emacs users, and that just seems silly to me.

When I proposed this idea on the emacs subreddit a while ago ([here](https://www.reddit.com/r/emacs/comments/3fhbg9/an_emacs_package_for_general_utility_functions/)), people insisted that these types of things belong in the emacs core, but I'm not entirely convinced that that is the case. It takes too long for new additions to the core to trickle down to most users, which makes it difficult to iterate and aggregate quickly. I'd love to hear your and other's thoughts on this because you, as melpa maintainers are uniquely qualified to appraise the appropriateness of this concept.

purcell:
Yeah, I'm generally very much against the idea of packages which are a collection of unrelated functionality. I wasn't a fan of `crux`, and I've declined many packages which fit that general description: at the end of the day, such a package is much more likely to be a dumping ground for an individual user's configuration helper functions than to be a package which is generally useful to many users.

Much better to provide individual smaller packages, even if they just contain one or two functions. For example, check out my [unfill](https://github.com/purcell/unfill/blob/master/unfill.el) package, which does the fill-and-unfill thing you mentioned. `join-paths` would fit nicely in the `f` package.

purcell:
(Though the toggle command in that `unfill` package is pretty crap: I'll replace it with Artur's version... :-) )

tarsius:
@IvanMalison I think the reasons given by @purcell for not including this package are sound. Do you intend to explain why you still think this should be available from Melpa? Or have you given up on that and we can close this issue?

IvanMalison:
@tarsius I don't have much more to say, other than that I'm still not totally convinced by those arguments. I would really like to share, for example, the compose implementation in this package between some other packages that I maintain, and without something like this, I would have to include separate copies of it in each package.

@purcell writes
> at the end of the day, such a package is much more likely to be a dumping ground for an individual user's configuration helper functions than to be a package which is generally useful to many users.

I don't think that is a criticism that can be fairly levied against this package. Perhaps it does not appeal to either of you, but I think that the prefix-selector macro, in particular, is potentially useful to a large subset of emacs users. The practice of grouping several conceptually related actions in single keybinding using universal prefix argument is an established emacs idiom, but writing a function to do so is much more tedious than it needs to be. Have you really never found yourself writing something like:

``` emacs-lisp
(defun my-avy (arg)
  (interactive "p")
  (cond ((equal arg 1)
         (call-interactively 'avy-goto-char))
        ((equal arg 4)
         (call-interactively 'avy-goto-word-1))))
```
That may not seem like a ton of code, but I actually got this wrong the first time I wrote it, because I forgot exactly how the non-raw prefix value works.

prefix-selector reduces this to
``` emacs-lisp
(emit-prefix-selector my-avy avy-goto-char avy-goto-word-1)
```
and it has the added benefit that it interprets numeric prefix arguments and universal prefix arguments in an unsurprising way. The [implementation](https://github.com/IvanMalison/emit/blob/master/emit.el#L113) of this macro is sufficiently complicated that it deserves a community maintained home, but it is small enough that it probably does not deserve its own package.

With all of that said, it seems that there is a very strong consensus against this package. I have a hard time seeing what harm its existence would do, but I appreciate the time that you guys have spent looking at this and trying to explain why it is a bad idea. Feel free to close this request if you feel there is no way forward.

GuiltyDolphin:
@IvanMalison I don't think there's anything wrong with the *individual* functions, but as @purcell mentioned:

> Much better to provide individual smaller packages, even if they just contain one or two functions. For example, check out my unfill package, which does the fill-and-unfill thing you mentioned. join-paths would fit nicely in the f package.

Unrelated functionality just isn't that useful - but small packages providing *specific* and *related* functions could be much more handy - a toolbox instead of a garage, if you take my drift. We already have the garage - Emacs, everything else should really be a toolbox.

Look at [dash.el](https://github.com/magnars/dash.el) for example - it aims to do one thing (list manipulations), and it does that *very* well.

Think of it as separation of concerns - if each package focuses on *one* thing, then you can easily pick and choose packages to specialise your Emacs - if you have a 'general' package you would likely end up with a great deal of functionality you wouldn't use and would ultimately just clutter your config.

Just make loads of teensy packages which you can add to later :smile:

tarsius:
@purcell I think we need you to make a decision here.

I had another look at the code and while I don't understand much of what's going on, it appears things actually are fairly related. There's come functions for composing functions, then there are some functions for giving names to those compositions, and some functions for creating commands that dispatch a "member of a composition" based on the prefix argument.

But I think the package description "This package provides functions that are generally useful for configuring Emacs." is bad.

IvanMalison:
I sort of already conceded above

> Feel free to close this request if you feel there is no way forward.

because it seems that the consensus is against anything like this.

I think @purcell made his thoughts pretty clear above (he's not a fan)

> I had another look at the code and while I don't understand much of what's going on, it appears things actually are fairly related.

If you, an experienced emacs user, has trouble seeing whats going on with these functions, then its probably pretty unlikely that most members of the community will find any of this to be much use.

> But I think the package description "This package provides functions that are generally useful for configuring Emacs." is bad.

What would be a better name. The functions do depend on each other in some cases, but they aren't really aimed at any particular end.

tarsius:
A bit of review:

* The name of `emit-prefix-selector` should contain the word "define". I also don't think this is a particular useful way to define a dispatcher command. I for one would never be able to remember how many prefix arguments are required to invoke a certain variant. I recommend you look into the `hydra` package instead.

(In the following I am not trying to pile on. After you said:

> If you, an experienced emacs user, has trouble seeing whats going on with these functions, then its probably pretty unlikely that most members of the community will find any of this to be much use.

I was actually trying to come up with a more balanced review, but turns out that even after reading the code more carefully I still don't understand what is going on.)

* There's a lot of macro stuff going on here. It's tempting to do this, but over the years I have come to appreciate a bit a bit of boiler-plate. By using such macros you make it much harder for others to understand what is going on. Or make potential readers not even want to read it - that's what I meant when I said "I don't understand much of what's going on": I don't want to invest the time that would be required to understand this because I think what this does actually isn't as much as the amount of code implies. That's a bit harsh, sorry. But I think it serves no one to make something simple complicated for the sake of having a nice abstraction.

 There is nothing wrong with e.g. a macro to define your particular flavor of prefix arguments, but you are going overboard by defining a macro, that uses another macro, that uses another macro, that defines a macro, which expands to another macro, which defines an alias, which calls another macro. That other macro is defined separately and expands to the definition of the interactive command we are doing all this for. I have no idea why any of the stuff described in the first sentence is even necessary or even just makes things easier.

You can read more about why I think macro abuse should be avoided here: http://emacs.stackexchange.com/questions/16375/assigning-same-value-to-multiple-variables/16377#16377.

@purcell I vote for not adding this.

PR #4158, opened 2016-08-16
eval-expression improvement

hydandata:
https://github.com/jwiegley/eval-expr

- Emacs lisp mode syntax table
- Can fix mistakes
- Can display results in a buffer if its too big
- Pretty Printing and more


purcell:
Thanks!

PR #4160, opened 2016-08-18
Added version-regexp for enh-ruby-mode

zenspider:
I think this should fix the issue for stable? @Trevoke ?

I tested this via the usual path. But I'm unsure that it also tests the stable packaging/data...

### DIFF
--- a/recipes/enh-ruby-mode
+++ b/recipes/enh-ruby-mode
@@ -1,3 +1,4 @@
 (enh-ruby-mode :repo "zenspider/enhanced-ruby-mode"
                :fetcher github
+               :version-regexp "[^0-9]*\\(.*\\)"
                :files ("*.el" ("ruby" "ruby/erm.rb" "ruby/erm_buffer.rb")))


Trevoke:
I did a quick check in emacs with the regexp builder and I *think* that should do it.

Since your tags are of the form `rel2015-07-11`, using that regexp and passing the result to `version-to-list` gives us `(2015 -4 7 -4 11)` . I think that's OK for Melpa but I don't exactly know what it does with it, so we do need to wait for @purcell to weigh in.

purcell:
No, that's not a good format for us, I think, exactly because of the `version-to-list` conversion you quoted. Other projects use something like `20150711`, though, so perhaps that would be an option? (`v20150711` would also be handled out of the box.)

zenspider:
@purcell hrm? Explain "because of the `version-to-list` conversion you quoted"?

purcell:
Dashes are only used in Emacs notion of versions to denote suffixes like `-alpha`, `-git` etc, so it handles those in a weird way. Dots would of course be fine.

Trevoke:
He is referring to my comment in this thread.

zenspider:
@purcell the part I'm not understanding is how this is a problem. You get a list of numbers. As long as I'm consistent with my tagging scheme, they should sort fine, no?

purcell:
> @purcell the part I'm not understanding is how this is a problem. You get a list of numbers. As long as I'm consistent with my tagging scheme, they should sort fine, no?

Yeah, so Emacs takes the `(2015 -4 7 -4 11)` version number in the available package list, and expands it to get the URL of the file it corresponds to on the package server: `enh-ruby-mode-2015snapshot7snapshot11.el` in this case. So that's pretty nonsensical, right? Ultimately, we're talking about version numbers for Emacs-specific things, so we all have to go along with the Emacs conventions (whether or not we like them) or end up with a mess like this.

zenspider:
OK. That makes sense. I didn't realize you were using it for packaging as well...

So... there's nothing wrong with this patch... it's entirely with my data. I've (hopefully?) renamed all my tags to match `vYYYYMMDD`. This should fix the packaging for melpa-stable, yes?

purcell:
Yeah. In fact, this patch should now be unnecessary, because the default tag pattern is already picking up your new release tags.

PR #4161, opened 2016-08-18
Add flimenu recipe

IvanMalison:
Summary: This package provides functions and a mode to flatten imenu indexes
Repository: https://github.com/IvanMalison/flimenu

### DIFF
--- a/recipes/flimenu
+++ b/recipes/flimenu
@@ -1,0 +1,1 @@
+(flimenu :fetcher github :repo "IvanMalison/flimenu")


syohex:
`advice-add` was introduced at Emacs 24.4, but this package requires Emacs 24 or higher versions.

purcell:
Thanks. You might consider dropping the `dash` dependency because you're only using it for `--first`, and I think `cl-find` with a `:test` option will do exactly the same thing.

PR #4162, opened 2016-08-18
Pull Request template

zonuexe:
This template may be able to reduce the work of maintainers.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,21 @@
+
+
+### Brief summary of what the package does
+
+[Please write an introduction of the package.]
+
+### Direct link to the package repository
+
+https://example.com/your/awesome_package
+
+### Your association with the package
+
+[Are you the maintainer? have you contributed? do you just like the package a lot?]
+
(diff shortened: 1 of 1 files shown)


tarsius:
I think it would be a good idea to add a template. Maybe not this one though. @purcell knows best what should go in there.

I have added an issue template for Magit and that really helped. Previously I have added images of hands pointing at information that I really wanted users to read. That didn't really help, very many still ignored it. Adding an issue template on the other hand helped a lot.

Maybe something like *Before writing a new recipe, please read the [contributing guidelines](https://github.com/melpa/melpa/blob/master/CONTRIBUTING.md).* is enough!

(I would also add *Please make sure you include a link to the repository of the package you are adding.*, because, for packages hosted on github, I find it very annoying having to copy the partial url from the recipe and then append that to the url in the address bar after removing the parts that have to be remove.)

tarsius:
@purcell Since you haven't merged this a month later, I'm gonna assume that you won't ever. But I would still like to encourage you again to add something like this. I am pretty sure it would help!

purcell:
Sorry, I was too slow looking at this, but I did intend to give it some thought, and I appreciate your help. In the end I've committed a slight variation on this template, so let's see how things go!

PR #4164, opened 2016-08-19
Add eslint-fix recipe

codesuki:
eslint-fix lets you format javascript files using ESLint.
ESLint provides more formatting and style options compared to js2-mode.
It's a very simple script but it proved so useful to me that I thought I would share it.

### DIFF
--- a/recipes/eslint-fix
+++ b/recipes/eslint-fix
@@ -1,0 +1,3 @@
+(eslint-fix :repo "codesuki/eslint-fix"
+			:fetcher github
+			:files ("eslint-fix.el"))
--- a/recipes/eslint-fix
+++ b/recipes/eslint-fix
@@ -1,3 +1,2 @@
 (eslint-fix :repo "codesuki/eslint-fix"
-			:fetcher github
-			:files ("eslint-fix.el"))
+			:fetcher github)


syohex:
You need not to specify `:files` property. Please read https://github.com/melpa/melpa#recipe-format 

> :files optional property specifying the elisp and info files used to build the package. Please do not override this unless the default value (below) is adequate, which it should usually be:

```
("*.el" "*.el.in" "dir"
 "*.info" "*.texi" "*.texinfo"
```

syohex:
I sent PR(https://github.com/codesuki/eslint-fix/pull/1). Please see it.

codesuki:
Thank you for your help! I removed the `files` property.

purcell:
Thanks! Looks good, and I'll go ahead and merge, but there are a couple of things that could be improved:
- Use `error` rather than `message` in this case
- Not every buffer will have a non-`nil` value of `buffer-file-name`, because it's fine to have buffers that aren't backed by files. You should probably use `shell-command-on-region` instead, which will work in both cases.
- Consider making the `eslint` path a `defcustom` (with a corresponding `defgroup` to contain it, of course)

codesuki:
Thanks for your suggestions and merging!

I will make some changes! 
About `shell-command-on-region`, I have a few questions.
I wanted to use it before but `eslint` only works on files. If the input it not a file it will throw an error.
I could write the contents of a buffer not backed by a file to a temporary file, format it and replace the buffer contents. What do you think?

purcell:
> I wanted to use it before but eslint only works on files. If the input it not a file it will throw an error.
> I could write the contents of a buffer not backed by a file to a temporary file, format it and replace the buffer contents. What do you think?

Yes, you can use `make-temp-file`, and then `write-region` to write the current buffer's contents to it, and then open the file, run `eslint` on it with `shell-command-on-region`, and finally copy the modified contents back to the original buffer.

PR #4165, opened 2016-08-19
Add yara-mode, using github fetcher

pstray:
Thought it would be nice to have yara-mode in melpa.

### DIFF
--- a/recipes/yara-mode
+++ b/recipes/yara-mode
@@ -1,0 +1,3 @@
+(yara-mode
+ :fetcher github
+ :repo "binjo/yara-mode")


syohex:
I have sent PR. Please check it(https://github.com/binjo/yara-mode/pull/1/files).
And you should write package summary at first line(after `---`)

pstray:
Was that last part of the comment for me or for binjo?

purcell:
> Was that last part of the comment for me or for binjo?

Either of you could do it, thanks to the magic of Pull Requests. ;-)

syohex:
> Was that last part of the comment for me or for binjo?

Current first line is [this](https://github.com/binjo/yara-mode/blob/470732ed534b027f98e8ce65b6c4279215179bf1/yara-mode.el#L1)

```lisp
;;; yara-mode.el ---
```

This line should be like

```lisp
;;; yara-mode.el --- Major mode of yara fule file
```

Paclage summary is displayed in `package-list-packages`

binjo:
Hi,

Thank you @pstray and @syohex , I just committed 2 minor changes in my own repo.
Let me know if other edit needs, hope it can get more pull request in the future. ;)

purcell:
Thanks - looks good. Don't rebind `C-j`, though -- that's more of a personal preference than a requirement for this major mode.

purcell:
Also:
- `(define-key yara-mode-map [remap comment-dwim] 'yara-comment-dwim)` should just be placed at the top level, after the `define-derived-mode` form: otherwise you're just altering the global `yara-mode-map` every time you start the mode, which has no benefit.
- `font-lock-defaults` is always buffer-local, so there's no need to use `make-local-variable` there.

binjo:
Thank you purcell, committed as your suggestions.

PR #4166, opened 2016-08-19
Add x-path-walker recipe

Lompik:


syohex:
I have sent PR(https://github.com/Lompik/x-path-walker/pull/1).

Why didn't you add `helm`(or `helm-core`) to dependency header ? Can users use this package without `helm` ?

Lompik:
I added `helm` dependency for now. I ha in mind to add option for other backends like avy.

syohex:
I sent PR again about helm(https://github.com/Lompik/x-path-walker/pull/2)

purcell:
> I added helm dependency for now. I ha in mind to add option for other backends like avy.

That's usually a sign you're doing something wrong. If you're going that way, you should instead consider a core package, and then separate add-on packages for the `helm` and `avy` extensions.

Lompik:
I see. I don't foresee an avy backend anytime soon but if that is happening I'll split the package.

purcell:
Great, thanks

PR #4167, opened 2016-08-21
recipes/x-path-walker: Include xpathwalker module

marsam:
x-path-walker needs the `xpathwalker` which is added to [python path](https://github.com/lompik/x-path-walker/blob/6c6d839/x-path-walker.el#L36-L39) before calling the command

cc: @Lompik

### DIFF
--- a/recipes/x-path-walker
+++ b/recipes/x-path-walker
@@ -1,2 +1,3 @@
 (x-path-walker   :fetcher github
-                 :repo "lompik/x-path-walker")
+                 :repo "lompik/x-path-walker"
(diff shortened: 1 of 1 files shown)


Lompik:
Yes this looks right. I did not realize it would not pull the whole repo. Thanks !

purcell:
Thanks @marsam. 

> Yes this looks right. I did not realize it would not pull the whole repo. Thanks !

The README describes how to build/install/test the package before submitting, so that you know it's working. :-)

Lompik:
Guilty as charged :). Good to know for next time !

PR #4168, opened 2016-08-22
y-ret-or-n-p recipe

SanCoder-Q:
**y-ret-or-n-p.el** is a simple emacs plugin for replacing the original function `yes-or-no-p` and `y-or-n-p` in order to make your agreement faster.

https://github.com/SanCoder-Q/y-ret-or-n-p.el

### DIFF
--- a/recipes/y-ret-or-n-p
+++ b/recipes/y-ret-or-n-p
@@ -1,0 +1,1 @@
+(y-ret-or-n-p :fetcher github :repo "sancoder-q/y-ret-or-n-p.el")


purcell:
Thanks. Quick feedback:
- In the `Commentary` section, please don't just refer people to the github page. There should be some text in this section to describe what the package is and how to use it, so maybe copy some of the README into the Commentary.
- It's bad practice to copy a big core Emacs function, make a tiny change and then turn it into a package. There's usually an easier solution, e.g. something like the following, which I figured out by reading the docstring:

```el
(defun y-ret-or-n-p (prompt)
  (let ((query-replace-map (let ((map (copy-keymap query-replace-map)))
                             (define-key map (kbd "RET") 'act)
                             map)))
    (y-or-n-p prompt)))
```

SanCoder-Q:
@purcell Thanks for your quick feedback, and for reviewing my code. Firstly, I agree with you for both feedbacks. 
For the second one, I've tried to give a name hiding and original function call like your suggestion to make it happen. But it is hard to change answer when it's non-interactive mode. Also, it is hard to manipulate the query message from outside. Any idea?

tarsius:
> But it is hard to change answer when it's non-interactive mode. Also, it is hard to manipulate the query message from outside. Any idea?

I don't think you can do this in this case and that it is therefore okay to copy the original definition. But you should add a comment which states when you last synchronized your definition with the original and what Emacs version that corresponds to.

SanCoder-Q:
I have added some description and warning to remind the user that this package could increase the risk of confirming by mistake. Also, thanks for @dotemacs , an illustration table has been added to explain how the package works.

tarsius:
I have looked at this again and noticed that what your package does can essentially be achieved  using just

```lisp
(define-key query-replace-map "\r" 'act)
```

I don't think it is necessary to mention "RET" in the prompt because, beside "y", there already are a few other keys that can be used to confirm, namely "SPC" and "Y", and they also aren't mentioned in the prompt.

purcell:
Yep, agreed. I'm going to decline this, sorry, for the reasons @tarsius and I have outlined above.

PR #4169, opened 2016-08-22
Recipe programming-mode

Liu233w:
https://github.com/Liu233w/programming-mode.el

Input method to exchange your number line with the mark above it.

For example, when you press 1, ! will be entered. If ! was binding to function other than self-insert-command, it will be called. But if Emacs have evil installed, it's only worked at insert-state. Pressing S+1 will enter 1, etc.

It's a input method, so it's only worked on insert mode in current buffer, so chords like C-x 2 will act as normal, and the key won't be translated in minibuffer.

### DIFF
--- a/recipes/programming-mode
+++ b/recipes/programming-mode
@@ -1,0 +1,1 @@
+(programming-mode :repo "Liu233w/programming-mode.el" :fetcher github)


syohex:
I suppose package name `programming-mode` is confused. Because Emacs already has `prog-mode` it is a base mode of major-mode of programming language mode.

Liu233w:
OK, I'll change it



发自我的小米手机
在 Syohei YOSHIDA <notifications@github.com>，2016年8月22日 下午3:54写道：

I suppose package name programming-mode is confused. Because Emacs already has prog-mode it is a base mode of major-mode of programming language mode.

PR #4171, opened 2016-08-22
emagician-fix-spell-memory: upgrade, move repo, rename

jonnay:
https://github.com/jonnay/fix-muscle-memory

- Old package was in my starter kit dir, now has it's own github repo
- Old package name was kind of goofy, has a new name now

This commit also deletes the old recipe.  I set :old-names in the new
recipe.

Let me know if I need to do anything differently with the rename.  Thank you!


purcell:
Great, thanks.

PR #4174, opened 2016-08-23
pcap-mode recipe

orgcandman:
This package implements pcap-mode, a wireshark-like frontend integrated
into emacs.

### DIFF
--- a/recipes/pcap-mode
+++ b/recipes/pcap-mode
@@ -1,0 +1,1 @@
+(pcap-mode :repo "orgcandman/pcap-mode" :fetcher github)


syohex:
- All functions and global variable should starts with package prefix because Emacs has only one name space. This is for avoiding name confliction.
- Major-mode function should be defined by `define-derived-mode` rather than `defun`
- `add-hook` configuration should be buffer local
- autoload cookies(`;;;###autoload`) are missing. They are necessary for lazy loading

orgcandman:
1. Okay, will change all the functions and variables to have pcap-mode-- as a prefix.
2. No problem, will do that.
3. it is?  what is currently non-local?  kill-buffer-hook is always buffer local.
4. okay - I think just the pcap-mode itself should be autoloaded.

orgcandman:
I have updated pcap-mode - please confirm if it is now acceptable.

syohex:
> it is? what is currently non-local? kill-buffer-hook is always buffer local.

It is 4th argument of `add-hook`.

orgcandman:
It is currently t for the 4th argument - that is local, yes?

syohex:
> It is currently t for the 4th argument - that is local, yes?

Yes

orgcandman:
So, are there any other changes needed?  Is this able to be merged, now?

purcell:
Thanks, and thanks also @syohex for jumping in with advice while I've been busy for a few days. Quick feedback from me:
- `pcap-mode-group` should just be `pcap-mode`
- The `pcap-mode--` prefix conventionally denotes private symbols, while `pcap-mode-` would denote public symbols: you probably want the latter in most cases here. Sorry for any confusion.
- When building shell commands, consider being paranoid and using `shell-quote-argument` liberally
- `M-x checkdoc` will guide you through tidying up docstrings.
- Now that you're using `define-derived-mode` (yay!), you can delete [this line](https://github.com/orgcandman/pcap-mode/blob/365b994f1f0f77b55fe6096c75eeb55829954db3/pcap-mode.el#L293), [this line](https://github.com/orgcandman/pcap-mode/blob/365b994f1f0f77b55fe6096c75eeb55829954db3/pcap-mode.el#L298) and [this form](https://github.com/orgcandman/pcap-mode/blob/365b994f1f0f77b55fe6096c75eeb55829954db3/pcap-mode.el#L302).
- Feel free to autoload the `auto-mode-alist` modification

Hope that helps. This looks really useful - nice work!

orgcandman:
Thanks.  I've done changes based on the feedback.  I'm still working on a solution that would properly quote arguments (because filters and switches need to be treated differently).  Is it acceptable with just escaping filename and capture interface?

purcell:
> Thanks. I've done changes based on the feedback. I'm still working on a solution that would properly quote arguments (because filters and switches need to be treated differently). Is it acceptable with just escaping filename and capture interface?

Probably, yep.

PR #4175, opened 2016-08-24
Move base16-emacs to new location

belak:
**What the package does:** base16-theme is a collection of themes which are generated from sets of 16 colors.

**Package repository:** https://github.com/belak/base16-emacs

**Package association:** I am the new maintainer.

I recently took over the base16-emacs project. I'm moving the theme files into a separate directory, so I figured now would be a good time to make sure the repo location is updated as well.

I have tested `make recipes/base16-theme` and `package-install-file` to ensure this package still works as designed.

### DIFF
--- a/recipes/base16-theme
+++ b/recipes/base16-theme
@@ -1,1 +1,1 @@
-(base16-theme :repo "mkaito/base16-emacs" :fetcher github)
+(base16-theme :repo "belak/base16-emacs" :fetcher github :files ("*.el" "build/*.el"))


belak:
Ah, I see, it errors... what would the best way be to resolve this? Should it just be broken for a little bit?

belak:
I put the only copy of the theme files in the root and I'll move them after this gets merged.

belak:
I've updated the PR description with all relevant information as specified in CONTRIBUTING.md.

purcell:
Thanks! Do you plan to submit a subsequent PR removing the `"*.el"` pattern?

purcell:
(Or perhaps you're going to keep the `base16-theme-pkg.el` file in the root... in which case you should use `:defaults` instead of `"*.el"` please.)

belak:
The idea was that `base16-theme.el` would stay in the root (because that defines the important part) and all the schemes would go in build so you can see the important stuff. There's nothing in build right now because I didn't want to break the package while the migration was happening.

What does `:defaults` actually do? I think, because of the separate build dir, that wouldn't work though

belak:
So, would it be `(base16-theme :repo "belak/base16-emacs" :fetcher github :files (:defaults "build/*.el"))`?

purcell:
> The idea was that base16-theme.el would stay in the root (because that defines the important part) and all the schemes would go in build so you can see the important stuff. There's nothing in build right now because I didn't want to break the package while the migration was happening.

:+1:


> What does :defaults actually do? I think, because of the separate build dir, that wouldn't work.

It expands to `*.el`, plus some other stuff, and excludes test files.


> So, would it be (base16-theme :repo "belak/base16-emacs" :fetcher github :files (:defaults "build/*.el"))?

Yeah, exactly!

PR #4177, opened 2016-08-24
Add launch-mode, using github fetcher

iory:
launch-mode is a major mode for editing ROS's launch(http://wiki.ros.org/roslaunch/XML) file.

### DIFF
--- a/recipes/launch-mode
+++ b/recipes/launch-mode
@@ -1,0 +1,3 @@
+(launch-mode :fetcher github
+              :repo "iory/launch-mode"
+              :files (:defaults "*.py"))


syohex:
Does it requires `helm` absolutely ? If so, `helm` should be added to package dependency header.

purcell:
Thanks. Quick feedback:

- Please use the standard elisp boilerplate structure, as generated by `M-x auto-insert`. In particular, you should have a short `Commentary:` section.
- The keymap setup code is very unconventional, what with the prefix and concat-ing of key strings. Use a separate [prefix keymap](https://www.gnu.org/software/emacs/manual/html_node/emacs/Prefix-Keymaps.html) here, and allow the user to bind it to whatever prefix they like in `launch-mode-map`.
- `C-c r` is a key binding [reserved for end users](https://www.gnu.org/software/emacs/manual/html_node/elisp/Key-Binding-Conventions.html), though, so you may not use that prefix.
- In Lisp, `)` doesn't like to live on its own line, either in our recipes or your code!

Hope that helps!

tarsius:
@iory Here's a friendly ping for you.

iory:
Sorry my reply is late.

@syohex 
Using ido instead of helm.
I think ido is emacs's default mode. This solution is OK?

@purcell 
I added some commens and delete prefix key map.

> In Lisp, ) doesn't like to live on its own line, either in our recipes or your code!

Oh, I did not known that. Fixed it.

syohex:
- I have sent PR https://github.com/iory/launch-mode/pull/2
- [This code](https://github.com/iory/launch-mode/blob/d057bddf65a876030450118e33e8572a10a241e4/launch-find-definition.el#L144) uses undefined function `launch-goto-offset`. Is this correct ?

iory:
@syohex 
Thank you for your PR.
I added missed launch-goto-offset function. https://github.com/iory/launch-mode/blob/master/launch-find-definition.el#L168

purcell:
The lonely `)` are still there in https://github.com/iory/launch-mode/blob/master/launch-find-definition.el, but I'll go ahead and merge now in any case.

PR #4179, opened 2016-08-25
Summary: Some calendar dates for Norway

unhammer:
Link: https://github.com/unhammer/calendar-norway.el

Association: Author.

### DIFF
--- a/recipes/calendar-norway
+++ b/recipes/calendar-norway
@@ -1,0 +1,1 @@
+(calendar-norway :fetcher github :repo "unhammer/calendar-norway.el")


purcell:
Thanks! Quick feedback:

- Please remove the redundant "for Emacs" from the package description line
- It's very unusual placing `eval-after-load` forms inside a function. Why would you not just put all the code that's in `calendar-norway-common-settings` at the top level of the file instead?

unhammer:
Thanks for the feedback =D Both are implemented now. I guess lazy-loading like that was a bit overkill …

purcell:
Yay, thanks!

PR #4180, opened 2016-08-25
Add evil-swap-keys recipe

wbolster:
please add this new package, written by me.

https://github.com/wbolster/evil-swap-keys

### DIFF
--- a/recipes/evil-swap-keys
+++ b/recipes/evil-swap-keys
@@ -1,0 +1,1 @@
+(evil-swap-keys :repo "wbolster/evil-swap-keys" :fetcher github)


syohex:
I sent PR for older Emacs. `map-apply` function is very new, only emacs 25 users can use it.
Please see it(https://github.com/wbolster/evil-swap-keys/pull/1)

wbolster:
thanks @syohex, the issue is fixed in master now.

wbolster:
ping. any other things i need to take care of or is this good to go in now?

syohex:
I suppose it looks good.

purcell:
What's the relationship between this and [number-lock](https://github.com/Liu233w/number-lock.el/blob/master/number-lock.el)? They seem to have similar goals. Your package is larger and more documented, so I'm wondering if it has to be specific to `evil`? Perhaps an approach like that of `number-lock` could make it more widely applicable? I feel a bit weird about merging both this and #4170 at almost the same time! /cc @Liu233w

wbolster:
yeah, i also noticed this package... after writing mine. see 
https://github.com/Liu233w/number-lock.el/issues/1 since i contacted the maintainer

my package evil-swap-keys is very evil specific since keyboard handling is modal in evil, which means naive key remappings will not work correctly. my package handles evil states and various evil commands (even from other evil extension packages) in an intelligent way. all these things do not apply to vanilla emacs.

my package does offer much more features, both available keyboard swaps and how they ate applied (buffer local, major mode specific).

also, my package is much more flexible to configure.

and it has docs.

purcell:
Thanks. I've merged both, because I think they have distinct use cases.

PR #4181, opened 2016-08-25
Added recipe for transfer-sh package

Brillow:
Please consider adding this package to melpa. It includes a function to upload regions or buffers to the open-source file-sharing service transfer.sh.


purcell:
Looks good. You should almost certainly be using `shell-quote-argument` to escape the strings you're passing to `curl`, to prevent nasty suprises.

purcell:
Also, consider adding an `;;;###autoload` cookie comment for the `transfer-sh-upload` command.

PR #4182, opened 2016-08-26
Add po-mode, included in the gettext repository

knu:
`po-mode` is included in the gettext repository which is too huge (just to clone) for such a small piece of elisp software.  So if you put it in your local Cask file, it takes a long time to install or check for updates for the single package.

It seems there have been some interest in putting it into GNU ELPA, but there has not been much progress as of yet.
https://lists.gnu.org/archive/html/bug-gettext/2015-07/msg00040.html

There are some elisp libraries that depend on po-mode, such as po-mode+ and mo-mode which are currently not packaged, so it would be beneficial to have it in MELPA.

### DIFF
--- a/recipes/po-mode
+++ b/recipes/po-mode
@@ -1,4 +1,4 @@
 (po-mode
  :fetcher git
  :url "git://git.savannah.gnu.org/gettext.git"
- :files ("gettext-tools/misc/*.el"))
+ :files ("gettext-tools/misc/po-mode.el"))
--- a/recipes/po-mode
+++ b/recipes/po-mode
@@ -1,0 +1,4 @@
+(po-mode
+ :fetcher git
+ :url "git://git.savannah.gnu.org/gettext.git"
+ :files ("gettext-tools/misc/*.el"))


purcell:
How huge is the repo? We have packages for a couple of libraries (erlang-mode?) which live in very large parent repositories, and it's not ideal for us either.

knu:
I guess it takes a long time to fetch it probably because the repo has a very long history and Cask performs deep-clone (git clone without `--depth`) just to check out the latest version of the file.

purcell:
> I guess it takes a long time to fetch it probably because the repo has a very long history and Cask performs deep-clone (git clone without --depth) just to check out the latest version of the file.

Yeah, and we'll do the same, because we potentially need the version history of any file mentioned by the recipe's `:files` list.

But hey, it's only (!) 64MB total, so it's fine for us to add this.

I think, though, that I'll only add `po-mode.el` -- `po-compat` should be unnecessary in any Emacs with `package.el`, and `start-po.el` is a hand-written autoloads file. You might consider submitting a pull request upstream to add `;;;###autoload` cookie comments for the necessary forms in `po-mode.el`, so that the `po-mode-autoloads.el` file generated by `package.el` will be equivalent to `start-po.el`.

purcell:
In fact the package formatting is invalid, and therefore the metadata can't be read by `package-buffer-info`:
```el
;;; po-mode.el -- major mode for GNU gettext PO files
```
should be
```el
;;; po-mode.el --- major mode for GNU gettext PO files
```
(Note the extra `-`)

Any chance you could submit an upstream pull request to fix that? I'm not very keen to add it otherwise, since it will show up as "No description available" in MELPA.

knu:
Thanks for taking a close look.  I will!

knu:
OK, my patches have been merged to the upstream.  How does it look now?

purcell:
Great job, thanks!

manuel-uberti:
Thanks a lot for this one

PR #4183, opened 2016-08-27
Replace "*.el" with :defaults in base16-theme

belak:
This is a follow up to #4175 as recommended by @purcell, since it just makes sense.


purcell:
Great, thanks!

PR #4184, opened 2016-08-27
habitica recipe

abrochard:
habitica.el is an extension to interact with https://habitica.com/static/front, a RPG style habit tracker.
I like the concept of getting exp and loot when you keep good habits, so I wanted to make it accessible via an org-mode interface.

https://github.com/abrochard/emacs-habitica

Looking forward to your feedback.

### DIFF
--- a/recipes/habitica
+++ b/recipes/habitica
@@ -1,0 +1,1 @@
+(habitica :repo "abrochard/emacs-habitica" :fetcher github)


syohex:
- Missing autoload cookie`;;;###autoload`
- You should not use `C-c t` for default binding. It breaks [Emacs key binding convention](https://www.gnu.org/software/emacs/manual/html_node/elisp/Key-Binding-Conventions.html). It says `C-c letter` is reserved for users.

abrochard:
- Ah! Just added in this [commit](https://github.com/abrochard/emacs-habitica/commit/fc86ce1bdd7440692d8ddde4da37f2da607167f2). Is it the correct location?

- I don't think I am using it? Everything is behind the prefix `C-x t`

syohex:
Oh sorry, I mistook.

purcell:
Thanks!

abrochard:
Thank you!

PR #4189, opened 2016-08-29
Skip unneeded files in skewer-reload-stylesheets recipe

NateEag:
I added a lame test helper file a while ago, and only recently noticed
that the original recipe picked it up for packaging, where it has no
use.


purcell:
Thanks for this. The fault was the `"*.el"`, which should have been `:defaults`, since that excludes tests automatically. So the new version should be `:files (:defaults "*.js")` - would you mind updating the recipe please?

NateEag:
Thanks for pointing me in the right direction here - new commit pushed.

Sorry for the slow response - real life is pretty busy right now.

purcell:
Super, thanks.

PR #4190, opened 2016-08-30
Changed Repo location

jonnay:
The old location was kinda goofy.  Sorry about that.

Also spaces not tabs in recipe.


purcell:
> The old location was kinda goofy. Sorry about that.

No problem.

> Also spaces not tabs in recipe.

Yay!

PR #4193, opened 2016-08-31
update spray recipe repository URL.

stardiviner:


syohex:
This is duplicated with #4192

PR #4194, opened 2016-08-31
Add dired-launch recipe

thomp:
dired-launch.el provides a launcher for dired-mode.  In a nutshell, it lets you select a file and then launch an external application with that file.

https://github.com/thomp/dired-launch

I wrote the package and, finding that I use it frequently, thought others might benefit from the package as well.

### DIFF
--- a/recipes/dired-launch
+++ b/recipes/dired-launch
@@ -1,0 +1,1 @@
+(dired-launch :fetcher github :repo "thomp/dired-launch")


syohex:
- All global variable and functions should start with same package prefix for avoiding name confliction
  - https://github.com/thomp/dired-launch/blob/d8a9d793d343887b1bc80573ebfd8c3ff8396e9b/dired-launch.el#L18)
  - Use `dired-launch-` instead of `dl-`
- Missing autoload cookie(`;;;###autoload`)
  - https://github.com/thomp/dired-launch/blob/d8a9d793d343887b1bc80573ebfd8c3ff8396e9b/dired-launch.el#L38
  - https://github.com/thomp/dired-launch/blob/d8a9d793d343887b1bc80573ebfd8c3ff8396e9b/dired-launch.el#L49
- Use `add-hook` instead of `setq`. `setq` overwrites previous value.
  - https://github.com/thomp/dired-launch/blob/d8a9d793d343887b1bc80573ebfd8c3ff8396e9b/dired-launch.el#L60
- `dired-mode` assigns both `l` and `L` in default keymap. So you should not assign those keys by default.
  - https://github.com/thomp/dired-launch/blob/d8a9d793d343887b1bc80573ebfd8c3ff8396e9b/dired-launch.el#L62-L63
  - https://github.com/thomp/dired-launch/blob/d8a9d793d343887b1bc80573ebfd8c3ff8396e9b/dired-launch.el#L67-L68

thomp:
The above should be addressed in https://github.com/thomp/dired-launch/commit/1cd3e79fe3c1147164423d9045ca6ee7526bb5d2

purcell:
Thanks. More feedback:
- `"C-c l"` and `"C-c L"` are [reserved for end user customization](https://www.gnu.org/software/emacs/manual/html_node/elisp/Key-Binding-Conventions.html) so you may not use those bindings
- `(concat "open " (dired-get-filename))` will fail if the filename contains a space
- It's not true that if `system-type` is not `windows`, then it must be a mac. You should check more explicitly.
- Instead of `dired-launch-default-key-bindings` and modifying `dired-mode-hook`, just define a little non-global minor mode with `define-minor-mode`, then you can tell the user to add that mode to their `dired-mode-hook`, and you won't have to modify `dired-mode-map` etc.

thomp:
Thanks -- great focus on code quality. Above concerns should be addressed by https://github.com/thomp/dired-launch/commit/3e0961dd55746dcc3a8ee125f04e8b309908a45f and https://github.com/thomp/dired-launch/commit/9afa143081014d081a4fd44a9dead23b797f7827 .

purcell:
Thanks. Sent a further fix in https://github.com/thomp/dired-launch/pull/2.

purcell:
Great, thanks!

PR #4197, opened 2016-09-02
Add ereader

bddean:
Ereader.el provides ereader-mode to view epub ebooks. Org-ebook.el provides functions for linking to epub books from Org and opening them in ereader-mode. I think they belong in the same package because they are pretty tightly coupled.

I wrote it and maintain it.

See repository at https://github.com/bddean/emacs-ereader

Thanks

### DIFF
--- a/recipes/ereader
+++ b/recipes/ereader
@@ -1,0 +1,2 @@
+(ereader :fetcher github
+         :repo "bddean/emacs-ereader")


purcell:
Thanks. Quick feedback:

- `(picture)` is not a valid package dependency, because the mandatory version number is missing. But in any case, you shouldn't depend on built-in packages.
- `"C-c r"` is [reserved for end-user customization](https://www.gnu.org/software/emacs/manual/html_node/elisp/Key-Binding-Conventions.html), so you must not use that binding
- Re. `(put 'ereader-mode 'mode-class 'special)` - should you maybe just use `special-mode` as the parent mode instead of `view-mode`?

Looks good otherwise!

bddean:
Thanks for the response. Those should be fixed in commit 2dcec1e.

vyp:
in case you just forgot, you may also want to put a free software license on them. considering it will be distributed on melpa and everything.

bddean:
Ok -- I added put the GNU Pubic License in at commit 91f9db8

purcell:
Super, thanks!

PR #4200, opened 2016-09-02
Add zoutline

abo-abo:
It's a small library that I want `lispy` and `worf` to depend on.

### DIFF
--- a/recipes/zoutline
+++ b/recipes/zoutline
@@ -1,0 +1,2 @@
+(zoutline :repo "abo-abo/zoutline"
+          :fetcher github)


tarsius:
Thanks!

tarsius:
Here's what @purcell would have told you had I not already merged this: You should add: `;;; zoutline.el ends here` and a library commentary.

abo-abo:
> You should add: ;;; zoutline.el ends here and a library commentary.

Why, though? Which part of Emacs makes use of that comment?

tarsius:
It will keep `package-buffer-info` from raising an error about that line missing. Originally the convention of adding the "ends here" line was created so that users could tell whether the file they manually downloaded was corrupted or not. Personally I don't think we should have to add this line, but currently we have to.

tarsius:
And the commentary is displayed by `describe-package`.

abo-abo:
OK, thanks for the info. Would be cool if the package manager did not need this in the future.

purcell:
> OK, thanks for the info. Would be cool if the package manager did not need this in the future.

I don't know that I agree -- yes, that line's a bit silly, but I'd rather have a slightly outdated formatting standard than a bunch of files which don't look consistent. :-)

purcell:
(As a general rule, I'd recommend that all package authors begin a new file with `M-x auto-insert`. It really helps get the structure right, and has completion for the standard keywords etc.)

PR #4201, opened 2016-09-02
Add ox-tufte recipe

dakrone:
This adds the recipe for the ox-tufte org-mode export backend -
https://github.com/dakrone/ox-tufte

This is a specialized HTML backend supporting an unmodified Tufte CSS
stylesheet - https://edwardtufte.github.io/tufte-css/


purcell:
Thanks!

PR #4203, opened 2016-09-03
Add plain-theme recipe

yegortimoshenko:
https://github.com/yegortimoshenko/plain-theme

### DIFF
--- a/recipes/plain-theme
+++ b/recipes/plain-theme
@@ -1,0 +1,1 @@
+(plain-theme :fetcher github :repo "yegortimoshenko/plain-theme")


purcell:
Wouldn't it make more sense to just turn off `font-lock-mode`?

yegortimoshenko:
No, see [the rationale](https://github.com/yegortimoshenko/plain-theme#rationale).

Turning off `font-lock-mode` will disable useful things that are not syntax highlighting: CIDER error underlining, `markdown-mode` italics and underlined URLs. And `org-mode` is nearly useless without `font-lock-mode` turned on.

purcell:
Cool, thanks for the pointer. Makes sense!

purcell:
P.S. It's a bug if a package like CIDER ever explicitly enables `font-lock-mode` -- they should never do that. /cc @bbatsov

yegortimoshenko:
Oops, it looks like this issue was solved 3 months ago, see clojure-emacs/cider#1768. Now, that it has been fixed, I don't know other packages that conflict with `font-lock-mode` off, so I'll remove it from my rationale. I still hold the first argument, though.

purcell:
Fair enough. :-)

PR #4204, opened 2016-09-05
Add nix-buffer package.

shlevy:
This package adds a command to modify the buffer state according to
a nix expression.

The package repository is at https://github.com/shlevy/nix-buffer.

### DIFF
--- a/recipes/nix-buffer
+++ b/recipes/nix-buffer
@@ -1,0 +1,2 @@
+(nix-buffer :fetcher github :repo "shlevy/nix-buffer"
+            :version-regexp "v\\.\\(.*\\)")


syohex:
I have sent PR(https://github.com/shlevy/nix-buffer/pull/1) about package summary.

shlevy:
Pushed a fix, thanks!

purcell:
Thanks. You've redefined `f-traverse-upwards`, which is a big no-no (see the [elisp conventions](https://www.gnu.org/s/emacs/manual/html_node/elisp/Coding-Conventions.html)). Instead, define a locally-renamed version of it, if you must use it at all -- seems like the built-in `locate-dominating-file` would do the same thing anyway.

shlevy:
Fixed, thanks

purcell:
Great, thanks.

PR #4206, opened 2016-09-05
Add org-clock-csv recipe.

atheriel:
This commit adds a recipe for the [`org-clock-csv`](https://github.com/atheriel/org-clock-csv) package, which extracts clock entries from org files and converts them into CSV format.

It's intended to facilitate analysis of clock data in external programs (e.g. R, Python, even Excel).

I wrote and maintain the package. I did some searching for similar approaches, but couldn't find anything at the time, so as far as I know there's no duplication of functionality. I thought that it might appeal to the community.

### DIFF
--- a/recipes/org-clock-csv
+++ b/recipes/org-clock-csv
@@ -1,0 +1,1 @@
+(org-clock-csv :repo "atheriel/org-clock-csv" :fetcher github)


purcell:
Thanks, merged!

I'd encourage you to run `M-x checkdoc` and fix the docstrings as directed. (`flycheck` displays the same information, if you use that.)

PR #4208, opened 2016-09-05
rename org-eww to org-preview-html

lujun9972:


purcell:
Thanks. This said "rename", but was actually just the addition of the new package. Will there be another pull request to delete `org-eww`? I think that's necessary, right?

purcell:
Sorry, ignore my comment above -- seems like things are fine.

PR #4209, opened 2016-09-05
Fix helm recipe.

thierryvolpiatto:
* recipes/helm: helm-match-plugin.el have been removed to helm-multi-match.el.


purcell:
Great, thanks.

PR #4210, opened 2016-09-05
Add swift3-mode recipe

taku0:

Major-mode for Apple's Swift programming language.

This is a fork of chrisbarrett/swift-mode with a new indentation engine, supporting Swift 3.

https://github.com/taku0/swift3-mode

I'm the maintainer.

None

### Did you read [CONTRIBUTING.md](https://github.com/melpa/melpa/blob/master/CONTRIBUTING.md)?

Yes.

### DIFF
--- a/recipes/swift3-mode
+++ b/recipes/swift3-mode
@@ -1,0 +1,3 @@
+(swift3-mode
+ :fetcher github
+ :repo "taku0/swift3-mode")


purcell:
Thanks!

The metadata is not correct here: you have a `-pkg.el` file, which means that any metadata in `swift3-mode.el` will be ignored. You've got a `Package-Requires` header in there which specifies a dependency on Emacs 24.4, and that dependency will therefore be masked. I suggest you delete the `-pkg.el` file (MELPA will generate one) and just put all the metadata directly in the `swift3-mode.el` file.

In the meantime, all the identifiers are in the form `swift-mode*`, but they should be `swift3-mode*`, otherwise it will be impossible to have both this package and `swift-mode` installed simultaneously without them clashing.

In any case, perhaps it would make more sense to submit your indentation engine changes to the upstream `swift-mode`? Is there a reason why that would not be possible?

taku0:
Removed `-pkg.el` and renamed symbol prefixes.

> Is there a reason why that would not be possible?

I made several PR to the chrisbarrett/swift-mode but they rejected the PRs due to increased complexity.

https://github.com/chrisbarrett/swift-mode/pull/74
https://github.com/chrisbarrett/swift-mode/pull/126

chrisbarrett/swift-mode doesn't seem to be active now. Both the code and the Issues/PRs are left for a long time with critical bugs (https://github.com/chrisbarrett/swift-mode/issues/125#issuecomment-244169790).

purcell:
Thanks. I'll merge this now, but you also need to rename the `defgroup` from `swift` to `swift3`, plus the references to that name from `defcustom` forms, e.g. [this one](https://github.com/taku0/swift3-mode/blob/604ffeb94028e9207aaa91de7ebd28d711f6dc79/swift3-mode-indent.el#L38).

Another small tip is that [this line](https://github.com/taku0/swift3-mode/blob/604ffeb94028e9207aaa91de7ebd28d711f6dc79/swift3-mode.el#L138) is redundant, because `define-derived-mode` generates it for you. :-)

purcell:
I'm sorry to hear that the changes couldn't be integrated into `swift-mode`. Given that there are a bunch of open issues there, if I were @chrisbarrett I'd be inclined to simply make you a maintainer so that you can help fix things centrally without needing to fork. That'd be much better for everyone, I would think. What do all you high-volume `swift-mode` contributors think? (cc @ap4y, @bbatsov, @uk-ar)

bbatsov:
I'm fine with the idea. I definitely think it's a mistake to start another major mode.

chrisbarrett:
I've opened chrisbarrett/swift-mode/issues/133 for transferring swift-mode an organisation.

purcell:
Super, thanks Chris! @taku0's work looks really good, so I hope you can all sort something out. Ping me later if - as I hope - this `swift3-mode` package can be deleted! :-)

taku0:
Thanks all.
Migration will take a little time. swift3-mode lacks some features and automated tests swift-mode has.

PR #4211, opened 2016-09-06
Mac Classic Theme

codeswish:
TextMate inspired Emacs 24 theme

### DIFF
--- a/recipes/mac-classic-theme
+++ b/recipes/mac-classic-theme
@@ -1,0 +1,1 @@
+('mac-classic-theme :fetcher github :repo "ahobson/mac-classic-theme")


vyp:
@codeswish No need for the `'` in the recipe. :)

milkypostman:
would you add a screenshot to the README?

syohex:
This package does not conform Emacs packaging format.

- https://www.gnu.org/software/emacs/manual/html_node/elisp/Packaging-Basics.html#Packaging-Basics
- https://www.gnu.org/software/emacs/manual/html_node/elisp/Simple-Packages.html#Simple-Packages

codeswish:
I'll fix these issues and submit a new pull request.

PR #4213, opened 2016-09-07
Add go-add-tags

syohex:
`go-add-tags` is a command inspired by vim-go GoAddTags command. It inserts go struct tags.

https://github.com/syohex/emacs-go-add-tags


syohex:
Thanks

tarsius:
@milkypostman @purcell It might be a good idea to give @syohex access too. It felt kinda strange merging this, considering that he is doing much more than me.

syohex:
Thanks

PR #4214, opened 2016-09-07
Add ng2-mode recipe

AdamNiederer:
Hi there!

This is a major mode for Angular2 development. For now, it just highlights some domain-specific syntax atop its parent modes.
https://github.com/AdamNiederer/ng2-mode

### DIFF
--- a/recipes/ng2-mode
+++ b/recipes/ng2-mode
@@ -1,0 +1,1 @@
+(ng2-mode :fetcher github :repo "AdamNiederer/ng2-mode")


syohex:
I sent PR about packaging(https://github.com/AdamNiederer/ng2-mode/pull/1).

And you should add autoload cookie (`;;;###autoload`) to major modes and `auto-mode-alist` setting for lazy loading. (https://www.gnu.org/software/emacs/manual/html_node/elisp/Autoload.html)

AdamNiederer:
I just added the autoload comments and merged the packaging PR. Thanks a lot for the help!

purcell:
Thanks. Quick review comments:
- Your package is called `ng2-mode`, but there's no matching file in the package. Instead, you've got `ng2-ts.el` and `ng2-html.el`. The package metadata (including dependencies) will be picked up from the matching file.
- The defgroup should probably be called `ng2` rather than `angular2`
- [These lines](https://github.com/AdamNiederer/ng2-mode/blob/8a872183e9fb053414f879f2f60b5c6fbb5e755b/ng2-ts.el#L69-L71) are unnecessary - define-derived-mode generates them for you, though you should provide its NAME and DOCSTRING params.
- Same in ng2-html.el

I feel like maybe this is two separate packages.

AdamNiederer:
Hi - thanks for the review!
I'd prefer having one package to promote interoperability between the two modes.
Everything else should be resolved.

purcell:
Cool. Sent you https://github.com/AdamNiederer/ng2-mode/pull/2.

PR #4216, opened 2016-09-08
Added recipe for auto-img-link-insert

tashrifsanil:
auto-img-link-insert is an emacs package that makes inserting images from the web much easier and quicker

### DIFF
--- a/recipes/auto-img-link-insert
+++ b/recipes/auto-img-link-insert
@@ -1,0 +1,2 @@
+
+(auto-img-link-insert :fetcher github :repo "tashrifsanil/auto-img-link-insert")


syohex:
- All global variables and functions should start with package prefix for avoiding name conflict.
- `provide` statement is missing
- package footer is missing
- autoload cookie is missing
- Do not use absolute path for starting process. If user does not install command there, then `start-process` is failed. BTW you can use `make-directory` function instead of `mkdir`. non-UNIX platform like Windows does not have `mkdir`. Emacs function is portable than command.
- You should use `buffer-file-name` or similar function instead of `c-get-current-file`. It is a function of `cc-mode`.
- There are byte-compile warnings.

See also
- https://www.gnu.org/software/emacs/manual/html_node/elisp/Packaging.html#Packaging

tashrifsanil:
I've made changes to make **auto-img-link-insert** more suitable for packaging

purcell:
Thanks. Quick review feedback:
- `(require 'subr-x)` means you need to declare a dependency on Emacs 24.4.
- *Every* top-level identifier should have the name prefix `auto-img-link-insert-`, or _maybe_ `autoimg-`.
- The header is still malformatted. Start with an empty file, run `M-x auto-insert`, answer the prompts, and insert your code in the middle.
- Still missing a `(provide)` form
- I'm pretty sure those `defvar` should not be global variables
- The code in `extract-file-format` is repetitive -- there are built-in functions which deal with file extensions, and you should simply need a list of image file extensions. I'm pretty sure there'll already be such a list in a built-in Emacs library though.
- You've got a "Resources" string hard-coded

Overall I feel like this is code which is useful to you, but that isn't yet likely to be useful for many other people, and is not clean enough for wider distribution.

tashrifsanil:
Thanks for the feedback. I do aplogize about the constant back and forth replies. I'm a complete beginner when it comes to elisp, and this is my first time writing an elisp package, let alone for submission. So I do appreciate your patience with me.

I've fixed the following things:

-  Removed the need for `(require 'subr-x)`
-  Added top level identifiers
- Made an attempt at fixing the header. I used `M-x auto-insert`, but it wouldn't let me proceed after the tags, so the rest I did by hand by looking at other MELPA packages.
- Removed the `defvars`

Some things though I need to clarify, spefically:
- The hard coded "Resources" string
- The `extract-file-format` function

The hard coded "Resources" string: I hard coded the "Resources" string because that is the name of the directory that the package will create, it doesn't matter if it doesn't exist because it will be created. The way the package works is that it will create a "Resources" folder at the current directory of the current file and download the user specificed web image to that directory. So I'm a bit confused as to how to un-hard code this. Should I just declare a variable that has value "Resources" ?

The next thing I want to clarify is the `extract-file-format` function. The thing is that I can't use an emacs functions to get the file format because it's a web link. For example: Suppose I have a link https://images.duckduckgo.com/iu/?u=http%3A%2F%2F4.bp.blogspot.com%2F-pGUXC8ugtWk%2FUSqy_ppFbCI%2FAAAAAAAAD6Y%2FyRSgg40BXlw%2Fs1600%2Fcat%2Bpictures.jpg&f=1  . Notice that the `.jpg` isn't at the very end of the link, and there are other characters are at the end of it. An emacs file extension function would return the file format in this case, because they assume that the file format is a the very end and there are no other characters at the end of it. That is why I had to make use of a custom function. 

And lastly I do agree with you on your last comment. It's just that when the code actually worked,  and the fact that It was useful to me, I felt like sharing it. MELPA was a means to an end on that note. Lastly, I do appreciate all of you going through my code

syohex:
- You should use `auto-img--` instead of `auto-img/` for private functions(This is emacs lisp convention)
- Some functions are still old name
- You should use `let` for local variable, this causes a lot of byte-compile warnings as below.
- `string-remove-prefix`, `string-remove-suffix` are `subr-x` functions

```
In auto-img/get-current-raw-file-name:
auto-img-link-insert.el:34:59:Warning: assignment to free variable
    ‘current-file-name’
auto-img-link-insert.el:35:47:Warning: reference to free variable
    ‘current-file-name’
auto-img-link-insert.el:38:53:Warning: assignment to free variable
    ‘current-file-ext’
auto-img-link-insert.el:37:53:Warning: assignment to free variable
    ‘current-file-dir’
auto-img-link-insert.el:37:32:Warning: reference to free variable
    ‘current-file-dir’
auto-img-link-insert.el:38:9:Warning: assignment to free variable
    ‘current-raw-file-name’
auto-img-link-insert.el:38:32:Warning: reference to free variable
    ‘current-file-ext’
auto-img-link-insert.el:38:70:Warning: reference to free variable
    ‘current-raw-file-name’

In auto-img/create-img-res-dir:
auto-img-link-insert.el:45:42:Warning: assignment to free variable
    ‘current-file-name’
auto-img-link-insert.el:45:22:Warning: reference to free variable
    ‘current-file-name’
auto-img-link-insert.el:46:30:Warning: assignment to free variable
    ‘current-dir’
auto-img-link-insert.el:46:30:Warning: reference to free variable
    ‘current-dir’
auto-img-link-insert.el:48:22:Warning: assignment to free variable
    ‘img-res-dir’
auto-img-link-insert.el:51:9:Warning: reference to free variable
    ‘img-res-dir’

In auto-img/get-local-img-file-loc:
auto-img-link-insert.el:60:3:Warning: assignment to free variable
    ‘img-local-file-loc’
auto-img-link-insert.el:60:3:Warning: reference to free variable
    ‘img-local-file-loc’

In auto-img-link-insert:
auto-img-link-insert.el:70:70:Warning: assignment to free variable ‘img-type’
auto-img-link-insert.el:70:61:Warning: reference to free variable ‘img-type’
auto-img-link-insert.el:76:23:Warning: assignment to free variable
    ‘img-local-file-loc’
auto-img-link-insert.el:75:18:Warning: reference to free variable
    ‘img-local-file-loc’

In auto-img/embed-img-at-cursor:
auto-img-link-insert.el:82:24:Warning: reference to free variable
    ‘img-caption’
auto-img-link-insert.el:85:30:Warning: reference to free variable ‘img-name’
auto-img-link-insert.el:86:24:Warning: reference to free variable
    ‘img-local-file-loc’

In end of data:
auto-img-link-insert.el:92:1:Warning: the following functions are not known to be defined:
    string-remove-prefix, string-remove-suffix, embed-img-at-cursor
```

tashrifsanil:
Thanks Syohei! I'll try to fix this and try again. By the way this may seem a very stupid question to ask but what are you using to compile elisp? I was just evaluating my code in an org-mode source block. Not a very good way of doing things I know.

syohex:
> what are you using to compile elisp?

Open dired, type `B`.

byte-compile

tashrifsanil:
Thanks Syohei! I've fixed the byte compile warnings. Also just a quick question do I have to close and reopen the pull request when I make a change, just so travis cl can perform checks or is just commenting fine? For now, since I don't know the answer to that I will be closing and repopening the pull request

purcell:
> Made an attempt at fixing the header. I used M-x auto-insert, but it wouldn't let me proceed after the tags, so the rest I did by hand by looking at other MELPA packages.

You can just hit "return" when done entering the tags, then it continues.

In any case, I fixed this and more in https://github.com/tashrifsanil/auto-img-link-insert/pull/1.

> So I'm a bit confused as to how to un-hard code this. Should I just declare a variable that has value "Resources" ?

Yes.

> Notice that the .jpg isn't at the very end of the link, and there are other characters are at the end of it. An emacs file extension function would return the file format in this case, because they assume that the file format is a the very end and there are no other characters at the end of it. That is why I had to make use of a custom function.

Yup, and my fix above reflects that.

You've still got the byte compilation issues to address, so I'll leave that to you. :-)

tashrifsanil:
Thank you so much purcell ! I appreciate the changes. I think the tag issue has something to with spacemacs (my emacs configuration). By the way are byte compile warnings still there. When I compiled it using the method Syohei suggested, I had no warnings whatsover.
byte-compile-1
byte-compile-2

purcell:
If you enabled `flycheck`, it'll show you everything that needs fixing in the file. :-)

tashrifsanil:
Gosh, I'm sorry. I probably should have added subroutine comments. Anyway  I have fixed all the erros that flycheck pointed out. This may sound like a stupid question, but you mentioned that when using `(require 'subrx)` I needed to include a dependency on emacs 24.4. How would I go about that, would  I do that as a comment at the heading, if so what is the syntax for it, an example would be nice? Again I am sorry for the inconveniences

purcell:
You'd add a `;; Package-Requires: ((emacs "24.4"))` header to the metadata, next to `;; Version:` etc.

tashrifsanil:
Alright it's  done, I have added it as a package dependency. I added it after the keywords heading cause, that's how the `powerline` package does it.  Let me know if there is anything else that needs to fixed, or is it fine now?

purcell:
Looking good!

tashrifsanil:
Phew! I appreciate your patience with me everyone, and the things you've shown me. Have a good one everyone

purcell:
Happy to help - thanks for the time you've put into this process.

tashrifsanil:
Hey purcell, do commits awaiting-upstream usually take a long time to actually get committed upstream? I ask because it's been a day, and I just wanted to know if this is the norm because of the size of MELPA or if there's an issue with my repository?

PR #4217, opened 2016-09-08
Add recipe for elscreen-fr

rocher:
ElScreen extension to show screen names in the window title, see [elscreen-fr](https://github.com/rocher/elscreen-fr).


purcell:
Thanks, looks good. You'll need to declare a package dependency on `seq`, though, using the version in `elpa.gnu.org`.

rocher:
Thanks for the feedback, I've fixed that package requirement.

purcell:
Great, thanks.

PR #4218, opened 2016-09-08
Add sort-words.el recipe

dotemacs:
As per subject, sort-words.el added

https://github.com/dotemacs/sort-words.el

### DIFF
--- a/recipes/sort-words
+++ b/recipes/sort-words
@@ -1,0 +1,1 @@
+(sort-words :fetcher github :repo "dotemacs/sort-words.el")


syohex:
- `(provide 'sort-words)` is missing
- `list-to-string` should be named `sort-words-list-to-string` or remove it and use its content directly.

And I suppose you should not check by `(use-region-p)` for function usage(See `sort-lines` implementation)

dotemacs:
Good points @syohex, thanks!

dotemacs:
Hey @syohex 

the omission that shouldn't have been there, have now been fixed:

- prefix for function name: **list-to-string** correctly named **sort-words-list-to-string**
- **provided** added

https://github.com/dotemacs/sort-words.el/pull/3

dotemacs:
And I've removed the (use-region-p) check

purcell:
Thanks, looks good. Maybe run `M-x checkdoc`, which will help you tidy up the docstrings so they conform to the conventions. :-)

PR #4219, opened 2016-09-08
Add inherit-local package.

shlevy:
This package adds infrastructure for inheriting buffer-local
variables.

The package repository is at https://github.com/shlevy/inherit-local.

### DIFF
--- a/recipes/inherit-local
+++ b/recipes/inherit-local
@@ -1,0 +1,2 @@
+(inherit-local :fetcher github :repo "shlevy/inherit-local"
+               :version-regexp "v\\.\\(.*\\)")


purcell:
Thanks! Interesting. An example would really help, as it's not immediately clear what a use case might be.

You should declare a `Package-Requires` dependency on Emacs 24.3, because `setq-local` was not available in earlier versions. Alternatively, use the classic combination of `set` and `make-local-variable` instead.

shlevy:
@purcell Where is the appropriate place to put an example?

syohex:
@shlevy https://github.com/shlevy/inherit-local/blob/8f838e0801b1e3286b65bfe5d578ccd609c51898/inherit-local.el#L8

For example
- https://github.com/emacs-mirror/emacs/blob/1dd4f26ab6c1f14628d9fcf03b0cca7e54d52302/lisp/emacs-lisp/let-alist.el#L6

shlevy:
@syohex @purcell asked me to give an example as well as add the package requires dep, I'm not sure where the example should live

syohex:
Example.

```lisp
;;; inherit-local.el --- Inherited buffer-local variables

;; Copyright (C) 2016 Shea Levy

;; Author: Shea Levy
;; URL: https://github.com/shlevy/inherit-local/tree-master/
;; Version : 1.0
;; Package-Requires: ((emacs "24.3"))

;;; Commentary:
```

Diff.

```diff
diff --git a/inherit-local.el b/inherit-local.el
index 0eb8562..12705ea 100644
--- a/inherit-local.el
+++ b/inherit-local.el
@@ -5,6 +5,7 @@
 ;; Author: Shea Levy
 ;; URL: https://github.com/shlevy/inherit-local/tree-master/
 ;; Version : 1.0
+;; Package-Requires: ((emacs "24.3"))
 
 ;;; Commentary:
 
```

shlevy:
@syohex Sorry, I think you're misunderstanding what I'm asking for. @purcell wants an example of how my code should be used, I'm asking where I should put that example. I know how to add the emacs version requirement.

syohex:
Ah sorry.

syohex:
I suppose README or commentary section.

shlevy:
Updated.

syohex:
Looks good

shlevy:
If it is necessary for merging I'll do that, but to be honest I don't see the point in duplicating the documentation and worry it will get out of sync if it's in two places.

purcell:
Commentary is the main place, please, so that the code can be understood in isolation. README is a nice-to-have, but not essential.

Merging now - thanks!

PR #4220, opened 2016-09-08
Adds recipe for white-theme

anler:
Just another theme ^^

https://github.com/anler/white-theme.el

### DIFF
--- a/recipes/white-theme
+++ b/recipes/white-theme
@@ -1,0 +1,1 @@
+(white-theme :repo "anler/white-theme.el" :fetcher github)


syohex:
There are no `*.el` file. Did you push `.el` file ?

anler:
@syohex totally forgot about it, sorry! is pushed now

purcell:
Thanks. The identifiers named `basic-colors-*` should all be called `white-theme-*` instead, to avoid name clashes. Looks good otherwise.

anler:
@purcell done!

purcell:
Thanks!

anler:
\0/ thank you guys!

PR #3281, opened 2015-11-08
Added receipe for avk-themes

avkoval:
Hello!
This package contains 3 new themes for Emacs. 
I host those themes with BitBucket here:   https://avkoval@bitbucket.org/avkoval/avk-emacs-themes
I made those themes myself on my own taste.
Thank you.

PR #4222, opened 2016-09-09
Added receipe for avk-emacs-themes

avkoval:
Hello!
(I am not sure how to reopen https://github.com/melpa/melpa/pull/3281#issuecomment-155384162) so posting it again, hopefully with all required fixes. ;-)

This package contains 3 new themes for Emacs.
I host those themes with https://github.com/avkoval/avk-emacs-themes
Thank you.

### DIFF
--- a/recipes/avk-emacs-themes
+++ b/recipes/avk-emacs-themes
@@ -1,0 +1,2 @@
+(avk-emacs-themes :fetcher git
+                  :url "https://github.com/avkoval/avk-emacs-themes")
--- a/recipes/avk-emacs-themes
+++ b/recipes/avk-emacs-themes
@@ -1,2 +1,2 @@
-(avk-emacs-themes :fetcher git
-                  :url "https://github.com/avkoval/avk-emacs-themes")
+(avk-emacs-themes :fetcher github
+                  :repo "avkoval/avk-emacs-themes")


purcell:
That's great, thanks, but would you mind using the `github` fetcher instead, and `:repo "avkoval/avk-emacs-themes"` instead of `:url ...`? You can just push a further commit to the recipe branch and I'll squash them when I merge.

syohex:
You should use `provide-theme` instead of `provide`
- https://github.com/avkoval/avk-emacs-themes/blob/master/avk-darkblue-white-theme.el#L187
- https://github.com/avkoval/avk-emacs-themes/blob/master/avk-darkblue-yellow-theme.el#L193

This line is not necessary.
- https://github.com/avkoval/avk-emacs-themes/blob/master/avk-daylight-theme.el#L176

avkoval:
Fixed fetcher and provide-theme. Won't remove last line (... ends here) until its very necessary as then flycheck shows warning there...

purcell:
> Won't remove last line (... ends here) until its very necessary as then flycheck shows warning there...

The line to be removed was the `(provide ...)` form, which was redundant.

purcell:
Otherwise looks good, so merging now. Thanks!

PR #4224, opened 2016-09-09
Add mips-mode

hlissner:
A simple major mode for MIPS assembly code, based off haxor-mode, and written [mainly] for the MIPS Assembly track on [exercism.io](http://exercism.io/).

* I am the author
* Repo: https://github.com/hlissner/emacs-mips-mode

### DIFF
--- a/recipes/mips-mode
+++ b/recipes/mips-mode
@@ -1,0 +1,1 @@
+(mips-mode :repo "hlissner/emacs-mips-mode" :fetcher github)


syohex:
I think assembly code uses 8 tab width in most case, Linux kernel, FreeBSD, golang etc. So I think default value of tab-width should be `tab-width`(user's default value) or 8 instead of 2. (I have never seen 2 tab width in assembly code)

hlissner:
Good point! I've changed it to `tab-width`

purcell:
Thanks!

PR #4225, opened 2016-09-09
Add recipe for evil-org

GuiltyDolphin:
Unfortunately, "evil-org" is already taken by "evil-org-mode", so I've used "evil-Org" instead.

It looks like this might cause some issues? The `provide` is `evil-org` - (which is the same as for `evil-org-mode`).

Would you recommend re-naming _everything_ (package-wise) to `evil-Org` - bit of a pain :/? Perhaps just the main `provide` needs to change, and everything else can stay the same? That being said, it does seem to compile quite happily with the [testing](https://github.com/melpa/melpa/blob/master/CONTRIBUTING.md#testing) instructions.

---

Repo: https://github.com/GuiltyDolphin/evil-org
Summary: Provides Evil integration for Org (including context-sensitive bindings).

### DIFF
--- a/recipes/evil-Org
+++ b/recipes/evil-Org
@@ -1,0 +1,1 @@
+(evil-Org :fetcher github :repo "GuiltyDolphin/evil-org")


purcell:
> Unfortunately, "evil-org" is already taken by "evil-org-mode", so I've used "evil-Org" instead.

Ugh, no capitals please. That's not standard, and I don't think a capital letter is going to stop people getting confused! :-)  You should choose a different package name altogether, with a different `provide`d feature name and identifier prefix.

purcell:
I'll go ahead and close this for now pending a rename. :-)

GuiltyDolphin:
@purcell Do you have any recommendations? The other package using the save `provide` uses `evil-org-mode` on `el-get`, the repo etc.

Perhaps `org-evil`? Or `evilorg`? Hmm... I think I could get used to `org-evil`...

I'll ping you again once I've re-named - thanks.

GuiltyDolphin:
@purcell Okay - I've rebranded to `org-evil`.

Want me to update this PR with the new recipe? (I can rebase and remove the old if yea wish).

GuiltyDolphin:
@purcell I've updated the commentary with a bit more info - it'll probably need updating later when features *other* than bindings are included.

Do you have any other comments/issues?

PR #4227, opened 2016-09-09
Add recipe for monitor

GuiltyDolphin:
Repo: https://github.com/GuiltyDolphin/monitor
Summary: Provides utilities for generating 'hooks' (monitoring expressions etc.)

### DIFF
--- a/recipes/monitor
+++ b/recipes/monitor
@@ -1,0 +1,1 @@
+(monitor :fetcher github :repo "GuiltyDolphin/monitor")


purcell:
Thanks!

There's no information at all about what this does or how to use it, so perhaps you could add a `Commentary` with motivating examples?

Separately, I'd suggest merging everything into `hook.el`, because the artificial separation into two tiny modules isn't very emacsy.

And finally, `hooks` are already something very specific in Emacs, and I don't think that what you're doing is related to that, so I'm pretty sure this isn't a good package name for this functionality.

syohex:
Your package is enabled only by loading(Adding an hook to `post-command-hook` makes Emacs slow) and there no disabled interface(Package users need to remove it manually by `(remove-hook 'post-command-hook 'hook--check-monitored)` for disabling it). I suppose you could add setup interface and disabled interface or add global-minor-mode.

GuiltyDolphin:
@syohex Hmm, good idea - being able to enable/disable it would be useful.

@purcell Naming ain't my thing, is it? Yes, I realised after re-writing this a bit that it should probably be called something like `monitor` instead (which is what it does); I've got quite a bit of re-naming to do!

Thanks for the comments.

GuiltyDolphin:
@purcell I've renamed the package to `monitor` (as it `monitor`s expressions).

purcell:
Much nicer, thanks - merged!

The `:after-hook` in the `define-minor-mode` form looks unnecessary: you should be able to simply make its argument the body of the `define-minor-mode` form, right?

GuiltyDolphin:
@purcell Hmm, I'm not entirely sure - I'll have a look - thanks.

PR #4228, opened 2016-09-10
Add doom-theme

hlissner:
An opinionated UI plugin/theme extracted from my [emacs.d](https://github.com/hlissner/.emacs.d), inspired by the One Dark/Light UI theme in Atom. Includes optional dimming of non-source buffers & minibuffer, a neotree theme with font icons, and (soon) a mode-line config.

* I am the author
* Repo: https://github.com/hlissner/emacs-doom-theme
* ~~Note: The recipe intentionally excludes a file that doesn't exist yet.~~

### DIFF
--- a/recipes/doom-theme
+++ b/recipes/doom-theme
@@ -1,0 +1,3 @@
+(doom-theme :repo "hlissner/emacs-doom-theme"
+            :fetcher github
+            :files (:defaults (:exclude "doom-modeline.el")))
--- a/recipes/doom-theme
+++ b/recipes/doom-theme
@@ -1,3 +1,0 @@
-(doom-theme :repo "hlissner/emacs-doom-theme"
-            :fetcher github
-            :files (:defaults (:exclude "doom-modeline.el")))
--- a/recipes/doom-themes
+++ b/recipes/doom-themes
@@ -1,0 +1,1 @@
+(doom-themes :repo "hlissner/emacs-doom-theme" :fetcher github)


purcell:
Thanks. A few bits of feedback:
- `setq-local` requires Emacs 24.3 or greater, and `advice-add` requires 24.4, so you need to declare a package dependency on the appropriate Emacs version.
- `doom|` and `doom*` are not conventional prefixes. Use `doom--` instead.
- You've got a `doom-theme.el` file, which the Emacs theme system will discover and expect to be able to load as a theme, but it doesn't contain a `provide-theme` form. Probably the code there should be in `doom.el` instead, but not sure.
- `doom-init` is problematic, because it does a bunch of stuff based on whether the frame from which it was called is a graphical one, but in reality there can be multiple frames in an Emacs instance, some graphical and some not. You should instead be using face definitions which have multiple classes for different display types, then you won't need all the `advice` and `find-file-hook` magic, which are not really appropriate in theme code.

Hope that helps!

hlissner:
Thanks for the feedback! 

> You've got a doom-theme.el file, which the Emacs theme system will discover and expect to be able to load as a theme, but it doesn't contain a provide-theme form.

doom-theme.el is just a glue file. However, I'll move the color themes into a `themes/` subdirectory and add that to `custom-theme-load-path` instead. Will that work? Also, do I have to change the recipe or will melpa recurse subdirectories automatically?

> You should instead be using face definitions which have multiple classes for different display types, then you won't need all the advice and find-file-hook magic, which are not really appropriate in theme code.

I'll look into face classes, but unfortunately, I think the find-file hook and advice functions will still be necessary to get dimming to work reliably. Unless face classes can be per-buffer predicate functions or something.  I understand it's inappropriate for themes, but would it help if I simply made `doom-init` optional? Or possibly turn it into a minor mode?

purcell:
> doom-theme.el is just a glue file. However, I'll move the color themes into a themes/ subdirectory and add that to custom-theme-load-path instead. Would that alright? Also, do I have to change the recipe or will melpa recurse subdirectories automatically?

No, we don't recurse by default, and a subdir isn't a great workaround, I'd say. Maybe just call the package "doom-themes" instead, with "doom-themes.el"?

> I'll look into face classes (didn't know about them until now), but I think the find-file hook and advice functions will still be necessary for buffer-local face remapping. Unless a face class can be a per-buffer predicate function or something. I understand it's inappropriate for themes, but would it help if I simply made doom-init optional? Or possibly turn it into a minor mode (disabled by default)?

Yeah, a minor mode would be a good way to handle it.

hlissner:
Alright, done! 

- [X] Renamed package and file to `doom-themes` (updated the recipe too)
- [X] Fixed the unconventional function prefixes
- [X] Added the Emacs 24.4 dependency
- [X] Made doom-init into `doom-buffer-mode`
- [X] Using face classes (still needs work, but I'll keep at it)

Are we good to go?

purcell:
Yay, great - merging now!

PR #4229, opened 2016-09-10
Fix version-regexps for inherit-local and nix-buffer

shlevy:


purcell:
Can you not just remove the version regexps altogether? Looks like the defaults would do the trick here, no?

shlevy:
Ah, I didn't realize there is a default, what is it?

purcell:
`"^[rRvV]?\\(.*\\)$"`

shlevy:
@purcell Thanks, updated

purcell:
Thanks!

PR #4230, opened 2016-09-10
Recipe projectile-variable

zonuexe:
- Store project local variables (property) using [Projectile](https://github.com/bbatsov/projectile) and [Symbol Plists](https://www.gnu.org/software/emacs/manual/html_node/elisp/Symbol-Plists.html).
- https://github.com/zonuexe/projectile-variable
- I am author


purcell:
Thanks!

PR #4233, opened 2016-09-10
Add recipe for gulp-task-runner

NicolasPetton:
This package provides a straightforward integration of [Gulp](http://gulpjs.com/) in Emacs. 

Evaluate `M-x gulp` to get the list of gulp tasks of a project and run a task.

- [x] I'm the author of the package
- [x] The package is on github: https://github.com/NicolasPetton/gulp-task-runner
- [x] The package builds (`C-c C-c` in the recipe buffer)
- [x] The package installs properly via `package-install-file`

### DIFF
--- a/recipes/gulp-task-runner
+++ b/recipes/gulp-task-runner
@@ -1,0 +1,1 @@
+(gulp-task-runner :fetcher github :repo "NicolasPetton/gulp-task-runner")


purcell:
Thanks. Why not just call this "gulp"? It's odd to have the package be `gulp-task-runner`, and then use the `gulp` prefix internally.

purcell:
Also:
- Add an `;;;###autoload`
- `shell-command-to-string` will save a line or two
- Rather than simply calling `compile`, consider using `compilation-start` and naming the buffer `*gulp*` or similar, to avoid clobbering the user's existing `*compilation*` buffer.

syohex:
> shell-command-to-string will save a line or two

I suppose `process-lines` saves lines more.

NicolasPetton:

> Thanks. Why not just call this "gulp"? It's odd to have the package be
> `gulp-task-runner`, and then use the `gulp` prefix internally.

gulp.el is an obsolete package in emacs.

NicolasPetton:
:+1: Thanks guys for the feedback.  I pushed some improvements.

NicolasPetton:
@purcell Is it better now?

syohex:
Merged

PR #4234, opened 2016-09-11
Add sphinx-mode

Fuco1:
I can't believe nobody has written sphinx-mode until now... so I decided to make one.

So far it adds native fontification of `.. code-block::` blocks, but I have some more planned with references and navigation.

Spread the word, I don't want to write all of this myself :D

### DIFF
--- a/recipes/sphinx-mode
+++ b/recipes/sphinx-mode
@@ -1,0 +1,1 @@
+(sphinx-mode :fetcher github :repo "Fuco1/sphinx-mode")


syohex:
- Missing second argument of `defgroup`
- Missing autoload cookie(`;;;###autoload`)
- `dash.el` functions are not used, but it is in package dependencies.

Fuco1:
Dash is my default dependency, it got autoinserted.  I removed it for now.  Fixed the other issues as well.

Fuco1:
Hi, what's the status? Anything else I should fix? :)

purcell:
This is okay to merge, but I'd strongly suggest you delete the `-pkg.el` file: it's redundant, and you should just put the correct metadata in the `sphinx-mode.el` file and use Cask's `package-file` directive.

Fuco1:
@purcell I don't understand.  Without the pkg file you can't install the package.  What if someone doesn't have Cask and doesn't go through melpa?

What is `package-file` directive?  Also, this package has more than one file where I always assumed metadata in "main" file doesn't work.

syohex:
You can also write package dependencies in `Package-Requires` header.

> What is package-file directive? 

[package-file](http://cask.readthedocs.io/en/latest/guide/dsl.html?highlight=package-file#el.function.package-file) reads package metadata(dependencies etc) from the file. So if you write `Package-Requirements` header in the file, you need not to declare package dependencies in Cask file.

Fuco1:
Would there still be the problem with multifile packages?

xuchunyang:
`-pkg.el` is required for multi-file packages. I guess melpa will build `-pkg.el` automatically and ignore your `-pkg.el`, so it is not recommended by purcell. But considering your need, you should provide `-pkg.el` and melpa is ok with it.

purcell:
> Would there still be the problem with multifile packages?

Problem? No. If you use `package-file`, then `Cask` can still generate multifile packages for you -- it will generate a `-pkg.el` file for you from the metadata in the main `.el` file, which is generally what you want.

And MELPA will automatically generate a `-pkg.el` file, similarly, when your package has more than one file. :-)

purcell:
> [from @xuchunyang] But considering your need, you should provide -pkg.el and melpa is ok with it.

No, we'd prefer if you omit it.

PR #4237, opened 2016-09-12
Add recipe for prompts

GuiltyDolphin:
Repo: https://github.com/GuiltyDolphin/prompts.el
Summary: Provides utilities for text-prompts and read-completion.

### DIFF
--- a/recipes/prompts
+++ b/recipes/prompts
@@ -1,0 +1,1 @@
+(prompts :fetcher github :repo "GuiltyDolphin/prompts.el")


purcell:
Is this package really more than a nice implementation of the `read-variable` function which is conspicuously missing from Emacs core? I'm not sure it is, currently, or what the future plans are, so why not just cut this down to the bare minimum and call it `read-variable`?

GuiltyDolphin:
@purcell I'm hoping to do a couple of things:

* Provide more consistent prompts - that is to say, provide better means of normalizing and displaying default values (the current method of doing this seems to vary between prompt-functions, and makes some assumptions about the format of the prompt string that may not be true).
* Provide more completing-read implementations - things like `read-keymap`, variable reading limited to certain prefixes (possibly intelligent from mode) (e.g., read an `emaps` function/command) etc.
* Try and provide improved versions of some existing readers - with perhaps an option to wrap the existing ones with advice.

I've also had some ideas (and come up with a few implementations) of things like `read-regex`, which prompts for a value that must match a regular expression, failing that it re-prompts, that kind of thing.

purcell:
Makes sense, and sounds useful. I feel like a more distinctive name would be better - there's already at least one `prompt-` function in Emacs core, and choosing a _very_ general single-word library name like this can feel a bit grandiose. There isn't much precedent for it among popular libraries. Now, it's not _very_ different, but how about "prompts"?

Even

GuiltyDolphin:
@purcell Hmm, mayhaps - I still quite like `prompt`. I guess `prompts` does make sense given the additional 'prompts' it provides.

I'll bow to your wisdom and rename it to `prompts` - and update the GH repo to `prompts.el` while I'm at it. Thanks for the advice :)

GuiltyDolphin:
@purcell It is done.

GuiltyDolphin:
@purcell By the way - I've received a suggestion to update the GH repo of `monitor` to `monitor.el` to make it easier to identify as an Emacs package.

I think this is a fairly sensible suggestion, but I'm a bit worried about whether or not that might affect some in-place systems (I think GitHub will redirect?); since it is in its early infancy still, do you think it would be fine to update the GH repo to `monitor.el` as well as the `MELPA` recipe?

purcell:
> It is done.

Awesome - merged!

> since it is in its early infancy still, do you think it would be fine to update the GH repo to monitor.el as well as the MELPA recipe?

Sure. I don't bother with that naming convention myself, but it kinda makes sense. Other people would use `emacs-prompts`. Whatever you choose to rename it to, Github will redirect, so MELPA won't break, but we'd appreciate a PR to update the repo path in the recipe anyway.

PR #4240, opened 2016-09-14
Add recipe for gitter

xuchunyang:
- Summary: A Gitter client
- Link: https://github.com/xuchunyang/gitter.el
- Association: I am the author

### DIFF
--- a/recipes/gitter
+++ b/recipes/gitter
@@ -1,0 +1,2 @@
+(gitter :fetcher github
+        :repo "xuchunyang/gitter.el")


syohex:
- https://github.com/xuchunyang/gitter.el/blob/ad8775c31849d2348469420f37bc78201b3d3490/gitter.el#L166

Please check `fboundp` for avoiding overwriting macro

- https://github.com/xuchunyang/gitter.el/blob/ad8775c31849d2348469420f37bc78201b3d3490/gitter.el#L176

Please check `(require 'subr-x nil 'no-error)` instead of `featurep`. `featurep` returns `t` when the library is already loaded. So it may return `nil` even Emacs 24.4 or higher versions users if they does not load library yet.

xuchunyang:
@syohex I have fixed the issues you mentioned. Thanks.

syohex:
Merged

purcell:
I wouldn't have merged this quite yet because there are a couple more issues:
- You shouldn't be patching `defvar-local` etc when they're missing. The [elisp coding conventions](https://www.gnu.org/s/emacs/manual/html_node/elisp/Coding-Conventions.html) are quite clear about how to handle that correctly. Or just depend on Emacs 24.3 or 24.4 if you need that stuff.
- The use of various `(require ...)` forms in `gitter--markup-emoji` isn't good practice, because if a user has multiple libraries installed, you shouldn't be deciding what to load. Either depend on a specific package, or make those markup methods a switchable option.

PR #4245, opened 2016-09-15
Add recipe for sexy-monochrome-theme

nuncostans:


syohex:
I sent PR for packaging format(https://github.com/nuncostans/sexy-monochrome-theme/pull/1). Your package does not conform for packaging format. Please see packaging document before writing package. (https://www.gnu.org/software/emacs/manual/html_node/elisp/Packaging.html)

And why do you specify `no-byte-compile` ? If it is not necessary, please remove it.

nuncostans:
I have fixed packaging format. I think

syohex:
Merged

PR #4246, opened 2016-09-15
Add recipe for conda.el

necaris:
An environment manager for users of the Anaconda open data science toolchain and the `conda` tool for managing environments. Conda is primarily aimed at Python code, but also supports R and other general-purpose packages.

Available on GitHub at: https://github.com/necaris/conda.el -- I am the maintainer.

### DIFF
--- a/recipes/conda
+++ b/recipes/conda
@@ -1,0 +1,1 @@
+(conda :repo "necaris/conda.el" :fetcher github)


syohex:
- `advice-add` and `advice-remove` were introduced at Emacs 24.4 but minimum Emacs version of this package is 24.0.

purcell:
Agree re. `advice-add` etc.

Further feedback from me:

- Please omit `:files` here, since the default is fine.
- `in Emacs` is redundant.
- `conda-preactivate-hook` etc. should be `defcustom`s

(Obviously all sorts of hilarious things can go wrong when trying to temporarily set or apply/remove changes to `exec-path`, but it's hard to avoid that with packages like this that interface with very un-emacsy tools like `rbenv`, `virtualenv` etc.)

necaris:
Thank you @syohex and @purcell for the review! Addressed your comments in https://github.com/necaris/conda.el/commit/83aae1718c626c6dfb4878a29f341a19b076c92c and https://github.com/melpa/melpa/pull/4246/commits/1ecc2acb9292a6d8ff17b78f201f227b51dbc43c :-)

syohex:
Could you squash commits ?

necaris:
@syohex done!

syohex:
Thanks merged

PR #4247, opened 2016-09-16
Add fsbot-data-browser recipe

Benaiah:
Simple package to view the entry database of the resident bot at #emacs on Freenode.

### DIFF
--- a/recipes/fsbot-data-browser
+++ b/recipes/fsbot-data-browser
@@ -1,0 +1,1 @@
+(fsbot-data-browser :repo "benaiah/fsbot-data-browser" :fetcher github)


syohex:
I have sent PR(https://github.com/Benaiah/fsbot-data-browser/pull/1) about autoload cookie.

Benaiah:
I’ve merged it and made a new release (for melpa-stable).

syohex:
Merged

purcell:
@Benaiah You're only using `-map` from `dash` -- why not just use the built-in `mapc` or `mapcar` instead, and drop the dependency?

Benaiah:
@purcell good point, I think it just got left in from when I first wrote the function. I intentionally avoided request.el for the same reason, so I'll replace that.

PR #4249, opened 2016-09-16
Add recipe for magithub

vermiculus:
Magithub is basic integration of the 'hub' GitHub client into Magit's
popup interface.  See hub.github.com for more details.

I'm also submitting this PR for more selfish reasons: this is officially
testing the fork/pull-request flow with the package :smile:

Here's to a successful test :beers:

### DIFF
--- a/recipes/magithub
+++ b/recipes/magithub
@@ -1,4 +1,4 @@
 (magithub :fetcher github
           :repo "vermiculus/magithub"
-          :files ("lisp/magithub.el"
+          :files ("lisp/magithub*.el"
                   "LICENSE"))
--- a/recipes/magithub
+++ b/recipes/magithub
@@ -1,0 +1,4 @@
+(magithub :fetcher github
+          :repo "vermiculus/magithub"
+          :files ("lisp/magithub.el"
+                  "LICENSE"))
--- a/recipes/magithub
+++ b/recipes/magithub
@@ -1,4 +1,4 @@
 (magithub :fetcher github
           :repo "vermiculus/magithub"
-          :files ("lisp/magithub*.el"
+          :files ("magithub*.el"
                   "LICENSE"))


syohex:
I have sent PR(https://github.com/vermiculus/magithub/pull/1). Please check package with minimum configuration before PR to melpa.

recipes/magithub
```diff
@@ -0,0 +1,4 @@
+(magithub :fetcher github
+          :repo "vermiculus/magithub"
+          :files ("lisp/magithub.el"
```
syohex: Please use wildcard or put it at top directory and use `:defaults`(Please see recipe format document(https://github.com/melpa/melpa#recipe-format)). This configuration let us updating recipe if you add other files in the future.

purcell:
My comments for @vermiculus:

- `eval-after-load "magit"` is kinda pointless at the end of a file which `require`s `magit` -- instead, just put the code in that block at the top level, and tell people to add `(eval-after-load "magit" '(require 'magithub))` to their configs.
- `implemented as an Emacs package.` doesn't add anything to the `Commentary` of this Emacs package. ;-)

(Re. the wildcard, @syohex, it would also have been okay to omit the wildcard here: while adding it might seem helpful, it might also lead to accidentally bundling future test files. In a subdirectory, it's probably better to leave the author to specify the file patterns. Of course, I'd agree with your suggestion of putting the elisp at the top level of the repo, since that's more idiomatic, and it's an elisp-only repo.)

purcell:
P.S. Looks handy!

vermiculus:
I based the file itself heavily off of `magit.el` since I don't have the best track record when it comes to conforming to the emacs package ecosystem 😓

I've implemented your suggestions 😄 Turns out you're don't think as clearly at 1am.

vermiculus:
P.S. Thanks!

recipes/magithub
vermiculus: What would using the `:defaults` parameter look like? Something like below?

(magithub :fetcher github
          :repo "vermiculus/magithub"
          :files (:defaults
                  "LICENSE"))

syohex:
@vermiculus Could you remove `:files` attribute ? It is no longer necessary and enough to use default configuration(and I suppose LICENSE file is not necessary, because there is GPL3 description in package file). And please squash commits.

Silex:
@vermiculus: nice!

Does that mean https://github.com/dgtized/github-clone.el becomes more or less obsolete? If I understood correctly, `hub` does everything `github-clone` does and more.

Well, ok maybe the advantage of `github-clone` is that it does not depend on `hub`.

vermiculus:
@Silex thanks!

Yes, the functionality of `github-clone.el` seems (at least on the surface) to be a subset of that of `magithub` -- but the goal of magithub is not necessarily to supercede these tools, but to bring the functionality into the magit workflow.

I'm like to replace dependence on `hub` with elisp-only implementations (or perhaps use the elisp implementations as fallbacks if `hub` is indeed faster), but the current implementation is still useful to get the user-experience I'm looking for.

:smile:

recipes/magithub
purcell: You'd just use `(magithub :fetcher github :repo "vermiculus/magithub")` and omit the LICENSE -- we don't normally package those. I've added that recipe manually now, and closed this PR.

> I based the file itself heavily off of magit.el since I don't have the best track record when it comes to conforming to the emacs package ecosystem

`M-x auto-insert` is your friend. Run it in a fresh, empty emacs lisp file. :-)

PR #4252, opened 2016-09-19
Add suggestion-box recipe

yuutayamada:
This package shows information on the cursor. Mainly I'm going to use this package to support nim-mode, but I think the idea can be used for other programming major-modes as well.
(link: https://github.com/yuutayamada/suggestion-box-el)

### DIFF
--- a/recipes/suggestion-box
+++ b/recipes/suggestion-box
@@ -1,0 +1,1 @@
+(suggestion-box :repo "yuutayamada/suggestion-box-el" :fetcher github)


syohex:
Is it necessary to define functions by `defgeneric` ? There are no `defmethod`.

yuutayamada:
Hi, thank you for your comment:)

I might have misreading about defgeneric's document (this was first
attempt for me to implement defgeneric), but my idea here is providing
basic functionality for FUNCTION(...) style (for inside parenthesis)
and if people don't like default behavior, they can partially change
via `defmethod'.

syohex:
> I might have misreading about defgeneric's document (this was first
> attempt for me to implement defgeneric), but my idea here is providing
> basic functionality for FUNCTION(...) style (for inside parenthesis)
> and if people don't like default behavior, they can partially change
> via `defmethod'.

In such case, `defadvice` or nadvice functions/macros should be used. `defmethod` should be used by definition function for specific type.

```lisp
(cl-defgeneric somefunc (arg)
  (message "%s" arg))

(cl-defmethod somefunc ((arg string))
  (message  "[STRING] %s" arg))

(somefunc 10)
;; => "10"

(somefunc "hello")
;; => "[STRING] hello"
```

yuutayamada:
Should I really use `defadvice`?  Personally I want to leave `defadvice` for end users.

I'm not sure this is convincing, but I implemented this package using `defgenerics` because I saw [xref-js2](https://github.com/NicolasPetton/xref-js2/blob/master/xref-js2.el)'s implementation. (I agree with my package still need some work to do though)

But I'm ok closing this PR and re-challenge to another PR when I did more concrete implementation. (maybe I'll mimic xref's implementation then)

tarsius:
> Should I really use defadvice? Personally I want to leave defadvice for end users.

I think what @syohex meant is that you should use just `defun` and users who want to modify your functions should use `defadvice`. (I would recommend the new `advice-add`.)

yuutayamada:
I reduced defgeneric/defmethod as much as I can and updated the document.

syohex:
Please write documents(docstring) for public API(autoloaded functions).

yuutayamada:
Done

syohex:
Thanks

yuutayamada:
Thank you for merging and review:)

PR #4254, opened 2016-09-20
Add recipe for ede-php-autoload

stevenremot:

ede-php-autoload is an implementation of PHP autoloading in Emacs Lisp, that can be used as a SemanticDB backend.

https://github.com/stevenremot/ede-php-autoload

I am the package maintainer

**None**

### Checklist

### DIFF
--- a/recipes/ede-php-autoload
+++ b/recipes/ede-php-autoload
@@ -1,0 +1,3 @@
+(ede-php-autoload :fetcher github
+                  :repo "stevenremot/ede-php-autoload"
+                  :files (:defaults "ede-php-autoload"))


recipes/ede-php-autoload
```diff
@@ -0,0 +1,3 @@
+(ede-php-autoload :fetcher github
+                  :repo "stevenremot/ede-php-autoload"
+                  :files ("*.el" "ede-php-autoload"))
```
syohex: Is it enough to use `:defaults` instead of `*.el` ?

See
- https://github.com/melpa/melpa#recipe-format

stevenremot:
I missed this paragraph in the doc! I tried it and it has the same result, so this recipe uses `:defaults` now.

syohex:
Good

purcell:
There are a few issues with the code here, and I think this recipe was merged too quickly.

- I don't see how the feature loading can possibly work here. The file `ede-php-autoload/class-loader.el` defines the feature `class-loader`, and `(require 'ede-php-autoload/class-loader)` will fail. That's not how feature loading works in Emacs. What am I missing here?
- I suggest you remove the `-pkg.el` file because MELPA doesn't need it.
- Don't forget to change `"*.el"` to `:defaults` in the `Cask` file too! :-)
- The cases in `ede-php-autoload-composer.el` where you're checking parameters with `stringp` are bogus: what if someone passes a number? The function will silently do nothing. So instead of `(when (stringp path) ...)`, just write `(when path ...)`

stevenremot:
Thank you very much for the review @purcell!

> I don't see how the feature loading can possibly work here. The file ede-php-autoload/class-loader.el defines the feature class-loader, and (require 'ede-php-autoload/class-loader) will fail. That's not how feature loading works in Emacs. What am I missing here?

I am not sure which part should fail to you. The file `class-loader.el` is properly placed in the ` ede-php-autoload`  directory at the root of the package, so the file is properly retrieved by Emacs. Then in the file the provide line properly declares ` 'ede-php-autoload/class-loader`. This should be enough for Emacs to properly load the feature as far as I know.

> I suggest you remove the -pkg.el file because MELPA doesn't need it.

Done

> Don't forget to change "*.el" to :defaults in the Cask file too! :-)

I tried it but apparently Cask does not recognize ` :defaults`:

`Wrong type argument: stringp, :defaults`

> The cases in ede-php-autoload-composer.el where you're checking parameters with stringp are bogus: what if someone passes a number? The function will silently do nothing. So instead of (when (stringp path) ...), just write (when path ...)

If I understand correctly, I should let non-string pass to make the code fail loudly? I applied modifications in this way.

purcell:
> This should be enough for Emacs to properly load the feature as far as I know.

Yes, true. I has misread the `provide` form, evidently. This is quite an unusual arrangement, but an interesting way of doing things.

> I tried it but apparently Cask does not recognize :defaults:

Interesting! It uses `package-build` under the covers, just like MELPA...

> If I understand correctly, I should let non-string pass to make the code fail loudly? I applied modifications in this way.

Yep. Thanks!

stevenremot:
> Interesting! It uses package-build under the covers, just like MELPA...

I tried upgrading cask in case of my version would be a bit out of date, but it did not change anything...

The config line in my Caks file is: `(files :defaults "ede-php-autoload")`  so it seems correct to me...

May cask rely on an older version of package-build that didn't have the ` :defaults`  keyword ?

purcell:
> May cask rely on an older version of package-build that didn't have the :defaults keyword ?

No, I don't think so. Ah well, never mind!

PR #4255, opened 2016-09-21
Add recipe for forth-mode.

larsbrinkhoff:

Programming language mode for Forth.  This is different from the old forth.el and gforth.el.

http://github.com/larsbrinkhoff/forth-mode

I am the author.

### Checklist

### DIFF
--- a/recipes/forth-mode
+++ b/recipes/forth-mode
@@ -1,0 +1,2 @@
+(forth-mode :fetcher github
+            :repo "larsbrinkhoff/forth-mode")


syohex:
I have sent PR about packaging.

syohex:
- I think `forth-interaction-mode.el` function and variable definitions should be moved into forth-mode.el and remove it. Because both files depends on each other.
- Definitions of `forth-indent` and `forth-indent-comment` are missing
- There are too many byte-compile warnings. Please check.
- Is parse.el necessary ?
- I suppose `forth` command should be renamed, for example`run-forth` like other similar commands(`run-python`, `run-ruby` etc)
- And `ensure-forth` should be renamed. It should start with `forth-` prefix.
- Please write any documents for users.

larsbrinkhoff:
Hello @syohex,

This package, while usable as is (I use it myself), is still in an early stage, and is expected to grow.  Hence, I have divided some parts into separate files.  `forth-interaction-mode` is not, or should not, be necessary for using `forth-mode`.

`parse.el` provides functionality to interactively display information about the code.

I will certainly remove all warnings and adress other concerns, but it will not be done immediately since I do not have much spare time.

Are there any particular issues that blocks this submission?

purcell:
I agree with that feedback, and now's the best time to tidy things up if it's to be useful for others.

`parse.el`, if necessary, should be `forth-parse.el`, and have a `provide` form added.

Some more notes from me :-)
- We'd generally suggest dropping the `-pkg.el` file, and instead just putting the correct metadata in `forth-mode.el`. We'll generate the `-pkg.el` file for you.
- Definitions of `forth-forward-sexp` etc. are empty: best to remove placeholder code like this until it does something useful
- Don't write `(add-hook (make-local-variable 'before-save-hook) #'forth-blockify)`: instead, use `(add-hook 'before-save-hook #'forth-blockify nil t)` (i.e. use the LOCAL argument to `add-hook`

syohex:
> This package, while usable as is (I use it myself), is still in an early stage, and is expected to grow. Hence, I have divided some parts into separate files. forth-interaction-mode is not, or should not, be necessary for using forth-mode.

I see. However cycle dependencies should be avoided. Please consider function definition position(file).

larsbrinkhoff:
Thank you @syohex and @purcell,

> However cycle dependencies should be avoided.

I certainly agree.  The intent is that `forth-mode.el` should be usable standalone without any other file loaded.  The other files will require `forth-mode.el` but no other file.

For now, `forth-mode` does call the `forth-block-mode` and `forth-interactive-send` functions.  This should only happen in special circumstances: a unusally formatted Forth source code file in the first case, and a user request in the other.

Is it possible to retain this functionality without having any cyclic dependencies?  I was under the impression that it would be possible to autoload a select few functions on request.

purcell:
> Is it possible to retain this functionality without having any cyclic dependencies? I was under the impression that it would be possible to autoload a select few functions on request.

Yes, you can do that, and I'm going to merge this PR now (thanks!), but I feel like particularly the separation of `forth-block-mode.el` is excessive - it sounds like you're trying to minimise the loading of code, but that's just a few functions plus its dependency on `cl`, so you're not saving anything worthwhile IMO. (PS. You should declare a package dependency on version 0.5 of `cl-lib`, and use that instead of the deprecated `cl` library -- updating the code is usually as simple as adding a `cl-` prefix to the `cl` symbols, eg. `incf` => `cl-incf`.)

larsbrinkhoff:
Thanks!

PR #4256, opened 2016-09-21
Added recipe for evil-colemak-minimal

bmallred:

Colemak configuration for evil-mode with minimal adjustments.

https://github.com/bmallred/evil-colemak-minimal

Maintainer

None

### Checklist

### DIFF
--- a/recipes/evil-colemak-minimal
+++ b/recipes/evil-colemak-minimal
@@ -1,0 +1,4 @@
+(evil-colemak-minimal
+ :repo "bmallred/evil-colemak-minimal"
+ :fetcher github
+ :files ("evil-colemak-minimal.el"))


purcell:
Thanks! Quick feedback:
- Don't use the `evil-colemak-minimal/` - use `evil-colemak-minimal-` instead, as per the [elisp conventions](https://www.gnu.org/s/emacs/manual/html_node/elisp/Coding-Conventions.html).
- Prefix all the top-level symbols such as `(defconst left "s")`, or `let`-bind those values for the specific block of code which uses them to make the definitions local
- `evil-colemak-minimal-mode-toggle` looks bogus, because minor modes are already for toggling stuff. What are you trying to do there?

bmallred:
@purcell thanks for the feedback! I will make the adjustments in the first two bullets and will remove the toggle function. It was more of an attempt as a convenience function than anything else.

Is it preferable to ```let```-bind over ```defconst```?

purcell:
> It was more of an attempt as a convenience function than anything else.

When user's call the minor mode function itself interactively, it acts as a toggle. So I think Emacs already has your back. :-)

> Is it preferable to let-bind over defconst?

They do different things. If you want to define a constant which is accessible by any code in Emacs, then `defconst` is the way to go. If you want to introduce a variable which is only relevant to a block of code, then `let` is your friend.

bmallred:
@purcell good to know! All the changes have been committed and pushed. Thank you for the code review :)

purcell:
Thanks!

PR #4259, opened 2016-09-22
Add parinfer receipe

DogLooksGood:

Parinfer mode for S-expression edit.

https://github.com/DogLooksGood/parinfer-mode

I'm maintainer.

None

### Checklist

### DIFF
--- a/recipes/parinfer
+++ b/recipes/parinfer
@@ -1,0 +1,2 @@
+(parinfer :repo "DogLooksGood/parinfer-mode"
+          :fetcher github)


syohex:
I have sent PR(https://github.com/DogLooksGood/parinfer-mode/pull/5). And there are some byte-compile warnings please check them. (I think `goto-char` + `beginning-of-line` can be replaced with `forward-line`)

DogLooksGood:
Thanks for PR and tips. I've fixed the byte-compile warnings.

syohex:
BTW No window Emacs users(`emacs -nw`) cannot use `C-9`. So I recommend you to use another key binding or bind alternative key binding for such users.

DogLooksGood:
@syohex I've remove the default keybinding to let user customize their own.  The instruction for this has been added to README.

syohex:
Looks good.

purcell:
Thanks. A little more feedback:
- Don't include installation instructions in the `Commentary` section: that text will be displayed on the MELPA package package. Instead, maybe create an `;;; Installation:` comment section.
- Docstrings for (at least) your `(interactive)` functions would be good: `M-x checkdoc` will help you there
- Instead of the package description `Parinfer on Emacs`, which doesn't tell the user anything about what the package does, use `"Simpler Lisp editing"`, since that is how `parinfer` describes itself.

Hope that helps!

DogLooksGood:
@purcell Thanks for the advice. I have done these work and pushed to my repo.

syohex:
Thanks.

PR #4261, opened 2016-09-22
Add bar-cursor

ruscfx:

bar-cursor is an Emacs Lisp package that changes the Emacs cursor from a block into a bar. In overwrite-mode, the cursor will change into a block.

@ajsquared forked bar-cursor from the file by Joseph L. Casadonte Jr. found on his [webpage](http://www.northbound-train.com/emacs.html#MyPackages).

https://github.com/ajsquared/bar-cursor

I am not a maintainer, I have not contributed, I just like the package.

None

### Checklist

### DIFF
--- a/recipes/bar-cursor
+++ b/recipes/bar-cursor
@@ -1,0 +1,2 @@
+(bar-cursor :repo "ajsquared/bar-cursor"
+:fetcher github)


syohex:
Looks good

purcell:
Thanks. Since you're maintaining this now, a couple of suggestions:
- Drop the `byte-compile-dynamic` business
- Convert `bar-cursor-mode` to a call to `define-minor-mode`
- `(if (not frame) ...)` can be rewritten `(unless frame ...)`

PR #4262, opened 2016-09-23
auto-org-md added

jamcha-aa:

auto-org-md.el export a markdown file automatically when you save an org-file

https://github.com/jamcha-aa/auto-org-md

I am a maintainer.

None.

### Checklist

### DIFF
--- a/recipes/auto-org-md
+++ b/recipes/auto-org-md
@@ -1,0 +1,1 @@
+(auto-org-md :repo "jamcha-aa/auto-org-md" :fetcher github)


syohex:
- Please add enable/disable interface. Current implementation enables only loading this package and there is no disable interface(User need to disable it by `remove-hook` manually.
- I think it is enough to use buffer local hook(4th argument of `add-hook`)
- I think that checking major-mode is better than checking file extension(`.org`)

jamcha-aa:
Thank you for your advices. I have modified this program following your comments.
- I added `auto-org-md-mode` in order to toggle this function.
- `add-hook` became buffer local.
- file check became `(eq major-mode 'org-mode)`

purcell:
Thanks. More feedback:
- The `auto-org-md/` prefix should instead be `auto-org-md-` throughout, as per the [elisp conventions](https://www.gnu.org/s/emacs/manual/html_node/elisp/Coding-Conventions.html)
- `auto-org-md-mode` will always be `boundp` when that function runs, so you can remove that check
- Rather than check `(eq major-mode 'org-mode)` every time, why not just report an error when enabling the mode if the current mode is not `org-mode`? Also, maybe use `derived-mode-p` for that check instead of `eq`.

jamcha-aa:
Thank you for fruitful advises. I modified my program as below.
- `auto-org-md-` prefix became simplified.
- `auto-org-md-mode` was modified according to [ergoemacs' elisp-toggle-command](http://ergoemacs.org/emacs/elisp_toggle_command.html)
- `derived-mode-p` was used to check major-mode instead of `eq`.
- `auto-org-md-mode` docstring became `"cycle auto-org-md-mode between on/off"`

purcell:
Thanks!

jamcha-aa:
Thanks for merging and reviews.

PR #4263, opened 2016-09-23
Add emacs-helper recipes

tumashu:

This package is tumashu's emacs-config, somebody of my emacs community want
to it become a package, and I will update it frequently.

https://github.com/tumashu/emacs-helper

I am maintainer and use it every day.

None

### Checklist

### DIFF
--- a/recipes/emacs-helper
+++ b/recipes/emacs-helper
@@ -1,0 +1,1 @@
+(emacs-helper :fetcher github :repo "tumashu/emacs-helper")


syohex:
I think this is very personal configurations. I suppose it is better to prepare install script like [prelude](https://github.com/bbatsov/prelude/) or use [spacemacs](https://github.com/syl20bnr/spacemacs/) like setup(Add init.el and `git clone` it as `~/.emacs`) rather than MELPA package.

tumashu:
This package will be used in different way: I want to it grow as a simple emacs "book" for beginner instead of a "config",  it will be the sourse of documents: http://tumashu.github.io/emacs-helper/

tumashu:
Although there are few "documents" at moment.

purcell:
Yes, this is really personal configuration, and not a coherent package with a specific purpose which will be useful to many people. So, sorry, but we've declined packages like this previously, and I'm going to decline this one too.

PR #4266, opened 2016-09-23
Add open-in-msvs recipe

evgeny-panasyuk:

Emacs package which provides function open-in-msvs to open current file:line:column in Microsoft Visual Studio.

https://github.com/evgeny-panasyuk/open-in-msvs.el

I am creator of this package.

None

### Checklist

### DIFF
--- a/recipes/open-in-msvs
+++ b/recipes/open-in-msvs
@@ -1,0 +1,5 @@
+(open-in-msvs
+ :repo "evgeny-panasyuk/open-in-msvs.el"
+ :fetcher github
+ :files (:defaults "*.vbs"))
+


syohex:
- I think `(eq system-type 'windows-nt)` check should be removed. Because there are possibility of other system-types, cygwin etc(Bash on windows).
- No autoload cookie
- Use `-` separator for variable and functions instead of `/`.

evgeny-panasyuk:
Thank you for review!

- I removed windows-nt check
- Added autoload
- Changed "namespace" separator from / to --

syohex:
And
- I suppose that using `shell-quote-argument` is better than wrapping double quote for escaping shell argument or use `call-process` instead of `call-process-shell-command`. `call-process` arguments are not necessary to escaping characters.

evgeny-panasyuk:
I changed to shell-quote-argument: [commit](https://github.com/evgeny-panasyuk/open-in-msvs.el/commit/c075a7dcf8d013ac3a6e334dc288e28e238f3d9b).
Looks like call-process does not work for .vbs scripts.

syohex:
Looks good.

purcell:
Thanks. More feedback from me:
- The `:files` spec in the recipe could more neatly be written `(:defaults "*.vbs")`
- `Provides function to open current file:line:column in Microsoft Visual Studio` should be shortened, e.g. to `Open current file:line:column in Microsoft Visual Studio`
- Either omit your email address or include a proper one. It's not helpful to include `?????[dot]P???????[at]gmail.???` in the code. Note that your email address can trivially be scraped from github anyway, so I suggest not worrying about publishing it.

Otherwise looks good. :-)

evgeny-panasyuk:
Thank you for feedback!

- Shortened :file spec
- Shortened description
- Removed obfuscated e-mail address

syohex:
Could you squash commits ?

purcell:
Thanks!

PR #4267, opened 2016-09-24
Modify sage-shell-mode recipe

stakemori:
I set the :files keyword of [sage-shell-mode](https://github.com/stakemori/sage-shell-mode) too strict. Unexpectedly, my package is growing. So I would like to change the :files keyword.

### DIFF
--- a/recipes/sage-shell-mode
+++ b/recipes/sage-shell-mode
@@ -1,3 +1,2 @@
-(sage-shell-mode :repo "stakemori/sage-shell-mode"
-                 :fetcher github
-                 :files ("sage-shell-mode.el" "*.py"))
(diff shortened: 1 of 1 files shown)


syohex:
- Could you check whether `:defaults` is available ?
- I suppose documents should be put either top level directory or `doc/`

stakemori:
Thanks for the remark. I use `:default`. I don't understand the second remark. Sorry. The only document in the repo is `README.org` and I copied the default value of :files keyword in the previous commit.

stakemori:
Sorry, I pushed further fix.

stakemori:
I just transferred the repository. So I also changed the `:repo` keyword.

purcell:
Thanks!

PR #4268, opened 2016-09-24
Add recipe for org-wiki.

caiorss:

Org-wiki is an org-mode extension that provides tools to manage and build a desktop wiki where each wiki page is an org-mode file.

https://github.com/caiorss/org-wiki

Author and maintainer. 

None
### Checklist

### DIFF
--- a/recipes/org-wiki
+++ b/recipes/org-wiki
@@ -1,0 +1,1 @@
+(org-wiki :repo "caiorss/org-wiki" :fetcher github)


syohex:
I have sent PR(https://github.com/caiorss/org-wiki/pull/6) about package header.

And please check following comment
- There are many byte-compile warnings
  - Function `directory-list` is not defined, this causes run-time error
- no autoload cookie
- Please use `-` separator(`--` can be used for private functions/variables) instead of `/` for emacs lisp coding convention

caiorss:
Why not  use the slash separator (/) instead of dash (-) because it works like a Clojure namespaces that uses <namespace >/<function> syntax ? 

It is easier in this way to group the functions by common prefix  separating the prefix of the function and grouping them by pseudo "namespaces" that is more completion friendly and also better to the eyes. Why not innovate ? The slash convention doesn't mess anything on Emacs.

By using the slash it is more easier to know to what package the function belongs.

syohex:
> Why not use the slash separator (/) instead of dash (-) because it works like a Clojure namespaces that uses / syntax ? 

It is Emacs Lisp coding convention.

https://www.gnu.org/software/emacs/manual/html_node/elisp/Coding-Conventions.html

caiorss:
All right, I have updated the code according to the Elisp coding conventions. 

But this conventions should be updated. Emacs lacks namespaces and the user has a hard time to spot visually to what package the function belongs. This notation taken from Clojure separates better the prefix of package name from the function name and it makes the functions more search friendly and completion friendly.

purcell:
> But this conventions should be updated. Emacs lacks namespaces and the user has a hard time to spot visually to what package the function belongs. This notation taken from Clojure separates better the prefix of package name from the function name and it makes the functions more search friendly and completion friendly.

We're not here to discuss language design, sorry, but you can bring it up on the emacs mailing list if you feel strongly.

Whatever you think of them currently, the elisp conventions are there to support uniformity of code, so that we can all comfortably read and contribute to each other's code, so this is a matter of community benefit rather than individual preferences. (And I say this as someone who has worked a lot with Clojure, so I agree that its namespaces are nice.)

Some additional specific feedback:
- Don't indent subsequent docstring lines: `  Example:` etc. should be `Example:`.
- `M-x checkdoc` will help you tidy up docstrings. And in general your formatting is not very idiomatic, with lots of random empty lines and orphaned trailing `))`.
- There's a function `org-open-file-with-system` which I think might do the same as your big `org-wiki-xdg-open` function
- `helm-core` is likely an adequate dependency, rather than all of `helm`

Hope that helps!

caiorss:
Thanks for the feedback. 

> M-x checkdoc will help you tidy up docstrings

I've fixed the docstrings. 

> And in general your formatting is not very idiomatic, with lots of random empty lines and orphaned trailing ))

I have fixed the trailing parenthesis and the empty lines. 

> There's a function org-open-file-with-system which I think might do the same as your big org-wiki-xdg-open function

I have removed the function org-open-file-with-system, but the function org-wiki-xdg-open didn't do the same as that. It opens any file with system's default application like *.dwg files with Autocad or *.odf spreadsheet with Libreoffice calc. 

>  helm-core is likely an adequate dependency, rather than all of helm 
Ok. I did it.

purcell:
> I've fixed the docstrings.

At least one still has its subsequent lines indented.

> I have fixed the trailing parenthesis and the empty lines.

Not as far as I can see. There are still lots of empty random lines before/after `(interactive)` forms etc, and groups of `)` on their own lines.

tarsius:
@caiorss are you still planing on making the requested changes, or should we close this?

alphapapa:
@caiorss Ping? :)

tarsius:
It has been half a year and I am closing this now. Feel free to reopen when you have the time to finish this.

Atlas48:
As org-wiki appears to be unmaintained, I've decided to make my own fork, [atlas46/org-lwiki](/atlas46/org-lwiki), and am taking feedback for how to improve the codebase.

PR #4269, opened 2016-09-25
Add hex recipe

AdamNiederer:
Hi everybody. I wrote a small base conversion utility for emacs. It can convert numbers interactively, or convert and replace a number at the point. It can deal with prefixed numbers (`0x`, `0b`, etc), and comes with some fairly accurate base inference for unprefixed numbers.

https://github.com/AdamNiederer/hex


syohex:
I have sent PR(https://github.com/AdamNiederer/hex/pull/1)

AdamNiederer:
Merged the PR, and wrote some more tests.

purcell:
Thanks. This is cool, but I feel like all this should already be possible using `calc`, or at least the implementation of this would be a lot easier using `calc` functions.

I'm also a little bit wary of merging a new package with such a general name. There's a good chance there are `hex.el` files out in the wild, or that there might be one in the elisp standard library sooner or later, so maybe a more unique name (`hexconvert`? `hexwiz`?) would be a good idea.

(I also feel like I've seen a very similar package before, but I can't find it now...)

AdamNiederer:
I haven't used `calc` much, but it doesn't appear to have anything for general base conversion I wrote this primarily for the at-point conversion feature - it helps a lot for assemblers without binary/hex literals.
Would `0xc` be an acceptable name for the package?

purcell:
> I haven't used calc much, but it doesn't appear to have anything for general base conversion.

This looks promising: http://stackoverflow.com/a/1859918/247020

I _suspect_ a little digging there might allow you to remove a bunch of code, which is always a good thing. :-)

> I wrote this primarily for the at-point conversion feature - it helps a lot for assemblers without binary/hex literals.
> Would 0xc be an acceptable name for the package?

Yep, certainly.

AdamNiederer:
It looks like `math-format-radix` (from `calc-bin.el`) is affected by a global from `calc.el`. 

`math-convert-radix-digits` from the same file is undocumented and I can't get a coherent result from it.

purcell:
Thanks. You'll have to rename the recipe file from `hex` to `0xc` too.

> It looks like math-format-radix (from calc-bin.el) is responsible for the functionality, but it's affected by a global from calc.el.

You can let-bind it as needed:

```
ELISP> (let ((calc-number-radix 16)) (math-format-radix 100))
"64"
```

purcell:
Super, thanks Adam.

PR #4270, opened 2016-09-25
Add livedown

shime:

Live offline Markdown previews.

https://github.com/shime/emacs-livedown

I'm an author.

**None**
### Checklist

### DIFF
--- a/recipes/livedown
+++ b/recipes/livedown
@@ -1,0 +1,1 @@
+(livedown :fetcher github :repo "shime/emacs-livedown")


syohex:
I have sent PR(https://github.com/shime/emacs-livedown/pull/7)

And
- Use `-` separator instead of `:` separator for [Emacs Lisp Coding Convention](https://www.gnu.org/software/emacs/manual/html_node/elisp/Coding-Conventions.html#Coding-Conventions)
- No autoload cookie

shime:
Merged your PR, changed the separator and added autoload cookies.

purcell:
Thanks! A little more feedback from me:
- "for Emacs" in the package description is redundant and should be removed
- Similarly, don't use strings like `emacs-livedown`
- The parent group of the `livedown` custom group cannot be itself - instead, place it in one of the standard custom groups (`files`?)
- `livedown-autostart` is bogus, because the file may be loaded before that variable is set. Instead, you should define a global minor mode, which will automatically give you a place to start/stop the server
- These shell commands with `&` 
- When building command strings with `%s`, you need to use the `shell-quote-argument` unless you want bad things to happen when certain characters are found in file paths

Hope that helps!

tarsius:
@shime Are you planing to make the suggested changes? We cannot merge this before you do, or explain why (some of that) is unnecessary/undesirable.

@purcell What's that supposed to mean?

> These shell commands with `&`

purcell:
Hah, sorry. Those `livedown stop` shell commands seem to use `&` to suppress errors, but they could also lead to race conditions.

tarsius:
It has been half a year and I am closing this now. Feel free to reopen when you have the time to finish this.

PR #4271, opened 2016-09-25
Add wosc-import

wosc:

Helper functions to organize Python import statements

https://bitbucket.org/wosc/import-el

I'm the author.

n/a

### Checklist

- [X] I've read CONTRIBUTING.md
- [X] I've built and installed the package using the instructions in the README

### DIFF
--- a/recipes/wosc-import
+++ b/recipes/wosc-import
@@ -1,0 +1,2 @@
+(wosc-import :repo "wosc/import-el" :fetcher bitbucket)
+


syohex:
- no `provide` statement
- Use `--` separator instead of `_` prefix for private functions

wosc:
@syohex Thanks for your help! I've updated the code (I must admit I'm not very familiar with the elisp conventions yet, sorry about that.)

purcell:
Thanks. Does this add anything new beyond the https://github.com/Wilfred/pyimport and https://github.com/emacs-pe/pyimpsort.el packages?

wosc:
@purcell Thanks for letting me know about these. I've checked them out, and as usual ;), all three do _different_ things:
* wilfred/pyimport adds import lines by looking them up from other open buffers, and can remove superfluous ones by utilising flycheck results. Sorts lines alphabetically with `sort-lines`.
* emacs-pe/pyimport does not add/remove lines, but sorts them in groups (stdlib, site-packages, local) using an external process.
* wosc/import-el adds import lines by creating their text by itself, sorts alphabetically with `sort-lines`. Does not remove lines.

So I guess I could ask @Wilfred if they are open to including another way of adding imports that does not rely on open-buffer lookup -- that would neatly fold my package into theirs.

tarsius:
@Wilfred @wosc any news on this?

Wilfred:
@wosc Great summary of the existing options! I've often noticed that developers who know how the alternatives work, end up writing better tools :)

I'd be open to a patch on pyimport if you think it's a natural addition. wosc-import is prompting interactively, whereas pyimport uses a heuristic to find the containing package name.

Perhaps we could fall back to prompting the user, wosc-import style, if we don't find an example in an existing buffer? What do you think @wosc?

wosc:
@Wilfred I was thinking along the same lines: heuristics first, prompt second (maybe have an option to disable either one or something like that; would need to try it to see how it plays out). So, perfect, I think we're on the same track here. Good to hear you're open for a patch, so I'll try to come up with one. @tarsius I think this means the Request

tarsius:
Okay, thanks you two!

PR #4274, opened 2016-09-26
Add recipe for refs.el

Wilfred:

refs.el is an Emacs package for finding references to functions, macros or variables. It works by parsing the elisp and analysing it.

refs_screenshot

https://github.com/Wilfred/refs.el

I'm the author.

### Checklist

### DIFF
--- a/recipes/elisp-refs
+++ b/recipes/elisp-refs
@@ -1,0 +1,3 @@
+(elisp-refs :repo "Wilfred/refs.el"
+            :fetcher github
+            :files (:defaults (:exclude "elisp-refs-bench.el")))
--- a/recipes/refs
+++ b/recipes/refs
@@ -1,0 +1,3 @@
+(refs :repo "Wilfred/refs.el"
+      :fetcher github
+      :files ("refs.el"))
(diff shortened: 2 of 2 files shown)


syohex:
- no `autoload` cookie
- I suppose `:files` attribute should be 
  - `:defaults` + `:exclude`
  - Or move benchmark file under `benchmark` and use default value(= not specify)

purcell:
Is it wise to use such a general name for a package which deals with only elisp refs? Why not `erefs`, for example?

Wilfred:
OK, I've moved to using `:defaults` in the recipe.

There are already autoload cookies, have I missed any places where I should have them @syohex?

Regarding the name, I like it because I think it reads well: `refs-function` -- "show me the code that references this function". I may also generalise it to other languages in future.

Wilfred:
@purcell OK, I agree. Looking at other packages, like `elisp-slime-nav`, it's worth clarifying.

I've rebranded the package as elisp-refs. Does everything look good now? :)

purcell:
Super, thanks! Merged. :-)

PR #4278, opened 2016-09-26
Add ansible-vault-mode recipe

zellio:

Provides the ability to edit and create ansible-vault files seamlessly without the user needing to directly invoke the `ansible-vault` command.

https://github.com/zellio/ansible-vault-mode

I am the primary author and maintainer. 

### Checklist

### DIFF
--- a/recipes/ansible-vault
+++ b/recipes/ansible-vault
@@ -1,0 +1,2 @@
+(ansible-vault :fetcher github
+               :repo "zellio/ansible-vault-mode")


syohex:
I have sent PR(https://github.com/zellio/ansible-vault-mode/pull/1)

- Keyword header lists `Keywords: org-mode, elisp, project` however is this package related to org-mode(and elisp, project) ?

zellio:
@syohex - I've merged your request, thank you for the suggestions.

Additionally I've updated the doc strings and keywords. The previously listed keywords were copied from one of my other projects which I had used as a template for this file and forgot to update.

syohex:
- I suppose minor-mode should define disabled case like [this](https://github.com/dgutov/diff-hl/blob/f8ce39d157144ee4b22d386dd40673ea7f87ecd7/diff-hl.el#L478-L514)
- Should any check be necessary before calling `shell-command-on-region` ? `shell-command-on-region` with `replace` argument replaces buffer content even if the command is failed.  If there is any validation command, I suppose buffer content should be validated with such tool. (Sorry I don't understand `ansible-vault` well. It may be no problem)

zellio:
`minor-mode` function has been improved as suggested.

No validation function is required for the buffer. `ansible-vault` is designed to operate on an entire file normally and the entire buffer is standing in as a proxy for "Current state of the file" which is what it is.

There should possibly  be some logic around ensuring the existence of the `ansible-vault` program and of the password file but that is currently out of scope.

purcell:
Thanks. Almost ready to merge this, but the package description says `Major mode for editing ansible vault files`, while there's no major mode defined here, so a better description is needed. :-)

zellio:
Purcell - Corrected in the latest release

purcell:
Great, thanks!

PR #4283, opened 2016-09-27
Add recipe for gobgen

gergelypolonkai:

Generate GObject boilerplate code based on input provided through a form.

https://github.com/gergelypolonkai/gobgen.el

Author and maintainer

None

### Checklist

### DIFF
--- a/recipes/gobgen
+++ b/recipes/gobgen
@@ -1,3 +1,2 @@
 (gobgen :fetcher github
-        :repo "gergelypolonkai/gobgen.el"
-        :files ("gobgen.el"))
+        :repo "gergelypolonkai/gobgen.el")
--- a/recipes/gobgen
+++ b/recipes/gobgen
@@ -1,0 +1,3 @@
+(gobgen :fetcher github
+        :repo "gergelypolonkai/gobgen.el"
+        :files ("gobgen.el"))


syohex:
- no autoload cookie
- There are too many byte-compile warnings
  - `lexical-binding: t;` is set, but this package depends on dynamic scope
  - Some functions are missing

gergelypolonkai:
* I have rebased to current master.
* The `lexical-binding:t` was a left-around from a faulty copy-paste, sorry about that.
* I fixed the compile warnings, and also made the variables buffer-local
* …and fixed that call to a missing function

recipes/gobgen
```diff
@@ -0,0 +1,3 @@
+(gobgen :fetcher github
+        :repo "gergelypolonkai/gobgen.el"
+        :files ("gobgen.el"))
```
syohex: Please don't specify `:files` attribute. In this case, it is enough to use default value(`:defaults`). Please read https://github.com/melpa/melpa#recipe-format .

syohex:
- Some macros(`defvar-local`) was introduced at Emacs 24.3, but this package does not specify minimum Emacs version(Example https://github.com/company-mode/company-mode/blob/d8ca520e99f40db1a0d2727d9b46c1db26b81f12/company.el#L10)
- `(eval-when-compile (require 'wid-edit))`, `eval-when-compile` is not necessary. Because your package uses wid-edit function. You can use `eval-when-compile` when you use only package macros.

gergelypolonkai:
I have updated both the recipe and the package as advised.

syohex:
https://github.com/gergelypolonkai/gobgen.el/pull/1

I have sent PR. Please check it.

gergelypolonkai:
Thank you, I have merged it.

purcell:
Thanks! Merging this now.

You might consider defining a major mode to simplify setup of the widgets window: you've got code in the `gobgen` function such as `(kill-all-local-variables)` and `(use-local-map widget-keymap)` which would be generated automatically by `define-derived-mode`. So you'd have code like

```el
(define-derived-mode gobgen-mode special-mode ...)
```
and then instead of `widget-keymap` you'd use the generated `gobgen-mode-map`.

PR #4284, opened 2016-09-28
Add env-var-import

ajsquared:

On OS X, a GUI Emacs instance will have a different environment from one started in the shell. This library works around that by copying environment variables from the shell.

https://github.com/ajsquared/env-var-import

I am the maintainer.

None

### Checklist

### DIFF
--- a/recipes/env-var-import
+++ b/recipes/env-var-import
@@ -1,0 +1,2 @@
+(env-var-import :repo "ajsquared/env-var-import"
+:fetcher github)


syohex:
What differences is this package from [exec-shell-from-path](https://github.com/purcell/exec-path-from-shell) ?

ajsquared:
I didn't know that existed! Thanks for the link @syohex, I don't think this is necessary anymore.

purcell:
Great minds think alike, eh? :-)

purcell:
P.S. So many edge cases - you definitely wouldn't have wanted to maintain this. :-)

PR #4286, opened 2016-09-28
Adds a recipe for the "node" package

cowboyd:
https://github.com/cowboyd/node.el

My experience working with Node JS in Emacs has been that it's not as
good as working with other languages. There's nothing terrible, but it's
just a bunch of paper cuts, so I decided to collect my various functions
and hacks into a single place that can be shared.

this package currently adds support for node stack traces for `(next-error)` and
`(previous-error)` navigation within compilation buffers. Planned
features also include:
- cleaning the crap out of compilation buffers when npm / mocha and
  friends don't pay attention to the TERM=dumb
- lookup local executables for running commands within `node_modules`
- less yak shaving for setting up the debugger.
### Checklist

### DIFF
--- a/recipes/node-tools
+++ b/recipes/node-tools
@@ -1,0 +1,3 @@
+(node-tools
+ :repo "cowboyd/node-tools.el"
+ :fetcher github)


syohex:
`node` is too generic name. I suppose it should be named like `node-compile`, `nodejs-compile` etc.

cowboyd:
I'm working more features than just compilation though, so `node-compile` etc... is too specific. `node-helpers` ? `node-utils`?

cowboyd:
@syohex I've renamed to `node-tools`. Let me know if there's anything else!

purcell:
Hey Charles! Small world. Quick feedback:
- Having renamed the package, any top-level identifiers should now be named `node-tools-*`
- The `initialize` function is pretty redundant: calling it would have the same effect as defining the compilation regexps at the top level and just `require`-ing the file, which I'd say would be preferable.

I'd honestly say there isn't enough here to publish it yet, though - maybe hold off until there are a few more bits? The plans (and name) suggest a grab-bag of config, which we've found doesn't easily translate to a widely re-usable package.

cowboyd:
Hey Steve!

Thanks for the feeback, and truth be told, I was hoping to meet you here :)

> The initialize function is pretty redundant: calling it would have the same effect as defining the compilation regexps at the top level and just require-ing the file, which I'd say would be preferable.

The reason I wrapped this in an initialize function was to make it zero-config. Can I just autoload those forms directly?

The reason I'm submitting this for publication now is twofold. The first is that I have wider plan for it, and for the things I want to build on top of it, and the second, is that I have a team of 8 people here that I want to share this with immediately rather than copy around a snippet that decays with each cut and paste.

The next step is to add to this package functions that resolve local node executables from within in a project. Our specific use-case for wanting to do this is  so that we can find the `eslint` executable in projects that have it. This will open up some key possibilities. The first is automagically enabling flycheck in eslint'd projects without needing to have eslint installed globally (this is problematic because it requires both a separate eslint installation and setup step, and also because eslint versions can and do vary from project to project). Another is to configure our other packages like `mocha-snippets` and other snippets to obey project-local eslint preferences durring snippet expansion. Of course, the eslint functionality would go into a different package.

Consuming this as a package from the beginning means that my team gets to have working compilation buffers asap, and then publishing and sharing the follow-on work is much easier.

I could break the node-tools into three separate packages?

* `node-compilation-regexen`
* `node-find-local-executable`
* `node-inspector-launcher`

But it does seems like overkill to publish one function per package.

purcell:
> The reason I wrapped this in an initialize function was to make it zero-config. Can I just autoload those forms directly?

Yes you can! The compilation regexp is kind of a special case, though, because in general it's not permitted for packages to auto-enable themselves: as with the very many packages in the emacs standard library, the ability to load a package should not be equated with the user's desire to actually use that package.

> The reason I'm submitting this for publication now is twofold. The first is that I have wider plan for it, and for the things I want to build on top of it, and the second, is that I have a team of 8 people here that I want to share this with immediately rather than copy around a snippet that decays with each cut and paste.

:+1:
 
> The next step is to add to this package functions that resolve local node executables from within in a project. Our specific use-case for wanting to do this is so that we can find the eslint executable in projects that have it. This will open up some key possibilities. The first is automagically enabling flycheck in eslint'd projects without needing to have eslint installed globally (this is problematic because it requires both a separate eslint installation and setup step, and also because eslint versions can and do vary from project to project). Another is to configure our other packages like mocha-snippets and other snippets to obey project-local eslint preferences durring snippet expansion. Of course, the eslint functionality would go into a different package.

Hmm, makes sense. Have you thought about providing a global/local minor mode pair in this package such that when the local minor mode (`node-tools-path-mode`?) is enabled, its hook would allow each function in that hook access to info about the calculated current node environment? This would allow you to easily piggy-back flycheck and mocha-snippets reconfiguration onto it by adding hook functions, without necessarily putting it all into the same package.

That minor mode could perhaps simply set the `exec-path` variable buffer-locally, and if you're lucky, subsequently-loaded minor modes like `flycheck` would then automatically find the correct executables. (Downside is that the PATH env-var wouldn't get set, so there'd potentially be edge-cases where you'd hope `shell-command` would do the right thing, but it wouldn't.)


> Consuming this as a package from the beginning means that my team gets to have working compilation buffers asap, and then publishing and sharing the follow-on work is much easier.
> 
> I could break the node-tools into three separate packages?
> 
>     node-compilation-regexen
>     node-find-local-executable
>     node-inspector-launcher
> 
> But it does seems like overkill to publish one function per package.

Nah, multiple packages does seem like overkill.

The main thing we'd want to avoid is that the package would be opinionated config useful only to your team -- we've seen some "packages" that amount to a my-way-or-the-highway collection of `(setq ...)` and `add-hook` calls. It doesn't sound like that will be the case, but a flexible implementation is key.

Happy to help with further advice.

cowboyd:
> Yes you can! The compilation regexp is kind of a special case, though, because in general it's not permitted for packages to auto-enable themselves: as with the very many packages in the emacs standard library, the ability to load a package should not be equated with the user's desire to actually use that package.

So does that mean it's ok in the case of compilation regexen?

> Hmm, makes sense. Have you thought about providing a global/local minor mode pair in this package such that when the local minor mode (node-tools-path-mode?) is enabled, its hook would allow each function in that hook access to info about the calculated current node environment? This would allow you to easily piggy-back flycheck and mocha-snippets reconfiguration onto it by adding hook functions, without necessarily putting it all into the same package.
>
>That minor mode could perhaps simply set the exec-path variable buffer-locally, and if you're lucky, subsequently-loaded minor modes like flycheck would then automatically find the correct executables. (Downside is that the PATH env-var wouldn't get set, so there'd potentially be edge-cases where you'd hope shell-command would do the right thing, but it wouldn't.)

I haven't thought about it until now, but I'm curious about how that would work. Is there a guide on how global/minor modes work? What would be the role of the global mode, and what would be the role of the minor mode?

One problem is that npm does not have a flat install hierarchy, so there is no one directory that I can point to that that contains the executables. Instead, if you want to run an executable called `eslint`, then I need some function to say "hey, I want to use 'eslint' where is it?"  Even so, I should be able to do this with a mode-hook. But the more I think about it, this sounds like an _npm_ problem, not a node problem, so perhaps the answer is to try out [npm-mode][1] and see if I can't accomplish what I want with it and try and extend it with what I need if necessary, and then have a totally separate package for the compilation error regex called `node-compilation-error-regex`

So in summary.

1. error regex goes into its own, focused package
1. I work with the `npm-mode` author if necessary to add the functionality I need.

Does that strike you as a good plan?

[1]: https://github.com/mojochao/npm-mode

purcell:
Yeah, good plan. You know, you could probably even justify throwing the compilation regexp into npm-mode.

As far as a function for finding the path of `eslint`, there's already `executable-find`: it looks for matching files in directories listed in the `exec-path` variable. That covers most process-starting functionality in Emacs, but not the cases where you run things via your user shell, e.g. `M-!` (`shell-command`).

Minor modes are the regular minor modes which are initialized on a per-buffer basis: you could define one which would modify a version of the `exec-path` variable made buffer-local (e.g. with `(set (make-local-variable 'exec-path) ...)`) according to the `default-directory` variable value in that buffer and any associated `node_modules` path, and then anything launched from that buffer would see that `exec-path` value. You'd use `define-minor-mode` for this.

Then you can make a globalized version of that minor mode, which basically turns that minor mode on in every buffer: see `define-globalized-minor-mode` for this. The same basic mechanism is used in a bunch of popular libraries, e.g. `projectile`.

`npm-mode` already has such a mode pair, so you maybe start of by experimenting with adding functions to `npm-mode-hook` which set `exec-path` locally as above, and see where/if that breaks down. :-)

tarsius:
@cowboyd How are things progressing?

cowboyd:
@tarsius I haven't revisited the above this year, so it's a bit stale in my brain, but I've been mollified by using the 

1. `add-node-modules-path`
2. the compilation error regexp contained in this PR
3. Always making sure I'm using yarn or npm version > 3.0 (which uses a flat install)

And that seems to catch 99% of cases including using the proper `eslint` executable and configuration.

I really ought to still publish the regexp, just not sure where to put it. I could put it in `npm-mode`, but that does feel like a bit of a hack, especially since there is a compelling alternative now in `yarn` `node-compilation` maybe?

tarsius:
> Is there any particular issue you're having?

I am trying to determine whether to close this pr, while not stepping on anyone's toes ;-)

I don't use javascript myself, so the things you mention don't mean much to me. But I do get the impression that you have essentially given up on the package this pr is about and have come to the conclusion that you can achieve the same with a bit of configuration, the main issue being how to distribute some regexp that is core to that approach to other users. If that is correct, then we should probably close this pr.

cowboyd:
Thanks for the consideration. I think it's fine to close, and I'll see if I can't chew on it a little more and open a separate PR if needs be.

tarsius:
Thanks! Happy chewing ;-)

PR #4288, opened 2016-09-28
Add org-clock-today recipe

mallt:

Minor mode to show the total clocked time with org of the current day in the mode line.

https://github.com/mallt/org-clock-today-mode

I am the maintainer.

**None**

### Checklist

### DIFF
--- a/recipes/org-clock-today
+++ b/recipes/org-clock-today
@@ -1,0 +1,2 @@
+(org-clock-today :fetcher github
+                 :repo "mallt/org-clock-today-mode")


syohex:
-  All global functions/variables should start with package prefix
- Please implement disabling case of minor-mode.`(org-clock-today-mode -1)`
- Is it necessary to re-define`org-clock-update-mode-line` ? Can `cl-letf` be used for it instead of re-defining ?

mallt:
Thanks a lot for the comments @syohex!

I've implemented your suggestions, except for the cl-letf which did not seem to have the desired effect. I use an advice now instead of the redefinition of the org-clock-update-mode-line function.

Thanks again!

purcell:
Thanks. Further notes from me:
- The package description `Minor mode to show the total clocked time of the current day in the mode line` can/should be shortened to `Show the total clocked time of the current day in the mode line`
- Parent group for the `defgroup` should probably be `org` or `org-clock` rather than `convenience`
- Using `advice-add` (from `nadvice.el`) requires that you declare a package dependency on at least Emacs 24.4, since that library was unavailable in older versions.
- The `provide` form for the package is missing.

mallt:
Thanks a lot @purcell! I've updated the repo accordingly.

purcell:
Super, thanks.

PR #4289, opened 2016-09-29
Add recipe for zhcosin/chinese-number

zhcosin:

Converting number format between Arabic and Chinese.

https://github.com/zhcosin/chinese-number

I am the author and maintainer, and some other has contribute for this package.

None

### Checklist

### DIFF
--- a/recipes/chinese-number
+++ b/recipes/chinese-number
@@ -1,0 +1,2 @@
+
+(chinese-number :repo "zhcosin/chinese-number" :fetcher github)


syohex:
I have sent PR(https://github.com/zhcosin/chinese-number/pull/2)

zhcosin:
I have merged it already.

syohex:
- no autoload cookies
- Public interfaces should use `-` separator, not `--`. And private functions/variables uses `--` separator instead of `-` separator

zhcosin:
Now I had add the autoload commentary at the public interface function, and adjust the separator in function name, thanks for your reminder.

syohex:
I have sent PR(https://github.com/zhcosin/chinese-number/pull/5) about autoload.
Is README updated about interface name ?

zhcosin:
Thanks for your carefulness, and I was very sorry for my carelessness, I have commit a modify fix this problem.

syohex:
Looks good.

purcell:
Thanks. I sent you a fix in https://github.com/zhcosin/chinese-number/pull/6.

You shouldn't really use the `cnum-` prefix here. It's okay if you want to use it internally, because I realise that some of the identifiers are very long, but your public functions at least should all have the `chinese-number-` name prefix.

zhcosin:
Thanks, I will review problems like this after the Chinese National Day.

zhcosin:
Now I fixed the prefix problem by modify almost all function name, use the `chinese-number--` as the prefix for all variables and functions, so some function name was very long.

purcell:
Super, thanks!

PR #4294, opened 2016-10-01
Add direnv to MELPA

jml:

The direnv package loads environment variables from direnv into the
file-local environment variables for a buffer. This is mostly useful
when the environment variables defined in direnv control compiler or
test behaviour.

https://github.com/jml/direnv-el

I am the author & maintainer of the direnv Emacs package, although not of [direnv](http://direnv.net/) itself.

**None**

### Checklist

### DIFF
--- a/recipes/direnv
+++ b/recipes/direnv
@@ -1,0 +1,1 @@
+(direnv :fetcher github :repo "jml/direnv-el")


syohex:
- no autoload cookie
- `process-environment` is not buffer local variable. Is it no problem about changing `process-environment`  globally ?

jml:
> no autoload cookie

I'm not aware of this—it didn't come up in any of the automated checks I ran. I discovered https://www.gnu.org/software/emacs/manual/html_node/elisp/Autoload.html. I'm guessing that I need to put `;;;###autoload` before the definition of `direnv-load-environment`? Why would I want to do that?

> `process-environment` is not buffer local variable. Is it no problem about changing process-environment globally ?

It's not a problem changing it globally, because there's not really much else you can do, and direnv only sets environment variables to values that you have explicitly approved. After my most recent changes, they'll get unset when you change out of buffer as well as when you open a new one.

Silex:
> I'm guessing that I need to put ;;;###autoload before the definition of direnv-load-environment? Why would I want to do that?

TL;DR autoloads cookie makes it so your methods can be called *without* loading the `.el` file first. The `.el` file is then loaded only when your method is called. The reason for this mechanism is speed, it allows for your config to load very fast and the full `.el` is loaded only when needed, yet you can bind keys to your functions.

jml:
Thanks. I've added a cookie in what I hope is the correct way.

purcell:
Thanks.

PR #4295, opened 2016-10-01
Add ts-comint package.

josteink:

ts-comint is a package which provides a TypeScript comint REPL to emacs.

https://github.com/josteink/ts-comint

I'm the author.

None

### Checklist

### DIFF
--- a/recipes/ts-comint
+++ b/recipes/ts-comint
@@ -1,0 +1,2 @@
+(ts-comint
+ :fetcher github :repo "josteink/ts-comint")


syohex:
- All global variable and function names should have package prefix, `ts-comint-` or `inferior-ts-` or `inf-ts-` etc, except `run-ts`(This naming are used in some packages, `run-python`, `run-octave`, etc)
- Don't assign `C-c l` binding for key binding convention(See [Key binding convention](https://www.gnu.org/software/emacs/manual/html_node/elisp/Key-Binding-Conventions.html#Key-Binding-Conventions))
- I suppose REPL buffer name should be `*Typescript*`, `*Inferior Typescript*` rather than`*ts*`.  `*ts*` is too short and it may be used other packages.

I have sent some PRs, please check them.
- https://github.com/josteink/ts-comint/pull/2
- https://github.com/josteink/ts-comint/pull/3

`js-comint` which you fork is a old package and it does not confirm Emacs packaging convention while this package is new, so this package should conform Emacs packaging convention.

purcell:
Thanks @josteink & @syohex. I agree with that feedback. Couple more small points:
- The `defgroup` is a child of itself. Change the `:group` to `'languages` or similar.
- The `(use-local-map ..)` is redundant: `define-derived-mode` generates that code for you.
- `M-x checkdoc` will help you fix up a couple of the docstrings.
- The code inside the `interactive` form of `run-ts` looks suspicious: shouldn't you also set `inferior-ts-program-arguments` etc when this function is called non-interactively? In general, using `setq` inside `interactive` forms to set the command arguments shouldn't be necessary, since the interactive form is there to do that for you.
- I feel like you should be passing the `local` argument to `add-to-list` when modifying `comint-preoutput-filter-functions`, since that var doesn't appear to be buffer-local by default

Hope that helps!

josteink:
Thanks for the pulls. Merged.

Also thanks for the references on key binding conventions. As far as I can see, I'm not in violation, as I'm merely suggesting bindings I'm user-defined hooks, not actually adding them myself.

As for the rest... They all seem like good suggestions. I'll see what I can get done.

josteink:
> `js-comint` which you fork is a old package and it does not confirm Emacs packaging convention while this package is new, so this package should conform Emacs packaging convention.

And here silly me thought forking something would save me considerable effort. Seriously though: I appreciate the tips and guidance provided.

I think I should have cleaned up in everything suggested so far. If you still have feedback about things you want improved, please let me know.

Otherwise I think it's getting ready for publishing a v0.0.1.

josteink:
Any news here?

syohex:
https://github.com/josteink/ts-comint/blob/f159407198e2f8042e7a3c1bda84faaca57a8d7b/ts-comint.el#L94-L95

`ts-prompt-regexp` is unused. Isn't it used for `comint-prompt-regexp` ? If it is unused, please remove.

josteink:
That seems correct. I've pushed a new commit to address this, and also replace nodejs-style module-loading with something which should work better for Typescript.

purcell:
Thanks!

josteink:
Awesome! Thanks :)

PR #4297, opened 2016-10-02
Add mosey.el

alphapapa:

Howdy. Why don’t ya mosey on in here. Have yourself a sit-down.

`mosey` makes it easy to mosey back and forth in your buffers. Just pass mosey a list of functions that move the point to certain places, and it’ll mosey the point between those places. Tell it `:backward` if you want to mosey on back, otherwise it’ll mosey on ahead. Tell it to `:cycle` if you want it to loop around when it gets to one end or the other.

To make it easier for ya, just pass a list of point-moving functions to defmosey, and it’ll cook up four functions: `mosey/forward`, `mosey/backward`, `mosey/forward-cycle`, and `mosey/backward-cycle`.

https://github.com/alphapapa/mosey.el

Chief cook and bottle-washer.

I try not to talk to myself too much...

### Checklist

### DIFF
--- a/recipes/mosey
+++ b/recipes/mosey
@@ -1,0 +1,1 @@
+(mosey :fetcher github :repo "alphapapa/mosey.el")


syohex:
- Please use `-` separator instead of `/` separator for Emacs Lisp coding convention

alphapapa:
> Please use - separator instead of / separator for Emacs Lisp coding convention

Hi syohex,

I understand that that's the historical convention, however there are now many packages which use ` /` to separate their "namespace", as it were, such as https://github.com/magnars/multiple-cursors.el, https://github.com/magnars/expand-region.el, https://github.com/Malabarba/smart-mode-line, http://github.com/k-talo/volatile-highlights.el, and even John Wiegley's own https://github.com/jwiegley/use-package, all of which are also on MELPA.  So I would prefer to use the slash here if that's okay.  :)

syohex:
Yes. It is difficult to fix such packages because many users uses them before registered MELPA. However your package is a new package so it is good to confirm Emacs lisp coding convention.

alphapapa:
Ok, I changed `mosey/` to `mosey-`.  Thanks for your help.

purcell:
Thanks. `with-eval-after-load` is discouraged except for in end user configuration files. What you probably want instead is `(eval-when-compile (require 'org))`.

alphapapa:
@purcell Thanks, Steve.  I had forgotten about `eval-when-compile`, but I'm not sure how to use it here.  The code in question modifies the function `mosey-forward` which doesn't yet exist at compile time, so it throws an error when compiled.  But I'm not sure why `mosey-forward` doesn't exist yet, since it should have been defined by the call to `defmosey`.  I guess I have a lot to learn about how the byte-compiling process works.

Anyway, I will probably just remove that whole part for now, because I think it needs some polishing before it works really well with Org.

Thanks as always for your help with smoothing the rough edges.  :)

PR #4300, opened 2016-10-05
Added recipe for ryo-modal

Kungsgeten:

`ryo-modal` provides a minor-mode and toolkit letting the user create their own modal key bindings. It has no default keybindings. It is similar to https://github.com/mrkkrp/modalka but provides more features.

https://github.com/Kungsgeten/ryo-modal

**None**

### Checklist

(Not 100% on the last part, but managed to build it with `make recipes/ryo-modal` and it works fine if I `eval-buffer` in `emacs -Q`.

### DIFF
--- a/recipes/ryo-modal
+++ b/recipes/ryo-modal
@@ -1,0 +1,1 @@
+(ryo-modal :repo "Kungsgeten/ryo-modal" :fetcher github)


syohex:
https://github.com/Kungsgeten/ryo-modal/blob/75234eef7e400a92b567744ab114f54d93aabaa1/ryo-modal.el#L175

This line is evaluated only by loading And there is no interface for disable it. `post-command-hook` makes emacs too slow. So I suppose it should be enabled at enable `ryo-modal-mode` and disabled at disable ` ryo-modal-mode`.

Kungsgeten:
I haven't noticed it being slow, but perhaps it is. The problem of enabling/disabling when running `ryo-modal-mode` is that the cursor color is bound by frame, and not buffer. This means that the cursor color will be the same color even if you switch to a buffer where `ryo-modal-mode` isn't activated. I tried searching for ways to set the cursor color locally, but couldn't find any.

As you can see in https://github.com/Kungsgeten/ryo-modal/blob/75234eef7e400a92b567744ab114f54d93aabaa1/ryo-modal.el#L161 the cursor shape is changed upon enabling/disabling the mode. I guess an alternative could be to have the cursor color switching disabled by default, and then have a function to add/remove the hook?

syohex:
Ah, I see. However I think enable and disable command should be added. Because that setting is evaluated only `(require 'ryo-modal)`.

Kungsgeten:
I've modified the code so that the hook is added/removed upon enabling/disabling `ryo-modal-mode`. The hook isn't added if `ryo-modal-cursor-color` is `nil`, so that's a way of disabling it.

1aafb5df8a94e5e2f50ebf4822a7d6c5bc6e7aa5

Kungsgeten:
@syohex I've made some more commits today (new features). Nothing to do with your latest comment though. Do you think my previous mentioned fix is enough, or do I need to do something more?

purcell:
1. Feels like changing the per-frame cursor using a per-buffer mode is always going to be problematic. Why not use a coloured modeline lighter instead?
2. Don't change the global value of `post-command-hook` inside a non-global minor mode. Instead, pass the `LOCAL` flag when calling `(add-hook 'post-command-hook #'ryo-modal--cursor-color-update)` and `remove-hook`.

Kungsgeten:
@purcell 1. Mainly because your eyes tend to be where the cursor is, not on the modeline. I feel like cursor color is a good way to indicate the mode (modal or insert) you're in. I could just use cursor shape, but personally I do not like that and I feel that most Emacs users have grown to like the box shaped cursor. A coloured lighter is a good idea though, I should probably add that as well.

2. If the purpose of the hook is to switch cursor color -- depending on if the buffer is in `ryo-modal-mode` or not -- the hook needs to be global. Otherwise the color won't change back to the normal color when switching to a buffer where `ryo-modal-mode` is inactive. What I really wan't is something like `switch-buffer-hook`, but I can't seem to find something like that in Emacs.

purcell:
Cool, thanks.

PR #4304, opened 2016-10-07
Added webpaste.el package

etu:

This mode allows to paste whole buffers or parts of buffers to pastebin-like services. It supports more than one service and will failover if one service fails. More services can easily be added over time and prefered services can easily be configured.

https://github.com/etu/webpaste.el

I built and maintain it.

**None**

### Checklist

### DIFF
--- a/recipes/webpaste
+++ b/recipes/webpaste
@@ -1,0 +1,1 @@
+(webpaste :repo "etu/webpaste.el" :fetcher github)


alphapapa:
Hi, a couple of suggestions:

1.  Does it really require Emacs 25.1?  A lot of people are still on Emacs 24.

2.  The keywords field uses specific keywords, described in the info page (`helm-info-emacs` makes it easy to find the right page):

> To make it easier to find packages related to a topic, most packages are associated with one or more "keywords" based on what they do.  Type `C-h p' (`finder-by-keyword') to bring up a list of package keywords, together with a description of what the keywords mean.  To view a list of packages for a given keyword, type <RET> on that line; this displays the list of packages in a Package Menu buffer (*note Package Menu::).

etu:
1) I set 25 due to creation to `save-mark-and-excursion` and changes in `save-excursion` but found a good way to make a macro for compability, so I've changed it to 24.

2) I changed keywords to: `convenience, comm, paste` and hope that's fine :)

purcell:
Thanks - looks useful! Couple of quick points:
- You should use `defgroup` to define a custom group to hold your various `defcustom`s.
- Don't define your own version of `save-mark-and-excursion` when missing. Instead, define it with a prefix: the [elisp conventions](https://www.gnu.org/software/emacs/manual/html_node/elisp/Coding-Conventions.html) discuss this.

alphapapa:
> I changed keywords to: convenience, comm, paste and hope that's fine :)

Well, I think the idea is that only the keywords in the built-in list should be used (i.e. `paste` isn't in the list), but I confess that I'm not an expert on this.  I guess it won't hurt anything, but I'm not sure it will have any effect either.  They're really more like categories than "keywords" (i.e. search-engine or HTML META keywords).  I'm not sure they're really useful anyway, so maybe it doesn't matter.  :)

purcell:
> Well, I think the idea is that only the keywords in the built-in list should be used (i.e. paste isn't in the list), but I confess that I'm not an expert on this.

Yep, that's the ideal, but the keywords in packages "out in the wild" are all over the place, so it's hard to be really picky. :-)

etu:
@purcell 
I've added defgroup also renamed the macro wrapping save-excursion.

@alphapapa 
Well I've seen others using that keyword as well, so it's not a "new" keyword it's just a non-standard one ;)

alphapapa:
> Yep, that's the ideal, but the keywords in packages "out in the wild" are all over the place, so it's hard to be really picky. :-)

Well, so much for trying to help reduce your workload reviewing PRs.  haha  ;)

Seriously though, would it be worth adding some code to flycheck-package to catch this?  Seems like it wouldn't be too hard to do.

purcell:
> Would it be worth adding some code to flycheck-package to catch this? Seems like it wouldn't be too hard to do.

Great idea!

purcell:
@alphapapa Are you volunteering to send me a pull request for `flycheck-package`? :-)

etu:
@purcell Thanks for merging! Now I just need to add more providers to make it more fault-resistent :)

alphapapa:
> Are you volunteering to send me a pull request for flycheck-package? :-)

@purcell I'm short on free time at the moment, but I'll see what I can do...

PR #4305, opened 2016-10-08
Add recipe for build-helper

afonso360:

```build-helper``` allows control over multiple build systems with one single interface.

Users can simply execute commands on a compile buffer and have their command history
recorded per project. Users can also link build-helper with other build systems and execute
elisp functions provided by other projects.

https://github.com/afonso360/build-helper/

**None**

### Checklist

### DIFF
--- a/recipes/build-helper
+++ b/recipes/build-helper
@@ -1,0 +1,1 @@
+(build-helper :repo "afonso360/build-helper" :fetcher github) 


syohex:
Should `build-helper-setup` be autoload function ?

afonso360:
Yes, it probably should be.

Fixed in the latest commit

syohex:
Looks good.

purcell:
Thanks. Quick feedback:
- `:group 'build-helper` means your `defgroup` is its own parent, which I don't think is what you want. ;-)
- Use `:group` in your `defcustom` forms
- `build-helper--functions` should be what users customize, IMO: they can then modify it using the custom interface, or via directory-local variables (which trivially give you per-major-mode configs). `build-helper-add-function` doesn't really add anything other than the requirement that users write elisp to customize the package, and your hand-rolled approach isn't idiomatic.

afonso360:
Based on your feedback:

Variables are now in their appropriate groups and the package is now grouped under programming tools.

Moved `build-helper--functions` and `build-helper--comint` to be public and gave them an appropriate customize interface.

I haven't removed `build-helper-add-function` because I still think its useful, but I agree that it shouldn't be the only way to add functions (this also applies to `comint` settings).

@purcell @syohex Thank you for taking the time to review my package.

purcell:
Thanks!

PR #4306, opened 2016-10-09
some refactoring in package-build.el

mr1yh1:


package-build/package-build.el
```diff
@@ -44,6 +44,7 @@
 (require 'package)
 (require 'lisp-mnt)
 (require 'json)
+(require 'subr-x)
```
dunn: This will mean adding `(emacs "24.4")` to the Package-Requires field above.

purcell:
Thanks, but package-build now lives [in its own repo](https://github.com/melpa/package-build), so please could you file changes in that location instead?

package-build/package-build.el
```diff
@@ -146,6 +147,10 @@ function for access to this function")
     (:exclude ".dir-locals.el" "test.el" "tests.el" "*-test.el" "*-tests.el"))
   "Default value for :files attribute in recipes.")
 
+(defun mappend (fn seq)
+  "Apply `FN' to the `SEQ', and append results."
+  (apply #'append (mapcar fn seq)))
+
```
purcell: `cl-mapcon` would do this too, I think. You definitely shouldn't introduce new unprefixed helper functions.

mr1yh1: yes i totally ignored that lack of namespace thing. 
it seems it's safe to use cl-mapcan there.

- what do you think about a separate doc-folder option in recipe spec ? 
(ECB requires customization after install, because it's searching info file under info-help.)

package-build/package-build.el
mr1yh1: ok, i will remove the function and change the regex original function was using.

package-build/package-build.el
Silex: Also, the commit message has a typo (Impleme**n**ting)

mr1yh1: maybe that was the way  cosmos telling me that my implementation is incomplete.

PR #4311, opened 2016-10-11
[add] receipe for multitran, initial import

zevlg:

Nifty Emacs interface to http://multitran.com

Supports multiple languages, including Esperanto, Latin and Luxembourgish.

https://github.com/zevlg/multitran.el

I'm author

None

### Checklist

Also checked for byte compilation

### DIFF
--- a/recipes/multitran
+++ b/recipes/multitran
@@ -1,0 +1,1 @@
+(multitran :fetcher github :repo "zevlg/multitran.el")


syohex:
- I suppose it is better to define major-mode `define-derived-mode` rather than `defun`
- https://github.com/zevlg/multitran.el/blob/e0954896d3b9a56a787eb740039a2fec1d9ee77a/multitran.el#L600 . `current-word` returns `nil`, then the package searches `nil` word. Is it intended behavior ?

syohex:
BTW please use `x` instead of `*` for check-box. This is GFM specification.

zevlg:
thanks, I've committed suggestions you made - https://github.com/zevlg/multitran.el/commit/bc174d031e9649793bc8759a928605d85fb69ff3

purcell:
Thanks. Sent you https://github.com/zevlg/multitran.el/pull/2.

However, you should also fix the face definitions: instead of defining new colour specs, you should inherit existing faces (e.g. `:inherit comment`, `:inherit link` etc.) so that the faces look good with all existing themes.

zevlg:
@purcell nice suggestion about faces definition, `shadow' and `link' quite suits

zevlg:
Pushed faces related changes - https://github.com/zevlg/multitran.el/commit/0a18683644b63aaf9a55a433c23ff4c83bceb915

purcell:
Nice, thanks.

PR #4312, opened 2016-10-11
Add recipe for hledger-mode.

narendraj9:

It's a major-mode for writing hledger-mode files. hledger is a Haskell ledger clone. 

https://github.com/narendraj9/hledger-mode.git

I am the creator and the maintainer of the package. 

**None**

### Checklist

### DIFF
--- a/recipes/hledger-mode
+++ b/recipes/hledger-mode
@@ -1,0 +1,1 @@
+(hledger-mode :fetcher github :repo "narendraj9/hledger-mode" :files ("*.el"))


syohex:
I have sent PR.

https://github.com/narendraj9/hledger-mode/pull/5

syohex:
- There are too many byte-compile warnings. Please check them.
- Please don't define function without package prefix(https://github.com/narendraj9/hledger-mode/blob/8d6e8a5e135d6721cb9a0058ddff99ecadd579b6/hledger-mode.el#L76-L82). Use anonymous function or function whose name starts with package prefix.

narendraj9:
Thanks. Will fix it.

narendraj9:
Fixed most of the byte-compile warnings. Added emacs version to Package-Requires.

syohex:
I have sent PR(https://github.com/narendraj9/hledger-mode/pull/6)

narendraj9:
Hey thanks! :-) This is my first major-mode. Now, hopefully, with your help, all issues are fixed. Is there anything else needed to add it to melpa?

purcell:
Thanks. Quick feedback from me:
- `(add-to-list 'auto-mode-alist '("\\.journal\\'" . hledger-mode))` is inappropriate, I think, because the file extension is not unique to `hledger`. Maybe suggest that the user adds this themselves.
- `company-hledger` and `ac-source-hledger-source` don't belong in this mode. To byte-compile cleanly, they would require those packages to be installed, and this should be optional. Instead, provide a `completion-at-point` function, which `company` will then automatically use. Similarly, `(setq-local ac-sources '(ac-source-hledger-source))` is not appropriate, because it can override user preferences.
- `font-lock-defaults` is always buffer-local, so `setq` is fine: `setq-local` is not required.
- `hledger-mode` should not be derived from `prog-mode`, I would say: use `fundamental-mode` as the parent.
- `(use-local-map hledger-view-mode-map)` is redundant: `define-derived-mode` will generate this form for you. Same with `(use-local-map hledger-mode-map)`.
- Use `M-x checkdoc` to help you tidy up the docstrings.
- The Mailgun stuff doesn't belong in a major mode. Perhaps make a separate package for mailgun support. Emacs has built-in email support, and you should use that by default, and make the delivery method configurable. (Also: you're using `async`, but there's no package dependency declare for that.)
- Why enable `hl-line-mode` in one place, then disable it elsewhere? Seems fragile. Leave this to users.
- `hledger-webservice.el` reads like personal code that would not be widely useful for other people. Or maybe the default URL should be blank, rather than your own server?
- In `hledger-helpers`, why use `cl`? You're using `cl-lib` elsewhere, and should avoid `cl` altogether, since it is deprecated. (The filename is not a useful indicator of what the assorted code in it is for.)

Hope that helps! :-)

narendraj9:
> (add-to-list 'auto-mode-alist '("\\.journal\\'" . hledger-mode)) is inappropriate, I think, because the file extension is not unique to hledger. Maybe suggest that the user adds this themselves.

Correct. Will fix it.

> company-hledger and ac-source-hledger-source don't belong in this mode. To byte-compile cleanly, they would require those packages to be installed, and this should be optional. Instead, provide a completion-at-point function, which company will then automatically use. Similarly, (setq-local ac-sources '(ac-source-hledger-source)) is not appropriate, because it can override user preferences.

This won't interfere with the user preferences because: 1. company-hledger is just a function definition 2. ac-sources becomes local to the buffer. 

But it makes sense to just define functions and sources and ask the user to add them to a mode hook. So will rename the `company-hledger` to something that starts with `hledger-`. Add instructions for users.

> The Mailgun stuff doesn't belong in a major mode. Perhaps make a separate package for mailgun support. Emacs has built-in email support, and you should use that by default, and make the delivery method configurable. (Also: you're using async, but there's no package dependency declared for that.)

I couldn't find support for multi-part form data directly in Emacs. That is why I had to write functions to prepare the multipart boundary and all that to send a post request of that kind. If there is built-in support for multipart/form-data, I would happily remove it.

> Why enable hl-line-mode in one place, then disable it elsewhere? Seems fragile. Leave this to users.

Makes sense.

> hledger-webservice.el reads like personal code that would not be widely useful for other people. Or maybe the default URL should be blank, rather than your own server?

I mention in the `README.md` that people are free to use the webapp hosted at [services.vicarie.in](services.vicaire.in). So, that url can be used by others as well. 

 
> font-lock-defaults is always buffer-local, so setq is fine: setq-local is not required.

> (use-local-map hledger-view-mode-map) is redundant: define-derived-mode will generate this form for you. Same with (use-local-map hledger-mode-map).

> hledger-mode should not be derived from prog-mode, I would say: use fundamental-mode as the parent.

> Use M-x checkdoc to help you tidy up the docstrings.


Thanks. Will fix these! :)

narendraj9:
@purcell Fixed! :)

purcell:
> This won't interfere with the user preferences because: 1. company-hledger is just a function definition 2. ac-sources becomes local to the buffer.
> 
> But it makes sense to just define functions and sources and ask the user to add them to a mode hook. So will rename the company-hledger to something that starts with hledger-. Add instructions for users.

You shouldn't have code specific to auto-complete or company here at all: to byte-compile cleanly, users would have to have both installed. Where add-ons have additional dependencies, you'd normally provide them as separate packages with the appropriate dependencies declared, e.g. a `company-hledger` package which depends on `hledger-mode` and `company`. But as I said, I'm pretty sure that you should simply define a completion-at-point function instead anyway, because that will just work with company. (Also, if in doubt, leave out the ac source.)

> I mention in the README.md that people are free to use the webapp hosted at services.vicarie.in. So, that url can be used by others as well.

Cool.

narendraj9:
For completions, I think I have fixed everything. This package
byte-compiles without company-mode or auto-complete. There are normal
function definitions that one can (if one chooses to) use as backend
functions for company.

Best,
Narendra Joshi

narendraj9:
This has been possible mainly because company-backend functions do not use any free variables.

purcell:
Super, thanks. Sorry for the delay.

narendraj9:
Thanks! :-)

PR #4313, opened 2016-10-12
Add recipe for godoctor

microamp:

Emacs frontend for [godoctor](http://gorefactor.org/), a refactoring tool for the Go programming language.

https://github.com/microamp/godoctor.el

None

### Checklist

### DIFF
--- a/recipes/godoctor
+++ b/recipes/godoctor
@@ -1,0 +1,2 @@
+(godoctor :fetcher github
+          :repo "microamp/godoctor.el")


syohex:
- I suppose not dry-run commands should do save-buffer before calling godoctor and do revert-buffer after it. How about ?

microamp:
How about simply making it complain when the current buffer is not saved? This behaviour is consistent with `go-rename` AFAIK. 

Something like
```emacs-lisp
(when (buffer-modified-p)
  (error "Please save the current buffer before invoking godoctor"))
```
?

syohex:
No problem.

microamp:
The change discussed above has been merged into `master`, https://github.com/microamp/godoctor.el/pull/3

Please have a look when you can. Thanks!

syohex:
Buffer is not updated after file changed by godoctor. All users don't enable `auto-revert-mode`. I thinks you need to add `revert-buffer`.

microamp:
Thanks for spotting it. Should've handled that. Here's the latest change. https://github.com/microamp/godoctor.el/pull/4

syohex:
Looks good.

purcell:
Thanks. Quick feedback:
- The parent group for your `defgroup` should not be `go` if that is not guaranteed to be defined, which it isn't in this case. Use `languages`.
- "Emacs" in the package description is redundant. All frontends implemented as Emacs packages are Emacs frontends!

microamp:
Thanks for your comments. The changes suggested above are now in `master`. https://github.com/microamp/godoctor.el/pull/5

> All frontends implemented as Emacs packages are Emacs frontends!

And you're indeed right! :)

purcell:
Super, thanks.

PR #4314, opened 2016-10-12
add the opener package to recipes

0robustus1:

The opener package allows evil users to open the resource
behind a http(s) URL in a buffer.

https://github.com/0robustus1/opener.el

**None**

### Checklist

### DIFF
--- a/recipes/opener
+++ b/recipes/opener
@@ -1,0 +1,4 @@
+(opener
+  :fetcher github
+  :repo "0robustus1/opener.el"
+  :files ("opener.el" "opener.texi"))


syohex:
- I suppose package name should be evil-opener because it must require evil as dependency
 - Or separate package
- This package sets some configuration and overwrite default evil configuration only by loading. I suppose it is better to provide some setup function or similar interface

0robustus1:
I was actually intending to make it less explicitly dependent on evil with time, therefore i'd like the 'evil'-less name.

Regarding the second point. That is quite a good one. The `gf` command actually gets rebound to opener, which will however fallback to the default action of evil (`find-file-at-point`: https://bitbucket.org/lyro/evil/src/f2648b841f9bcded8764ce63410065d2b8c5f856/evil-maps.el?at=default&fileviewer=file-view-default#evil-maps.el-73). But seeing as evil might change that default-behaviour, i can define a function that will activate the evil-mappings.

0robustus1:
Ok so i ultimately went with the separate package idea. This package is now evil-less, and is a dependency of the evil-opener package, which will provide a minor-mode to govern the remapping of keys. I created a PR for the evil-opener recipes here: #4317.

syohex:
> Regarding the second point. That is quite a good one. The gf command actually gets rebound to opener, which will however fallback to the default action of evil (find-file-at-point: https://bitbucket.org/lyro/evil/src/f2648b841f9bcded8764ce63410065d2b8c5f856/evil-maps.el?at=default&fileviewer=file-view-default#evil-maps.el-73). But seeing as evil might change that default-behaviour, i can define a function that will activate the evil-mappings.

I see

syohex:
https://github.com/0robustus1/opener.el/pull/1

I have sent PR.

syohex:
BTW Should `opener-url-browser-f` be renamed to `opener-url-browser-function` or `opener-url-browser-fun`?

0robustus1:
I guess `opener-url-browser-function` is fine. Thx for the PR :)

0robustus1:
I merged your pull-request and changed the name of the `opener-url-browser-f` variable to `opener-url-browser-function`.

purcell:
Quick feedback:
- `opener-open` should take `url-or-file` as an argument, and read it inside the `interactive` form.
- `lexical-let` is obsolete, and comes from the deprecated `cl` package. Instead, declare `lexical-binding: t` for the file, and use a regular lambda. Also add `(emacs "24")` to the `Package-Requires:` list if you do this.
- It's more conventional to use `_` in place of `dontcare`.
- `BANG` isn't a clear argument name.
- Instead of `(string-match-p (regexp-quote ".") file-segment)`, how about `(file-name-extension file-segment)`?

Hope that helps!

0robustus1:
@purcell thanks for the feedback, i really didn't expect to receive so much help here :)

* opener-open should take url-or-file as an argument, and read it inside the interactive form.
  * *done*
* lexical-let is obsolete, and comes from the deprecated cl package. Instead, declare lexical-binding: t for the file, and use a regular lambda. Also add (emacs "24") to the Package-Requires: list if you do this.
  * A little bit trickier, i still need the `(function* ...)`, as *request.el* needs it for the callbacks. I therefore required *cl-lib*, that isn't deprecated afaik, right?
* It's more conventional to use _ in place of dontcare.
  * Heard of both conventions, went with yours ;)
* BANG isn't a clear argument name.
  * good point, it's a remnant from when the definition was actually an evil-ex-command. Changed it to the more *speakable* `force-buffer`.
* Instead of (string-match-p (regexp-quote ".") file-segment), how about (file-name-extension file-segment)?
  * Originally i thought of it as a predicate function (meaning t/nil), however upon checking string-match-p actually return the index of the match, therefore `file-name-extension` seems to be a better fit. Changed it

purcell:
Nice!

> A little bit trickier, i still need the (function* ...), as request.el needs it for the callbacks. I therefore required cl-lib, that isn't deprecated afaik, right?

Yep, that's fine. You should add `(cl-lib "0.5")` to the package list, though.

Ping me when done and one of us will merge the recipe. :-)

0robustus1:
@purcell done

purcell:
Thanks.

PR #4317, opened 2016-10-13
Add recipe for evil-opener

0robustus1:

The opener package allows evil users to open the resource
behind a http(s) URL in a buffer. It utilizes the opener-package (see #4314) to
do this. This package provides commands and remappings in evil-mode for convenience purposes. It governs this with a minor-mode.

https://github.com/0robustus1/opener.el

**None**

### Checklist

### DIFF
--- a/recipes/evil-opener
+++ b/recipes/evil-opener
@@ -1,0 +1,4 @@
+(evil-opener
+  :fetcher github
+  :repo "0robustus1/opener.el"
+  :files ("evil-opener.el" "evil-opener.texi"))


syohex:
- Missing `opener` in dependencies(https://github.com/0robustus1/opener.el/blob/master/evil-opener.el#L5). (We should register opener package at first.)
- Recent packages does not use `turn-on-` or `turn-off-` interfaces. I suppose it is enough to use `M-x evil-opener-mode`. It toggles enabling mode.

0robustus1:
* Sure. I'll adjust that once we have the opener package registered.
* Oh ok? I was taking [evil-surround](https://melpa.org/#/evil-surround) as an example. Also if i understood the documentation of [define-globalized-minor-mode](https://www.gnu.org/software/emacs/manual/html_node/elisp/Defining-Minor-Modes.html#index-define_002dglobalized_002dminor_002dmode) correctly, the second argument has to be a turn-on function. If i just supply the `evil-opener-mode` function there, it would be a turn-on/turn-off function which doesn't seem match the required semantics. Could you provide me with an example of a package using it without turn-on/turn-off functions defined?

purcell:
> If i just supply the evil-opener-mode function there, it would be a turn-on/turn-off function which doesn't seem match the required semantics. Could you provide me with an example of a package using it without turn-on/turn-off functions defined?

My `page-break-lines` package does that. Point is that the `turn-on` and `turn-off` functions shouldn't be part of the public interface to the package: the mode function itself should be the canonical way to toggle the mode. For `define-globalized-minor-mode` you can inline a lambda, e.g. `(lambda () (evil-opener-mode 1))`.

0robustus1:
I went with the lambda-option you suggested, but it doesn't seem that your `page-break-lines` package is using that pattern: it still contains the autoloaded [turn-on/off functions](https://github.com/purcell/page-break-lines/blob/master/page-break-lines.el#L101-L109).

purcell:
> but it doesn't seem that your page-break-lines package is using that pattern: it still contains the autoloaded turn-on/off functions.

Ugh, right - gotta scrap those!

0robustus1:
Now that *opener* is merged and built i adjusted the package-requires in *evil-opener* to include opener.

purcell:
Thanks. I'm going to merge this, but there's a further change needed on your side: it's bad practice to depend on snapshot versions of other packages. So tag your repo, e.g. 0.1, and then depend on `(opener "0.1")`. Similarly, your `evil` dependency should be on a version like `1.2.12` -- see MELPA Stable, on which your package will appear once you've added a version tag. Avoiding snapshot version dependencies allows packages in MELPA Stable to be installed without the user also needing to enable MELPA.

0robustus1:
Sure i will tag both correctly and make the adjustment.

0robustus1:
And thank you for the input and reviewing this :)

0robustus1:
Just to document this: I added the tag and changed the package-requires for both opener and evil-opener accordingly (meaning to require actual versions instead of snapshots).

PR #4318, opened 2016-10-13
add recipe wolfram.el

stardiviner:

Wolfram Alpha integration in Emacs

Allows you to query Wolfram Alpha from within Emacs.

https://github.com/hsjunnesson/wolfram.el

Like the package.

**None**

### Checklist

### DIFF
--- a/recipes/wolfram
+++ b/recipes/wolfram
@@ -1,0 +1,2 @@
+(wolfram :fetcher github
+         :repo "hsjunnesson/wolfram.el")


syohex:
- Wrong `interactive` usage
- Use undefined function(`wolfram--xml-for-query`)
- No autoload cookie
- `defgroup` should set parent group attribute

purcell:
> Wrong interactive usage

I think that's okay, but `use-region-p` is the correct way to decide whether to use the region, not checking `transient-mark-mode` and `region-active`.

syohex:
https://github.com/hsjunnesson/wolfram.el/blob/6d83f4af77cb722aa9ba0cbb5187df4ae4728f08/wolfram.el#L144-L149

> Wrong interactive usage

My explanation was insufficient, I meant `interactive` position is wrong. It must be after docstring or arguments.

purcell:
> It must be after docstring or arguments.

Ah yes - I missed that! :-)

stardiviner:
I have updated my PR. fixed all.

stardiviner:
Ok, upstream merged and updated.

syohex:
- Missing second argument of `wolfram--async-xml-for-query` https://github.com/hsjunnesson/wolfram.el/blob/bab68aabc0066b4b4474a02551c9b20b0b7209be/wolfram.el#L62

stardiviner:
It does not exist in latest `master` version.

stardiviner:
I checked out code, found that function `wolfram--pods-for-query` is not used in program.
So temporary this `callback` argument issue can be ignored. I will post an issue on `wolfram.el`. Hope author can fix it.
So this recipe can be merged for now.

purcell:
Thanks.

Issue #4310, opened 2016-10-10
Preparing to migrate `puml-mode` and `plantuml-mode`


skuro:
Hey there,

as part of an active effort to merge efforts in supporting PlantUML mode in Emacs, [it was agreed](https://github.com/zwz/plantuml-mode/issues/8) to have one single `plantuml-mode` major mode, which will use the current `puml-mode` repository.

The merge will be finalized at the end of the week, when I will proceed with the following actions:

- merge the renamed mode into `master`
- rename my GitHub repository (`skuro/puml-mode` -> `skuro/plantuml-mode`)
- open a PR to melpa to delete [`puml-mode`](https://github.com/melpa/melpa/blob/master/recipes/puml-mode) and update [`plantuml-mode`](https://github.com/melpa/melpa/blob/master/recipes/plantuml-mode) to point it to my repository

I'd like to know from MELPA if I'm missing anything here, if there's any extra detail that you guys will need to approve the move and, ultimately, if I'm good to go with the merge.

Thanks

syohex:
I suppose you should check [spacemacs](https://github.com/syl20bnr/spacemacs/) too. Because many users use spacemacs and it uses `puml-mode` now. (This is not related to MELPA, sorry).

skuro:
@syohex good point, I'll take that offline though as it's not 100% MELPA related. Thanks!

skuro:
Ok, given the silence I gather that I didn't raise any eyebrow from MELPA. Closing for now.

PR #4320, opened 2016-10-14
Preparing to the merge with `plantuml-mode`

skuro:

This PR is to update the location of the existing package `puml-mode`. The repository got renamed to `plantuml-mode` and the old code is now in a dedicated `puml-mode`.

See #4310 for more details.

https://github.com/skuro/plantuml-mode/tree/puml-mode

I'm the maintainer of the package

None

### Checklist

### DIFF
--- a/recipes/puml-mode
+++ b/recipes/puml-mode
@@ -1,1 +1,1 @@
-(puml-mode :fetcher github :repo "skuro/puml-mode")
+(puml-mode :fetcher github :repo "skuro/plantuml-mode" :branch "puml-mode")


skuro:
On a second thought, I don't think MELPA supports having multiple packages in the same repository. We're all better off with a single PR where `puml-mode` is removed and `plantuml-mode` is switched to its new place.

purcell:
> I don't think MELPA supports having multiple packages in the same repository. We're all better off with a single PR where puml-mode is removed and plantuml-mode is switched to its new place.

We do support multiple packages in the same repo, but all the same, let's just do this the straightforward way in a single PR.

skuro:
Okido. But then one question before we go: how would I tell MELPA stable which tags belong to which repository?

purcell:
A tag applies to the whole repository, so you probably don't want to get into that situation. However, we support a `:version-regexp` option for extracting stable version numbers from tags, so you could potentially have version tags like `packageA-1.0` and `packageB-2.0`. Please don't do that though. :-)

skuro:
Gotcha. I'll just keep two separate repos.

PR #4322, opened 2016-10-15
Add eshell-up to MELPA

peterwvj:

package for quickly navigating to a specific parent directory in `eshell` without having to repeatedly typing `cd ..`.

https://github.com/peterwvj/eshell-up

I am the maintainer of the package.

**None needed** 

### Checklist

### DIFF
--- a/recipes/eshell-up
+++ b/recipes/eshell-up
@@ -1,0 +1,1 @@
+(eshell-up :fetcher github :repo "peterwvj/eshell-up")


purcell:
Thanks. Quick feedback:
- `go-up` isn't a great name because it's so general. `eshell-up` might be better.
- All your functions must have a package-specific name prefix, so `find-sub-str` should be `go-up--find-sub-str` etc. See the [elisp conventions](https://www.gnu.org/s/emacs/manual/html_node/elisp/Coding-Conventions.html) for more on this.
- `find-parent-dir` looks bogus to me: there are lots of built-in functions for handling filenames and directories, and you certainly shouldn't need to write code dealing with platform-specific filename separators. I feel like this package should need about half a dozen lines of code in total.

Hope that helps!

peterwvj:
First, thank you for your feedback - it really helped improve the code.

I have addressed all the issues you raised, and the code is now much more concise I think. Note that I moved the package to a new repository called `eshell-/up`.

Can we progress the PR?

purcell:
Looks great - thanks! Merging now.

PR #4326, opened 2016-10-16
Add recipe for pygen

JackCrawley:

Python code generation with Elpy and Python-Mode.

It includes a number of commands to automatically generate and manipulate python code.

https://github.com/JackCrawley/pygen

Author & maintainer

**None needed**
### Checklist

### DIFF
--- a/recipes/pygen
+++ b/recipes/pygen
@@ -1,0 +1,2 @@
+(pygen :repo "JackCrawley/pygen"
+       :fetcher github)


syohex:
https://github.com/JackCrawley/pygen/pull/1

I have sent PR about package format. Please check whether package format is valid before sending PR.

syohex:
- No autoload cookie
- Should `pygen-minor-mode` be renamed to `pygen-mode` ?
 - keymap is `pygen-mode-map` while hook is  `pygen-minor-mode-hook`. I think this is confused
- Emacs in terminal(`emacs -nw`) cannot use `C-=` key

purcell:
I agree with the above feedback, and that pull request will need to be merged before we can do anything here. Additionally:
- Don't use `(when (not ...) ...)` everywhere - there's `unless` for that!
- Code like [this](https://github.com/JackCrawley/pygen/blob/fac14c704cae060191f8c96ec5365e995c3c9bc3/pygen.el#L323-L333) is very clunky. Whenever you're doing lots of `setq` of local variables, you should be using `let` properly instead, and chaining expressions instead of mutating vars. Throughout the file you're using `let` simply to declare uninitialised local vars, making the code more imperative, verbose & fragile.

JackCrawley:
Thanks for the feedback guys. I'll apply these suggestions and resubmit. Should I close this request and open a new one once things are fixed?

alphapapa:
@JackCrawley No, just update this one when you're ready.  :)

JackCrawley:
Hey guys, finally had some time to implement the suggestions you gave me. With any luck it should adhere to the specifications now. The only suggestion I haven't implemented is using `(match-string)` on [this](https://github.com/JackCrawley/pygen/blob/fac14c704cae060191f8c96ec5365e995c3c9bc3/pygen.el#L324) function. I find it easier to read when I search for the correct pattern separately to the setting of the substring, but I can rewrite it if you think it's better. 

I've never used autoload cookies before. As far as I can tell, it's appropriate to put one in front of the `define-major-mode` macro, as the function `pygen-mode` is the entry point to the package. Is this correct?

alphapapa:
Well it's up to you, of course, but here's an example you might want to consider:

Replace this:

```el
(while (re-search-forward "[^ \n\t\\\\].*," nil t)
              (let (current-argument)
                (setq current-argument (buffer-substring-no-properties previous-position (1- (point))))
                (setq current-argument (pygen-parse-single-argument current-argument))
                (when current-argument
                  (push current-argument arguments)
                  (setq previous-position (point)))))
```
with this:
```el
;; Outside the defun
(defconst comma-separated-args-re (rx (not (any space "\\"))
                                  (group (minimal-match (1+ anything)))
                                  ","))
...
;; Inside the defun
  (while (re-search-forward comma-separated-args-re nil t)
    (-when-let (arg (pygen-parse-single-argument (match-string-no-properties 1)))
      (push arg arguments)))

```

purcell:
> As far as I can tell, it's appropriate to put one in front of the define-major-mode macro, as the function pygen-mode is the entry point to the package. This is what I've done at the moment. Is this correct?

Yep.

Something else I noticed is that you're using `in-string-p`, which only appeared in Emacs 25.1. You'd need to either depend on that version (not recommended) or make your own `pygen-in-string-p` function, which could simply use the expression `(elt (syntax-ppss) 3)`.

JackCrawley:
@alphapapa 
`dash.el` is very cool. I knew of it but I didn't realise how useful it was! Same with `rx`. Very useful! 

Your suggestion is a lot clearer. I uncovered an error when I was rewriting it so the structure is a little different but I've rewritten it along those lines. Thanks! :)

@purcell 
Thanks for the heads up. I've extracted it into a function that calls `in-string-p` if it's bound, otherwise it falls back to `(elt (syntax-ppss) 3)`. I figure `in-string-p` may be more efficient than using `syntax-ppss`, or it may become that way in the future. Is this bad form? Should I functions like this for the sake of predictability?

purcell:
I just looked, and `in-string-p` is defined like this:

```el
(defun in-string-p ()
  "Return non-nil if point is in a string."
  (declare (obsolete "use (nth 3 (syntax-ppss)) instead." "25.1"))
  (let ((orig (point)))
    (save-excursion
      (beginning-of-defun)
      (nth 3 (parse-partial-sexp (point) orig)))))
```

so it's actually marked as obsolete, and you should apparently use exactly the snippet I suggested. :-)

JackCrawley:
Awesome, I've updated it to use the recommended snippet instead.

purcell:
Thanks.

PR #4327, opened 2016-10-17
add recipe for add-node-modules-path

codesuki:
Provides a function that searches for and adds the node_modules/.bin directory to the buffer exec_path. E.g. support project local eslint (and other tools) installations.

https://github.com/codesuki/add-node-modules-path

Maintainer

**None needed**

### Checklist

### DIFF
--- a/recipes/add-node-modules-path
+++ b/recipes/add-node-modules-path
@@ -1,0 +1,1 @@
+(add-node-modules-path :repo "codesuki/add-node-modules-path" :fetcher github)


purcell:
Hah! This does what I was suggesting to @cowboyd after a discussion related to #4286. If this is working for you, @codesuki, I'm thinking the best approach is probably to merge this recipe, and then @cowboyd can package up (or submit upstream) any other node-related magic he comes up with.

codesuki:
Ah :) 
Cool it seems to help! I made it a while back and thought it could be helpful for other people using flycheck etc with local node modules. 
I am using it so don't have to pollute my global namespace just to be able to use the packages from emacs. So far no problems.

cowboyd:
@codesuki This looks fantastic!

One problem I do have is that when using `npm@2`, the `eslint` executable could be buried as a sub-dependency and will not be inside `node_modules/.bin`. Have you encountered this problem? And if so, how have you dealt with it.

codesuki:
@cowboyd sorry no idea! 
I am always using the latest npm version so this never happened to me.

codesuki:
Do I need to change something?

syohex:
Is it enough to set only `exec-path` ? `call-process` or `start-process` works however `shell-command`(`shell-command-to-string`) does not work. It is necessary to set environment variable `PATH` for them.

purcell:
> Is it enough to set only exec-path ?

I discussed that in #4286 - it's correct that this isn't a 100% perfect fix, but it seems to be working for the intended cases, so I'll go ahead and merge.

PR #4328, opened 2016-10-17
add js-import recipe

jakoblind:

Automatically import JavaScript files from the current project or dependencies

https://github.com/jakoblind/js-import

I am creator and maintainer

**None needed**

### Checklist

### DIFF
--- a/recipes/js-import
+++ b/recipes/js-import
@@ -1,0 +1,1 @@
+(js-import :fetcher github :repo "jakoblind/js-import")


syohex:
- What is different from `https://github.com/galooshi/emacs-import-js` ?
- No `Package-Requires` header
 - This package requires `f`, `projectile`, `dash`
 - This package requires Emacs 24.4 or higher version because it uses `subr-x`
- Use `js-import-` prefix instead of `jsi/`.
 - Use `-` separator instead of `/` for emacs lisp coding convention
 - `js-import--` should be used for private functions/global variables
- No autoload cookie

jakoblind:
I have updated the code based on your comments.

The difference between this package and `https://github.com/galooshi/emacs-import-js` is that this package doesn't have dependency on any NPM applications.

purcell:
Thanks. A little more feedback:
- Please depend on stable versions of packages, not snapshot versions. You can generally find the latest stable version by looking on stable.melpa.org.
- Docstring for `js-import` is missing. (`M-x checkdoc` for the win!)
- If you're always assuming a `package.json`, then you almost certainly don't need to depend on and use `projectile` - just use `locate-dominating-file`. :-)

jakoblind:
I have updated the code. 

I also use projectile for getting all files in the project, so I still would need the dependency even if I would use `locate-dominating-file`.

purcell:
Cool, thanks. Merged!

Issue #4330, opened 2016-10-17
go-guru source has moved


mdorman:
Per https://go.googlesource.com/tools/+/69f6f5b782e1f090edb33f68be67d96673a8059e

purcell:
Fixed by #4331.

PR #4331, opened 2016-10-17
Update go-guru and go-mode recipes

syohex:
### Checklist

It is hosted at dominikh/go-mode. So go-mode recipe need to update too.
This is related to #4330. CC: @mdorman


purcell:
Super, thanks.

PR #4333, opened 2016-10-18
Add kdeconnect recipe

carldotac:

This package provides helper functions to use [KDE Connect](https://community.kde.org/KDEConnect) functionality, like sending files to Android devices or making your phone ring, from within Emacs.

This is my first Emacs package. While I have tested its build and looked over past pull requests for common errors, I'm still pretty green with Emacs Lisp. So if I should change anything just let me know!

https://github.com/carldotac/kdeconnect.el

**None needed**

### Checklist

### DIFF
--- a/recipes/kdeconnect
+++ b/recipes/kdeconnect
@@ -1,0 +1,1 @@
+(kdeconnect :fetcher github :repo "carldotac/kdeconnect.el")


syohex:
https://github.com/carldotac/kdeconnect.el/blob/a3ffa9f3033cbc4feb63b0a547cf983fac060db8/kdeconnect.el#L88-L89

Does this error message depends on locale ? If it depends, it should use exit code instead of message content.

syohex:
https://github.com/albertvaka/kdeconnect-kde

It looks error messages depend on locale. For example, Japanese environment users may get `デバイスは見つかりませんでした` instead of `No devices found`.

carldotac:
Thank you for that! Unfortunately, returning no devices isn't an error from kdeconnect-cli's perspective, so I can't use stderr.

To make kdeconnect.el locale-agnostic, I've added a variable where the user can place their desired devices instead of parsing shell output. This also makes switching between devices much quicker.

syohex:
Thanks for updating code. BTW Environment variable can be replaced temporary as below.

```lisp
(let ((process-environment (cons "LANG=C" process-environment)))
  (shell-command-to-string "kdeconnect-cli -a --id-only"))
```

purcell:
Thanks - will merge in a moment, but just a couple of notes:

Firstly, `kdeconnect-active-device` should probably be quoted using `shell-quote-argument` before inserting it into command strings, just to be safe.

Secondly,
```el
(defun kdeconnect-select-active-device ()
  "Choose the active device from all available ones."
  (interactive)
  (setq kdeconnect-active-device
        (completing-read
         "Select a device: "
         (split-string kdeconnect-devices "," t)
nil t "")))
```
would be a little nicer/more idiomatic written like this IMO:
```el
(defun kdeconnect-select-active-device (name)
  "Set the active device."
  (interactive
   (list (completing-read
          "Select a device: "
          (split-string kdeconnect-devices "," t)
          nil t ""))))
  (setq kdeconnect-active-device name)
```
then you can also use it as a regular function.

PR #4341, opened 2016-10-20
Add recipe for dr-racket-like-unicode

david-christiansen:

This package implements the Unicode input interface from DrRacket, the editor shipped with Racket. In this interface, one writes a TeX-style shortcut, then presses a key to transform the TeX-code immediately to the left of point into its Unicode equivalent.

This is nice when only using a few characters not on one's keyboard, as it doesn't get in the way as much as a separate input method.

https://github.com/david-christiansen/dr-racket-like-unicode

I am the author.

None needed.

### Checklist

This is a package for performing unicode input in the style of DrRacket,
which can be convenient when only occaisionally using characters not on
one's keyboard.

### DIFF
--- a/recipes/dr-racket-like-unicode
+++ b/recipes/dr-racket-like-unicode
@@ -1,0 +1,3 @@
+(dr-racket-like-unicode
+ :fetcher github
+ :repo "david-christiansen/dr-racket-like-unicode")


syohex:
- No autoload cookie
- Regular expression is wrong some characters(https://github.com/david-christiansen/dr-racket-like-unicode/pull/1)

david-christiansen:
Right. The cookie is added and the PR merged. Thanks!

syohex:
BTW what differences are this package from https://github.com/apg/unipoint ?

PR #4343, opened 2016-10-21
indent-tools: indent, move around indentation units (perfect for yaml…

vindarel:

The package allows to indent, navigate and act on indentation units:
- indent the current indented block, or all the blocks with the same indentation
- move around, for example in a big yaml file: go to the next block of the same indentation, go to its parent, child, etc
- other related actions: fold, kill, copy, comment, all presented in nice hydras.

Thought for yaml, python, jade.

https://gitlab.com/emacs-stuff/indent-tools

Author and maintainer.

**None needed**
### Checklist
- [X] I've read [CONTRIBUTING.md](https://github.com/melpa/melpa/blob/master/CONTRIBUTING.md) => yes, but it isn't linked in melpa's readme anymore, too bad imo.

Thanks for the review and the merge !

### DIFF
--- a/recipes/indent-tools
+++ b/recipes/indent-tools
@@ -1,0 +1,1 @@
+(indent-tools :fetcher gitlab :repo "emacs-stuff/indent-tools")


purcell:
Thanks. I'm not really convinced this is ready for distribution yet:
- There's a bunch of special-case code for a handful of modes, e.g. `indent-tools--indentation-offset` and that code is not extensible.
- It's not advisable for a general-purpose library like this to assume that all users will want `hydra`: you should use the standard minor mode and keymap facilities instead, then you can add a hydra for personal use if it's helpful to you.
- Not only is `indent-tools--last-beg` unused, but it seems like a hack. That var will be shared across buffers (because it's not declared buffer-local), which is obviously not ideal. What you really want is to just select the region then call `comment-dwim` instead of `comment-region`.

alphapapa:
Just curious, did you implement the folding yourself?  There's already a package that handles folding very well, `origami-mode`.  It would probably be worth using its functionality.

vindarel:
Hi, thanks for the review.

- indeed, that is my [issue n°1](https://gitlab.com/emacs-stuff/indent-tools/issues/1): how to get the indentation of the current mode, generically ? For notice, the code is a little extensible: we can change the default indentation offset that will be used if we don't know the current global mode (`defvar indent-tools-indent-offset`). I reckon that should be more generic. After your review I began to make an alist of modes -> variables to get the indentation, but finally I don't think it's a progress. Progress would be users asking to support another mode and that support added in the core.
- good point, I created a minor mode and the hydra isn't bound by default ([commit](https://gitlab.com/emacs-stuff/indent-tools/commit/b077975c90698afa6995f78f1d6f55116aad34a4)).
- indeed, I removed that.

vindarel:
@alphapapa I used yafolding https://github.com/zenozeng/yafolding.el (just because I knew it and it's based on indentation)

vindarel:
hey, lil' up ?

alphapapa:
> how to get the indentation of the current mode, generically ?

Does the `standard-indent` variable change by mode?  I'm not sure.

Anyway, you might be able to determine it yourself by opening a new temporary buffer, setting the major mode, calling the appropriate indentation function, and measuring how much it indented.

purcell:
I feel like the autoloads will cause `hydra` to be loaded at Emacs startup time even for people who have no interest in having or using `hydra`, which really isn't optimal.

Using `standard-indent` as suggested by @alphapapa is a great idea. Ideally you'd have an `alist` mapping major modes to functions which can return the indentation for that major mode, then everything would be nicely extensible.

vindarel:
Nice feedback both of you, thanks, I'll try something !

vindarel:
Hi,
So I wrote the alist: https://gitlab.com/emacs-stuff/indent-tools/commit/ea9d6833562cbeb22914f4120e3c33a17bff4d41 (I renamed them to `indent-tools-indentation-of-xxx`)
and I use `standard-indent` instead of my own variable (https://gitlab.com/emacs-stuff/indent-tools/commit/30c71dcc7ffc25845a4442d2815275155b3a67ff)

---
I didn't manage to determine the current indentation level of any given mode myself, I was trapped in a sort of recursive problem: once I insert a string in a temp buffer with the major-mode activated, I don't know of any command to indent the line accordingly:`indent-according-to-mode` doesn't indent when not necessary, `indent-rigidly` needs an indentation target, which is calculated by my code.

alphapapa:
> I don't know of any command to indent the line accordingly:indent-according-to-mode doesn't indent when not necessary

I guess you could do this:

1.  Have a snippet of text for each mode, containing code in the appropriate language, that is not indented properly, with a marker for the cursor position.
2.  Insert the snippet into the temp buffer and position the cursor.
3.  Call the indentation function.
4.  Measure the resulting indentation.

vindarel:
> Have a snippet of text for each mode, containing code in the appropriate language

that is "impossible" :) this last function we are discussing would be to mesure the indentation *of a language we didn't take into account yet*. Thus I can't insert code of the appropriate language. When I know the language, I can find the indentation in some variables, like I did in the alist with python, jade and yaml.

Do we understand each other ?

So, you asked:
- to have an alist of modes -> function that returns the indentation: OK
- use the default variable `standard-indent`: OK
- you had the idea to determine the indentation ourselves: I leave it to the extensible alist (in which a function could use the method you described, if necessary).

purcell:
Merged now, thanks.

vindarel:
Awesome :)
Thanks again for the helpful reviews.

PR #4344, opened 2016-10-22
Add recipe for smmry

microamp:

A simple client that talks to [SMMRY](http://smmry.com/) [API](http://smmry.com/api)

https://github.com/microamp/smmry.el

None needed

### Checklist

### DIFF
--- a/recipes/smmry
+++ b/recipes/smmry
@@ -1,0 +1,2 @@
+(smmry :fetcher github
+       :repo "microamp/smmry.el")


syohex:
- https://github.com/microamp/smmry.el/blob/1db45082eea34efbab7120e3b03ca060068f3504/smmry.el#L125 code is always evaluated and this command always failed
- `string-trim` is defined at subr-x.el but this packaged does not load it. And subr-x is introduced at Emacs 24.4 but this packages does not require minimum Emacs version
- There are some byte-compile warnings about using free variables

microamp:
* Oops. It must have been a last minute `magit-discard` action I did. Things got out of order. Thanks for spotting.
* I removed the `subr-x` dependency after verifying that it wasn't actually needed. The package still requires `json` though which was introduced in Emacs 23.1. So I added `Package-Requires: ((emacs "23.1"))`.
* Byte-compile warnings should not be there anymore.

Thanks for reviewing again. Much appreciated.

microamp:
Here is the PR addressing the issues reported above: https://github.com/microamp/smmry.el/pull/2

purcell:
Declaring a dependency on `((emacs "23.1"))` is actually counterproductive, because `package.el` for Emacs 23.x does not know about the `emacs` pseudopackage, so this clause actually makes the package uninstallable on Emacs < 24. :-)

microamp:
I see. Thanks for the info. TIL. :)

It's been removed from `Package-Requires`.

purcell:
Thanks!

PR #4347, opened 2016-10-23
Add seoul256-colors-theme to melpa

anandpiyer:

This is an Emacs theme port of seoul256 color scheme for Vim.

https://github.com/anandpiyer/seoul256-emacs

Author and maintainer.

**None needed**

### Checklist

### DIFF
--- a/recipes/seoul256-colors-theme
+++ b/recipes/seoul256-colors-theme
@@ -1,0 +1,1 @@
+(seoul256-colors-theme :fetcher github :repo "anandpiyer/seoul256-emacs")


syohex:
Please don't define global function/variable whose name does not start with package prefix for name conflict. (This is emacs lisp coding convention)

purcell:
Thanks. More feedback from me:
- "for Emacs" is redundant and should be removed. All Emacs themes are for Emacs!
- You're using `seoul256` in some places and `seoul256-colors` elsewhere. Either call the theme `seoul256-theme` (recommended) and use the `seoul256-` prefix for identifiers, or use `seoul256-colors-` throughout.
- You don't need `defvar`s for all those constants: just `let`-bind them over the `custom-theme-set-faces` form

anandpiyer:
Thank you @syohex and @purcell! 

I pushed a commit that incorporates your suggestions. Can you take a look and see if it is fine?

@purcell, I had named the package seoul256-theme initially, but had to change when I found that melpa had an existing seoul256-theme package.

purcell:
Hmmm, wouldn't it make sense to consolidate the themes, then? On the other hand, your theme looks more complete - perhaps @ChrisDavison would be happy for us to replace his Seoul256 theme with this one?

anandpiyer:
@purcell @ChrisDavison totally open to consolidation. Please let me know how to proceed.

ChrisDavison:
@purcell @anandpiyer 👍

purcell:
Cool, so if I've understood that thumbs-up from @ChrisDavison correctly, let's go with renaming this to `seoul256-theme`, and we'll change the existing recipe to point at @anandpiyer's version. :-)

anandpiyer:
@ChrisDavison thanks! 

@purcell, renaming done and pushed to the repo.

PR #4348, opened 2016-10-23
add recipe for atomic-chrome

alpha22jp:

This is an Emacs version of [Atomic Chrome](https://atom.io/packages/atomic-chrome) which is an extension for Google Chrome browser that allows you to edit text areas of the browser in Emacs. It's similar to [Edit with Emacs](https://www.emacswiki.org/emacs/Edit_with_Emacs), but has some advantages as below with the help of websocket.

* The input on Emacs is reflected to the browser instantly and continuously.
* You can use both the browser and Emacs at the same time. They are updated to the same content bi-directionally.

https://github.com/alpha22jp/atomic-chrome.git

I'm a maintainer of this package.

**None needed**

### Checklist

### DIFF
--- a/recipes/atomic-chrome
+++ b/recipes/atomic-chrome
@@ -1,0 +1,1 @@
+(atomic-chrome :repo "alpha22jp/atomic-chrome" :fetcher github)
--- a/recipes/emacs-chrome
+++ b/recipes/emacs-chrome
@@ -1,0 +1,1 @@
+(emacs-chrome :repo "alpha22jp/emacs-chrome" :fetcher github)
--- a/recipes/emacs-chrome
+++ b/recipes/emacs-chrome
@@ -1,1 +1,0 @@
-(emacs-chrome :repo "alpha22jp/emacs-chrome" :fetcher github)


syohex:
- README says this package requires Emacs 24.4 or higher versions but there is no minimum Emacs version in Package-Requires
- You should name better package name, for example `chrome-edit` etc. I suppose **emacs** in package name makes no sense
- Is it necessary to call hook in minor-mode manually(I suppose hook functions are called two times) ? `define-minor-mode` macro is expanded with `run-hooks` code.

purcell:
Agreed, this should almost certainly be called `atomic-chrome`.

Further feedback from me:
- You need a `defgroup` to define the parent group for all the `defcustom` items.
- Don't add advice to `save-buffers-kill-emacs` -- there's `kill-emacs-hook` for that

alpha22jp:
Thanks for reviewing. I modified some points as below.

* Added emacs version dependency to Package-Requires.
* Renamed package name to "atomic-chrome".
* Removed unnecessary call of `run-hooks`.
* Added `defgroup` and set parent group to "applications".

> Don't add advice to `save-buffers-kill-emacs` -- there's `kill-emacs-hook` for that

I first tried to use `kill-emacs-hook`, but it didn't help me being prompted to kill the websocket server process. So I borrowed this idea from [edit-server](https://github.com/stsquad/emacs_chrome/blob/master/servers/edit-server.el) package. Could you tell me what is the proper way to do it?

purcell:
Thanks!

> first tried to use kill-emacs-hook, but it didn't help me being prompted to kill the websocket server process. So I borrowed this idea from edit-server package. Could you tell me what is the proper way to do it?

Hmm. I'm surprised that didn't work. Not sure then, sorry!

PR #4350, opened 2016-10-24
Add creamsody-theme

jasonm23:

Creamsody is a color theme for Emacs

https://github.com/emacsfodder/emacs-theme-creamsody

I'm the package author

n/a

### Checklist

### DIFF
--- a/recipes/creamsody-theme
+++ b/recipes/creamsody-theme
@@ -1,0 +1,1 @@
+(creamsody-theme :fetcher github :repo "emacsfodder/emacs-theme-creamsody")


purcell:
Nice, thanks. Quick feedback:
- Please remove `A color theme for Emacs.` throughout - it's redundant
- Please avoid unsafe local variables: `;; eval: (when (fboundp 'rainbow-mode) (rainbow-mode 1))`. [Here's a better way to do that](https://github.com/purcell/emacs.d/blob/61dacdcba53e3a7bd2120161f7825cd976e64ff6/lisp/init-lisp.el#L250-L261).
- Since you're defining all those colour vars in a big `let` scope, you might like to skip the `creamsody-` prefix for them

jasonm23:
- unsafe local variables...

        ;; eval: (when (fboundp 'rainbow-mode) (rainbow-mode 1))

    I'm going to leave this as it is.  This is for people editing the theme in Emacs.

    It's not loaded by `require`, `load-library` or `load-file` so there's no issue for people who are just using the theme.  Please tell me if I'm wrong about this.

- removed the superfluous "a color theme for emacs"

- the `creamsody-` prefix helps with autocompleting color names.  It'll stay too.

jasonm23:
emacsfodder/emacs-theme-creamsody @ 47ba10f

purcell:
Whether or not to use `rainbow-mode` is a user preference, it's not a property of the file, so this is a misguided policy.

jasonm23:
It's a fair point, but it's just an opinion.  

The case can be made that it helps people discover a package which they may not otherwise know about.  Package discovery is a haphazard affair.  Marking rainbow mode as a dependency would be _misguided_.

That's the reason why it's in my themes.

It's also the policy I've chosen, you may disagree, but it's not _misguided_ Steve, it's just not in line with your opinion.  I'm sorry about that, but I would hope that you appreciate the motivation.

jasonm23:
Please let me know if this recipe merge is blocked.  Cheers.

purcell:
> It's a fair point, but it's just an opinion.

It's also right there in the [elisp manual](https://www.gnu.org/software/emacs/manual/html_node/emacs/Specifying-File-Variables.html#Specifying-File-Variables):

> Note, however, that it is often a mistake to specify minor modes this way. Minor modes represent individual user preferences, and it may be inappropriate to impose your preferences on another user who might edit the file. If you wish to automatically enable or disable a minor mode in a situation-dependent way, it is often better to do it in a major mode hook (see Hooks). 

I should have explained this more fully, sorry.

I'm not going to block acceptance of this recipe over this minor point, but here's why I spend time on points like this which you might consider pedantic: the more I can give feedback which aligns code in MELPA with the elisp conventions, the better off we'll all be in the long term, because elisp authors learn from the code already out there. I like to encourage authors to put the community and its conventions ahead of their personal preferences, however defensible those seem in isolation.

jasonm23:
Thank you Steve, I don't generally disagree, but I consider theme elisp a special case.

These are quite often first contact elisp for newcomers to Emacs, so switching on Rainbow mode for them, if they have it will help out a lot.

This isn't a personal editing preference. Think of it as an agenda item for guerilla accessibility of emacslisp.

jasonm23:
I think a good course of action is to add a comment to the locals block which states the intent and makes a recommendation not to generally use these vars in packages.

jasonm23:
BTW if you read the manual page:

> it _may be_ inappropriate

That's very much open to interpretation.

PR #4351, opened 2016-10-24
Add recipe for virtual-desktops

chep:

This package allows you to save/retore a frame configuration: windows and buffers.

https://github.com/chep/virtual-desktops.el

**None needed**

### Checklist

### DIFF
--- a/recipes/virtual-desktops
+++ b/recipes/virtual-desktops
@@ -1,0 +1,1 @@
+(virtual-desktops :fetcher github :repo "chep/virtual-desktops.el")


purcell:
Thanks! There are already multiple packages in MELPA which do this, including `spaces` and `simple-screen`, so I'm not keen to add another unless it brings something unique, and I'm not sure that's the case here, sorry. Additionally, there's a lot of code here which I think could/should just be handled by `set-window-configuration` etc.

(Related: the built-in `desktop.el` can handle all of the save/restore functionality, persisting it across restarts, and it's possible to then put a "switching" layer on top, which is what the `desktop-registry` package does.)

PR #4352, opened 2016-10-24
youtube-this

maximus12793:

A simple plugin to watch youtube inside emacs!

https://github.com/maximus12793/youtube-this

I am the maintainer

### Checklist

### DIFF
--- a/recipes/#youtube-this#
+++ b/recipes/#youtube-this#
@@ -1,2 +1,0 @@
-(youtube-this :fetcher github :repo "maximus12793/youtube-this")
-c
--- a/recipes/#youtube-this#
+++ b/recipes/#youtube-this#
@@ -1,0 +1,2 @@
+(youtube-this :fetcher github :repo "maximus12793/youtube-this")
+c
--- a/recipes/helm-youtube
+++ b/recipes/helm-youtube
@@ -1,0 +1,1 @@
+(helm-youtube :fetcher github :repo "maximus12793/helm-youtube")
--- a/recipes/youtube-this
+++ b/recipes/youtube-this
@@ -1,1 +1,0 @@
-(youtube-this :fetcher github :repo "maximus12793/youtube-this")
--- a/recipes/youtube-this
+++ b/recipes/youtube-this
@@ -1,0 +1,1 @@
+(youtube-this :fetcher github :repo "maximus12793/youtube-this")


syohex:
- This package uses  cl.el macros but it is not loaded. This causes byte-compile error(However cl-lib should be used instead cl.el for coding convention)
- This package requires `helm`, `request` but there is no package dependency header(`Package-Requires:`)
- `-pkg.el` file should be removed
- Don't use free variables
- Don't declare function whose name does not start with package prefix
- `Version:` should be version-string(ex `1`, `1.2`, `1.2.3`)

maximus12793:
youtube-this.el:51:4:Warning: misplaced interactive spec: `(interactive)'
unsure how to get rid of this but everything else should be good to go!

maximus12793:
@

alphapapa:
The description is misleading: this package does not enable the watching of YouTube videos in Emacs.  It provides a way to select a video which is then opened in the external browser.

maximus12793:
I have changed the description to say "Query youtube via emacs and play videos in a browser of your choice". Will that work? Beyond this I do not see any other changes. 

I may try to embed the video in emacs **later** as another project but I believe that would slow it down significantly. (and can update package if it is desirable)

xuchunyang:
> youtube-this.el:51:4:Warning: misplaced interactive spec: `(interactive)'
> unsure how to get rid of this but everything else should be good to go!

You should place `(interactive)` at the beginning of the function.

By the way, you should not call

```
(require 'helm-config)
(helm-autoresize-mode 1)
```

in your package, it is probably unneeded by your package, and more importantly, it  overrides users's configuration, which is bad. If you need it for yourself, simply put it to your init.el.

maximus12793:
modified per @xuchunyang 's request

alphapapa:
There are a few more issues, e.g. [here](https://github.com/maximus12793/youtube-this/blob/afabd7e0ab2384b011a5651145797e1c1d6eb9fb/youtube-this.el#L81):

*  `defvar` is for defining variables that maintain global state, not for variables used in a function.  Use `let` for this.
*  "Earmuffs" are used for defining special/global variables, and these are not those kind of variables.  But earmuffs are not used in elisp conventions anyway.  If you were defining a global variable, it should be prefixed with your package's name and have a docstring in the `defvar`.
*  CamelCase is not typically used in elisp, so e.g. `videoId` should be `video-id`.
*  You could improve the loop by writing it something like this (note especially the use of `helm-build-sync-source`):
```el
(defun youtube-this-wrapper (json)
  (let* ((results (cl-loop for element in (cdar json)
                           collect (list (cdr (youtube-this-tree-assoc 'title x))
                                         (cdr (youtube-this-tree-assoc 'videoId x)))))
         (source (helm-build-sync-source "youtube-this"
                                         :candidates results
                                         ... )))
    (helm :sources source)))
```
*  One other suggestion: since you always get the `cdr` of the result from `youtube-this-tree-assoc`, you could just have it return what you need.
* I may be wrong about this, but I think instead of `(eval-when-compile (require 'cl-lib))` you should just use `(require 'cl-lib)`.  You can test this by installing your package in the MELPA sandbox (`make sandbox`), which installs and byte-compiles it in a clean config.
*  Please be sure to give your functions docstrings.  If you run `flycheck-package`, it will mark these issues, and if you fix all of the ones it finds, it will make reviewing the package easier.  :)

maximus12793:
@alphapapa could u do a pull with ur suggested optimization my lisp is very novice so I am not sure what you mean. Everything else has been updated/checked with flycheck-package

purcell:
Great review feedback @alphapa, thanks. Good job fixing up the code, in general.

I'd add a couple of points:
- Since `helm` is mandatory for this, it should probably be called something like `helm-youtube` instead. `helm` is an Emacs lifestyle choice, and not for everyone. :-)
- You definitely don't want this: `setq helm-sources`
- Please add `(cl-lib "0.5")` to the `Package-Requires` list: `flycheck-package` should have mentioned that to you.

maximus12793:
@purcell I have updated it to helm-youtube etc. but should I create a new pull now that the package is different or are u able to change the name? everything else is the same. And is the setq able to be replaced by let? Again my elisp is not great so am not sure what is the best usage here or how to get @alphapapa suggestion about youtube-this-wrapper

alphapapa:
@maximus12793 Sorry, I don't have time right now to send a PR.  Regarding `cl-loop`, everything I know about it is from reading in the Emacs info files and trial-and-error.  `C-h f cl-loop RET` and click the link at the bottom.  ;)

The reason you shouldn't set `helm-source` is that it's an internal Helm variable.  Just use `let` and name your own variable `source` or something that doesn't interfere with the `helm-` namespace.

maximus12793:
@alphapapa yea @syohex got it working as you described. Thanks for the tips! @purcell hopefully it is good to go now?

syohex:
Please add variable of API key as @alphapapa said. With current implementation, users need to rewrite code for setting their own API keys(And they also need to re-byte-compile file). This is ugly.

maximus12793:
@syohex not sure how to do this. I have tried 

`(defgroup helm-youtube nil
  "Helm-youtube settings"
  :group 'tools
)

(defcustom helm-youtube-key "NONE"
  "Your Google API key";; INSERT YOUR KEY FROM GOOGLE ACCOUNT!!!
  :group 'helm-youtube
)

("key" . (format "%s" helm-youtube-key)))
("key" . helm-youtube-key))
`

However I get sequencep errors etc. so suspect its some issue with putting a variable into a (k,v) value. Any idea whats wrong?

syohex:
Did you unquote it ?

```
 (request
   "https://www.googleapis.com/youtube/v3/search"
   :params `(("part" . "snippet")
             ("q" . ,(read-string "Search YouTube: "))
             ("type" . "video")
             ("maxResults" . "20")
             ("key" . ,helm-youtube-key))
```

maximus12793:
@syohex  yes 

http://imgur.com/a/7b9Pl

alphapapa:
> @syohex yes 

It's not unquoted in that screenshot.

maximus12793:
@alphapapa what does 'it's' refer to? helm-youtube-key is unquoted everything else was initially quoted.

alphapapa:
@maximus12793 I don't understand what you mean.  `helm-youtube-key` is not unquoted in that screenshot.  Compare the screenshot to the code that @syohex posted.

maximus12793:
@alphapapa sorry didn't realize ,var was a thing. Fixed. yAny other changes?

alphapapa:
The comma unquotes forms inside backquoted forms.  It's a lisp thing.  It's especially important in macros, but it is also useful when building lists.  The PCL book has some info about them in the macro chapter here: http://www.gigamonkeys.com/book/macros-defining-your-own.html

purcell:
Sent you a further small fix in https://github.com/maximus12793/helm-youtube/pull/4, but I'll go ahead and merge this now anyway. Thanks!

PR #4353, opened 2016-10-24
[new] debian-changelog-mode + debian-bug recipes

ardumont:

This is a major mode for Debian changelog files.

https://github.com/pcarrier/debian-changelog-mode

I use and like the package.
- [X] [Notified the upstream maintainer ](https://github.com/pcarrier/debian-changelog-mode/issues/1)
### Checklist
- [X] I've built and installed the package using the instructions in the [README](https://github.com/melpa/melpa/blob/master/README.md) - C-c C-c method.

Cheers,

### DIFF
--- a/recipes/debian-bug
+++ b/recipes/debian-bug
@@ -1,0 +1,4 @@
+(debian-bug :fetcher cvs
+            :url ":pserver:anonymous@cvs.alioth.debian.org:/cvs/pkg-goodies-el"
+            :module "emacs-goodies-el"
+            :files ("elisp/debian-el/debian-bug.el"))
--- a/recipes/debian-changelog-mode
+++ b/recipes/debian-changelog-mode
@@ -1,0 +1,4 @@
+(debian-changelog-mode :fetcher cvs
+            :url ":pserver:anonymous@cvs.alioth.debian.org:/cvs/pkg-goodies-el"
+            :module "emacs-goodies-el"
+            :files ("elisp/dpkg-dev-el/debian-changelog-mode.el"))


syohex:
This package uses some [debian-bug.el](https://apt-browse.org/browse/debian/wheezy/main/all/debian-el/35.2+nmu1/file/usr/share/emacs/site-lisp/debian-el/debian-bug.el) features. Can this package work without it ?

ardumont:
> This package uses some debian-bug.el features.

Ah yes, indeed.

>  Can this package work without it ?

Well, partially then.
The part that i used works (which are not in relation to debian-bug).
Those parts are the one to manipulate the changelog file (add, edit, finalize, etc...).

--

Taking a look a tad further, debian-bug is from the debian-el's debian package https://packages.debian.org/sid/debian-el (cvs...).

```sh
$ debcheckout debian-el
declared cvs repository at :pserver:anonymous@cvs.alioth.debian.org:/cvs/pkg-goodies-el emacs-goodies-el
...
```

I don't know how to package that part though.

ardumont:
> I don't know how to package that part though.

Mmm, i guess fetching a tarball could do it.
Although, reading again melpa's readme, the cvs fetcher exists :D so no worry apparently for that part...

ardumont:
Well, i got as far as:

```lisp
(debian-bug :fetcher cvs
            :url ":pserver:anonymous@cvs.alioth.debian.org:/cvs/pkg-goodies-el"
            :module "emacs-goodies-el"
            :files ("elisp/debian-el/debian-bug.el"))
```

but now, since it's missing the first `;` in the header (only 2 instead of 3 `;`), the install fails:
`Package lacks a header`...

ardumont:
Hello,

So, i'm unclear as to what is the next reasonable thing to do:
- Drop this altogether (it'd be a shame, i like my [packaging straight](https://github.com/ardumont/dot-files/blob/master/nixos/emacs.nix#L6-L10) :)
- Propose a patch to debian-changelog-mode and then provide some packaging that works for debian-bug (blocked for now, cf. [earlier comment](https://github.com/melpa/melpa/pull/4353#issuecomment-255763283)).

Thanks for any pointers

Cheers,

purcell:
I'd be keen for you to get `debian-bug` fixed upstream, then package both that and `debian-changelog-mode`.

ardumont:
@purcell hello, and thx. I'll see what i can do.

Note that, I don't know how much time this will take to propose a change or make it change in debian-bug.el though.

ardumont:
> (update: i mean the upstream debian-bug.el file is missing the expected ;;; header)

Indeed, i can confirm this is the issue:
- Installing from the buffer using `M-: (package-install-from-buffer)` (from package-build) directly from the upstream `debian-bug.el` raises the error about missing header (for lack of a better method).
- Fixing by adding one ';' and same modus operandi raises no longer that error but another one about missing package version. This is not a problem in the melpa context since that's added before that.

Conclusion: So the upstream fix seems to simply add that missed ';'.

Also, taking a closer look at the upstream repository debian-el, only a small portion of .el files are lacking it. So it might be a simpler matter of asking for it :D.

```
$ find -iname "*.el" | xargs head -1 | grep '^;; .*' | grep '.el' | wc -l
10
$ find -iname '*.el' | xargs head -1 | grep '^;;; .*' | grep '.el' | wc -l
180
```

ardumont:
For the sake of completeness, after further analysis, it seems that the original recipe (from that PR) will also have to change to point to the real upstream debian-changelog-mode (which is not in github but on the same cvs repository mentioned here).

From the checked-out cvs repository:
```
$ find -iname debian-changelog-mode
./elisp/dpkg-dev-el/debian-changelog-mode.el
```

ardumont:
["Bug report"](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=842566) sent upstream with a patch about it.

ardumont:
> "Bug report" sent upstream with a patch about it.

And upstream merged the patch ^^.

So, I have:
- rebased the code to the latest master
- added 1 commit with the debian-bug recipe [as detailed](https://github.com/melpa/melpa/pull/4353#issuecomment-255763283)
- updated the initial recipe debian-changelog-mode to use the right upstream repository [as mentioned previously](https://github.com/melpa/melpa/pull/4353#issuecomment-257145840)
- checked both recipes work and installs correctly and runs (they do)

Cheers,

purcell:
Great work!

PR #4355, opened 2016-10-24
Fix conditional in readme (unless -> if)

Stebalien:


purcell:
Good catch - thanks very much!

PR #4360, opened 2016-10-25
Correct when forks will be accepted.

Stebalien:
I assume this was supposed to say "forks will *only* be accepted in extreme circumstances". Otherwise, this should probably read "forks will never be accepted, not even in extreme circumstances."


Stebalien:
(nevermind, I read "accepted except" as just "accepted").

PR #4363, opened 2016-10-26
add recipe sdcv.

stardiviner:

Emacs interface for command sdcv.

https://github.com/stardiviner/sdcv.el

maintainer

no need

### Checklist

### DIFF
--- a/recipes/sdcv
+++ b/recipes/sdcv
@@ -1,0 +1,2 @@
+(sdcv :fetcher github
+      :repo "stardiviner/sdcv.el")


purcell:
Thanks. Quick feedback:
- Parent of defgroup should be `'editing`, not `'edit`.
- Be careful of code like `(shell-command-to-string (format "espeak -v en %s &" word))` because words containing quotes, spaces or `;` will potentially cause very bad things to happen. Use `shell-quote-argument`.

stardiviner:
@purcell fixed now.

purcell:
Thanks.

PR #4364, opened 2016-10-26
Add recipe for frames-only-mode

davidshepherd7:

Makes Emacs use Emacs frames (i.e. OS windows) instead of Emacs windows where possible. 

https://github.com/davidshepherd7/frames-only-mode

Maintainer

**None needed**
### Checklist

### DIFF
--- a/recipes/frames-only-mode
+++ b/recipes/frames-only-mode
@@ -1,0 +1,1 @@
+(frames-only-mode :repo "davidshepherd7/frames-only-mode" :fetcher github)


syohex:
- Minimum package version should be specified. Minimum version of `seq` is not specified
- `advice-add`, `advice-remove` requires Emacs 24.4 or higher versions
- No autoload cookie

davidshepherd7:
Thanks I've fixed those issues now.

purcell:
Thanks. Couple more review points from me:
- `(require 'org-agenda)` and `(require 'ediff-wind)` shouldn't be required here: why would we need to load all of `org` to enable this mode?
- That list of vars to set revertably seems like it should be a `defcustom` or similar, not hard-coded inside the mode function.

davidshepherd7:
Would it be reasonable to require those files when the mode is run then?
Ideally I would like to make that revertible set function work with
eval-after-load, but I think I would need to sit down and think about for a
while to make it work correctly and I'm pretty busy at the moment.

purcell:
> Would it be reasonable to require those files when the mode is run then?
> Ideally I would like to make that revertible set function work with
> eval-after-load, but I think I would need to sit down and think about for a
> while to make it work correctly and I'm pretty busy at the moment.

No, I don't think you should ever pro-actively require other libraries just so you know about their variables. Why not only set/revert vars which are `boundp`?

davidshepherd7:
The problem is that then when you e.g. run magit for the first time after enabling frames-only-mode then the settings won't be applied unless you toggle frames-only-mode off and on again (unless I'm missing something?).

Anyway I'm getting less busy now :) so hopefully I'll have a chance to figure out how to integrate `eval-after-load` into that `revertable-set` function soon.

davidshepherd7:
@purcell I think I've sorted it now. I don't `require` any external stuff anymore (and there's a [test](https://github.com/davidshepherd7/frames-only-mode/blob/master/load-test/loading-test.el) to make sure). I was looking at the [eval-after-load docs](https://www.gnu.org/software/emacs/manual/html_node/elisp/Hooks-for-Loading.html) and found this:

>  If you need to examine and set the variables defined in another library (those meant for outside use), you can do it immediately—there is no need to wait until the library is loaded.

So at least until I run into a case where it breaks that's what I'm doing.

davidshepherd7:
Also if anyone has any suggestions for tidying up all the `car`/`cdr` stuff [here](https://github.com/davidshepherd7/frames-only-mode/blob/master/frames-only-mode.el#L131) that would be great. I tried `defstruct` but it ended up incredibly verbose because of the namespacing.

syohex:
> Also if anyone has any suggestions for tidying up all the car/cdr stuff here that would be great. I tried defstruct but it ended up incredibly verbose because of the namespacing.

Use `cl-first`, `cl-second` or `nth` functions instead of `car`, `cdr`

purcell:
Thanks!

PR #4366, opened 2016-10-27
Add reciepe to include drone ci launcher

olymk2:

Runs drone exec with in current project, drone is a ci service like jenkins it uses docker and lets you run locally with the same environment

https://github.com/olymk2/emacs-drone

I have created this package, its my first go at coding in lisp

**None needed**
### Checklist

### DIFF
--- a/recipes/drone
+++ b/recipes/drone
@@ -1,0 +1,3 @@
+(drone
+ :repo "olymk2/emacs-drone"
+ :fetcher github)


olymk2:
Any comments welcome, I don't mind admitting i am new to emacs and lisp in particular :)

syohex:
- Package description is nonsense
- No package dependency header
 - However I think it is enough to use `locate-dominating-file` instead of `projectile` function
- No `provide`
- No autoload cookie
- Use setting `let` + `default-directory` instead of `cd directory;` in shell command

olymk2:
Thanks for the suggestions @syohex I have a question regarding locate-dominating-file

from what i can tell locate-dominating-file searches all descendants all the way back to root i am wondering if this is perhaps not desired behaviour it could run a file not related to the project, perhaps this does not really matter but i am interested in your thoughts, it would be nice to not rely on projectile.

I will address the other issues I think i have found out everything i need to make those changes.

olymk2:
@syohex I was thinking about the project root issue some more and decided your correct so i have pushed with the amendments.

https://github.com/olymk2/emacs-drone/commit/bdcc68b93aeadb8395ec32211748ea58b1d3e392

Anything else need addressing, I am making these changes in another much larger package and integrating tests so this will be helpful when i put in my next PR.

olymk2:
Just updated the description, your correct it was nonsense i think i misunderstood that part and changed the summary previously.

now change the description in the header [history](https://github.com/olymk2/emacs-drone/commits/master)

purcell:
Thanks. Quick feedback from me now:
- You should either do nothing (or call `error`) when `dc-drone-root` returns `nil`, because the following code makes no sense otherwise.
- `shell-command` is easy, but it's meant for end users, and isn't a great way to run specific processes like this because they end up competing for the user's `*Shell command*` buffer. Better would be to handle `*drone*` buffer(s) yourself, use `compilation-start` etc. After all, you will probably also want to detect when the process fails so that you can report it when that happens.

So in its current form, I'm not sure there's enough here to be worth distributing, but with a little more polish that would change.

olymk2:
@purcell great thanks for those tips i will go and research and implement those changes :)

olymk2:
@purcell I have made these changes based on your comments, let me know if that what you intended ?

https://github.com/olymk2/emacs-drone/commit/b849f315fd8ae2f9b9ac8b7e569aa40f1e0d9adb

Also a general question, does melpa add packages from a specific commit ?
Just curious what happens if i updated my repo after something is included in melpa ?

purcell:
That wasn't quite what I meant. I've sent you some fixes in https://github.com/olymk2/emacs-drone/pull/1.

MELPA will re-build whenever you make further commits. If you tag your repo with a version number, corresponding packages will also appear on stable.melpa.org.

PR #4367, opened 2016-10-27
Add cmd-to-echo recipe

mallt:

Show the output of long-running commands (f.ex. javascript bundlers) in the echo area.

https://github.com/mallt/cmd-to-echo

I am the maintainer

/
### Checklist

### DIFF
--- a/recipes/cmd-to-echo
+++ b/recipes/cmd-to-echo
@@ -1,0 +1,2 @@
+(cmd-to-echo :fetcher github
+             :repo "mallt/cmd-to-echo")


syohex:
- `compgen` works only on bash. zsh implements it too but it is a shell function so it does not work with `shell-command-to-string`. I suppose `bash -c 'compgen -c'` is better than it. For example `(process-lines "bash" "-c" "compgen -c")`

mallt:
Thank you @syohex! I've replaced the shell-command-to-string with process-lines.

purcell:
This looks really handy, but tying this code to `bash` by calling out to the shell is a bad idea, sorry. Just read a string, and consider using `shell-command-history` or `compile-history`.

mallt:
Hi @purcell, maybe I could also ignore any errors for the completion command:
`(completing-read "Command: "
                 (ignore-errors
                     (process-lines "bash" "-c" "compgen -c")))`

This way people using bash will still get the command completions.

Thanks!

purcell:
It's still not a great idea IMO. No comparable package does anything like this merely for the sake of convenient completions.

In fact, I dug a bit deeper, and there's a standard `read-shell-command` function in `simple.el` -- you should just use that! :-)  Then if you want to add some magic completion for that function, you can `fset` it to a custom version something more to your liking, without changing this package directly.

mallt:
Thanks a lot for the help @purcell, that is indeed exactly what I was looking for! I'm sorry to have missed that function... I've updated the repo.

purcell:
Yay, great - merged! I'll have to add this to my own config.

It occurred to me that it might be quite nice to make a further command which could be run in a `*Shell command*` or `*Compilation*` buffer, such that it would hide the buffer and then start piping output through to the echo area...

mallt:
Thanks!

That would be nice indeed, I quickly looked at it but I think this might require tweaking the internals of `shell-command` or `compile` as I don't immediately see an option to specify a custom filter for these functions.

PR #4369, opened 2016-10-28
Drop unmaintained themes

Fanael:
I don't maintain them anymore, they have pretty bad face coverage and I really want to delete them. Better to drop them from MELPA first before deleting the repos.


tarsius:
@purcell I am strongly in favor of doing this, no point keeping unmaintained themes around.

PR #4373, opened 2016-10-29
persp-mode-fr

rocher:

In 'persp-mode', show the list of perspectives in the GUI window title. It also indicates the currently selected one.

https://github.com/rocher/persp-mode-fr

Maintainer.

None.

### Checklist


purcell:
Thanks.
- There's a mix of `persp-fr` and `persp-mode-fr` prefixes being used here: they should match, and so I'd suggest renaming the package to `persp-fr`, and prefixing all its identifiers consistently to match.
- You're using `add-function`, which is only available in very new Emacs versions, so you'll need to either change that or declare a package dependency on the minimum Emacs version. `flycheck-package` can help you with that.

rocher:
Hi Steve,

Thank you very much for your feedback. 

You're right, it should be named `persp-fr-mode` and use only `persp-fr` as prefix. Anyway, I will rename it `persp-fr` and will fix Emacs dependencies, so I will send a new push request for `persp-fr`.

PR #4375, opened 2016-10-29
Add chinese-pyim-wbdict recipe

tumashu:

Some wubi dicts for chinese-pyim

https://github.com/tumashu/chinese-pyim-wbdict

I'm  the maintainer

None needed

### Checklist

- [+] I've read [CONTRIBUTING.md](https://github.com/melpa/melpa/blob/master/CONTRIBUTING.md)
- [+] I've built and installed the package using the instructions in the [README](https://github.com/melpa/melpa/blob/master/README.md)

### DIFF
--- a/recipes/chinese-pyim-wbdict
+++ b/recipes/chinese-pyim-wbdict
@@ -1,0 +1,3 @@
+(chinese-pyim-wbdict :fetcher github
+                     :repo "tumashu/chinese-pyim-wbdict"
+                     :files (:defaults "*.pyim" "*.pyim.gz"))


tumashu:
Wubi is a kind chinese input method which is based the sharp of chinese char.

syohex:
- Why does not it require `chinese-pyim` package as package dependency ? Your package does not work without it at all.
- Why does this package allow byte-compiling ? (`no-byte-compile: t`)

tumashu:
Fixed, thanks!
```
Why does this package allow byte-compiling ? (no-byte-compile: t)
```

tumashu:
This package is just a data package, it can be installed without chinese-pyim. if chinese-pyim has not been installed, this package do nothing but message "you should install chinese-pyim".
```
Why does not it require chinese-pyim package as package dependency ? Your package does not work without it at all.
```

purcell:
> This package is just a data package, it can be installed without chinese-pyim. if chinese-pyim has not been installed, this package do nothing but message "you should install chinese-pyim".

Yes, that's exactly why it should depend on `chinese-pyim`. It's not useful without it.

tumashu:
Fixed

```
Yes, that's exactly why it should depend on chinese-pyim. It's not useful without it.
```

purcell:
Super, thanks.

PR #4377, opened 2016-10-30
Add recipe for persp-fr  (was persp-mode-fr)

rocher:

In persp-mode, show perspective list in the GUI window title

http://github.com/rocher/persp-fr

Maintainer

None
### Checklist


alphapapa:
Would it be a good idea to make the name a bit more descriptive, something like `persp-frame-name` or `persp-name-in-frame`?

syohex:
Could you squash commits ? There are a lot of unrelated commits.

purcell:
> Would it be a good idea to make the name a bit more descriptive, something like persp-frame-name or persp-name-in-frame?

Kinda agree, but I already requested a renaming, from `persp-mode-fr`, and I'm not going to impose another, so I'll just squash & merge this now. :-)

rocher:
Well, in fact, right now I'm thinking about a new mode that use the frame name to show screen/perspective lists for both `elscreen` and `persp-mode`. The idea is to use a common high level API and to adapt it to different *back-ends* (it's so different the way `elscreen` and `persp-mode` are implemented!!). This new mode would replace `elscreen-fr` and `persp-fr` (I'm the maintainer of both), and will be open to adapt new Emacs screen/perspective/window/frame managers.

So let's leave `persp-fr` for the moment until the new mode is ready. It has no name yet, so I'm open to hear your suggestions for the name of this new Emacs mode `;-)`:

- `frame-name-manager` ?
- `frame-modeline`  (kind of `modeline`, but in the frame name) ?
- ...

PR #4378, opened 2016-10-30
waf-mode added

valden:

Waf build tool integration for Emacs

https://bitbucket.org/dvalchuk/waf-mode

I'm a maintainer of that package.

None needed
### Checklist

### DIFF
--- a/recipes/waf-mode
+++ b/recipes/waf-mode
@@ -1,0 +1,1 @@
+(waf-mode :fetcher bitbucket :repo "dvalchuk/waf-mode")
--- a/recipes/waf-mode
+++ b/recipes/waf-mode
@@ -1,1 +1,1 @@
-(waf-mode :fetcher bitbucket :repo "dvalchuk/waf-mode")
+(waf-mode :fetcher git :url "https://bitbucket.org/dvalchuk/waf-mode")


syohex:
- Use `git` fetcher instead of `bitbucket` when you use git on bitbucket. (`bitbucket` feature supports only mercurial)
- There is no package header, author, copyright, license, package description, package dependencies, commentary section etc. Please read packaging document
- Don't use `C-c w` prefix for emacs lisp coding convention. Please read packaging document
- File pattern of `auto-mode-alist` should be improved(for example `wscript\\'` etc). `wscript` matches against `awscriptb`.  And use `add-to-list` for avoiding registering multiple times

valden:
I've fixed all your notes.

syohex:
`C-b` prefix overwrites`backward-char` key binding. I thinks it is bad for Emacs users except evil users.

valden:
But it is not forbidden by emacs lisp coding convention. Ok - what is your suggestion? Maybe you know free prefix?

alezost:
@valden, Keys for minor modes are rather strict; most of the times they begin with `C-c`.

See (info "(elisp) Keymaps and Minor Modes")

valden:
I've changed prefix to `C-c f`

syohex:
https://www.gnu.org/software/emacs/manual/html_node/elisp/Key-Binding-Conventions.html#Key-Binding-Conventions

Key binding conventions say 

> Don’t define C-c letter as a key in Lisp programs. Sequences consisting of C-c and a letter (either upper or lower case) are reserved for users; they are the only sequences reserved for users, so do not block them.

valden:
@syohex And below on the same page:

> Sequences consisting of C-c followed by any other ASCII punctuation or symbol character are allocated for minor modes. Using them in a major mode is not absolutely prohibited, but if you do that, the major mode binding may be shadowed from time to time by minor modes.

purcell:
But `f` is not punctuation or a symbol.

purcell:
More feedback from me:
- "for Emacs" in the package description is redundant and should be removed in the three places where it appears: all elisp packages are for Emacs. :-)
- You should set `compilation-scroll-output` as a local variable, not globally.
- Avoid hanging `)`: this is bad style.
- Your `condition-case` form can probably just be `ignore-errors`
- The `auto-mode-alist` modification feels out of place here, since there's no direct connection between this code and `python-mode`
- Is there a `wscript` in every waf-enabled project directory? If so, then you should avoid the heavyweight `projectile` dependency, and just use Emacs' own `locate-dominating-file` function instead.
- I imagine you plan for users to add `waf-conditionally-enable` to a hook, but if so, you should mark it as `;;;###autoload`.

Hope that helps!

valden:
@purcell thank you for feedback. 
Regarding "Avoid hanging ): this is bad style." - Is it requirement or I can use my style? Because as for me "))))))" is a bad style - it is difficult to read that code without editor (f.e. here in github web interface).

valden:
@purcell regarded your note: 

> The `auto-mode-alist` modification feels out of place here, since there's no direct connection between this code and `python-mode`

The Waf building tool is writtent in python and it's scripts (wscript) files are true python code. So I do not see why Emacs should not treat them as python code.

purcell:
Yeah, I guess the `auto-mode-alist` modification is probably okay.

Regarding the `)`, I strongly suggest you reconsider the idea of "my style" vs. the community conventions: those conventions are there to allow everyone in the elisp community to work together consistently, which is more important than individual preferences. There are tools (e.g. `rainbow-delimiters`, `hl-sexp-mode`) which make it much easier to work with `)))` etc. if you find them confusing.

valden:
@purcell regarding your comment:
> You should set `compilation-scroll-output` as a local variable, not globally.
I tried to replace `(setq compilation-scroll-output t)` with `(setq-local compilation-scroll-output t)` and it does not work. Maybe you meant another solution?

purcell:
> I tried to replace (setq compilation-scroll-output t) with (setq-local compilation-scroll-output t) and it does not work. Maybe you meant another solution?

No, that's the right solution, but you need to make sure that the code in `waf-setup-compile-buffer` is executed with the compilation buffer as the current buffer. So `with-current-buffer` is what you need. And similarly, you should add to the *local* version of `compilation-filter-hook`, by passing the `LOCAL` flag to `add-hook`.

valden:
Please review it again, It seems I've finished resolving of all your feedbacks.

valden:
@syohex fixed.

purcell:
> Please review it again, It seems I've finished resolving of all your feedbacks.

The redundant "for Emacs" is still there, as is the `projectile` dependency. Those can both be eliminated.

tarsius:
@valden Any progress on the remaining issues?

valden:
@purcell can you suggest how to eliminate the projectile dependency?
eliminate `(projectile-project-root)` is easy but what I can do with following snippet?
lines: 55-58
`(save-some-buffers (not compilation-ask-about-save)  (lambda ()  (projectile-project-buffer-p (current-buffer)  default-directory)))`

Here I save only **this project** related files before compilation.

valden:
@tarsius everything is done except "projectile dependency"

valden:
@purcell For now I see only one option: move outside that lambda and call it here only if it was defined (like I did it with key prefix). Is it ok?

purcell:
You can replace `projectile-project-root` like this:

```el
(defun waf--project-root (&optional file)
  "Return the waf project root for FILE, or for `default-directory' if not specified."
  (or (locate-dominating-file (or file default-directory) "wscript")
      (error "No wscript found")))
```

Then your cmd function becomes this:
```el
(defun waf--run-build-cmd (cmd)
  "Call `waf CMD' in the root of the project."
  (let* ((waf-root (waf--project-root))
         (default-directory waf-root)
         (cmd (concat "waf " cmd)))
    (save-some-buffers (not compilation-ask-about-save)
                       (lambda ()
                         (and default-directory
                              (string-prefix-p default-directory waf-root))))
    (compilation-start cmd)))
```

and `waf-conditionally-enable` becomes:

```
(defun waf-conditionally-enable ()
  "Enable `waf-mode' only when a `wscript' file is present in project root."
  (when (ignore-errors (waf--project-root))
    (waf-mode 1)))
```

You can also just write `waf--run-build-cmd` like this:

```el
(defun waf--run-build-cmd (cmd)
  "Call `waf CMD' in the root of the project."
  (let* ((waf-root (waf--project-root))
         (default-directory waf-root)
         (cmd (concat "waf " cmd))
         (compilation-save-buffers-predicate
          (lambda ()
            (and default-directory
                 (string-prefix-p default-directory waf-root)))))
    (compile cmd)))
```

valden:
Thank you, @purcell .

valden:
Please review my changes. I've done with your notes.

syohex:
I think you should write more document for users into README.md or commentary section
- How to enable this minor-mode
- Default key bindings
- etc

valden:
@syohex done. 
Anything else?

purcell:
Nice, thanks!

PR #4379, opened 2016-10-31
Delete flycheck-google-cpplint

lunaryorn:
There's no one in Flycheck who still maintains this extensions; it hasn't been touched since years.

I'll move it out of Flycheck org into my account for archiving purpose in case someone's still interested, but if none steps up I'll eventually make it private to avoid the hassle of having yet another public repo in my account that I don't care for.  

@tarsius Iirc you're maintaining the emacs-mirror.  Do you per chance also have a place for repos that aren't maintained anymore?

/cc @cpitclaudel

### DIFF
--- a/recipes/flycheck-google-cpplint
+++ b/recipes/flycheck-google-cpplint
@@ -1,1 +1,0 @@
-(flycheck-google-cpplint :repo "flycheck/flycheck-google-cpplint" :fetcher github)


tarsius:
> Do you per chance also have a place for repos that aren't maintained anymore?

Yes, the [Emacs**attic**](https://github.com/emacsattic). Quoting from the [Emacsmirror homepage](https://emacsmirror.net):

> If a package’s upstream disappears, then a copy remains available on the mirror. Once its upstream has disappeared a package is usually moved from the Emacsmirror to the Emacsattic, where it is no longer updated.

There's also an Emacsorphanage:

> If other mirrored packages still depend on a package whose upstream has disappeared or no longer maintains the package, then it is copied to the Emacsorphanage. The Emacsmirror then mirrors that repository from the orphanage. The orphanage repository serves as a temporary home until someone volunteers to take over as maintainer. While a package is in the orphanage, the maintainer of the Emacsmirror makes an effort to merge pull requests from users.

---

I plan to wait for @purcell to remove this and `frame-restore` from Melpa, and then I will move these packages from the mirror to the attic too.

tarsius:
By the way, I very much appreciate it when maintainers explicitly deprecate their old packages as you do here. There are so many packages now, it really helps knowing which of those one does not have to consider anymore when looking for cool new stuff.

lunaryorn:
@tarsius Thanks for the information, and for taking care of the emacsmirror 👍 

I presume what you said in your comment holds for #4380 as well?

tarsius:
@purcell I think this should be removed.

tarsius:
And so I have ;)

tarsius:
Ups, this is a pr, ah well, I created a new commit.

thekondr:
Hi guys,

I recently found out that this package is missing and I was very sad about it until I realised spacemacs allows to install packages directly from github.

I don't think that it is not maintained. Last commit was [less than one year ago](/flycheck/flycheck-google-cpplint/commit/a97c9ed05d3e7ecb28923f90c56b5fe518024abe). It just works fine enough to not require any new changes.

Hope you will bring it back :)

purcell:
@thekondr Please can you check in with the new Flycheck maintainers, perhaps by opening an issue on the upstream repo? If someone (them or you!) is happy to maintain that repo, then I'm happy to restore the recipe here.

jcs090218:
@purcell I am the new maintainer for this package now. Can you consider restore the package back to melpa? Thanks! :)

tarsius:
Link to the repository: https://github.com/flycheck/flycheck-google-cpplint.

tarsius:
Added it again.

jcs090218:
@tarsius Thank you! :)

PR #4380, opened 2016-10-31
Delete frame-restore

lunaryorn:
frame-restore is a kludge for a missing feature in Emacs 24.3 and earlier.  Since Emacs 24.4 the built-in Desktop Mode can restore frames in a much better and more powerful way than this package.

As such, I'm not maintaining frame-restore.el anymore since I've got no intention of ever using Emacs 24.3 again 8)

I've anyone wants to take the package fine with me, but if no one steps up I'll be making the repository private at some point and eventually delete it.

@tarsius I guess you're mirroring the repo anyway so that would do no harm would it?

### DIFF
--- a/recipes/frame-restore
+++ b/recipes/frame-restore
@@ -1,1 +1,0 @@
-(frame-restore :repo "lunaryorn/frame-restore.el" :fetcher github)


purcell:
I include it in my emacs config as a fallback for older Emacs versions, but I’d be equally happy to drop it, I guess.

lunaryorn:
@purcell Oh, didn't know that.  In that case I'm fine with keeping it, though I'd like to get it out of my Github account to be honest :blush:

purcell:
Thing is, I don’t know if anybody even uses that, and recently I’m feeling a bit more inclined to just ensure graceful degradation in old emacs versions rather than trying to selectively switch between old and new functionality.

lunaryorn:
@purcell So I understand that you actually rather prefer to remove it than to keep it?

purcell:
Yep, and now I've removed it in my config, so I've simply gone ahead and merged this. :-)

lunaryorn:
@purcell Thank you very much.

@tarsius Can I just delete my repository and the mirror will keep up, or do you need to take some action first?

tarsius:
@lunaryorn No need to wait for me.

lunaryorn:
@tarsius Thank you very much. Great to know that the repository is mirrored for all that are still interested, your work on the emacsmirror is really awesome :+1: Thanks a lot!

lunaryorn:
@tarsius I've made the repo private, if there are any issues I can quickly make it public again.

If nothing bad happens I'll hard-delete the repo by the end of next week.

PR #4381, opened 2016-10-31
add dante

jyp:

Emacs mode for Interactive Haskell

[[http://stable.melpa.org/packages/dante-badge.svg]]

Dante is a fork of Intero mode. It steals many good ideas from Intero,
it aims for light weightedness. Dante has no dependency on "Stack" nor
on any custom Haskell code. Dante's Emacs code is about half as long
as that of Intero. Flychecking is optional (yet recommended).

https://github.com/jyp/dante

**None needed**

### Checklist

### DIFF
--- a/recipes/dante
+++ b/recipes/dante
@@ -1,0 +1,3 @@
+(dante :repo "jyp/dante"
+       :fetcher github
+       :files ("dante.el"))
--- a/recipes/dante
+++ b/recipes/dante
@@ -1,3 +1,2 @@
 (dante :repo "jyp/dante"
-       :fetcher github
-       :files ("dante.el"))
+       :fetcher github)


syohex:
- Please do not global functions/variables/macros which don't start with package prefix. In this case, `cps-bind` and `cps-let` for name conflict.
- This package requires haskell-mode however it does not load haskell-mode.

jyp:
I have fixed those two problems in dante/master. Thanks for pointing them out!

purcell:
Thanks - this is really interesting!

Quick feedback from me:
- The `:files` spec is not necessary here, and should be omitted from the recipe. The README covers this a little.
- You've got a package description in the header, but it should be in the `Commentary:` section of the header so it can be parsed out by the various tools (including ours) which expect standard elisp boilerplate formatting.
- The keybindings you're using are reserved for major modes, as per the [elisp manual](https://www.gnu.org/software/emacs/manual/html_node/elisp/Key-Binding-Conventions.html). Are they "borrowed" from `intero`?
- `(flycheck-select-checker 'haskell-ghc)` should probably be replaced with code that simply triggers `flycheck` to re-select the default checker itself, in case that would not be `haskell-ghc` given the user's config.

jyp:
Thanks for all the good points. Regarding the keybindings: I've changed them to fit the convention (as far as I understand).

Yet, to answer your question: indeed, Intero steals keybindings of the form C-c C-<letter>. There is a long tradition of doing so in interactive "layers" over the (major) Haskell mode. Perhaps the Haskell major mode should officialize the tradition (maybe it does already).

purcell:
> Yet, to answer your question: indeed, Intero steals keybindings of the form C-c C-letter. There is a long tradition of doing so in interactive "layers" over the (major) Haskell mode. Perhaps the Haskell major mode should officialize the tradition (maybe it does already).

Yeah, I think haskell mode and its extensions have been a bit of a "Wild West" for some time! :-)  I'm keen to try `dante` out myself, as a mostly-satisfied `intero` user. I'd love to know (or be able to read about) the differences between the two: presumably `intero` is able to do a bunch of stuff which `dante` can't.

jyp:
@purcell : Intero is "stack only", which is a no-go for me. Dante will work with any GHCi.
Additionally, Dante:
* supports xref-find-definitions / xref-find-references
* has a different approach to Haskell evaluation
* does not support eldoc nor hoogle

purcell:
Cool - thanks for the notes. You should put those in the README too. :-)

PR #4382, opened 2016-10-31
Remove flycheck-protobuf

marsam:
The protobuf checker is already part of flycheck upstream

See:
+ https://github.com/edvorg/flycheck-protobuf/issues/3 
+ https://github.com/flycheck/flycheck/issues/790
+ https://github.com/flycheck/flycheck/pull/1125 

cc: @edvorg, @lunaryorn

### DIFF
--- a/recipes/flycheck-protobuf
+++ b/recipes/flycheck-protobuf
@@ -1,3 +1,0 @@
-(flycheck-protobuf
-    :fetcher github
-    :repo "edvorg/flycheck-protobuf")


milkypostman:
@marsam i think we should add `:old-names (flycheck-protobuf)` to the *flycheck* recipe. Could you do that in this PR?

purcell:
@milkypostman In this particular case, I'd probably not worry about that: web searches will show up `flycheck` for `flycheck-protobuf`, and there's no point aggregating download stats here. Is there a trade-off I'm missing?

milkypostman:
No reason just completeness. I'm ok either way.

purcell:
Cool.

PR #4386, opened 2016-11-02
Add textx-mode

novakboskov:

Basic support for editing [TextX](https://github.com/igordejanovic/textX) files.

https://github.com/novakboskov/textx-mode

I am maintainer.

**None needed**

### Checklist

### DIFF
--- a/recipes/textx-mode
+++ b/recipes/textx-mode
@@ -1,0 +1,1 @@
+(textx-mode :fetcher github :repo "novakboskov/textx-mode")


syohex:
This package requires Emacs 25.1 or higher versions. However it looks Emacs 24.3 can work. Which features requires Emacs 25.1 ?

novakboskov:
@syohex It doesn't require any "special power" or Emacs 25.1. The sole reason I declared that version is that I've tested it only against that version because it is one which I'm currently using. Should I change that to 24.3?

syohex:
> Should I change that to 24.3?

Yes if you possible. Because many people still use older version Emacs(Some Linux distribution like ubuntu does not provide Emacs 25.1 yet).

novakboskov:
@syohex OK, changed.

purcell:
Thanks. Quick feedback:
- You should include a `defgroup` form to declare the `textx` group
- `define-derived-mode` will generate the mode map and hook vars, so you don't need to declare those yourself
- `font-lock-defaults` is always buffer-local, so `setq` is sufficient, rather than `setq-local`

novakboskov:
@purcell Thank you. :)
I did it that way in the latest commit.

syohex:
Thanks. Merged.

PR #4387, opened 2016-11-02
Add recipe for rjsx-mode

felipeochoa:

This mode implements a JSX parser according to the spec [here](https://facebook.github.io/jsx/). When enabled, it plugs the JSX parser into js2-mode's own parser to build the JSX AST. Currently the mode handles syntax checking and highlighting, and when you use js2-jsx-mode (bundled with j2-mode), also indentation.

https://github.com/felipeochoa/rjsx-mode

I'm the author of the package

**None needed**

### Checklist

### DIFF
--- a/recipes/rjsx-mode
+++ b/recipes/rjsx-mode
@@ -1,0 +1,4 @@
+(rjsx-mode
+ :fetcher github
+ :repo "felipeochoa/rjsx-mode")
+ 


felipeochoa:
There are unfortunately a series of "Messages should start with a capital letter" flycheck warnings, which are due to `rjsx-mode` following the style of `js2-mode` (whose message/error reporting this package uses). Other than that, there are no compiler warnings and the package builds on my Windows machine as well as Travis!

purcell:
Thanks! `"Real support for JSX in Emacs"` should simply be `"Support for JSX."`. Elisp code is always for things in Emacs!

Looks good otherwise. I think it would be appropriate here to include an autoloaded form which adds an entry to `auto-mode-alist`, since I don't think there are any other packages which aim to support `.jsx` files.

felipeochoa:
Ok will do. Your suggestion has made me realize that I should probably require `js2` to be loaded when `rjsx` is loaded, or at least warn the user if not. Is there a standard practice for minor modes that only work with certain major modes?

For the `.jsx` `auto-mode-alist` entry, is that just a comment like the following?
```elisp
;;;###autoload (add-to-list 'auto-mode-alist '("\\.jsx\\'" . rjsx-mode))
```

purcell:
Hmmm, I see. Maybe this should actually be a major mode, with `js2-mode` as the parent mode? Feels like it -- and that was why I suggested the `auto-mode-alist` entry.

felipeochoa:
Ok -- it's now a major mode and has the auto-mode-alist cookie. The `defgroup` comment is also updated. Let me know if you have any other thoughts!

purcell:
Looks good - merging now.

BTW, you can switch
```
;;;###autoload (add-to-list 'auto-mode-alist '("\\.jsx\\'" . rjsx-mode))
```
to
```
;;;###autoload
(add-to-list 'auto-mode-alist '("\\.jsx\\'" . rjsx-mode))
```

PR #4388, opened 2016-11-03
Add tile recipe

IvanMalison:
Summary: tile provides a way to interactively cycle through pre-defined custom layout definitions à la XMonad
Repository: https://github.com/IvanMalison/tile

### DIFF
--- a/recipes/tile
+++ b/recipes/tile
@@ -1,0 +1,2 @@
+(tile :repo "IvanMalison/tile"
+      :fetcher github)


purcell:
Thanks -- this is very cool!

The `Commentary` section talks about `multi-line`, so I presume you've just pasted it in from elsewhere. Proper docs about how to customise the package will be essential for users, and that's really not obvious by looking at the code. Is there anything you can do which would allow at least some basic customisation using the `customize` system? e.g. `tile-strategies` could be a `defcustom` at least.

IvanMalison:
> Is there anything you can do which would allow at least some basic customisation using the customize system?

I thought about doing this, but the thing is that anything offered through customize is going to be a bit stunted as far as the amount of customization it will offer. I must admit that I've never really used customize myself and I'm not an expert in its use. My main concern with this case is that it would require the specification of entire lists of customized entries -- there's not really a good way to allow a user to enable or disable specific layouts.

I suppose that there are ways that I could restructure things specifically for defcustom that would allow customize variables that specifically toggle certain layouts, but this might be to the detriment of the code quality...

IvanMalison:
Okay, I came up with something that allows the toggling of specific layouts.

I'd be curious to know what you think about the defaults I chose. Wasn't sure how to choose those.

purcell:
Yeah, not sure if that's necessarily better: I guess the order of the strategies is something the user should also be able to easily customise. My thought was that if the user could define the order of strategies - and even new strategy definitions - using plain data in one or more custom variables, then that would be the best experience for most users: a `setq` or two and they'd be done.

In any case, I'll go ahead and merge this now. :-)

PR #4389, opened 2016-11-04
Add apib-mode recipe

w-vi:

Major mode for editing API Blueprint files [https://apiblueprint.org/](https://apiblueprint.org/)

https://github.com/w-vi/apib-mode

Maintainer

**None needed**

### Checklist

### DIFF
--- a/recipes/apib-mode
+++ b/recipes/apib-mode
@@ -1,0 +1,1 @@
+(apib-mode :fetcher github :repo "w-vi/apib-mode")


syohex:
Please check code and use the package with plain Emacs at least once.

- I cannot load your library. What is `defconst-mode-local` ?
- `nil` is not a function. This package uses `nil` as function at much parts.
- Too many byte-compile warnings.

w-vi:
Fixed
- `defconst-mode-local` comes from cedet so I have removed it
- cleaned up the `nil` is not a function
- No byte-compile warnings when run with clean emacs anymore.

Thanks a lot for the feedback and review it is my first package so I am learning how to do it properly.

purcell:
Quick feedback:
- You can & should remove `in GNU Emacs` from the package description
- Don't use `concat` to build command strings unless you're also using `shell-quote-argument` to make filenames safe: they might contain spaces or quotes
- Adding new global compilation patterns every time you open a file in this mode isn't a good idea. What you probably want instead is to a specific function that end users can call, e.g. `apib-run-drafter`, rather than trying to modify the built-in compilation machinery.
- Define your font lock keywords globally as a `defconst`, then add them by name (e.g. `apib-mode-font-lock-keywords`) in the mode function.
- Don't use `(executable-find "drafter")` at load time, because `exec-path` may or may not get modified before or after loading `apib-mode`: instead, just use the plain string `"drafter"` as the default.

w-vi:
I have updated the package significantly, includes now some sanity test using `ert`. Have tried it using the package in clean emacs as well.

w-vi:
Is there anything I can do to push it?

purcell:
Thanks, looking better! Just a couple more small things, then we should be ready to merge:
- You'll need a `defgroup` to define the `apib-mode` custom group, and place it in one of the standard customisation groups.
- [This sort of thing](https://github.com/w-vi/apib-mode/blob/5386f3ad2c454eccab22d2e9849b762b153e49ee/apib-mode.el#L155-L160) isn't safe when filenames contain spaces, quotes or other characters with special meanings to the shell. Use `shell-quote-argument` to escape command arguments.

w-vi:
Thanks a lot for the feedback I have learnt so much. Will fix those.

w-vi:
Updated so there is a proper group now.
Regarding the `shell-quote-argument` I've setup a [test where there is a space in filename](https://github.com/w-vi/apib-mode/blob/master/apib-mode-test.el#L62) to check where the quoting needs to take place and the bit you have pointed out it seems that I should not do it as then I'm getting error that the file was not found. Looking at `call-process` documentation I am not sure how it works but not quoting works.

syohex:
Please don't put unrelated files(`install-deps.el`, `load-deps`) which matches against default file patterns at top directory and please check package content again if you add files. 

```
% make recipes/apib-mode
% tar tf packages/apib-mode-20161202.117.tar
apib-mode-20161202.117/
apib-mode-20161202.117/apib-mode-pkg.el
apib-mode-20161202.117/apib-mode.el
apib-mode-20161202.117/install-deps.el
apib-mode-20161202.117/load-deps.el
```

w-vi:
Sorry I forgot to update melpa recipe and after your comment decided to move the files to subdirectory as a cleaner solution. I hope that now everything is OK.

syohex:
Looks good

purcell:

w-vi:
I thank you for teaching me.

PR #4391, opened 2016-11-04
Don't exclude jdee-test.el

pwojnowski:
This is an update to the current JDEE recipe to not exclude *test.el, because it's name of our core modules.

JDEE keeps tests in test/*, but uses jdee-test.el as one of packages. Default files recipe excludes one of core file, breaking the whole package.


purcell:
Thanks. Why not just `:files (:defaults "jdee-test.el")`? I think that would be equivalent.

pwojnowski:
Yes, it would be even better. I corrected the recipe. Thanks!

purcell:
Super, thanks.

PR #4392, opened 2016-11-05
add counsel-bbdb recipe

redguardtoo:

Quick search&input email from BBDB database based on ivy-mode

https://github.com/redguardtoo/counsel-bbdb

### Checklist

### DIFF
--- a/recipes/counsel-bbdb
+++ b/recipes/counsel-bbdb
@@ -1,0 +1,1 @@
+(counsel-bbdb :repo "redguardtoo/counsel-bbdb" :fetcher github)


syohex:
Looks good.

purcell:
This should be `ivy-bbdb`, rather than `counsel-bbdb`, I think.

redguardtoo:
@purcell , I'm following the creator of counsel/ivy's file naming convention. Ivy provides basic infrastructure and counsel provides utilities people uses.

http://oremacs.com/2015/04/19/git-grep-ivy/

https://github.com/abo-abo/swiper/commit/e9db958

purcell:
> @purcell , I'm following the creator of counsel/ivy's file naming convention. Ivy provides basic infrastructure and counsel provides utilities people uses.

Yes, but this isn't the Counsel package. There are more `ivy-*` packages than `counsel-*` packages, and I expect that the author just chose "counsel" as the name for _his_ package of completion add-ons, not as a community convention. What do you think, @abo-abo?

redguardtoo:
I'm ok with any namespace.

abo-abo:
@purcell My idea was to keep the `ivy-` namespace small, so that the most fundamental and basic customization features are more easily discoverable via completion.  Then `counsel-` would be a large namespace containing also new namespaces like `counsel-bbdb-`.

But looking at MELPA, there are already more 3rd party `ivy-` packages than `counsel-` packages (approx 6 vs 3). So consistency is already gone.

The naming should always be up to the author. My personal preference is `counsel-bbdb`.

purcell:
Cool, thanks.

PR #4393, opened 2016-11-06
Emacs dashboard package

rakanalh:

Dashboard is an extraction of Spacemacs-buffer into a standalone templates which provides functionality of:
1. Displaying banners
2. Recently modified files
3. List of bookmarks
4. Projectile recent projects

https://github.com/rakanalh/emacs-dashboard

I am the maintainer

None needed

### Checklist

### DIFF
--- a/recipes/dashboard
+++ b/recipes/dashboard
@@ -1,0 +1,3 @@
+(dashboard :fetcher github
+	   :repo "rakanalh/emacs-dashboard"
+	   :files ("dashboard.el" "banner.txt"))


purcell:
Thanks. I sent some fixes in https://github.com/rakanalh/emacs-dashboard/pull/6. 

Overall I'm not completely sure this is going to be widely useful in its current form. For example, the sections are hard-coded, which makes sense for spacemacs: generally, though, plenty of people won't be using projectile or bookmarks, so I feel that what you should be providing here is a bunch of tools so that people can easily make their own dashboards.

rakanalh:
Hello @purcell,

I agree with your concern. I actually do have a plan to enable section selection in configs and it's part of the wishlist. However, i don't think it's useful to do now given that i only provide 3 sections. I am aiming to implement a couple more sections list in the near future and will be providing list customization along with it. I wanted to also postpone submitting to MELPA until those are done but due to some demand i decided to  go for it now.

Is this request a blocker for merging this PR?

Thanks

syohex:
I have sent PR https://github.com/rakanalh/emacs-dashboard/pull/11

rakanalh:
@syohex Thanks! Merged

rakanalh:
I also just merged a change which basically enables users to select which widgets (lists) are displayed on their dashboards. More widgets to come soon.

@purcell This the kinda customization you were referring to, correct?

purcell:
> This the kinda customization you were referring to, correct?

Yes, though you should use `defcustom` etc. rather than `defvar`. My concern is really where things go from here, because the dependency on `projectile` is a huge code smell for a general-purpose package like this, because only a subset of your users will even want to have it installed: if you plan to start adding more predefined sections, with dependencies on corresponding packages, then that's the wrong approach. I'd rather see a truly general and vanilla package which didn't have the projectile dependency, and this would force you to make it really easy for users to add their own custom sections. (With the right set-up, it would be just a line or two to define a projectile section.)

rakanalh:
I do see your concern. Am not very well-versed with elisp but one idea is to check if a certain package exists or not, if that's possible? Because i would want to have the dashboard as batteries included where its covenient for the users to select one of many dashboard widgets available with `Dashboard` and these widgets can and will have dependencies. Providing code snippets that users  can paste into their emacs.d is not the idea i had about this because it would add too much noise to their configs that they have to somehow understand.

In this case, what would you think the best approach is?

rakanalh:
@purcell I added a check in Dashboard's code to require projectile in case the package exists. Otherwise, if the `projects` widget is specified in the list (removed from default), then it will show an error that the package doesn't exist.
https://github.com/rakanalh/emacs-dashboard/blob/master/dashboard.el#L210

If this approach is suitable, i'll probably follow it for the rest of the "battery-included" widgets that depend on other packages.

I'd love your feedback on this.

Thanks

purcell:
> I added a check in Dashboard's code to require projectile in case the package exists.

One of the basic principles of installable emacs packages is that you can have them installed without having them enabled, much like the very many packages included in the Emacs standard library. So trying to load `projectile` isn't appropriate, because the user might not even use `projectile`, and you certainly shouldn't enable `projectile-mode` on the user's behalf just because you can load `projectile`.

*However*, you could certainly check (at runtime, not load-time) that `projectile-mode` is enabled. In general, the batteries-included approach can be made to work, but I think you need each section to not only be an "insert" function, but also a function which reports whether that section is available for use, much like how `auto-complete`, `company` and `flycheck` items work.

Does that make sense?

rakanalh:
Makes total sense. I changed the code so that it checks for projectile-mode to be activated.
https://github.com/rakanalh/emacs-dashboard/blob/master/dashboard.el#L210

purcell:
Cool - merged.

rakanalh:
Woohoo thanks!

PR #4395, opened 2016-11-07
add fluxus-mode

defaultxr:

Fluxus-mode is a mode for interfacing with the Fluxus live coding environment. Fluxus can be used to program VJ-style 3D graphics or full games in real-time using a Racket-based language with functions for creating 3D graphics, as well as sound, among other things. This package allows you to use Emacs to write code for Fluxus instead of having to use Fluxus's built-in editor. Fluxus-mode connects to Fluxus using the OSC (open sound control) protocol and can send code to it to be evaluated by its interpreter. It also has functions to redirect Fluxus's error output to a buffer within Emacs, among other convenience utilities. Fluxus's home page is located at http://www.pawfal.org/fluxus/ .

https://github.com/defaultxr/fluxus-mode

I created and maintain this package.

**None needed**

### Checklist

### DIFF
--- a/recipes/fluxus-mode
+++ b/recipes/fluxus-mode
@@ -1,0 +1,3 @@
+(fluxus-mode
+ :fetcher github
+ :repo "defaultxr/fluxus-mode")


syohex:
Please remove `pkg.el` file. `fluxus-mode.el` has package information and it is not consistent with `pkg.el`.

defaultxr:
Removed `fluxus-mode-pkg.el`.

syohex:
`fluxus.scm` is not in package(it is not specified in `files:` attribute). Is it ok ?

defaultxr:
`fluxus.scm` is a file for Fluxus, not Emacs. It's the Fluxus half of the connection between the two programs. So it shouldn't be in the `files:` attribute.

purcell:
Thanks. Quick feedback:
- The `with-eval-after-load` block for smartparens is out of place here, and should be placed in your personal config instead. Conditional configuration/extension of other possibly-installed packages isn't something that a package should generally be doing. (Also, `with-eval-after-load` is intended for user configs, as noted in the elisp manual.)
- Add an `;;;###autoload` cookie comment for `fluxus-start`.
- The F5-59 key bindings are [reserved for users](https://www.gnu.org/software/emacs/manual/html_node/elisp/Key-Binding-Conventions.html), so you shouldn't bind them

Looks good otherwise. :-)

defaultxr:
Thanks for the feedback - I've made those changes as well.

purcell:
Super, thanks.

PR #4397, opened 2016-11-09
Add all-the-icons-dired

jtbm37:

Adds dired support for [all-the-icons](https://github.com/domtronn/all-the-icons.el) package.

https://github.com/jtbm37/all-the-icons-dired

Maintainer

**None needed**

### Checklist

### DIFF
--- a/recipes/all-the-icons-dired
+++ b/recipes/all-the-icons-dired
@@ -1,0 +1,3 @@
+(all-the-icons-dired
+ :repo "jtbm37/all-the-icons-dired"
+ :fetcher github)


syohex:
https://github.com/jtbm37/all-the-icons-dired/blob/d74cc4472b0a9107e414f12e0ac129de9e3b34b7/all-the-icons-dired.el#L73-L78

This `if` clause always evaluate `then` part.  Should it be `all-the-icons-dired-mode` instead of `t` ?

jtbm37:
You're absolutely correct. Some left over from testing. Will fix this after work.

jtbm37:
Fixed.

purcell:
You've defined a local minor mode, which implies it can be toggled in any given dired window, but its definition changes the global value for the `dired-after-readin-hook`. I think you should instead change the local version, using `add-hook`'s LOCAL flag.

jtbm37:
Thank you for reviewing the code.
Fixed.

purcell:
Thanks.

PR #4399, opened 2016-11-10
sudoku recipe import

zevlg:

Play sudoku in GNU Emacs.   Includes such features as:

* Custom puzzle editor
* Save/Load puzzles
* Puzzles downloader from websudoku.com and menneske.no
* Puzzles solver

and much more

https://github.com/zevlg/sudoku.el

I'm the author and maintainer

None needed

### Checklist

### DIFF
--- a/recipes/sudoku
+++ b/recipes/sudoku
@@ -1,0 +1,1 @@
+(sudoku :fetcher github :repo "zevlg/sudoku.el")


tarsius:
There's already another `sudoku.el`: https://sourceforge.net/projects/sudoku-elisp/. Have you looked at that?

zevlg:
Yeah, I've seen that implementation. Even more, I was inspired by it to write this one

Jr2075's implementation was missing crucial features, and there wes several stuckpoints in code to extend it, so I decided to totally rewrite it

purcell:
Cool, but what should we do to avoid confusion? Even if you believe your sudoku.el is better, we would generally consider that name to have been claimed by the previous version. Perhaps a variation on the name?

zevlg:
@jkr what do you think about naming?  Should we keep plain `sudoku.el` name, but interfering with your initial name, or choose another one?

zevlg:
Will wait couple days for @jkr response, then rename to `sudoku+' on silence

jkr:
Haven't really thought about it in a decade -- consider it abandoned and do what you think best. If that's renaming, do so with my blessing.

zevlg:
Thanks, great!  Let's then keep plain `sudoku.el` for convenience

purcell:
Super, thanks.

PR #4400, opened 2016-11-11
Add flycheck-demjsonlint recipe

z4139jq:

This package provides a [flycheck][flycheck-ref] checker for [json-mode][json-mode-ref] using jsonlint from [demjson][demjson-ref].
The built in JSON checker uses [zaach/jsonlint][jsonlint-ref] which has no rules support, but jsonlint from demjson has a set of options let you control which checks are to be performed, e.g. trailing-comma is allowed in .eslintrc but forbidden in package.json.

https://github.com/z4139jq/flycheck-demjsonlint

I am the maintainer and the contributor of this package.

**None needed**

### Checklist

- [ ] I've read [CONTRIBUTING.md](https://github.com/melpa/melpa/blob/master/CONTRIBUTING.md)

[demjson-ref]: https://github.com/dmeranda/demjson "demjson"
[json-mode-ref]: https://github.com/joshwnj/json-mode "json-mode"
[jsonlint-ref]: https://github.com/zaach/jsonlint "jsonlint"
[flycheck-ref]: http://www.flycheck.org "Flycheck"

### DIFF
--- a/recipes/flycheck-demjsonlint
+++ b/recipes/flycheck-demjsonlint
@@ -1,0 +1,1 @@
+(flycheck-demjsonlint :repo "z4139jq/flycheck-demjsonlint" :fetcher github)


purcell:
Thanks. Quick feedback:
- You should include the necessary information in the `Commentary` section of the `.el` file rather than referring people to the README: the code should be self-documenting, and the `Commentary` section is used to provide information used on the package page on MELPA and in the Emacs package menu.
- The current recipe won't include the `demjsonlint` file. It that correct? It looks like you're assuming it will be there.
- If someone wants to tweak their own flycheck configuration, they should be able to use the checker without `flycheck-demjsonlint-setup`: however, that function sets the path to `demjsonlint`. Instead, set the default value for the `-executable` variable at the top level: you can use something like this:
```el
(setq flycheck-json-demjsonlint-executable
      (expand-file-name "demjsonlint"
                        (file-name-directory (or load-file-name (buffer-file-name)))))
```

z4139jq:
Thanks for your help, I have fixed those issues you mentioned.
But I have mis-deleted the fork and lost track of the compare branch, so I have to close this PR, give another one to commit the change, sorry for this.

PR #4402, opened 2016-11-11
Ignore `plantuml-mode` test files

skuro:

The package is already included in MELPA, but in future releases more files will be added to support testing, including the PlantUML JAR file itself. To avoid killing MELPA with useless files, the recipe now has a `:files` whitelist including just the `plantuml-mode.el` file.

https://github.com/skuro/plantuml-mode

I'm the maintainer

**None needed**

### Checklist

### DIFF
--- a/recipes/plantuml-mode
+++ b/recipes/plantuml-mode
@@ -1,2 +1,3 @@
 (plantuml-mode :fetcher github
-               :repo "skuro/plantuml-mode")
+               :repo "skuro/plantuml-mode"
+               :files ("plantuml-mode.el"))


syohex:
I suppose it is better to use sub directory instead of specifying each files.

Default pattern is here.  For example you can avoid bundling test files by putting them into `test/` directory.

```
("*.el" "*.el.in" "dir"
 "*.info" "*.texi" "*.texinfo"
```

skuro:
Tests are already within the `tests` folder, and I can make the most worrisome one (a JAR file) to fall into it (it's currently inside a `bin` folder, which I assume it's also ignored, given the pattern you provided). If that's the case I will simply close the PR.

purcell:
Correct, so I'll go ahead and close this - thanks!

Also note that you can easily verify that the existing recipe does not include any new files by using the simple steps described in the README to build the package locally.

PR #4406, opened 2016-11-13
Add recipe for flycheck-title

Wilfred:

flycheck-title is a small package that allows you to see flycheck messages in the Emacs frame title, keeping the minibuffer free for other things.

https://github.com/Wilfred/flycheck-title

I'm the author.

None needed.

### Checklist

### DIFF
--- a/recipes/flycheck-title
+++ b/recipes/flycheck-title
@@ -1,0 +1,2 @@
+(flycheck-title :repo "Wilfred/flycheck-title"
+                :fetcher github)


purcell:
Thanks, looks handy!
- In the body of `flycheck-title-mode`, why check `(null ...)` in the conditional? I'd expect to see you check the value of the `flycheck-title-mode` variable there instead.
- Would it make sense to open this up a little so that you expose a function which users can themselves set as their `flycheck-display-errors-function`, or use to build their own compound display-errors function? e.g. what if people would like their errors in both the modeline and frame title?
- Re. the `Commentary` text, I'd encourage you to not simply refer people to the README: this is the place for the library to document itself, in the grand Emacs tradition. :-)

Wilfred:
> I'd expect to see you check the value of the flycheck-title-mode variable there instead.

:) So would I! However, this doesn't protect a minor mode from being activated twice: https://github.com/flycheck/flycheck-pos-tip/issues/23#issuecomment-260134664

> Would it make sense to open this up a little so that you expose a function which users can themselves set as their flycheck-display-errors-function, or use to build their own compound display-errors function?

Other flycheck packages override `flycheck-display-errors-function` too, e.g. `flycheck-pos-tip`. This gives a good out-of-the-box experience for new users. We could explore adding that functionality if users request it, but I'm not sure it's a common workflow worth supporting in v1.

> Commentary text

Fixed, thanks.

purcell:
👍

PR #4412, opened 2016-11-14
Add temporary-persistent recipe please

kostafey:

temporary-persistent - easy way to switch temp buffers and keep them persistent.
Like notes. But simpler ;). No special mode, no timers.

https://github.com/kostafey/temporary-persistent

### DIFF
--- a/recipes/temporary-persistent
+++ b/recipes/temporary-persistent
@@ -1,0 +1,2 @@
+(temporary-persistent :repo "kostafey/temporary-persistent"
+                      :fetcher github)


syohex:
There are too many byte-compile warnings please check them.

kostafey:
@syohex, Thank you, fixed.

purcell:
Please can you add a `Commentary` section so that people might have a clue about how to use this? :-)

Overall I'm not exactly sure why you're reinventing the `buffer-file-name` functionality. Seems like you could use the built-in file/buffer machinery for much of this.

Also, don't use `melpa.milkbox.net` in the MELPA setup instructions, please - melpa.org is the way to go, or - better - refer people to our setup instructions, which will always be up-to-date.

kostafey:
@purcell Ok, fixed. Thank you for code checking.

syohex:
Looks good

purcell:

PR #4413, opened 2016-11-14
Added recipe for vital-dark-theme

tashrifsanil:

Vital Dark theme is a minimal Emacs color theme that highlights important elements of your code, and is meant to be easy on the eyes. It's an attempted clone of the vital dark syntax theme for emacs. This theme uses spacemacs-theme as base. There's only a dark variant and it should work 256 color terminals.

https://github.com/tashrifsanil/vital-dark-theme

I guess I'm all of the above

### Checklist

### DIFF
--- a/recipes/vital-dark-theme
+++ b/recipes/vital-dark-theme
@@ -1,0 +1,1 @@
+(vital-dark-theme :fetcher github :repo "tashrifsanil/vital-dark-theme")


syohex:
- Please remove byte-compiled file from repository
- Please remove autoload file from repository.

Both files are generated by package.el at installing.

tashrifsanil:
Sorry about that, I've removed the byte-compiled file and the autoload file

purcell:
Thanks. Quick feedback:
- `"A dark color theme for Emacs"` -- all Emacs themes are for Emacs, so you should remove `for Emacs` from this! :-)
- It would be preferable for you to use the standard boilerplate comments in `vital-dark-theme.el`: just 2 of several advantages would be that the `-pkg.el` file would then not be necessary, and you would have included information about the licence under which you are distributing this code. (Ironically, you have all the necessary boilerplate in the `-common` file, but not this one.)
- `dyn-let`, `true-color-p` etc. should be prefixed with the package name.
- Testing for display capabilities at the time you create the theme is a completely invalid approach, since the theme will potentially be used on many future frames with different capabilities. The correct approach is to specify different conditional class variants for each of the faces, then Emacs will automatically select the variant for the current frame based upon its display capabilities.

Hope that helps!

tashrifsanil:
Thanks for your reply purcell. I've made changes based on your first 3 points. But the your last point is where I'm kind of at a loss. I'm only beginning to understand some of the capabilities of the colour theme, and most of it was shamelessly taken from the dark variant of the [spacemacs theme](https://github.com/nashamri/spacemacs-theme). Needless to say the look of the theme came to be by fiddling around with the spacemacs theme. I'm kind of unsure about how as to go about your last point, an example code snippet or something would be very helpful

purcell:
> I'm kind of unsure about how as to go about your last point, an example code snippet or something would be very helpful

Sure. Here's how the built-in `font-lock-keyword-face` is defined:
```el
(defface font-lock-keyword-face
  '((((class grayscale) (background light)) :foreground "LightGray" :weight bold)
    (((class grayscale) (background dark))  :foreground "DimGray" :weight bold)
    (((class color) (min-colors 88) (background light)) :foreground "Purple")
    (((class color) (min-colors 88) (background dark))  :foreground "Cyan1")
    (((class color) (min-colors 16) (background light)) :foreground "Purple")
    (((class color) (min-colors 16) (background dark))  :foreground "Cyan")
    (((class color) (min-colors 8)) :foreground "cyan" :weight bold)
    (t :weight bold))
  "Font Lock mode face used to highlight keywords."
  :group 'font-lock-faces)
```
Note that this single face definition has variants with different class specs, and Emacs then chooses the first variant which matches the specs of the current frame. It's pretty elegant, and completely standard. :-)

tashrifsanil:
I do apologize for the late response, but I'm kind of having issues wrapping my head around what specifically I'm supposed to change. Do I modify everything after the basics section, and do I declare each of the faces with defface as shown in the snippet? I do apologize if this comes off as overly trivial

purcell:
Yep, such that each face entry under basics would have a couple of variants defined. The true-colour variant would have the class `((class color) (min-colors 16777216))` and the other would have a class like `((class color) (min-colors 89))`.

tarsius:
@tashrifsanil Any progress on this?

tarsius:
This has been open for more than half a year now. Closing.
Feel free to open a new pull request when you are ready.

PR #4415, opened 2016-11-16
Added recipe for language-detection

andreasjansson:

Automatic programming language detection from snippets of code.

https://github.com/andreasjansson/language-detection.el

I'm the maintainer.

None needed

### Checklist

### DIFF
--- a/recipes/language-detection
+++ b/recipes/language-detection
@@ -1,0 +1,2 @@
+(language-detection :fetcher github
+                    :repo "andreasjansson/language-detection.el")


recipes/language-detection
```diff
@@ -0,0 +1,3 @@
+(language-detection :fetcher github
+                    :repo "andreasjansson/language-detection.el"
+                    :files ("language-detection.el"))
```
syohex: You need not to specify `:files` attribute, it is enough to use default value.

Please read 
- https://github.com/melpa/melpa#recipe-format

syohex:
- No autoload cookie
- No `provide`

andreasjansson:
Thanks for review @syohex. Addressed comments in https://github.com/andreasjansson/language-detection.el/commit/3b180a5a0416286051e39e511ce51400c21310f0

Fanael:
You declare a dependency on `cl-lib` but don't use it. Could you change `loop` to `cl-loop` and let the old `cl` library rest in peace?

andreasjansson:
Thanks @Fanael, I changed the [`(require 'cl)` to `(require 'cl-lib)`](https://github.com/andreasjansson/language-detection.el/blob/master/language-detection.el#L72) and [`loop` to `cl-loop`](https://github.com/andreasjansson/language-detection.el/blob/master/language-detection.el#L96). I also [removed the dependency on cl-lib](https://github.com/andreasjansson/language-detection.el/blob/master/language-detection.el#L7) since I assume that's bundled with Emacs and doesn't need an explicit require. Is that correct?

Fanael:
It's only bundled with Emacs 24.3 and newer. Either declare a dependency on that, or put the `cl-lib` dependency back in.

andreasjansson:
I put the [`cl-lib` dependency back in](https://github.com/andreasjansson/language-detection.el/blob/master/language-detection.el#L7).

purcell:
Thanks. I think `language-detection-buffer` would also be useful programmatically, so perhaps change the body of it so that it always returns the detected language, and only prints it when called interactively.

PR #4416, opened 2016-11-18
Add projectile-ripgrep

nlamirault:

Ripgrep integration into Projectile

https://github.com/nlamirault/ripgrep.el

I'm the maintainer

### Checklist

Signed-off-by: Nicolas Lamirault <nicolas.lamirault@gmail.com>

### DIFF
--- a/recipes/projectile-ripgrep
+++ b/recipes/projectile-ripgrep
@@ -1,0 +1,3 @@
+(projectile-ripgrep :repo "nlamirault/ripgrep.el"
+                 :fetcher github
+                 :files ("projectile-ripgrep.el"))


syohex:
I have sent PR https://github.com/nlamirault/ripgrep.el/pull/11

- Missing dependency header
- Some needless code

purcell:
Thanks, merging not, but I noticed one thing:

```el
(read-from-minibuffer "Ripgrep search for: " (thing-at-point 'symbol))
```

Here, in some modes, symbol constituent characters can sometimes be regexp special characters, so you might consider calling `regexp-quote` on the result.

Additionally, you should use the higher-level `read-string` function rather than `read-from-minibuffer`, because then you can provide an input history ring for the user's convenience.

PR #4426, opened 2016-11-25
add new recipe major-mode-icons

stardiviner:

Show icon for major-mode on icon.

https://github.com/stardiviner/major-mode-icons

Maintainer

### Checklist

### DIFF
--- a/recipes/major-mode-icons
+++ b/recipes/major-mode-icons
@@ -1,0 +1,4 @@
+(major-mode-icons :fetcher github
+                  :repo "stardiviner/major-mode-icons"
+                  :files (:defaults "icons")
+                  )


syohex:
- Please conform Emacs Lisp package convention
  - missing some package headers(author, dependencies, license etc)
- Please don't use `/` separator, use `-`  instead of it(`--` can be used for private symbols)
- Please check too many byte-compile warnings

recipes/major-mode-icons
```diff
@@ -0,0 +1,3 @@
+(major-mode-icons :fetcher github
+                  :repo "stardiviner/major-mode-icons"
+                  )
```
syohex: `icons/` directory is missing.

syohex:
https://github.com/stardiviner/major-mode-icons/tree/master/icons

There are bug.xpm and Bug.xpm. Case insensitive file system like NTFS, HFS+ can see it as only one file.

stardiviner:
@syohex All fixed. check it out the latest push.

syohex:
- cl-lib functions are used but cl-lib is not loaded
- Don't use cl.el macro, use cl-lib macro instead 
- map-keys is introduced at Emacs 25.1 and you should load `map.el` for using it
- What is `activate` function ?
- There are two `Commentary` sections
- `Package-Requires` header position is wrong
-  Make no sense `;;; -*- coding: utf-8 -*-` at second line, it should be first line or buffer local variable section
- `This file is part of GNU Emacs.` Really ?
- `Copyright (C) 2016 Free Software Foundation, Inc` Really ?

Please check carefully before PR.

stardiviner:
> Don't use cl.el macro, use cl-lib macro instead
What do you mean this?
Does `cl-some` has another name in `cl-lib`?

syohex:
`case` is cl.el macro. Use `cl-case` instead of it. You can check it by byte-compiling with minimum configuration.

stardiviner:
I updated my recipe to include `icons/` directory.

purcell:
- Instead of `:files ("*.el" ("icons" "icons/*"))`, please just use `:files (:defaults "icons")`, which is equivalent.
- The `;;; Commentary:` section is misplaced. Please follow the standard formatting, as inserted by the `auto-insert` command.
- `cl-some` is autoloaded, so you don't need to require `cl-extra`
- The Emacs 25.1 dependency ensures that you'll already have a recent `cl-lib`, so you can remove the `(cl-lib "0.5")` dependency.
- It's possible to install `clojure-mode` but not `cider`, in which case your code will break. Same with `python-mode` and `pyvenv-mode`.
- The `post-command-hook` looks unnecessary. You can ask for the selected window at any time.

Overall, it looks like you should just modify `mode-line-format` in order to add an `(:eval ...)` section. What you're doing appears to be a hand-rolled version of that.

purcell:
Also, you can get the preferred display name of a mode just by inspecting the `mode-name` variable in a buffer with that mode, so there's no need to hard-code a list of names, which might not match what the user wants.

stardiviner:
I have changed my source code, waiting for merge. (maybe some places I missed, can figure out) Thanks for you guys review my code.

syohex:
Please don't use `--` separator for customize variables. `--` means private function/variable but customize variables are public, not private.

stardiviner:
@syohex Updated.

syohex:
Merged

PR #4429, opened 2016-11-29
Add recipe ivy-rich.

Yevgnen:

Provides more friendly interfaces for `ivy`.

https://github.com/Yevgnen/ivy-rich

maintainer

### Checklist

### DIFF
--- a/recipes/ivy-rich
+++ b/recipes/ivy-rich
@@ -1,0 +1,2 @@
+(ivy-rich :repo "yevgnen/ivy-rich" :fetcher github :files
+               ("*.el"))


syohex:
- missing `projectile` dependency
- `subr-x` requires Emacs 24.4 or higher version
- Please write package description(Short description at first line after `---` )

Yevgnen:
@syohex Thanks for your quick reply and advices! I've updated the source. Please check it.

syohex:
It looks package has bug.

https://github.com/Yevgnen/ivy-rich/issues/1

Yevgnen:
@syohex Thanks for reporting it!  I'll looking into it soon.

alphapapa:
Just curious, how exactly does your package improve on `ivy-switch-buffer`?  It would be good to show what the difference is.

Also, in the screenshot, why is there a list of file paths under the buffer list?

Yevgnen:
@alphapapa Currently not much really.  I just make it more like `helm` as I was a `helm` user. The list of full path is the original `ivy` feature, which is called `virtual buffers`. It could be also formatted though. You can see some discussion [here](https://github.com/abo-abo/swiper/issues/256)

Personally, for some `ivy` extensions, the default display of candidates is really unfriendly. For example, here's the `org-ref` using `ivy`

screen shot 2016-11-29 at 11 24 19 pm

I can't see the even basic alignment. 

Compared to `helm` with the default settings

screen shot 2016-11-29 at 11 28 12 pm

I've also wrote a `ivy-org-ref-rich` for it too. Even now it's just like a toy.

In my opinion, this(at least a more slightly friendly interface) should be done by the origin author of the package. But there are so many package, I don't want to advice them one by one and most of them seem not much interested in it. I rather to wrote it myself, and share them to anyone who is interested in it.

purcell:

- `--` in a symbol prefix denotes private symbols, so don't use `ivy--`
- `ivy-switch-buffer-rich` would be better called `ivy-rich-switch-buffer`, since it's in a package called `ivy-rich`
- There's package metadata in both `.el` files, but we'll only read the data in `ivy-rich`. Why not just make this one file, called `ivy-rich-switch-buffer`? Then at least it would make (a little more) sense that it depends on `projectile`.
- In any case, don't assume that the user has activated or wants to use `projectile`. Check whether `projectile-mode` is true.

Yevgnen:
Thanks for your relay and suggestions. I've updated the package into melpa convention. Please check when you have time.

syohex:
Looks good.

syohex:
Merged

PR #4430, opened 2016-11-29
add proportional package

ksjogo:

Globalized minor mode for having a proportional font by default in all buffers, some modes relying on space formatting (tabulated, some popup packages) are switched to mono.

https://github.com/ksjogo/proportional

Maintainer

### Checklist

### DIFF
--- a/recipes/proportional
+++ b/recipes/proportional
@@ -1,0 +1,1 @@
+(proportional :repo "ksjogo/proportional" :fetcher github)


syohex:
You should use face functions instead of `setq` for setting `variable-pitch` face. `setq` causes byte-compile warning about using free variable.

```lisp
(setq variable-pitch `((t :family ,proportional-font)))
```

ksjogo:
Thanks for the tip, applied it.

alphapapa:
Hey, this package looks interesting.  I use proportional fonts in some modes myself, and something to do it more generically would be nice.  Here are some things I noticed:

1.  Your package uses `hydra`, `which-key`, and `use-package`, but doesn't declare dependencies on them.

2.  AFAIK, you shouldn't make calls to `use-package` in yours; those are for users to call in their configs.  Instead, you should `require` the packages you use at the top of the file, and then use them directly.

3.  In your minor-mode, you make a lot of calls to `add-hook` and `remove-hook` with hardcoded lists of modes.  It would be better to define customizable variables of lists of modes in which to use proportional and fixed fonts, and then use those lists.

4.  In that vein, since fixed-width fonts are generally the default in Emacs, it might make more sense to consider proportional fonts the exception rather than the rule.  Right now it seems like your mode changes the default expectation to proportional and then makes exceptions to change some modes back to monospace fonts, which seems a bit backward to me.  :)

ksjogo:
Thanks for the feedback.
1. The use-package dependency is declared in the headers, added a require now. 
2. which-key and lv-message are `:defer t`, so they are not loaded by us, but the config (should only?) be added, when somebody else loads them, add if guards in case my assumption was wrong. In general the intention of this package is to be a one-stop drop in for proportional fonts, I do not want users to have to add more stuff to their config, as these get crufty and I think a lot of effort is duplicated there. 
3. Absolutely right, changed that
4. The main idea behind this mode is to have everything non-monospaced by default so if something new is added it will be fine to go, so

alphapapa:
> which-key and lv-message are :defer t, so they are not loaded by us, but the config (should only?) be added, when somebody else loads them, add if guards in case my assumption was wrong. In general the intention of this package is to be a one-stop drop in for proportional fonts, I do not want users to have to add more stuff to their config, as these get crufty and I think a lot of effort is duplicated there and use-package is great for that.

Sorry, I think this is a bad idea.  Remember, `use-package` is intended for users to use in their configuration.  What happens if someone else (like me) has `(use-package hydra...` in their config, and then your package does the same thing?  It's not designed to be used this way.  This will likely cause very strange behavior that will be hard for users to track down to your package.

A better way is to use `with-eval-after-load`, which will only activate after the specified package is loaded.

But there are some deeper issues with the way you're doing this in general.  For example:

```el
(progn
    (which-key--init-buffer)
    (with-current-buffer which-key--buffer
(proportional-use-monospace)))
```

Here you're calling an internal `which-key` function, one that's only intended to be called by `which-key` itself.  Maybe this works now, but what if `which-key` makes changes in the future which break this behavior?  It could end up causing more weird behavior that will be very difficult for users to track down; non-developer users might be simply unable to do so.

This may also be problematic:

```el
(defadvice lv-message (after proportional)
    (with-current-buffer (get-buffer " *LV*")
      (buffer-face-mode t)
      (proportional-use-monospace)))
  (when proportional-mode
(ad-enable-advice 'lv-message 'after 'proportional))
```

I'm not sure what the `*LV*` buffer is--something Hydra-related I guess--but this is another case where changes in the package you're manipulating might break this.

By the way, `advice-add` supersedes `defadvice`.

In general, having package-specific code in your package is a fragile solution.  You will be left playing catch-up whenever those packages make incompatible changes.

This is why I suggest considering setting the default Emacs font to proportional, and then adding mode-specific hooks to set desired modes back to monospace.  Hooks are a more flexible, canonical way to hook into other packages' functionality.

ksjogo:
Afaik the use-package will kinda compile down to the `eval-after-load` case and so far I did not ran into any problem with having multiple `use-package` for the same package and I could not find anything in the code which would cause a problem there.  And the `:defer` is resulting in an addtion, "which will only activate after the specified package is loaded", so that seems fine to me.

There were some problems with the whick-key hook, so I had to patch the internals. I should probably add some version requirement to allow better checks against changes here. 
Yes, the `*LV*` thing is hydra stuff also.
Will add `boundp`checks against all stuff I am calling.
Absolutely agreeing on hooks being better, some packages have a bit strange font behavior requiring the advices though.
Actually in the beginning I was only hook into some stuff to make it proportional as possible but that result in a lot more code and hacks, as most ui was overwriting it somehow and new modes wouldn't be handled. In general the amount of buffers/faces/... of (wanted to be proportional) stuff is larger than the other side, thus fixing the other side.

syohex:
https://github.com/ksjogo/proportional/commit/21fe8093998c63f95aefacc0a9aef419d9deb84f#diff-5349cc753128120ff6492d5b1d1bd9b8R130

`if-let` requires Emacs 25.1 or higher version.

purcell:
I agree with the feedback from @alphapapa: `use-package` isn't an appropriate way to handle optional other packages. The user might not even be using the package system. I'm also wary of the hard-coded default font family names, which will only work on some platforms.

But anyway, there's `variable-pitch-mode`, so why not just define a globalised version of that which chooses not to enable itself in certain major modes? All the machinery for that is built into every Emacs. What am I missing?

tarsius:
I also agree that the use of `use-package` is not appropriate. @ksjogo do you intend to change that? What about the suggestion about `variable-pitch-mode`?

PR #4435, opened 2016-12-02
add recipe for binary-line-navigation minor mode

mgrachten:

This package provides _binary line navigation_, a quicker alternative to using forward/backward character/word when moving the cursor within long lines: the forward and backward commands allow you to approximate any position in a line with steps of logarithmically decreasing size, analogous to [binary search](https://en.wikipedia.org/wiki/Binary_search_algorithm).

https://github.com/mgrachten/bln-mode

Author/Maintainer

None needed

### Checklist

### DIFF
--- a/recipes/bln-mode
+++ b/recipes/bln-mode
@@ -1,0 +1,3 @@
+(bln-mode
+ :repo "mgrachten/bln-mode"
+ :fetcher github)


syohex:
- Don't define global variable without package prefix(`beg`, `end`, `prev-mid`). And use `defvar` instead of `setq` for declaring global variables
- Write package comment at `;;; Commentary:` section

mgrachten:
Thanks for the quick feedback. I fixed the issues you mentioned.

purcell:

PR #4437, opened 2016-12-04
Correctly extract htmlize version from the git tag

hniksic:
Change as discussed in https://github.com/hniksic/emacs-htmlize/issues/3


purcell:
Thanks!

PR #4439, opened 2016-12-05
add: js-format recipes

futurist:

Format js code using node, by region/buffer to local http server and then back, quickly and customizable.

https://github.com/futurist/js-format.el

I'm the Creator and Maintainer

None needed

### Checklist

# js-format 

Format js code (region/buffer) using js, the default formatter is [standard](http://standardjs.com/), but you can add more.

See the README.md file for details:

https://github.com/futurist/js-format.el

### DIFF
--- a/recipes/js-format
+++ b/recipes/js-format
@@ -1,3 +1,3 @@
 (js-format :fetcher github
            :repo "futurist/js-format.el"
-           :files (:defaults "*.js" "*.json"))
+           :files (:defaults "*.js" "*.json" "styles"))
--- a/recipes/js-format
+++ b/recipes/js-format
@@ -1,3 +1,3 @@
 (js-format :fetcher github
            :repo "futurist/js-format.el"
-           :files ("*.js" "*.el" "*.json"))
+           :files (:defaults "*.js" "*.json"))
--- a/recipes/js-format
+++ b/recipes/js-format
@@ -1,0 +1,3 @@
+(js-format :fetcher github
+           :repo "futurist/js-format.el"
+           :files ("*.js" "*.el" "*.json"))


syohex:
I have sent PR.

https://github.com/futurist/js-format.el/pull/1

recipes/js-format
```diff
@@ -0,0 +1,3 @@
+(js-format :fetcher github
+           :repo "futurist/js-format.el"
+           :files ("*.js" "*.el" "*.json"))
```
syohex: `:defaults` is preferable to `"*.el"`. 

https://github.com/melpa/melpa#recipe-format

futurist:
Changed, please review

syohex:
https://github.com/futurist/js-format.el/blob/2330c672105816e6eaf67a3f319c15315340a051/js-format.el#L254

This line should be `let` + `default-directory` instead of `cd some-directory && some-command`. Some shells like fish does not recognize `&&` syntax.

```lisp
(let ((default-directory some-directory))
  (shell-command-to-string "some-command"))
```

futurist:
See this PR:

https://github.com/futurist/js-format.el/commit/130569ea36e23cc47216e9d2c01673891d104eee

syohex:
Good. By the way, `#'` is not necessary for lambda expression.

syohex:
- https://github.com/futurist/js-format.el/blob/130569ea36e23cc47216e9d2c01673891d104eee/js-format.el#L245
- https://github.com/futurist/js-format.el/blob/130569ea36e23cc47216e9d2c01673891d104eee/js-format.el#L250

Process sentinel function is called when process status is changed. So the process might not exits and still be alive. I think it is better to check whether process exits as below before deleting process.

```lisp
(when (eq (process-status proc) 'exit)
  (delete-process proc))
```

futurist:
updated

futurist:
updated

futurist:
updated as 69664fa8443996e3f8b0f657ad0e46cb43a36b1e

futurist:
Is it safe to untouch the `delete-process` and only check for `exit` state? The `delete-process` is from my old code, but seems not necessary here. see 

https://github.com/futurist/js-format.el/commit/9c8d448e6ac1333cd859acd5bd2eb4d76575d4b1

syohex:
Looks good about code.

And could you improve package description ? Current description is here.

```
Format javascript code using node
```

MELPA already registered similar package [jsfmt](https://github.com/brettlangdon/jsfmt.el).  jsfmt also uses node.js. I think it is better to clarify differences between jsfmt.el and this package. For example, use `standard formatter` instead of `node`.  Please consider.

futurist:
In fact, it can format any string using JS, see:

https://github.com/camilojd/sequeljs

Using this lib you can format SQL.

`js2-mode` dependency can be removed in the future, js2-mode now just help mark region to format.

In other word, it can **pluggable**. I will think and reword the docs later.

purcell:
Thanks. Why would you use a server process for this, rather than just running a js script directly? That seems like a _really_ unusual choice.

futurist:
@purcell, I've tried `call-process` to run a JS code, but every time there's a lag, since starting a new node is a heavy operation. And the output/input pipe not easily controlled if run as deamon.

With need of formatting region constantly, or even, auto formatting when press <kbd>RETURN</kbd>, that lag is fatal, system will use more resource, since malloc and free constantly.

Spawning a server instead, giving a fast response from socket, and you can format remotely (setup a format server in your workplace)

futurist:
I have some basic idea about below features (**future?**):

1. Should present a socket server **list**, try all the server before spawn a local server, a fail connection test is fast in `url-retrieve` ? I've tried it's fast but need more test.

2. Format string according to different parser/formatter available in JS world, and pack them (autoload them) with package (from NPM when needed? ), some parser that can be used for formatter:

 Markdown: https://github.com/evilstreak/markdown-js

 RegExp: https://github.com/JexCheng/regulex

 HTML: https://github.com/fb55/htmlparser2

 CSS: https://github.com/postcss/postcss

 CSV: https://github.com/C2FO/fast-csv

 YAML: https://github.com/nodeca/js-yaml

 SQL: https://github.com/forward/sql-parser

3. Is it a too big idea about formatting?  Is it run out of the way? Should it be just for JS?

futurist:
I'm updating the package and docs, please review:

https://github.com/futurist/js-format.el

For the above message of features:

1.  It should done by users to choose one from their list of format server

2. Already done for now, by adding a new style folder, and register the style name, it can extend.

3. After test, it can really work with other languages, not only for JS. Using JS+NPM is powerful.

syohex:
Looks good.

BTW https://github.com/futurist/js-format.el/blob/master/server.js format is very strange. It uses both tab and space indentation and some indentations are wrong. I suppose it should be formatted by your package.

futurist:
Thank you! The new commit https://github.com/futurist/js-format.el/commit/60e3b991db12ccbc117113349d96302b3f94a04b fixed that!

syohex:
Merged

futurist:
Thank you for the job!

I have a question, does MELPA allow code form stackoverflow?

I want to publish another package: [temp-execute.el](https://github.com/futurist/temp-execute.el)

But a code form stackoverflow:

https://github.com/futurist/temp-execute.el/blob/master/temp-execute.el#L83

If I keep the link and copyright, is that ok to publish to MELPA?

syohex:
I suppose it is no problem if copyright is clear.

syohex:
@futurist BTW there is similar tool [quickrun](https://github.com/syohex/emacs-quickrun/) developed by me. I suppose quickrun has more features than it.

futurist:
Oh, I should know **quickrun** eariler, that will save me more time!

Can quickrun provide a key to quickly remove the output buffer?

syohex:
Yes, it removes temporary files when command is finished

PR #4441, opened 2016-12-06
Add recipe for sensible-defaults

hrs:

sensible-defaults is a collection of well-named functions that tweak some of Emacs' settings. These functions are intended to be called from a user's `.emacs` file, and do things like ensure that files end in newlines, enable syntax highlighting everywhere, and make `dired` file sizes human-readable.

https://github.com/hrs/sensible-defaults.el

**None needed**

### Checklist

(As a warning, `package-lint` errors since there are `/`s in the function names. I'm reluctant to change that, since those names are part of the public interface and we have a few users, but I'm open to it.)

sensible-defaults is a collection of functions that tweak some of Emacs'
settings. These functions are intended to be called from a user's .emacs
file, and do things like ensure that files end in newlines, enable
syntax highlighting everywhere, and make dired file sizes
human-readable.

### DIFF
--- a/recipes/sensible-defaults
+++ b/recipes/sensible-defaults
@@ -1,0 +1,2 @@
+(sensible-defaults :repo "hrs/sensible-defaults.el"
+                   :fetcher github)


syohex:
- No `provide`
- No package footer
- No autoload cookie
- `-` should be used as variable/function name separator instead of `/`

purcell:
Thanks for taking the time to submit this.

This code is really personal configuration rather than a widely usable package, sorry, and in most cases the functions wrap very simple variable or mode settings that users should arguably either learn about and access directly, or discover and modify via the Customize interface.

I'm going to decline this, then.

tzz:
@purcell there are many schemes for managing emacs configurations and settings overlays that get distributed ad-hoc. I think it would be really useful if there was a "SMELPA" for settings to help the community share and distribute such tools. In my mind they are similar to themes, which also don't really provide code.

Spacemacs has done this in a very useful way with its layers.

purcell:
Thanks @tzz, that's a really interesting idea. I'm wary of distributing free-form config code as packages for several reasons: primarily, I believe it actually discourages people from pulling apart code and publishing the widely-useful portions of it as well-formed packages.

One area which I think has been under-explored is the use of the custom theme system to define defaults: we tend to think of custom themes as colour themes, but they include variables, and can be used to layer settings over the defaults, while still respecting user-level customisation.

PR #4442, opened 2016-12-08
Add recipe for importmagic.el

anachronic:

Fix python imports using [importmagic](https://github.com/alecthomas/importmagic).

https://github.com/anachronic/importmagic.el

None needed

### Checklist

Please go easy on me as I'm not very familiar with Emacs Lisp.

### DIFF
--- a/recipes/importmagic
+++ b/recipes/importmagic
@@ -1,0 +1,3 @@
+(importmagic :fetcher github
+             :repo "anachronic/importmagic.el"
+             :files ("importmagic.el" "importmagicserver.py"))
--- a/recipes/importmagic
+++ b/recipes/importmagic
@@ -1,3 +1,3 @@
 (importmagic :fetcher github
              :repo "anachronic/importmagic.el"
-             :files ("importmagic.el" "importmagicserver.py"))
+             :files (:defaults "importmagicserver.py"))


recipes/importmagic
```diff
@@ -0,0 +1,3 @@
+(importmagic :fetcher github
+             :repo "anachronic/importmagic.el"
+             :files ("importmagic.el" "importmagicserver.py"))
```
syohex: Use `:defaults` value instead of specifying each `.el` files.

Please read https://github.com/melpa/melpa#recipe-format

syohex:
Please don't use `C-c i` key binding for Emacs lisp key binding convention.

> Don’t define C-c letter as a key in Lisp programs. Sequences consisting of C-c and a letter (either upper or lower case) are reserved for users; they are the only sequences reserved for users, so do not block them. 

https://www.gnu.org/software/emacs/manual/html_node/elisp/Key-Binding-Conventions.html#Key-Binding-Conventions

anachronic:
Done in both sides: project repo and melpa repo.

syohex:
Looks good.

purcell:
`C-c C-l` is still a binding reserved for major modes. You should at least define the keymap at the top level so it can be customised.

anachronic:
@purcell Thanks for merging.

Is it ok if I leave a suggestion for the user and don't provide a keymap for `importmagic-mode`?

purcell:
> Is it ok if I leave a suggestion for the user and don't provide a keymap for importmagic-mode?

Yep, that's probably a good way to do it.

PR #4451, opened 2016-12-09
Add recipe for winum

deb0ch:

Window numbers for Emacs: Navigate windows and frames using numbers !

This package is an extended and actively maintained version of the https://github.com/nschum/window-numbering.el package by Nikolaj Schumacher, with some ideas and code taken from https://github.com/abo-abo/ace-window.

This version brings, among other things, support for number sets across multiple frames, giving the user a smoother experience of multi-screen Emacs.

https://github.com/deb0ch/winum.el

Maintainer

**None needed**

### Checklist

### DIFF
--- a/recipes/winum
+++ b/recipes/winum
@@ -1,0 +1,1 @@
+(winum :fetcher github :repo "deb0ch/winum.el")


purcell:
Thanks. Quick feedback:
- Use `cl-lib` instead of `cl`. The latter is deprecated, and you're using functions from the former anyway, without `require`-ing them. (Add `(cl-lib "0.5")` to your `Package-Requires` header accordingly.)
- Don't override `M-0` or `M-m` etc by default: they are all standard and widely-used bindings. Perhaps provide a prefix map which contains bindings for `0`..`9` etc., and then suggest that the user should bind a global key to that prefix map.
- Your functions should be called `winum-select-...` rather than `select-window-...`, to match the package name.

deb0ch:
Done, thanks for your feedback ! :smiley_cat: 

I just didn't unbind `M-0`. It would break the logic and convenience of the package.

If the user wants to opt-out of this, how to do it is described in the README with code snippet and examples.

purcell:
> I just didn't unbind M-0. It would break the logic and convenience of the package.

Then you should provide an autoloaded command which sets up those keybindings. You should never override them by default without the user explicitly choosing to do so.

deb0ch:
Can you point me to any documentation / discussion / whatever that would tell me what is allowed and what is forbidden in this standard ?

I couldn't find anything in my google searches, maybe I was using the wrong keywords.

I could consider a prefix keymap, like I did with `M-m`, but with your answer I really don't know what I can and cannot use as a default prefix.

Do you have a solution to propose that would not impair on the convenience of the default bindings ?

deb0ch:
Back in 2013 it was already bound that way in `window-numbering`, but you still integrated it without requiring a change in the bindings:

https://github.com/nschum/window-numbering.el/pull/2
https://github.com/melpa/melpa/commit/ce1dc80f69894736b276885e4ec3ce571a8612c9

What changed since then ?

Keeping bindings similar to the old package has its importance to me, because I would like the change for people who are willing to adopt this new version to be as smooth as possible.

syohex:
There are some byte-compile warning about using variables before declaring them. Could you fix them ?

deb0ch:
@purcell OMG I just realized that `M-<number>` could be used in vanilla emacs to specify prefix args 

:dizzy_face::boom::gun:

Sorry, I just didn't know what I was talking about. Indeed that should absolutely change, I don't even know how it could be that way in the old package.

Thanks for pointing that out, I wouldn't have realized.

Fanael:
And `M-m` is `back-to-indentation`.

deb0ch:
@fanael Thanks you for that insight, didn't know that as well :smile_cat:

@syohex Did you have time to consider my point ? Readability or warnings ?

Waiting for your input to push.

deb0ch:
@purcell Done changing the keybindings. I didn't find anything other than a long and painful `C-x w`, but at least it is not shadowing any of the native Emacs key bindings. I'm giving tips in the README for setting shorter bindings.
This is what you wanted, right ? Or did I misunderstand you ?

@syohex Fixed the warnings. The separation between public and internal stuff is not anymore as clear as I would have wanted it to be, but at least there is no more warnings.

Let me know if more things need to be changed.

Cheers

purcell:
:+1:

PR #4457, opened 2016-12-14
pull request for recipe xah-format-code

xahlee:

command xah-reformat-lines that reformat text from single line to multi-lines, by strictly exchange whitespaces with singe newline char. Similar to fill-region, but more useful for source code.

https://github.com/xahlee/xah-reformat-code

Am author of the package.

None needed

### Checklist

- [*] I've read [CONTRIBUTING.md](https://github.com/melpa/melpa/blob/master/CONTRIBUTING.md)

- [*] I've used [package-lint](https://github.com/purcell/package-lint) to check for packaging issues, and addressed its feedback

( got error
imenu-unavailable-error: imenu unavailable: "This buffer cannot use ‘imenu-default-create-index-function’"
Filed a bug report on that.
https://github.com/purcell/package-lint/issues/49
)

- [*] I've built and installed the package using the instructions in the [README](https://github.com/melpa/melpa/blob/master/README.md)

Thank you.


syohex:
Please use topic branch instead of master branch. There are some unrelated commits.

- Use same package prefix for all exported symbols. There are two prefixes, `xah-replace`, `xah-reformat`

xahlee:
Thanks for the fast response.

Can we consider xah- as the prefix?

For now i'd like to keep the name xah-replace-whitespaces-to-one-space, as that's technically accurate for that command.

About branching, should i close and do a pull request fresh?

Thanks.

purcell:
> Can we consider xah- as the prefix?

I'd advise against this, even though you're only going to conflict with your other packages. It's just cleaner to use a consistent prefix. My suggestion would be to rename the package to `xah-reformat`, and then use `xah-reformat-` as the prefix for all its top-level symbols.

xahlee:
ok. Thanks guys. Closing this and will do a fresh pull request with issues fixed.

PR #4458, opened 2016-12-14
Add guix recipe

alezost:

Interface for GNU Guix package manager

https://github.com/alezost/guix.el

I'm the maintainer.

None needed

### Checklist

### DIFF
--- a/recipes/guix
+++ b/recipes/guix
@@ -1,0 +1,8 @@
+(guix
+ :fetcher github
+ :repo "alezost/guix.el"
+ :files ("elisp/*.el"
+         "doc/*.texi"
+         ("images" "images/*.svg")
+         ("scheme" "scheme/*")
+         (:exclude "scheme/Makefile.am")))


vyp:
Why not put guix.el in http://git.savannah.gnu.org/cgit/guix.git/tree/emacs ?

alezost:
> Why not put guix.el in http://git.savannah.gnu.org/cgit/guix.git/tree/emacs ?

Because "guix.el" is not maintained as a part of Guix anymore.

The story is: in 2014 I wrote this "guix.el", and it was megred into the
Guix repo (the link you posted).  Then it has been maintained inside the
Guix source for all this time, but not anymore.  Now it is a separate
project again, and I think it will be removed from the Guix source tree
soon.

I've made the announcement about it on guix-devel list:
http://lists.gnu.org/archive/html/guix-devel/2016-12/msg00597.html

vyp:
Oooh my bad, I didn't check the lists first! And yes, now I actually remember reading the 'alt-guix' thread where you explained earlier.

alezost:
Oh no, not your bad! You asked before I sent that announcement :-)

purcell:
Thanks. Quick feedback:
- In the recipe, `("scheme" "scheme/*")` can simply be `"scheme"`.
- You have several files relating to `hydra`, but no package dependency on it.

Looks good otherwise.

alezost:
> In the recipe, ("scheme" "scheme/*") can simply be "scheme".

Sorry, I should have written that I tried "scheme", but this leaves
`scheme/Makefile.am` file, which should be excluded.

> You have several files relating to `hydra`, but no package dependency on it.

If you mean Emacs package [hydra](https://github.com/abo-abo/hydra),
then there is no related code.  It happens that the same name is used
for the Guix build farm: https://hydra.gnu.org, and all the
`guix-hydra-...` code is the interface for this Hydra.

alezost:
I have modified the recipe to add "images" directory as well.  Is it OK
to install a couple of images?

These images are (optionally) needed for `M-x guix-about` command I've
just committed.  It displays a fancy splash screen that looks like this:

M-x guix-about

purcell:
Looks good to me, so I'm merging now. (I'll note, however, that not all Emacs are built with the optional SVG support enabled, so the images won't work universally.)

alezost:
Thanks for merging!

And thanks for the note.  It's not a problem: if SVG support is
disabled, the image will not be displayed (`create-image` checks
if an image type is available).

PR #4461, opened 2016-12-15
Add mode for editing JastAdd AST files

rudi:
This mode adds support for JastAdd's AST definition files
(see http://jastadd.org/)

Upstream is at https://github.com/rudi/jastadd-ast-mode

I'm the author of the package.  package-lint and flycheck-package don't seem
to complain.


purcell:
You're using `prog-mode`, so you should add a `;; Package-Requires: ((emacs "24"))` header. I'll go ahead and merge this now though - thanks.

rudi:
Many thanks for the review!  Added the Package-Requires header.

PR #4466, opened 2016-12-18
Add recipe for python-switch-quotes

werehuman:

Cycle between apostrophes and quotes in python strings.

https://github.com/werehuman/python-switch-quotes

Maintainer.

***None needed***

### Checklist

### DIFF
--- a/recipes/python-switch-quotes
+++ b/recipes/python-switch-quotes
@@ -1,4 +1,2 @@
 (python-switch-quotes :fetcher github
-                      :repo "werehuman/python-switch-quotes"
-                      :files ("python-switch-quotes.el"
-                              "README.md"))
+                      :repo "werehuman/python-switch-quotes")
--- a/recipes/python-switch-quotes
+++ b/recipes/python-switch-quotes
@@ -1,0 +1,4 @@
+(python-switch-quotes :fetcher github
+                      :repo "werehuman/python-switch-quotes"
+                      :files ("python-switch-quotes.el"
+                              "README.md"))


syohex:
I have sent PR about minimum Emacs version.
- https://github.com/werehuman/python-switch-quotes/pull/1

recipes/python-switch-quotes
```diff
@@ -0,0 +1,4 @@
+(python-switch-quotes :fetcher github
+                      :repo "werehuman/python-switch-quotes"
+                      :files ("python-switch-quotes.el"
```
syohex:
```lisp
("*.el" "*.el.in" "dir"
 "*.info" "*.texi" "*.texinfo"
```

Could you rename or move test file name and remove `:files` attribute ? Current test file name is not matched against default `:exclude` pattern(above snippet).
- Renaming: `test-python-switch-quotes.el` -> `python-switch-quotes-test.el`
- Moving: `test-python-switch-quotes.el` -> `test/test-python-switch-quotes.el`

And I suppose README.md is not necessary. It is enough by package comment and docstring.

syohex:
Thanks for fixing.

This package assign key binding(`C-c '`) only by loading. This may override user key configuration even if user does not use this package. I propose that

- Remove its code(and users assign key manually)
- Provide minor-mode(default mode map uses `C-c '` key) or setup function

How about ?

werehuman:
I think that the first variant is better. https://github.com/werehuman/python-switch-quotes/commit/ee36b435070cf19d2156c161d57fc18ed73684ac

syohex:
Thanks!!

PR #4468, opened 2016-12-19
Take releases off the "stable" branch

hniksic:
MELPA seems to always use the latest htmlize commit from "master", which is incorrect - those commits can be unstable and are not meant to be used by end users - except contributors and those who want to test the bleeding-edge code.

Added the :branch "stable" to instruct the fetcher to take releases as they appear on the stable branch.

### DIFF
--- a/recipes/htmlize
+++ b/recipes/htmlize
@@ -1,4 +1,5 @@
 (htmlize
  :repo "hniksic/emacs-htmlize"
  :fetcher github
- :version-regexp "release/\\(.*\\)")
+ :version-regexp "release/\\(.*\\)"
+ :branch "stable")


syohex:
There is [melpa-stable](https://stable.melpa.org/#/) for such purpose. melpa-stable provides latest tag version. Such users should use melpa-stable instead of melpa.

milkypostman:
Hrvoje, there are other threads about this which you can look for in the issue tracker. We don't track additional branches in melpa for a number of reasons. I will list a few. 

1. The purpose of melpa is to avoid burdening the author with release process. Having to maintain a stable branch means that the package maintained needs to *release* stable versions. In practice package authors forget to update the stable branch. 

2. MELPA is intended to replace git clones from master. This a stable branch breaks this model. 

3. The master branch should never be "broken" by design. It can happen certainly, but experimental features and partial work should be done in a separate branch then merged into master. Otherwise it breaks the source tree for other developers. This doesn't mean bug free or certainly not stable, but it means a "development" version which is what we want for melpa. 

If you would like to tag stable releases feel free to direct your users to melpa-stable. Mostly that is less advertised because it is less maintained by myself and the other maintainers. I simple don't want stable packages, and am willing to pay the cost of occasional (very very rare in practice) breakages. But stable was designed for this purpose and users who truly want a stable experience.

hniksic:
With due respect, please allow individual project to define what "master" means for them. Otherwise you are simply forcing me to move actual development to a branch of another name, and use "master" to mean "stable". I don't see the point of that exercise.

Why does your fetcher allow arbitrary branch names if you are not going to use them?

Silex:
@hniksic: I think you are free to make MELPA fetch from the `stable` branch if that's still what you want. In your original post you seemed misinformed about what MELPA is (it is unstable) and didn't know about MELPA-stable, so your PR didn't make sense in that regard. If after all this you still want to treat master differently than most people do, I don't see why you shouldn't be able to.

@milkypostman: I think you'd keep the PR open and possibly merge it if @hniksic wants to go through.

That being said, I have two concerns:

- Almost all git workflows treat "master" as something stable, so your workflow is a bit peculiar here.
- It does not look like this repo gets much activity, so I don't quite understand the need for a "stable" branch. Just a thought :wink:

milkypostman:
@hniksic i think "stable" is the wrong term and we should think of it as "not-broken", either way it's not "release".

the question to answer is: if I wanted to collaborate (fork) your project and begin work, which branch should I use?

If the answer is "stable" then you should change your repository to set the default/main branch to stable and melpa will pick that up. this is a git setting, exposed by github, which directs people to the repo owners working branch of choice. If the answer is "master" then things are working as intended.

we allow arbitrary branch names for very specific cases to fix bugs, notice in all 2316 recipes only 14 use this tag and of those that do, the ones that have the use-case you mention here were committed in 2014.

PR #4469, opened 2016-12-20
Add recipe for perspeen

seudut:

This is package is intended to make it much more convenient to work on multiple project (work space) at the same time. Keep each work space has its own window-configuration, ido buffers and the tabs

https://github.com/seudut/perspeen

Maintainer

**None needed**

### Checklist

### DIFF
--- a/recipes/perspeen
+++ b/recipes/perspeen
@@ -1,0 +1,1 @@
+(perspeen :fetcher github :repo "seudut/perspeen")


syohex:
Thanks for PR.

This package uses super(hyper) key prefix by default. However, some super key prefix is often used by OS(system). For example, `s-e` is bound for opening explore on Windows.  `s-t` is bound for opening terminal emulator on some Linux desktop environment. In such environment, there is conflicting key bindings and system key binding is used instead of emacs binding.
And super(hyper) key cannot be used on terminal emacs usually.

I suppose other prefix key should be used rather than super key or use customizable prefix key like elscreen.

seudut:
Thanks @syohex for the suggestion. I have changed the default prefix key as `C-z`, and let user be able to set this.

Silex:
Wow, *another* windows configuration package!

I'll ask the same question I asked in #4148

> Hi! Not related to your PR but concerning your package, I was looking for something like this since a long time. Can you elaborate a little bit the difference with desktop-save-mode and maybe others ?
> 
> Full list of similar packages at https://www.emacswiki.org/emacs/SessionManagement

seudut:
Yes, @Silex , I have planned to write down some stuff to introduce more detail about this . Maybe in the README or maybe another page.

In short, this package is intended to combine some features of `perspective` and `elscreen`, that's why it's named as `perspeen`.  Also the implementation are inspired from these two package.
 
Why I write this package is because of some previous work habit. Before switch to Emacs, I am using vim+tmux+zsh and often working multiple workspaces (projects) at the same time. 
In my work flow, each tmux window is a workspace(project). There is a vim and zsh in the each tmux window. In the vim, it has multiple buffers of current project, it also has multiple tabs and its own window layout.
So each workspace  are separate, and has it own buffers, tabs. Just like the `perspective`, while `perspective` doesn't support multiple tabs.
Each tabs in the same workspace share the same buffer list of current project. And has its own windows layout. just like the `elscreen`.

After using Emacs, I have always been looking for some package for this, as I find the flow (tmux+vim+zsh) are very efficient for me, especially when working multiple projects at the same time.
I have already been using `perspective` and `elscreen` for a long time, but find they still cannot meet my requirement.  So, at last, I decide to try write this for myself. Although it still has some bugs, and sometimes the behaviors are still not as my expected, but the basic feature can work well now.

In a word, `perspective` doesn't support multi-tab (when both `perspective` and `elscreen` enabled , all the project share the same tabs, that's not what I want), but this package does. 
`elscreen` cannot separate the buffers against project (like perspective), but this package can.

Any suggestion are welcome.

PR #4474, opened 2016-12-23
winum: move repo name to 'emacs-winum' to follow conventions

deb0ch:
I realized not long after first submitting the recipe that a widely used convention for naming Emacs packages was to use "emacs-" as a prefix to the package name.

It gives more visiblity when searching on Google and Github, gives a standardized way of finding emacs packages, and IMO is just a better name.

Currently, I can't find even find the package on Google when searching for `github deb0ch emacs winum`.

For the transition to be seamless, here is my strategy:
- I created a new repo `emacs-winum`, which is a mirror of the original repo
- This PR points to that mirror
- when this PR is merged, I will quickly between two `melpa` builds:
  - delete the mirror
  - rename the `winum.el` repo to `emacs-winum`, so that the melpa recipe will still be valid
  - update the file header with the correct address of the repo

I think that it is still time to make this change, as this package release is still very fresh.

### DIFF
--- a/recipes/winum
+++ b/recipes/winum
@@ -1,1 +1,1 @@
-(winum :fetcher github :repo "deb0ch/winum.el")
+(winum :fetcher github :repo "deb0ch/emacs-winum")


purcell:
Why so complicated? Just rename the repo using github, and github will handle the redirection automatically so that MELPA keeps working. Once you've done that, ping me here and I'll merge this PR.

deb0ch:
Oh, I didn't know that GitHub would handle such redirections, so convenient ! :open_mouth: 

Ok, I just renamed it, you can go ahead :smile_cat:

PR #4475, opened 2016-12-24
add company-erlang recipe

s-kostyaev:

`company-erlang` is company backend for erlang based on `ivy-erlang-complete`.

https://github.com/s-kostyaev/company-erlang

I am the author.

### Checklist

### DIFF
--- a/recipes/company-erlang
+++ b/recipes/company-erlang
@@ -1,0 +1,2 @@
+(company-erlang :repo "s-kostyaev/company-erlang" :fetcher github
+                :files (:defaults))


syohex:
- Could you remove pkg file from repository ? It is enough to use package header of `company-erlang.el`.

s-kostyaev:
@syohex done.

syohex:
Merged

s-kostyaev:
Thanks

purcell:
I'd suggest you apply version tags to both this and `ivy-erlang-complete`, so that you can depend on a non-snapshot version of the latter from this package, and so that both get included in MELPA Stable.

s-kostyaev:
I think it's good idea. Where can I read how to do this?

purcell:
Just add git tags, e.g. "0.1" or "v1.2", then use those numbers (rather than snapshot version numbers) when specifying dependencies in `Package-Requires:` headers. Packages for which there are version tags are automatically included in MELPA Stable, without any further steps being necessary on your part.

s-kostyaev:
@purcell thanks for clarifying. Done.

PR #4479, opened 2016-12-26
Add rust-playground recipe

grafov:

GNU/Emacs mode that setup local playground for code snippets in Rust
language similar to play.rust-lang.org.

https://github.com/grafov/rust-playground

I am author of the package.

### Checklist

### DIFF
--- a/recipes/rust-playground
+++ b/recipes/rust-playground
@@ -1,0 +1,1 @@
+(rust-playground :repo "grafov/rust-playground" :fetcher github)


syohex:
- Could you put `Version:` header ?

grafov:
I thought this header should be added by MELPA build scripts. My another plugin `go-playground` (already passed to MELPA) has not this header too. Sure I will add the header to both packages.

syohex:
BTW `package-lint` says.

```
1:1: error: package.el cannot parse this buffer: Package lacks a "Version" or "Package-Version" header
1:1: warning: "Version:" or "Package-Version:" header is missing. MELPA will handle this, but other archives will not.
```

grafov:
Yes it reported this ...but I ignored the warning because of my previous package has same notice but  was passed to MELPA successfully :)

grafov:
@syohex I just added `Version` to the package header.

syohex:
Thanks

purcell:
Some belated feedback:
- You should probably add a `defgroup rust-playground` in which to place your custom items -- that group can, of course, live inside the `rust-mode` group.
- You're making command lines by concatenating file and directory paths which might contain spaces or other special characters. Use `shell-quote-argument` to avoid this.

grafov:
Thank for your feedback.

I already done the subgroup in previous commits (see line 50) as:

```elisp
 (defgroup rust-playground nil
  "Options specific to Rust Playground."
  :group 'rust-mode)
```

Shell arg for compiler call string fixed just now.

purcell:
Awesome.

PR #4480, opened 2016-12-29
Add multiple elisp modules and banner files

rakanalh:
Dashboard is an extraction of Spacemacs-buffer into a standalone package.

This modification enables the packaging of multiple lisp files and multiple banner files that the new implementation is providing. The code base has not been merged into master as i don't want to break installations for existing users. However, the new implementation exists in the following PR: https://github.com/rakanalh/emacs-dashboard/pull/29

Once this has been approved and merged, i will merge my change into master so that we're in sync.

https://github.com/rakanalh/emacs-dashboard

I am the maintainer

None needed

### Checklist

### DIFF
--- a/recipes/dashboard
+++ b/recipes/dashboard
@@ -1,3 +1,5 @@
 (dashboard :fetcher github
 	   :repo "rakanalh/emacs-dashboard"
-	   :files ("dashboard.el" "banner.txt"))
+	   :files ("dashboard*.el"
+		   "banners/*.png"
+		   "banners/*.txt"))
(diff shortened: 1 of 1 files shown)


purcell:
Thanks. This change will cause the `.txt` and `.png` files to be copied to the root directory of the resulting package. Is that what you intended? I would suggest that's not the best choice. If you simply want to preserve the contents of the `banners` subdirectory inside the resulting package, a single `:files` entry of `"banners"` would be sufficient.

rakanalh:
You're right, it wasn't the intended behaviour. I've modified the PR to your comment.

Thanks alot

purcell:
Thanks. I committed a different version of this change, in e9a7934, because there's a better way to include multiple "*.el" files.

PR #4481, opened 2016-12-30
adds the github theme

tx46:

The package provides the GitHub color theme, trying to stay true to the theme used for syntax highlighting on the GitHub website.

https://github.com/xxxxx/emacs-github-theme

Maintainer

**None needed*

### Checklist

### DIFF
--- a/recipes/github-theme
+++ b/recipes/github-theme
@@ -1,0 +1,3 @@
+(github-theme
+ :repo "philiparvidsson/emacs-github-theme"
+ :fetcher github)


purcell:
Thanks. Quick feedback:
- Please remove "for Emacs" from the package description: all Emacs packages are for Emacs, so this wording is redundant. `package-lint` would have warned you about this, so please consider using it in future.
- You should remove `;; eval: (when (require 'rainbow-mode nil t) (rainbow-mode 1))` from the local variables. It's an unsafe variable, and if you like to use rainbow-mode when editing themes, a better way to do that is with code in your emacs config [like this](https://github.com/purcell/emacs.d/blob/38bbef8ad964474930aa3219c81f96852c384767/lisp/init-lisp.el#L250-L261).

Looks good otherwise.

philiparvidsson:
Should be fixed now.

purcell:
Thanks!

PR #4483, opened 2016-12-30
Add info-buffer recipe

llvilanova:

Interactive command (info-display-manual-in-buffer) for showing each info topic on its own buffer. With prefix, always creates a new buffer even if the topic is already opened.

https://github.com/llvilanova/info-buffer

Author & maintainer

None needed

### Checklist

### DIFF
--- a/recipes/info-buffer
+++ b/recipes/info-buffer
@@ -1,0 +1,3 @@
+(info-buffer
+ :fetcher github
+ :repo "llvilanova/info-buffer")


syohex:
Looks good.

purcell:
Thanks - looks good to me. I've got a couple of quick comments, but I'll go ahead and merge the recipe in the meantime:
- The prompt for `info` says `"Info file name: "`, but you're using `"Manual name: "` -- surely these should match?
- `Display info topics on separate buffers` should be `Display info topics in separate buffers` :-)

llvilanova:
Quick one, how do I manage new versions of the package? Do you detect new tags or should I send a new pull request?

Also, I'm not sure about the prompt. C-u M-x info asks specifically for a file, while info-buffer asks for the name of a registered "info manual", not a file.

syohex:
> Quick one, how do I manage new versions of the package? Do you detect new tags or should I send a new pull request?

You update `info-buffer` repository then melpa updates after some hours. (You add new tag, then melpa-stable updates it too). You need not to send pull request again.

PR #4485, opened 2017-01-01
Add rg.el recipe

dajva:

This is an Emacs frontend to grep/ag replacement ripgrep. It mimics (and reuses) the Emacs builtin rgrep command.

https://github.com/dajva/rg.el

Maintainer

**None needed**

### Checklist

### DIFF
--- a/recipes/rg
+++ b/recipes/rg
@@ -1,4 +1,3 @@
 (rg
  :fetcher github
- :repo "dajva/rg.el"
- :branch "melpa")
+ :repo "dajva/rg.el")
--- a/recipes/rg
+++ b/recipes/rg
@@ -1,0 +1,4 @@
+(rg
+ :fetcher github
+ :repo "dajva/rg.el"
+ :branch "melpa")


syohex:
What is different between this package and [ripgrep](https://github.com/nlamirault/ripgrep.el) package ?

dajva:
The main difference is that this package has an interface to rg's file type filtering. This package's interface and features is also essentially the same as Emacs builtin `rgrep` which provides a seamless transition. I personally is used to `rgrep` and prefer its interface and features.

dajva:
Another benefit with `rg.el` is that grep addon packages like wgrep (I have not tried others) works out of the box. I have not really looked into `ripgrep.el` that much and since I am not really an elisp programmer I can't judge if it would be easy to incoroporate all this into `ripgrep.el`.

purcell:
I'm open to merging this. Some quick feedback re. the code:
- [This whole block](https://github.com/dajva/rg.el/blob/d9d4899dc7a87f5095b0794a8834ebcd0b49add1/rg.el#L109-L121) is probably just a single `cl-find` with a lambda. Or at the very least, a `cl-dolist` with a `cl-return`. Require `cl-lib` for that, and declare a package dependency on `(cl-lib "0.5")`.
- You're missing a `defgroup` for `'rg`, but you're using that group in your `defcustom`
- ` from within Emacs` in the package description is redundant and should be removed. All elisp code is for doing stuff from within Emacs! `package-lint.el` should have pointed that out to you when you ran it...
- If this is meant to work exactly like `rgrep`, you might consider calling the package `rgrep-rg` instead.
- The indentation is all weird and unconventional. `C-x h` followed by `M-C-\` would fix that! :-)

dajva:
Thanks for your feedback.

I have fixed the issues now. I did not rename the package since the mapping between `rg.el` and `rgrep` is not an exact match. The main difference is the type alias system which needs to follow `rg`'s internals to make sense.

https://github.com/dajva/rg.el/compare/d9d4899dc7a87f5095b0794a8834ebcd0b49add1...melpa

syohex:
```
rg --no-heading --color always --colors match:fg:red --type-add 'gn:*.gn' --type-add 'gn:*.gni' --type-add 'g\
yp:*.gyp' --type-add 'gyp:*.gypi' --type all --color -i when
error: The argument '--color <WHEN>' was provided more than once, but cannot be used multiple times
```

I got above error. If you specify `--colors` option by default(by `rg-command` variable), I suppose you should not pass `<C>` placeholder to `grep-expand-template`(`<C>` placeholder is expanded to `--color -i`)

dajva:
I have not tested this in Emacs 25.1 where the `--color` flag was moved from `grep-options` to `<C>` placeholder. I will work on a fix.

dajva:
Fix is done now.

recipes/rg
```diff
@@ -0,0 +1,4 @@
+(rg
+ :fetcher github
+ :repo "dajva/rg.el"
+ :branch "melpa")
```
syohex: I suppose it is better to use `git tag` and melpa-stable rather than own branch for stable release.

purcell: Agreed. We won't pull off a branch like that, because you might forget to update it. MELPA wants to build packages which automatically track your ongoing development, and you can apply version tags to control what "release" packages get built in MELPA Stable.

dajva: Oh, ok. Seems a bit backwards to me but sure, I'll change it.
The question is to what? Is `master` always assumed to be a development branch or what do you expect there?

purcell:
Great, thanks.

PR #4486, opened 2017-01-02
Add untitled-new-buffer recipe

zonuexe:

Open untitled new buffer like other text editors in Emacs

https://github.com/zonuexe/untitled-new-buffer.el

I'm author.

None needed

### Checklist

### DIFF
--- a/recipes/untitled-new-buffer
+++ b/recipes/untitled-new-buffer
@@ -1,0 +1,1 @@
+(untitled-new-buffer :fetcher github :repo "zonuexe/untitled-new-buffer.el")


syohex:
What is different between [scratch-el](https://github.com/cbbrowne/scratch-el) and this package ? There are some similar packages in MELPA(Please search `scratch` or similar word in https://melpa.org/#/)

tarsius:
While it would be nice for the author to answer that question here, in the README, and the library Commentary section, https://www.reddit.com/r/emacs/comments/5lna4d/open_untitled_new_buffer_like_other_text_editors/ gives some insights. As I understand it `scratch.el` provide fancy features a long-time Emacs users might wish for, while `untitled-new-buffer` is for new Emacs users who find it strange that new "files" have to be named before being saved for the first time.

PR #4487, opened 2017-01-03
Add ivy-youtube recipe

squiter:

This package is a `ivy` version of `helm-youtube`. It uses `ivy` to list youtube videos and launch the browser to watch selected video.

https://github.com/squiter/ivy-youtube

I'm the author.

None needed.

### Checklist

### DIFF
--- a/recipes/ivy-youtube
+++ b/recipes/ivy-youtube
@@ -1,0 +1,1 @@
+(ivy-youtube :fetcher github :repo "squiter/ivy-youtube")


syohex:
Looks good

tarsius:
@syohex Weren't you given commit access and the permission to merge if it "Looks good"? Just asking because I am currently going through the list of open pull requests, adding `awaiting upstream` and otherwise pinging Steven.

@purcell If done so by "asking for a review" from you (using the Reviewers pull-down on the right). Is it sufficiently clear what I mean when I do that? Maybe we could also add a `lgtm` tag -- kinda the opposite of `awaiting upstream`.

purcell:
>  If done so by "asking for a review" from you

Works for me, but probably a "lgtm" tag would be less specific to me, and therefore better in the long run.

Regarding the package itself, here's my feedback:
- Don't depend on `swiper` -- depend on `ivy`, instead.

Looks good otherwise!

squiter:
@purcell you are right! I fixed this dependency problem, thanks.

syohex:
Merged

PR #4489, opened 2017-01-05
adding `buffer-manage` recipe

plandes:

Provides support to multiple managed buffers (i.e. inferior shells, SQL sessions, etc.).

https://github.com/plandes/buffer-manage

Original author and maintainer.

This is the second submission, this time from a topic branch created from an earlier commit before the merges (sorry for any extra work that created!).  Thanks again.

### Checklist

### DIFF
--- a/recipes/buffer-manage
+++ b/recipes/buffer-manage
@@ -1,1 +1,1 @@
-(buffer-manage :fetcher github :repo "plandes/buffer-manage" :files ("*.el"))
+(buffer-manage :fetcher github :repo "plandes/buffer-manage")
--- a/recipes/buffer-manage
+++ b/recipes/buffer-manage
@@ -1,1 +1,1 @@
-(buffer-manage :fetcher github :repo "plandes/buffer-manage" :files ("lisp/*.el"))
+(buffer-manage :fetcher github :repo "plandes/buffer-manage" :files ("*.el"))
--- a/recipes/buffer-manage
+++ b/recipes/buffer-manage
@@ -1,0 +1,1 @@
+(buffer-manage :fetcher github :repo "plandes/buffer-manage" :files ("lisp/*.el"))


syohex:
This package requires Emacs 24 or higher versions. However there are some warnings at loading on Emacs  24.3 or lower versions. Is this no problem ? And there are some byte-compile warnings too.

```
buffer-manage.el:Warning: `(entries (slot-value this (quote entries)))' is a
    malformed function
buffer-manage.el:Warning: `(last-switched-to (slot-value this (quote
    last-switched-to)))' is a malformed function
buffer-manage.el:Warning: `(entries (slot-value this (quote entries)))' is a
    malformed function
buffer-manage.el:Warning: `(last-switched-to (slot-value this (quote
    last-switched-to)))' is a malformed function
buffer-manage.el:Warning: `(include-frame-wins (slot-value this (quote
    include-frame-wins)))' is a malformed function
buffer-manage.el:Warning: `(read-history (slot-value this (quote
    read-history)))' is a malformed function
buffer-manage.el:Warning: `(entries (slot-value this (quote entries)))' is a
    malformed function
buffer-manage.el:Warning: `(last-switched-to (slot-value this (quote
    last-switched-to)))' is a malformed function
buffer-manage.el:Warning: `(entries (slot-value this (quote entries)))' is a
    malformed function
buffer-manage.el:Warning: `(last-switched-to (slot-value this (quote
    last-switched-to)))' is a malformed function
buffer-manage.el:Warning: `(include-frame-wins (slot-value this (quote
    include-frame-wins)))' is a malformed function
buffer-manage.el:Warning: `(read-history (slot-value this (quote
    read-history)))' is a malformed function
```

plandes:
It uses eieio heavily, which has changed a lot across 24 versions.  I am on 24.5 and have no problem fixing it to that.  In that case how do I enforce that?

Bonus question: What debian packages do I use to load 24.5?  I have another package that is failing on a Travis build because it's using an early version of 24 (I think 24.0).  Any ideas?

Thanks!

plandes:
Disregard the last comment.

I have moved the build to be more like the MELPA build and I am no longer getting the warnings on my travis build.  Also I have committed a new change on my MELPA forked repo as the location of the lisp files has changed to the root.  Let me know if I need to need to create a new pull request for that.

recipes/buffer-manage
```diff
@@ -0,0 +1 @@
+(buffer-manage :fetcher github :repo "plandes/buffer-manage" :files ("*.el"))
```
syohex: It is not necessary to specify `:files` attribute because it is enough to use default value. Please see https://github.com/melpa/melpa#recipe-format

tarsius:
> I have moved the build to be more like the MELPA build and I am no longer getting the warnings on my travis build.

Those that mean that this package now works on all 24 versions or that you now use 24.5 for testing on Travis? IMHO it would be fine to depend on >=24.5 and be done with it.

plandes:
@tarsius Thank for the input and I agree--we'll all be (or already moving) to 25 anyway.  I have made 24.5 the dependency in the elisp file and the travis build and everything checks out.  I have also made a bug fix.

@syohex Thanks for making me aware of that compliance detail.  That change has been made and pushed to my melpa fork.

It appears ready to ship.

Thanks all!

tarsius:
You might even want to depend on 25! `eieio` has again changed considerably in that release. I wasn't aware that it changed much during the 24 period, but from 24.* to 25 it changed a lot, and for the better.

plandes:
@tarsius Good point!  I will do that but after I move to 25 myself :)  I'll be moving to 25 soon and then go back and get everything 25 ready.

plandes:
@syohex There was one last change I just committed.  However, everything should check out.  Let me know if there are any other changes you want to make.

plandes:
Will someone please tell me if there is something else I need to do to get this merged?

Do you want me to move to Emacs 25 or just be more patient :)

plandes:
I have updated this Emacs 25

syohex:
Merged

plandes:
Thanks!

PR #4491, opened 2017-01-07
add helm-etags-plus recipe

jixiuf:
Another Etags helm.el interface, it support multiple tag files.

https://github.com/jixiuf/helm-etags-plus

I am the maintainer

### DIFF
--- a/recipes/ctags-update
+++ b/recipes/ctags-update
@@ -1,3 +1,2 @@
-(ctags-update :repo "jixiuf/helm-etags-plus"
-              :fetcher github
-              :files ("ctags-update.el"))
+(ctags-update :repo "jixiuf/ctags-update"
+              :fetcher github)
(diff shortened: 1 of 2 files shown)


syohex:
- You should name `helm-etags-plus.el` instead of `helm-etags+.el` if you name package `helm-etags-plus`. (Tests on travis is failed by this naming issue).
- I suppose `ctags-update.el` should be moved to other repository. (And then you need not to specify `:files` attribute in recipe.

purcell:
Some quick feedback:
- Top-level symbols like `pos-in-same-symbol-p` are not allowed: all such symbols should have the `helm-etags-plus-` prefix.
- Similarly, symbols like `helm-c-source-etags-plus-select` should probably be of the form `helm-etags-plus-c-source-select` instead.
- `M-x checkdoc` would have some advice on tidying up some of those docstrings.

Looks good otherwise.

jixiuf:
updated

purcell:
Super, thanks - merged!

PR #4495, opened 2017-01-08
ctags-update: get code from new repo

mdorman:
Per comment in https://github.com/jixiuf/helm-etags-plus/commit/beac9f54d1c26e19b59b009fb26fd3d78688fa82


syohex:
This is already updated by #4491

PR #4500, opened 2017-01-13
Add lazily recipe

justbur:

lazily provides a macro to execute forms in a config file while catching any
void-variable or void-function errors. The bad forms are stored and retried
after every new load using after-load-functions.

https://github.com/justbur/emacs-lazily

Author/maintainer

**None needed**

### Checklist

### DIFF
--- a/recipes/lazily
+++ b/recipes/lazily
@@ -1,0 +1,1 @@
+(lazily :repo "justbur/emacs-lazily" :fetcher github)


purcell:
Interesting. Seems to me that this is likely to result in forms being executed over and over again in quite unpredictable ways at startup, thus slowing everything down. And any typos which result in legitimate `void-variable` errors etc. will never be surfaced to the user. I'm quite wary of promoting this approach: once on MELPA this package will be found by Emacs newbies and they won't realise the pitfalls.

justbur:
> Interesting. Seems to me that this is likely to result in forms being executed over and over again in quite unpredictable ways at startup, thus slowing everything down. 

I've set it up to check if for example there is a void-variable error that the variable is defined before the form is tried again to avoid this issue. I've also added two variants, `lazily-do` and `lazily-do-all`. One stops as soon as any error is found and stores all remaining forms. The `-all` version keeps going. 

> And any typos which result in legitimate void-variable errors etc. will never be surfaced to the user. I'm quite wary of promoting this approach: once on MELPA this package will be found by Emacs newbies and they won't realise the pitfalls.

Fair enough. This is an experiment as far as I'm concerned. Agreed that typos are one of the main issues. I added a function that produces a report of everything found. If you're not comfortable with it on MELPA, I have no problem with that.

PR #4501, opened 2017-01-13
adding recipe bshell

plandes:

Manage multiple shells in Emacs buffers

https://github.com/plandes/bshell

Contributor and maintainer.

None

### Checklist

### DIFF
--- a/recipes/bshell
+++ b/recipes/bshell
@@ -1,0 +1,1 @@
+(bshell :fetcher github :repo "plandes/bshell")


purcell:
Thanks. Quick feedback:
- The package description ("interactive object oriented shell") seems unconnected to what this package actually does.
- Certain keybindings you're using (e.g. `"C-c g"`, `"C-c r"`) are [reserved for end user customisation](https://www.gnu.org/software/emacs/manual/html_node/elisp/Key-Binding-Conventions.html), so you should choose different keybindings which fit the conventions
- The `PATH` and `MANPATH` munging seems tangential to the primary goal of the package, so it seems out of place and likely to break things for some users.
- Instead of `(if (not COND) FORM)`, prefer `(unless COND FORM)`
- Re. the code which looks for `~`: would the standard function `abbreviate-file-name` help here?
- You've defined the `bshell` custom group as being a child of itself, which is invalid. Please instead choose one of the standard top-level groups.
- The `Commentary` text says what's in the package, but not really how to use it. A little basic usage info would be desirable there, since the docstrings elsewhere are pretty sparse.

Hope that helps!

plandes:
Steve:

Thanks for the feedback and I agree on all points.  I have implemented all suggestions and committed: https://github.com/plandes/bshell/commit/58210b68e808433cb8f36f52c500806954a44379

purcell:
Thanks!

PR #4504, opened 2017-01-14
add recipe eldoc-overlay-mode

stardiviner:

https://github.com/stardiviner/eldoc-overlay-mode

maintainer

**None needed***

### Checklist

### DIFF
--- a/recipes/eldoc-overlay-mode
+++ b/recipes/eldoc-overlay-mode
@@ -1,0 +1,3 @@
+(eldoc-overlay-mode
+  :fetcher github
+  :repo "stardiviner/eldoc-overlay-mode")


syohex:
- No autoload cookie
- `eldoc-message-function` was introduced at Emacs 24.4. But this package requires Emacs 24.3 or higher versions
- Minor mode lighter should start with space
- URL repository is wrong

purcell:
Additionally, don't define a face with only one of `:background` and `:foreground`. You should always provide both, and ideally two versions for dark and light background classes. Often the best choice is to `:inherit` a built-in face by default, then the user (or themes) can override the face as desired.

stardiviner:
I have updated my code, fixed all of upper issues. Thanks for your reviews.

syohex:
Merged

PR #4505, opened 2017-01-15
added recipe github-pullrequest

jakoblind:

Emacs package to smoothly create and checkout Github pull requests. Uses the Github access token for authorization to Github.

https://github.com/jakoblind/github-pullrequest

I am creator/maintainer

**None needed**

### Checklist

### DIFF
--- a/recipes/github-pullrequest
+++ b/recipes/github-pullrequest
@@ -1,0 +1,1 @@
+(github-pullrequest :fetcher github :repo "jakoblind/github-pullrequest")


syohex:
What is different between this package and [magit-ph-pulls](https://github.com/sigma/magit-gh-pulls) ? It looks magit-ph-pulls provides more features.

Review comment
- No autoload cookies
- Package dependencies are not correct(https://github.com/jakoblind/github-pullrequest/pull/12)

jakoblind:
`magit-ph-pulls` does not support Github access token as authorization. Therefore it is not possible to use with accounts that uses 2f auth.

syohex:
I see.

syohex:
Merged.

purcell:
> magit-ph-pulls does not support Github access token as authorization. Therefore it is not possible to use with accounts that uses 2f auth.

Maybe there's an opportunity to submit this behaviour as an enhancement to that package, though? Seems a shame to duplicate effort in different directions.

jakoblind:
@purcell yes that was my initial thoughts. I worked for a few hours but had to give up because the code for  `magit-ph-pulls` (and `gh` which is uses) was way too complex. I read about others having similar experiences working with that code base. 

So my conclusion was to rewrite it from scratch with code readability as one of the top prioritization.

PR #4513, opened 2017-01-21
Add hacker-typer recipe

therockmandolinist:

For amusement, implements a customizable version of http://hackertyper.com in emacs.

https://github.com/therockmandolinist/emacs-hacker-typer

Maintainer

**None needed**

### Checklist

### Note

This is my first package and I consider myself fairly new to emacs-lisp, I understand (and bear with me) if it's riddled with errors or terrible ways of doing things.

### DIFF
--- a/recipes/hacker-typer
+++ b/recipes/hacker-typer
@@ -1,0 +1,1 @@
+(hacker-typer :fetcher github :repo "therockmandolinist/emacs-hacker-typer")


syohex:
I have sent PR with small fix. https://github.com/therockmandolinist/emacs-hacker-typer/pull/1

therockmandolinist:
Merged, thanks!

I'm not sure if this is the right place to ask, but I have one question/hesitation about my package. One of the functions sets `inhibit-redisplay` to `t` while it opens a file and removes comments from it. Is it possible that this isn't the recommended way to do this and I should call an async process to do this externally instead? In case there's user input during the process leading to unexpected issues, or something like that.

syohex:
[note] `file-name-case-insensitive-p` is very newer API. It is implemented only in development Emacs, in other word stable Emacs users cannot use it.

therockmandolinist:
Thanks for the heads up. I took that directly from the code for set-auto-mode. I've updated it to just include the try for case-sensitive and fallback to case-insensitive without using `file-name-case-insensitive-p`.

purcell:
Thanks. `hacker-typer for emacs` isn't a great package description: can you find something better please? "for emacs" is redundant, and the description doesn't tell a potential user what the package actually does.

therockmandolinist:
@purcell that's a fair point. how specific/technical does it have to be? how about something like: "Pretend to write code like a pro"

purcell:
I like it!

therockmandolinist:
@purcell thanks for the feedback, I've committed the changes. 

Also, a question for future reference: if a later version of the package requires async, for example - would I just add the appropriate info to the `;; Package Requires` line?

purcell:
> Also, a question for future reference: if a later version of the package requires async, for example - would I just add the appropriate info to the ;; Package Requires line?

Yes, exactly. Merging now - thanks!

PR #4520, opened 2017-01-25
add org-parser.

zck:

This package parses an org file into a list of structure objects, so it can be programmatically worked with.

https://bitbucket.org/zck/org-parser.el

I'm the creator/maintainer

None needed

### Checklist

- [ ] I've used [package-lint](https://github.com/purcell/package-lint) to check for packaging issues, and addressed its feedback -- All errors fixed but some naming recommendations.

### DIFF
--- a/recipes/org-parser
+++ b/recipes/org-parser
@@ -1,0 +1,1 @@
+(org-parser :fetcher bitbucket :repo "zck/org-parser.el")


syohex:
- `HomePage` header is invalid
- No documentation
- No license
- No package dependencies header
- No autoload cookie
- Use `-` separator instead of `-` for emacs lisp coding convention

zck:
I'll look at these things.

If they're required (specifically the middle four things), might I suggest they be added to CONTRIBUTING.md?

syohex:
They are written in Emacs Lisp convention. And CONTRIBUTING.md refers it as `The package should follow the conventions of Emacs Lisp`.(https://www.gnu.org/software/emacs/manual/html_node/elisp/Tips.html)

zck:
I made a bunch of changes. I think the "package dependencies header" is what's referred to as package-version [here](https://www.gnu.org/software/emacs/manual/html_node/elisp/Library-Headers.html#Library-Headers)?

syohex:
It is `Package-Requires`. And it is still wrong. `seq.el` was bundled since Emacs 25, Emacs 24.5 and 24.5 does not bundle seq.el

zck:
Updated to require emacs 25.1.

purcell:
Thanks. Looks useful! Some quick feedback:
- `org-parser/parse-buffer` etc. should use the `org-parser-` prefix, not `org-parser/`, as per the elisp community conventions. You might prefer the other naming scheme, but the elisp conventions are there to allow the wider community to more easily share and contribute to each other's code, which is more important than personal preferences. Use `org-parser--` for package-private functions.
- Similarly, `title-line?` would more conventionally be `title-line-p` in elisp

Cheers!

tarsius:
What does this offer over `org-element.el`, which is part of Org?

zck:
> org-parser/parse-buffer etc. should use the org-parser- prefix, not org-parser/, as per the elisp community conventions. 

I sense I'm fighting a very losing battle, but I find this to cause confusion, especially for packages like this one that don't have an obvious one-word prefix. And, when I've been looking at code, the downside of having it be unreadable is larger than the upside of following the conventions. And (although not the majority) there are already things in org-mode and shipped with Emacs with this style; for example, org-plot. Similarly, things that are already on melpa use this scheme: for example yasnippet and eclim.

> What does this offer over `org-element.el`, which is part of Org?

I find org-element to be rather difficult to use. Also, results are overly tied to the buffer the data came from. For example, there's a lot of information in an `org-element` list about where the blocks lived in the original source file. In `org-parser`, not only is the information more relevant to what you want, you can also programatically rearrange elements, and output those rearranged elements back out as an org file. I'm not arguing that it's superior for all use cases, just that I want to use it for things I'm writing, which I also want to distribute on melpa.

purcell:
Yep, there's already some poor and inconsistent naming among packages in MELPA, and some of those are lost causes because they have become too popular to easily fix. For new packages, though, we have an opportunity to be consistent and community-friendly, and should do so. (And note that I was a Clojure guy and broadly prefer those naming conventions, so I understand where you're coming from, but I stick to the official conventions for elisp nonetheless.)

Re. `org-element`, I wonder if there's an opportunity to provide an upstream patch to that library? Making that more suited for a variety of use cases might be more valuable and effective than adding a new library... No idea if that's practical.

zck:
Sigh. Updated the code.

purcell:
> Sigh. Updated the code.

Super, merged - thanks!

zck:
Thanks!

PR #4524, opened 2017-01-27
Add fuff

joelmo:

Provides a command (fuff-find-file) that lists files with find in completing read. Also provides a hook for ido that switches to fuff-find-file from ido-find-file, when a starting point directory is entered. These starting points can be customized.

https://github.com/joelmo/fuff

I wrote the package. I did not find any of these to suite my needs: projectile, find-file-in-project, find-file-in-repository. Fuff is similar to these, biggest difference it the ido switch. 

If there is any problems with this package, I will fix it...

### Checklist

### DIFF
--- a/recipes/fuff
+++ b/recipes/fuff
@@ -1,0 +1,1 @@
+(fuff :repo "joelmo/fuff" :fetcher github)


purcell:
Thanks. Some feedback:
- Why not blacklist files rather than whitelist them? Seems like a better option, and there are built-in blacklists you can re-use.
- It feels like none of this code should be specific to `ido` -- just use `completing-read`, which will be overridden by `ido-everywhere`/`ido-ubiquitous`/`ivy`/`helm` according to the user's preferences.
- Don't autoload `(add-hook 'ido-setup-hook 'fuff-ido-setup)`
- You need to declare a package dependency on `seq`, which is new in Emacs 25.1 but available in ELPA for older Emacs versions.
- Overall, this seems like a non-caching and less flexible alternative to `projectile`'s `projectile-find-file`, and it's not entirely clear that it is going to be useful to a wider audience.

purcell:
I re-read your description, and understand better what you're aiming for, so please disregard the note about this vs. projectile.

joelmo:
I didn't find any built-in blacklists that seemed suitable, if you can give an example I can look at it. I thought about using `auto-mode-alist`. But then I would still want to have an option to filter out modes (image-mode) with patterns that shouldn't be displayed. And the pattern have to be modified for find.
I will fix the other points.

purcell:
Check out `grep-find-ignored-files` and `grep-find-ignored-directories`.

joelmo:
Fuff is now using those. The whitelist is removed but I will support it if anyone needs.

purcell:
Super, thanks.

PR #4527, opened 2017-01-28
Remove rustfmt recipe

Wilfred:
This package has been deprecated by the author as the functionality is
now part of rust-mode.


purcell:
Thanks!

PR #4529, opened 2017-01-29
Added the emacs-dokuwiki recipe.

accidentalrebel:

Provides a way to edit a remote Dokuwiki wiki on Emacs.  Uses Dokuwiki's XML-RPC API.

https://github.com/accidentalrebel/emacs-dokuwiki

**None needed**

### Checklist

### DIFF
--- a/recipes/dokuwiki
+++ b/recipes/dokuwiki
@@ -1,3 +1,3 @@
-(emacs-dokuwiki
+(dokuwiki
  :fetcher github
  :repo "accidentalrebel/emacs-dokuwiki")
--- a/recipes/emacs-dokuwiki
+++ b/recipes/emacs-dokuwiki
@@ -1,0 +1,3 @@
+(emacs-dokuwiki
+ :fetcher github
+ :repo "accidentalrebel/emacs-dokuwiki")


syohex:
- I suppose package name should be `dokuwiki` instead of `emacs-dokuwiki`.
- No autoload cookies.

accidentalrebel:
Hi there. I am unfamiliar with the accepted conventions on emacs package names. I can change if needed.

I am sorry I am unfamiliar with autoload cookies. Allow me to research on this for a bit. Can you point me in the right direction?

Thanks!

purcell:
Thanks for this.

> Hi there. I am unfamiliar with the accepted conventions on emacs package names. I can change if needed.

Yes please. The convention is that the "emacs-" prefix is redundant because all emacs packages are for emacs, so it is undesirable. Rename the `.el` file, the recipe, and the contents of both accordingly. (ie. replace all `emacs-dokuwiki` with `dokuwiki`.

> I am sorry I am unfamiliar with autoload cookies. Allow me to research on this for a bit. Can you point me in the right direction?

See the [emacs lisp manual page on this](https://www.gnu.org/software/emacs/manual/html_node/elisp/Autoload.html). Usually it just means adding `;;;###autoload` on the line before the main `interactive` commands that someone might use to trigger start using your package.

More feedback:

- Don't include `Emacs-dokuwiki: ` in your messages: just write the message, not which code generated it.
- `nil` is the only "false" value in Lisp. So `(if (eq (string-match-p ".dwiki" (buffer-name)) nil) ...` should be written `(if (not (string-match-p ".dwiki" (buffer-name))) ...`, or just write this as a guard clause using `(unless (string-match-p ".dwiki" (buffer-name))) (error ...))`
- Similarly, `(when (not (eq page-content nil)) ...)` should be written (when page-content ...)` 
- Same with `(if (not (equal emacs-dokuwiki-xml-rpc-url nil))` etc.
- Don't use `(interactive)` and `read-string` separately: just use `(interactive "sEnter page name: ")`, and add `page-name` to the function args. Look at examples of code which use `interactive` to see what I mean.
- Use `defcustom` and `defgroup` for customisable vars like `emacs-dokuwiki-login-user-name`.
- You should probably write a minor mode for the local buffers, and enable it automatically in those buffers so that `C-x C-s` will trigger your `save-page` function.

Hope that helps!

accidentalrebel:
Hi there, @purcell. I appreciate all your comments and suggestions.

I've implemented almost all of these. Just need to test it out thoroughly. Will update the new changes when I'm done.

Thanks!

accidentalrebel:
Updated the pull request with lots of changes: https://github.com/melpa/melpa/pull/4529/commits/4293d464d14790c75c59a2deda76a3ce3827c1eb

Changes include:
- Changed package from `emac-dokuwiki` to `dokuwiki`
- Added an autoload cookie
- Removed unnecessary prefix to the message logs
- Refactored conditionals
- Made full use of interactive forms instead of `read-string` whenever applicable
- Made use of `defcustom` and `defgroup` for customization variables

### Checklist

purcell:
Thanks, nice work! Merging now. Note that `(defun dokuwiki-open-page(page-name-or-url)` should be written `(defun dokuwiki-open-page (page-name-or-url)` -- note the space before the arg list.

accidentalrebel:
Hi @purcell 

Ah I see. After adding a space before the arg list, flycheck-package showed me other errors to fix. It seems I need to add a space after every function declaration name. Will fix this now and update.

Thanks again!

PR #4531, opened 2017-01-30
Add cov

AdamNiederer:

Displays gcov data on your code

https://github.com/AdamNiederer/cov

I am the maintainer of this project.

### Checklist

### DIFF
--- a/recipes/cov
+++ b/recipes/cov
@@ -1,0 +1,1 @@
+(cov :fetcher github :repo "AdamNiederer/cov")


syohex:
I have sent PR. https://github.com/AdamNiederer/cov/pull/5

purcell:
Thanks. How does this differ from "coverage" and "coverlay"?

AdamNiederer:
Coverage.el only supports a specific coverage tool for Ruby, and coverlay only supports lcov. Both don't have profiling support.
This package is also written to be easily accept other coverage backends without much modification, so it should accept more formats than both of them, eventually.

purcell:
Thanks, and sorry for the delay. Would be nice if you could ping the authors of the other packages and join forces -- having a standard coverage tool which works across languages would be a good outcome, and it would be a shame if `cov` never gained support for additional backends.

PR #4532, opened 2017-01-31
Add recipe for buffer-watcher

NicolasPetton:

Buffer-watcher makes it easy to run shell scripts per file type/directory when a buffer is saved.

https://github.com/nicolaspetton/buffer-watcher

**None needed**

### Checklist

### DIFF
--- a/recipes/jade
+++ b/recipes/jade
@@ -1,1 +1,3 @@
-(jade :fetcher github :repo "NicolasPetton/jade")
+(jade :fetcher github
+      :repo "NicolasPetton/jade"
+      :files ("*.el" "doc/texinfo/*.texi"))


syohex:
What's different from [firestarter](https://github.com/wasamasa/firestarter) ?

NicolasPetton:
@syohex I didn't about this package, but from what I saw, it's a bit different: to run shell commands for buffers of a specific mode in a project with firestarter, you have to use `.dir-locals.el` in the project directories, while `buffer-watcher` uses a custom var `buffer-watcher-commands` that you can put in `custom.el` or anywhere in your `.emacs.d`.

mattiasb:
FWIW the dir-locals stuff doesn't mandate using actual `.dir-locals.el` files. 

A (modified) example from my config (this sets code style rather than firestarter-settings though):
```elisp
(c-add-style
 "my-style"
 '("stroustrup"
   (c-basic-offset  . 2)
   (c-offsets-alist . ((innamespace       . 0)
                       (substatement-open . 0)
                       (inline-open       . 0)))))

(dir-locals-set-class-variables
 'my-code
 '((c++-mode . ((c-file-style . "my-style")))))

(dir-locals-set-directory-class "~/Code/my-repo/" 'my-code)
```

syohex:
This package sets `buffer-watcher-run-commands` to `after-save-hook` only by loading. I suppose this package provides global minor-mode instead of it.

manuel-uberti:
Any chance of merging this?

manuel-uberti:
Anything? :)

purcell:
@manuel-uberti - this is currently closed, so not on our radar. Is that wrong?

manuel-uberti:
Maybe @NicolasPetton can tell us more.

NicolasPetton:
This PR was closed because firestarter offers similar features. I don't mind but I still use buffer-watcher, and if the package can be useful for others, we could reopen the PR.

manuel-uberti:
Well, I prefer the approach of `buffer-watcher` with a defcustom instead of using dir-locals, but I see your point.

If it's not much of a problem, I'd very much like `buffer-watcher` on MELPA.

purcell:
Added in c8681776. An autoloaded global minor mode would be slightly neater than enabling this by simply calling `require`.

manuel-uberti:
Thank you @purcell. A minor mode would be much appreciated, of course :+1:

PR #4533, opened 2017-02-01
Add recipe for build-status

sshaw:

Mode line build status indicator

https://github.com/sshaw/build-status

Author

### Checklist

### DIFF
--- a/recipes/build-status
+++ b/recipes/build-status
@@ -1,0 +1,3 @@
+(build-status
+ :fetcher github
+ :repo "sshaw/build-status")


syohex:
- `--` separator should not be used for `defcustom` variable, because it is not private variable. `--` means private variable/functions by convention
- Define group by `defcustom` variable for `M-x customize`

sshaw:
Okay, I dropped the `defcustom` and just went with `defvar` without the "private" convention.

purcell:
Thanks, looks good.

Rather than the alist of background colours, you should define some faces so that customisation via themes will be straightforward. Define a base face, e.g. `build-status-face`, then `:inherit` that in a number of more specific faces, e.g. `build-status-failed-face`, adding a `:background` colour there.

(You'd also generally place those `defface`s in a `defgroup`, sorry!)

sshaw:
@purcell nice suggestion, though what's the purpose for the base face, for `class color`?

purcell:
> what's the purpose of the base face

It gives you a common place to set defaults for the other faces, e.g. `bold` etc. But it's not strictly necessary, of course.

sshaw:
Checkout my latest update. 

One question. If the face has no attributes the mode line text changes. Do you think [using `face-differs-from-default-p`](https://github.com/sshaw/build-status/blob/6b4822a65d0c1433434f85824aae9f6d550bd308/build-status.el#L300) is the way to check this?

sshaw:
Also, the package linter wants me to specify a containing group in my `defgroup`. Suggestions?

purcell:
> Also, the package linter suggests I specify a containing group in my defgroup. Suggestions?

`'programming`?

purcell:
> Checkout my latest update.

Looks good. It's generally best to specify both background and foreground colours, otherwise you'll get caught out when users have (for example) "green" as a default foreground colour. There are some built-in (and therefore widely-themed) faces you can optionally inherit from in some cases, e.g. `successs`, `warning`, `error`.

> One question. If the face has no attributes the mode line text changes. Do you think using face-differs-from-default-p is the way to check this?

No, don't inspect faces at runtime to decide behaviour. Just pick one branch of that conditional and use it in all cases.

sshaw:
> It's generally best to specify both background and foreground colours...

`build-status-face` inherits from `mode-line`, which I think would keep my faces somewhat in-line with the user's defaults. This is not sufficient?

>> One question. If the face has no attributes the mode line text changes. Do you think using face-differs-from-default-p is the way to check this?

> No, don't inspect faces at runtime to decide behaviour. Just pick one branch of that conditional and use it in all cases.

Hurmmm... In general that makes sense, but in this case it will look odd as there are extra spaces. but no background color. With a background it not an issue as it becomes a well-defined box.

purcell:
> build-status-face inherits from mode-line, which I think would keep my faces somewhat in-line with the user's defaults. This is not sufficient?

But `mode-line` could have `:background 'green`, right?

> Hurmmm... In general that makes sense, but in this case it will look odd as there are extra spaces but no background color. With a background it not an issue as it becomes a well-defined box.

I still don't understand what you're going for with the extra spaces, and switching based on font similarity is a non-starter. In any case, you shouldn't have spaces in your mode lighter, because mode lighters for separate minor modes are conventionally _separated_ by spaces. I'd expect your lighter here to look something like ` Build:-`, ` Build:P`, ` Build:?` etc.

sshaw:
> But mode-line could have :background 'green, right?

And so can the `success` face. There's always chance of a conflict so I don't see your point here unless you're saying the probability of a conflict is smaller.

> I still don't understand what you're going for with the extra spaces,

The spaces form a box that indicate the build's status:

example-passing

If there's no custom face for the given status then the lighter looks odds with the extra spaces since there is no background color to form the "box". (I considered `:box`, but that doesn't look as good). 

> and switching based on font similarity is a non-starter.

Why is this a problem?

> In any case, you shouldn't have spaces in your mode lighter, because mode lighters for separate minor modes are conventionally separated by spaces.
> That's usually accomplished by always showing the lighter when the mode is enabled, and prefixing it with a single space.

Yes, this is what's currently being done. Have you even used the mode or looked at the screenshots in the README?

> So I'd expect your lighter here to look something like Build:-, Build:P, Build:? etc.

This is not a hard requirement. There are plenty of lighters that do no prefix themselves with the mode name. `display-time-mode` and  `display-battery-mode` come to mind.

This is a visual design decision. The "box style" lighter looks much better than `Build:X`.
If I was using a image would we even be having this conversation?

tarsius:
I think both lines of though are valid. And while I tend to agree with @purcell that it would be preferable to follow the conventions here, I appreciate the argument that this is (visual) design decision, and as such the decision is up to the package author. This isn't a grave violation of the conventions, it's just a reasoned departure. 

(Personally I would append an additional character to the `-:---` part of the mode-line, and set the mode-lighter to `""`.)

purcell:
Thanks guys.

PR #4534, opened 2017-02-01
add zpresent.

zck:

Zpresent.el is a package to present org files.

https://bitbucket.org/zck/zpresent.el

None needed

### Checklist

### DIFF
--- a/recipes/zpresent
+++ b/recipes/zpresent
@@ -1,0 +1,1 @@
+(zpresent :fetcher bitbucket :repo "zck/zpresent.el")


syohex:
Some similar tools are already registered in MELPA. What is different between this package and those packages.

- https://github.com/rlister/org-present
- https://github.com/takaxp/org-tree-slide

zck:
I don't like those packages, so I decided to write my own.

zck:
To be a bit more specific (because I suspect you will want that), this is more of a "traditional" presentation mode. The presented slides don't look like org-mode documents, and they're not editable like org-mode documents. You can move back and forth between the presentation and other buffers, even the original org doc.

To be honest, I'm also not real sure why it matters if there are multiple packages for similar purposes. Isn't Melpa a place for end-users to choose what packages they want to use?

purcell:
> Isn't Melpa a place for end-users to choose what packages they want to use?

Yes, but it's not necessarily a place for authors to publish whatever they want. We check all submissions for overall quality, duplication of functionality w.r.t. other packages, and adherence to basic packaging rules. All of this gives the end users a better experience.

Regarding this case, if zpresent were very similar to an existing package, we'd likely suggest that you work with the other author to jointly make a single package.

That's not the case here, so I'll go ahead and merge.

Regarding the code itself:
- You should move the major mode declaration lower in the file, to avoid forward references to all the commands. 
- `zpresent-increase-multiplier` and `zpresent-bullet` etc. should be `defcustom`s in a `defgroup`.

PR #4537, opened 2017-02-02
Add recipe for emacs-async-await

chuntaro:

This is a simple implementation of Async/Await.  
Inspired by the Async/Await implementation of TypeScript.

This implementation uses generator.el included in Emacs 25.

https://github.com/chuntaro/emacs-async-await

Author, maintainer

None needed

### Checklist

### DIFF
--- a/recipes/async-await
+++ b/recipes/async-await
@@ -1,0 +1,1 @@
+(async-await :repo "chuntaro/emacs-async-await" :fetcher github)


syohex:
- Please don't define functions which does not start with package prefix for naming convention. `a/a--is-error` etc. And please don't use `/` separator for coding convention. 
- Is `await` global macro necessary ? Is it enough to use local macro by `macro-let` ?
- Please don't use melpa package version like `(promise "20170118.431")` and use package version for other package manager`(promise "1.0")`.

chuntaro:
I modified it.

a/a-* -> async-await-*
(promise "20170118.431") -> (promise "1.0")

> Is await global macro necessary ? Is it enough to use local macro by macro-let ?

What I wanted to do was defmacro, which works only with the required file.
I do not know how to do it, so I commented out.

purcell:
Thanks. More quick feedback:
- The package should probably not include `async-await-examples.el`. The best way to achieve this would be to place it in an `examples` subdirectory in your repo.
- Please remove `for Emacs` from the package description: it's redundant, because all elisp packages are for Emacs. :-)

chuntaro:
I modified it.

- Move `async-await-examples.el` to `examples` subdirectory
- Remove `for Emacs`

purcell:
Nice, thanks!

chuntaro:
Thanks!

PR #4539, opened 2017-02-04
Fix not including examples subdirectory

chuntaro:
Since the examples subdirectory is not included, please merge.

### DIFF
--- a/recipes/async-await
+++ b/recipes/async-await
@@ -1,0 +1,1 @@
+(async-await :repo "chuntaro/emacs-async-await" :fetcher github)
--- a/recipes/async-await
+++ b/recipes/async-await
@@ -1,1 +1,2 @@
-(async-await :repo "chuntaro/emacs-async-await" :fetcher github)
+(async-await :repo "chuntaro/emacs-async-await" :fetcher github
+             :files ("async-await.el" ("examples" "examples/async-await-examples.el")))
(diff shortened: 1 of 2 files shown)


recipes/promise
```diff
@@ -1 +1,2 @@
-(promise :repo "chuntaro/emacs-promise" :fetcher github)
+(promise :repo "chuntaro/emacs-promise" :fetcher github
+         :files ("*.el" ("examples" "examples/*.el")))
```
milkypostman: instead of "*.el" you should use `:defaults`

recipes/async-await
```diff
@@ -1 +1,2 @@
-(async-await :repo "chuntaro/emacs-async-await" :fetcher github)
+(async-await :repo "chuntaro/emacs-async-await" :fetcher github
+             :files ("async-await.el" ("examples" "examples/async-await-examples.el")))
```
milkypostman: same here.

milkypostman:
this branch is out of date, and normally we don't include these files with the packages. the main reason is that it bloats the package. If the examples are really important they should probably go in the README.md file in the project or be linked to. You could alternatively create emacs documentation files which we automatically build.

purcell:
Agreed. Separate files of examples are not to be included in installed packages.

PR #4541, opened 2017-02-05
Add helm-perspeen recipe

jimo1001:

Simple helm interface for perspeen.

https://github.com/jimo1001/helm-perspeen

maintainer.

**None needed**.

### Checklist

### DIFF
--- a/recipes/helm-perspeen
+++ b/recipes/helm-perspeen
@@ -1,0 +1,1 @@
+(helm-perspeen :repo "jimo1001/helm-perspeen" :fetcher github)


syohex:
Is this line necessary ? 
- https://github.com/jimo1001/helm-perspeen/blob/2c25eb5132bc452a78a6373947d98e8706236ee4/helm-perspeen.el#L107

You declare `helm-projectile` as package dependency, so this `require` never fails. This s-expression is evaluated at `package-initialize` and `helm-projectile` and its dependencies are loaded at that time. This makes emacs initialization slow and is bad for package lazy loading.

jimo1001:
Surely, this line is unnecessary.
The integration with `helm-projectile` is an optional feature, so I removed from `Package-Requires`.
https://github.com/jimo1001/helm-perspeen/commit/90826312f85658d4bfb71b83e4d5c3ed0187da5a

syohex:
`with-eval-after-load` was introduced at Emacs 24.4. Please declare minimum Emacs version in `Package-Requires` header. Or use `eval-after-load`(+ `quote`) instead of it.

jimo1001:
replaced the using function from `with-eval-after-load` to `eval-after-load`.
https://github.com/jimo1001/helm-perspeen/commit/1d10a83b44cfe71eb3fe372cec10c640fe8b3deb

syohex:
Merged

PR #4544, opened 2017-02-06
Add picpocket recipe

johanclaesson:

Picpocket is an image viewer which requires GNU Emacs 24.4+ compiled with ImageMagick.
It have commands for adding tags to images.  The tags are saved to disk.

https://github.com/johanclaesson/picpocket

Author and maintainer.

None needed

### Checklist

- [not really] I've used [package-lint](https://github.com/purcell/package-lint) to check for packaging issues, and addressed its feedback

Regarding package-lint i use the prefix picp- as well as picpocket- since picpocket- is rather long.
package-lint does not like this.  This is probably unacceptable.  But personally i feel it makes sense.
Therefore i like to ask one thing before i change it.  I am thinking of changing package-lint so that it would accept the prefixes listed in a file local variable (say package-lint-prefixes) in addition to the package name.  Unless you think this is a terrible idea then i might attempt to add this behaviour to 
package-lint.

This is not a big deal.  If you don't like the idea then i will just change to use picpocket- prefix everywhere.

### DIFF
--- a/recipes/picpocket
+++ b/recipes/picpocket
@@ -1,0 +1,1 @@
+(picpocket :fetcher github :repo "johanclaesson/picpocket")


purcell:
Thanks! Quick feedback:
- The `defgroup` specifies itself as its own parent: please choose a standard group instead, e.g. `'applications`.
- The `picp-` prefix should be `picpocket-` throughout.

Looks good otherwise!

johanclaesson:
Thanks for the feedback.  
The defgroup parent is changed to multimedia.
All prefixes are changed to picpocket-.

johanclaesson:
Great, thanks!

PR #4548, opened 2017-02-08
Remove ebib-handy

tumashu:
ebib-handy cannot work again for the change of ebib, and I have no time and interesting to maintain it , so please remove it


syohex:
Removed

tumashu:
thanks

PR #4551, opened 2017-02-08
adding go-guru.el and go-rename.el

Darren:
Since guru has moved elisp files from golang.org/x/tools to go-mode
please add these files to go-mode package


syohex:
Both packages are already registered in MELPA as another package.

- https://melpa.org/#/go-guru
- https://melpa.org/#/go-rename

Please check them.

PR #4561, opened 2017-02-13
Add nocomments-mode

Lindydancer:

Minor mode that let you temporarily hide comments in a buffer..

https://github.com/Lindydancer/nocomments-mode

Author

None needed

### Checklist

### DIFF
--- a/recipes/nocomments-mode
+++ b/recipes/nocomments-mode
@@ -1,0 +1,1 @@
+(nocomments-mode :repo "Lindydancer/nocomments-mode" :fetcher github)


syohex:
What different from [hide-comnt](https://www.emacswiki.org/emacs/hide-comnt.el) ? I know hide-comnt does not provides minor-mode, however, do this package and it provide same features ?

Lindydancer:
Hi!

There are some differences between the two packages. They use two different
technical solutions. `hide-comnt' use the `invisible' text property whereas
`nocomment-mode' use `font-lock' to set the `display' property to display a
space instead of characters. Effectively, this mean:

* `hide-comnt' collapses the lines containing the comments, whereas
`nomments-mode' ensures that lines that doesn't contain comments stay in
place. I prefer the latter as I think it is easier to find my way around in
a buffer if text doesn't move around just because I hide the comments.
* Editing commands will still work as they normally do in `nocomments-mode'
(i.e. they can step into comments), whereas in `hide-comnt' they will act
as though the comments aren't there.
* `hide-comnt' doesn't update the buffer automatically, e.g. if you paste
in section with comments, those comments will be visible.

    -- Anders

syohex:
I see. Merged.

Lindydancer:
Thanks!

    -- Anders

PR #4571, opened 2017-02-19
Added recipe messages-are-flowing

legoscia:

messages-are-flowing makes it easier to send "flowed" email messages from Emacs, meaning that the recipient's email client is free to rewrap lines within paragraphs. This makes the message more readable when viewed on a narrow screen, such as a phone.

https://github.com/legoscia/messages-are-flowing

I am the maintainer.

**None needed**

### Checklist

   > It gives one warning: `You should depend on (emacs "24") if you need lexical-binding.`  
   > Not sure if that's the right thing to do - the code works the same with lexical or dynamic binding, though I guess the byte compiler will generate more efficient byte code with lexical binding enabled. I would expect the code to work in Emacs 23, so strictly speaking there is no need to depend on Emacs 24, and there is also no need to remove the `lexical-binding` setting, as it will be ignored if an old Emacs doesn't understand it.

### DIFF
--- a/recipes/messages-are-flowing
+++ b/recipes/messages-are-flowing
@@ -1,0 +1,1 @@
+(messages-are-flowing :repo "legoscia/messages-are-flowing" :fetcher github)


syohex:
Merged

PR #4574, opened 2017-02-19
Add elquery

AdamNiederer:

It's the most web-scale Javascript framework not written in Javascript. When they add an Emacs Lisp VM to Firefox, you'll all be thanking me.

(for real, it's an HTML parsing, manipulation, and output library)

https://github.com/AdamNiederer/elquery

I wrote it this weekend

### Checklist

### DIFF
--- a/recipes/elquery
+++ b/recipes/elquery
@@ -1,0 +1,1 @@
+(elquery :fetcher github :repo "AdamNiederer/elquery")


syohex:
I have sent PR. Please check 
- https://github.com/AdamNiederer/elquery/pull/1

And please check following stuffs.
- `elq-rm-prop` references free variable `props`
- `elq--new-node` parenthesis looks strange. And `,` is used at out of backtick list
- `elq--set-prop` is undefined, but it is used.

purcell:
Looks fun! Since this is a new package, you should take the opportunity to match up the package name the symbol prefix: just use `elquery-`, rather than `elq-`. It'd also be worth running `M-x checkdoc`, since there are a quite a few docstrings to tidy up.

AdamNiederer:
Namespaces and docstrings have been cleaned up. I suppose we can shorten the function names in production with uglify.el anyway :).

purcell:
> Namespaces and docstrings have been cleaned up.

Oh... the docstrings still produce about 50 checkdoc warnings when I try it out...

purcell:
Well, `checkdoc` specifies the official standard format, so your objections don't really hold up, sorry. One initial sentence, which may exceed 80 chars, and then following sentences. I agree that the requirement to mention arguments in order can be a pain, but that only applies to the first sentence, so there's a way around it: write a summary in the first sentence, and introduce the arguments in following sentences.

P.S. Flycheck lets you see checkdoc warnings directly in the buffer, which helps get things right.

AdamNiederer:
Docstrings now conform to checkdoc, save for one stylistic false positive.

purcell:
Yay, great - thanks. Sorry for the slow response!

PR #4578, opened 2017-02-22
recipe for sayid

bpiel:

Sayid is a debugger for clojure. This package is an emacs nrepl client to interface with sayid.

https://github.com/bpiel/sayid

author

**None needed**

### Checklist

### DIFF
--- a/recipes/sayid
+++ b/recipes/sayid
@@ -1,0 +1,3 @@
+(sayid :fetcher github
+       :repo "bpiel/sayid"
+       :files ("src/el/*.el"))


syohex:
- Package dependency header is missing(I have sent PR. https://github.com/bpiel/sayid/pull/20)
- [This line](https://github.com/bpiel/sayid/blob/01bf777cb15a4f236bc44842712e9ca82fed7f55/src/el/sayid.el#L28) should be removed. Its default value is `nil`. A package should not set other package custom variable.
- `clojure-mode-map` is modified only by installing this package(code is [here](https://github.com/bpiel/sayid/blob/01bf777cb15a4f236bc44842712e9ca82fed7f55/src/el/sayid.el#L879-L881)). You should provide minor-mode or some setup function instead of this.
- You should use `define-derived-mode` for defining major-mode instead of `defun`

bpiel:
@syohex Thanks for the quick response. I'll get on this asap.

bpiel:
@syohex I believe I've addressed all of your points. Ready for another look. Thanks

purcell:
Quick feedback from me:
- The keybinding `C-c s` is reserved for end user customisation, and you should therefore not use it. See the corresponding [elisp manual page](https://www.gnu.org/software/emacs/manual/html_node/elisp/Key-Binding-Conventions.html). What you probably need to do is provide a prefix map and then tell people in your README how to bind it to a prefix of their choice, which could then be `C-c s`.
- You can remove `(run-hooks 'sayid-mode-hook)`: `define-derived-mode` will expand to code which does this for you. Same for your other modes.
- Just merge `sayid-mode.el` into `sayid.el`... there's little reason to split this across two files, particularly when the latter has hard-coded strings which document the bindings defined in the former. I'd personally merge all the files into one, given the current shape of things.
- Consider using the build-in Emacs machinery for generating mode binding documentation rather than building a big help string with that info.

bpiel:
@syohex @purcell 

I addressed @purcell's first three points and I believe this is ready for another look.

Regarding the last point about docs: Although I'm not fond of my implementation, I like that the user is able to see concise docs showing only keybindings for sayid and that descriptions are provided, instead of just command names. If there's a more idiomatic way to achieve that, please let me know. I couldn't find anything built-in that met that criteria.

thank you

purcell:
Sounds reasonable, I also can't find exactly the built-in stuff you'd need. Perhaps worth a look at `hydra`. Anyhow, merging now.

bpiel:
@purcell @syohex thanks!

PR #4581, opened 2017-02-23
Add nikola.el recipe

drym3r:

Simple wrapper for nikola in emacs

https://git.daemons.cf/drymer/nikola.el

I'm the author

Nope.

### Checklist

### DIFF
--- a/recipes/nikola
+++ b/recipes/nikola
@@ -1,0 +1,1 @@
+(nikola :fetcher git :url "https://git.daemons.cf/drymer/nikola.el/")


syohex:
- `defcustom` value should have `:type` attributes(Newer emacs warns without `:types` attribute)
- No autoload cookie
- You should use `cl-search` instead of `search` and load `cl-lib` for using it

drym3r:
[x] - defcustom value should have :type attributes(Newer emacs warns without :types attribute): Done
[ ] - No autoload cookie: not sure if I understand it, I should ;;;###autoload before all commands?
[x] - You should use cl-search instead of search and load cl-lib for using it: Done, but why is that?

purcell:
> You should use cl-search instead of search and load cl-lib for using it: Done, but why is that?

The unprefixed names are deprecated. `cl` has been replaced by `cl-lib`.

More quick feedback from me:
- You should run `M-x checkdoc`, to get some help fixing up the invalid docstrings.
- You shouldn't be sending `C-u` programmatically, e.g. in `(kbd "C-u")(read-only-mode 0)`. See the docstring for `read-only-mode` to find out what you should be doing instead in that case.
- I think your lambdas which modify an `output` var in the enclosing scope will not work safely (or as you expect) when `lexical-binding` is not enabled for the file.
- Small typo: `get it's version` => `get its version`

drym3r:
Thanks for the feedback!

> You should run M-x checkdoc, to get some help fixing up the invalid docstrings.

You mean there should be more invalid docstrings? Because checkdoc doesn't show any.

> You shouldn't be sending C-u programmatically, e.g. in (kbd "C-u")(read-only-mode 0). See the docstring for read-only-mode to find out what you should be doing instead in that case.

Sure, changed.

> I think your lambdas which modify an output var in the enclosing scope will not work safely (or as you expect) when lexical-binding is not enabled for the file.

I'm not sure what this means. This is the first thing I write with elisp, if you have some useful information I could read, I'd appreciate it.

> Small typo: get it's version => get its version

Changed.

purcell:
> You mean there should be more invalid docstrings? Because checkdoc doesn't show any.

Erk, not sure what my confusion was there. Sorry, please ignore.

> > I think your lambdas which modify an output var in the enclosing scope will not work safely (or as you expect) when lexical-binding is not enabled for the file.
> 
> I'm not sure what this means. This is the first thing I write with elisp, if you have some useful information I could read, I'd appreciate it.

You create a lambda which is evaluated later by `async`, but its job is to modify `output` in the enclosing scope. For that to behave well, I think you will need lexical binding, as is usual in other non-lisp languages with lambdas. In short, add `-*- lexical-binding: t -*-` at the end of the first line of the file, and check it still works.

drym3r:
I don't really understand what you say but I changed it and it works. I'll just add that the way I use it, it's the recommended, as you can see in the [example](https://github.com/jwiegley/emacs-async).

purcell:
Great, thanks - merged!

It's worth reading up on the difference between dynamic and lexical binding in general, and in emacs lisp particularly.

PR #4582, opened 2017-02-23
Add luarocks recipe

marsam:

Includes helpers for [LuaRocks](https://luarocks.org/)

https://github.com/emacs-pe/luarocks.el

maintainer

None needed

### Checklist

### DIFF
--- a/recipes/luarocks
+++ b/recipes/luarocks
@@ -1,0 +1,1 @@
+(luarocks :fetcher github :repo "emacs-pe/luarocks.el")


syohex:
I have sent PR(https://github.com/emacs-pe/luarocks.el/pull/1). Please check it.

purcell:
Quick feedback from me:
- "for Emacs" is redundant in `"LuaRocks tools for Emacs."` and should be removed. Same with ` in this Emacs`. later.
- Why use `(parse-colon-path (getenv "PATH"))` instead of `exec-path`? They should always be in sync anyway, otherwise all sorts of things can go wrong.

marsam:
> "for Emacs" is redundant in "LuaRocks tools for Emacs." and should be removed. Same with in this Emacs. later.

Thanks for the suggestion, I've removed it.

> Why use (parse-colon-path (getenv "PATH")) instead of exec-path? They should always be in sync anyway, otherwise all sorts of things can go wrong.

I think `exec-path` is initialized from `PATH` on startup, so further modifications on `PATH`, can make them out of sync. e.g https://github.com/jorgenschaefer/elpy/issues/1086#issuecomment-282291034

purcell:
> I think exec-path is initialized from PATH on startup, so further modifications on PATH, can make them out of sync. e.g jorgenschaefer/elpy#1086 (comment)

True, but then that's a bug elsewhere, including in the case you referenced. Emacs code generally relies on `exec-path` being correct at all times. You shouldn't have to add code to work around that IMO. Anyhow, I'll leave that choice to you.

PR #4585, opened 2017-02-25
Add recipe for apiwrap

vermiculus:

[`API-Wrap.el`](https://github.com/vermiculus/apiwrap.el) is a tool to interface with the APIs of your favorite services.  These macros make it easy to define efficient and consistently-documented Elisp functions that use a natural syntax for application development.

https://github.com/vermiculus/apiwrap.el

Author

**None needed**

### Checklist

### DIFF
--- a/recipes/apiwrap
+++ b/recipes/apiwrap
@@ -1,0 +1,1 @@
+(apiwrap :fetcher github :repo "vermiculus/apiwrap.el")


syohex:
I have sent PR about cl-lib. Please check it. https://github.com/vermiculus/apiwrap.el/pull/1

vermiculus:
Resolved 😄

purcell:
When posting PRs like this, please don't delete the template which github inserts. The information we ask you to enter there is very helpful to us.

The code (at least the `apiwrap-new-backend` macro) appears prone to issues with multiple expansion of parameters, so you should probably rewrite it to use `gensym`. See the Practical Common Lisp book for a great overview of these issues: http://gigamonkeys.com/book/macros-defining-your-own.html

vermiculus:
@purcell That appears to be a current bug in Magithub, then – I've been reorganizing the internals to not use the `hub` command line tool. (That is to say, I never removed the template because it never made it in 😦).

I've updated my comment and will be addressing the subsequent points shortly.

purcell:
> That appears to be a current bug in Magithub, then

Aha, cool -- it did occur to me to wonder whether the templates work with non-web workflows for opening PRs...

vermiculus:
> the code […] appears prone to issues with multiple expansion of parameters

@purcell After re-reading the article you linked, I don't see where I'm leaking any bindings. One expansion (`macroexpand-1`) of

```elisp
  (apiwrap-new-backend "GitHub" "ghubp"
    ((repo . "REPO is a repository alist of the form returned by `/user/repos'.")
     (org  . "ORG is an organization alist of the form returned by `/user/orgs'."))
    (lambda (version link) (format "https://developer.github.com/v%d/%s" version link))
    ghub-get ghub-put ghub-head ghub-post ghub-patch ghub-delete)
```

gives me

```elisp
(defmacro ghubp-defget (resource doc version link &optional object internal-resource &rest kwargs)
    "Define a new GET resource wrapper function. [...]"
    (declare (indent defun) (doc-string 2))
    (when (keywordp object)
      (setq kwargs (cons object (cons internal-resource kwargs))
            object nil internal-resource nil))
    (when (keywordp internal-resource)
      (setq kwargs (cons internal-resource kwargs)
            internal-resource nil))
    (apiwrap--defresource "ghubp" 'get 'ghub-get
      (lambda (version link)
        (format "https://developer.github.com/v%d/%s" version link))
      '((repo . "REPO is a repository alist of the form returned by `/user/repos'.")
        (org . "ORG is an organization alist of the form returned by `/user/orgs'.")
        (thread . "THREAD is a thread"))
      'nil kwargs resource doc version link object internal-resource))
```

One expansion (`macroexpand-1`) of

```elisp
(ghubp-defget "/issues" "List all [...]" 3 "issues/#list-issues")
```

gives me

```elisp
(prog1
    (defun ghubp-get-issues (&optional data &rest params)
      "List all [...]"
      (declare (indent defun))
      (apply 'ghub-get "/issues"
             (if (keywordp data)
                 (list (funcall #'identity (apiwrap-plist->alist (cons data params))))
               (list (funcall #'identity (apiwrap-plist->alist params)) data))))
  (put 'ghubp-get-issues 'apiwrap-prefix 'ghubp)
  (put 'ghubp-get-issues 'apiwrap-version 3)
  (put 'ghubp-get-issues 'apiwrap-method 'get)
  (put 'ghubp-get-issues 'apiwrap-endpoint "/issues")
  (put 'ghubp-get-issues 'apiwrap-documentation "https://developer.github.com/v3/issues/#list-issues"))
```

All the let-bindings I have in the macrocode are strictly in an evaluation context. Am I missing something? …should I be generating symbols for the `defmacro` generation? How will that impact docstrings?

purcell:
> Am I missing something?

Yeah, so the error case occurs when a user writes something like:
```el
(apiwrap-new-backend "blah" (some-expensive-operation-returning-prefix) ...)
```
because then you'll get multiple calls to `some-expensive-operation-returning-prefix` in the macro-expanded code. So you should generate a `let` which binds a `gensym`-d name to that value, to ensure it is only expanded once.

vermiculus:
I've been working on this for most of the day and am starting to think myself into circles. Would it be alright if I just errored out if these arguments aren't string literals? I don't think your use case is terribly likely.

purcell:
Yeah, that can be a way of solving it, but it's also a bit lame, because then you'd stop users from storing params in a simple var, e.g. `my-api-prefix` and then using that to reduce duplication.

If I were tackling this myself, I'd look at unpicking the nested levels of quoting and unquoting, because it's hard to see what's going on at runtime vs. compile time. I've been in this situation before myself, so I recognise the confusion that can set in! :-)

vermiculus:
@purcell I'll look into simplifying the quoting (even if only for my own sanity), but I don't see where duplication can come into play. Params in particular are meant to be specified exactly once in `apiwrap-new-backend`; they propagate to all macros from there.

vermiculus:
vermiculus/apiwrap.el@ae1d361be61527598443ede3c79d411343848bd3 should resolve this 😄

syohex:
- `->>` is dash.el macro. You should add dash dependency(and load it) or use `thread-last` of `subr-x` macro
- `gensym` is cl function. You should use `cl-gensym` instead of it and load `cl-lib`
- Don't use `add-to-list` for local variable in`lexical-binding:t`. This causes byte-compile error, you should use `cl-pushnew` or `push` instead of it

vermiculus:
Addressed all concerns. I've verified that the package (and files that use the package) compile appropriately with no warnings.

syohex:
Merged

PR #4587, opened 2017-02-26
Added recipe ivy-todo

Kungsgeten:

`ivy-todo` lets you manage org-mode TODO lists using [ivy](https://github.com/abo-abo/swiper). The primary use case is to manage TODO lists of the project you're currently working on (ivy-todo will guess using [projectile](https://github.com/bbatsov/projectile) or vc.el).

A somewhat similar package is [org-projectile](https://github.com/IvanMalison/org-projectile) but `ivy-todo` isn't bound to projectile and you do not necessary need to manage project TODOs with it. The workflow is also different compared to `org-projectile`.

https://github.com/Kungsgeten/ivy-todo

None needed

### Checklist

Sorry, I haven't used the last one. I've submitted packages to Melpa before though, so I hope it'll work.

### DIFF
--- a/recipes/ivy-todo
+++ b/recipes/ivy-todo
@@ -1,0 +1,1 @@
+(ivy-todo :repo "Kungsgeten/ivy-todo" :fetcher github)


Kungsgeten:
Hi! Sorry for disturbing, but it's been a while and I haven't heard anything regarding this pull request. Perhaps I'm spoiled because my previous pull requests has been handled quickly :) @purcell @syohex

syohex:
- Emacs has `vc.el` since very old versions and `(require 'vc nil t)` never fails. But `vc-root-dir` was introduced at Emacs 25. This package requires Emacs 24.3 or higher versions but 24.3 and 24.4 and 24.5 does not have `vc-root-dir` and then exception is thrown on such Emacs.

Kungsgeten:
Good catch! I guess this commit should fix it: https://github.com/Kungsgeten/ivy-todo/commit/01b8e99e764ba50f75d8e07a5c8267bca5dbcdca

purcell:
This is still going to explode for users with neither a recent Emacs or projectile. 

The right way to handle this IMO is to provide a `defcustom` that defines the function for finding the project root, then fail at runtime if it is unset. By all means set the default value for that defcustom to either projectile-project-root or vc-root if you find they are defined at the time your library is loaded.

Kungsgeten:
@purcell It might be a good idea to provide a `defcustom` like you suggest, in order to optimize `ivy-todo--guess-headline-name` a bit. That function is usually run once per file though, and isn't particullary slow. It will not explode Emacs: If neither `projectile`, `find-file-in-project` or `vc` cond clause triggers it will return `nil` and the user just enters the TODO list name manually (see lines 85--86). `ivy-todo` doesn't have to be used in projects, it can manage general TODO lists.

purcell:
I'll go ahead and merge, but I'd definitely suggest moving to configuration rather than run-time tests.

Thanks!

PR #4588, opened 2017-02-26
Add recipe for flow-mode

an-sh:

Major mode for flowtype.org derived from web-mode. Very minimal, but better than nothing.

https://github.com/an-sh/flow-mode

Maintainer + enthusiastic user. Some code is borrowed from the original [flow-for-emacs](https://github.com/flowtype/flow-for-emacs) snippet, hence the original licence is left intact.

**None needed**

### Checklist

### DIFF
--- a/recipes/flow-mode
+++ b/recipes/flow-mode
@@ -1,0 +1,1 @@
+(flow-mode :fetcher github :repo "an-sh/flow-mode")


syohex:
- I have sent PR(https://github.com/an-sh/flow-mode/pull/1), please check it.
- You should use `process-file`(`call-process`) instead of `shell-command`(`shell-command-to-string`) because if `flow-binary` contains spaces, then command is failed.

an-sh:
Thanks for pointing that out, i've rewritten flow iterations using `call-process`.

syohex:
https://github.com/an-sh/flow-mode/blob/9538632fbbed2dea0ab7addfa463db8be6fd3d4a/flow-mode.el#L176-L177

I suppose you should use named function instead of lambda function. Lambda function can not be removed from hook and can be added to hook multiple times.

an-sh:
fixed

syohex:
Merged

PR #4589, opened 2017-02-26
Add remember-last-theme recipe

anler:

This package makes Emacs remember the last used theme, so nex time is opened it uses that same theme.

https://github.com/anler/remember-last-theme

I'm the maintainer

None needed

### Checklist

### DIFF
--- a/recipes/remember-last-theme
+++ b/recipes/remember-last-theme
@@ -1,0 +1,1 @@
+(remember-last-theme :repo "anler/remember-last-theme" :fetcher github)


purcell:
Thanks, but I think you're reinventing Emacs' customization system here for a single variable. `customize-themes` already stores the selected theme(s) for the next startup. Presumably you're trying to make the result of arbitrary `load-theme`  calls persistent. I wouldn't say this is advisable on an ad-hoc basis, because users who care about this should use the custom interface instead, so I'm not super keen to add this package.

(If you must write code for this case, then using `custom-set-variable` to force the saving `custom-enabled-themes` would be a better way to handle this than to advise `load-theme`.)

anler:
Ok, I'm a bit lazy and having to call `customize-themes` and select the theme is a bit too much work, I already have a keybinding to changing themes (by invoking `helm-themes`) and advising `load-theme` is the only way I've found to persist it. I'll try to find a way that doesn't reinvent Emacs though 😄 
Thanks!

purcell:
Yeah, I'm guessing it's be possible to just use something like this:

```el
(defun force-save-themes ()
  (custom-set-variables (list 'custom-enabled-themes custom-enabled-themes))
  (custom-save-all))

(add-hook 'kill-emacs-hook 'force-save-themes)
```

anler:
Thanks a lot @purcell ! I really appreciate the care all of you put in melpa 🙏

anler:
Hi again! I have changed the original approach and now I'm using customize api plus a couple of hooks. Do you think it's useful enough to deserve the merge as a package?

purcell:
Yes, I think so. First, I've sent you some small fixes in https://github.com/anler/remember-last-theme/issues/1.

- Add an `;;;###autoload` for the

anler:
Thanks @purcell How about now?

purcell:
Yay, thanks.

anler:
Awesome! Thank you! 😃

PR #4592, opened 2017-02-28
Add recipe for el-exiftool

arunisaac:

el-exiftool is an elisp wrapper around ExifTool.  ExifTool supports reading and writing metadata in various formats including EXIF, XMP and IPTC.

https://git.systemreboot.net/el-exiftool

Author and maintainer

### Checklist

### DIFF
--- a/recipes/el-exiftool
+++ b/recipes/el-exiftool
@@ -1,0 +1,3 @@
+(el-exiftool
+ :fetcher git
+ :url "https://git.systemreboot.net/el-exiftool/")


syohex:
- Please load `tq` for using its function
- Don't load `cl.el` at load time. You should load cl-lib.el instead of it and use cl-lib functions/macros. However it looks cl-lib functions are not used in your package.
- Unused local variable name should start with `_`. Please check byte-compiling result.

arunisaac:
Thanks for the feedback. I've made all three changes. However, since I'm using `cl-destructuring-bind` and `cl-function`, both macros from `cl-macs`, I'm requiring `cl-macs` instead of `cl-lib` as you suggested. Is that alright?

syohex:
- Please load `cl-lib` instead of `cl-macs`. 3rd party should load `cl-lib`.

arunisaac:
Done. `cl-lib` loaded instead of `cl-macs`.

syohex:
Merged

purcell:
I'd have suggested you rename this to `exiftool`. The `el-` prefix is redundant and unhelpful. If you'd like to do that now, I'd encourage it: just rename the source file and the symbols in it, then submit a subsequent PR to rename and update the MELPA recipe.

arunisaac:
Ok, I'll rename the library to `exiftool`.

purcell:
Yay, thanks!

PR #4594, opened 2017-02-28
Add literal-string-mode package

joodie:
This package provides a minor mode for editing long string
literals (especially docstrings) in a dedicated buffer.

The source repository is at
https://github.com/joodie/literal-string-mode

I am the author of the literal-string-mode package.

### DIFF
--- a/recipes/literal-string
+++ b/recipes/literal-string
@@ -1,0 +1,1 @@
+(literal-string :fetcher github :repo "joodie/literal-string-mode")


syohex:
- I have sent PR please check this(https://github.com/joodie/literal-string-mode/pull/1)
- `defgroup` should specify `:group`(parent group) attribute
- `defcustom` variable should specify `:type`  attribute
-  There are some byte-compile warnings, please check.

joodie:
@syohex thanks for the speedy response. I merged your PR and fixed other issues, as far as I can see.

Please let me know if there is anything else I can do to get this PR merged.

Thanks again,
Joost.

Fanael:
I believe it should be built upon edit-indirect – should be easy to do, edit-indirect has enough hooks to transform the text, so indenting and dedenting could be done there – instead of reimplementing the same functionality yet again.

purcell:
Yep, agreed. Re-use is good!

joodie:
@Fanael & @purcell - Thanks for suggesting `edit-indirect` - I did not know it existed and it looks like a good library for replacing a large chunk of `literal-string-mode`. Since my time for writing & maintaining public code is limited I prefer to offload as much functionality as possible to other people - who will probably do a better job anyway ;-) I will definitely try to get the the next version of `literal-string-mode` based on `edit-indirect`.

Will this PR be on hold until that happens or can the recipe be included in melpa with the library in the current state?

purcell:
> Will this PR be on hold until that happens or can the recipe be included in melpa with the library in the current state?

Have merged for now, but please do refactor to use `edit-indirect` when you get a chance! There's nothing quite like deleting code. :-)

joodie:
Thanks!

Work is underway to make the change. Needed a few tweaks to edit-indirect to make it functional. Those just got accepted.

PR #4595, opened 2017-03-01
Rename el-exiftool to exiftool

arunisaac:
I've renamed `el-exiftool` to `exiftool` as discussed in #4592


syohex:
Would you add `:old-names` attribute ?

arunisaac:
Done.

syohex:
Thanks

PR #4596, opened 2017-03-02
Add recipe for neato-graph-bar

RobertCochran:

Provides updating graph bars for CPU and memory/swap, inspired by `htop`.
Useful to have in an Emacs window to quickly eyeball CPU or memory usage.

https://gitlab.com/RobertCochran/neato-graph-bar

Author and maintainer.

Well, that's me, so **None needed**

### Checklist

I didn't try the build in the sandbox, but it worked on my normal Emacs session.
I only depend on Emacs 24.3 and `cl-lib`, so there shouldn't be any dependency problems,
but you never know.

Please let me know if there's anything I need to do or fix.

Thanks!

### DIFF
--- a/recipes/neato-graph-bar
+++ b/recipes/neato-graph-bar
@@ -1,0 +1,3 @@
+(neato-graph-bar :fetcher gitlab
+                 :repo "RobertCochran/neato-graph-bar"
+                 :branch "release")


syohex:
- Please add `;;; Code:` meta comment at beginning of code(https://gitlab.com/RobertCochran/neato-graph-bar/blob/master/neato-graph-bar.el#L22)
- `cl-lib` was bundled since Emacs 24.3 so you need not specify `cl-lib` dependency
- `for Emacs` in package description is redundant

syohex:
And I suppose you should use git tag and MELPA stable for stable release instead of `release` branch.

RobertCochran:
> `cl-lib` was bundled since Emacs 24.3 so you need not specify cl-lib dependency
Okay, fair enough. I'll fix this when I have time.

> `for Emacs` in package description is redundant
You're not wrong, but what does it hurt?

> And I suppose you should use git tag and MELPA stable for stable release instead of `release` branch
To be perfectly honest, I've never used the tags mechanism and don't want to mess things up. But I've realized how that messes up MELPA stable... I'll look into it tomorrow if I have time.

> Please add `;;; Code:` meta comment at beginning of code(https://gitlab.com/RobertCochran/neato-graph-bar/blob/master/neato-graph-bar.el#L22)
I'm not doing this unless you have a *really* good reason for insisting (no, "it's tradition/the standard/similar" is not a good reason). Quite honestly, the header and footer are something I didn't particularly want to include either. What is the rationale? The code is literally right there. Why do we need a comment that states the obvious? Are there tools that expect to find it?

purcell:
> You're not wrong, but what does it hurt?

It's just noise in the package list, and therefore undesirable. `package-lint` should have pointed this out when you ran it on your code.

> I don't like this convention. Quite honestly, the header and footer are something I didn't particularly want to include either. What is the rationale here? The code is literally right there. Why do we need a comment that states the obvious? Are there tools that expect to find it?

We're not going to take time out to justify individual points of the established Emacs community standards, sorry. Standardised conventions like this exist to support both tools and other contributors: consistency across the community trumps individual preferences, so we encourage authors to follow the conventions by default when publishing code here for community use. (P.S. Flycheck will help you spot issues like this.)

alphapapa:
> Why do we need a comment that states the obvious? Are there tools that expect to find it?

Actually, yes, having those headings makes it easy to navigate the buffer, e.g. with tools like `navi-mode`.  Try this, open up a random, large elisp file from the Emacs source (which typically have several headings to organize the file), then run `navi-mode` (after installing it, of course), and then press `8`.

tarsius:
@RobertCochran It would be a shame if this package would not be added just because you don't want to add `;;; Code:`, and us insist on doing it because that line is not only a convention but also serves a real purpose as explained by @alphapapa.

RobertCochran:
I'll admit that it was initially frustration, but I think that I'll just accept that that's how things have to be. Gotta find time to *do* it, but I'm not going to argue any more.

purcell:
@RobertCochran Thanks - sent you https://gitlab.com/RobertCochran/neato-graph-bar/merge_requests/1.

RobertCochran:
Merged. Thanks back for doing those things for me.

purcell:
Thanks - merged here too! :-)

purcell:
Also: welcome to MELPA!

alphapapa:
By the way, another really nice thing about having headers is, if you add this to your `emacs-lisp-mode-hook`, you can fold/expand sections of code with them.  It helps navigating big files a lot!

```el
(defun ap/el-mode-outline-hook ()
  (setq outline-level 'ap/el-outline-level)
  (setq outline-regexp "\\(;;[;]\\{1,8\\} \\|\\((defun\\)\\)")
  (outline-minor-mode)

(defun ap/el-outline-level ()
  (or
   ;; Commented outline heading
   (and (string-match (rx
		       (* space)
		       (group (one-or-more (syntax comment-start)))
		       (one-or-more space))
		      (match-string 0))
	(- (match-end 0) (match-beginning 0) 1))

   ;; Lisp def heading
   ;; Add 8 (the highest standard outline level) to every keyword
   ;; heading
   (+ 8 (- (match-end 0) (match-beginning 0)))))
```

That uses the built-in `outline-minor-mode`, but there are also other packages that provide configurable folding that could make use of the headers.

Happy hacking!

PR #4597, opened 2017-03-02
post blog-minimal recipe

thiefuniverse:

Blog-minimal is a simple static site generator based on org mode. It's so simple and others can hack it quickly and then create and modify their own site.

https://github.com/thiefuniverse/blog-minimal

maintainer

**None needed**

### Checklist

### DIFF
--- a/recipes/blog-minimal
+++ b/recipes/blog-minimal
@@ -1,0 +1,3 @@
+(blog-minimal :fetcher github
+	      :repo "thiefuniverse/blog-minimal"
+	      :files ("*.el" "media" "templates" "about.org"))
--- a/recipes/blog-minimal
+++ b/recipes/blog-minimal
@@ -1,3 +1,3 @@
 (blog-minimal :fetcher github
 	      :repo "thiefuniverse/blog-minimal"
-	      :files ("*.el" "media" "templates" "about.org"))
+	      :files (:defaults "media" "templates" "about.org"))


purcell:
Thanks. Quick feedback:
- `:files ("*.el" "media" "templates" "about.org")` in the recipe should instead be `:files (:defaults "media" "templates" "about.org")`
- In the "Commentary" section of the main .el file, don't just point people at the website: it's best if the code itself provides the documentation a user will need, and the Commentary section will form the long package summary on the MELPA page for this package.
- You can skip the `-pkg.el` file because MELPA will generate it for you: instead it's often easier to just include a `;; Package-Requires: ...` header in the ain .el file
- `bm.el` is a pre-existing library for bookmarks. Your `bm-vars.el` and `bm-utils.el` should be called `blog-minimal-vars.el` and `blog-minimal-utils.el` for consistency.
- Similarly, `bm/` is not a good prefix. Firstly, `/` should be `-`, and secondly `bm` is not globally unique and should instead match the package name, ie. use `blog-minimal-` throughout. See the [elisp coding conventions](https://www.gnu.org/s/emacs/manual/html_node/elisp/Coding-Conventions.html) for more info.
- Use `M-x checkdoc` to get advice on how to format your elisp docstrings to match the conventions.
- You're using `s-` functions, but have not declared a dependency on the `s` package.

Hope that helps!

thiefuniverse:
Thanks for your guide, I have updated it.

syohex:
It is not enough yet. Please check https://github.com/thiefuniverse/blog-minimal/pull/4 and its comment.

thiefuniverse:
Thanks for your patience. :smile:

purcell:
Great, thanks.

PR #4598, opened 2017-03-02
Add recipe for project-shells

hying-caritas:

Manage the shell buffers of each project.

https://github.com/hying-caritas/project-shells.git

I am the maintainer.

**None needed**

### Checklist

### DIFF
--- a/recipes/project-shells
+++ b/recipes/project-shells
@@ -1,0 +1,1 @@
+(project-shells :repo "hying-caritas/project-shells" :fetcher github)


syohex:
I have sent PR please check it https://github.com/hying-caritas/project-shells/pull/1

syohex:
And there is no autoload cookie.

hying-caritas:
Thanks for pointing out this.  I just added autoload cookies.

purcell:
Thanks. Quick feedback:
- `project-shells-default-shell-name` should probably default to `shell-file-name`, not simply `"sh"`. (Also, `"/bin/sh"` is hard-coded later in the file.)
- The code looks like it will fail by default if projectile is not installed and loaded.
- Is it possible that you should be using `shell-quote-argument` instead of writing your own `project-shells--escape-sh`?
- Instead of the complex `project-shells-setup` code and custom variables holding various keybindings, a simpler and more idiomatic solution would be to create a small minor mode with the appropriate keymap, then enable that minor mode in the necessary buffers.

hying-caritas:
Thanks for comments and pointing out the issues.

- project-shells-default-shell-name is just the name of the shell buffer, not the shell inferior program to run, in fact (shell) is called for shell buffers, so explicit-shell-file-name, (getenv "ESHELL"), or shell-file-name will be used.  Later "/bin/sh" is hard-coded to run in ansi-term, but I then run "exec" explicit-shell-name project-shells-term-args.  I do this because I want to use different command line arguments for shell buffer and terminal buffer.

- Thanks to point this out.  Will check symbol binding and default to empty project name and home directory when the project related function definition is nil.

- Will change the code to use shell-quote-argument

- Will create a small minor mode.

purcell:
> project-shells-default-shell-name is just the name of the shell buffer, not the shell inferior program to run

Oh, okay - sorry!

hying-caritas:
I have changed the code to address your comments.

purcell:
Great, thanks.

PR #4599, opened 2017-03-03
Recipe: helm-sql-mode

eric-hansen:

Integrates Helm with sql-mode to let the user choose which DB to connect to (right now only Postgres is supported).

https://github.com/eric-hansen/helm-sql-mode

Maintainer/author

**None needed**

### Checklist

### DIFF
--- a/recipes/helm-sql-mode
+++ b/recipes/helm-sql-mode
@@ -1,0 +1,1 @@
+(helm-sql-mode :repo "eric-hansen/helm-sql-mode" :fetcher github)


eric-hansen:
Right now I'm not sure how to get non-Postgres DB connections to work due to how sql-connect works in this fashion, but if anyone has ideas that would be great.

syohex:
There are some packaging issue and I have sent PR please check https://github.com/eric-hansen/helm-sql-mode/pull/1

eric-hansen:
PR has been merged, thanks for the fixes!

purcell:
Thanks. Quick feedback from me too:
- This package doesn't provide a mode, so don't call it `helm-sql-mode`. Perhaps `helm-sql-connect`? (That would also be a better name than `helm-sql-connect-to` for the main interactive command: maybe swap the names of the two functions?)
- You're not using the "value" [here](https://github.com/eric-hansen/helm-sql-mode/blob/8bf9a419ca15d8761fd3f6a73a9e99177d65c9b3/helm-sql-mode.el#L26)
- `message` seems an odd way of providing candidate names to Helm: normally your candidate function would just return a plain list of keys.

eric-hansen:
Thanks all, I've decided to decline this PR in favor of renaming this to helm-sql-connect, and so will make a PR of that instead.

PR #4602, opened 2017-03-04
Add elisp-docstring-mode

Fuco1:
#### elisp-docstring-mode

A simple major-mode for editing content of Emacs Lisp docstrings.  Very useful in combination with [string-edit](https://github.com/magnars/string-edit.el).

#### Features

Provides font-lock.  In combination with string-edit, there is support for automatic escaping and unascaping of quotes.

https://github.com/Fuco1/elisp-docstring-mode

Author/Maintainer

### DIFF
--- a/recipes/elisp-docstring-mode
+++ b/recipes/elisp-docstring-mode
@@ -1,0 +1,1 @@
+(elisp-docstring-mode :fetcher github :repo "Fuco1/elisp-docstring-mode")


syohex:
- There is no autoload cookie(https://github.com/Fuco1/elisp-docstring-mode/pull/1)

Fuco1:
@syohex Thanks, merged!

purcell:
Cool, thanks. PS, `font-lock-defaults` is always buffer-local, so you can simply `setq` it.

tarsius:
But except for being noise, making it explicitly buffer local, isn't really an issue. So I am merging now. You should still change that though.

purcell:
Sorry, I think I thought I had clicked to merge already.

Fuco1:
@purcell huh? But it is merged, not sure what you mean.

Thanks for the tip though, I'll change it when I get back home :)

purcell:
> uh? But it is merged, not sure what you mean.

@tarsius had commented above that he was merging this. I was saying that I thought I had already done that. :-)

In any case, we all got the right outcome!

PR #4603, opened 2017-03-04
Add auto-read-only recipe

zonuexe:

Automatically make the buffer-file to read-only based on buffer-file-name. For example, it can protect library code provided by third parties.

https://github.com/zonuexe/auto-read-only.el

I'm the author.

None needed.

### Checklist

### DIFF
--- a/recipes/auto-read-only
+++ b/recipes/auto-read-only
@@ -1,0 +1,1 @@
+(auto-read-only :fetcher github :repo "zonuexe/auto-read-only.el")


zonuexe:
(I found some issues in this package.)

zonuexe:
Several problems I had were resolved.

purcell:
Thanks. A couple more small issues:
- You have an invalid trailing `-*-` on the first line
- I think you should be modifying `find-file-hook`, not advising the `find-file` function

zonuexe:
@purcell thank you for your review!

I fixed issues.
https://github.com/zonuexe/auto-read-only.el/commit/16e7240593a1d9a866ed0a3811dc164abfe2bd70 https://github.com/zonuexe/auto-read-only.el/commit/bfa6ac11da9c05ffe7a4206effeca60010eb5046

marsam:
Hi:
what differences are this package from [hardhat](https://github.com/rolandwalker/hardhat)?

tarsius:
@zonuexe: @marsam is asking that because when another package already provides the same functionality, then it might be preferable to not add a new one. But maybe your package is "better" for some reason. You should compare your package with the competition in the readme and here.

zonuexe:
@tarsius @marsam 
Hmm... I did not know hardhat, but it seems to have a function close to my ideals.
I considered shifting to hardhat for several days.
But, hardhat seems to be doing too complicated things for compatibility.  It has defects that interfere with behavior such as `package-install`.

I decided to maintain auto-read-only for myself.
https://github.com/zonuexe/auto-read-only.el/commit/79654f8fc024f383ae7af05487c1345738236500

purcell:
Thanks for the explanation. In this case, I'm happy to include this package in addition to hardhat.

PR #4604, opened 2017-03-04
Add magit-lfs

Ailrun:

Integrate git-lfs with magit

https://github.com/ailrun/magit-lfs

Maintainer

**None needed**

### Checklist


purcell:
Thanks. Quick feedback:
- Please just use the default file list (ie. remove `:files` completely). We wouldn't normally include files like `CHANGES` in the built packages.
- Don't depend on a snapshot version of `magit`: use the latest MELPA Stable version instead.
- The custom group description for `magit-lfs` is empty.

Ailrun:
@purcell 

I fix it. Thank you for feedback.

Anyway, I can't access melpa-stable version of `magit` via emacs and [melpa-stable](https://stable.melpa.org/) page with my firefox.
So I just use latest release version of `magit` in [magit release page](https://github.com/magit/magit/releases).

Is there any way to solve these problems?

purcell:
It looks like the disk is full on the MELPA Stable server, sorry. I’ll look into that and get things working again.

purcell:
Hey @milkypostman, don't suppose you can see what's filled up the MELPA Stable server. It has 11GB in "working", just like the main MELPA server, and nothing else is obviously huge, so I'm unsure how it has filled 20GB, or how to fix this. Can you bump up the disk size maybe?

tarsius:
* You are using but not requiring `dash`. (Okay so `magit` does that for you, but since it is a third-party package you should still make this explicit. I for one like to be able to immediately tell if a library uses thing s like `dash` and `cl-lib`.)
* It seems unnecessary to put some stuff into a separate `magit-lfs-core.el`.

Ailrun:
@tarsius I update it.

tarsius:
There's an extra dash in the copyright line: `;; Copyright (C) 2017- Junyoung Clare Jang`, probably a typo. You should remove that or tools that try to extract the year(s) will choke on it.

Ailrun:
@tarsius I fix it too. Thanks.

PR #4606, opened 2017-03-05
Add recipe for pomidor

TatriX:

Pomidor is a simple pomodoro timer.

https://github.com/TatriX/pomidor

I'm a maintainer.

**None needed**

### Checklist

### DIFF
--- a/recipes/pomidor
+++ b/recipes/pomidor
@@ -1,1 +1,1 @@
-(pomidor :fetcher github :repo "TatriX/pomidor" :files ("pomidor.el" "tick.wav" "tack.wav" "overwork.wav"))
+(pomidor :fetcher github :repo "TatriX/pomidor" :files (:defaults "tick.wav" "tack.wav" "overwork.wav"))
--- a/recipes/pomidor
+++ b/recipes/pomidor
@@ -1,0 +1,1 @@
+(pomidor :fetcher github :repo "TatriX/pomidor" :files ("pomidor.el" "tick.wav" "tack.wav" "overwork.wav"))


recipes/pomidor
```diff
@@ -0,0 +1 @@
+(pomidor :fetcher github :repo "TatriX/pomidor" :files ("pomidor.el" "tick.wav" "tack.wav" "overwork.wav"))
```
syohex: Please use `:defaults` instead of `pomidor.el`.

syohex:
I have sent PR about packaging issue. Please check it(https://github.com/TatriX/pomidor/pull/2)

purcell:
Thanks. Quick feedback from me:
- There's no standard "productivity" custom group. Please choose a different parent, e.g. `applications` or `environment`.
- Prefer `defconst` to `defvar` for values like `pomidor-dir` etc.
- Don't incude a font or width in `pomidor-time-face`: those are very personal and platform-specific settings. Also, don't set `:foreground` unless you are going to provide variants for `dark` and `light` backgrounds. Consider just `:inherit`-ing standard faces like `warning`, `success`, `font-lock-comment-face` by default, so that themes can easily work.
- Use the `async` package to play the sound asynchronously: it will handle creating a child process for you.
- You should pass the `LOCAL` flag to `(add-hook 'kill-buffer-hook ...), otherwise you'll be cancelling the timer when _any_ buffer is killed.

Hope that helps!

purcell:
Also, there was an excellent suggestion on Reddit that you use `alert` to notify the user: this is definitely the way to go. I'm not sure why `(alert "test" :style 'libnotify)` didn't work for you, but you generally shouldn't hard-code the style: `(alert "test")` should do the right thing on all platforms, and if you provide `:severity` and/or `:category`, users can apply their own preferences to control how the notifications appear.

TatriX:
@purcell thank you. I've pushed some changes based on your feedback.

Where can I find info about `async`?

How do  you think `alert` should be used from the user side? e.g. how can I disable notification? How can I adjust what and how is shown?

TatriX:
So I should call from my side:
```list
(alert "Make a pause" :severity 'high :category 'pomidor)
```
This way users can control appearance or for example ignore the notification. But how should I make the message customizable? Probably I should add a function which will be called to produce a message. Then user will be able to override it.

purcell:
> This way users can control appearance or for example ignore the notification.

Yes, exactly. 

> But how should I make the message customizable? Probably I should add a function which will be called to produce a message. Then user will be able to override it.

Maybe use a `defcustom` to define a variable for each message?

Regarding `async` and the sounds, I think you would just use one of the `async` functions to call `play-sound-file` asynchronously. Check out the README.

TatriX:
Hm, I've tried async and it seems it's worse then current implementation.
It has problems with passing params:
```lisp
(async-start (lambda () (play-sound-file pomidor-sound-overwork)) 'ignore)
;; error in process sentinel: Symbol’s value as variable is void: pomidor-sound-overwork
```
Then it's a dependency, when actually I need only 3 lines of code.
Maybe I'm doing something wrong?

purcell:
Yeah, so that lambda will run in a separate process, where `pomidor-sound-overwork` will not be defined. You'd need to use [async-inject-variables](https://github.com/jwiegley/emacs-async#async-inject-variables). If you can get it to work, it's still better than rolling your own sub-process execution code IMO.

TatriX:
I read async-start code and it's almost identical to my own solution (for my "play audio file" case).
So I think making a dependency on the library to make such a simple thing isn't the best idea. Less dependencies — less complexity. IMO ;)

What else should be resolved to merge this PR?

recipes/pomidor
TatriX: It's done, but I think github already told you so.

syohex:
Merged

PR #4607, opened 2017-03-06
add helm spotify plus to melpa

wandersoncferreira:
Improved version of a old package called `helm-spotify. 
Helm is used only here to narrow the candidates we get from the Spotify API instead of directing communicating to the API through the candidate buffer.

https://github.com/wandersoncferreira/helm-spotify-plus

I am one of the maintainers.

None needed.

### Checklist

### DIFF
--- a/recipes/helm-spotify-plus
+++ b/recipes/helm-spotify-plus
@@ -1,0 +1,2 @@
+(helm-spotify-plus :repo "wandersoncferreira/helm-spotify-plus"
+                   :fetcher github)


purcell:
Thanks! Quick feedback:

- Which package is the "multi" library from? It feels like you're missing a dependency.
- You're missing docstrings for (at least) a bunch of the `interactive` commands. (Hint: `M-x checkdoc`)
- Do you need all of `helm`, or would the `helm-core` package be sufficient? I expect it would.
- Don't use `setq` to introduce otherwise-unbound local variables like `track-name`, `new-url`: instead, use `let` or `let*`.

wandersoncferreira:
Hi @purcell ,

- The "multi" is a package itself. It provides the **defmulti-method** that I am using.

- Thanks for the tips about the docstring, they are fixed.

- Yes, helm-core is already enough. I fixed that too.

- I changed the setq. I am using let instead now. Thanks.

- There were a couple of setq to define magic numbers that are already in defvar forms.

Hope everything is fine now. Thanks!

syohex:
No, there is no package dependencies. Please check https://github.com/wandersoncferreira/helm-spotify-plus/pull/18.

And this package cannot be  byte-compiled. Please check too.

```
helm-spotify-plus.el:48:47:Error: Wrong type argument: listp, "Play app in OS systems."
```

wandersoncferreira:
@syohex thanks for your help.

I fixed the problems to byte-compile the code.

purcell:
Awesome, thanks. There's an extra unrelated commit in this pull-request, though, for `python-experiment`. Please can you rebase the pull request to remove that commit, then force-push it back here so we can cleanly merge it?

wandersoncferreira:
@purcell I was preparing another submission and I added the python-experiment commit by mistake. It was too late at night to think about rebasing hehehe. It is correct now. Thanks!

wandersoncferreira:
Do I need to do anything else to have this PR merged?

purcell:
Merged now, thanks for your patience.

PR #4610, opened 2017-03-07
add recipe for docker-robe

aki2o:

A plugin lets robe.el work under docker container.

https://github.com/aki2o/emacs-docker-robe

I'm a maintainer.

**None needed**

### Checklist

### DIFF
--- a/recipes/docker-robe
+++ b/recipes/docker-robe
@@ -1,0 +1,2 @@
+(docker-robe :repo "aki2o/emacs-docker-robe"
+             :fetcher github)


syohex:
- Please use `-` separator instead of `:`(also use `--` instead of `::` for private symbols) for emacs lisp convention
- There is no autoload cookie
- Missing cl-lib dependency in dependency header
- Please put docstring to entry point functions at least

Silex:
Just wanted to mention https://github.com/emacs-pe/docker-tramp.el allows to run shells inside containers, might simplify the implementation.

aki2o:
fixed.
But, not changed separator because I'd like to use not `-` as separator when the package name includes `-`.

I know it's recommended.
But, I think it can't force that to third party plugin like this pull request.
At present, other many plugins exists which uses not `-` separator char.

Silex:
https://yourlogicalfallacyis.com/bandwagon :wink:

purcell:
Where there is a clear community standard, choosing your own preference anyway means you're not interested in the community being able to work smoothly with your code. And that's fine, but if that's the case, why share that code with the community via MELPA? ;-)

tarsius:
As @purcell said, you are of course free to break with the conventions if you don't like them. But it is okay if we don't include the package here because you have done so.

PR #4611, opened 2017-03-07
Replace direnv package with alternative implementation

wbolster:
This changes the direnv package to use this one instead:

https://github.com/wbolster/emacs-direnv

...which is fine according to the author of the current melpa direnv package:

https://github.com/jml/direnv-el/issues/3#issuecomment-284720244


syohex:
Merged

wbolster:
done.

https://melpa.org/#/direnv

PR #4612, opened 2017-03-07
Add org-wikish

Benaiah:

Simple Org wiki system.

https://github.com/benaiah/org-wikish

I wrote `org-wikish` over the last couple weeks.

**None needed**

### Checklist

### DIFF
--- a/recipes/org-wikish
+++ b/recipes/org-wikish
@@ -1,0 +1,1 @@
+(org-wikish :fetcher github :repo "benaiah/org-wikish")


syohex:
- `org-wikish-wiki-directory` should be custom variable(defined by `defcustom`) and its autoload cookie is not necessary
- `org-wikish-link-word-at-point-and-enter` is used but it is not defined.

Benaiah:
@syohex just pushed fixes for those.

syohex:
Sorry, I should have explained in more detail.

- Please define group by `defgroup` for `defcustom` variables and specify its type attribyte(`:type`)
- I suppose `org-wikish-wiki-directory` should have better default value. Current default value `/your/directory/here/` causes error in most case(because there is no such directory). For example, `~/.emacs.d/org-wikish-wiki`(Use `user-emacs-directory` instead of `~/.emacs.d`) etc.

purcell:
> Use user-emacs-directory instead of ~/.emacs.d

See also the useful function `locate-user-emacs-file`.

tarsius:
Upstream still hasn't responded to my pull request implementing our suggestions, so I am closing this now.

PR #4614, opened 2017-03-08
Add dired-hide-dotfiles recipe

mattiasb:

Hide dotfiles in dired.

https://github.com/mattiasb/dired-hide-dotfiles

I'm the maintainer.

**None needed**

### Checklist

- [ ✓ ] I've read [CONTRIBUTING.md](https://github.com/melpa/melpa/blob/master/CONTRIBUTING.md)
- [ ✓ ] I've used [package-lint](https://github.com/purcell/package-lint) to check for packaging issues, and addressed its feedback
- [ ✓ ] I've built and installed the package using the instructions in the [README](https://github.com/melpa/melpa/blob/master/README.md)

### DIFF
--- a/recipes/dired-hide-dotfiles
+++ b/recipes/dired-hide-dotfiles
@@ -1,0 +1,2 @@
+(dired-hide-dotfiles :fetcher github
+                     :repo "mattiasb/dired-hide-dotfiles")


syohex:
I think [dired-filter](https://github.com/Fuco1/dired-hacks#dired-filter) provides same feature and more features about filtering. How about ?

Review memo
- This package does not use any Emacs 25.1 features, it works on Emacs 24.1

mattiasb:
Are there an easy way to check what version is required for your code?

Yeah I tried dired-filter before writing this. It has a rather advanced concept of filtering that felt interesting but kind of got in the way for me. All i wanted was a simple toggle-mode, similar to dired-omit but for dot-files.

mattiasb:
Pushed a `25.1` → `24.1` Requires-fix.

purcell:
> Are there an easy way to check what version is required for your code?

You'd really need to test with other versions to be sure. But `package-lint` will at least warn you about uses of functions/libraries added in Emacs versions newer than the one in your `Package-Requires` header.

mattiasb:
Yeah that's what I gathered. I don't have easy access to older versions of Emacs though. Personally I'd prefer just going back to the 25.1 require, since that's what I've tested for, but either is fine with me.

purcell:
Cool, then we're good to merge, but I'd ask you to first remove the redundant `:files` spec in the recipe entirely please.

mattiasb:
Ok, so basically:
1. Fix the recipe
2. Go back to 25.1?

mattiasb:
Done in the package [here](https://github.com/mattiasb/dired-hide-dotfiles/commit/8e5946573b765fc90f3a9d4f7e98572b8b540ee0) and force-pushed an updated recipe.

syohex:
https://github.com/mattiasb/dired-hide-dotfiles/blob/8e5946573b765fc90f3a9d4f7e98572b8b540ee0/dired-hide-dotfiles.el#L62

Is this line regexp `"^\\."`, not `"^\\\."` ?

mattiasb:
I believe you have to escape both the backslash and the dot. That is `\\` → `\` and `\.` → `.`.

xuchunyang:
@mattiasb The regexp is `^\.`, in Elisp string, it written as `"^\\."`, though `"^\\\."` is the same, but you don't need (and should not) to escape `.` in Elisp string, since it is confusing.

```elisp
(string= "^\\." "^\\\.")
     => t
```

mattiasb:
You are correct of course. :)

Fixed!

syohex:
Merged

mattiasb:
Thanks! :)

Boruch-Baum:
@syohex wrote: https://github.com/melpa/melpa/pull/4614#issuecomment-285665418

> I think [dired-filter](https://github.com/Fuco1/dired-hacks#dired-filter) provides same feature and more features about filtering. How about ?

This package can result in a TRIPLE over-do, because emacs itself has `dired-omit-mode` built-in, so a user could potentially end up being confused by the three packages. Package `dired-filter` does warn that its own omit-mode filter should not be used in conjunction with `dired-omit-mode` itself.

PR #4616, opened 2017-03-09
Add recipe for historian

raxod502:

Historian.el stores the results of completing-read and similar
functions persistently. This provides a way to give completion
candidates that are more frequently or more recently used a better
position in the candidates list.

https://github.com/PythonNut/historian.el

I'm an enthusiastic user---I had been wanting something that would
properly generalize `smex` for quite a while.

**None needed**

### Checklist

### DIFF
--- a/recipes/historian
+++ b/recipes/historian
@@ -1,0 +1,2 @@
+(historian :fetcher github :repo "PythonNut/historian.el"
+           :files ("historian.el"))


purcell:
This is nice. I wonder if @abo-abo is aware of your work on this library?

manuel-uberti:
Out of curiosity: does this make `smex` obsolete?

abo-abo:
@purcell I wasn't aware, thanks for the pointer. There was some discussion for a similar feature at https://github.com/abo-abo/swiper/issues/436, but it's not finalized.

raxod502:
@manuel-uberti Unfortunately, not quite yet. It only needs one additional feature, though: displaying the last few selected items at the top of the list when there is no query typed in. This should be a pretty easy feature to add, however. Once I have time I would like to make it a complete replacement for `smex` and also write some documentation. Neither me nor the author tend to have oodles of free time, unfortunately.

PR #4620, opened 2017-03-11
add python-experiment to melpa

wandersoncferreira:
This mode was created to easily generate a Python Inferior session with several objects already filled. Something I really do everyday is to help people using Pandas, numpy or other libraries. It is nice to be able to have a mock-example every time at hand to show up that cool feature you just learned to do.

In this version that is already a Faker object loaded in case you have this library installed in your environment.

https://github.com/wandersoncferreira/python-experiment

I'm the maintainer.

None needed.

### Checklist

Please confirm with `x`:

Hope to have learned a lot from my previous submission of helm-spotify-plus. Everything checked. Thanks for the tips!

### DIFF
--- a/recipes/python-experiment
+++ b/recipes/python-experiment
@@ -1,0 +1,2 @@
+(python-experiment :repo "wandersoncferreira/python-experiment"
+                   :fetcher github)


syohex:
- This package cannot work with older python-mode. Minimum package version should be specified(https://github.com/wandersoncferreira/python-experiment/pull/8)
- No autoload cookie
- [These configurations](https://github.com/wandersoncferreira/python-experiment/blob/cb54de08b0f9504d24054404c73b27aa0e2b018b/python-experiment.el#L174-L178) should be provided by minor-mode or setup function. Because this part overrides user configuration(global key map) only by loading.

wandersoncferreira:
@syohex thanks for the suggestions. The behavior I want is to be able to use the keys either in **python-mode** buffers as in **inferior-python-mode** buffers. I used the _python-mode-hook_ and _inferior-python-mode-hook_ to accomplish that. If you have any better suggestion I would love to hear.

autoload added too!

Thanks again!

wandersoncferreira:
Sorry, do I need to review anything else? I think I fixed the points above. Thanks!

purcell:
Quick feedback:
- The body of the `Commentary` section should start with `;;`, not `;;;`
- I can't tell from the package description or the `Commentary` anything about what the package is for, so you're unlikely to find any interested users of this package without providing better information
- Provide a minor mode with its own keymap: don't add keybindings to `python-mode`. In any case, many of the function keys are [reserved for end-user customisation](https://www.gnu.org/software/emacs/manual/html_node/elisp/Key-Binding-Conventions.html), and you should not bind them on the user's behalf.
- There's a lot of unidiomatic use of `nil`, `t` etc. in conditionals.
- Don't keep using `(concat user-emacs-directory ...)`. Use `(locate-user-emacs-file ...)` instead, and make functions or variables to reduce duplication.
- The `vc-backend` stuff is very special-case and inappropriate.

Overall this doesn't look mature enough for wider publishing, I think: it looks closer to a few helpers for your personal workflow. I'm going to close this for now, then, sorry.

wandersoncferreira:
Not a problem at all!! Really thanks for your feedback. Still learning about [E]Lisp in general. I will work on the points you highlighted.


Appreciate you took some time to provide me these guidances.

PR #4621, opened 2017-03-13
Added recipe for lastpass

storvik:

A LastPass command line wrapper that uses lpass command line utility to manage passwords in Emacs. 

https://github.com/storvik/emacs-lastpass

Maintainer

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/elpass
+++ b/recipes/elpass
@@ -1,2 +1,0 @@
-(elpass :repo "storvik/elpass"
-        :fetcher github)
--- a/recipes/elpass
+++ b/recipes/elpass
@@ -1,0 +1,2 @@
+(elpass :repo "storvik/elpass"
+        :fetcher github)
--- a/recipes/lastpass
+++ b/recipes/lastpass
@@ -1,0 +1,2 @@
+(lastpass :repo "storvik/emacs-lastpass"
+          :fetcher github)


syohex:
- Current package description is `Elisp LastPASS`.  It should be `lastpass command wrapper` or better description

storvik:
@syohex consider it fixed!

purcell:
Thanks!

The name "elpass" is a bit generic and confusing, particularly since there are other elisp packages which integrate with the "pass" tool. Would you consider renaming this package to simple "lastpass"? (You can still call the containing repo something like "emacs-lastpass" if you prefer.)

Regarding the code, I feel like you should be using `shell-quote-argument` when building those command line strings. Looks good otherwise!

storvik:
@purcell Thanks for the feedback. 

I will rename the package to "lastpass" and update the pull request. 

Didn't know about the `shell-quote-argument` before, but it seems to fit this perfectly. I'll try to get this implemented as soon as possible.

storvik:
@purcell I have updated the pull request and added shell-quote-argument. Hope you find that everything is in order now.

purcell:
Yes, looks great - sorry for the delay!

storvik:
No problem, thank you very much for the feedback. Useful for a new elisp developer!

syohex:
I think `lastpass-location` is not good for users. If users install lastpass command by package manager except homebrew(default configuration), it is installed at `/usr/bin` usually. Such users need to change `lastpass-location` value in current implementation.

It is not necessary to specify full path for executing external command if `PATH` and `execute-path` are set correctly(`/usr/bin`, `/bin`, `/usr/local/bin` are set by default usually). I think that `lastpass-location` should be `"lastpass"`( or `(executable-find "lastpass")`) rather than directory path.

storvik:
That's a valid point. I'll rewrite this part to use executable-find. Thanks!

purcell:
Just `"lastpass"` would be better than using `executable-find`. They're pretty much equivalent, except that the latter "locks-in" the location at the time it is called based on the $PATH at the time. So if someone changes `exec-path` (or $PATH) after `lastpass-location` has been defined using `executable-find`, then it won't work.

storvik:
That's probably the best solution. I think I'll remove the location variable entirely, as it's no need to have this option.

PR #4624, opened 2017-03-14
New: hack-time-mode

marcowahl:

Fork time (to some extend) 

https://gitlab.com/marcowahl/hack-time-mode

maintainer+contributor+enthusiastic user

**None needed**

### Checklist

Please confirm with `x`:

Warnings

### DIFF
--- a/recipes/hack-time-mode
+++ b/recipes/hack-time-mode
@@ -1,1 +1,1 @@
-(cask-mode :repo "marcowahl/hack-time" :fetcher gitlab)
+(hack-time-mode :repo "marcowahl/hack-time-mode" :fetcher gitlab)
--- a/recipes/hack-time-mode
+++ b/recipes/hack-time-mode
@@ -1,0 +1,1 @@
+(cask-mode :repo "marcowahl/hack-time" :fetcher gitlab)


syohex:
- Some functions are undefined, please check function names by byte-compiling or using all features of this package
- You should load `org.el` for using `org-read-data`. It is not an autoload function.

marcowahl:
Thanks for your suggestions!

- AFAICT hack-time-mode now also byte-compiles cleanly. 
- And I put a (require 'org) immediately in front of the call to org-read-data.

purcell:
I'll go ahead and merge this, but some feedback:

- For `org-read-data`, don't `(require 'org)`: instead, use `declare-function` at the top level.
- Consider making `target-date` a parameter to `hack-time-mode-set-current-time`: you can then change the `interactive` form to `(interactive (list (org-read-date)))` and everything will work the same as currently, but more flexibly.
- Note that your `prologue ends here` is currently part of the `Commentary` section, and will be included in the extracted documentation for the package which is displayed on the MELPA page and via the Emacs package list buffer when pressing `?`.
- Don't indent the subsequent lines in the minor mode docstring

marcowahl:
Thanks for your feedback!  I think I fixed these issues.

Best regards and thanks for including the package into melpa.

PR #4626, opened 2017-03-15
add recipes pocket-mode

lujun9972:

[Manage your pocket]

https://github.com/lujun9972/pocket-mode

I am the maintainer

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/pocket-mode
+++ b/recipes/pocket-mode
@@ -1,0 +1,1 @@
+(pocket-mode :fetcher github :repo "lujun9972/pocket-mode")


syohex:
- `tabulated-list-print` can accept only one optional argument on older Emacs(24.4, 24.5). This causes exception in `pocket-refresh`
- I suppose `list-pocket` should be `pocket-list` for package naming convention.

lujun9972:
+ I have removed the second argument of `tabulated-list-print`
+ according to http://www.gnu.org/software/emacs/manual/html_node/elisp/Tabulated-List-Mode.html#Tabulated-List-Mode.
```
The derived mode should also define a listing command. 
This, not the mode command, is what the user calls (e.g., M-x list-processes). 
The listing command should create or switch to a buffer, turn on the derived mode, specify the tabulated data, and finally call tabulated-list-print to populate the buffer. 
```
So I think `list-pocket` is the way following naming convention

purcell:
`pocket-list` would still be preferable, please, and will not produce `package-lint` warnings.

In the meantime, I'll go ahead and merge this, which looks like a handy package!

PR #4628, opened 2017-03-15
Add recipe for comint-intercept

hying-caritas:

Intercept the input in comint-mode.  This can be used to run eshell command, or run some command in a terminal buffer from the command line in shell buffer.

https://github.com/hying-caritas/comint-intercept

I am the maintainer.

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/comint-intercept
+++ b/recipes/comint-intercept
@@ -1,0 +1,1 @@
+(comint-intercept :repo "hying-caritas/comint-intercept" :fetcher github)


syohex:
- Don't load `cl-extra`. It is bundled since Emacs 24.5. Your package requires Emacs 24.3 or higher versions but 24.3 and 24.5 cannot load your package by it.

hying-caritas:
Thanks for pointing this out.  I had used cl-some before, but now I don't really need cl-extra anymore.  So I removed cl-extra requiring.

purcell:
Thanks. Can you add a docstring for the minor mode, please?

I feel like all the configuration for this is a little inflexible. Did you consider just having an alist of `(regexp . runner-function)` pairs? The regexp would let you match `e.*` or `top`, and `runner-function` would be passed the command. You'd provide functions like `comint-intercept-run-in-eshell` etc.

hying-caritas:
There's a default docstring for the minor mode.  You think that is not good enough.  In addition to the contents of default one, which other information do you want?  Some brief introduction for the functionality?

For the configuration, Yes, it is inflexible.  I done that mainly for simplicity, performance and requirement.  The current configuration is easy to customize, has good performance (because only 4 regular expressions are needed to match, instead of tens).  And I have no clear requirement for more complex configuration.  If you want something more flexible for your usage, I can add another configuration method as you suggested.  What do you think about that?

purcell:
> There's a default docstring for the minor mode. You think that is not good enough. In addition to the contents of default one, which other information do you want? Some brief introduction for the functionality?

You'd usually just provide a sentence about what the minor mode is for, e.g. `"Intercept comint input and send it to other buffers."`

> If you want something more flexible for your usage, I can add another configuration method as you suggested. What do you think about that?

It's not about _my_ usage, but you're sharing this package on MELPA so that it will hopefully be useful for many other people, and when I review code I think about whether it is likely to work well for most people. In this case, I wasn't sure, so I was suggesting you consider how other people might want to work.

hying-caritas:
>> There's a default docstring for the minor mode. You think that is not good enough. In addition to the contents of default one, which other information do you want? Some brief introduction for the functionality?
>
> You'd usually just provide a sentence about what the minor mode is for, e.g. "Intercept comint input and send it to other buffers."

Sure.  I will add it.

>> If you want something more flexible for your usage, I can add another configuration method as you suggested. What do you think about that?
>
> It's not about my usage, but you're sharing this package on MELPA so that it will hopefully be useful for many other people, and when I review code I think about whether it is likely to work well for most people. In this case, I wasn't sure, so I was suggesting you consider how other people might want to work.

Got it.  I will add another more flexible configuration method to make complex thing possible.  And make necessary helper functions public.

hying-caritas:
I just updated the code accordingly.

purcell:
Looks good, thanks - merged!

PR #4630, opened 2017-03-15
Add recipe for quoted-scratch

narendraj9:

This package lets you add new quotes to a collection of Quotes. Displays random quote from the collection in the scratch buffer. It also allows you to have a quote from htttp://quote.rest/qod.json in your scratch buffer on startup. I have also included some Auroville qualities with the quote text in Chinese and English.

https://github.com/narendraj9/quoted-scratch

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/quoted-scratch
+++ b/recipes/quoted-scratch
@@ -1,0 +1,4 @@
+(quoted-scratch :fetcher github
+                :repo "narendraj9/quoted-scratch"
+                :files ("*.el*"
+                        ("_assets/" "_assets/quotes.txt")))


syohex:
- Load url.el for using its features
- Please use `-` separator instead of `/` for coding convention

narendraj9:
@syohex Fixed the above two issues. :) Thanks.

syohex:
- There is an issue on older Emacs(https://github.com/narendraj9/quoted-scratch/pull/1)

narendraj9:
@syohex Merged your pull request. Thanks! :)

narendraj9:
@syohex Is there anything else I should change?

purcell:
Quick feedback:
- Use `:defaults` in the recipe, please, rather than `"*.el*`
- Don't use the `qs-` prefix: use `quoted-scratch-`
- When setting faces, don't specify only one colour: either provide `dark` _and_ `light`-background versions of the face, or (better) `:inherit` from a standard face so that the default appearance matches the user's theme.
- Most of the code in this package is a reinvention of code in `fortune.el`, particularly the `fortune` function. Please just use the code there instead.

The particular combination of quote and Auroville quality makes this feel like a _very_ personal set of configuration, so I'm not sure I can see this being widely appealing, and my current feeling is that you should probably keep it in your own Emacs config rather than trying to package it here.

narendraj9:
@purcell The Auroville quality can be hidden with a custom parameter that is provided. I think you have a valid point. This might not be useful for a wide set of people.

purcell:
I was thinking about this: the generalisation would be to create a kind of `scratch-format` package, which would provide a `scratch-format` variable a little like the existing `mode-line-format`. Users could customise that variable as desired (including with `:eval` forms), and the package would ensure that the variable was re-applied to the scratch buffer when changed.

A recently-added package which is a little similar is [scratch-message](https://github.com/thisirs/scratch-message/blob/master/scratch-message.el), by @thisirs. That package is quite similar to this one, including a predefined selection of quotes.

Rather than everyone reinventing the wheel, perhaps it'd be worth you and @thisirs building upon `scratch-message` so that it works for both of you, possibly using the format variable idea above? I'd be really keen to see what you might come up with.

In any case, there are so many overlapping scratch-related packages (see also [fortune-cookie](https://github.com/andschwa/fortune-cookie), for example) that I'm not keen to add this one too, sorry.

narendraj9:
@purcell The generalisation idea is really cool. Thanks! :+1:

PR #4633, opened 2017-03-16
Add terminal-here

davidshepherd7:

Add commands to open an external terminal emulator in the current directory or current project.

https://github.com/davidshepherd7/terminal-here

Maintainer

**None needed**

### Checklist

Please confirm with `x`:

- [x ] I've built and installed the package using the instructions in the [README](https://github.com/melpa/melpa/blob/master/README.md)

### DIFF
--- a/recipes/terminal-here
+++ b/recipes/terminal-here
@@ -1,0 +1,2 @@
+(terminal-here :repo "davidshepherd7/terminal-here"
+               :fetcher github)


syohex:
- There is no autoload cookie

davidshepherd7:
Thanks, added them.

Would it be reasonable to add a check for some autoloads existing in `package-lint`? Or might some packages not need autoloads?

purcell:
It's actually hard to know which functions might need autoloading, otherwise I'd have added the check to `package-lint`. It's also indeed the case that some package might legitimately not need any autoloads.

purcell:
Thanks. Sent you a [PR with a few fixes](https://github.com/davidshepherd7/terminal-here/pull/1).

davidshepherd7:
Hey, I think this is ready to merge now.

tarsius:
> ;; Copyright (C) 2015 Free Software Foundation, Inc.

Did you write this in 2015? Did you actually assign the copyright to the fsf?

davidshepherd7:
> Did you write this in 2015? Did you actually assign the copyright to the fsf?

Ah, no I copied the boilerplate from somewhere, fixed now.

How do I actually go about assigning copyright? I sent an email to the mailing list once (as instructed in [the manual](https://www.gnu.org/software/emacs/manual/html_node/emacs/Copyright-Assignment.html#Copyright-Assignment) but no one ever replied to me. Do I just keep requesting it until someone responds?

tarsius:
> How do I actually go about assigning copyright?

Well, if you don't actually plan to add your packages to Emacs or GNU Elpa (which is considered "part of Emacs"), there's not much to gain from this. It would however make it harder for you to accept contributions from other people who have not signed the same thing.

> Do I just keep requesting it until someone responds?

Yeah, I must just have fallen though the cracks.

tarsius:
... much like this pr.

Silex:
Ah, great package! I had this idea since like 2 years ago now :disappointed: 

Thanks for doing it :+1:

PR #4634, opened 2017-03-16
Add evil-lion package

edkolev:

Defines a minor mode (evil-lion-mode) which adds two evil-mode operators `gl` and `gL` for left and right align. Port of [vim-lion](https://github.com/tommcdo/vim-lion)

https://github.com/edkolev/evil-lion

Maintainer.

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/evil-lion
+++ b/recipes/evil-lion
@@ -1,0 +1,3 @@
+(evil-lion :repo "edkolev/evil-lion"
+           :fetcher github
+           :files ("evil-lion.el"))


edkolev:
There's one byte-compile error, which I'm not sure how to best fix - this new package uses `(align-region)` from emacs' `align.el`. But the `align-region` function has no autoload cookie.

The approach I took is, right before calling:
```
      (when (not (fboundp 'align-region))
        (require 'align))
      (align-region .........)
```

But maybe this is a better appraoch:
```
(autoload 'align-region "align.el")
```

However, the second approach prints an annoying message on first use `uncompressing align.el.gz...done`. That's why I went with the first approach.

syohex:
- Dependency header is insufficient(https://github.com/edkolev/evil-lion/pull/1)
- Why don't you load `align.el` at top level ?

edkolev:
> Why don't you load align.el at top level ?

If possible, I would like to load `align.el` the first time the `gl`/`gL` evil operator is pressed, instead of when `(evil-lion-mode)` is called.

I'll try to split evil-lion.el into two files (evil-lion-core.el and evil-lion.el), maybe this way I'll be able to achieve the desired behaviour.

edkolev:
The byte compile warning is now silenced with `(declare-function align-region "align")` in https://github.com/edkolev/evil-lion/commit/af12db645c50c3e677ed6287265989d745143e59

purcell:
> The byte compile warning is now silenced with (declare-function align-region "align") in edkolev/evil-lion@af12db6

Then now you don't need to `(require 'align-region)` at all. You can drop the `fboundp` conditional wrapper around `declare-function` too.

edkolev:
@purcell thanks - the `fboundp` is now removed https://github.com/edkolev/evil-lion/commit/039c82f5183a01b6e008b1028b48f1ec33f74eff

However, I can't get align.el to be loaded without this:
```
      (when (not (fboundp 'align-region))
        (require 'align))
```

The `align-region` function isn't autoloaded (no autoload cookie in align.el) - maybe this is the reason `declare-function` isn't sufficient to load align.el?

purcell:
Sorry, you're right. But you can use `(autoload 'align-region "align")` instead of `declare-function`. I think it's a little nicer than loading a feature explicitly from inside a function.

(P.S. `unless` is nicer than `(when (not ...) ...)`)

edkolev:
Thanks!

Also thanks for `unless` :) https://github.com/edkolev/evil-lion/commit/875ca7bde23d6102356675af0e6c27775eb21d17

PR #4635, opened 2017-03-17
Adding recipe for org-sync-snippets

abrochard:

Simple extension to export snippets to org-mode and vice versa.

https://github.com/abrochard/org-sync-snippets

Maintainer

None needed

### Checklist

Please confirm with `x`:

NB: follow up to [#4631](https://github.com/melpa/melpa/pull/4631) to remove exta commits.

### DIFF
--- a/recipes/org-sync-snippets
+++ b/recipes/org-sync-snippets
@@ -1,0 +1,1 @@
+(org-sync-snippets :repo "abrochard/org-sync-snippets" :fetcher github)


purcell:
Thanks. Quick feedback:

- You haven't defined the `'oss` custom group with `defgroup`. In any case, it should be called `org-sync-snippets`, not `oss`.
- Please use `org-sync-snippets-` as the symbol prefix throughout, rather than `oss-`.
- Rather than `(if (not COND) BODY)`, just use `(unless COND BODY)`
- When writing snippets to the org file, it would be more idiomatic to use something like `with-temp-file` and `insert`, rather than building a big string with `concat` and then finally dumping it.
- Turn the output org filename and directory into function parameters so that the functions are more widely useful. You can then provide wrapper commands which apply the defaults.
- Instead of `(concat user-emacs-directory "snippets/")`, use `(locate-user-emacs-file "snippets")`, or at least use `expand-file-name` rather than `concat`.

Hope that helps!

abrochard:
Thank you so much for the feedback. It is really helpful!
I applied the changes that you suggested. 
Also, I was under the impression that `insert` was less efficient than `f-write`. Is it true but we prefer something more idiomatic? Or was I just mislead?

tarsius:
> I was under the impression that `insert` was less efficient than `f-write`.

`f-write` is a wrapper around `insert`, so no.

abrochard:
Oh wow. Ahahah ok. Thanks for the clarification.

purcell:
Nice! Merging now.

abrochard:
Thank you!

PR #4636, opened 2017-03-17
Add recipe for easy-hugo

masasam:

Emacs package for writing blogs made with hugo

https://github.com/masasam/emacs-easy-hugo

I'm the author

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/easy-hugo
+++ b/recipes/easy-hugo
@@ -1,0 +1,3 @@
+(easy-hugo
+ :repo "masasam/emacs-easy-hugo"
+ :fetcher github)


syohex:
- I suppose you should use `delete-directory` function instead of `rm -rf` for portability(like Windows etc)

masasam:
Thank you for review.
I fixed it.

purcell:
Thanks. Some feedback:

- The parent group of the `easy-hugo` custom group is currently defined as itself. Change it to a pre-existing group, e.g. `'tools`.
- `(when (null COND) BODY)` is much more clearly written as `(unless COND BODY)`
- Use `shell-quote-argument` when building file paths: if `easy-hugo-root` contains a space or quote, your current code will fail badly.
- Don't build file paths with `concat`. Use `expand-file-name` instead.

masasam:
Thank you for review.
I fixed it.

masasam:
- Don't build file paths with concat. Use expand-file-name instead.
I am sorry, I re-corrected this, so please review again

purcell:
Thanks. I sent you a fix in https://github.com/masasam/emacs-easy-hugo/pull/1, mainly for missing `:group` declarations in your `defcustom` forms.

PR #4637, opened 2017-03-17
Remove counsel-ripgrep

julienXX:
Hi I'm the maintainer of counsel-ripgrep and I deleted this this package since ripgrep support is already in counsel. It should be removed from Melpa.


purcell:
Thanks.

PR #4638, opened 2017-03-18
Rename flow-mode into flow-minor-mode.

an-sh:

A package rename, based on users feedback/patches https://github.com/an-sh/flow-minor-mode/issues/2 . Also deletes the old one to free the name, not sure how to do it correctly.

### DIFF
--- a/recipes/flow-minor-mode
+++ b/recipes/flow-minor-mode
@@ -1,0 +1,4 @@
+(flow-minor-mode
+ :fetcher github
+ :repo "an-sh/flow-minor-mode"
+ :old-names (flow-mode))
--- a/recipes/flow-mode
+++ b/recipes/flow-mode
@@ -1,1 +1,0 @@
-(flow-mode :fetcher github :repo "an-sh/flow-mode")


syohex:
Could you  add `:old-names` attribute and squash commits ?

purcell:
It's not clear to me that you'd need to rename the whole package, but it's fine if that's your preference. But if so, then `flow-minor-mode.el` should now `provide` the new feature name, not `flow-mode`, and you'll need to fix the header and footer comments there which currently still mentions `flow-mode.el`.

an-sh:
I've renamed everything to flow-minor-mode, also can i keep `flow-` prefix for package symbols or should i rename them too?

purcell:
> I've renamed everything to flow-minor-mode, also can i keep flow- prefix for package symbols or should i rename them too?

If you're renaming to "free up" the `flow-mode` name for someone else to use, you should really leave the `flow-` prefix for them, and use `flow-minor-` or `flow-minor-mode-` here.

an-sh:
Sorry for the delay, prefix should be fixed by now.

purcell:
Great, thanks.

PR #4647, opened 2017-03-22
Add uptimes.el

davep:

uptimes.el provides a simple system for tracking and displaying the uptimes of your Emacs sessions. Simply loading uptimes.el from your ~/.emacs file (or equivalent) will start the tracking of any session.

https://github.com/davep/uptimes.el

Maintainer?

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/uptimes
+++ b/recipes/uptimes
@@ -1,0 +1,1 @@
+(uptimes :fetcher github :repo "davep/uptimes.el")


purcell:
Thanks. Quick feedback:
- Don't hard-code `~/.emacs.uptimes`: use `(locate-user-emacs-file ".emacs.uptimes")` instead. (Since the file contains sexps, you might consider calling it `.uptimes.el`.)
- Don't use the obsolete and deprecated `cl` library. You're using `cl-flet`, but that's declared in `cl-lib` these days: you should require that library instead, and declare `(cl-lib "0.5")` in a `Package-Requires` header line.
- Don't shim `defgroup` and `defcustom`. There's no remotely recent Emacs version that doesn't have them.
- Maybe make `uptimes-save-uptimes` interactive too. (And `uptimes-save` would be a sufficiently clear name.)
- `uptimes-this` isn't a great name, IMO. How about `uptimes-current` instead?
- Why the `(eval-when (load eval) ...)` block? Feels like it could just be a `progn`.

Hope that helps!

davep:
That's great feedback @purcell -- very helpful. Most of it betrays the fact that this code is around 18 years old (as is my rusty elisp knowledge; the last time I coded in elisp in a non-trivial way `cl` was still a thing, for example, and emacsen that lacked`defgroup` and `defcustom` were a common occurrence).

Couple of things:

> Maybe make `uptimes-save-uptimes` interactive too. (And `uptimes-save` would be a sufficiently clear name.)

Given the age of the code, and given that it's been out there and in use for a long time, I'm loath to change the name of functions now. I guess I could provide an alias, of course.

> `uptimes-this` isn't a great name, IMO. How about `uptimes-current` instead?

Ditto this.

> Why the `(eval-when (load eval) ...)` block? Feels like it could just be a `progn`.

I believe the idea was to ensure that uptime tracking only took place when the file was loaded or the content of the buffer was evaluated -- but not when the code was being compiled. My recollection is somewhat fuzzy at this point, I have to admit. I'll look back over this and see if it makes sense.

davep:
I'd also note, while I appreciate this advice:

> Since the file contains sexps, you might consider calling it .uptimes.el.)

I'm not sure it's a good plan to have the code ignore people's existing uptime history if they should install the next version of the file. (the advice on how to better locate it in the first place is, of course, spot on).

davep:
Actually, given that `locate-user-emacs-file` will result in the default database location possibly being different anyway... I might as well change the name and also make some of the other breaking changes mentioned and bump the major version number.

davep:
All suggested changes have been implemented in [v3.0 of uptimes.el](https://github.com/davep/uptimes.el/releases/tag/v3.0). Hopefully this will mean it's in a good state to go into melpa.

Thanks again for the feedback.

purcell:
Looks great, thanks! I might have to install this myself...

PR #4650, opened 2017-03-23
Add org-edit-latex package

et2010:

org-edit-latex is an extension for org-mode. It let you edit a latex fragment just like editing a src block.

https://github.com/et2010/org-edit-latex

Yes, I'm the maintainer, the contributor and an enthusiastic user 😄 

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/org-edit-latex
+++ b/recipes/org-edit-latex
@@ -1,0 +1,1 @@
+(org-edit-latex :repo "et2010/org-edit-latex" :fetcher github)


syohex:
- [This line](https://github.com/et2010/org-lfe/blob/f6fe19e01d0dc064c5878590f5ef23811fa02891/org-lfe.el#L43) may break user configuration only by loading. If you need to modify it, you should provide minor-mode and set it as buffer local variable
- [These configurations](https://github.com/et2010/org-lfe/blob/f6fe19e01d0dc064c5878590f5ef23811fa02891/org-lfe.el#L111-L112) are enabled only by loading. You should provide minor-mode or setup function

et2010:
Thanks for reviewing! I'll try to fix those problems.

et2010:
Fixed! @syohex

et2010:
BTW I'm not so sure about the package name. Which one do you think is better? org-latex-editor or org-lfe?

purcell:
Just use `define-minor-mode` instead of recreating the toggle behaviour with `org-edit-latex-toggle`!

Also, your current behaviour is global, because your function advice does not check if the toggle is enabled in the *current* buffer.

I've sent you a PR containing an untested fix for these issues: https://github.com/melpa/melpa/pull/4650#issuecomment-288680598

et2010:
Merged! Thanks for the fix, it works like a charm! @purcell

purcell:
Great, thanks.

PR #4651, opened 2017-03-23
vimgolf.el now lives at the root

timvisher:


purcell:
Thanks! Would you mind also removing the `:files` spec entirely now, since the default will do the trick nicely?

timvisher:
There you are, my friend. :)

purcell:
Perfect, thanks Tim.

PR #4652, opened 2017-03-23
Add recipe for wotd

cute-jumper:

Show word-of-the-day from various online sources

https://github.com/cute-jumper/emacs-word-of-the-day

I'm the maintainer.

None needed.

### Checklist

Please confirm with `x`:

**QUESTION**: I use the acronym `wotd` as the package prefix since I think `word-of-the-day` is a little bit too long. `package-lint` complains about that. Is this not acceptable?

### DIFF
--- a/recipes/wotd
+++ b/recipes/wotd
@@ -1,0 +1,2 @@
+(wotd :repo "cute-jumper/emacs-word-of-the-day"
+      :fetcher github)


purcell:
Thanks!
- I think you should simply rename the file/package/recipe to `wotd`.
- `eval` should be unnecessary, and expands/executes the `wotd--def-parser` macro at runtime. Just modify `wotd--def-parser` to generate the "get" functions directly, e.g. `wotd--get-merriam-webster`.

cute-jumper:
@purcell Thanks, Steve. I've updated the package, the file name and the recipe. Let me know if there is any other problem.

purcell:
Great, thanks!

PR #4653, opened 2017-03-25
Add obfusurl.el

davep:

`obfusurl.el` provides `obfusurl`, a command that will obfuscate an
URL under the cursor. This might be useful if you are writing out an URL
for someone but the URL itself might spoil the surprise.

For example, this:

```
<URL:http://www.davep.org/emacs/>
```

is turned into this:

```
<URL:http://www.davep.org/%65%6d%61%63%73/>
```

https://github.com/davep/obfusurl.el

Maintainer

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/obfusurl
+++ b/recipes/obfusurl
@@ -1,0 +1,1 @@
+(obfusurl :fetcher github :repo "davep/obfusurl.el")


purcell:
Nice, thanks!

PR #4656, opened 2017-03-26
Add quick-peek

cpitclaudel:

A utility library to display inline pop-ups.  Extracted from company-coq.

https://github.com/cpitclaudel/quick-peek/

Creator

### DIFF
--- a/recipes/quick-peek
+++ b/recipes/quick-peek
@@ -1,0 +1,1 @@
+(quick-peek :fetcher github :repo "cpitclaudel/quick-peek")


syohex:
- I have sent PR about package dependencies(https://github.com/cpitclaudel/quick-peek/pull/1)

cpitclaudel:
Thanks, I merged it.  I don't think it needs autoloads.

alphapapa:
> I don't think it needs autoloads.

The user-facing commands should have autoloads, otherwise they won't be available to the user unless he runs `(require 'quick-peek)`.  :)

cpitclaudel:
I'm happy to add autoloads, but I'd like to understand better: since none of these commands are interactive, they will all be used from scripts; given this, why are autoloads needed?  (I just looked at similar 'library' packages, like `pos-tip` and `dash`, and neither of these have autoloads)

alphapapa:
LOL, AFAIK you're right.  I was thinking about the commands I shared on the other issue.  Sorry.  :)

cpitclaudel:
No, I think it's fair.  The "proper" way to do this would be to put your code in a separate file (with autoloads on the interactive commands) and that file would require `quick-peek`.  In practice that sounds like a pain, so the alternative would be to `(require 'quick-peek)` in each of your function, but that's a pain too (admittedly milder).  So I added autoloads instead :)

alphapapa:
> The "proper" way to do this would be to put your code in a separate file (with autoloads on the interactive commands) and that file would require quick-peek. 

That is indeed the proper way, and I'm pretty sure that that's what you should do.  @syohex may correct me if I'm wrong, as he is far more of an expert than I am.

> In practice that sounds like a pain

Not really, it's the standard way to do it.  You actually don't want autoloads on non-interactive functions, because that way your package's loading can be deferred until it's required by another package.  If every function in every package had an autoload, every package would load when `(package-initiallize)` was run, IIUC.

cpitclaudel:
>> The "proper" way to do this would be to put your code in a separate file (with autoloads on the interactive commands) and that file would require quick-peek.

> That is indeed the proper way, and I'm pretty sure that that's what you should do. @syohex may correct me if I'm wrong, as he is far more of an expert than I am.

Sorry, I think you misunderstood me.  I was talking about *your* code (the snippets you posted on the quick-peek tracker).  There are no interactive commands in quick-peek, so there's nothing to move to a separate file — and it would not make sense to move them even if there were interactive commands, since quick-peek is already a proper package (unlike the typical init file)

>> In practice that sounds like a pain

> Not really, it's the standard way to do it.

Not for snippets in an init file (I thought that was your use case?).

> You actually don't want autoloads on non-interactive functions, because that way your package's loading can be deferred until it's required by another package. If every function in every package had an autoload, every package would load when (package-initiallize) was run, IIUC.

That's incorrect.

alphapapa:
> Sorry, I think you misunderstood me. I was talking about your code (the snippets you posted on the quick-peek tracker). There are no interactive commands in quick-peek, so there's nothing to move to a separate file — and it would not make sense to move them even if there were interactive commands, since quick-peek is already a proper package (unlike the typical init file)

Actually I understood, but I did not write clearly, sorry.  :)  Yes, I meant that I should have the autoloads on my interactive commands, therefore you don't need autoloads in your package.

> Not for snippets in an init file (I thought that was your use case?).

I guess this is a matter of preference.  :)  I use `use-package`, so I can do something like:

```elisp
(use-package quick-peek
  :config (defun my-quick-peek () (interactive) ...))
```

That way I don't need an explicit `(require 'quick-peek)`.  I guess if someone is posting random functions from their init file, it might be helpful to others if they `require` any packages they need, but I prefer not to do it that way.

> That's incorrect.

I'm not sure exactly what you mean.  What I mean is, if every function in every package had an autoload, (as opposed to only those that need it) it would make `package-initialize` slower.

Sorry, I haven't been communicating very clearly tonight!  D:

cpitclaudel:
> I'm not sure exactly what you mean. What I mean is, if every function in every package had an autoload, (as opposed to only those that need it) it would make package-initialize slower.

Technically yes, by a very small amount.  It wouldn't cause every package to be loaded on (package-initialize), though.

cpitclaudel:
@purcell Ping?

purcell:
Thanks, merged - sorry for the delay!

cpitclaudel:
Thanks you!

PR #4657, opened 2017-03-26
added swagger-mode

Nooby:

Compile [Swagger API](http://swagger.io/) definitions and preview the API help in a browser. 

https://github.com/Nooby/swagger-mode

Author

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/swagger-mode
+++ b/recipes/swagger-mode
@@ -1,0 +1,3 @@
+(swagger-mode
+ :repo "Nooby/swagger-mode"
+ :fetcher github)


syohex:
- I have sent PR about setting hook(https://github.com/Nooby/swagger-mode/pull/1)
- Please don't use `C-c` and a letter binding(like `C-c c`, `C-c v` for Emacs Lisp keybinding convention(https://www.gnu.org/software/emacs/manual/html_node/elisp/Key-Binding-Conventions.html)
- BTW space indentation is better than tab indentation if you set `tab-width` other than 8. Indentations of this package are strange with default Emacs configuration.

Nooby:
Thank you @syohex for the review and the PR.

- I Merged your PR. The after-save-hook is now buffer local.
- I changed the `C-c` keybinding to `M-s` (so `C-c c` is now `M-s c`)
- I changed the Tabs to spaces. (and fixed my Emacs settings to only use spaces from now on)

purcell:
- `M-s` is a default keybinding, which you should avoid shadowing. Follow the guidelines in the [key binding conventions](https://www.gnu.org/software/emacs/manual/html_node/elisp/Key-Binding-Conventions.html). Ideally you would define a prefix keymap to contain "c", "e" and "p", and then you can bind that prefix map to a permitted single configurable prefix like `C-'`, or leave the user to define that.
- `(bound-and-true-p swagger-mode)` can just be `swagger-mode`, because that symbol will _always_ be bound when that code is executed.
- `(if (not cond) body)` can more neatly be written `(unless cond body)`
- This package is specific to `swagger-codegen`. To avoid suggesting that it does everything swagger-related, I think you should rename the package to `swagger-codegen.el`.

Hope that helps!

tarsius:
@Nooby friendly ping!

tarsius:
I am closing this due to long silence. Please feel free to reopen this once the above issue have been addressed.

PR #4659, opened 2017-03-27
Add recipe for helm-navi-outshine

alphapapa:

This package lets you navigate through a buffer using the headings and keywords provided by `outshine` and `navi-mode`, presented in a Helm buffer.

https://github.com/emacs-helm/helm-navi-outshine

Author and maintainer.

**None needed**

### Checklist

Please confirm with `x`:

### Note minor problem

The package builds and installs correctly for me, but unfortunately I'm having a minor problem with the package descriptor which causes it to not install correctly in the sandbox, because no dependencies are listed:

```elisp
(helm-navi-outshine .
                    [(20170327 353)
                     nil "No description available." tar nil])
```

I can't figure out why it's leaving out the dependencies and package description.  The package headers are correct, and there are no flycheck-package/package-lint errors.

Thanks for your help.

### DIFF
--- a/recipes/helm-navi
+++ b/recipes/helm-navi
@@ -1,0 +1,1 @@
+(helm-navi :fetcher github :repo "emacs-helm/helm-navi")


syohex:
You need to put/rename a file(`helm-navi-outshine.el`)  same as package name or put `*-pkg.el` file into repository.

purcell:
`helm-navi` and `helm-outshine` should be separate packages, as far as I can tell. I suggest you put each in its own package, and submit two recipes.

purcell:
I haven't reviewed the code in detail yet, but I can immediately see some other little issues, e.g. `cl-lib` should be `require`-d explicitly for `cl-loop`, and some of the `defvar` should be `defcustom` and specify their containing group, otherwise `defgroup helm-outshine` is a bit pointless.

alphapapa:
Thanks @syohex and @purcell for your help.  I've renamed the package and fixed the problems you mentioned, sending new PR now.

PR #4665, opened 2017-03-29
Makefile: Add INSTALL argument to sandbox target

alphapapa:
e.g. run `make sandbox INSTALL=helm` to make the sandbox and automatically install the "helm" package.  This saves the user from having to wait for `package-initialize` to finish, manually run `package-install`, and finally type in the package name and hit `RET`.

When trying to debug issues with packages installed into a sandbox, it becomes tedious doing all that manually over and over again.  This makes it much faster, since you can just run it directly from the shell history.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -112,11 +112,12 @@ packages will be available for installation along with those already
 in MELPA:
 
 ```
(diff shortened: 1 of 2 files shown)


purcell:
Nice, thanks.

PR #4666, opened 2017-03-30
Add recipe for kubernetes-el

chrisbarrett:
This package adds a live-updating interactive UI for managing Kubernetes pods. It uses Magit's rendering and popup systems, and usage should be pretty self-explanatory for Magit users.

I'm the maintainer. Repo here: https://github.com/chrisbarrett/kubernetes-el

- I've run checkdoc and purcell's package linter, and only get a spurious warning about explicitly specifying the Emacs version.

- I've verified the melpa recipe builds locally on my fork.

Let me know if you need any more information to proceed here. :)

### DIFF
--- a/recipes/kubernetes
+++ b/recipes/kubernetes
@@ -1,0 +1,1 @@
+(kubernetes :repo "chrisbarrett/kubernetes-el" :fetcher github)


purcell:
Thanks. I'll go ahead and merge, but a couple of points of feedback:
- Though small, `kubernetes-evil` should ideally be a separate little package, since it will cause byte-compilation issues at installation time for non-evil users. It's trivial to accomplish this, so why not? Then you can avoid the hacky soft-`require` and use of `eval-after-load` (the latter of which should generally not be used in libraries).
- If possible, `:inherit` from standard faces for all your faces, so that themes work out of the box.

chrisbarrett:
Thanks for the feedback, I'll issue a separate PR to separate the packages.

purcell:
Thanks. You can keep them in the same repo, and just add a `:files ("the-file.el")` spec to each recipe: once you've added dependencies on `kubernetes` and `evil` for `kubernetes-evil`, everything should work nicely. (Not sure if splitting this will affect your Cask config though.)

PR #4668, opened 2017-03-31
Added recipe for julia-repl.

tpapp:
Run an inferior Julia REPL in a terminal inside Emacs.

The github repository is at https://github.com/tpapp/julia-repl.

I am the original author/maintainer.

### Checklist

Please confirm with `x`:

- [X ] I've read [CONTRIBUTING.md](https://github.com/melpa/melpa/blob/master/CONTRIBUTING.md)
- [ X] I've used [package-lint](https://github.com/purcell/package-lint) to check for packaging issues, and addressed its feedback
- [ X] I've built and installed the package using the instructions in the [README](https://github.com/melpa/melpa/blob/master/README.md)

### DIFF
--- a/recipes/julia-repl
+++ b/recipes/julia-repl
@@ -1,0 +1,1 @@
+(julia-repl :fetcher github :repo "tpapp/julia-repl")


alphapapa:
Hi, I saw a few small things you might want to change.  I sent you a couple as comments on commits, here are a few more:

* You should `defgroup` a group before the first `defcustom`.
* Not sure, but `(define-key term-raw-map (kbd "M-x") #'execute-extended-command)` might not be a good idea, because I think that will affect every buffer using that map.

I don't know anything about Julia, but looks like you did a great job!

tpapp:
Thanks for your help. In the latest 2 commits, I added the `defgroup`, and at least documented the issue with `M-x` and made it customizable. Since `ansi-term` shares maps between buffers, I see no other workaround for this.

alphapapa:
> Since ansi-term shares maps between buffers, I see no other workaround for this.

Can you copy the ansi-term map, modify it, and set your package's terminal buffers to the copy?

tpapp:
Unfortunately the fix is not that simple. `ansi-term` has two functions, `term-char-mode` and `term-line-mode` that switch between the two maps which are hardcoded variables. I could add a `defadvice` to functions using `term-char-mode`, but that would be an extremely brittle and ugly hack. Other than a rewrite of `ansi-term`, I don't have a better idea than what I am doing currently.

alphapapa:
Yeah, sounds like it's better to leave that alone.  :)

tpapp:
This is my first "serious" Emacs library, and I appreciate all the help and patience. I am wondering if there is anything that I should do to get the library to MELPA.

alphapapa:
Usually Steve and/or Syohei will look at it and then merge the recipe when they think it's ready.  They're pretty busy, so sometimes it takes a few days.  I'm still waiting for my new package to get merged too.  :)

syohex:
Sorry belated reply

https://github.com/tpapp/julia-repl/blob/420eb955dd01f8252de8f6b81109d45699d83a7d/julia-repl.el#L44-L52

- This code is evaluated only by loading.
- It looks that it changes default behavior. I suppose it should effect only julia REPL buffer, for example checking `julia-repl-mode` etc 
- You should not use `lambda` function, and use named function. Because `lambda` function cannot be removed(it cannot be specified by name).

tarsius:
@tpapp friendly ping!

tpapp:
@tarsius : Thanks for the ping. I have to admit I forgot about this, but will update the library ASAP. 

@syhoex: Very good points. However, it looks like fixed the issue on the Julia end, so the workaround may not be needed anymore, so I can remove the whole stuff. I will experiment with the change and get back here within a few days.

tpapp:
@tarsius: I think I fixed all the issues above. I have been using the [latest master](https://github.com/tpapp/julia-repl) for a few days now without problems on various Emacs versions and it works fine. So if possible, it would be great to have it added to MELPA.

purcell:
Thanks! `"C-c m"` is a binding [reserved for end users](https://www.gnu.org/software/emacs/manual/html_node/elisp/Key-Binding-Conventions.html) and may not be used here. And the code which declares the keymap is very odd -- does it even work? Should just be a `defvar` with a block to define the map.

Looks good otherwise.

tpapp:
Fixed that, thanks.

AFAIK the alist is valid for a keymap, according to the [manual](https://www.gnu.org/software/emacs/manual/html_node/elisp/Defining-Minor-Modes.html).

purcell:
Wow, that's pretty special! :-)  Merging now...

PR #4669, opened 2017-03-31
Added recipe org-brain

Kungsgeten:

Uses org-mode files to create a concept map. Its like a combination of a wiki and a mind map. It is heavily inspired by the software The Brain.

https://github.com/Kungsgeten/org-brain

I'm the developer and maintainer.

None needed

### Checklist

Please confirm with `x`:

The byte-compiles (and package-lint) complains about using `goto-line`, since it is meant for interactive use only. The documentation says that its "usually wrong", but in my use case (generate a buffer's content from scratch) it works fine.

### DIFF
--- a/recipes/org-brain
+++ b/recipes/org-brain
@@ -1,0 +1,1 @@
+(org-brain :repo "Kungsgeten/org-brain" :fetcher github)


purcell:
Nice, thanks.

PR #4671, opened 2017-04-02
add recipe for 'challenger-deep-theme' package

MaxSt:

Dark theme for Emacs

https://github.com/MaxSt/challenger-deep

Maintainer

 **None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/challenger-deep-theme
+++ b/recipes/challenger-deep-theme
@@ -1,0 +1,1 @@
+(challenger-deep-theme :fetcher github :repo "MaxSt/challenger-deep")


purcell:
Thanks for this, but all those `window-system` tests are incorrect and undesirable. If you have multiple variants of the faces for different displays, the way to handle it is to put all the variants in the face definitions, so that Emacs will choose the right ones according to the frame on which the face is later displayed.

So after loading the theme in an emacs instance, the correct colours will be displayed both in graphical and non-graphical frames connected to that instance. You'll be able to see that's not the case here by launching a graphical Emacs frame, loading the theme, and then connecting to the same Emacs in a terminal using emacsclient -t.

The emacs manual has more on this: https://www.gnu.org/software/emacs/manual/html_node/elisp/Defining-Faces.html#Defining-Faces

(Additionally, the theme description line says "Orbit Theme", which appears to be incorrect.)

MaxSt:
Thank you for your detailed feedback! 👍 
I just pushed a new version with support for 256 color displays

purcell:
@MaxSt Very nice!

PR #4677, opened 2017-04-02
Transfer maintainership of outshine, outorg, and navi-mode

alphapapa:
Thorsten has graciously agreed to transfer maintainership of the Outshine/Outorg/Navi packages to me. tj64/outshine#64


purcell:
Thanks!

PR #4679, opened 2017-04-04
Added recipe for playerctl.el

thomasluquet:

- [Playerctl](https://github.com/acrisci/playerctl) let you control your music player (Spotify, vlc, audacious, bmp, xmms2, and others) from shell.
- playerctl.el is just a simple binding to play your music from emacs

https://github.com/thomasluquet/playerctl.el

Creator

None needed

### Checklist

### DIFF
--- a/recipes/playerctl
+++ b/recipes/playerctl
@@ -1,0 +1,1 @@
+(playerctl :repo "thomasluquet/playerctl.el" :fetcher github)


syohex:
- There are some byte-compile warnings
   - I think you can use `start-process` instead of `start-process-shell-command` for avoiding warnings
   - And 2nd argument of `start-process-shell-command` is buffer name. You should set `nil`(buffer is unused) or better name(like `*playerctl*` etc)

thomasluquet:
Thank you for your feed back,
i apply your recommandation , but i am not sure for the autoload cookies
(it is my first lisp project !)

purcell:
> [x] I've used package-lint to check for packaging issues, and addressed its feedback

Strange - it should have told you that the first line was invalid! `Playerctl.el` should be `playerctl.el`.

Additional feedback:
- You should run `M-x checkdoc` and fix up the docstrings as directed.
- Your minor mode declaration appears to be pointless, since you're recommending that people modify `global-map` in order to bind keys to the `playerctl` functions.

thomasluquet:
I used package-lint but it don't find issues.

I have change the autoad and the 'P' case in comment 😄

purcell:
> I used package-lint but it don't find issues.

Thanks - that was actually due to [an Emacs bug](https://debbugs.gnu.org/cgi/pkgreport.cgi?include=subject%3Apackage-buffer-info;package=emacs)!

I sent you a few fixes in https://github.com/thomasluquet/playerctl.el/pull/1.

thomasluquet:
I have merge. Thank you for everything . i have learn a lot.

PR #4686, opened 2017-04-07
Add add-hooks

nickmccurdy:

Function for setting multiple hooks

https://github.com/nickmccurdy/add-hooks

Author and maintainer

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/add-hooks
+++ b/recipes/add-hooks
@@ -1,0 +1,1 @@
+(add-hooks :repo "nickmccurdy/add-hooks" :fetcher github)


purcell:
Thanks. There's no reason for that to be a macro, though, and it just makes the code more brittle: if it were a simple function, you'd be able to pass in a quoted list, the result of a function call, or a variable containing a list of hooks to add.

Also, please include docs in the `Commentary` section rather than referring people to the README, which won't be included in the package. Better yet, include some examples in the docstring for `add-hooks`.

nickmccurdy:
Thank you for your feedback, I converted it to a function taking a quoted list and updated its docstring to include the rest of the readme contents.

purcell:
Yay, much better! To be consistently prefixed, and match up with the package name, you should rename `add-hook*` to `add-hooks*`. (Also, the current example for `add-hook*` uses `add-hooks` instead of `add-hook*`.)

nickmccurdy:
Thanks. I intentionally named `add-hook*` that way since it's a modification of `add-hook`, and I didn't want to imply that it's a modification of `add-hooks` (since `add-hooks` uses `add-hook*` internally). Do you think `add-hooks-pair` would be a more appropriate name for `add-hook*`? You're totally right about the example typo, I just fixed that.

purcell:
> Do you think add-hooks-pair would be a more appropriate name for add-hook*?

Yes, that would be good.

nickmccurdy:
I renamed `add-hook*` to `add-hooks-pair`.

purcell:
Awesome - merged!

PR #4695, opened 2017-04-10
Add evil-org-improved

Somelauw:

Integrate evil-mode and org-mode by defining key bindings and motions.

https://github.com/Somelauw/evil-org-improved

I'm the maintainer. It was originally forked from https://github.com/edwtjo/evil-org-mode by @edwtjo, but by now it has diverted significantly from the original.

The discussion about forking started here https://github.com/edwtjo/evil-org-mode/issues/39

### Checklist

Please confirm with `x`:

- [ ] I've used package-lint to check for packaging issues, and addressed its feedback
There are some warnings if that matters, but most of them were from before I forked it. I'll try to get rid of them.

### DIFF
--- a/recipes/evil-org-improved
+++ b/recipes/evil-org-improved
@@ -1,0 +1,1 @@
+(evil-org-improved :fetcher github :repo "Somelauw/evil-org-improved")


purcell:
You can't use the same identifiers in a forked package, sorry, so you'll need to change all your `evil-org-*` to `evil-org-improved-*`. Every package must have a unique namespace.

Forks, in general, are something we try hard to avoid, because they're a confusing outcome for the community.

The *best* solution here would be if @edwtjo is happy to nominate your fork as the new official maintained version of `evil-org`, which seems like a good idea to me. You'd then rename your fork back to `evil-org.el`, and we would update our existing `evil-org` recipe to point at your fork. How about it, @edwtjo ?

Somelauw:
@purcell hmm, I'm not not yet sure about the namespace title, because "evil-org-improved-" is a bit long.

edwtjo:
Sure thing, what do you guys need from me, an updated README.md in my repo?

edkolev:
@Somelauw do you think the dependency on evil-leader can be removed or at least made optional?

Somelauw:
@edkolev It should already be an optional dependency. Let me know if you are unable to use it without evil-leader. I'll remove it from "Package-Requires" in the comment later.

Somelauw:
I have renamed evil-org-improved to evil-org. If I understand correctly, I should make a new pull request that updates the evil-org recipe.

purcell:
> I have renamed evil-org-improved to evil-org. If I understand correctly, I should make a new pull request that updates the evil-org recipe.

Yes please, or push different commits to this pull request branch.

You might consider recreating your repo as a fork of @edwtjo's original `evil-org-mode`, to make its history a little more clear.

Once we've updated the recipe then yes, @edwtjo, it'd be nice if you added a note to your repo.

Somelauw:
On the suggestion of @purcell, I have recreated the repo as https://github.com/Somelauw/evil-org-mode.
This has been done in a new pull request https://github.com/melpa/melpa/pull/4740 (already merged).

@edwtjo Can you add a note that links to https://github.com/Somelauw/evil-org-mode?

edwtjo:
Updated README in the old repo, thanks!

edkolev:
Sorry for polluting this thread.

@Somelauw issues aren't enabled in your forked repo. I wanted to open an issue to share some recommendations, like:
- I think having `use-package` install instructions is a good improvement, something like:
```
(use-package evil-org
  :ensure t
  :after org
  :config
  (add-hook 'org-mode-hook 'evil-org-mode))
```
- the README should list all keybindings, no need to click link "Full overview of bindings and configuration" to keythemes.org
- not all keybindings are clear to me, e.g. I need to check the source for `evil-org-delete-char` to see what it does -- users should be able to most if not all information from the README

Thanks for this package

Somelauw:
@edkolev Thanks. I have enabled issues.

> - I think having use-package install instructions is a good improvement, something like:

They look good to me. I'll add them.

I'll have to think about how to address your other points, because I want to documentation to be complete, but not overwhelming. In org-mode it's easy to fold information, but on Github everything appears on a single page.

PR #4696, opened 2017-04-10
Add `spaceline-all-the-icons' theme recipe

domtronn:

This is a theme for [Spaceline](https://github.com/TheBB/spaceline) using the [All The Icons](https://github.com/domtronn/all-the-icons.el) package.

https://github.com/domtronn/spaceline-all-the-icons.el

I am the current maintainer

**None Needed**

### Checklist

The `package-lint` returns some issues with namespacing inside [`spaceline-all-the-icons-segments.el`](https://github.com/domtronn/spaceline-all-the-icons.el/blob/master/spaceline-all-the-icons-segments.el) since the functions don't end in `-segments` but they're all namespaced with `spaceline-all-the-icons` which I wanted to do since it's shorter and it's the main package namespace anyway 😅

### DIFF
--- a/recipes/spaceline-all-the-icons
+++ b/recipes/spaceline-all-the-icons
@@ -1,0 +1,3 @@
+(spaceline-all-the-icons
+ :fetcher github
+ :repo "domtronn/spaceline-all-the-icons.el")


purcell:
You've put a dependency on `all-the-icons` in `spaceline-all-the-icons-segments.el`, but that's not the main package `.el` file, so those dependencies are ignored. MELPA only looks at dependencies in `spaceline-all-the-icons.el`, so you should move all dependencies there, and omit them elsewhere to avoid confusion.

domtronn:
Thanks! I didn't realise that's how it worked, I've updated this now

purcell:
Awesome, thanks.

PR #4699, opened 2017-04-11
add recipe for smart-hungry-delete

hrehfeld:

Hungrily deletes whitespace between cursor and next word, paren or delimiter while honoring some rules about where space should be left to separate words and parentheses.

https://github.com/hrehfeld/emacs-smart-hungry-delete

Author.

**None needed**

### Checklist

Please confirm with `x`:

 (however, I've left package-name/function-name instead of package-name-function-name, as other packages (hydra mc use-package yas) seem to do it as well and I find it to be much clearer)

Thanks!

### DIFF
--- a/recipes/smart-hungry-delete
+++ b/recipes/smart-hungry-delete
@@ -1,0 +1,1 @@
+(smart-hungry-delete :repo "hrehfeld/emacs-smart-hungry-delete" :fetcher github)


alphapapa:
> (however, I've left package-name/function-name instead of package-name-function-name, as other packages (hydra mc use-package yas) seem to do it as well and I find it to be much clearer)

Personally, I tend to agree with you, however MELPA is firm about this requirement--that's why that check was added to `package-lint`, to avoid having to manually ask authors to fix it...  ;)

purcell:
Standards like the Elisp coding conventions exist to support both tools and other contributors: consistency across the community trumps individual preferences, so we encourage authors to follow the conventions by default when publishing code here for community use.

Where there is a clear community standard, choosing your own preference anyway means the community can't work as smoothly with your code. (And that's fine, but then why share that code with the community via MELPA?) I personally like the Clojure symbol naming style: when writing Emacs packages, I follow the local standards.

Some older (and popular) packages do not follow these conventions. Those packages don't negate the best practice. Indeed some, like `yasnippet`, [have been fixed](joaotavora/yasnippet@a51adba) over time.

hrehfeld:
Fixed.

purcell:
LOL, thanks. Merging now. This looks really nice, btw.

hrehfeld:
thanks for all your work! :+1:

PR #4700, opened 2017-04-11
added monochromatic theme.

diasbruno:

Simple monochromatic theme.

[github:diasbruno/monochromatic-theme](https://github.com/diasbruno/monochromatic-theme)

Author.

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/monochromatic-theme
+++ b/recipes/monochromatic-theme
@@ -1,0 +1,1 @@
+(monochromatic-theme :repo "diasbruno/monochromatic-theme" :fetcher github)


purcell:
Thanks! A few things to address here:

- There are four monochrome themes already in MELPA. What's the difference here? Could you use/extend one of the existing themes?
- It's far from monochrome, so I'm not sure that it's as advertised! Perhaps a better name and/or description is in order.
- Yellow tooltips are not going to work if the background is yellowish.
- You should never set specific fonts in themes: on numerous Emacs platforms, Menlo and/or Arial will not be available, and themes have no business changing the user's font family and size preferences.

diasbruno:
Hi @purcell. In this theme you can choose background and foreground colors, but probably it still needs some fixes for sure (the idea is to highlight only errors and warning). I couldn't find a theme that does this, so I've created this one. 
Those fonts I actually forget to remove (thanks).

diasbruno:
Yeah, probably rename this package is better.

diasbruno:
Fixed the fonts and tooltip.
Should I close this PR and create a new one when I rename the package or should I keep this one opened?

purcell:
> Should I close this PR and create a new one when I rename the package or should I keep this one opened?

Probably just open a fresh one. I'll go ahead and close this, then. :-)

PR #4704, opened 2017-04-13
Add hierarchy

DamienCassou:

Library to create, query, navigate and display hierarchy
  structures.

http://github.com/DamienCassou/hierarchy

package maintainer

n/a

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/hierarchy
+++ b/recipes/hierarchy
@@ -1,0 +1,1 @@
+(hierarchy :fetcher github :repo "DamienCassou/hierarchy")


purcell:
Thanks, looks great!

Just one small thing: rather that specifying a `:files` spec, can I suggest you move the example `.el` file into an `examples` subdirectory in your repo?

alphapapa:
This looks really interesting!  The readme is really good!  The only thing that could make it better is if you had some example use cases (other than the hypothetical animal kingdom one ;).

purcell:
Merci Damien!

DamienCassou:
Thank you both for the feedback.

> can I suggest you move the example .el file into an examples subdirectory in your repo?

done

> had some example use cases

done. Please tell me what you think.

alphapapa:
Thanks, that helps some.  I guess I'm just a novice, but I don't understand the `parentfn` API for building the hierarchy.  It seems like an awkward way of passing the same information that a simple list of lists would have.  I'm assuming there's an academic or theoretical basis that is way over my head, but not knowing that makes it hard for me to understand how and when I would use this library.  This is just FYI, not a complaint.  :)

DamienCassou:
@alphapapa thanks for the discussion. Could you please copy/paste this text to a new issue at http://github.com/DamienCassou/hierarchy so we can discuss it there? I think it's worth thinking and talking about it.

DamienCassou:
Thanks for merging Steve (and thanks for saying "Merci" in French :-)).

PR #4707, opened 2017-04-13
add recipe jemdoc-mode

drdv:

Major mode for editting jemdoc files

https://github.com/drdv/jemdoc-mode

Author and maintainer

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/jemdoc-mode
+++ b/recipes/jemdoc-mode
@@ -1,0 +1,1 @@
+(jemdoc-mode :repo "drdv/jemdoc-mode" :fetcher github)


tarsius:
It seems @purcell has overlooked this pr, so I will get the ball rolling with some suggestions:

* You are adding faces to the custom group `jemdoc-mode-faces` without having defined that first.
* In `jemdoc-mode-highlight-*-html-text` you probably don't need `(when ... t)`, since this will only return `t` if `...` itself is non-nil. Do the callers really expect the value to be `nil`/`t` as opposed to `nil`/non-nil? (If so, you should use `and` instead of `when`, since this is about a value, not side-effects.
* Make this either more or less explicti:

    ```lisp
      (let ((region-beg)
    	(region-end))
    ```

I.e. either add to `nil`s or change to `(let (region-beg region-end)`.
* There are some trailing parens on their own lines.
* Some lines in `jemdoc-mode-edit-code-block` (and elsewhere) are to long and can easily be made shorter.
* `jemdoc-mode`'s indentation is messed up.

drdv:
@tarsius: Thanks for your helpful suggestions! I have fixed the code.

purcell:
More feedback from me:

- Don't use `package-installed-p`. What if `font-lock+` is installed manually? Instead, use `featurep`. But in fact, don't do that: just `(require 'font-lock+ nil t)` at the top level instead.
- You can use `defvar-local` where applicable instead of `defvar` + `make-local-variable`.
- `jemdoc-mode-map`, `jemdoc-mode-font-lock-syntax-table` and `jemdoc-mode-font-lock-keywords` (and possibly others) should not be buffer-local.
- `s/editting/editing/`, `s/delimeters/delimiters/` :-)
- `:group 'Faces` - `Faces` is not a predefined group, but `faces` is. In any case you should probably put this in a new `jemdoc` or `jemdoc-mode` group.
- Try to use `:inherit` for more of your faces. Defining only a foreground colour is almost certain to look bad for many people.
- `:inherit` from sensible faces. e.g. there's already an `italic` face.
- `(comment-start . "#")` shouldn't be included in `font-lock-defaults`, I think.

Hope that helps.

drdv:
@purcell: I fixed the code following your comments (they were very helpful). 

- I have added a `jemdoc-mode` group to `:group 'languages` but I am not sure whether this is reasonable
- Regarding your last comment: now I set `comment-start` using `(set (make-local-variable 'comment-start) "#")` but as far as I can tell, this is precisely what is done  in `font-lock-set-defaults` (am I missing something?)

Thanks

purcell:
> Regarding your last comment: now I set comment-start using (set (make-local-variable 'comment-start) "#") but as far as I can tell, this is precisely what is done in font-lock-set-defaults (am I missing something?)

Maybe, but the point is that `comment-start` has nothing to do with `font-lock`. In theory, you could disable `font-lock-mode`, and you'd still want comment behaviour to work the same way. That's why it's not idiomatic to use `font-lock-defaults` to set lots of variables like that.

Anyhow, merging now - thanks!

PR #4708, opened 2017-04-14
Add org-sticky-header recipe

alphapapa:
This package displays in the header-line the Org heading for the node that's at the top of the window.  This way, if the heading for the text at the top of the window is beyond the top of the window, you don't forget which heading the text belongs to.

https://github.com/alphapapa/org-sticky-header

Please confirm with `x`:

-  [x] I've built and installed the package using the instructions in the [README](https://github.com/melpa/melpa/blob/master/README.md)

Thanks.


purcell:
Thanks!

PR #4710, opened 2017-04-15
Add org-recent-headings

alphapapa:
This package lets you quickly jump to recently used Org headings using Helm, Ivy, or plain-ol' `completing-read`.

https://github.com/alphapapa/org-recent-headings

Author and maintainer.

Please confirm with `x`:

### DIFF
--- a/recipes/org-recent-headings
+++ b/recipes/org-recent-headings
@@ -1,0 +1,1 @@
+(org-recent-headings :fetcher github :repo "alphapapa/org-recent-headings")


purcell:
`(recentf)` is not a valid dependency in older Emacsen, including (I think Emacs 24.4). Now, you're requiring a relatively recent Emacs, but you should make sure that older Emacsen will understand the package metadata enough to know that they cannot install this package. In other words, you should normally add a version number to fix this. However, in this particular case, `recentf` is not a separately-installable package anyway, so you should simply remove it from `Package-Requires` -- the rule is to never depend on internal packages that are not installable from a package archive.

Regarding the code itself, I question the need for `ivy` and `helm`-specific code in there: when `ivy` or `helm` are installed, they will generally override `completing-read`. `org-recent-headings` should therefore do the job for both those completion backends, and others.

(Note that `with-eval-after-load` is made for user configs, and its use in libraries/packages is a code smell. Even though it is explicitly deferred, the code in those blocks will still produce byte compilation warnings when users don't have both `ivy` and `helm` installed.)

alphapapa:
Hey Steve,

Thanks, I wasn't sure about the `recentf` dependency.  I removed it.

Yeah, that's a good point about Helm and Ivy overriding `completing-read`.  I've had a user of one of my other packages request an Ivy version of its commands, and I would like to experiment with it myself, so having dedicated commands make that much easier (I guess the alternative would be to only have either Helm or Ivy installed at one time, which would be a pain).

Another benefit of having a Helm source built-in is that users can use that source in other Helm commands.  For example, I added it to my `helm-find-files` command that displays buffers, recent files, etc.  Considering the trial-and-error I had to go through to get the source working (and I've made Helm sources before, too), I'd like to save users from the same trouble.

I do realize that the `with-eval-after-load` is stinky.  :)  I'd prefer not to use it, but is there another way to avoid defining functions unless their dependencies are available?  ISTR a `(when (require 'ivy nil t) (defun ... ))` pattern; should I use that?

Thanks for reviewing this, your feedback is always much appreciated.

purcell:
> I do realize that the with-eval-after-load is stinky. :) I'd prefer not to use it, but is there another way to avoid defining functions unless their dependencies are available?

Yes, separate packaging! :-) While it seems overkill in this case, the correct solution is ultimately to package the `helm` and `ivy` add-ons separately. Would anyone expect a package which tracks recent org headings to provide `helm` and `ivy` backends, anyway? It seems like a muddling of the package's core purpose, in that you shouldn't have to update `org-recent-headings` when a new completion package gets popular.

> ISTR a (when (require 'ivy nil t) (defun ... )) pattern; should I use that?

That's worse, because it means your package will load `ivy` and/or `helm`, even if the user doesn't want to use those packages: remember we need to assume that users can have packages installed without necessarily using them.

I realise I'm being somewhat pedantic, so it's fine with me if you want to leave things as-is, and be willing to split out separate packages later if the `helm`/`ivy` functionality gets bigger.

I do think it's helpful to outline the above thinking, though, because elisp authors learn from code found in the wild, so we all have a responsibility to model best practices.

alphapapa:
Yeah, separate packaging would do it, but it seems like overkill for what would be essentially a single function in each one.

> That's worse, because it means your package will load ivy and/or helm, even if the user doesn't want to use those packages: remember we need to assume that users can have packages installed without necessarily using them.

Ah, good point.  It seems strange to me to have Helm or Ivy installed without having them loaded, but you're right that we shouldn't assume or force that.

> I realise I'm being somewhat pedantic, so it's fine with me if you want to leave things as-is, and be willing to split out separate packages later if the helm/ivy functionality gets bigger.

I think that's what I'd prefer right now.  I don't expect the helm/ivy-specific part to grow much, if at all.

> I do think it's helpful to outline the above thinking, though, because elisp authors learn from code found in the wild, so we all have a responsibility to model best practices.

Absolutely!  I've learned so much from looking at other packages.

Thanks for your help as always.

purcell:
Thx!

PR #4711, opened 2017-04-16
Added recipe for exsqlaim-mode

ahmadnazir:

Minor mode for sql-mode that makes it possible to **use variables inside sql queries**.

Here is an example:

```
@db   = `test`
@id   =  1234
@name = "John Doe"

SELECT * FROM @db.users WHERE id = @id OR name = @name;
```

https://github.com/ahmadnazir/exsqlaim-mode

Maintainer

Not applicable

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/exsqlaim-mode
+++ b/recipes/exsqlaim-mode
@@ -1,0 +1,3 @@
+(exsqlaim-mode
+ :repo "ahmadnazir/exsqlaim-mode"
+ :fetcher github)


alphapapa:
Hey, looks very nice!  Few things I noticed:

* `lexical-binding: t` must be set in the first line of the file. See https://www.gnu.org/software/emacs/manual/html_node/elisp/Using-Lexical-Binding.html#Using-Lexical-Binding

* You should do the `add-hook` when your minor mode is activated, and remove it when deactivated.  As it is now, you add the hook when your package is loaded, and it's never removed.

* You don't need to autoload every function.  In this package, I think the only thing that needs the autoload cookie is the `define-minor-mode`.

purcell:
Thanks! I agree with @alphapapa's notes above, except for the `add-hook` point: that's equivalent to unconditionally enabling `exsqlaim-mode` globally, so you should instead either provide a global minor mode (in addition to the regular minor mode) or - probably better - just tell users to add the hook themselves.

More feedback from me:
- Somewhere you should note that this is a minor mode which should be enabled in `sql-mode`. Or, perhaps you should turn this into a major mode which is a submode of `sql-mode`?
- You're not handling the case where a user disables the mode: you should restore the `font-lock-keywords`.
- Define the font lock keywords using a variable so that you don't have to include the regexp literal in the minor mode definition.

ahmadnazir:
Thanks for the feedback @alphapapa and @purcell 

As suggested, I have made the following changes now:

- `exqlaim-mode` is not enabled by default. The user has to do it. Tests and README file has been updated accordingly.
- `lexical-binding: t` is moved to the first line. Maybe, this should be reported
- autoload annotations have been removed
- `font-lock` keywords are being removed when the mode is disabled
- Regexp literal isn't being used inside the minor mode definition

purcell:
Thanks, looking much better! Couple more little points now:
- `(bound-and-true-p exsqlaim-mode)` can just be `exsqlaim-mode`, since that var will _always_ be `boundp` when that code is executed.
- [This block](https://github.com/ahmadnazir/exsqlaim-mode/blob/b6ae8e58e797ddd3de8455ba052d450f40840973/exsqlaim-mode.el#L118-L121) looks strange because it will get executed immediately after [this line](https://github.com/ahmadnazir/exsqlaim-mode/blob/b6ae8e58e797ddd3de8455ba052d450f40840973/exsqlaim-mode.el#L126) anyway - it looks redundant.
- You shouldn't need to call `font-lock-fontify-buffer` explicitly.

ahmadnazir:
Thanks for the feedback @purcell,

Pull request updated except that I need to call font-lock-fontify-buffer explicitly. This is a visual thing - the tests still pass without fontifying the buffer. But you can see the problem when you manually enable the mode and the variables aren't highlighted immediately even if the font-lock is updated.

purcell:
Awesome, thanks - merged!

PR #4713, opened 2017-04-18
Add frame-mode

IvanMalison:
`frame-mode ` configures `display-buffer-alist` so that calls to display-buffer that result in a buffer being displayed somewhere that does not replace the current buffer always use a different frame instead of using a different window.

I am aware of the existence of https://github.com/davidshepherd7/frames-only-mode , but that package did not suit my needs.

This package differs from that one in a few ways:

- frame-mode uses display-buffer-alist which would seem to be a less brittle way to approach the problem than installing a bunch of hooks everywhere
- frame-mode doesnt try to destroy frames as agressively as frames-only-mode
- frames-only-mode seems to include a lot of preferences that are peculiar to the author's setup. frame-mode seeks to be more general and it really only provides the frames instead of windows behavior

https://github.com/IvanMalison/frame-mode

Maintainer

- [x] I've built and installed the package using the instructions in the [README]

### DIFF
--- a/recipes/frame-mode
+++ b/recipes/frame-mode
@@ -1,0 +1,1 @@
+(frame-mode :fetcher github :repo "IvanMalison/frame-mode")


alphapapa:
Hi, nice job!  Couple of things I noticed:

* AFAIK you don't need to declare `cl-lib` as a dependency because it's part of Emacs.
* You declare a dependency on Dash but there's no `require` for it.

IvanMalison:
> It doesn't look like you remove your frame-mode-display-buffer-alist-entry from display-buffer-alist when the mode is deactivated.

Yeah, that is by design. The predicate function checks if frame-mode is acting before actually doing anything.

> AFAIK you don't need to declare cl-lib as a dependency because it's part of Emacs.

It was introduced in 24.3 https://elpa.gnu.org/packages/cl-lib.html I had the requirement at 24.1

> You declare a dependency on Dash but there's no require for it.

Yeah I decided not to use it.

purcell:
Thanks!

PR #4714, opened 2017-04-18
Add info file to ledger-mode recipe

JalapenoGremlin:

This is just a modification to an existing recipe for ledger-mode. This change only adds the "doc/*.texi" files to the recipe so that the ledger-mode.info file can be generated, as well.

https://github.com/ledger/ledger-mode

I'm just a user, who likes documentation with the code.

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/ledger-mode
+++ b/recipes/ledger-mode
@@ -1,1 +1,1 @@
-(ledger-mode :fetcher github :repo "ledger/ledger-mode" :files ("*.el") :old-names (ldg-mode))
+(ledger-mode :fetcher github :repo "ledger/ledger-mode" :old-names (ldg-mode))


purcell:
I think the solution here is to simply remove `:files` altogether, which will result in all those files being includes. Would you mind testing that and amending the recipe accordingly please?

JalapenoGremlin:
Yep. That does exactly the correct thing.  So.... (my github fu may be off).

purcell:
Super, thanks!

jaccarmac:
@JalapenoGremlin This recipe change is causing problems for those trying to install `ledger-mode`, as the mode depends on loading a `ledger-test.el` which is excluded by the default recipe setting from MELPA. I'm in the middle of provisioning a new dev machine so can't add to the PR myself just yet, but this patch should be pretty close:

```
modified   recipes/ledger-mode
@@ -1 +1,7 @@
-(ledger-mode :fetcher github :repo "ledger/ledger-mode" :old-names (ldg-mode))
+(ledger-mode :fetcher github
+             :repo "ledger/ledger-mode"
+             :files ("*.el" "*.el.in" "dir" "*.info" "*.texi" "*.texinfo"
+                     "doc/dir" "doc/*.info" "doc/*.texi" "doc/*.texinfo"
+                     (:exclude ".dir-locals.el" "test.el" "tests.el"
+                               "*-tests.el"))
+             :old-names (ldg-mode))
```

Thanks!

purcell:
Thanks, good catch. Just `:files (:defaults "ledger-test.el")` is sufficient here, and I've applied that in 1549048b.

JalapenoGremlin:
Wow. Sorry about that, @jaccarmac. I didn't have a problem loading the package locally. Good catch. 
I'll review the CONTRIB more closely next time so that I don't break things in the future -- Perhaps I missed something that could have caught this before submitting the pull request.

jaccarmac:
Sweet, didn't realize that was a feature @purcell. Thanks for the trick!

No worries @JalapenoGremlin, I've dealt with weird packaging issues before when my environment wasn't perfectly clean. I'm just happy you left a paper trail to follow so I could hunt down the issue easily :smile:!

PR #4717, opened 2017-04-18
ssh-config-mode-el: Add *.txt to get ssh-config-keywords.txt.

jhgorrell:
Includes the `ssh-config-keywords.txt` file which is a new file.


purcell:
Thanks!

PR #4723, opened 2017-04-23
recipes/pamparam: Add pamparam

abo-abo:

Simple and fast flashcards

https://github.com/abo-abo/pamparam

maintainer

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/pamparam
+++ b/recipes/pamparam
@@ -1,2 +1,3 @@
 (pamparam :repo "abo-abo/pamparam"
-          :fetcher github)
+          :fetcher github
+          :files (:defaults (:exclude "elpa.el")))
--- a/recipes/pamparam
+++ b/recipes/pamparam
@@ -1,0 +1,2 @@
+(pamparam :repo "abo-abo/pamparam"
+          :fetcher github)


purcell:
Thanks, this looks great! Some feedback:

- The recipe will include `elpa.el`. I'd suggest you keep distributed and non-distributed elisp files in separate dirs in your repo.
- You `require` hydra, but there's no declared dependency on it.
- The `pam-` prefix should be `pamparam-` throughout. `package-lint` should have told you about this... is it broken?

abo-abo:
> The recipe will include elpa.el. I'd suggest you keep distributed and non-distributed elisp files in separate dirs in your repo.

I tried once to move it elsewhere. It was too tricky w.r.t. paths.

> You require hydra, but there's no declared dependency on it.

`lispy` depends on `hydra`.

> The pam- prefix should be pamparam- throughout. package-lint should have told you about this... is it broken?

`package-lint` didn't complain. But I'd really prefer to keep the `pam-` prefix. It's unique enough, and is more convenient for programming and discoverability.

purcell:
> I tried once to move it elsewhere. It was too tricky w.r.t. paths.

Okay, then you'll need something like `:files (:defaults (:exclude "elpa.el"))` in the recipe.

> lispy depends on hydra.

It might not always depend on `hydra` - it's good style to depend explicitly on all the packages you use directly, not rely on transitive dependencies.

> But I'd really prefer to keep the pam- prefix. It's unique enough, and is more convenient for programming and discoverability.

Please don't. `pam` already means something else in Linux land, and bytes are cheap. `pamparam` is still a short prefix by elisp standards.

abo-abo:
> pam already means something else in Linux land

It's nothing too relevant to user land. In the unlikely event of `pam.el` ever coming up, I can make a rename to `pamparam-`.

alphapapa:
> It's nothing too relevant to user land. In the unlikely event of pam.el ever coming up, I can make a rename to pamparam-.

But then users would have to fix their configs.  :(

tarsius:
@purcell Do you still insist on the prefix being changed?
@abo-abo Do you still insist on the prefix kept as is?

It's not up to me to make a decision, but if neither of you changes his mind, then I will close this issue. :wink:

abo-abo:
@tarsius Renaming the prefix isn't a blocker for me, although keeping the same one would be more convenient.  I've switched jobs a few months ago, and the current job strain prevents me from putting a lot of time into my personal projects. I already maintain many packages, I probably shouldn't add one more for now. If my job situation improves, I'll rename the prefix and request another merge. You can keep the PR open or close it, as you prefer.

Pamparam is a good package, I use it every day and wanted to share it. I just don't want to make any maintenance promises right now.

purcell:
Cool, I sent a PR in https://github.com/abo-abo/pamparam/pull/6.

PR #4724, opened 2017-04-24
Add mastodon recipe

jdenen:

Client for [Mastodon](https://github.com/tootsuite/mastodon), the federated microblogging service.

https://github.com/jdenen/mastodon.el

I'm the maintainer.

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/mastodon
+++ b/recipes/mastodon
@@ -1,0 +1,1 @@
+(mastodon :fetcher github :repo "jdenen/mastodon.el" :files ("lisp/*.el"))


purcell:
Thanks, looks good. Couple of small things I noticed:

- Soft-loading your own features, e.g. `(require 'mastodon-client nil t)`, doesn't make much sense.
- Use `package-file` rather than `package` in `Cask`, to avoid repeating your metadata: like MELPA, Cask will happily read the metadata from `mastodon.el` if you let it.
- Avoid version vars like `mastodon-version`: they are redundant and won't be correct in snapshot builds, so you'll just get confusing bug reports.

In any case, this is ready to merge, and you can address the above separately. :-)

jdenen:
Thanks for the feedback! This is my first Emacs package so I'm still working my way through the conventions.

PR #4726, opened 2017-04-29
It is very simple mode to mimic the concept of leader key in VIM.

wcy123:
http://github.com/wcy123/leader-key-mode

I am the maintainer.

    key - mode

### DIFF
--- a/recipes/leader-key-mode
+++ b/recipes/leader-key-mode
@@ -1,0 +1,1 @@
+(leader-key-mode :repo "wcy123/leader-key-mode" :fetcher github)


alphapapa:
You might want to look at https://github.com/noctuid/general.el

wcy123:
Thank you very much for your feedback. I think leader-key-mode.el is not as same as general.el. Most importantly, I try my best to keep it as simple as possible. In fact, almost the whole implementation is just no more than 8 lines elisp source code which is listed in the web site.

Even through general.el has many more features than leader-key-mode.el, I think leader-key-mode.el is still very useful as a simple demo code, to show how `emulation-mode-map-alists` works in emacs. This might inspire people coming out with more exciting ideas and implement it.

I submit this PR to share it with my friends, because it is easier for them to install it via `package-install`.

alphapapa:
Generally a file this small is not suitable for releasing as a package.  It's essentially some key binds, which any user can add to their own config.  Demonstrations are very helpful, but they belong more on a blog post than a package.

I don't say this to discourage you, but it's just not what is generally suitable for a package.

If you really want to submit it as a package, you at least need to follow the MELPA submission guidelines.

PR #4728, opened 2017-05-03
Add point-pos recipe

alezost:

Saves and restores point positions.  I always thought that it can't be
useful for anyone but me, but it looks like it can:

  https://github.com/alezost/point-pos.el/issues/1

So I wonder would it make sense to add it to MELPA?

https://github.com/alezost/point-pos.el

I'm the maintainer

None needed

### Checklist

### DIFF
--- a/recipes/point-pos
+++ b/recipes/point-pos
@@ -1,0 +1,1 @@
+(point-pos :repo "alezost/point-pos.el" :fetcher github)


purcell:
I know you've looked at bookmarks to some extent, but have you looked at all the packages? Really feels like this must exist already!

Re. the code itself, you'd want to add a bunch of autoload cookies, and I also feel like you're probably reinventing a lot of code that's already in [ring.el](https://www.gnu.org/software/emacs/manual/html_node/elisp/Rings.html) for you to use.

alezost:
> I know you've looked at bookmarks to some extent, but have you looked
> at all the packages?

No, actually I didn't search for similar packages at all :smile:
I don't insist on adding this package to MELPA, feel free to reject it!

> Really feels like this must exist already!

The alternatives surely can exist.  Perhaps @cmpitg knows: somehow he
found my package and decided to use it for some reason.

> Re. the code itself, you'd want to add a bunch of autoload cookies,

No, I think there should be only one autoload - for `point-pos-save`.
The other commands can't be used until a first position will be saved so
autoloading them seems useless to me.

> and I also feel like you're probably reinventing a lot of code that's
> already in ring.el for you to use.

Hm, I'm not sure it will suit the needs of `point-pos`, but I didn't
know about `ring.el`, thanks for pointing!

purcell:
> I think there should be only one autoload - for point-pos-save.

Ah, yup, and I didn't notice that autoload when I reviewed.

Well, I'll go ahead and merge this for now, and you can follow up with any further changes that make sense to you - thanks! :-)

kaushalmodi:
@alezost Seems like the [history][1] package is pretty close. Would you like to compare with that?

@cmpitg https://github.com/alezost/point-pos.el/issues/2

[1]: https://github.com/boyw165/history

alezost:
Thanks for merging!
@kaushalmodi I replied in the issue you opened, thanks!

cmpitg:
@alezost @kaushalmodi Thanks for keeping me in the loop :-). Since this issue is about getting `point-pos` into MELPA, I'll reply in alezost/point-pos.el#2.

PR #4729, opened 2017-05-04
Added unicode-picker recipe.

accidentalrebel:

Unicode-picker is a visual picker tool that makes searching and picking unicode characters easier.

It searches unicode characters using regular expressions and displays the results to a dedicated buffer. Selected characters from dedicated buffer are inserted back to the point from the calling buffer.

https://github.com/accidentalrebel/emacs-unicode-picker

Maintainer.

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/unicode-picker
+++ b/recipes/unicode-picker
@@ -1,0 +1,1 @@
+(unicode-picker :fetcher github :repo "accidentalrebel/emacs-unicode-picker")


alphapapa:
Hey, this looks very cool!  I like the way the UI works in the GIF.  It looks like a nice alternative to `helm-unicode` for some situations.

Here are a few suggestions:

1.  `(add-hook 'post-command-hook 'unicode-picker--post-command-listener)`  It's not a good idea to add your listener function to this global hook.  It's especially not good to do it and then never remove it.  :)  It would be best if you could figure out a different way to do this.
2.  You might consider using a keymap instead of `local-set-key` in a function.
3.  You should define a customization group with `defgroup`, and then you don't need `:group unicode-picker` in subsequent `defucstom`s.
4.  I'm not sure what this is for:
```el
 (if (fboundp 'devenv-smart-open-elisp-output-window)
	  (devenv-smart-open-elisp-output-window unicode-picker--buffer-name)
```
  But I'm guessing it's something particular to your personal config.  It would be better to leave this out of your package, and then add this customization in your personal config with, e.g. advice.

5.  You might want to use a persistent buffer so you don't need to activate `unicode-picker-mode` every time the user calls it.
6.  `(setq index (+ index 1))`  can be simply `(incf index)`.
7.  There are some dangling `)` here and there.  :)
8.  Instead of, e.g. `(select-window (get-buffer-window unicode-picker--caller-buffer)` you might be able to simply do `(with-current-buffer unicode-picker--caller-buffer ...`.  You should definitely be able to do that instead of `(switch-to-buffer unicode-picker--buffer-name)`.
9.  In general, you might consider if there are ways you can use fewer special variables.
10.  You might want to give some more functions `--` names.  I'm guessing that only `unicode-picker` is intended to be called by the user.

That list got longer than I expected.  I don't mean to be nit-picky--this looks like a really useful package!   Thanks for making it.  :)

accidentalrebel:
Hi @alphapapa,

Thank you for your suggestions. I swear, everytime I send a package to melpa I always learn something new.

I have some clarifications, if you do not mind:

**There are some dangling ) here and there. :)**
Which ones exactly? When you refer to dangling ")" is it the one for unicode-picker-insert-character? How about the one in unicode-picker--post-command-listener? Is that considered dangling? I am honestly not yet sure of this convention.

**In general, you might consider if there are ways you can use fewer special variables.**
Are you specifically talking about the different defvars I've defined? Does this mean I should consider putting them inside "lets" as much as possible?

**You might want to give some more functions -- names. I'm guessing that only unicode-picker is intended to be called by the user.**
Sorry, I don't quite follow. I have two other functions that can be called by the user if they wish, specifically unicode-picker-insert-character and unicode-picker-insert-character-then-return. Is this what you meant with "functions -- names"?

I understand your other points and I think they are something that I should implement. Will do these changes over the following days.

Thank you very much!

alphapapa:
> Thank you for your suggestions. I swear, everytime I send a package to melpa I always learn something new.

Ditto!  Steve and syohex always teach me something new.

By dangling `)`, I mean ones that are on a line by themselves, like at https://github.com/accidentalrebel/emacs-unicode-picker/blob/master/unicode-picker.el#L93.  Sometimes this is useful, like when defining a long list with each token on its own line, to make diffs less noisy, but generally the convention is to not do this.

> In general, you might consider if there are ways you can use fewer special variables.

What I mean is, you store global state in several special variables.  This isn't necessarily a problem--Emacs does plenty of this--but in general it makes it easier to reason about and debug code if you avoid this as much as possible.  I would guess that if you refactor to avoid that listener function, you'll need less of this.

> Sorry, I don't quite follow. I have two other functions that can be called by the user if they wish, specifically unicode-picker-insert-character and unicode-picker-insert-character-then-return. Is this what you meant with "functions -- names"?

Well, I was looking at these:

    (local-set-key (kbd "RET") 'unicode-picker-insert-character-then-return)
    (local-set-key (kbd "SPC") 'unicode-picker-insert-character)

Those are called by the user, but IIUC, they are only usable within the unicode-picker window/buffer, so they're not functions or commands that a user would call independently.  They're not *exactly* private functions, but they sort of are, because they're only useful within their own mode.  This is just a matter of style, really, so it's not a criticism.  :)

accidentalrebel:
Thanks for the clarifications. I understand the reasoning behind them. Will make sure to implement them as soon as I am able this weekend.

Thanks a lot!

tarsius:
Friendly ping!

tarsius:
Pong?

accidentalrebel:
Hi @tarsius apologies for not getting back sooner. I have not come back to this project in a while and might not be able to in the foreseeable future. Should I take this off here for now and just resend if ever I do get back to it?

Thanks!

tarsius:
> Should I take this off here for now and just resend if ever I do get back to it?

Yes, let's close this for now.

Cheers!

PR #4730, opened 2017-05-05
Add soji package

joecannatti:

Soji is a Mindful Workday tool for emacs.
it's essentially a combination of the Pomodoro Technique, jounaling, and note taking.

[Soji on Github](https://github.com/joecannatti/soji-emacs)

Maintainer

None Needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/soji
+++ b/recipes/soji
@@ -1,0 +1,1 @@
+(soji :fetcher github :repo "joecannatti/soji-emacs")


purcell:
Thanks. Quick feedback:
- You've got package metadata in 3 places, and none of your package dependencies will currently be seen by MELPA. You can easily fix this: delete the `-pkg.el` file, use cask's `package-file` directive instead of `package` and `depends-on`, and check `soji.el` with `package-lint` -- for example, the first line should start `;;;`, not `;;`, and you'd add your dependencies on `org-pomodoro` in there.
- Don't use `cl`, use `cl-lib` instead.
- Don't call your `defgroup` `soji-settings` -- it's more idiomatic to just call it `soji`.
- All the `defcustom` default values look like your personal settings. Try changing them to defaults that will work for new users.
- `soji-dim` and `soji-bright` do not belong in this package. Instead, provide hooks so that users can add their own special effects.
- Code like `(add-hook 'org-pomodoro-started-hook 'soji-bright)` modifies the behaviour of `org-pomodoro`, which is undesirable. 
- Don't modify `org-agenda-custom-commands` for the user. Perhaps provide a set of custom commands that the user can _optionally_ include in their own agenda settings. Really, it looks like all of your agenda commands and capture templates should simply be deleted from the file.
- You should provide more documentation (e.g. in the `Commentary` section) if you want this to be useful to other people.

A lot of this appears to be a very personal system, not really a reusable package that lots of other people will easily incorporate into their own workflows, so I feel like it's not yet ready for wider distribution.

tarsius:
Friendly ping!

joecannatti:
@tarsius Thanks! This can be closed for now. The comments from @purcell were very helpful and got me re-thinking what I had in mind with this. Thanks! Closing

PR #4731, opened 2017-05-05
Add turing-machine recipe

therockmandolinist:

A single-tape Turing machine simulator (with an editing major mode), based on http://morphett.info/turing/turing.html.

https://github.com/therockmandolinist/turing-machine

Maintainer

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/turing-machine
+++ b/recipes/turing-machine
@@ -1,0 +1,1 @@
+(turing-machine :fetcher github :repo "therockmandolinist/turing-machine")


therockmandolinist:
Right now I'm updating a buffer and forcing redisplay -- I'd love to know if there's a better way to display rapidly changing propertized text.

purcell:
> Right now I'm updating a buffer and forcing redisplay -- I'd love to know if there's a better way to display rapidly changing propertized text.

Not sure, sorry. Maybe just `font-lock-ensure`? BTW, you should use syntax tables to set up the `;` comments -- there should be no need to add font-lock patterns for that, since syntax tables handle it more directly.

PR #4737, opened 2017-05-07
add helm-frame recipe

chee:

it opens your helm buffer in a dedicated frame in the middle of the screen, similar to rofi or Alfred.

https://gitlab.com/chee/helm-frame

I am the maintainer and contributor and an enthusiastic user.

**None needed**

### Checklist

Please confirm with `x`:

_**note,** i used a non-standard separator `/` for the commands; there is a potential for conflict with `helm-frame-$command` between `helm-frame` and a possible `helm` command. ie, if i had a command called `dog` and helm added a command called `frame-dog`, both would be called `helm-frame-dog`_

### DIFF
--- a/recipes/helm-frame
+++ b/recipes/helm-frame
@@ -1,0 +1,1 @@
+(helm-frame :fetcher gitlab :repo "chee/helm-frame")


chee:
screenshot of helm-frame working in sandbox
screenshot of `helm-frame` working in sandbox with only `helm` and `helm-frame` installed

purcell:
Thanks. Quick feedback:

- `(if frame frame nil)` is exactly the same as just `frame`.
- There's no need to `autoload` your commands.

Also:

> note, i used a non-standard separator / for the commands; there is a potential for conflict with helm-frame-$command between helm-frame and a possible helm command. ie, if i had a command called dog and helm added a command called frame-dog, both would be called helm-frame-dog

It doesn't matter, you should still use `-` rather than `/`. That's just how the simplistic namespacing in emacs lisp works.

Code like the following will always iterate over every item in a list, even if the first item is the one you want:

```el
  (let (current-monitor)
    (dolist (monitor (display-monitor-attributes-list) current-monitor)
      (if
        (member (or frame (window-frame)) (assoc 'frames monitor))
        (setq current-monitor monitor))))
```

So instead, `(require 'cl-lib)` and use something like `cl-find-if`, or `cl-loop`:

```el
  (cl-loop for monitor in (display-monitor-attributes-list)
    if (member (or frame (window-frame)) (assoc 'frames monitor))  
    return monitor)
```
or
```el
  (cl-find-if (lambda (m) (member (or frame (window-frame)) (assoc 'frames m)) )
              (display-monitor-attributes-list))
```

chee:
thank you for the feedback 

i've made those changes now

lol @ `(if frame frame nil)`

purcell:
Sorry for the long delay in merging - this got lost in the pile of pull requests. :-)

PR #4739, opened 2017-05-07
Uncaught TypeError: Cannot read property 'totalPackages' of undefined

### LINKED ISSUE #4738: Something wrong with melpa.org
Hi,

I just went to melpa.org and all it shows is this:

image

szkl:
Fixes #4738.

I started to have this error and a broken homepage recently. I've investigated the error message.

Error is thrown in the callback passed to `_.reduce` call [here](https://github.com/melpa/melpa/blob/a372bd61b8e09086124f4b5ae3a837d6220def38/html/js/melpa.js#L141).

I've figured the actual error is `actually some (~23) packages missing `props` property in the archive data by debugging. Error is swallowed for some reason.

Underlying error message can be reproduced by loading a package page directly.

```
Uncaught TypeError: Cannot read property 'commit' of null
```

Reference to `props` property introduced [here](https://github.com/melpa/melpa/pull/4736/files#diff-16a4b1562f252501410d42a063023b9fR153) breaks the execution.

### DIFF
--- a/html/js/melpa.js
+++ b/html/js/melpa.js
@@ -151,7 +151,7 @@
       pkgs.push(new melpa.Package({
         name: name,
         version: version,
-        commit: built.props.commit,
+        commit: props.commit || '',
         dependencies: deps,
         description: built.desc.replace(/\s*\[((?:source: )?\w+)\]$/, ""),
         source: recipe.fetcher,
--- a/html/js/melpa.js
+++ b/html/js/melpa.js
@@ -140,6 +140,7 @@
     var listed = _.intersection(_.keys(archive), _.keys(recipes));
     return new melpa.PackageList(_(listed).reduce(function(pkgs, name) {
       var built = archive[name];
+      var props = built.props || {};
       var recipe = recipes[name];
       var version = built.ver.join(".");
       var deps = _.map(built.deps || [], function (ver, name) {


html/js/melpa.js
```diff
@@ -150,7 +151,7 @@
       pkgs.push(new melpa.Package({
         name: name,
         version: version,
-        commit: built.props.commit,
+        commit: props.commit || '',
```
purcell: Was this change actually necessary? It looks undesirable to me, since it will make every package look like it has commit info.

milkypostman:
Also the pages are now showing "github03994829303828" as if appending the
hash to the link text.

We should probably discuss this in the original pull request.

html/js/melpa.js
szkl: I think it needs to be done in here or other places referencing `commit` property. Because `null` would be `"null"` in string concatenation. I didn't follow-up [this](https://github.com/melpa/melpa/pull/4736/files#diff-16a4b1562f252501410d42a063023b9fR410) reference any further. After reading the comment [below](https://github.com/melpa/melpa/pull/4739#issuecomment-299770740) I think it is either not needed or needs to be revised to build a proper URI for commit history page. @purcell what do you think?

szkl: > Or, rather, the default value of '' looks undesirable. :-)

@purcell got it :-) I think it's a matter of preference since JavaScript is not a typed language. `commit` property is assumed to be a string then my thinking is that the default value should be a string too. If we keep that aside then `''`, `null`, `undefined` all of them are equal as a boolean value and `props.commit` expression will yield `undefined`.

purcell: Yeah, the point is that code elsewhere should be able to say something like `if (package.props.commit) { ... }` in order to only render commit info in the page if it is present.

szkl: @purcell That's okay. Empty string is [false](https://www.ecma-international.org/ecma-262/7.0/index.html#sec-toboolean).

purcell:
> That's okay. Empty string is false.

That may be true, but it doesn't make it _right_. ;-)

PR #4742, opened 2017-05-08
New package mandoku-tls

cwittern:

This Emacs package provides access to the Thesaurus Linguae Sericae
(TLS) initiated and directed by Christoph Harbsmeier. It will use the
lexicon database from https://github.com/tls-kr/tls-org.

It can be used from Sinomacs (https://github.com/mandoku/sinomacs) or
installed as an Emacs package from Melpa.

https://github.com/mandoku/mandoku-tls

maintainer

Not needed.

### Checklist

Please confirm with `x`:

- [ x ] I've read [CONTRIBUTING.md](https://github.com/melpa/melpa/blob/master/CONTRIBUTING.md)
- [ x ] I've used [package-lint](https://github.com/purcell/package-lint) to check for packaging issues, and addressed its feedback
- [ x ] My elisp byte-compiles cleanly
- [ x ] `M-x checkdoc` is happy with my docstrings
- [ x ] I've built and installed the package using the instructions in the [README](https://github.com/melpa/melpa/blob/master/README.md)

### DIFF
--- a/recipes/helm-charinfo
+++ b/recipes/helm-charinfo
@@ -1,0 +1,1 @@
+(smex :repo "mandoku/helm-charinfo" :fetcher github)
--- a/recipes/helm-charinfo
+++ b/recipes/helm-charinfo
@@ -1,1 +1,1 @@
-(smex :repo "mandoku/helm-charinfo" :fetcher github)
+(helm-charinfo :repo "mandoku/helm-charinfo" :fetcher github)
--- a/recipes/mandoku-tls
+++ b/recipes/mandoku-tls
@@ -1,0 +1,1 @@
+(mandoku :fetcher github :repo "mandoku/mandoku-tls" )
--- a/recipes/mandoku-tls
+++ b/recipes/mandoku-tls
@@ -1,1 +1,1 @@
-(mandoku :fetcher github :repo "mandoku/mandoku-tls" )
+(mandoku-tls :fetcher github :repo "mandoku/mandoku-tls" )


purcell:
This package will bundle `helm-charinfo.el`: you should delete that since it seems unnecessary. The `-pkg.el.in` file is also unnecessary and can be deleted.

- `defvar mandoku-tls-root-path` should be a `defconst` so that it gets redefined when you replace and reload a package.
- You should add a `defgroup` to place your custom items, and specify its containing group.
- You're defining a `helm` source and using `defhydra` in here. For the latter you *must* depend on `hydra`, and for the former you should also add a package dependency on `helm`. (If the `helm` functionality is optional, then the correctly-packaged solution would be to provide a separate `helm-mandoku-tls` package, so that it could have proper package dependencies and would byte-compile cleanly.)
- You're also modifying org settings (`org-cycle-hook`, `org-add-link-type`), and using various `org` functions, so you should `(require 'org)`
- `M-s` and `M-t` are built-in bindings: don't override them globally. Create a prefix map for everything after the `M-t` and _suggest_ top-level bindings for the user in your docstrings.

Hope that helps!

krptest:
Dear Steve,
Thanks for taking the time to review my package.  This was extremely helpful.  I have now updated the package following your suggestions with the exception of one point:  mandoku-tls requires (and declares a dependency on)  mandoku, which in turn requires and declares a dependency on org.  This led me to believe that adding that again to mandoku-tls  would be unnecessary. Is this a mistaken notion?  Is the proper way to do this to also declare indirect dependencies?  Or is this necessary because I directly interact with some of org code?
All the best,
Christian

purcell:
It's always best to declare explicit dependencies for the code you're using: the transitive dependencies could change at some future point, and then your code might stop working.

cwittern:
I see.  Will do so.  Thanks.  Christian

purcell:
> I have now updated the package following your suggestions with the exception of one point

I still see `helm-charinfo.el` in the repo, though. If it's really to be part of this package, you should call that `mandoku-helm-charinfo.el`.

The new dependency on `github-clone` seems wrong, too. And where possible, you should avoid depending on the MELPA snapshot versions of other packages: instead, depend on the latest stable version (e.g. on MELPA Stable) so that if you tag your repo with a version, users of MELPA Stable will be able to install it using only packages in MELPA Stable.

cwittern:
Dear Steve,
I now moved helm-charinfo into a separate package, for which I will create a PR in a moment. I also tried to use references to melpa stable version where available.  I think I addressed all your concerns now.
All the best,
Christian

cwittern:
Dear Steve,
I tried to make a new PR for the new package, but it seems it got added to this one for some reason. If this is wrong, please tell me how to fix it.
All the best,  Christian

purcell:
Great, thanks - merged!

cwittern:
Thank you, great to know this is on melpa now!  All the best,  Christian

PR #4744, opened 2017-05-08
Add atom-tabs.el recipe

domtronn:
A reimplementation of Tabs in the style of the Atom Editor using [`all-the-icons`](https://github.com/domtronn/all-the-icons.el)

https://github.com/domtornn/atom-tabs.el

I'm the creator/maintainer

### DIFF
--- a/recipes/atom-tabs
+++ b/recipes/atom-tabs
@@ -1,0 +1,3 @@
+(atom-tabs
+ :fetcher github
+ :repo "domtronn/atom-tabs.el")


Silex:
Working link to repo is https://github.com/domtronn/atom-tabs.el

purcell:
Thanks. A few things I spotted:
- You've got `defcustom` vars like `atom-tabs-show--tab-numbers?`, but if a var is public enough to be customizable, you shouldn't be using `--` in the name. And `?` may be typical in Ruby etc., but `-p` (predicate) is conventional in Lisp. (Same with args like `active?`.)
- Add a `defgroup` for your custom group (with an appropriate parent `:group` set) otherwise the custom items will be stranded.
- The docstring for your keymap says `"Keymap for Projectile mode."`... :-/
- If you're depending on functions from libraries, `require` the corresponding library: examples are `cl-reduce` and `cl-incf`, from `cl-lib`.
- Don't hard-code faces `:family "Arial Narrow" :height 2.5`, particularly font families. Use `defface`, declare a simple face that will work everywhere, and then people can customize it in themes if they wish.
- Code like `(face-background 'powerline-inactive1)` assumes `powerline` is available, but that doesn't appear to be a requirement.
- Where you have `(advice-add 'find-file :after 'atom-tabs--add-recent)` etc., there are generally standard hooks you should use instead of advice, e.g. `find-file-hooks`.
- What happens if you have multiple frames, and you cycle tabs in one, but not the other? Does everything still work? Can you have tabs in one frame but not another?
- Don't hard-code colours like `#63B2FF`, provide faces instead.
- Prefer mode names to regexps for blacklists and whitelists. Much less brittle, and much easier to specify.

Hope that helps!

Alexander-Miller:
What this package needs is some screenshots in the readme. People are much more likely to use it when they look at your readme and like what they see. Especially since trying it for yourself involves first installing the fonts from all-the-icons.

tarsius:
Friendly ping!

domtronn:
Apologies that this is taking a while! I started a new job so haven't had much time to work on this. I'm also still looking at ways to optimise the performance as at the moment I'm not happy with it... 

I am still intending on completing it though, but if you'd like to close this PR for now, I can re open it when I'm finally happy with it? :)

tarsius:
https://github.com/domtornn/atom-tabs.el no longer exists so I am closing this.

PR #4747, opened 2017-05-09
Update omnisharp: remove override on :files, not needed anymore

razzmatazz:
This is a small update on the omnisharp recipe where it has been refactored to have a little bit more canonical file structure with all the files on this multi-file package having now moved to the root directory of the repo.

See https://github.com/OmniSharp/omnisharp-emacs


purcell:
I like this sort of change - thanks!

PR #4748, opened 2017-05-09
added caleb theme

avelino:

The disco music theme for Emacs

https://github.com/avelino/caleb-theme

Maintainer

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/caleb-theme
+++ b/recipes/caleb-theme
@@ -1,0 +1,1 @@
+(caleb-theme :fetcher github :repo "avelino/caleb-theme")


purcell:
Thanks for this, but all those `display-graphic-p` tests are incorrect and undesirable. If you have multiple variants of the faces for different displays, the way to handle it is to put all the variants in the face definitions, so that Emacs will choose the right ones according to the frame on which the face is later displayed.

So after loading the theme in an emacs instance, the correct colours will be displayed both in graphical and non-graphical frames connected to that instance. You'll be able to see that's not the case here by launching a graphical Emacs frame, loading the theme, and then connecting to the same Emacs in a terminal using `emacsclient -t`.

The emacs manual has more on this: https://www.gnu.org/software/emacs/manual/html_node/elisp/Defining-Faces.html#Defining-Faces

I gave this same feedback to another theme author recently (#4671) and he did a great job of fixing up his theme to work more smoothly in multiple frame types: hopefully it'll be easy for you to do the same. :-)

tarsius:
Friendly ping!

avelino:
Theme deprecated, sorry

PR #4749, opened 2017-05-13
Add shx recipe

riscy:

shx or "shell-extras" extends comint-mode.  It parses simple markup in the output stream (enabling plots and graphics to be embedded in the shell) and adds several command-line functions which plug into Emacs (for example, use :e <filename> to edit a file).

https://github.com/riscy/shx-for-emacs/

Maintainer.

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/shx
+++ b/recipes/shx
@@ -1,3 +1,2 @@
 (shx :repo "riscy/shx-for-emacs"
-     :fetcher github
-     :files ("shx.el"))
+     :fetcher github)
--- a/recipes/shx
+++ b/recipes/shx
@@ -1,0 +1,3 @@
+(shx :repo "riscy/shx-for-emacs"
+     :fetcher github
+     :files ("shx.el"))


purcell:
Thanks. This looks cool!

Quick feedback:
- Why does the recipe not include `shx-split.el`? Maybe put that file in a separate dir (e.g. `tool`, `misc`) if it's not required, because recipes that list specific files are more brittle.
- `(string= (shx--current-input) "")` => you're requiring Emacs 24.4, so `string-empty-p` is an option.
- `(add-hook 'shell-mode-hook 'shx-for-shell-mode) ; always run in shell-mode` is a bit of a hack: you should generally leave that choice to the user, or provide a globalized minor mode that activates `shx-for-shell-mode` in buffers having one of a user-customizable set of major modes.
- `shx-activate` and `shx-for-shell-mode` should be defined with `define-minor-mode` instead, so they can be toggled on/off cleanly, and have their own keymap instead of hacking the local map.

riscy:
Thanks so much for all you do.  This is great feedback -- I'll fix it up.

riscy:
- Moved `shx-split.el` and `shx-test.el` into a separate repository and subdirectory respectively, so the recipe no longer needs to specify the file.
- Used `string-empty-p` (and `string-trim-left`, `string-prefix-p`, ...) where appropriate.  Thanks for the pointer to that.
- Hackish `shx-activate` and `shx-for-shell-mode` are gone. Now `shx-mode` is a full-fledged minor mode that can be turned on and off -- much better.  Moreover, all the shx functions that advise comint-mode functions only work when `shx-mode` is on, so the advice effectively turns on and off with the minor mode.

These changes are all in the 0.0.1 release.

purcell:
Nice, thanks - merged!

PR #4753, opened 2017-05-14
Add flycheck-popup-tip recipe

Simplify:

Display Flycheck error messages using popup.el 

https://github.com/flycheck/flycheck-popup-tip

Maintainer

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/flycheck-popup-tip
+++ b/recipes/flycheck-popup-tip
@@ -1,0 +1,2 @@
+(flycheck-popup-tip :fetcher github
+                    :repo "flycheck/flycheck-popup-tip")


purcell:
Thanks. You've got just a single minor mode, which intercepts `post-command-hook`. This means that every command in every buffer - whether or not that buffer uses flycheck - is going to have all its commands inspected. This is very inefficient, and I'd be surprised if it's the only way to make this work. Instead, you should be able to make a non-global minor mode which keeps state in buffer-local variables, and modifies `post-command-hook` buffer-locally rather than globally. (You can still easily make a global minor mode to enable it globally, by adding/removing your minor mode to `flycheck-mode-hook` when the global minor mode is enabled/disabled.)

Simplify:
Hi @purcell, sorry for delay. I just changed code to implement non-global minor mode, makes more sense. Original minor-mode code came from `flycheck-pos-tip` that also implemented global minor-mode.

purcell:
Okay, great, but also note my comment about only modifying hooks locally: you should pass the `LOCAL` flag when calling `add-hook` and `remove-hook`.

tarsius:
Friendly ping!

Simplify:
@purcell , @tarsius , sorry about delay. Code is changed now to pass `local` flag when adding or removing hooks.

PR #4759, opened 2017-05-17
Add counsel-spotify to melpa

Lautaro-Garcia:

Allows the user to search for tracks, albums and artists in the Spotify API and play those with
the Spotify App. The frontend interface to narrow the possible results is Ivy.
It is simply a fork of `helm-spotify` but with Ivy as the interface and the possibility to search by
artist and album

https://github.com/Lautaro-Garcia/counsel-spotify

I am the mantainer

None needed

### Checklist

### DIFF
--- a/recipes/counsel-spotify
+++ b/recipes/counsel-spotify
@@ -1,0 +1,1 @@
+(counsel-spotify :fetcher github :repo "Lautaro-Garcia/counsel-spotify.el")


purcell:
Thanks. Quick feedback:

- You've declared `:group 'counsel-spotify`, but there's no corresponding `defgroup`, so you should add one, with an appropriate parent group, e.g. `applications`.
- When you insert a string into a shell command using `%s`, you should almost always be using `shell-quote-argument`, otherwise nasty things can happen if the inserted string contains certain characters.

Looks good otherwise. Hope that helps!

Lautaro-Garcia:
Thank you for your review! 
I added a `defgroup` and used `shell-quote-argument` for the shell commands

purcell:
Great, thanks - merged!

PR #4760, opened 2017-05-17
Add treemacs recipe.

Alexander-Miller:

It's a graphical-ish file explorer like neotree. The picture at the top of the readme explains it better than I can.

I made this.

None needed

### Checklist

Except for that one warning in treemacs-mode.el. I was actually hoping you might know a way to get around it while keeping the major mode definition in its own file.
Kinda. It's mostly just stuck on Documentation style. However the docs are in proper order as per the flycheck syntax cheker, meaning all functions and their arguments are documented, punctuation and line lengths are as the should be etc.

### DIFF
--- a/recipes/treemacs
+++ b/recipes/treemacs
@@ -1,0 +1,4 @@
+(treemacs
+ :fetcher github
+ :repo "Alexander-Miller/treemacs"
+ :files (:defaults "icons" (:exclude "treemacs-evil.el")))
--- a/recipes/treemacs-evil
+++ b/recipes/treemacs-evil
@@ -1,0 +1,4 @@
+(treemacs-evil
+ :fetcher github
+ :repo "Alexander-Miller/treemacs"
+ :files ("treemacs-evil.el"))


purcell:
Thanks! Some quick feedback after poking through a few of the files:

- `(:exclude "*tests.el")` is unnecessary, since those files are excluded by `:defaults`
- Don't `(require 'cl-macs)` - require `cl-lib` instead. In the case where `cl-lib` has been installed via ELPA, e.g. in older Emacs versions, that package does not include a file called `cl-macs`.
- Why not rename `treemacs-init` to simply `treemacs`? Much nicer.
- If you list the arg and directory name in the signature of `treemacs-init`, you can initialise them both in the `interactive` form, and also make the function more useful for people calling it from elisp.
- Where you use `f` and `s` functions in a file, make a habit of `require`-ing those libs at the top of the file. (e.g. in `treemacs.el`)
- `treemacs-build-tree` has code which is conditional upon the result of `(if (window-system)`, but it seems to me that the treemacs buffer would be visible in any type of frame in the current Emacs, so the same buffer should be made to work in either GUI or terminal frames simultaneously. But perhaps I'm missing something.
- `with-eval-after-load` only appeared in Emacs 24.4, but you don't declare a dependency on that Emacs version. The `eval-after-load` mechanism is really only meant for users' startup files. Calling it inside functions is quite off-the-wall: instead of the code in `treemacs--evil-config`, you should consider making a separate little `treemacs-evil` package which can have proper dependencies on `evil` and `treemacs`, and avoid all the "is it there, is it not there?" magic. That's what the package mechanism is for, and we'll be more than happy to host separate little packages like that.
- `setq-local` was only added in Emacs 24.3. You probably need to figure out your minimum Emacs version, and declare it. It's pretty easy to get a travis build running to test it, too.

Hope that helps!

Alexander-Miller:
Lots of changes coming up here. If you don't mind I'd prefer to push these to [a dev branch](https://github.com/Alexander-Miller/treemacs/tree/dev) while we're doing the review. When the discussion and code have stabilized I'll get everything on master.

> (:exclude "*tests.el") is unnecessary, since those files are excluded by :defaults

Funny, I could've sworn I've only seen the singular version in the readme. It's fixed now.

> Don't (require 'cl-macs)

Done.

> Why not rename treemacs-init to simply treemacs?

Done.

> If you list the arg and directory name in the signature of treemacs-init

I'm afraid I'll need some pointers here. How do I make sure you're only asked for a dir when there's a prefix arg? `(defun treemacs (&optional arg dir) (interactive "PD\nDir: ") ...` doesn't do it.

> treemacs-build-tree has code which is conditional upon the result of (if (window-system), but it seems to me that the treemacs buffer would be visible in any type of frame in the current Emacs,

Pretty much, yes. I wrote that fallback under the assumption that people would stick to a single emacs session either in a gui or in a terminal, or that at least they'd focus on working in either one for a time. This way compatibility with the current ui is only a single refresh away.

Treemacs is very much built on the idea that there's only 0 or 1 treemacs buffers around, and I don't think adding another special term-treemacs buffer is worth the additional complexity. I can expand the switch detection to also run in `treemacs-toggle`, but fixing this by adding more buffers runs the risk of treemacs getting weird with something like eyebrowse and persp and neotree being buggy with eyebrowse is pretty much why I started doing treemacs in the first place.

> Where you use f and s functions in a file, make a habit of require-ing those libs at the top of the file

Done.

> setq-local was only added in Emacs 24.3. You probably need to figure out your minimum Emacs version,

Done. This ought to be enough for now -  I've zero experience with travis, but I'll make learning about it a long term goal.

`> instead of the code in treemacs--evil-config, you should consider making a separate little treemacs-evil package

The whole evil/default config bit is one of the oldest parts of treemacs, written when I knew much less about elisp than I do now. The issue I had back then was that treemacs needed to know whether to be evil or not at the time it was being loaded, doing an evil initilization after the fact would always result in the keymaps not being set up right and the cursor being visible. I'll try and see if I can get around that now.

However I don't know that I see the value in extracting this into another package. If I can get this to work without the whole know-it-at-load-time business I can just put the evil bits into another file that needs to be explicitly required, same as with another package, but without the extra work of setting up and publishing a new repository and rewriting my readme.

At any rate first I need to figure out how to make evil work on its own first.

Alexander-Miller:
Ok, other than the interactive part I don't quite know how to do I think I'm now through with everything on that list. I've added more checks for gui/tui swicthes and got evil down to a single optional require.

Note in case you'll want to keept looking at the code, since it might be another point of contetion: please don't mind the silly method of starting two processes I'm currently using to check for git status. I've no clue why I ever thought that was a good idea and that code is very much obsolete. There's a feature branch ready to make things simple and async once my other process-future MR gets merged.

purcell:
Nice, looking good!

> However I don't know that I see the value in extracting this into another package. If I can get this to work without the whole know-it-at-load-time business I can just put the evil bits into another file that needs to be explicitly required, same as with another package, but without the extra work of setting up and publishing a new repository and rewriting my readme.

The `treemacs-evil.el` file will cause byte-compilation issues when installing in an Emacs without `evil`, which is why I'd suggest separating it out. 

You can have two separate `treemacs` and `treemacs-evil` packages living in the same repo easily enough: just make the recipe for the former `:exclude` the latter, and add a separate recipe for that other single file. This way, you can put proper dependency headers in `treemacs-evil.el` and everything will work beautifully.

Alexander-Miller:
> byte-compilation issues when installing in an Emacs without evil

I hadn't considered that. In this case it makes sense to go for another package.

Evil is excluded now.

Alexander-Miller:
I know the rules say 1 PR per package, but would you mind if I add another commit for the treemacs-evil recipe?

The two directly belong together, and it would prevent the issue of the treemacs-evil PR being either broken or delayed until treemacs proper gets merged.

purcell:
Yep, sure.

Alexander-Miller:
Never mind the PR closing - I just deleted the branch because I wanted to restart the CI build which was failing because it was cloning master, but all my changes are in dev.

I've added a new commit for treemacs-evil now. It does not have a treemacs version it needs, since it always works with master. It doesn't have its own version either, since it's an optional part of treemacs and does shares its version and the fewer version strings I need to remember to pull up, the better. I hope this setup is alright, as flycheck-package classifies the latter as a fullblown error.

purcell:
> I've added a new commit for treemacs-evil now. It does not have a treemacs version it needs, since it always works with master. It doesn't have its own version either, since it's an optional part of treemacs and does shares its version and the fewer version strings I need to remember to pull up, the better. I hope this setup is alright, as flycheck-package classifies the latter as a fullblown error.

Yep, I sent a fix for those in https://github.com/Alexander-Miller/treemacs/pull/14. You can use `Package-Version: 0` to keep flycheck-package happy, and `"."` was just plain invalid as a dependency version, while `"0"` does the trick.

Alexander-Miller:
Hey there. It's been a while now - is there anything else you think needs to be cleared up before merge is a go? I think I've fixed all the stuff you found.

Bunch of new stuff in master in the meantime - smoother git thanks to pfuture and autorefresh on changes in the displayed file system - if you want to try that out, too.

I'll also be changing things to a (optional?) treemacs-buffer-per-frame approach, so term compatibility should fully clear up as well. The spacemacs maintainers showed intereset in getting treemacs on board and framelocality is a thing that neotree struggles with a lot. It'll be a *lot* of complex changes though, and in the meantime I think the current system is sufficiently robust for MELPA.

purcell:
Great - merged, thanks.

Alexander-Miller:
👍

PR #4762, opened 2017-05-18
Add recipe for vdiff-magit

justbur:
I am the author of this package. I recently broke this file, which integrates my
vdiff package into magit, out into a separate repository. vdiff-magit.el used to be on melpa because it was included in the vdiff package before this change. 

See https://github.com/justbur/emacs-vdiff and
https://github.com/justbur/emacs-vdiff-magit

Thanks

### DIFF
--- a/recipes/vdiff-magit
+++ b/recipes/vdiff-magit
@@ -1,0 +1,1 @@
+(vdiff-magit :repo "justbur/emacs-vdiff-magit" :fetcher github)


purcell:
Thanks. If you can, please depend on a stable version of magit, otherwise if you tag your package to make a stable release, it will not be installable by users who have only MELPA Stable enabled.

justbur:
Done, thanks

PR #4763, opened 2017-05-18
Update fstar-mode: add etc/ directory

cpitclaudel:
The original recipe included only text files, but I need an icon for
notification popups.


purcell:
Thanks!

PR #4765, opened 2017-05-21
Add prettier-js

wyuenho:

Formats your JavaScript code using 'prettier' on file save.

https://github.com/prettier/prettier/blob/master/editors/emacs/prettier-js.el

An enthusiastic user

~~Package attributes are currently missing, waiting for prettier/prettier#1651~~

Merged

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/prettier-js
+++ b/recipes/prettier-js
@@ -1,0 +1,1 @@
+(prettier-js :fetcher github :repo "prettier/prettier" :files ("editors/emacs/prettier-js.el"))


purcell:
Thanks. Quick feedback:
- "Emacs minor mode to format JS code on file save": please remove the redundant "Emacs" in this. `package-lint` should have pointed this out.
- There is no `defgroup` form to declare the containing custom group. (Its parent `:group` should be `languages`.)
- Custom vars are public, and should not have the prefix `prettier-js--`: instead, they should begin `prettier-js-`. (`--` indicates package-private vars.)
- The functions should also use a consistent namespace, so `prettier` and `prettier-mode` should be renamed to `prettier-js` and `prettier-js-mode`.
- `prettier-before-save` is unneccessary, and the docstring is confusing and misleading: this function can be deleted, then just add/remove `prettier` to the hooks locally when the mode is enabled/disabled.
- Almost all the code in `prettier-js--delete-whole-line` appears unnecessary: just call `kill-whole-line` will `kill-ring` temporarily bound to `nil`. I've seen this code cargo-culted into a couple of other packages recently.

So this means a few upstream changes for you to coordinate first, please. Hope that helps.

rcoedo:
Thanks for the feedback @purcell.

I created a pull request fixing these issues, if I can borrow 5 minutes from you I'd love your feedback on these.

Cheers! 🍻

purcell:
Replaced by #4777.

PR #4767, opened 2017-05-22
Add qt-pro-mode

EricCrosson:

Provides a major-mode for editing Qt build-system files.

https://github.com/ericcrosson/qt-pro-mode

Maintainer, contributor, user

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/qt-pro-mode
+++ b/recipes/qt-pro-mode
@@ -1,0 +1,1 @@
+(qt-pro-mode :fetcher github :repo "EricCrosson/qt-pro-mode")


EricCrosson:
It seems like the ci build failed on the fetch of upstream `cask` resources

purcell:
Thanks. I re-ran the build and it looks fine.

As for the specifics, the code looks good, just one little oddity: `(eval-after-load 'qt-pro-mode #'qt-pro-mode-associate-files)` is nonsensical, since that form will only execute when `qt-pro-mode` is loaded. In other words, just promote the code inside `qt-pro-mode-associate-files` to the top level and you'll get the same effect.

Bonus tip: `define-derived-mode` will automatically set the syntax table correctly based on its name, so you can omit `:syntax-table qt-pro-mode-syntax-table`. That macro will also define a major mode map for you, so instead of declaring it explicitly yourself, you can simply put code at the top level which adds entries to the keymap, as long as you place that code after the `define-derived-mode` form.

purcell:
And re. `fic-mode`, you shouldn't decide for the user to enable this. It's orthogonal to this package, so you should leave it out: you can put it in your personal config easily enough: `(add-hook 'qt-pro-mode-hook 'fic-mode)`.

EricCrosson:
Thanks for the tips! I appreciate the code review. 

[This commit](https://github.com/EricCrosson/qt-pro-mode/commit/804ff95e0b060708c160347e58296320565769ed) implements your feedback.

purcell:
Great, thanks.

PR #4768, opened 2017-05-22
Remove :files in evil-goggles recipe

edkolev:

Adds global minor mode evil-goggles-mode, which, when on, briefly shows a colored overlay to indicate what text will be affected. For example, 2 dd will briefly show an overlay on the 2 lines which will be deleted before deleting them.

https://github.com/edkolev/evil-goggles

Maintainer.

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/evil-goggles
+++ b/recipes/evil-goggles
@@ -1,3 +1,1 @@
-(evil-goggles :repo "edkolev/evil-goggles"
-              :fetcher github
-              :files ("evil-goggles.el"))
+(evil-goggles :repo "edkolev/evil-goggles" :fetcher github)


edkolev:
I'll be splitting the single .el file into two files. This PR removes `:files` so melpa can pick up all `*.el` files.

Thanks

purcell:
Sure, that's a good idea anyway, but I'd suggest that it's not idiomatic to split a short (300-line) elisp package into multiple files.

edkolev:
I see.. I was planning to add a `...-faces.el` file where only `defface`s would be added. I'll reconsider and probably leave everything in the same file for now.

Thanks

purcell:
> I see.. I was planning to add a ...-faces.el file where only deffaces would be added. I'll reconsider and probably leave everything in the same file for now.

:+1:

Related tip, then: when defining new faces, try not to choose your own colours. Instead, use `:inherit` in the face definition to create your new faces as aliases for built-in faces. This way, themes will automatically look good, but users can still override the faces if desired.

PR #4769, opened 2017-05-23
add recipe flycheck-inline

stardiviner:

Display flycheck error message with inline style.

https://github.com/stardiviner/flycheck-inline

maintainer

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/flycheck-inline
+++ b/recipes/flycheck-inline
@@ -1,0 +1,1 @@
+(flycheck-inline :repo "stardiviner/flycheck-inline" :fetcher github)


stardiviner:
What does this CI error means?
```
  Scenario: read recipes
    Given read recipes
      Recipe 'flycheck-inline' contains mismatched package name 'flycheck-pos-tip'
2 scenarios (1 failed, 1 passed)
3 steps (1 failed, 0 skipped, 2 passed)
```

purcell:
> What does this CI error means?

Can you guess? ;-) Look at the content of your pull request -- there's an error with it. You've checked the box above which says "I've built and installed the package using the instructions in CONTRIBUTING.md", but that can't be true in this case.

stardiviner:
Damn, I forget to write the correct package name. I've updated recipe.

stardiviner:
@purcell I updated my recipe, and updated my package by adding a minor mode for easily toggle this minor mode.

purcell:
Thanks. Feedback on the code now:

- Please provide a very brief `Commentary` section describing how to use the code.
- `with-eval-after-load 'flycheck` is inappropriate: just `(require 'flycheck)` at the top level
- The minor mode is not a great solution, because when you disable it, it might not restore `flycheck-display-errors-function` to its previous value. Instead, just give the user instructions to `(setq flycheck-display-errors-function 'flycheck-inline)`, and add an `;;;###autoload` cookie comment for the `flycheck-inline` function.

stardiviner:
@purcell updated. Check out them.
Ask one question, why this minor mode toggle (disable) will not restore `flycheck-display-errors-function` to default value? even I set code like this?
```elisp
(define-minor-mode flycheck-inline-mode
  "This is a minor mode to display flycheck error message in inline way."
  :global t
  (if flycheck-inline-mode
      (setq flycheck-display-errors-function 'flycheck-inline)
    (setq flycheck-display-errors-function 'flycheck-display-error-messages)
    ))
```
If so, is there a correct solution?

purcell:
> Ask one question, why this minor mode toggle (disable) will not restore flycheck-display-errors-function to default value? even I set code like this?

Because `flycheck-display-errors-function` might not have been `flycheck-display-error-messages` when the mode was enabled.

What you're talking about here is a simple setting, not a toggled mode, so just remove the mode.

stardiviner:
What about enable this hook in major mode hook? Like:
```elisp
(add-hook 'clojure-mode-hook #'flycheck-inline-mode)
```
This will override flycheck default have been `flycheck-display-error-messages` value.

purcell:
> This will override flycheck default have been flycheck-display-error-messages value.

Only if someone hadn't previously set it to something different!

What you should probably do is just provide an autoloaded `flycheck-inline-enable` function which simply says `(setq-local flycheck-display-errors-function 'flycheck-inline)`.

purcell:
(I say `setq-local`, because if you want to add it to a major mode hook, presumably your intention is to change the local flycheck behaviour, not the global flycheck behaviour.)

stardiviner:
Ok, I will code, use `setq-local`, and update usage commentary.
I'm still uploading code.

stardiviner:
@purcell Hi, I updated my source code.

purcell:
> Hi, I updated my source code.

But the global minor mode is still there, and using `setq-local` in it is incorrect. As I suggested above, I think the mode should be removed and replaced with a simple "enable" command.

tarsius:
Friendly ping!

stardiviner:
@tarsius @purcell , Hi sorry for late response, I'm busy on life mess recently. I updated my code, check it out.

purcell:
Thanks. Sent you a couple of fixes in https://github.com/stardiviner/flycheck-inline/pull/1.

stardiviner:
@purcell merged. Thanks.

purcell:
Great!

PR #4770, opened 2017-05-23
Add pfuture recipe.

Alexander-Miller:

Wrapper around `make-process`, stuffing the process and its output into a single struct for convencience. Like `shell-command-to-string`, but async.

https://github.com/Alexander-Miller/process-future

I made this.

None needed.

### Checklist

It's complaining that I shouldn't use `pf-` as my function prefix, but I refuse to believe that naming a function `process-future-process-of` instead of `pf-process-of` is good practice, never mind the `process-future-future`struct and the resulting `process-future-future-process` accessor. I'd rather have the small namespace-y prefix.

### DIFF
--- a/recipes/pfuture
+++ b/recipes/pfuture
@@ -1,4 +1,3 @@
 (pfuture
  :repo "Alexander-Miller/pfuture"
- :fetcher github
- :files (:defaults))
+ :fetcher github)
--- a/recipes/pfuture
+++ b/recipes/pfuture
@@ -1,0 +1,4 @@
+(pfuture
+ :repo "Alexander-Miller/pfuture"
+ :fetcher github
+ :files (:defaults))


purcell:
Thanks, this looks handy.

> It's complaining that I shouldn't use pf- as my function prefix, but I refuse to believe that naming a function process-future-process-of instead of pf-process-of is good practice, never mind the process-future-futurestruct and the resulting process-future-future-process accessor. I'd rather have the small namespace-y prefix.

You're entitled to your preference, but I'm sorry to say that it's not a good one for the purposes of publishing this package for community use. Firstly, there are 676 (26x26) two-character prefixes available, and 4000+ packages in MELPA, so this obviously isn't a good general plan. And secondly, bytes are cheap, and lisp identifiers are already quite wordy.

By carefully choosing names, you can still have nice clear identifiers:

- process-future (instead of pf-go)
- process-future-await
- process-future-live-p (don't use `alive?`: it's not idiomatic in elisp, and this name matches `process-live-p`)

If you're really still offended by the names (and you shouldn't be), you might also consider renaming the package to `pfuture`, so that the identifiers are all of the form `pfuture-*`.

Alexander-Miller:
Oh I am not offended, that's really too strong a word. But naming things right is hard, and there are times I'd like to have something other than the overly long and descriptive identifiers I already deal with in my java day job. Part of the problem is of course that I couldn't think of anything better and process-future just stuck with me. pfuture however sounds good to me, so I'll do that.

Alexander-Miller:
All done.

purcell:
Awesome, looks great.

P.S. I'm not sure you need to depend on Emacs 25. I _suspect_ that depending on Emacs 24 and `cl-lib` 0.5 would do the trick.

Alexander-Miller:
Needing emacs 25 isn't about cl-lib. Flycheck-package told me to do it because I'm using `make-process`.

purcell:
Ah, cool.

PR #4775, opened 2017-05-27
Nubox color theme

martijnat:

Color theme with both dark and light version

https://github.com/martijnat/nubox

I am the maintainer

**None needed**

### Checklist

Please confirm with `x`:


purcell:
Thanks. Quick feedback:
- "A theme for personal use based of gruvbox" seems to no longer apply, if you're publishing this for wider use
- Please remove the `:files` spec from the recipe, which is unnecessary.
- There is no metadata for this package, because there is neither a `nubox-pkg.el` file nor a `nubox.el` containing standard elisp headers. One of these is mandatory if you really wish to publish this as a single "nubox" package. Otherwise, I'd suggest publishing each variant as its own package, which would mean submitting separate recipes, each with `:files` spec listing a single `.el` file.
- The formatting of the theme source is very messy and non-standard. If you're publishing this for wider use and possible contributions, now is the time to tidy it up and follow elisp standards. Try enabling `aggressive-indent-mode` and `flycheck`.
- Once installed, this package will not place its directory on the theme path, therefore the themes will not be loadable. This should have been obvious after building and installing the package locally, which you said you tried. Look in popular themes to find the snippet you'll need for this.

Hope that helps!

martijnat:
Thanks purcell,

I have updated the recipe and the package and will create a new pull request.

PR #4778, opened 2017-06-01
Add auto-minor-mode recipe.

joewreschnig:

This package lets you enable minor modes based on file name and contents. To find the right modes, it checks filenames against patterns in `auto-minor-mode-alist` and file contents against
`auto-minor-mode-magic-list`. These work like the built-in Emacs variables `auto-mode-alist` and `magic-mode-alist`.

(Versions of something like this have been swapped around for a long time, but usually without the magic alist equivalent, missing handling for some special cases, or requiring manual activation or activating at odd times.)

https://github.com/joewreschnig/auto-minor-mode

Maintainer.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/auto-minor-mode
+++ b/recipes/auto-minor-mode
@@ -1,0 +1,1 @@
+(auto-minor-mode :repo "joewreschnig/auto-minor-mode" :fetcher github)


purcell:
This is neat! Merging now.

I feel like those `defvar`s should be `defcustom`s, with a containing group, but I note that `auto-mode-alist` is not a `defcustom`, so who knows...?

I'm wary of the choice to follow `case-fold-search`: that's something a user might toggle at any time, so I think this is a sub-optimal decision.

joewreschnig:
Thank you for merging this so quickly!

Re. `defvar` vs . `defcustom - I don't like `customize` personally, and I really don't like it for non-trivial types (e.g. lambda functions) or things modified conditionally on feature/package availability. Probably the right approach is to have one manual list and one customizable list and run through both...

Re. `case-fold-search`, I'm open to changing especially if users find it unhelpful. I just picked the lazy option of ignoring it, but wanted to make sure the behavior was documented.

purcell:
Definitely agree re. lambdas in custom vars, but there's a `function` widget `:type` for those, and even if you don't use custom vars yourself, there's some value to making explicit which vars are *meant* to be customized by end users.

On the `case-fold-search` point, most "system-level" code that can be affected by case folding uses `let` to explicitly unset `case-fold-search`, to avoid unnecessary interactions. This package feels like a "system-level" package rather than an end-user package, so I reckon you should do the same, laziness notwithstanding. ;-)

PR #4779, opened 2017-06-02
Add json-navigator

DamienCassou:
View and navigate JSON structures

https://github.com/DamienCassou/json-navigator

package maintainer

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/json-navigator
+++ b/recipes/json-navigator
@@ -1,0 +1,1 @@
+(json-navigator :fetcher github :repo "DamienCassou/json-navigator")


purcell:
Thanks!

- There's no dependency on `hierarchy`, so you must declare one.
- Consider adding a command `json-navigator-region`, which allows browsing a selected portion of JSON text in a buffer. You could make its region default to the full contents of the current buffer.

DamienCassou:
Good ideas, thanks. I fixed both.

Note: I knew about `"There's no dependency on hierarchy"` but it took me some time to make package-lint happy about it (issue https://github.com/purcell/package-lint/issues/55). I fixed it by redoing what you do in https://github.com/purcell/package-lint/blob/master/run-tests.sh.

purcell:
Cool, merged. And now you can remove `(depends-on "hierarchy" "0.5.0")` from `Cask`.

PR #4782, opened 2017-06-04
Add solaire-mode

hlissner:

`solaire-mode` is an aesthetic plugin that helps visually distinguish
file-visiting windows from other types of windows (like popups or sidebars) by
giving them a slightly different -- often brighter -- background.

https://github.com/hlissner/emacs-solaire-mode

I am the maintainer.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/solaire-mode
+++ b/recipes/solaire-mode
@@ -1,0 +1,1 @@
+(solaire-mode :repo "hlissner/emacs-solaire-mode" :fetcher github)


purcell:
Thanks. The `eval-after-load "persp-mode"` is a bit out-of-place: normally `eval-after-load` is only for use in users' startup files. You can probably just execute that block's contents unconditionally at the top-level.

hlissner:
Huh, I was under the impression that `advice-add` would complain if the target function didn't exist at runtime. My mistake. In any case, I've done as you've suggested.

PR #4786, opened 2017-06-04
add new recipe yahtzee

drdv:

The yahtzee game

https://github.com/drdv/yahtzee

I am a maintainer

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/yahtzee
+++ b/recipes/yahtzee
@@ -1,0 +1,1 @@
+(yahtzee :repo "drdv/yahtzee" :fetcher github)


purcell:
Thanks, this is fun! Some quick feedback:

- For your face definitions, rather than defining specific colours, use `:inherit some-built-in-face`, e.g. `highlight`. This way, your package will immediately work with all existing themes, but will also be customisable.
- Instead of `(when (not cond) ...)`, use `(unless cond ...)`.
- Instead of `yahtzee-do-nothing`, use the built-in `ignore` function
- Prefer `"RET"` to `"C-m"`.
- `C-n` and `C-p` are confusing bindings, since they overload default Emacs navigation keys with very different meanings.
- If everything is encapsulated properly, it should be possible to start a game, rename the buffer and then start another game, without the games breaking each other. I don't think that's quite the case yet, but it is probably enough to use `defvar-local` for the variables which track game state, so that their settings are localised to the specific current game buffer.

Hope that helps!

drdv:
@purcell: Thanks a lot for your feedback. 

I have modified the code according to your suggestions. Unless I have overlooked something, now it should be possible to play the game in multiple buffers. Each time a new game is started using `M-x yahtzee`, a unique buffer name is created.

purcell:
Super, thanks!

PR #4792, opened 2017-06-10
Change github username in recipe

voloyev:
I had changed the username in Github, so I changed it in the recipe.


purcell:
Great, thanks.

PR #4795, opened 2017-06-11
New package hugsql-ghosts.

rkaercher:

Displays overlays of clojure defns generated by the [hugsql](https://www.hugsql.org) library in cider buffers.

https://github.com/rkaercher/hugsql-ghosts

I am the author.

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/hugsql-ghosts
+++ b/recipes/hugsql-ghosts
@@ -1,0 +1,1 @@
+(hugsql-ghosts :fetcher github :repo "rkaercher/hugsql-ghosts")


purcell:
Thanks for this.

Re. the following snippet

```el
;;;###autoload
(add-hook 'cider-mode-hook 'hugsql-ghosts-auto-show-ghosts)
;;;###autoload
(add-hook 'after-save-hook 'hugsql-ghosts-auto-show-ghosts)
```

the autoloads are inappropriate, because they cause the package to enable itself simply because it is installed. Just like the hundreds of built-in emacs packages, ELPA packages should not be enabled unconditionally. Instead, provide a function or describe a method which will activate the mode.

Additionally, I'm pretty sure you don't want this clojure-specific code running in response to any file anywhere in Emacs getting saved: you should probably be modifying `after-save-hook` buffer-locally in only `cider-mode` buffers.

rkaercher:
Ouch, you're right, these lines definitely do not belong there anymore, I just discovered autoloads shortly prior to making the PR and wasn't so sure what users would expect. 
As for the after save hooks, this was not really thought through from the sql side, I'll need to come up with a solution for this. I think I'll try going with suggesting users should add it to the sql-mode save hook and then check for an active cider connection (which I should do anyway).
Thanks for the feedback, I hope I can fix this tonight.

rkaercher:
I changed the script now to offer a simple function one can use to install a hook for cider mode that installs a buffer local hook which gets called every time the user reloads the clojure buffer which feel much more natural than the save hook when using it. Could you please review this once more?

purcell:
Great, thanks. Sorry for the delay.

PR #4796, opened 2017-06-11
Add fence-edit

aaronbieber:

Easily edit code in "fenced code blocks" in their appropriate major modes.

https://github.com/aaronbieber/fence-edit.el

Author/maintainer

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/fence-edit
+++ b/recipes/fence-edit
@@ -1,0 +1,3 @@
+(fence-edit
+  :fetcher github
+  :repo "aaronbieber/fence-edit.el")


purcell:
Hi! This kind of thing has been written before, and before we consider adding this to MELPA, I'd like to be sure that it doesn't reinvent the wheel. Have you ruled out any prior art? Specifically, "indirect buffers" are an ideal mechanism for editing sub-code blocks, and the [edit-indirect](https://melpa.org/#/edit-indirect) package supports this. [markdown-edit-indirect](https://melpa.org/#/markdown-edit-indirect) is built with this, and at the very least you should consider using `edit-indirect` in your package.

aaronbieber:
Thanks for the thoughtful feedback as always, Steve. I did originally try indirect buffers, but there seem to be a fair number of quirky behaviors when attempting to change major modes. Indirect buffers work brilliantly for things like Org Capture where some derived minor mode is applied and the major mode is the same, but I ran into trouble attempting to use a different major mode in the indirect buffer with my first solution.

So we have this, which yes, is a bit of a hack I suppose you could say, but it does what it says it does, and with no frills. I am submitting this for inclusion mostly because I received some positive feedback urging me to do so by users of this package (see linked issue), but if it doesn't belong here I will instruct users to install it directly from my repository.

Fanael:
But `edit-indirect` doesn't use indirect buffers.

aaronbieber:
I'll close this for now as I don't have time to investigate alternate implementations. Thanks for all the great feedback.

mnewt:
@purcell I would like to see the functionality of `fence-edit-code-at-point` somehow make it into melpa, and I'm willing to contribute if/where appropriate.

While I don't feel qualified to weigh in on the matter directly, it seems @aaronbieber, @Fanael, Org (`org-edit-src-code`), and `poporg` all have come to the conclusion that indirect buffers are not the ideal solution and that simply copying the region to a new buffer is a better approach.

I understand you don't want duplicate functionality in melpa and I completely agree that if there was one common package to do all of this that would be best. However, I'm not sure these packages are duplicating as much as it seems at first. They all basically use something along the lines of `insert-buffer-substring` to move text between buffers. So, the method is common, simple, and built in. It's all the functionality around that which differentiates them.

Despite the above reservations, I'm willing to take a crack at changing `fence-edit`'s internals to use indirect buffers if you and @aaronbieber feel that is the correct approach. Or I could graft `fence-edit-code-at-point` onto a different package directly or as a dependent.

What is your recommended path to get this functionality in melpa?

Issue #4798, opened 2017-06-12
Drop support for rarely used sources


Fuco1:
This is somewhat related to #4712.  If we want to have a way to fetch manifests and cache the dependency graphs we should ideally just use git that supports this best (with tags) or mercurial (but nobody uses mercurial).

I have counted 9 recipes using svn fetcher, 2 with darcs, 0 with fossil, ~0 with csv~, 4 with cvs, 1 with hg.

There is also 160 wiki packages which should become unavailable just because it's terrible. (see also #2342)

This is the first step towards making something that would be able to build packages on-the-fly (that is when user requests an old version not on the server, we would fetch it and build and serve).  The less sources we support the simpler the process becomes.

mdorman:
@Fuco1 Hmmm, did you actually look for `csv`?  Because that's not gonna find anything, but `cvs` will---admittedly, only 4, but `w3m` at least is a pretty prominent package.

Anyway, while I understand your utilitarian argument, I, personally, would be leery of instituting a monoculture---and I say that as the guy who once a week regenerates the melpa package list for `nix`, and thus has to deal with things like `cvs` servers that can't seem to get their act together.

Fuco1:
Silly me, I wrote `csv` yes :). 

The thing is, we already have a monoculture.  If someone is afraid of using git there is the emacs mirror project that can substitute for just the release snapshots.

It seems w3m is on github now: https://github.com/ecbrown/emacs-w3m (last update 10 days ago)

There is a lot of, well, crap, accumulated over the years.  I would very much like to donate my time to make things faster, simpler, cleaner and more convenient.  If only for myself as an author of 40+ emacs packages.

dunn:
How many use `bzr`?

Fuco1:
2 on my 3 months old copy, which is safe to assume is up to date with regard to bzr packages.

aplaice:
The `bitbucket` fetcher also uses mercurial. A year ago (when I compiled some stats) there were 37 sources fetching from there.

I personally think that locking oneself into a git-monoculture is not a great idea.

Also, in what way is a repository using, say `bzr` worse (in the context of MELPA stable) than one using `git` but not tags? Considering that there will always be `git` repositories without tags (people are lazy), and hence "non-stable" MELPA will presumably always be around, why can't the less frequently used sources just be excluded from MELPA stable?

dunn:
> why can't the less frequently used sources just be excluded from MELPA stable?

[...]

PR #4802, opened 2017-06-14
Remove bzr, cvs, darcs, fossil and svn support

tarsius:
This isn't the first time that I suggest that Melpa should drop support for little used fetchers. Unfortunately I cannot find those discussions anymore. If I remember correctly, @purcell's position was that he is not opposed on principal, but that he sees no reason to do so until these fetchers start causing issues.

The removal of some fetchers has recently come up again in #4798, where @Fuco1 said:

> All complex systems tend to break, this is a given in programming. Reducing complexity promotes experimentation and growth. When rolling out "partial support" is met without much enthusiams, then we should simply drop the support altogether.

I agree with that. I have contributed to `package-build.el` in the past and the existence of little used fetchers was indeed an issue for me.

----

This pull requests attempts to demonstrate how we can drop support for *three* vcs fetchers without losing much. I hope that by making the proposal very concrete, we can actually reach a decision on whether to do this or not.

Please see the commit messages for details. Here is a summary:

* Drop support for the `fossil` fetcher. Not a single package used it.
* Drop support for the `bzr` fetcher. The *two* recipes that still used `bzr` now import from stale repositories on the Emacs*orphanage*. The upstream repositories have not seen any commits in years. But in the unlikely event that upstream decided to resume work, we would *not* automatically notice that.
* Drop support for the `darcs` fetcher. The *two* recipes that still used `darcs` now import from active repositories on the Emacs*mirror*. These repositories are only updated periodically, so there sometimes will be delays.

So by migrating *four* packages we can drop support for *three* fetchers. There are small drawbacks as mentioned, but I think that the reduced complexity outweigh those by far. 

Support for `cvs` and `svn` cannot be dropped yet because this pr only migrates *seven* out of the remaining ~~*thirteen*~~ *twelve* packages to the Emacsmirror.

* *Two* `cvs`-using packages have not been migrated so far. Both of these packages actually live in the same repository and I have asked upstream to consider a move to `git` (but have not gotten a reply yet). If that's what it takes, then I am willing to periodically import the `cvs` repository myself.
* ~~*Four*~~ *Three* `svn`-using packages have not been migrated so far because they consist of more than one library, which means that the Emacsmirror cannot use its `file` fetcher.
  * `caml` last updated two years ago. Let's ask upstream to migrate, but there is a good change that would fail.
  * `confluence` last updated two years ago. Let's ask upstream to migrate.
  * `dic-lookup-w3m` actively maintained. Let's ask upstream to migrate.
  * ~~`ruby-electric` @knu the file in the svn repository contains a link to https://github.com/knu/ruby-electric.el. Is the git repository up-to-date/ahead? Old discussion in #1079.~~ [Given that the url in both the svn and git repository points at the git repository, I decided that lacking a response the git repository should be preferred.]

### DIFF
--- a/scripts/bootstrap
+++ b/scripts/bootstrap
@@ -9,18 +9,13 @@ cd ${HOME}
 sudo ${SUDOENV} apt-get -y update
 sudo ${SUDOENV} apt-get -y upgrade
 sudo ${SUDOENV} apt-get -y install \
-    bzr \
     curl \
-    cvs \
-    darcs \
     emacs24 emacs24-el emacs24-common-non-dfsg \
-    fossil \
     git \
     make \
     mercurial \
     nginx \
     runit \
-    subversion \
     texinfo \
     tmux
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,18 +3,13 @@ language: ruby
 addons:
   apt:
     packages:
-    - bzr
     - git
     - mercurial
-    - subversion
-    - cvs
-    - darcs
-    - fossil
     - texinfo
 
 before_install:
   - sudo apt-get -yq update
-  - sudo apt-get -yq install bzr git mercurial subversion cvs darcs fossil texinfo
+  - sudo apt-get -yq install git mercurial texinfo
 
   - git clone https://github.com/rejeep/evm.git $HOME/.evm
   - export PATH=$HOME/.evm/bin:$PATH
--- a/README.md
+++ b/README.md
@@ -105,42 +105,43 @@ the following form (`[...]` denotes optional or conditional values),
 
 ```lisp
 (<package-name>
- :fetcher [git|github|gitlab|bitbucket|bzr|hg|darcs|fossil|svn|cvs|wiki]
+ :fetcher [git|github|gitlab|bitbucket|hg|wiki]
  [:url "<repo url>"]
  [:repo "github-gitlab-or-bitbucket-user/repo-name"]
  [:version-regexp "<regexp>"]
- [:module "cvs-module"]
  [:files ("<file1>" ...)])
 ```
 
 - `package-name`
 a lisp symbol that has the same name as the package being specified.
 
-- `:fetcher` (one of `git, github, gitlab, bitbucket, bzr, hg, darcs,
-fossil, svn, cvs, wiki`) specifies the type of repository that `:url`
-points to. Right now package-build supports [git][git],
-[github][github], [gitlab][gitlab], [bitbucket][bitbucket],
-[bazaar (bzr)][bzr], [mercurial (hg)][hg], [subversion (svn)][svn],
-[cvs][cvs], [darcs][darcs], [fossil][fossil], and
-[EmacsWiki (deprecated)][emacswiki] as possible mechanisms for checking out
-the repository. (Note: `bitbucket` assumes `hg`: `git` repos hosted on
-bitbucket should use the `git` fetcher.)
+- `:fetcher` specifies the type of repository that `:url` or `:repo`
+  points to.  Melpa supports [`git`][git], [`github`][github],
+  [`gitlab`][gitlab], [`hg`][hg] (Mercurial), and [`wiki`][emacswiki]
+  (for packages from Emacswiki.org).
 
-    *package-build* uses
-the corresponding application to update files before building the
-package. In the case of the `github`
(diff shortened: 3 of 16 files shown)


package-build/package-build.el
```diff
@@ -803,9 +684,6 @@ Optionally PRETTY-PRINT the data."
          "--exclude=CVS"
          "--exclude=.git"
          "--exclude=_darcs"
```
dunn: This line can go too, right?

scripts/bootstrap
```diff
@@ -12,9 +12,7 @@ sudo ${SUDOENV} apt-get -y install \
     bzr \
```
dunn: Can kill bzr here too?

package-build/package-build.el
tarsius: Yes, thanks for catching that.

scripts/bootstrap
tarsius: ditto

tarsius:
Note that the changes to `package-build.el` should be made in its own repository. I haven't done that yet to make it easier to review all the changes in one place.

Fuco1:
@tarsius thank you very much for taking up the time, I know you are super busy.  Let me know if I can do some work, I will happily donate some time to this as well.

tarsius:
> Let me know if I can do some work,

You could send friendly mails to the maintainers of packages that still use `svn` and offer to help them convert to `git`.

purcell:
I feel a bit weird about relying on Emacsmirror, but I'd be happy to merge this if the conflicts get resolved. Would you then sync the changes back to the `package-build` repo, @tarsius?

tarsius:
I am now importing the full history of most of these packages into git repositories using a simple Makefile again. More on that later, this response is about two of the four packages that I do not import again (yet?).

`debian-bug` and `debian-changelog-mode` are only interesting to Debian users, who actually contribute to Debian, and who probably would not mind installing them using `apt-get install debian-el dpkg-dev-el` (which has the added benefit that these packages come with other closely related Emacs libraries).

`debian-bug` has been downloaded only 69 times from Melpa and `debian-changelog-mode` only 168 times.

So my suggestion is that those two packages are removed from Melpa.

purcell:
> debian-bug has been downloaded only 69 times from Melpa and debian-changelog-mode only 168 times.
> So my suggestion is that those two packages are removed from Melpa.
> 

Fine with me!

tarsius:
Okay will do it.

package-build/package-build.el
```diff
@@ -802,10 +583,6 @@ Optionally PRETTY-PRINT the data."
          "--exclude=.svn"
          "--exclude=CVS"
```
dunn: This can go too, probably.

package-build/package-build.el
```diff
@@ -1524,7 +1301,7 @@ If FILE-NAME is not specified, the default archive-contents file is used."
    (list (intern (read-string "Package name: "))
          (intern (completing-read "Fetcher: "
                                   (list "github" "gitlab" "bitbucket"
-                                        "git" "wiki" "bzr" "hg" "cvs" "svn")
+                                        "git" "wiki" "hg" "cvs" "svn")
```
dunn: We can't also drop `cvs` and `svn` from this list?

alphapapa:
Just a thought: developers who run other distros (or even, e.g. OS X) but also package their software for Debian may find `debian-changelog-mode` useful, and installing it manually by extracting it from the deb would be a pain.

joewreschnig:
I think it would be worthwhile to split off parts of this for each VCS, as each has different issues:

- Fossil is literally unused, and could go right away right now and no one would notice.
- CVS is fundamentally insecure, and at some point the bullet must be bit and we drop `w3m` (the only remaining repository) because of that alone.
- Bzr is dead, unofficially for certain if not officially yet. ([Even its official download page no longer works.](http://wiki.bazaar.canonical.com/Download)) 
- Darcs and Subversion should probably be discouraged, but are secure and actively maintained.

joewreschnig:
> Just a thought: developers who run other distros (or even, e.g. OS X) but also package their software for Debian may find debian-changelog-mode useful, and installing it manually by extracting it from the deb would be a pain.

This is the situation I'm in now, but what the removal is probably going to do is get me off my butt to fork and modernize the package, so I think it's win/win (other than my free time).

tarsius:
Ups...

PR #4804, opened 2017-06-15
Add recipe for answer set programs.

santifa:

A major mode for editing Potassco Answer Set Programs files (https://potassco.org/).

https://github.com/santifa/pasp-mode

Author

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/pasp-mode
+++ b/recipes/pasp-mode
@@ -1,0 +1,1 @@
+(pasp-mode :fetcher github :repo "santifa/pasp-mode")


purcell:
Thanks, merging now. Couple of tips:

- Since you're requiring emacs 24.3, you can use `setq-local` instead of `set` + `make-local-variable`.
- If you move your key binding definitions below `define-derived-mode`, you'll find it will have declared `pasp-mode-map` for you. :-)
- `use-local-map` and `set-syntax-table` are called for you by `define-derived-mode`, so you can remove those invocations.

PR #4809, opened 2017-06-18
Add livereload recipe

stapelberg:

Instructs browsers with the LiveReload extension to reload their tabs displaying the file which was just saved.

https://github.com/joaotavora/emacs-livereload

Best-effort maintainer (see https://github.com/joaotavora/emacs-livereload/issues/1#issuecomment-309271964)

None needed

### Checklist

Please confirm with `x`:

checkdoc identifies a few problems, but I think there’s value in having the package in melpa as-is.

### DIFF
--- a/recipes/livereload
+++ b/recipes/livereload
@@ -1,0 +1,3 @@
+(livereload :repo "joaotavora/emacs-livereload"
+	    :fetcher github
+	    :files (:defaults "livereload.js"))


purcell:
Thanks. Quick feedback:
- You need to declare a package dependency on `websocket`.
- You should remove `for Emacs` from the package description line, as `package-lint` will have suggested to you: it's redundant, since all Emacs packages are for Emacs. :-)  Maybe something like "Tell the LiveReload browser extension when files are saved"

stapelberg:
Thanks for the feedback! Both points addressed.

purcell:
Great, thanks! Welcome to MELPA.

PR #4811, opened 2017-06-19
Add recipe for makefile-executor

thiderman:

Various extensions to makefile-mode! Currently aimed at working with Makefile targets from a project perspective. If there are multiple makefiles, a `completing-read` between them is shown! When a Makefile is selected, another selection with all the possible targets is shown.

https://github.com/thiderman/makefile-executor.el

Maintainer!

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/makefile-executor
+++ b/recipes/makefile-executor
@@ -1,0 +1,1 @@
+(makefile-executor :repo "thiderman/makefile-executor.el" :fetcher github)


purcell:
Thanks, this looks handy! Quick feedback:
- Don't interpolate file paths into commands without using `shell-quote-argument`
- Rather than modify `makefile-mode-map`, you should generally provide a little minor mode, and then tell users to add it to `makefile-mode-hook`.
- Just adding `+` is a slightly lazy way to name this package, particularly when it's such a small and narrow add-on package. :-)

thiderman:
Great feedback! I'll address the points and make adjustments accordingly! Thanks! :)

thiderman:
I have addressed all the points of feedback and renamed the package! 💃 

I wasn't super sure of where to put the `shell-quote-argument`, mostly because all of the filenames involved are temporary files generated by emacs itself. Regardless, best practices are a good thing and I'd rather follow them than not!

purcell:
Thanks.

PR #4813, opened 2017-06-21
Add lean-mode recipe

jroesch:

This package adds a major mode for the Lean programming language and theorem prover.

https://github.com/leanprover/lean-mode

I am a maintainer of the package and of Lean itself. This mode was mostly written by @leodemoura @gebner @Kha. I merely factored it into a MELPA package.

None needed, I have push access to the package.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/lean-mode
+++ b/recipes/lean-mode
@@ -1,0 +1,3 @@
+(lean-mode
+ :repo "leanprover/lean-mode"
+ :fetcher github)


jroesch:
One question are warnings during ELISP compilation acceptable? I was able to successfully compile the byte code, just curious.

purcell:
Thanks. Quick feedback:
- `eri.el` and `load-lean.el` don't fit the naming of the other files, and should perhaps be removed or renamed to use the `lean-` prefix. The latter, in particular, does not look like library code that should be distributed. The former might even make sense as a separate package upon which `lean-mode` would depend.
- `lean-pkg.el` is not a valid package descriptor file, which is confusing because that's what `-pkg.el` generally denotes. Here it's not going to break MELPA because MELPA would look for an optional `lean-mode-pkg.el`, but it's something to note.
- `lean-mode.el` has two leading package description lines.
- A major mode should not depend on `flycheck` or `company`, because those are two packages which are very much a matter of user preference, and outside the scope of a major mode. You can provide separate `flycheck-lean` and `company-lean` packages to provide that functionality, each of which would depend on `lean-mode`. (These can still be kept in the same repository, with the right arrangement of MELPA recipes.)
- There are functions which aren't properly namespaced, e.g. `replace-regex-return-position`.

jroesch:
Thanks for the feedback, I was on vacation for the past week and just got back. 

> A major mode should not depend on flycheck or company, because those are two packages which are very much a matter of user preference, and outside the scope of a major mode. You can provide separate flycheck-lean and company-lean packages to provide that functionality, each of which would depend on lean-mode. (These can still be kept in the same repository, with the right arrangement of MELPA recipes.)

From our perspective, which I concede may not be best practice in the Emacs community, splitting out functionality of `lean-mode` into multiple packages is a lot of extra work for us given the ratio of code to maintainers.

Is it possible to make these dependencies optional features of the package like other dependent type theory modes have done, for example `idris-mode` https://github.com/idris-hackers/idris-mode/blob/dc122c178c2a0ddda36fccdd0d3976fc7cd27245/idris-mode.el#L171. ?

Thanks again for the feedback 😄

purcell:
> Is it possible to make these dependencies optional features of the package like other dependent type theory modes have done, for example idris-mode

Yep, that's borderline okay, but not recommended. (I think the nested `eval` in the code you linked might be to mask the reasonable byte compilation issues which arise.) But if you want to do that for now, you can consider splitting later. From my perspective, it's super easy to split packages within one repo, so it doesn't have to be a lot of extra work for you.

tarsius:
@jroesch @Kha see this package is being worked on actively - have the above issues been addressed?

jroesch:
I chatted with @Kha and he was planning on addressing the feedback provided (since he actively maintains the mode) and updating this PR, maybe he can weigh in whether anything has changed?

Kha:
Yes, all issues should be resolved now*. I have updated the recipes here: https://github.com/Kha/melpa/commit/8d87ef9e1fbada91730d04ad6be54abab2725b6a. Perhaps @jroesch can merge that into this PR.

*Note that while I have created new `company-lean` and `helm-lean` packages, `lean-mode` still depends on flycheck, because I found this dependency to be too essential to remove. The one thing you expect from a plugin for a proof assistant is to tell you when you are wrong.

shlevy:
@purcell @tarsius Anything else needed here?

tarsius:
> Note that while I have created new `company-lean` and `helm-lean`

Additionally you should exclude them from the `lean-mode` using the `:files` keyword and add two new recipes so that these packages can be installed separately.

Kha:
I have done so at https://github.com/Kha/melpa/commit/8d87ef9e1fbada91730d04ad6be54abab2725b6a - I can open a new PR from there if that's easier.

tarsius:
Just force-push to the `lean-mode` branch.

Kha:
I can't, since I don't have write access to @jroesch's fork.

tarsius:
Ah! Yes, in that case just open a new pr.

PR #4814, opened 2017-06-21
Drop relative-line-numbers

Fanael:
It's buggy and unmaintained, and will be superseded by the display engine line numbering feature when it lands.

### DIFF
--- a/recipes/relative-line-numbers
+++ b/recipes/relative-line-numbers
@@ -1,1 +1,0 @@
-(relative-line-numbers :fetcher github :repo "Fanael/relative-line-numbers")


purcell:
I always wonder whether it's worth removing packages so aggressively: presumably this one is working for a number of people, and there's currently no viable alternative. Why not just note in the README that it's unmaintained, and turn off the issue tracker?

Fanael:
> there's currently no viable alternative

There's linum-relative and nlinum-relative.

purcell:
Oh cool. My mistake - merged! Thanks.

jgkamat:
@Fanael I've always loved your line number package, it's way faster than all the other ones I've tried, and I'm still going to be using it. I'm sad to see it go, but thanks for creating such a wonderful package! :smile:

PR #4819, opened 2017-06-24
Add bifocal recipe

riscy:

In bifocal-mode, paging up causes a comint-mode buffer to be split in two windows with a larger window on top (the head) and a smaller input window preserved on the bottom (the tail).  This helps with monitoring new output and entering text at the prompt (in the tail window), while reviewing previous output (in the head window).  Paging down all the way causes the split to disappear.

https://github.com/riscy/bifocal-mode

Maintainer

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/bifocal
+++ b/recipes/bifocal
@@ -1,0 +1,2 @@
+(bifocal :repo "riscy/bifocal-mode"
+         :fetcher github)


purcell:
Thanks. This is interesting.

BTW, the `turn-on` function in `define-globalized-minor-mode` doesn't need to be `interactive`. I feel like you could just define a regular global minor mode that adds/removes `bifocal-mode` from `comint-mode-hook`, since the current code causes the check to happen in every buffer.

riscy:
Awesome, thank you for the feedback!

riscy:
Following up -- I think just adding the `turn-on` to `comint-mode-hook` means the mode won't get turned off in existing buffers when someone calls `bifocal-global-mode` a second time.  Thanks again. 👍

PR #4823, opened 2017-06-28
+ config-general-mode: major mode for config files

TLINDEN:
config-general-mode

This major mode can be used to help editing config files parseable by Config::General (http://search.cpan.org/dist/Config-General/), a perl config file parser/writer used by many tools (e.g. Catalyst). 

https://github.com/TLINDEN/config-general-mode

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/config-general-mode
+++ b/recipes/config-general-mode
@@ -1,0 +1,2 @@
+(config-general-mode :repo "tlinden/config-general-mode" :fetcher github)
+


purcell:
Thanks! Sorry for the delay. Some quick feedback:

- Don't use `(if (not cond) (progn ...))`: just use `(unless cond ...)`
- Similarly, use `(when cond ...)` instead of `(if cond ...)` -- `if` is really only for expressions that have both an `if` and and `else` branch.
- `config-general-tab-or-complete` is unusual: rather than overriding completion behaviour (and `TAB` specifically, which not everyone will expect to complete) you should instead provide a `completion-at-point-function`, and then the user's preferred completion mechanism can use that function.
- You're doing a lot of things in `define-derived-mode` that that macro can handle itself, e.g. setting the mode name and symbol, running the hooks, setting the syntax table and keymap etc. So you can - and should - remove the redundant code.
- Unsetting local keybindings in your major mode initialisation code is weird and should be unnecessary.

Hope that helps!

TLINDEN:
Hello Steve,

thanks for your feedback!

- I fixed the `(if` issues (but not in `config-general--try-expand-dabbrev-all-buffers`, which is a patched version of `try-expand-dabbrev-all-buffers`)
- I have a problem with `(setq major-mode 'config-general-mode)`: if I remove this call, then the mode will be loaded but `describe-mode` doesn't mention it (whereas it does when I leave the call in the code)
- Similarly if I disable `(setq mode-name "C::G")` then the modeline displays "Fundamental"

Do you have any idea for the former two issues?


best regards,
Tom

TLINDEN:
PS: the keymap doesn't work as well, if I remove `(use-local-map config-general-mode-map)`. This is really weird.

purcell:
Sent you a bunch of fixes in https://github.com/TLINDEN/config-general-mode/pull/1.

TLINDEN:
Merged, thanks a lot!

purcell:
Merged here too - thanks! :-)

PR #4825, opened 2017-06-29
receipe for save-cwd

hardaker:

save-cwd is a minor mode that saves the current working directory of the file being edited every few seconds at idle-time

https://github.com/hardaker/elisp-save-cwd

Creator, owner, maintainer, person to throw stones at.

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/save-cwd
+++ b/recipes/save-cwd
@@ -1,0 +1,2 @@
+(save-cwd :repo "hardaker/elisp-save-cwd" :fetcher github)
+


purcell:
This is not really a great idea, IMO, and I don't think the current global minor mode implementation will even work reliably as intended. Emacs doesn't have a single global current working directory: every buffer can have a different one. There's nothing here to obviously tie the directory saved within the timer callback to the currently focused or active buffer.

I'm going to decline to merge this then, sorry, but if it works for you locally then great!

hardaker:
Bummer.  People do use it, it does work and the reason I submitted the request is people wanted it in melpa.

PR #4827, opened 2017-07-01
Add vc-msg

redguardtoo:

Show commit message of current line

https://github.com/redguardtoo/vc-msg

I'm the developer

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/vc-msg
+++ b/recipes/vc-msg
@@ -1,0 +1,1 @@
+(vc-msg :fetcher github :repo "redguardtoo/vc-msg")


purcell:
Thanks, and sorry for the slow reply. Quick feedback:
- Missing dependency on `cl-lib`.
- The various `defvar`s that are for end-users to modify should be `defcustoms` instead, with a containing `defgroup`.
- Typo: "messeage"
- Why all the code to determine the current VCS? `vc.el` provides this for you.
- Similarly, doesn't `vc-annotate` give you all the information you need for this library?

redguardtoo:
Thanks. I fixed typo and replace `defvar` with `defcustom`, now `vc-msg` requires `Emacs 24.3` which has `cl-lib` built in.

The purpose of this packages is to give user full freedom to support any VCS (for example, Perforce support is not built into `vc.el`). I have to write new code to support Perforce. Other open source VCS is just one liner like `(locate-dominating-file default-directory algorithm)`, it will not make my code cleaner to use `vc.el`

similar reason for `vc-annotate`. for example, user can customize `git blame` flag like `-M`.

This package is like `git-messenger` but supports any alien VCS.

redguardtoo:
@tarsius , anything else to change before being merged?

redguardtoo:
Thanks

PR #4829, opened 2017-07-02
Add rufo-mode

danielma:

Automatically formats ruby files with [rufo][]

https://github.com/danielma/rufo-mode.el

Author and maintainer.

None needed

### Checklist

Please confirm with `x`:

[rufo]: https://github.com/asterite/rufo

### DIFF
--- a/recipes/rufo
+++ b/recipes/rufo
@@ -1,0 +1,1 @@
+(rufo :repo "danielma/rufo.el" :fetcher github)
--- a/recipes/rufo-mode
+++ b/recipes/rufo-mode
@@ -1,0 +1,1 @@
+(rufo-mode :repo "danielma/rufo-mode.el" :fetcher github)
--- a/recipes/rufo-mode
+++ b/recipes/rufo-mode
@@ -1,1 +1,0 @@
-(rufo-mode :repo "danielma/rufo-mode.el" :fetcher github)


purcell:
Thanks. Quick feedback:
- Don't [build command strings](https://github.com/danielma/rufo-mode.el/blob/85ad3dfbcce460b7f430762b6da3d20bfb13089e/rufo-mode.el#L151) without using `shell-quote-argument`, or you'll have a bad time!
- Just use `executable-find` to [see if a command is available](https://github.com/danielma/rufo-mode.el/blob/85ad3dfbcce460b7f430762b6da3d20bfb13089e/rufo-mode.el#L154-L165).
- A bunch of this code has been cargo-culted from elsewhere (gofmt.el?). Do you really need the RCS patch support?
- Similarly, [rufo-mode--delete-whole-line](https://github.com/danielma/rufo-mode.el/blob/85ad3dfbcce460b7f430762b6da3d20bfb13089e/rufo-mode.el#L70) is bogus, as I've told [other authors who've copied that code from elsewhere](https://github.com/melpa/melpa/pull/4765#issuecomment-304495297).
- You should probably credit the original author.
- I'm not suggesting you rename this library (necessarily) but it's conventional to use the `-mode` suffix in package names only when they provide a major mode. There are plenty of exceptions, but `rufo.el` would have been fine here.

Hope that helps!

danielma:
Thanks for all the great feedback. I've done my best to update the package to match your suggestions, with a few exceptions.

> Don't build command strings without using shell-quote-argument, or you'll have a bad time!

Done!

> Just use executable-find to see if a command is available.

The reason I _didn't_ do this is because the package also supports usage through `bundler`, and so in that case the function is checking both that `(executable-find "bundle")` exists and that `rufo` is available in the context of `bundle exec`.
 
> A bunch of this code has been cargo-culted from elsewhere (gofmt.el?). Do you really need the RCS patch support?

I would say yes, it does need the RCS patch support. The [`rufo` README](https://github.com/asterite/rufo#tips-for-editor-plugins-implementors) specifically recommends that all editor plugins preserve cursor position, compute a diff between the old and new versions, and apply that diff rather than just reloading the buffer from disk.

> Similarly, rufo-mode--delete-whole-line is bogus, as I've told other authors who've copied that code from elsewhere.

Great. Removed it.

> You should probably credit the original author.

I added a section to the code and to the README crediting the `go-mode.el` authors.

> I'm not suggesting you rename this library (necessarily) but it's conventional to use the -mode suffix in package names only when they provide a major mode. There are plenty of exceptions, but rufo.el would have been fine here.

If you say it's the convention, I will happily follow. I've renamed to [`rufo.el`](https://github.com/danielma/rufo.el).

purcell:
> The reason I didn't do this is because the package also supports usage through bundler

I thought that might be the case.

> I would say yes, it does need the RCS patch support.

Awesome.

> Great. Removed it.

Ah. I didn't realise it wasn't even being called. Great.

I'll go ahead and merge this now, because I think it's a more solid package than the identically-named one submitted in #4831.

PR #4830, opened 2017-07-02
Add recipe for a.el

dotemacs:
As per subject, package that provides functions for dealing with
association lists and hash tables

Emacs Lisp functions for dealing with association lists and hash tables. Inspired by Clojure.

https://github.com/plexus/a.el

An enthusiastic user

Waiting for the maintainer @plexus to merge this PR:

https://github.com/plexus/a.el/pull/3 

and fix up the docstrings.

### Checklist

Please confirm with `x`:

Not the case, waiting on the maintainer @plexus to fix them

### DIFF
--- a/recipes/a
+++ b/recipes/a
@@ -1,0 +1,1 @@
+(a :repo "plexus/a.el" :fetcher github)


plexus:
Checkdoc is now *mostly* happy with the docstrings. It still complains about some things that I fundamentally disagree with and would rather not change. If full checkdoc compliance is a hard requirement for getting on to MELPA then I'll change them though. Thanks!

purcell:
Yep, please update the docstrings. It's easy enough, and you're sharing this for community benefit, so it's important to follow community conventions. Personal preferences are fine, for personal code. (I'm already biting my tongue about the one-character library name, and the `?` suffixes which should be `-p` for Emacs Lisp... while I'm a fan of Clojure it's important IMO to follow local conventions.)

plexus:
Hi @purcell, thanks for the feedback. Checkdoc should be happy now.

There are `a-has-key` and `a-equal` aliases for `a-has-key?` and `a-equal?`. I thought of adding the `-p`, but I think `has-key` and `equal` already imply that it's a predicate. I also prefer `has-key` over `key-p` to make a distinction between "having" and "being" a key. I hope that makes sense.

plexus:
Is there anything else we should address here or is this good to go? Thanks!

bbatsov:
I'm also mentioning here the new built-in https://github.com/emacs-mirror/emacs/blob/master/lisp/emacs-lisp/map.el as it seems the 3 of you were not aware of it. I think it might make sense to contribute to `map.el` if something important is missing there. //cc @NicolasPetton

bbatsov:
To be clear - I'm not saying this PR should not be merged. I just mentioned `map.el` for the sake of completeness and future collaboration.

dotemacs:
I updated the checklist, all the options are ticked now.

dotemacs:
@bbatsov regarding your second comment

> To be clear - I'm not saying this PR should not be merged. I just mentioned map.el for the sake of completeness and future collaboration.

That's how I understood it, thanks

NicolasPetton:
@plexus I'd be happy to discuss the design and future development of map.el with you, if you're interested.

plexus:
Thanks @NicolasPetton, I think it's great we have map.el, but it seems it has different design goals from a.el.

- `map-delete`, `map-put` mutate their arguments, a.el treats inputs as immutable
- value semantics

``` emacs-lisp
(map-elt (a-list "foo" 123) "foo")
;;=> nil

(a-get (a-list "foo" 123) "foo")
;;=> 123
```

- constructors for alist/map "literals" (`(a-list :foo 123)` / `(a-hash-map :foo 123)`)

So there seems to be enough reason for them to co-exist.

@purcell as far as I can tell we comply with all formal requirements for being included in MELPA. Is there something more we could/should do?

We have severeal packages in the works (Clojure/EDN parser and Unrepl client) that depend on `a.el`, and that we eventually would also like to get into MELPA, so it would be really helpful if we could get this in.

purcell:
Thanks all!

alphapapa:
@NicolasPetton 

> @plexus I'd be happy to discuss the design and future development of map.el with you, if you're interested.

I'm not him, but if I may: I think he has a point that this seems a bit counter-intuitive, or at least not as useful as it could be:

```el
(map-elt (a-list 1 "one") 1) => "one"

(map-elt (a-list "one" 1) "one") => nil
```

plexus:
Both options are valid. `map-elt` checks with eql because it's fast, as do
many other elisp functions. `a-get` uses equal because I'm happy to
sacrifice some performance to have these semantics.

alphapapa:
@plexus Good point.

By the way, am I missing something, or does it actually use `eq`?
> 
> map-elt is a compiled Lisp function in ‘map.el’.
> 
> (map-elt MAP KEY &optional DEFAULT)
> 
> Lookup KEY in MAP and return its associated value.
> If KEY is not found, return DEFAULT which defaults to nil.
> 
> If MAP is a list, ‘eql’ is used to lookup KEY.

```
(defun map-elt (map key &optional default)
  ...
  (map--dispatch map
    :list (alist-get key map default)
    :hash-table (gethash key map default)
    :array (if (and (>= key 0) (< key (seq-length map)))
               (seq-elt map key)
             default)))
```
```
(defun alist-get (key alist &optional default remove)
  ...
  (ignore remove) ;;Silence byte-compiler.
  (let ((x (assq key alist)))
    (if x (cdr x) default)))
```

> assq is a built-in function in ‘C source code’.
> 
> (assq KEY LIST)
> 
> Return non-nil if KEY is ‘eq’ to the car of an element of LIST.
> The value is actually the first element of LIST whose car is KEY.
> Elements of LIST that are not conses are ignored.

NicolasPetton:
`assoc` (and `alist-get`) now supports an optional `testfn` parameter, which can replace `eq`:
https://git.savannah.gnu.org/cgit/emacs.git/commit/?id=0bece6c6815cc59e181817a2765a4ea752f34f56

Support was also added to map.el:
https://git.savannah.gnu.org/cgit/emacs.git/commit/?id=76e1f7d00fbff7bf8183ba85db2f67a11aa2d5ce

alphapapa:
LOL, wow.  Well, I guess I can be forgiven for not noticing it, since it was only added in the last week.  Thanks, Nicolas.  :)

PR #4831, opened 2017-07-02
Add rufo

aleandros:

Automatically (or manually) formats ruby and enhanced-ruby buffers with [rufo](https://github.com/asterite/rufo)

https://github.com/aleandros/emacs-rufo

Creator + maintainer

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/rufo
+++ b/recipes/rufo
@@ -1,0 +1,1 @@
+(rufo :repo "aleandros/emacs-rufo" :fetcher github)


purcell:
Thanks. #4829 provides a similar package, which has more idiomatic code (partly because it was mostly copied from elsewhere).

This package is smaller (great!) but also quite unidiomatic, and it needs a fair bit work before it could be included here:
- Typo on the very first line
- `add-hook` shouldn't be called unconditionally: provide a minor mode that users can enable themselves, by adding it to their `ruby-mode-hook` if desired.
- Don't use `?` as a function suffix: `-p` is idiomatic in elisp.
- Don't execute `which` to find `rufo`: there's a function called `executable-find` which does this.
- Don't modify the *global* version of `before-save-hook` and then check that it's a `ruby-mode` file: instead, modify the local version of that hook so that you don't have to check! See the `LOCAL` argument to `add-hook`.
- Don't modify other modes' keymaps: put them in the keymap for your own minor mode. Also, `C-c f` is [reserved for end-user customisation](https://www.gnu.org/software/emacs/manual/html_node/elisp/Key-Binding-Conventions.html).

purcell:
In the end I merged the Rufo package submitted in #4829, which is also (now) called `rufo.el` -- I suggest you give that package a try and see how it works for you. I'm sure the author would be happy to have help.

aleandros:
Thanks a lot for the feedback! Not everyone is as thorough or takes the time to explain these things.

purcell:
No worries. I figure that the more people know about the kind of things I mentioned, the better off the community will be, and the less I'll be the one giving tips. :-)

aleandros:
We need "Effective Elisp" to be a thing.

PR #4833, opened 2017-07-03
Add snippets for ES-mode

dakrone:
This was previously an elisp-only repo, but now includes optional snippets for
yasnippet.

Relates to https://github.com/dakrone/es-mode/issues/36


purcell:
Great, thanks

PR #4834, opened 2017-07-03
Add eve-mode recipe

joshuafcole:
eve-mode is a major mode for editing [Eve documents][1] built on top of [polymode][2]. It provides syntax highlighting and simple eve-construct-aware indentation for authoring Eve documents (GFM-flavored markdown with code embedded as fenced code blocks).

https://github.com/witheve/emacs-eve-mode

Maintainer of both `eve-mode` and Eve itself.

My therapist informs me that conversing with myself in public is seen as unhealthy. ;)
**None needed**

### Checklist

Please confirm with `x`:

### Notes
As Eve documents are embedded in markdown, this package is intrinsically dependent on polymode. I originally patterned all of the polymode-related names (the three `defcustom`s) in the same fashion that polymode-markdown did, but this displeased package-lint. I attempted to address package-lint's feedback entirely, but with the most minimal changes possible. If I should do more or less, please let me know.

[1]: http://witheve.com/
[2]: https://github.com/vspinu/polymode

### DIFF
--- a/recipes/eve-mode
+++ b/recipes/eve-mode
@@ -1,0 +1,1 @@
+(eve-mode :repo "witheve/emacs-eve-mode" :fetcher github)


purcell:
Thanks. Quick feedback:

- There's no need to `(provide 'eve-block-mode)`. This is a new package, so who would be trying to write `(require 'eve-block-mode)`?
- Defining everything inside a `let` block is strange and while it (apparently) works thanks to the magic of `lexical-binding`, it would be greatly preferable for you to define variables and commands at the top level, not least so that they can be re-evaluated independently when you are developing this package.
- The big `let` block also prevents you from declaring `autoload` cookies for your major mode forms and the `auto-mode-alist` modification form, both of which would be helpful to users.

Hope that helps.

joshuafcole:
Thanks for the review!

> who would be trying to write (require 'eve-block-mode)

The thinking was that, since eve blocks can be edited outside of the context of a document, users would choose the require for the editing mode(s) they needed. That way, if the package was ever split in the future (so that people not interested in editing full Eve documents didn't end up with a dependency on `polymode`), users `.emacs` files would continue being correct. This is a very trivial detail and I'm happy to remove it anyway if it's considered poor style.

> Defining everything inside a let block is strange
> The big let block also prevents you from declaring autoload cookies

Sorry about that, I've spent much more time in clojure than elisp. I'll unwrap the package.

I'll notify you when I've addressed your concerns.

joshuafcole:
@purcell I believe I've addressed the concerns you listed. Please let me know if you have any further suggestions. https://github.com/witheve/emacs-eve-mode/commit/2224fb05d25be1209fb51f3c4f0d867410f56802

purcell:
Thanks, sorry for the slow response. I sent you a few little fixes in https://github.com/witheve/emacs-eve-mode/pull/1, but I'll go ahead and merge this PR in the meantime.

PR #4836, opened 2017-07-05
Tomatohead recipe added

leabravo:
Tomatohead is a simple Pomodoro timer which uses the header-line for visual feedback on the current progress of the countdown.

https://github.com/leabravo/tomatohead

I am the author.

None needed

Please confirm with `x`:

### DIFF
--- a/recipes/tomatohead
+++ b/recipes/tomatohead
@@ -1,0 +1,1 @@
+(tomatohead :repo "leabravo/tomatohead" :fetcher github)


purcell:
Thanks. Quick feedback:

- Various typos, e.g. "percentaje"
- Don't hard-code face colour attributes: instead, use `define-face` to make a custom face, and `:inherit` a built-in face by default, e.g. `success`.
- `frames` isn't a great parent group for the `tomatohead` custom group.

Hope that helps!

leabravo:
@purcell thank you very much for your input! I am going through the docs and at the same time fixing some code inconsitencies. Please let me know how to proceed when I get my repo updated. Once again many many thanks!

tarsius:
@leabravo Could you please take care of these minor issues now?

> Please let me know how to proceed when I get my repo updated.

Just tell us when things are ready to go :wink:

leabravo:
Hello @tarsius I had made the changed mentioned by the time I wrote the preceding comment (but forgot to comment, my bad). Please let me know if there's anything else I need to update and thanks for taking the time to take a look at the PR/code <3

tarsius:
> `frames` isn't a great parent group for the `tomatohead` custom group.

Nor is `faces`. I would suggest `applications` ("Applications written in Emacs").

> [use] `:inherit` [to derive your faces from] a built-in face by default, e.g. `success`.

You are not doing that but, while not optimal, I think that is okay. Might however want to look at the faces in `basic-faces` and if there are faces that look as appropriate and whose stated purpose does not conflict with the meaning of your own face, then use that. Doing so has the benefit that your faces will look as expected even for users who use a theme that is very different from the one you are using.

tarsius:
Please also add the standard permission statement to the library header:

```
;; This file is free software; you can redistribute it and/or modify
;; it under the terms of the GNU General Public License as published by
;; the Free Software Foundation; either version 3, or (at your option)
;; any later version.

;; This file is distributed in the hope that it will be useful,
;; but WITHOUT ANY WARRANTY; without even the implied warranty of
;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
;; GNU General Public License for more details.

;; For a full copy of the GNU General Public License
;; see <http://www.gnu.org/licenses/>.
```

Not that this sais "version 3, or later". Make sure that's what you want.

leabravo:
Noted, will update and get back to you ASAP. Thanks!

tarsius:
Ahem. Ping :wink:

leabravo:
Sorry about the delay. Updates pushed. Thanks @tarsius <3

Silex:
What is the difference between this package and https://github.com/baudtack/pomodoro.el ?

purcell:
Thanks!

- You've marked this as a non-global minor mode, but it clearly can't be toggled independently in separate buffers without them interfering. It should be a global minor mode.
- The code changes `header-line-format` in every buffer by looping through `buffer-list`, so you'll try to affect the minibuffer etc. This doesn't make much sense. Instead, provide a `tomatohead-local-mode` which sets/unsets `header-line-format` based on the global values of the various vars: your global minor mode then can be a globalised version of this mode - see `define-globalized-minor-mode`.

tarsius:
Slightly grumpy ping!

PR #4841, opened 2017-07-07
Add poly-ruby

knu:

poly-ruby introduces polymode for here-documents in a ruby script.

https://github.com/knu/poly-ruby.el

**None needed**

### Checklist

Please confirm with `x`:

I basically followed the naming conventions used in polymode, but I could change the names if it should be a must to satisfy package-lint.

### DIFF
--- a/recipes/poly-ruby
+++ b/recipes/poly-ruby
@@ -1,0 +1,3 @@
+(poly-ruby
+ :fetcher github
+ :repo "knu/poly-ruby.el")


purcell:
Thanks, this looks useful, and I've done similar things in the past using `mmm-mode` to colourise the content of Ruby heredocs.

What are polymode's restrictions on var/function naming? Ideally almost everything here would start with `poly-ruby-`, and I see a lot of `pm-ruby` and `/` dividers.

knu:
I don't think there's any restriction on naming, it just looks like [existing polymode modules](https://github.com/vspinu/polymode/tree/master/modes) follow its own naming conventions.

purcell:
Ah, great. Then please use the `poly-ruby-` prefix throughout. :-)

knu:
@purcell I've changed internal variables and functions to all have the `poly-ruby--` prefix.  For customs, I kept the names so they align with other polymode customs in the same groups.

purcell:
Makes sense. Thanks - merged!

PR #4842, opened 2017-07-07
Create herald-the-mode-line

marcowahl:

Herald the mode-line as message.

https://github.com/marcowahl/herald-the-mode-line

maintainer - contributor - enthusiastic user

**None needed** 

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/herald-the-mode-line
+++ b/recipes/herald-the-mode-line
@@ -1,0 +1,1 @@
+(herald-the-mode-line :fetcher github :repo "marcowahl/herald-the-mode-line")


purcell:
Thanks. Quick notes:
- `s/occurances/occurrences/`
- There's no need to escape `%`, because you can just say `(message "%s" (format-mode-line (or mode-line-format hide-mode-line)))`.

In fact, that then becomes the code for almost the entire library, in which case I really don't think this is worth adding as a package. :-)

marcowahl:
Thanks.  TIL "occurrence".

I think I'll apply your suggestion.  Thanks.

And agreed.  This little thing might not be worthy enough to be a package.

You shall decide, please.

purcell:
Yep, let's not add this. Thanks for taking the time to submit it though!

marcowahl:
Ok.

Thanks for the effort you invested in herald-the-mode-line.

PR #4850, opened 2017-07-14
Split calfw to calfw, calfw-howm, calfw-ical and calfw-org

tumashu:
This commit split calfw to four packages, which make deal with dependence easier.


purcell:
Super, thanks.

PR #4854, opened 2017-07-16
add inverse-acme-theme

dcjohnson:

Adds an emacs theme that looks like the inverse of the colors for Rob Pike's Acme editor. 

https://github.com/dcjohnson/inverse-acme-theme

Forked this from the code for the gruvbox-theme just for the color declarations. 
Currently maintain this package.

**none needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/inverse-acme-theme
+++ b/recipes/inverse-acme-theme
@@ -1,0 +1,1 @@
+(inverse-acme-theme :fetcher github :repo "dcjohnson/inverse-acme-theme")


purcell:
Thanks. You'll need to add a dependency on the `cl-lib` package. Look good otherwise!

BTW, [this](https://github.com/dcjohnson/inverse-acme-theme/blob/master/inverse-acme-theme.el#L475) is discouraged: minor modes are a matter of user preference, and therefore don't belong in local variables. If you like `rainbow-mode`, you can use a package like [auto-minor-mode](https://melpa.org/#/auto-minor-mode) to enable it in themes, or add a [snippet like this](https://github.com/purcell/emacs.d/blob/master/lisp/init-lisp.el#L248-L261) to your config.

purcell:
(Re. `cl-lib`: yes, this is also a dependency of `autothemer`, but it's best to not rely on transitive dependencies that might change over time.)

dcjohnson:
Thanks for the comments! 
I have updated the code accordingly.

purcell:
Great, thanks - merged!

purcell:
And welcome to MELPA! :-)

PR #4856, opened 2017-07-16
Correct syntax of .gitignore

raxod502:
According to the manual page for gitignore, using a `**` pattern without a prefix or suffix slash is "invalid". In particular, ripgrep will complain about it.

The new version has the same effect as was presumably intended, namely
ignoring all files anywhere in the repository that end in `.elc`.


purcell:
Thanks.

PR #4857, opened 2017-07-17
Add gitpatch recipe

tumashu:

```
;; Gitpatch is git-format patch toolkit, which let user easy handle
;; git-format patch without exit emacs.

;; 1. Send patch with `gitpatch-mail'

;;    `gitpatch-mail' can quick send a git-format patch file from magit,
;;    dired or ibuffer buffer.
```

https://github.com/tumashu/gitpatch

I'm the maintainer

**None needed** 

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/gitpatch
+++ b/recipes/gitpatch
@@ -1,0 +1,1 @@
+(gitpatch :fetcher github :repo "tumashu/gitpatch")
--- a/recipes/gnus-select-account
+++ b/recipes/gnus-select-account
@@ -1,0 +1,1 @@
+(gnus-select-account :fetcher github :repo "tumashu/gnus-select-account")


tumashu:
gnus-select-account is another package, which let user select an account before write a email.

purcell:
Thanks. In future you should put recipes in a separate pull request so they can be reviewed independently.

Regarding gitpatch, that mostly looks good, but "C-c i" is reserved for end-user customisation, so you should choose a different keybinding.

Regarding gnus-select-account, `defgroup` is missing.

tumashu:
```
Thanks. In future you should put recipes in a separate pull request so they can be reviewed independently.
```
Sorry, maybe next time I should use a seperate branch to pull request.

```
Regarding gitpatch, that mostly looks good, but "C-c i" is reserved for end-user customisation, so you should choose a different keybinding.

Regarding gnus-select-account, defgroup is missing.
```
1. "C-c i"  -> "C-x i"
2. gnus-select-account added

purcell:
`C-x i` is a built-in binding (`insert-file`), so I wouldn't recommend using that either.

tumashu:
```
C-x i is a built-in binding (insert-file), so I wouldn't recommend using that either.
```
I will find another key,

by the way: In my opinion,  if I write a mail to send patch file,  insert patch file is just  insert file :-),

tumashu:
I just leave it to user:-) and set nil by default

purcell:
Makes sense. Merged, thanks.

PR #4858, opened 2017-07-18
Term sets

swflint:

This builds upon my other buffer-sets package, providing functionality to automatically gather org-mode based notes files for a given term (year/{summer,fall,spring,etc}).

https://git.flintfam.org/swf-projects/buffer-sets

Maintainer.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/buffer-sets
+++ b/recipes/buffer-sets
@@ -1,1 +1,1 @@
-(buffer-sets :url "https://git.flintfam.org/swf-projects/buffer-sets.git" :fetcher git)
+(buffer-sets :url "https://git.flintfam.org/swf-projects/buffer-sets.git" :fetcher git :files ("buffer-sets.el"))
--- a/recipes/term-sets
+++ b/recipes/term-sets
@@ -1,0 +1,1 @@
+(term-sets :url "https://git.flintfam.org/swf-projects/buffer-sets.git" :fetcher git :files ("term-sets.el"))


purcell:
Thanks - I've applied the update to the `buffer-sets` recipe. I'm not sure we should add `term-sets`, though: not only is the name quite confusing, because "term" more typically means "terminal" for Emacs users, but I'm not sure how generally useful the provided functionality would be for other users. For example, year + term name is not universally applicable outside the US. So I feel like this functionality may simply belong in your personal config.

swflint:
If that is the case, then there was no reason to apply the update for `buffer-sets`.  And yes, the name might be a bit confusing, but it can be changed.

As it stands, there are plenty of packages that don't make sense outside of someone's config (angry-police captain, the nyan-cat whatever,  common-lisp-snippets, dad-joke, turing-machine, number-lock to name a few), and yet they are still in Melpa.

purcell:
> If that is the case, then there was no reason to apply the update for buffer-sets.

That update is necessary in order to exclude `term-sets` from the `buffer-sets` package. It's greatly preferred to place separate packages in separate repos for this reason.

> As it stands, there are plenty of packages that don't make sense outside of someone's config (angry-police captain, the nyan-cat whatever, common-lisp-snippets, dad-joke, turing-machine, number-lock to name a few), and yet they are still in Melpa.

I don't agree that any of those other packages are comparable: they are certainly frivolous, and have small audiences, but are also not heavily geared towards one person's situation and workflow in the same way that many packages I see are.

Term-sets is more in the latter category, which we usually do not include in MELPA, and absent even any `Commentary` section or a clear package description, it's not obvious to me how anyone else would find this package and realise not only that it might be helpful to them, but also how to use it. The same applies to `buffer-sets`, really, though the README goes some way towards mitigating that (the "why?" of it is still quite opaque).

The great thing about Emacs is we get to program it for our own workflows, but not all of those programs are worth distributing. The ones that are should be tidied up, documented and explained, and adjusted to take into consideration how others' workflows may differ from our own.

So I'm still open to including this, but here's what it would take:

- No hard-coded paths (there's still one in `term-sets-make-new-term-folder`).
- Better (non-empty) `Commentary` and clearer package description. Explain the workflow. Make the package discoverable and welcoming to the people it could help!
- Give it its own custom group instead of using `editing`.

Hope that helps!

tarsius:
> Term-sets is more in the latter category, which we usually do not include in MELPA

@swflint And additionally we haven't heard from you in a while. So I am closing this now.

PR #4865, opened 2017-07-20
csound-mode fix: remove `:defaults` to fix build error

hlolli:
https://github.com/melpa/melpa/pull/4797

Continuteing on what was discussed here, there's a build error popping up that is easily solved by removing the keyword `:defaults`.  In the end it will follow the pattern of cider https://github.com/melpa/melpa/blob/master/recipes/cider which does not use `:defaults` and works fine.


milkypostman:
Remove the `*.el` and it's better. `:defaults` includes this already and this is the reason for the error. we should deduplicate that's a different problem.

hlolli:
done

milkypostman:
Should build on the next cycle.

PR #4868, opened 2017-07-20
Add docker-compose-mode recipe

meqif:

Major mode for editing docker-compose files, providing completion of keys.

https://github.com/meqif/docker-compose-mode

Author

**None needed** if there is no problem.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/docker-compose-mode
+++ b/recipes/docker-compose-mode
@@ -1,0 +1,1 @@
+(docker-compose-mode :repo "meqif/docker-compose-mode" :fetcher github)


mattiasb:
Nice! I think this could be made even better by making it a `completion-at-point-functions` backend instead. That way you'd get support for the completion system shipped with vanilla Emacs, other 3rd-party frontends like autocomplete- and ido-at-point in addition to Company (which has a capf-backend).

mattiasb:
Take a look [here](https://github.com/mattiasb/mtg-deck-mode/blob/master/mtg-deck-mode.el#L138-L145) and [here](https://github.com/mattiasb/mtg-deck-mode/blob/master/mtg-deck-mode.el#L186-L187) for an example of how to set it up. :)

purcell:
> Nice! I think this could be made even better by making it a completion-at-point-functions backend instead. That way you'd get support for the completion system shipped with vanilla Emacs, other 3rd-party frontends like autocomplete- and ido-at-point in addition to Company (which has a capf-backend).

Yep, agreed. This would be even better IMO if it were not `company`-specific. (And I say that as a user and fan of `company`!)

meqif:
Thank you for your kind feedback. I took your suggestions to heart and replaced the `company-mode` backend with a capf one, and ended up creating a major mode as well.

mattiasb:
@meqif that's great! :)

purcell:
That rocks! Great outcome, and nice work. Merging now...

meqif:
Awesome, thank you! 🎉

tarsius:
@meqif, you have archived the repositories of `docker-compose-mode` and `flymake-diagnostic-at-point`. Have you moved them elsewhere and Melpa should get them from there? (If so, you should add a note to that effect here on Github.)

Or should these packages be removed from Melpa?

meqif:
Hi @tarsius 👋🏼 It's best to remove them, if you don't mind.

tarsius:
Done. Thanks for the quick response!

PR #4869, opened 2017-07-22
Rename osx-pseudo-daemon to mac-pseudo-daemon

DarwinAwardWinner:

This adds a new recipe mac-pseudo-daemon which is equivalent to the
old osx-pseudo-daemon. The new osx-pseudo-daemon package is just a
stub with obsolete alias definitions for backward comparibility.

(After a suitable period for everyone to switch over, I will rename
the repo and delete osx-pseudo-daemon.el entirely, and submit another 
update to the recipe.)

https://github.com/DarwinAwardWinner/osx-pseudo-daemon

I am the author.

None needed.

### Checklist

Please confirm with `x`:

Notes: `package-lint` complains aboud my decision to use an abbreviated prefix of `macpd-` for package functions rather than the full but quite unwieldly prefix of `mac-pseudo-daemon-`.

### DIFF
--- a/recipes/mac-pseudo-daemon
+++ b/recipes/mac-pseudo-daemon
@@ -1,0 +1,2 @@
+(mac-pseudo-daemon :repo "DarwinAwardWinner/osx-pseudo-daemon" :fetcher github
+                   :files ("mac-pseudo-daemon.el"))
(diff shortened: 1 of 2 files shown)


purcell:
Thanks, looks good. Couple of small points:

- You'll need to declare a package dependency on `cl-lib`.
- Please don't auto-enable the package via the mode's `:init-value` combined with `autoload`: as for the many libraries that ship with Emacs, it should be possible to have this installed without choosing to use it.

DarwinAwardWinner:
Ok, fixed in v2.1.

purcell:
Thanks, merged!

PR #4875, opened 2017-07-24
Added outline-toc recipe.

abingham:

Adds a table of contents sidebar for outline-mode compatible buffers.

https://github.com/abingham/outline-toc.el

Maintainer

None needed

### Checklist

### DIFF
--- a/recipes/outline-toc
+++ b/recipes/outline-toc
@@ -1,0 +1,2 @@
+(outline-toc :fetcher github
+             :repo "abingham/outline-toc.el")


purcell:
Thanks. Quick feedback:
- Don't ever hard-code a font family or size. Just make `outline-toc-font-face` an empty definition, then users can customise or theme it as desired.
- "for Emacs" in docstrings is redundant.
- Don't define your own `outline-toc-current-section` face colours: instead, `:inherit highlight`.
- You've got `defcustoms` with `--` in their name, but the `--` denotes private vars/functions, so this is contradictory. If they're really for end-user tweaking, use `-` instead of `--`.
- Re. `;; TODO: How do we specify "for all outline-mode" docs? Outline-mode is minor, I think...` -- there's `outline-mode-hook`, and.. the `outline-mode` variable.
- Minor mode lighters should have a leading space, so `"OToc"` => `" OToc"`

Hope that helps!

abingham:
Thanks for the feedback. I've worked them into the latest version. The code is still a bit of a mess, but I plan to clean it up as I work with the package a bit more.

alphapapa:
@abingham You might be able to build a lot of your code on top of `navi-mode`.  I think it already handles some things like that.  Although it's not structured as a library, it might still be a useful building block.

abingham:
@alphapapa Actually it looks like `imenu-list` might do almost everything I
need already. With the exception of synchronizing with navigation in the
master document, it provides all of the features I want in outline-toc. But
when I find some time to work on outline-toc some more, I'll definitely
look at navi-mode again.

PR #4877, opened 2017-07-24
Add alarm recipe.

kevindragon:

set or cancel an alarm

https://github.com/kevindragon/alerm.el

maintainer

None needed

### Checklist

Please confirm with `x`:


purcell:
Thanks for making your first MELPA submission!

This package has a few issues overall:

- GPL license statement is missing
- After setting a new alarm, you can't cancel a previous pending one
- You should use `interactive` to prompt for the time in `alarm-clock`, rather than calling `read-string` yourself.
- Hard-coding the playing of a sound isn't a good choice for many users: would be better to use something like the `alert` library to flexibly handle alerts.

Overall, the library has a very general name, and is currently tiny and not very widely useful, so I'd prefer not to accept it in its current form, sorry.

kevindragon:
Thanks for reply. 
This is my first contribute open source project. Although it is rejected, but I am happy as well, it will encourage me to learn more about open source and Emacs and Elisp, I will improve this package.

Thanks again.

PR #4880, opened 2017-07-25
Add windwow recipe

vijumathew:

Simple workspace management. Save and load buffer and window arrangements.

https://github.com/vijumathew/windwow

maintainer

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/windwow
+++ b/recipes/windwow
@@ -1,0 +1,1 @@
+(windwow :repo "vijumathew/windwow" :fetcher github)


purcell:
Thanks. How does this compare to (and differ from) the many existing packages that do similar things? I'm thinking of `desktop.el`, `elscreen`, `persp-mode` etc. Also, I'm surprised that the code itself doesn't seem to be using any of the built-in Emacs machinery for handling window configurations -- working with lists of split commands seems odd.

purcell:
(P.S. I love the name, and thanks for taking the time to make a first MELPA submission!)

vijumathew:
hey @purcell - thanks for taking a look at this and for the feedback! After seeing your comment I realized I used the Emacs term "window-configuration" when I didn't mean that so I changed that to "window-arrangement" to reduce any confusion that might have caused.

This is similar to `elscreen` and `persp-mode.el` in that it helps users to save and load settings about their workspace, but it differs in terms of what exactly it saves and how users can restore that information.

It saves data to let users restore window arrangements (how windows are split and placed in a frame) rather than saving the entire window configuration. Lists of split commands were used because they seemed to be a simple way to translate a window arrangement to data that can be used to restore that window arrangement. 

This package lets users restore buffer lists and and window arrangements independently of each other. An example use case of this is if you are working on several files with a particular window arrangement and you want to switch to a different arrangement easily without losing your buffers. This can be accomplished by calling `windwow-load-window-arrangement`.

And I am biased but I think this package is simpler to use than the other ones because it exposes a few functions for the user to call and they can be used independently of each other.

let me know what you think and if this answers your questions!

purcell:
Cool - thanks for the explanation. Welcome to MELPA!

PR #4884, opened 2017-07-27
Add sidebar recipe

sebastiencs:

sidebar is a customizable file explorer with git integration

https://github.com/sebastiencs/sidebar.el

Author and maintainer

None needed

### Checklist

### DIFF
--- a/recipes/sidebar
+++ b/recipes/sidebar
@@ -1,0 +1,1 @@
+(sidebar :repo "sebastiencs/sidebar.el" :fetcher github)


purcell:
Thanks for taking the time to make your first submission to MELPA! This package looks helpful, and I like that you've described in your README how its goals differ from those of existing similar packages.

Quick feedback:

- `(when (null cond)...)` is equivalent to `(unless cond ...)`
- Use `special-mode` as the parent mode for your major mode(s?), which is more conventional, and gives you automatic handling of the `q` binding, for example.
- I'm not sure the README comment about having two groups of faces is still accurate.
- Don't hard-code colours for new faces: instead, `:inherit` the many built-in faces, e.g. the various `dired` faces. This way, your faces will be automatically themed, but can still be customised.
- Beware unconventional bindings: `C-n` is normally "next line", for example, and no Emacs user would expect it to create a new directory. Consider copying dired bindings.
- `(remove-hook 'post-command-hook 'global-hl-line-highlight)` is going to break things
- I'd suggest that using `-let` when `let` is sufficient is bad style. Particularly when both are mixed in a single function.
- `?` suffixes aren't idiomatic in elisp, so e.g. `sidebar-exists?` would more idiomatically be named `sidebar-exists-p`.
- `defsubst` doesn't generally have any noticeable advantage over `defun`.
- Why are `sidebar--dir?` and `sidebar--opened?` macros? This doesn't seem necessary.

Hope that helps!

sebastiencs:
Thank you @purcell for taking the time to review my submission. It's really usefull to get this kind of feedback.  

> (remove-hook 'post-command-hook 'global-hl-line-highlight) is going to break things

What is your suggestion ? I don't need to highlight the current line because I already color it another way, and I make `post-command-hook` buffer-local, so what could be broken ?

For the others points, I will work on it and come back here to comment this PR.

Thank you

purcell:
> and I make post-command-hook buffer-local, so what could be broken ?

I missed that, sorry. It would be more idiomatic to pass the `LOCAL` flag to `add-hook`/`remove-hook` for this purpose.

tarsius:
@sebastiencs Are all issues resolved?

tarsius:
> Are all issues resolved?

There are no new commits, so no. I am closing for now. Feel free to open a new pr once these issues have been taken care of.

PR #4886, opened 2017-07-28
Add org-category-capture and org-projectile-helm recipes

IvanMalison:
I know that you generally encourage submitters to include only 1 recipe per PR, but since these are all sort of intertwined, and it's better if the building gets separated all in one go, I'd like to change them all together if that's okay. I'll actually need to submit another request after this goes through to exclude org-category-capture.el from the org-projectile.el build, but I think I need to include it in org-projectile.el for now, since I can't change the requirements to include org-category-capture until this goes out.

Repository todo management for org-mode

https://github.com/IvanMalison/org-projectile

Maintainer/Author

### Checklist

Please confirm with `x`:

Proof: https://gist.github.com/445a433bc747525b32a6929308d73246

### DIFF
--- a/recipes/org-category-capture
+++ b/recipes/org-category-capture
@@ -1,0 +1,3 @@
+(org-category-capture :repo "IvanMalison/org-projectile"
+                      :fetcher github
+                      :files ("org-category-capture*.el"))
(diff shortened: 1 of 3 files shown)


purcell:
Sure. Just send another PR with a further update to the `org-projectile` recipe once the dependency has been added to that package.

PR #4887, opened 2017-07-28
Add elvish-mode recipe

ALSchwalm:

`elvish-mode` provides a major mode for the language used in the Elvish shell: https://elvish.io/
This currently includes syntax highlighting and indentation.

https://github.com/ALSchwalm/elvish-mode

Author

None Needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/elvish-mode
+++ b/recipes/elvish-mode
@@ -1,0 +1,1 @@
+(elvish-mode :repo "ALSchwalm/elvish-mode" :fetcher github)


purcell:
Thanks! Quick feedback:

- `define-derived-mode` will declare `elvish-mode-map` for you, so you can omit the empty declaration of the latter.
- Use `prog-mode` as the parent mode.
- Parent custom group should be `languages`, not `applications`.
- You should add an autoload cookie for the major mode definition itself.

ALSchwalm:
Thanks for the feedback. Those should be fixed now.

purcell:
Awesome, thanks - merged!

PR #4894, opened 2017-07-29
New "lms" recipe

inigoserna:

Squeezebox / Logitech Media Server frontend; i.e., an squeezebox controller.

https://bitbucket.com/inigoserna/lms.el

I am the author.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/lms
+++ b/recipes/lms
@@ -1,0 +1,1 @@
+(lms :fetcher bitbucket :repo "inigoserna/lms.el")


purcell:
Thanks! Looks really complete.

- `seq` and `subr-x` are new libraries, so you'll need to declare a package dependency on the appropriate minimum Emacs version.
- `condition-case` with `error nil` can just be written `ignore-errors`
- Don't put `\n` at the end of your `error` messages
- All that `(setq track (plist-put track` code could just be `(setq track (list 'index  (- tot i 1) 'id val ...`, no?
- `C-r`, `C-w` and `C-p` are common built-in key bindings, and their use here will not be very natural.

inigoserna:
> seq and subr-x are new libraries, so you'll need to declare a package dependency on the appropriate minimum Emacs version.

I've written in the docs that it requires emacs 25+, but don't now how to declare it in the package. Any link, please?

> condition-case with error nil can just be written ignore-errors

Thanks! Didn't know about it.

> All that (setq track (plist-put track code could just be (setq track (list 'index (- tot i 1) 'id val ..., no?

Wow! It took me a couple of minutes inside IELM to understand what you mean... everything in LISP is a list!

>     Don't put \n at the end of your error messages

Sorry, a typo.

> C-r, C-w and C-p are common built-in key bindings, and their use here will not be very natural.

Buffers where these key bindings are used are only informative and read-only, they only display a few lines of text and user will no be able to change the contents, so no use for the default actions of these key bindings.

Please note I'm far to be an expert on elisp. Thanks for your comments and ideas!

purcell:
Awesome, merged! :-)

inigoserna:
Thanks.

PR #4896, opened 2017-07-29
Add snoopy-mode package

anmonteiro:

Snoopy Mode is a minor mode for number row unshifted character insertion in Emacs.

https://github.com/anmonteiro/snoopy-mode

Maintainer

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/snoopy
+++ b/recipes/snoopy
@@ -1,0 +1,2 @@
+(snoopy :fetcher github
+        :repo "anmonteiro/snoopy-mode")


purcell:
Thanks for taking the time to make a first MELPA submission!

- `:files` is unnecessary and should be removed here.
- You've called the package `snoopy-mode` while the file is `snoopy.el` -- this mismatch is invalid, so you should rename the recipe to `snoopy`.
- `(bound-and-true-p snoopy-mode)` should just be `snoopy-mode`, because that var will always be bound once the file is loaded.

Looks great otherwise!

anmonteiro:
@purcell thanks for the review, fixed the recipe issues.

purcell:
Great, thanks.

PR #4899, opened 2017-07-30
Add run-stuff package

ideasman42:

Context sensitive execute text under the cursor.

https://github.com/ideasman42/emacs-run-stuff

Maintainer

### Checklist

### DIFF
--- a/recipes/run-stuff
+++ b/recipes/run-stuff
@@ -1,0 +1,3 @@
+(run-stuff
+ :repo "ideasman42/emacs-run-stuff"
+ :fetcher github)


alphapapa:
This looks interesting.  How does it compare with running shell commands from Org source blocks with org-babel?

ideasman42:
@alphapapa I've never used Org source blocks but I've seen others use, so will attempt to answer.

It's more light-weight, If you paste in a path or a file, you can open can it without declaring any surrounding language or metadata.

Commands can be short or multi-line if you want, also the intent is to run things outside of emacs (open in a terminal, file manager, associated application).

Also it doesn't depend on org mode.

alphapapa:
Cool, thanks.

purcell:
Thanks. Quick feedback:

- "a simple package" is a frankly terrible package description. :-)
- There are lots of assumptions about how to execute things, and they are very specific to your individual platform and workflow. There are the `compile`, `term` and `eshell` facilities in Emacs for running commands, and they'd be impossible to integrate with the current code.
- There's already the configurable `browse-url-at-point` command for opening files/URLs at point, so unconditionally running the package's open command isn't a good choice for URLs.
- When would it make sense (or be safe) to split URLs and filenames over multiple lines with trailing `\`? It only seems relevant to executable commands.

I feel like this package tries to overgeneralise, and that it might be better to just make a `run-command-at-point` command, since the other cases are already well (and configurably) covered by `ffap`.

ideasman42:
-  _"a simple package"_ - agree, was left in by accident, corrected.
- Currently the assumption is that you either want to use an external terminal or run directly. If people want to use other kinds of execution - support for them would need to be added.
- Agree, the line splitting is mainly useful for long command lines and not for URL's. On the other hand, since the intention of this package is to perform context sensitive execution - once a string is found that  happens to be a URL - I'm not sure there is any point int _not_ running it in a browser. Note that I've committed a change to use `browse-url` instead of `call-process ...`.

There seems to be some conflict of priorities, I want to have a single keybinding to do something useful with the text under the cursor. Whereas existing tools do one thing _(which is good in general, just not what I'm after)_.

The thing is, I'm not sure you can do this easily by mixing individual commands `run-command-at-point` / `browse-url-at-point` / `ffap` ... etc. Since there needs to be some initial parsing of the text under the cursor to figure out what to do with it.

tarsius:
> The thing is, I'm not sure you can do this easily by mixing individual commands `run-command-at-point` / `browse-url-at-point` / `ffap` ... etc. Since there needs to be some initial parsing of the text under the cursor to figure out what to do with it.

Seems reasonable. @purcell should we merge?

purcell:
> Seems reasonable. @purcell should we merge?

Sure. Thanks for your patience, @ideasman42 .

PR #4904, opened 2017-07-31
add yasnippet-snippets configuration

AndreaCrotti:

"Official" collection of snippet for yasnippet.

https://github.com/AndreaCrotti/yasnippet-snippets

Maintainer

None needed

### Checklist

Please confirm with `x`:


AndreaCrotti:
Uhm yes that's the same error I get locally as well
```
File already exists: /tmp/yasnippet-snippets8412a75/yasnippet-snippets-20170430.1413/.yas-setup.el
```
any idea why it would complain about a file already existing on a clean build?

AndreaCrotti:
Ah wait I think I know, given the way I declared the `files` it's possible that they are being copied all in the same directory (even if coming from different directories).

purcell:
> Ah wait I think I know, given the way I declared the files it's possible that they are being copied all in the same directory (even if coming from different directories).

Yes, that would be it.

AndreaCrotti:
Yes ok thanks @purcell given that my project has something around 100 directories each containing files with no extension (plus some .el) files, what's the suggested way to get them all packaged up?

I guess I have to list all the directories, something like
`"abc-mode/*" xyz-mode/*` which in theory is fine if it's one off, but does it mean I have to change my recipe every single time I create snippets for a new mode?
Thanks

purcell:
All files listed at the top level in `:files` are copied into the root of the built package, so that's not what you want. There's a way to specify a `(source dest)` pair, but we're not really set up for what you're doing.

And anyway, this isn't a proper package without a top-level `yasnippet-snippets.el` file that contains metadata.

So here's an idea: move all of those subdirs into `snippets` (which is then trivial to include in `:files`), then make a `yasnippet-snippets.el` which adds `snippets` to the yas snippets path. This is the approach taken by the various other `*-snippets` packages in MELPA, and it works nicely. Check out [clojure-snippets](https://github.com/mpenet/clojure-snippets) for a good example.

AndreaCrotti:
Yes ok sounds good @purcell, thanks

By the way as an unrelated not, I noticed by looking around at other projects included in Melpa that many of them don't actually have a GPL license.
Is that actually not a strong requirement then and maybe just a suggestion?

purcell:
> By the way as an unrelated not, I noticed by looking around at other projects included in Melpa that many of them don't actually have a GPL license.
> Is that actually not a strong requirement then and maybe just a suggestion?

It's a strong requirement, and we're already in the process of reviewing the licenses of existing packages. That involves contacting the authors, where possible, but any non-compliant packages will likely be removed soon if the licenses remain incompatible with GPLv3.

alphapapa:
@purcell Does that mean that GPLv2 packages will be removed?

AndreaCrotti:
Oh yeah it's finally building now, I'm still using this other branch https://github.com/AndreaCrotti/yasnippet-snippets/pull/203
but I think it's all fine, I'll merge on master and remove the `:branch` from this branch.
What do you think @purcell ?

Thanks for your help by the way!

purcell:
That looks great. Want to remove `:branch` from the recipe, then I'll go ahead and merge?

Also, you can delete `yasnippet-snippets-pkg.el`, because we can get all the metadata we need from `yasnippet-snippets.el`, so there's no need to duplicate it.

AndreaCrotti:
All done @purcell thanks for your help and let me know if I need to something else.

purcell:
Great, thanks.

PR #4905, opened 2017-08-01
add outrespace

articuluxe:

Outrespace is a collection of utilities to manage c++ namespaces.

https://github.com/articuluxe/outrespace

I am the maintainer.

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/outrespace
+++ b/recipes/outrespace
@@ -1,0 +1,1 @@
+(outrespace :repo "articuluxe/outrespace" :fetcher github)


alphapapa:
Hi, I gotta say, I love the intro in the readme!  Very entertaining and clever.

Few things I've noticed:

* You need to declare your dependencies in the package requirements, like `ivy`.
* Not a problem, but after declaring your `defgroup`, you can leave out the `:group 'outrespace` from all the subsequent `def`s.
* Also not a problem, but `(let ((lst '()))) can be simply `(let (lst))`, as the empty list is the same as `nil`.  (You can also call it `list` since elisp is a lisp-2, but it's up to you.)
* Might want to remove this old code: https://github.com/articuluxe/outrespace/blob/master/outrespace.el#L156  :)
* Looking at outrespace--namespace-regexp, let me take a moment to evangelize the `rx` macro.  Many people aren't aware of it, and it makes writing and maintaining complex regexps much easier.  Up to you, of course.  :)
* I see you call a function to define keys in your map when you activate the mode.  Instead, you can define the map with `defvar` at the top of the file.

I can't offer any feedback related to the C++-related code, but overall this looks very well made, very consistent with Emacs conventions.  Great job.

articuluxe:
Hello,

Your feedback was very helpful. I have pushed fixes for the various issues you raised below. 

Thank you very much for taking the time to review the code!

purcell:
Thanks! A little more feedback from me:
- `"C-c n"` is specified in two separate places in the file, but it's also reserved for end-users, so you must not bind it. See the [keybinding conventions](https://www.gnu.org/software/emacs/manual/html_node/elisp/Key-Binding-Conventions.html) for more on this. Best to just tell the user in your docs how to bind their own preferred prefix key to your prefix map.
- Instead of `(make-variable-buffer-local (defvar ...`, use `defvar-local`. That's new as of Emacs 24.3. (So not *very* new...)
- Don't add `:bold t` to `:inherit highlight`. The purpose of inheriting is to get a well-themed appearance by default, and the user's theme might completely avoid bold.

Hope that helps.

articuluxe:
Great, thanks for the advice. 

I've pushed changes to address these comments. 

Thanks again

purcell:
Thanks, looks good. One further point is that you really shouldn't depend on and use `ivy` directly from this package: just use `completing-read`, and then the user's preferred completion system (`ido`, `ivy`, `helm`) will override it as necessary.

articuluxe:
Thanks for the advice.  I spent some time trying to undo this dependency, but ran into a problem.

Basically it seems as if emacs’s completing-read cannot handle multiple identical keys.  Please correct me if I’m wrong in this.

In c++, it is possible to define a namespace in multiple places, which means when jumping to a namespace, there may be multiple entries whose name is identical, differing only in their buffer location.
If there is no provision to handle identical keys, then jumping to a namespace by name is random.

Ivy, however, allows me to store each specific namespace in the cdr of each element of an alist, where the car is the namespace name.  This allows non-ambiguous navigation.

I hope I’ve explained this adequately.

alphapapa:
Ivy and Helm do make it easier by supporting both a "display" and a "real" value.  To do this with plain completing-read, you need to map the display value to the real value yourself.  It seems that you're right about it not accepting multiple identical keys, so I guess you'll have to make the keys unique by appending something to the identifiers.

articuluxe:
OK, I've made the ivy dependency non-mandatory. If ivy is installed, the current and preferable behavior is available. Otherwise the code falls back to standard completing-read, appending unique suffixes onto the namespace identifiers. 

Thank you.

purcell:
Great, merging now...

purcell:
Great, merging now... welcome to MELPA!

articuluxe:
Thank you!

PR #4907, opened 2017-08-01
Rename repos for some recipes

DarwinAwardWinner:
The ido-ubiquitous and osx-pseudo-daemon Github repos have been
renamed to ido-completing-readplus and mac-pseudo-daemon,
respectively.

All packages from these repos build cleanly with the new recipes on my system.


purcell:
Thanks

PR #4908, opened 2017-08-01
Update docker-compose-mode's recipe to exclude test and helper files

meqif:
This change will prevent the file `docker-compose-mode-helpers.el` and tests from being included in the `docker-compose-mode` package.

Major mode for editing docker-compose files

https://github.com/meqif/docker-compose-mode

Author

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/docker-compose-mode
+++ b/recipes/docker-compose-mode
@@ -1,4 +1,4 @@
 (docker-compose-mode
  :repo "meqif/docker-compose-mode"
  :fetcher github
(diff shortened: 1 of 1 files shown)


purcell:
Thanks, applied, with a small change: there was no need to exclude `tests/*.el`, because they wouldn't have been included by the default file spec anyway.

meqif:
I see, thank you!

PR #4909, opened 2017-08-02
Add himp recipe

mkcms:

Hides imports/comments/documentation at beginning of buffer automatically.
Uses `vimish-fold` to hide those regions.
Currently, python and java is supported, but it can be easily extended.

https://github.com/mkcms/himp

Author

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/himp
+++ b/recipes/himp
@@ -1,0 +1,2 @@
+(himp :fetcher github
+      :repo "mkcms/himp")


purcell:
Thanks. Quick comments:
- Why do you need `declare-function` if you're `require`-ing the corresponding file anyway?
- `himp-matchers` could be a `defcustom`. There are ways to annotate the type structure you're using there.
- The `himp-mode-toggle` command is unnecessary. The `himp-mode` command itself is already an interactive toggle.

mkcms:
Thanks, I commited with fixes/refactorings. I was using `declare-function` because I was getting compiler warnings but now they disappeared.

purcell:
Great, thanks! Merged... welcome to MELPA!

PR #4910, opened 2017-08-02
Add cubicle-mode for the model checker Cubicle

mebsout:

This package provides a major mode for editing files for the [model checker Cubicle](https://github.com/cubicle-model-checker/cubicle).

https://github.com/cubicle-model-checker/cubicle/blob/master/emacs/cubicle-mode.el

Maintainer

None needed

### Checklist

### DIFF
--- a/recipes/cubicle-mode
+++ b/recipes/cubicle-mode
@@ -1,0 +1,4 @@
+(cubicle-mode
+ :fetcher github
+ :repo "cubicle-model-checker/cubicle"
+ :files ("doc/emacs/cubicle-mode.el"))
--- a/recipes/cubicle-mode
+++ b/recipes/cubicle-mode
@@ -1,4 +1,4 @@
 (cubicle-mode
  :fetcher github
  :repo "cubicle-model-checker/cubicle"
- :files ("doc/emacs/cubicle-mode.el"))
+ :files ("emacs/cubicle-mode.el"))


purcell:
Thanks.
- You'll need to be specific about the license. You should include the standard GPL 3 header in the elisp file, as inserted by `M-x auto-insert` if you try that in an empty `.el` file.
- `"\\.cub$"` in the file regexp should be `"\\.cub\\'"`, to match end-of-string rather than end-of-line.
- It's not appropriate to override `compile-command` in a major mode. Users can set this in a file-local or directory-local variable if they wish. And, of course, `(buffer-file-name)` will sometimes be nil, and can change through the course of the buffer's lifetime, so you can't correctly set it in a major mode anyway.
- The box around the commentary text should be removed: it will confuse the core Emacs functions that are used to extract that text for the purposes of providing a package description (and also for the MELPA package page).

tarsius:
@mebsout Friendly ping!

mebsout:
Sorry for the delay, the previous comments have been addressed in https://github.com/cubicle-model-checker/cubicle/commit/8dbb250

Cheers

tarsius:
It doesn't seem appropriate to place an Emacs package below a folder intended for documentation.  You might even want to consider placing it in a completely separate repository. That doesn't really matter as far as Melpa is concerned but it helps adding the package to the [Emacsmirror](https://emacsmirror.net).

tarsius:
Thanks!

PR #4912, opened 2017-08-03
added recipe for catmacs

pymaximus:

Provides CAT client for Yaesu FT991A Transceiver

https://bitbucket.org/pymaximus/catmacs

Author/Developer/Maintainer

**None needed** 

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/catmacs
+++ b/recipes/catmacs
@@ -1,3 +1,3 @@
-(catmacs :url "https://pymaximus@bitbucket.org/pymaximus/catmacs.git"
-         :fetcher git
-         :files ("catmacs.el"))
+(catmacs :url "https://bitbucket.org/pymaximus/catmacs"
+         :fetcher git )
+         
--- a/recipes/catmacs
+++ b/recipes/catmacs
@@ -1,0 +1,3 @@
+(catmacs :url "https://pymaximus@bitbucket.org/pymaximus/catmacs.git"
+         :fetcher git
+         :files ("catmacs.el"))


alphapapa:
Hey there,

I can't speak to the package's intended purpose, but it looks neat!  Here's a little feedback:

* Not to discourage you, but you say that it's a prototype, so I'm not sure it belongs on MELPA yet.  Not that there are any hard rules here, though.
* You probably shouldn't bind `C-c m` in the global keymap when the package is loaded.  Better to define a global minor mode that activates and deactivates the keybindings.
* I don't know anything about the CAT protocol, so: is this package specific to that transceiver, or can it work with anything that uses CAT?  If it's specific to that transceiver, maybe it should be renamed appropriately.  `catmacs` brings to mind `cat for Emacs`, which could be feline in nature, or be related to `cat`, maybe something like `eless`.  :)

Thanks.

pymaximus:
Hi there,
Thanks for the feedback.

- I no longer consider this a prototype (although like most projects, it will improve over time). Updated the README 
  To reflect this.
- I removed the global key bindings, good point.
- I still like the sound of catmacs, and would be happy to evolve this over time to handle other compatible rigs.

Cheers / Frank

purcell:
Thanks. Quick feedback:
- The git URL in the recipe contains `pymaximums@`, and it shouldn't. That's not the proper public git URL for your project.
- `:files` spec is unnecessary and should be removed from the recipe.
- `C-c m` is [reserved for end-user customisation](https://www.gnu.org/software/emacs/manual/html_node/elisp/Key-Binding-Conventions.html) and should not be used. Instead, choose an allowed alternative or (better) provide a prefix map, and tell the user how to bind it to their prefix key of choice.

pymaximus:
Thanks for the feedback.

Made updates on the recipe. That should be reflected in the pull request now.

I will look at the prefix map idea (have not used that before). 

Cheers / Frank

pymaximus:
Hi,

Following up on the keymap and prefix idea (which is rather new to me)..

>> 	• C-c m is reserved for end-user customisation and should not be used. Instead, choose an allowed alternative or (better) provide a prefix map, and tell the user how to bind it to their prefix key of choice.

I defined a prefix map as follows ..

(define-minor-mode catmacs-mode
  "CAT client for Yaesu FT991A Transceiver"
  :lighter " catmacs"
  :keymap (let ((catmacs-keymap (make-sparse-keymap)))
            ;; prefix command
            (define-prefix-command 'catmacs-keymap). <======***
            ;; keys
            (define-key catmacs-keymap (kbd "f") 'catmacs-fa-set)
            (define-key catmacs-keymap (kbd "m") 'catmacs-md-set)
              <snip>


I can invoke the following from say scratch buffer to set a prefix key

(global-set-key (kbd "C-c m") 'catmacs-keymap)

And it works fine. I can invoke functions via C-c m prefix (e.g.: C-c m f for setting frequency)

I have a couple of questions.

Q1. Am I doing the prefix command part correctly when catmacs-mode is invoked, and
Q2. What is best practice for getting end user to issue their version of the  global-set-key invocation I show above and where
      Is the best place to document it. For completeness I could place a comment in the code for users who look at the
     Lisp, but for other end users, do I just say execute this or ???


Thanks / Frank

pymaximus:
> On 8 Aug 2017, at 2:04 pm, Frank Singleton <b17flyboy@gmail.com> wrote:
> 
> Hi,
> 
> Following up on the keymap and prefix idea (which is rather new to me)..
> 
>>> 	• C-c m is reserved for end-user customisation and should not be used. Instead, choose an allowed alternative or (better) provide a prefix map, and tell the user how to bind it to their prefix key of choice.
> 
> I defined a prefix map as follows ..
> 

Or is this approach preferred ??

(defvar catmacs-prefix "C-c m”) <=====

(define-minor-mode catmacs-mode
  "CAT client for Yaesu FT991A Transceiver"
  :lighter " catmacs"
  :keymap (let ((catmacs-keymap (make-sparse-keymap)))

            ;; keys
            (define-key catmacs-keymap (kbd (concat catmacs-prefix "f")) 'catmacs-fa-set)
            (define-key catmacs-keymap (kbd (concat catmacs-prefix "m")) 'catmacs-md-set)
            (define-key catmacs-keymap (kbd (concat catmacs-prefix "q")) 'catmacs-qr-set)

Guidance on this appreciated :-)


> (define-minor-mode catmacs-mode
>  "CAT client for Yaesu FT991A Transceiver"
>  :lighter " catmacs"
>  :keymap (let ((catmacs-keymap (make-sparse-keymap)))
>            ;; prefix command
>            (define-prefix-command 'catmacs-keymap). <======***
>            ;; keys
>            (define-key catmacs-keymap (kbd "f") 'catmacs-fa-set)
>            (define-key catmacs-keymap (kbd "m") 'catmacs-md-set)
>              <snip>
> 
> 
> I can invoke the following from say scratch buffer to set a prefix key
> 
> (global-set-key (kbd "C-c m") 'catmacs-keymap)
> 
> And it works fine. I can invoke functions via C-c m prefix (e.g.: C-c m f for setting frequency)
> 
> I have a couple of questions.
> 
> Q1. Am I doing the prefix command part correctly when catmacs-mode is invoked, and
> Q2. What is best practice for getting end user to issue their version of the  global-set-key invocation I show above and where
>      Is the best place to document it. For completeness I could place a comment in the code for users who look at the
>     Lisp, but for other end users, do I just say execute this or ???
> 
> 
> Thanks / Frank

purcell:
The former approach would be preferable (ie. with `define-prefix-command`).

pymaximus:
OK, just committed that change to catmacs repo.

purcell:
Great, merged now. Welcome to MELPA!

pymaximus:
> On 27 Aug 2017, at 4:05 pm, Steve Purcell <notifications@github.com> wrote:
> 
> Great, merged now. Welcome to MELPA!
> 
> 
Awesome, thanks !!

PR #4914, opened 2017-08-05
org-wc repo on github transferred to original author.

tesujimath:
Simple change of package origin.


purcell:
Thanks Simon. I see we're both software verterans in NZ with UK backgrounds. :-)

tesujimath:
Thanks Steve.  Yep, enjoying NZ now.

PR #4918, opened 2017-08-06
Update bbyac repository

baohaojun:
Package some .bbyac files with bbyac, to make it support programming languages completion.


purcell:
Thanks. Please use `:defaults` instead of `*.el`, then we're good to go.

baohaojun:
Sorry.  Thanks, I have updated accordingly.

purcell:
Great, thanks!

PR #4920, opened 2017-08-07
Add dotnet.el

julienXX:

Interact with dotnet CLI tool for .NET Core.

https://github.com/julienXX/dotnet.el

maintainer

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/dotnet
+++ b/recipes/dotnet
@@ -1,0 +1,1 @@
+(dotnet :repo "julienXX/dotnet.el" :fetcher github)


purcell:
Thanks! Quick feedback:
- You'll need to declare a package dependency on `cl-lib`. `package-lint` should have told you about this when you ran it.
- You're doing a lot of `read-string` in your commands: instead, add an arg to the commands and use `(interactive "sPackage name: ")` to have it magically read for you.
- Don't explicitly use `ido-completing-read`: just call `completing-read`. This is then overridden by various completion systems as preferred by the individual user.
- `C-c C-c` is a very common command binding. Instead, perhaps provide a prefix map that the user can bind to their preferred key.
Hope that helps!

julienXX:
> - You'll need to declare a package dependency on cl-lib. package-lint should have told you about this when you ran it.

Removed the dependency as I don't use it yet.

> - You're doing a lot of read-string in your commands: instead, add an arg to the commands and use (interactive "sPackage name: ") to have it magically read for you.

Thanks! That's a lot better.

> - Don't explicitly use ido-completing-read: just call completing-read. This is then overridden by various completion systems as preferred by the individual user.

Cool, I even have swiper completion now :)

> - C-c C-c is a very common command binding. Instead, perhaps provide a prefix map that the user can bind to their preferred key.

Good catch, it's been updated.

Thanks @purcell for the great suggestions!

purcell:
Thanks. Sent you some fixes in https://github.com/julienXX/dotnet.el/pull/1. Once merged, this PR is good to go... :-)

julienXX:
@purcell merged and thanks again for the lesson in emacs lisp :)

purcell:
No worries. Make sure you test it, though, in case I broke something. :-)

PR #4921, opened 2017-08-07
Comment tags recipe

vincekd:

Provides a minor mode for highlighting and jumping between comment tags like TODO, FIXME, etc.

The current packages are fairly outdated and don't allow easily jumping between tags/viewing them in all in one spot.

https://github.com/vincekd/comment-tags

None

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/comment-tags
+++ b/recipes/comment-tags
@@ -1,0 +1,3 @@
+(comment-tags
+ :fetcher github
+ :repo "vincekd/comment-tags")
--- a/recipes/comment-tags
+++ b/recipes/comment-tags
@@ -1,4 +1,3 @@
 (comment-tags
  :fetcher github
- :repo "vincekd/comment-tags"
- :files ("comment-tags.el"))
+ :repo "vincekd/comment-tags")
--- a/recipes/comment-tags
+++ b/recipes/comment-tags
@@ -1,3 +1,4 @@
 (comment-tags
  :fetcher github
- :repo "vincekd/comment-tags")
+ :repo "vincekd/comment-tags"
+ :files ("comment-tags.el"))


alphapapa:
This looks interesting.  Couple of thoughts:

* How does this compare with https://github.com/tarsius/hl-todo?  It's maintained by @tarsius and is actively developed.  Maybe you could collaborate.
* I think `C-c c` might not be a good prefix for your keymap because that's used by `org-capture`.

vincekd:
I think the biggest difference is just the ability to jump between tags with ido-completion and keywords, rather than just doing next, next. Also you can view all the tags in you all open buffers with the mode enabled, which I find invaluable when jumping between files.

According to the hl-todo description, it seems he's not very interested in maintaining things he doesn't use like the jumping between tags. (probably should have asked though).

alphapapa:
Cool.  You might want to put some screenshots up to help people understand how it works.

tarsius:
> Looking at the hl-todo description, it seems he's not very interested in maintaining things he doesn't use like the jumping between tags.

Yes, I am not particularly interested in adding any new features at this time, even if someone else implements them. I might add features at a later time but I have no immediate plans for anything.

kaushalmodi:
There is also [fixmee][1] that allows the FIXME/TODO/etc. navigation. 

May be collaborate with that package? If the author of that package is inactive, you can work with Melpa devs, I believe, to take over that package's maintainership. 

[1]: https://github.com/rolandwalker/fixmee/blob/master/README.markdown

alphapapa:
Maybe there could be a standard variable listing the keywords that both packages could use.  Otherwise keeping them in sync could be annoying.  :)

purcell:
Thanks. I have lots of little comments here, mostly about idiomatic elisp:

- Never call `ido-completing-read` directly. Just use `completing-read`, then the user's preferred completion framework can step in. `helm` users don't want you to decide they should use `ido`! :-)
- Don't hard-code foreground colours without also defining background colours. In fact, instead of `comment-tags-keyword-colors`, try `comment-tags-keyword-faces`. There are built-in faces that you can use like `success` and `error`.
- Similarly, any default face you use can `:inherit` a built-in face. So instead of `comment-tags-foreground-color` and `comment-tags-background-color`, consider `defface comment-tags-face`, with `:inherit highlight`. This way, everything looks good by default in any theme, but can still be customised.
- `C-c c` is reserved for end users: `package-lint` should have complained about that, if you've got a recent version.
- `(font-lock-mode 1)` should be unnecessary, and in theory you should not enable `font-lock-mode` if the user had disabled it. Instead, use `font-lock-flush` or similar.

(`fixmee` is a bit clunky really, so I'm not sure I'd necessarily suggest joining forces, though I hate to see parallel implementations of similar ideas because it's wasted effort, and also confusing for potential users.)

vincekd:
What might be a good alternative to `C-c c`? Maybe `C-c t`?

kaushalmodi:
Normally packages don't do the bindings. You can suggest the binding on the documentation, but leave it to the user to bind themselves. 

/Any `C-c` followed by upper/lower-case alphabet binding should not be bound by any package./

vincekd:
Alright I updated it to `C-c #` (which I think is reserved for minor modes) and addressed the other issues raised. Let me know if anything else needs to be fixed.

purcell:
Looks good. I'm going to edit the recipe before merging because the `:files` spec there is redundant - the defaults do the trick nicely in this case. Welcome to MELPA!

PR #4922, opened 2017-08-07
Add recipe for pelican-mode.

joewreschnig:

pelican-mode is an Emacs minor mode for editing pages and posts in
Pelican sites. Pelican is a static site generator which can process a
variety of text file formats. For more information, see URL
https://blog.getpelican.com/.

It's intended to be used alongside a major mode for the Pelican
document. Currently supported formats are Markdown, reStructuredText,
AsciiDoc, and Org.

https://git.korewanetadesu.com/pelican-mode.git

Maintainer/author/user.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/pelican-mode
+++ b/recipes/pelican-mode
@@ -1,0 +1,2 @@
+(pelican-mode :fetcher git
+              :url "https://git.korewanetadesu.com/pelican-mode.git")


purcell:
Thanks! Quick comments:
- `C-c <letter>` is [reserved for end-user customisation](https://www.gnu.org/software/emacs/manual/html_node/elisp/Key-Binding-Conventions.html), so you should choose different bindings, or simply provide a prefix map that the user can bind to a key of their choice.
- Rather than a global minor mode and `pelican-mode-enable-if-site`, it would be more conventional to provide a minor mode and a globalised minor mode: the latter would use `pelican-mode-enable-if-site` to enable the minor mode only in appropriate buffers.

joewreschnig:
> C-c <letter> is reserved for end-user customisation, so you should choose different bindings, or simply provide a prefix map that the user can bind to a key of their choice.

I've changed it to a prefix map. (I've not done this before, and based it on how Projectile does it.)

> Rather than a global minor mode and `pelican-mode-enable-if-site`, it would be more conventional to provide a minor mode and a globalised minor mode: the latter would use `pelican-mode-enable-if-site` to enable the minor mode only in appropriate buffers.

Unless I've misunderstood, this is exactly how it's implemented. Can you let me know which docs led you astray?

purcell:
> Unless I've misunderstood, this is exactly how it's implemented. Can you let me know which docs led you astray?

Ah, it was the `text-mode-hook` part which confused me. You could alternatively use `define-globalized-minor-mode`, in which case the "enable-if" function would check for `(derived-mode-p 'text-mode)`. It'd be slightly terser that way.

In any case, merging now... :-)

joewreschnig:
Thanks for the feedback and quick merge!

I've also filed some issues with `package-lint` based on it (which I hope to tackle in a couple weeks).

purcell:
Awesome, thanks Joe!

PR #4924, opened 2017-08-08
Specify cfml-mode.el file in cfml-mode recipe

am2605:

CFML (ColdFusion) syntax highlighting mode.

https://github.com/am2605/cfml-mode

Maintainer

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/cfml-mode
+++ b/recipes/cfml-mode
@@ -1,5 +1,5 @@
 (cfml-mode
  :repo "am2605/cfml-mode"
  :fetcher github
- :files (:defaults "mmm-mode")
+ :files ("cfml-mode.el")
  )


am2605:
My initial commit seemed to have some local dependencies (sorry).  I have updated the package to correct this.

purcell:
Thanks. Building and installing the package locally is a good way to find issues like this. I think it's on the submission checklist for that reason. Merging this now. :-)

purcell:
Actually, this can't yet be merged, because `cfml-mode` doesn't specify a package dependency on `cftag-mode`. And `cftag-mode.el` has a dependency on `mmm-mode`, which I think is bogus and should be removed. I'd suggest that I merge this and #4923 once those issues are fixed.

am2605:
Thanks.  Sorry about this.

I had done make recipes/cftag-mode and make recipes/cfml-mode as well as installed them into my local instance using C-c C-c but it didn't throw up any errors.

Is there something else I need to be doing in order to catch these before I submit so that I don't waste anyone's time?

________________________________

purcell:
Yeah, I guess that it wouldn't have been trivial to find the problem in this case: you'd have had to install the cut-down `cfml-mode`, restart Emacs so that `cftag-mode` wasn't still loaded, then you'd have found that `cfml-mode` didn't work.

am2605:
I am new to this, and I am happy to learn.  I'm just sorry I didn't get it first go(es).

I'll test it a few ways via a clean install before I resubmit.

________________________________

purcell:
Not to worry. No testing process is foolproof, and I'm absolutely happy to give feedback on submissions: they don't have to be perfect ahead of time. :-)

am2605:
Thanks Steve.

I've updated https://github.com/am2605/cfml-mode with the changes to the
Package-Requires, which I think now resolve this.  I don't believe I need
to resubmit any pull requests to MELPA as recipes/cfml-mode and
recipes/cftag-mode didn't need to change?

A question for the future - if I make further change in
https://github.com/am2605/cfml-mode are these automatically built into
MELPA or do I need to resubmit something again?

purcell:
Yep, that's great. Both PRs are now merged.

Re. future changes, we'll automatically rebuild new packages within a few hours of the code changing. :-)

alphapapa:
For future reference, this is what the MELPA sandbox is for, it makes testing installation on clean profiles very easy.  e.g.:

```
$ cd ~/src/melpa && rm -rf sandbox && make sandbox INSTALL=cfml-mode
```
Run that and all you have to do is wait for it to automatically install the package and all of its dependencies.  If there are no compiler errors, then you are probably good to go, although you might want to do some testing, depending on what your package does and what kind of changes you've made.

am2605:
Thank you!

________________________________

PR #4926, opened 2017-08-09
Recipe: Add highlight-function-calls

alphapapa:
This package highlights function symbols in function calls.  This makes them stand out from other symbols, which makes it easy to see where calls to other functions are.  Optionally, macros and special forms can be highlighted as well.  Also, a list of symbols can be excluded from highlighting; by default, ones like `+` / `-`, `<` / `>`, `error`, `require`, etc. are excluded.  Finally, the `not` function can be highlighted specially.

https://github.com/alphapapa/highlight-function-calls

Author and maintainer.

None needed.

### Checklist

Please confirm with `x`:

Thanks.

### DIFF
--- a/recipes/highlight-function-calls
+++ b/recipes/highlight-function-calls
@@ -1,0 +1,1 @@
+(highlight-function-calls :fetcher github :repo "alphapapa/highlight-function-calls")


purcell:
Thanks. Quick notes:
- Don't set `:foreground red` without also setting a background. But better still, just `:inherit` a built-in face (e.g. `error`) and don't add any further attributes.
- `highlight-function-calls--face` should not use the `--` separator, since it's a face and therefore customisable.
- The special handling of `not` seems out of place, and I think you should consider dropping it. Or (perhaps later) expose a customisable mapping of function symbol names to a face (if any) to apply. Much of your code would then become entries in the default mapping.

alphapapa:
Hi Steve,

> Don't set :foreground red without also setting a background. But better still, just :inherit a built-in face (e.g. error) and don't add any further attributes.

Thanks, I set it to inherit from `warning` for now.

> highlight-function-calls--face should not use the -- separator, since it's a face and therefore customisable.

Thanks, good catch.

> The special handling of not seems out of place, and I think you should consider dropping it. 

I realize it may seem unusual, but I've found it useful in other major modes.  And I've probably spent some time trying to figure out a bug when it was a `not` that I forgot to remove.  I'd like to keep it in here, otherwise I'd have to make a whole 'nother package just for it, and I definitely don't want to do that.  I've disabled it by default; is that acceptable?

> Or (perhaps later) expose a customisable mapping of function symbol names to a face (if any) to apply. Much of your code would then become entries in the default mapping.

I'm not sure what you mean by "default mapping."  But I like the idea of customizable mappings of function names to faces.  Might be interesting to tie it in to something like `nameless-mode` to give special highlighting to functions local to the package.  But I'm not sure if that would belong in a separate package, because it might make sense to also highlight package-local variables, etc.

Thanks for your feedback.

purcell:
> I'm not sure what you mean by "default mapping." But I like the idea of customizable mappings of function names to faces. Might be interesting to tie it in to something like nameless-mode to give special highlighting to functions local to the package. But I'm not sure if that would belong in a separate package, because it might make sense to also highlight package-local variables, etc. I'll think about adding this in the future.

By "default mapping" I mean that if you have a `defcustom` list of functions which each take a symbol name and return a face symbol or nil, then your code would call those in turn for each symbol until it got a face symbol. See `run-hook-with-args-until-success` for this. And then your current code would reduce to a function which returns a specific face for `not`, and another function which returns a face for other non-suppressed function names, so there'd be a "default mapping" containing just those two functions.

There's nothing magical about the `not` case, in that sense.

In any case, I'll merge this now, but I definitely encourage you to have a play with the above idea! :-)

(See also the `cl-lib-highlight` package, btw.)

alphapapa:
Thanks, Steve, now I understand what you mean.  That would be an elegant solution.  And thanks for the pointer to that package.

Fanael:
What's the difference between this package and [highlight-defined](https://github.com/Fanael/highlight-defined)?

alphapapa:
> What's the difference between this package and highlight-defined?

IIUC your package highlights symbols anywhere they occur.  The intention of this package is to only highlight them when they occur as function calls.

PR #4927, opened 2017-08-10
Add copy-file-on-save recipe

zonuexe:

Auto deployment (copy) a file when the file saved.

https://github.com/zonuexe/emacs-auto-deployment

I'm maintainer.

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/auto-deployment
+++ b/recipes/auto-deployment
@@ -1,0 +1,1 @@
+(auto-deployment :fetcher github :repo "zonuexe/emacs-auto-deployment")
--- a/recipes/auto-deployment
+++ b/recipes/auto-deployment
@@ -1,1 +1,0 @@
-(auto-deployment :fetcher github :repo "zonuexe/emacs-auto-deployment")
--- a/recipes/copy-file-on-save
+++ b/recipes/copy-file-on-save
@@ -1,0 +1,1 @@
+(copy-file-on-save :fetcher github :repo "zonuexe/emacs-auto-deployment")


recipes/auto-deployment
```diff
@@ -0,0 +1 @@
+(composer :fetcher github :repo "zonuexe/emacs-auto-deployment")
```
tarsius: The first element has to match the name of the package.

alphapapa:
Does this provide any benefit over, e.g. https://github.com/wasamasa/firestarter?  It even has an example of deployment:

```
# -*- firestarter: "rsync -avz -e ssh /src host:/dest" -*-
```

recipes/auto-deployment
zonuexe: @tarsius Thanks. It fixed.

zonuexe:
@alphapapa 
Firestarter is a nice product, but the aim of auto-deployment is different from that.
Auto-deployment focuses on synchronizing saved edited file.  This can possibly be realized with firestarter, but it will be slightly more complicated.

purcell:
Thanks. You'll need to either declare a package dependency on `cl-lib`, or on a newer Emacs version: `package-lint` should have warned you about that...?

It's not super clear from the `.el` file and its docstrings what the package does, and why a user might need it. The deployment method looks configurable, but only `copy-file` really does anything useful: setting the deployment method to a function would be pretty much the same as adding a custom function directly to `after-save-hook`.

My feeling is that you should simplify this slightly, and instead call it `copy-file-on-save`, which would also make its purpose clearer.

zonuexe:
@purcell 
I applied simplify and renamed to `copy-file-on-save`.  Please review this again.

purcell:
Looks great, thanks!

zonuexe:
Thanks!

PR #4928, opened 2017-08-10
Add company-terraform recipe

rafalcieslak:

This package provides a `company` backend for Terraform files, enabling auto-completion in Terraform sources.

https://github.com/rafalcieslak/emacs-company-terraform

I am the maintainer.

**None needed**

### Checklist

Please confirm with `x`:

I have also tested dependencies with `make sandbox`.

### DIFF
--- a/recipes/company-terraform
+++ b/recipes/company-terraform
@@ -1,0 +1,1 @@
+(company-terraform :fetcher github :repo "rafalcieslak/emacs-company-terraform")


alphapapa:
Looks pretty nice, although I don't use Terraform myself.  :)

Couple of suggestions, but these are a matter of style/preference:

* The nested `dolist` loops could be written as simple `cl-loop`s.
* Some of the long `cond`s that test on `context` could be written more succinctly with `pcase`. 

Of course, not everyone likes `loop` and `pcase`.  :)

rafalcieslak:
Thank you for your feedback! I really wanted to express the logic with pattern-matching - it makes perfect sense for recognizing `context` - but I wasn't aware that there is a macro that lets me do that so conveniently. So I'm really glad you directed me to `pcase`, it helped me clean up the control flow quite a lot. I've also remade nested lists using `cl-loop`, just to see how would they look like, but I'm not quite sure if they got better or worse.

Anyway, an updated / cleaned-up version is available in my repository.

purcell:
Super, thanks.

PR #4929, opened 2017-08-10
Updated org-mind-map

theodorewiles:

Creates directed graphs from org-mode files using graphviz

https://github.com/theodorewiles/org-mind-map

Developer and maintainer

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/org-mind-map
+++ b/recipes/org-mind-map
@@ -1,0 +1,1 @@
+(org-mind-map :repo "theodorewiles/org-mind-map"  :fetcher github :files ("org-mind-map.el"))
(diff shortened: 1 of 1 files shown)


purcell:
Thanks. Please remove the `:files` spec from the recipe, since the default will do what you want here. Further notes:

- Please use the GPLv3 boilerplate rather than v2. Version 3 is the version which covers recent Emacs versions and their extensions.
- You're using `dash`, but haven't declared a `Package-Requires` package dependency, so you should add one.
- You should use `shell-quote-argument` to make filenames and other command-line args safe when building commands with `concat`.
- Please fix the dangling `)` on their own lines. This isn't idiomatic Lisp formatting.
- Add `;;;###autoload` cookie comments for each of your three interactive entry-point commands.

Hope that helps!

theodorewiles:
Thanks for the review! I just implemented these edits.

purcell:
Great, thanks! Welcome to MELPA. :-)

PR #4931, opened 2017-08-11
Add the ‘hasky-stack’ package

mrkkrp:

Interface to the Stack Haskell development tool.

https://github.com/hasky-mode/hasky-stack

Author and maintainer.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/hasky-stack
+++ b/recipes/hasky-stack
@@ -1,0 +1,3 @@
+(hasky-stack
+ :repo "hasky-mode/hasky-stack"
+ :fetcher github)


mrkkrp:
Remaining issues from `package-lint`:

> 8:52: warning: Use a non-snapshot version number for dependency on "magit-popup" if possible.

I think `magit-popup` is versioned in this "snapshot" fashion and it's not possible to specify a non-snapshot version, although I may be wrong.

/cc @tarsius

tarsius:
> I think magit-popup is versioned in this "snapshot" fashion and it's not possible to specify a non-snapshot version, although I may be wrong.

When I create a release I set the version to something like `2.*` and in the very next version commit I set the "version" to something like `2017....`. So you should use non-snapshot version numbers.

mrkkrp:
OK, done.

purcell:
Nice!

PR #4934, opened 2017-08-12
Move nix-mode to separate repository

matthewbauer:
This adds the updated nix-mode.el version to the MELPA repository. Not sure what Nix maintainers think of this, but this nix-mode has some notable improvements in syntax highlighting. It is based on the same code as the original but with some improvements.

Same as older nix-mode, just with better highlighting support. Should make it easier for people to make changes, etc.

Because this moves an old package, I want to make sure everyone is okay with it before merging. These Nix people should probably sign off in some way:

- @edolstra
- @shlevy

If it's preferred to keep it within the NixOS org, maybe it can be made a separate repository? This is already done with NixOS/nix-idea. The main problem currently is that Emacs+Nix PRs sit around for a long time without anyone reviewing them.

Other people that might have input:

- @mdorman
- @ttuegel 
- @peti

See related issues: NixOS/nix#1419, matthewbauer/nix-mode#30

https://github.com/NixOS/nix-mode

maintainer

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/nix-mode
+++ b/recipes/nix-mode
@@ -1,2 +1,2 @@
-(nix-mode :repo "NixOS/nix" :fetcher github
-          :files ("misc/emacs/nix-mode.el"))
+(nix-mode :repo "NixOS/nix-mode" :fetcher github
+          :files ("nix-mode.el"))


purcell:
Hi - this is great, but we wouldn't normally change an existing recipe from an "official" source to an "unofficial" source without the blessing of the maintainers. A better solution is to get any changes/improvements into the official repo. (This wouldn't apply if you instead chose to rename your `nix-mode` to distinguish it, but it would arguably be better for users if there's a single excellent version of it.)

shlevy:
@matthewbauer I'd recommend an email to the nix devel mailing list asking to move your repo to the NixOS github org and making a case for your maintainership.

matthewbauer:
Ok I've updated it to reflect the new [NixOS/nix-mode](https://github.com/nixos/nix-mode) repo.

purcell:
Thanks!

PR #4936, opened 2017-08-12
Add rimero-theme recipe

yveszoundi:

Emacs theme with light colours on a dark background, for UI and terminal usage.

https://github.com/yveszoundi/emacs-rimero-theme

Creator/Maintainer

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/rimero-theme
+++ b/recipes/rimero-theme
@@ -1,0 +1,1 @@
+(rimero-theme :repo "yveszoundi/emacs-rimero-theme" :fetcher github)


purcell:
Thanks, and welcome to MELPA!

tarsius:
@yveszoundi Renaming `rimero-theme.el` to `rimero-dark-theme.el` broke this package on Melpa because every package must feature a library whose name matches the name of the package.

There are three ways forward and because the best two involve replacing the current recipe with one or more new recipes, I am deleting the current recipe.

1. Add a new library `rimero-themes.el` (plural), containing the metadata that we expect in the "main library". Beside `(provide 'rimero-themes)`, it doesn't have to contain any code.
2. Restore a library `rimero-theme.el` (singular). This has the disadvantage that `load-theme` will list that library as a theme (because of its name), regardless of whether it actually is a theme. If it is not and the user selects it, that results in an error. (This is the least favorable option.)
3. Add two separate recipes for the two themes. These recipes must set `:files` to exclude the other theme.

PR #4937, opened 2017-08-12
Add: shrink-path.el recipe

bandresen:

Provides functions that offer fish shell path truncation.
Directory /usr/share/emacs/site-lisp => /u/s/e/site-lisp

Also includes utility functions that make integration in eshell or the
modeline easier.

https://gitlab.com/bennya/shrink-path.el

Author

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/shrink-path
+++ b/recipes/shrink-path
@@ -1,0 +1,1 @@
+(shrink-path :fetcher gitlab :repo "bennya/shrink-path.el")


purcell:
Thanks! Quick comments:
- The autoload cookies should be removed: libraries or user config code requiring these functions should just `require` your library.
- I'm sure users would appreciate some examples of useful ways to use those functions, e.g. in `eshell`.

bandresen:
Thanks for the feedback.

There are two screenshots in the README. Both of those will be available in the DOOM Emacs configuration so when this is done (https://github.com/hlissner/.emacs.d/pull/176) I will link to the full configuration.

But for now I've included the eshell configuration in the README, as it's short enough that it can just be copy-pasted and only needs

I also removed the autoload cookies.

purcell:
:+1:

PR #4940, opened 2017-08-16
Add quiz.el

davep:

A simple trivia quiz game that uses [The Open Trivia Database](https://opentdb.com/) as the back end.

https://github.com/davep/quiz.el

Maintainer

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/quiz
+++ b/recipes/quiz
@@ -1,0 +1,1 @@
+(quiz :fetcher github :repo "davep/quiz.el")


purcell:
Thanks, this is fun. You should use `define-derived-mode` rather than rolling your own major mode function: that way you can extend `special-mode` (which gives you the `q` behaviour for free) and skip the boiler plate in your hand-rolled version.

davep:
Thanks. I'll take a look at that. I've got a few packages that might
benefit from that.

davep:
Now done.

Amusingly I've actually used this approach in much older code of mine, and seem to have forgotten about it. (including `quickurl` and `5x5` which are part of Emacs itself!)

purcell:
:-)  Merged now, thanks!

PR #4945, opened 2017-08-18
Move emacs-psci package

kRITZCREEK:
Fixes https://github.com/purescript-emacs/emacs-psci/issues/17.

We just moved the emacs-psci repository under the purescript-emacs organisation.


tarsius:
Thanks!

PR #4947, opened 2017-08-19
Restore bpr recipe (GPL-compatible license)

ilya-babanov:


purcell:
Thanks. Pending https://github.com/ilya-babanov/emacs-bpr/pull/9.

tarsius:
Thanks!

I have re-added this package in a batch commit.

PR #4950, opened 2017-08-23
Add recipe for `pulsectl`.

flexibeast:

`pulsectl` controls PulseAudio volumes from Emacs, via `pactl`.

https://github.com/flexibeast/pulsectl

i am the maintainer.

n/a

### Checklist

Please confirm with `x`:


alphapapa:
This looks very interesting!  It would help new users a lot if you provided some screenshots, especially for the sink selection.

flexibeast:
@alphapapa:

Thanks for your comments. :-)

Re. screenshots, i don't know how useful they'd be overall. Say i'm editing some ELisp in a buffer; i press `C-x / m`, and the currently-selected sink is muted. Consequently, there's nothing visual to see. Further, the currently-selected sink is specified by having called a `pulsectl-select-sink-by-*` command, which uses the minibuffer to prompt for a sink; so all a screenshot of that would show is a minibuffer prompt of "Sink name: ", which doesn't show much. (i myself use `ivy`, which shows a list of candidates underneath the prompt, but i can't assume users will be using `ivy`.)

Re. the package name, i'm willing to consider changing it, as long as:

* variable and function names didn't have to change to use the (to me, as maintainer) awkwardly-lonng `pulseaudio-control-` prefix; and

* that's not going to cause issues with getting the recipe added to MELPA. Might it require me to cancel the current pull request, delete the existing repo with the current name, create a new repo with the new name, then create a new pull request for adding the receipt to MELPA?

Re. autoloads, agreed; that was an oversight, sorryne
Re. lexical binding, i don't want to not support Emacs < 24.1 unnecessarily; if i add a file-local `lexical-binding t` setting, will that cause issues on Emacs < 24.1?

Re. using `call-process-region`, i'm not sure i understand why i should use this; what region am i going to be calling it on?

Re. `generate-new-buffer`, thank you for picking that up; i was experimenting with different designs using either `generate-new-buffer` or `with-temp-buffer`, and i meant to settle on `with-temp-buffer`. So i'll certainly change that.

Finally, does all the above need to be addressed before the recipe can be added to MELPA?

alphapapa:
I think a screenshot of showing the sink being selected would be helpful, but it's up to you, it's your package.  :)

> variable and function names didn't have to change to use the (to me, as maintainer) awkwardly-long pulseaudio-control- prefix

That's not really that long compared to some other packages.  You might consider using `nameless-mode` while developing.

> hat's not going to cause issues with getting the recipe added to MELPA. Might it require me to cancel the current pull request, delete the existing repo with the current name, create a new repo with the new name, then create a new pull request for adding the receipt to MELPA?

You could rename the repo without creating a new PR, but either way I don't think that's a problem.

> Re. autoloads, agreed; that was an oversight, sorry.

No need to apologize!  :)

> Re. lexical binding, i don't want to not support Emacs < 24.1 unnecessarily; if i add a file-local lexical-binding t setting, will that cause issues on Emacs < 24.1?

I don't know.  Emacs 24.1 was released 5 years ago, though.  But it's up to you which versions you want to support.  I just mentioned it as a best practice going forward.

> Re. using call-process-region, i'm not sure i understand why i should use this; what region am i going to be calling it on?

The point is to avoid calling a shell process, which in general is a good idea.  call-process-region need not operate on a region, it can simply insert into a buffer.

> Finally, does all the above need to be addressed before the recipe can be added to MELPA?

This is not up to me.  I'm not a staff member, merely a bystander who tries to help out by giving feedback on new PRs before Steve or Syohei get around to it.

flexibeast:
> I think a screenshot of showing the sink being selected would be helpful, but it's up to you, it's your package. :)

Okay, i'll see what i can come up with. :-)

> That's not really that long compared to some other packages. You might consider using nameless-mode while developing.

Hadn't heard of that! Will definitely try to keep it in mind next time.

> You could rename the repo without creating a new PR

Oh okay, i couldn't immediately see how GitHub lets one do that; i'll have to do some more research.

> I don't know. Emacs 24.1 was released 5 years ago, though. But it's up to you which versions you want to support. I just mentioned it as a best practice going forward.

\*nod\* Fair enough, thanks. i find lexical binding generally easier to reason about, myself, but i keep seeing people on various fora talking about how they're using < 24.1 (sometimes << 24.1!) for whatever reason. Again, i'll do some research, and if adding lexical binding doesn't break things for such users, i'll happily enable it.

> The point is to avoid calling a shell process, which in general is a good idea.

Agreed!

> call-process-region need not operate on a region, it can simply insert into a buffer.

Ah okay, i'll try to change over to using `call-process-region` instead, then.

> I'm not a staff member, merely a bystander who tries to help out by giving feedback on new PRs

Well, thank you very much for your input! 'Tis appreciated. :-)

flexibeast:
@alphapapa:

i'm having difficulties getting `call-process-region` to work. If, in `*scratch*`, i do:

    (call-process-region "-l" nil "ls" nil t)

then a directory listing is inserted into the buffer. But if i do:

    (call-process-region "list short sinks" nil "pactl" nil t)

i get "No valid command specified." And if i do:

    (call-process-region "-t get_workspaces" nil "i3-msg" nil t)

i get an empty array returned, even though running `i3-msg -t get_workspaces` from a shell returns a non-empty JSON array.

Any ideas as to what might be happening?

alphapapa:
The argument signature is different than the shell function.  Check the docs.

flexibeast:
Yep, the examples i provided were based on my reading of the `call-process-region` documentation. And yet `call-process-region` is returning the (seemingly) inconsistent results i noted.

alphapapa:
I don't understand.
```
call-process-region is a built-in function in ‘C source code’.

(call-process-region START END PROGRAM &optional DELETE BUFFER DISPLAY
&rest ARGS)

Send text from START to END to a synchronous process running PROGRAM.

START and END are normally buffer positions specifying the part of the
buffer to send to the process.
If START is nil, that means to use the entire buffer contents; END is
ignored.
If START is a string, then send that string to the process
instead of any buffer contents; END is ignored.
The remaining arguments are optional.
Delete the text if fourth arg DELETE is non-nil.

Insert output in BUFFER before point; t means current buffer; nil for
 BUFFER means discard it; 0 means discard and don’t wait; and ‘(:file
 FILE)’, where FILE is a file name string, means that it should be
 written to that file (if the file already exists it is overwritten).
BUFFER can also have the form (REAL-BUFFER STDERR-FILE); in that case,
REAL-BUFFER says what to do with standard output, as above,
while STDERR-FILE says what to do with standard error in the child.
STDERR-FILE may be nil (discard standard error output),
t (mix it with ordinary output), or a file name string.

Sixth arg DISPLAY non-nil means redisplay buffer as output is inserted.
Remaining args are passed to PROGRAM at startup as command args.

If BUFFER is 0, ‘call-process-region’ returns immediately with value nil.
Otherwise it waits for PROGRAM to terminate
and returns a numeric exit status or a signal description string.
If you quit, the process is killed with SIGINT, or SIGKILL if you quit again.
```
So you need to do something like:
```el
(call-process-region 0 0 "pactl"
                     nil '(t t) nil
                     "list short sinks")
```

flexibeast:
Evaluating your suggested command:

    (call-process-region 0 0 "pactl"
                         nil '(t t) nil
                         "list short sinks")

resulted in the error:

    Args out of range: #<buffer *scratch*>, 0, 0

which makes sense according to the documentation quoted:

    START and END are normally buffer positions specifying the part of the
buffer to send to the process.
    If START is nil, that means to use the entire buffer contents; END is
ignored.
    If START is a string, then send that string to the process
instead of any buffer contents; END is ignored.

And "0 0" aren't valid buffer positions, of course.

My previous examples assumed that the documentation stating:

    If START is a string, then send that string to the process
instead of any buffer contents

meant that I would pass command line arguments as a string in the START position. However, I see that near the bottom of the documentation, it says:

    Remaining args are passed to PROGRAM at startup as command args

Since ARGS isn't capitalised, I didn't realise the documentation was referring to the final function argument. So okay, how about this:

    (call-process-region
      "" ; Value for START - pass an empty string to the process' STDIN(?).
         ; We can't use `nil` here because that would mean "pass entire 
         ; buffer contents", and we can't use `0` because we're not going to 
         ; be using buffer contents at all. 
      nil ; Value for END.
      "pactl"; Value for PROGRAM.
      nil ; Value for DELETE - n/a since we have no buffer text to delete.
      t ; Value for BUFFER - insert output into current buffer.
      nil ; Value for DISPLAY - we don't need progressive output display.
      "list short sinks" ; Value for ARGS, "passed to PROGRAM at startup
                                ; as command args.
    )

The output is still:

    No valid command specified.

alphapapa:
`pactl` receives `"list short sinks"` as a single argument.  You need to pass the words as separate strings.

flexibeast:
That's it! i didn't at all consider that "list short sinks" is actually three arguments, not a single argument. Thank you for your time and patience. :-) i might think about suggesting some changes to the `call-process-region` documentation ....

At any rate, i can now move to using `call-process-region` instead of `call-process-shell-command`. Thanks again!

alphapapa:
Yeah, I don't like any of the process-calling functions in Emacs.  You might find this useful:

https://github.com/alphapapa/org-super-agenda/blob/master/test/test.el#L95

flexibeast:
That seems a nicer interface, indeed - i might try to put together something analogous, simplified for this particular use-case. Thanks!

flexibeast:
i've now:

* renamed the package from `pulsectl` to `pulseaudio-control`;
* updated the recipe to reflect the preceding;
* enabled lexical binding; and
* added a screenshot to the README.md.

@alphapapa: Thanks again for your advice and assistance!

purcell:
Sorry for the long delay on this - it's looking great. Just one nit: `(global-set-key (kbd "C-x /") 'pulseaudio-control-map)` is something that should be left for the user, so suggest that snippet in your README and/or provide a `pulseaudio-control-default-keybindings` command. It's quite possible someone would like a different global binding, so you should make it easy for them. :-)

flexibeast:
That's okay! i can imagine you've a lot on your plate, and anyway, it gave me time to incorporate @alphapapa's suggestions. :-)

The call to `global-set-key` is now inside a `pulseaudio-control-default-keybindings` command; i've updated the documentation accordingly. Thanks!

PR #4951, opened 2017-08-23
Add `snazzy-theme` recipe

weijiangan:

A high-contrast base16 theme based on @sindresorhus's Snazzy color scheme.

https://github.com/weijiangan/emacs-snazzy

I am the maintainer.

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/snazzy-theme
+++ b/recipes/snazzy-theme
@@ -1,0 +1,1 @@
+(snazzy-theme :repo "weijiangan/emacs-snazzy" :fetcher github)


purcell:
Hi, thanks for submitting a package to us!

You've checked `I've used package-lint to check for packaging issues, and addressed its feedback` above, but I can see at least two issues it would have pointed out. :-)

Please run package-lint and fix those issues then ping us here once done.

weijiangan:
I would lie to my grade school teacher, but never @purcell :(

screenshot from 2017-08-27 17-58-46

The only problem reported flycheck is the package base16-theme is not found, but I do not know how to get around the error, it seems to have something to do with location of the file? I'm not sure but this package sure works when you run `package-install-file` on it and I'm actually using the package as I'm typing this as seen in the screenshot.

If it's something less obscure i.e. elisp idioms or encouraged ways of writing elisp I ask your favor to point it out so I can look more into it as I'm still raw with elisp and I'm not very familiar with the best practices and stuff (I'm guessing it has something to do with require vs autoload?).

Thanks :)

purcell:
> I would lie to my grade school teacher, but never @purcell :(

LOL :-)

The two issues I spotted were:
- "Emacs" in the description line, which is redundant, and should be removed.
- `;; Package-Requires: ((base16-theme))` should be `;; Package-Requires: ((base16-theme "2.1"))`, since the former version-less format is not supported by `package.el` in older Emacsen.

But here's what `M-x package-lint-current-buffer` shows me locally, which includes a couple of further messages:

screen shot 2017-08-28 at 18 58 55

But the problem might be that the code in your screenshot is not the same as the current code in your github repo. Did you forget to push some changes?

weijiangan:
Oh my, you're right! I forgot to push the changes 😅 Silly me. I'll push the changes and update back

weijiangan:
I pushed the changes :)

purcell:
That's great, because it means I don't need to worry that there's a problem with `package-lint`! :-)

purcell:
Awesome, thanks! Welcome to MELPA. :)

weijiangan:
> That's great, because it means I don't need to worry that there's a problem with `package-lint`! :-)

Haha, sorry for the scare and misunderstanding, and thanks for making Emacs so awesome! 😄

PR #4954, opened 2017-08-26
Remove some packages

tumashu:
The three packages have been renamed to other packages:

chinese-fonts-setup  ->    cnfonts
chinese-pyim         ->    pyim
org-webpage          ->    org2web


purcell:
Thanks, merging now. Do you also want to add `:old-names` entries to the newer recipes? It allows the MELPA webpages to redirect, and also causes the download stats to be carried over.

tumashu:
No need， just a new start :-)

purcell:
Okay, cool.

PR #4955, opened 2017-08-26
Recipe for x509-mode

jobbflykt:

View certificates and other PKI-files using OpenSSL.

https://github.com/jobbflykt/x509-mode

Author. Maintainer.

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/x509-mode
+++ b/recipes/x509-mode
@@ -1,0 +1,3 @@
+(x509-mode :repo "jobbflykt/x509-mode"
+           :fetcher github
+           :files (:defaults "*.txt"))


purcell:
Thanks for submitting this! Looks really handy. Quick feedback:

- Your custom group should be called `x509` rather than `x509-mode`.
- `(require 'cl-seq)` won't work in earlier 24.x Emacsen with `cl-lib` from MELPA, because there's no such feature. You should just require `cl-lib`, IIRC.

jobbflykt:
Thanks for the feedback. I've fixed the `(require 'cl-lib)` and the custom groups are now named `x509` and `x509-faces`. Also realized that a few custom variables had the wrong `:type`.

purcell:
Awesome! Welcome to MELPA.

PR #4957, opened 2017-08-26
Create s12cpuv2-mode

AdamNiederer:

A major mode for the assembly language of the S12CPUV2 found on some Freescale microcontrollers

https://github.com/AdamNiederer/s12cpuv2-mode

Wrote it

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/s12cpuv2-mode
+++ b/recipes/s12cpuv2-mode
@@ -1,0 +1,1 @@
+(s12cpuv2-mode :fetcher github :repo "AdamNiederer/s12cpuv2-mode")


purcell:
Thanks. Quick feedback:
- Don't override `TAB` and `<backtab>`: instead, set your own `indent-line-function`, then all the standard Emacs machinery will work as the user expects, however they have customised it.
- `C-c l` is reserved for end users, and must not be used here. Recent versions of `package-lint` warn about this.
- You're using `setq-local`, which only appeared in Emacs 24.3, so you'll need to depend on that Emacs version at a minimum. `package-lint` should also have told you about this.
Looks good otherwise!

AdamNiederer:
Thanks for the feedback; I updated package-lint and fixed the issues.
Quick question: It doesn't look like `<backtab>` is bound to anything in `prog-mode`. Does binding it to something do anything unexpected? Removing the binding breaks one's ability to unindent lines with `<backtab>` from what I can see.

purcell:
> Quick question: It doesn't look like <backtab> is bound to anything in prog-mode. Does binding it to something do anything unexpected? Removing the binding breaks one's ability to unindent lines with <backtab> from what I can see.

You're right: I think it's reasonable in this case. I found a bunch of other packages which define `backtab` locally.

tarsius:
> You're right: I think it's reasonable in this case.

@purcell So we should merge?

purcell:
Yep. Thanks for your patience, @AdamNiederer.

PR #4958, opened 2017-08-26
Add recipe for dired-sidebar

jojojames:

Displays a tree browser/side directory browser leveraging dired to do most of the work

https://github.com/jojojames/dired-sidebar

Author/Maintainer

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/dired-sidebar
+++ b/recipes/dired-sidebar
@@ -1,0 +1,1 @@
+(dired-sidebar :fetcher github :repo "jojojames/dired-sidebar")


purcell:
Thanks for submitting this. Screenshots look very nice. Some quick concrete feedback on the code:

- Don't ever override font families or sizes: `dired-sidebar-font` should just be an empty declaration which the user can override as desired. (Also, `-font` in the name of a face should be avoided.)
- `buffer-face-mode-face` should be set buffer-locally, not globally.
- `Copied from `treemacs--set-width'.` isn't a helpful docstring for end users. Maybe a comment instead? :-)
- The `ace-window` handling is very special-case and should be removed: users of that package (including yourself!) could simply override the necessary keys in your minor mode keymap.

Hope that helps!

jojojames:
```
Don't ever override font families or sizes: dired-sidebar-font should just be an empty declaration which the user can override as desired. (Also, -font in the name of a face should be avoided.)
```
I was thinking setting a default font was fine since the font would only be set in the sidebar.

There's definitely a 'scratch my own itch' element to setting those values as the default.
I am thinking the user will have different options if they don't want to use those fonts,
either disable the custom font using the dired-sidebar-use-custom-font flag or change the face themselves?

Maybe a good compromise is to set dired-sidebar-use-custom-font to nil to disable the custom font by default?

I'm open to your original suggestion, just wanted to get your opinion on alternatives.

I'm also open to changing the name of that variable. Any suggestions would be great.

Thanks.
```
buffer-face-mode-face should be set buffer-locally, not globally.
```
Thanks for catching this. Will change it.
```
Copied fromtreemacs--set-width'.` isn't a helpful docstring for end users. Maybe a comment instead? :-)
```
Yeah, I wanted to give credit for the implementation. The original didn't have much documentation either though.

```
(defun treemacs--set-width (width)
  "Set the width of the treemacs buffer to WIDTH when it is created."
  (unless (one-window-p)
    (let ((window-size-fixed)
          (w (max width window-min-width)))
      (cond
       ((> (window-width) w)
        (shrink-window-horizontally  (- (window-width) w)))
       ((< (window-width) w)
        (enlarge-window-horizontally (- w (window-width))))))))
```

What do you think?

```
The ace-window handling is very special-case and should be removed: users of that package (including yourself!) could simply override the necessary keys in your minor mode keymap.
```
Hmnn, I wanted decent interop/nice defaults with other packages out of the box. (Similar to Neotree/Treemac's compatibility with other packages.)

What do you think about wrapping that keybind with a defcustom flag as well as checking for the feature before setting the bind?

Users would just be able to disable the flag if they don't want it.

purcell:
> Maybe a good compromise is to set dired-sidebar-use-custom-font to nil to disable the custom font by default?

It should be nil by default anyway IMO, and then the custom face should also be an empty `defface`, with no predefined fonts.

> I'm also open to changing the name of that variable. Any suggestions would be great.

Use `-face` instead of `font`.

> Yeah, I wanted to give credit for the implementation. The original didn't have much documentation either though.

Yup, I got that. But it's not helpful info for a user of the function, so doesn't belong in the docstring.

> What do you think about wrapping that keybind with a defcustom flag as well as checking for the feature before setting the bind?

You're really talking about reimplementing customisable keymaps for one special case! :-)  You can just provide a snippet for interested users which looks something like `(define-key dired-sidebar-mode-map [C-o] 'ace-window-whatever)`.

purcell:
Looking a bit more closely, I was misunderstanding what you were trying to accomplish with the `ace-window` handling, so the keymap suggestion wasn't helpful, I think!

What's the desired difference between, say, `C-o` and `RET`? Would you want both at different times, or just one? There's definitely a neater way of expressing this code which doesn't involve special-casing for `ace-window`, so I'm keen to hear what behaviour you're going for.

jojojames:
> It should be nil by default anyway IMO, and then the custom face should also be an empty defface, with no predefined fonts.

Ok, that's fair. I'll probably move those defaults to the README then.

~ Though it might be interesting to go for a 'remove what you want' approach vs a 'add what you want approach'.
~ Emacs has a lot of 'add what you want' when setting up these types of packages.

It's probably the same amount of work for users that care to remove/add stuff but it's definitely more work for users that want to be spoonfed stuff. :)

> Use -face instead of font.

Sounds good.

> Yup, I got that. But it's not helpful info for a user of the function, so doesn't belong in the docstring.

Sounds good, I'll move it out into a comment then.

> What's the desired difference between, say, C-o and RET? Would you want both at different times, or just one? There's definitely a neater way of expressing this code which doesn't involve special-casing for ace-window, so I'm keen to hear what behaviour you're going for.

RET should always open in the very next window.
C-o should leverage ace-window if it's available to let the user pick and choose the window they want to open the file in.

They are two different use cases for me.

purcell:
> RET should always open in the very next window.
> C-o should leverage ace-window if it's available to let the user pick and choose the window they want to open the file in.
> 
> They are two different use cases for me.

Makes sense. To solve this in an idiomatic Emacs manner, I'd probably make a `defcustom` of `:type function` that specifies the "alternate" select window method. Then if `dired-sidebar-find-file` is called with a prefix arg, it would use that function instead of `next-window`. This way you could either create a separate binding (e.g. C-o) or just use `C-u RET`. To use `ace-window` as the alternate method, you'd set that defcustom to `(lambda () (aw-select "Select window"))`.

jojojames:
Made the changes.

Alexander-Miller:
> Yeah, I wanted to give credit for the implementation. The original didn't have much documentation either though.

Best give credit to neotree then, that is where I looked up the window shrink/enlarge part. The doc is so sparse because the function is internal and fairly simple. The one directed at users looks like this:

```
(defun treemacs-reset-width (&optional arg)
  "Reset the width of the treemacs buffer to `treemacs-buffer-width'.
If a prefix argument ARG is provided read a new value for
`treemacs-buffer-width' first."
  (interactive "P")
  (let ((window-size-fixed nil))
    (when arg
      (setq treemacs-width
            (->> treemacs-width
                 (format "New Width (current = %s): ")
                 (read-number))))
    (treemacs--set-width treemacs-width)))
```

jojojames:
^ Thanks, I'll add a note for neotree too.

purcell:
Thanks for your patience: merged now. Welcome to MELPA!

jojojames:
Thanks @purcell !.

PR #4966, opened 2017-08-29
Add contextual-menubar

aaronjensen:

Displays the menubar if on a graphical display, but hides it if in a terminal.
It works for multiple frames on the same server.

https://github.com/aaronjensen/contextual-menubar

maintainer

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/contextual-menubar
+++ b/recipes/contextual-menubar
@@ -1,0 +1,1 @@
+(contextual-menubar :fetcher github :repo "aaronjensen/contextual-menubar")


purcell:
Thanks, this is handy, but you've got references to `eslint_d` in the description line and README, which have presumably been copied from elsewhere. Please ping me here when those are fixed so I can merge.

Also, rather than use `require` to enable this feature, you should probably instead provide an autoloaded command which executes the `add-hook` form. That way people can add that command to their init file or `after-init-hook` etc., or they can (potentially) `require` the library to use `contextual-menubar-show-or-hide-menubar` without it automatically modifying `after-make-frame-functions`.

aaronjensen:
Thanks @purcell. Is there a typical name for an installation command like you mention? Would something like `contextual-menubar-install-hook` be ok?

aaronjensen:
I've updated the repo, please let me know if that looks ok. Thanks!

yassinphilip:
@aaronjensen The header line "baseline" part still is : "use eslint_d to automatically fix js files" when that is not what the package does, it looks like some commit message..?

aaronjensen:
ah, oops. Thanks. Ok, I've searched for eslint and don't see anymore, but please let me know if you see anything else. Sorry 😖

yassinphilip:
> please let me know if you see anything else. 
Yes ; The filename/baseline header line must be on a single line, and not wrapped over two like it is now. Also, Capitalizing the first word would not hurt.
> Sorry
No PB, thanks for contributing!

aaronjensen:
Ok, fixed up. And yes, it did, but I did not rerun it after adding that function. It passes `checkdoc` and `package-lint` now.

purcell:
Thanks. I'll go ahead and merge this now, but I strongly suggest you rename `contextual-menubar-install-hook` to `contextual-menubar-install`.

PR #4971, opened 2017-08-30
Add euslisp-mode

iory:

euslisp-mode, a major mode for editing euslisp program files.

https://github.com/iory/Euslisp-mode

Maintainer

Reintroducing the package after licensing has been moved to GPLv3+, making it Emacs compatible.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/euslisp-mode
+++ b/recipes/euslisp-mode
@@ -1,0 +1,2 @@
+(euslisp-mode :fetcher github
+              :repo "iory/euslisp-mode")


purcell:
Hi! What's the story with the use of `exec-path-from-shell` combined with the code copied from it? Not only is that sort of functionality out of place in a major mode, I'm also wondering if there's a refactor I could make to `exec-path-from-shell` so that it would fit your use case better.

tarsius:
I have already restored the package recipe in the latest batch. So there's nothing to merge here, but @iory please look into the issue raised by @purcell anyway. If this were a new package, then it most likely would not get added until you have done so. Thanks!

Issue #4970, opened 2017-08-30
vc-fossil recipe won't bake


mdorman:
It doesn't seem to be able to find the source files (please ignore the eyebending paths---I'm on NixOS):
```
emacs -Q --batch -l /home/mdorman/src/melpa/package-build/package-build.el -l /nix/store/q6rjlwi6c1p3iypxxcig17q985fns3jr-emacs2nix-0.1.0.0/share/x86_64-linux-ghc-8.0.2/emacs2nix-0.1.0.0/checkout.el -f checkout /home/mdorman/src/melpa/recipes/vc-fossil vc-fossil /home/mdorman/src/emacs2nix/.workdir/vc-fossil
No matching file(s) found in /home/mdorman/src/emacs2nix/.workdir/vc-fossil/: (*.el *.el.in dir *.info *.texi *.texinfo doc/dir doc/*.info doc/*.texi doc/*.texinfo (:exclude .dir-locals.el test.el tests.el *-test.el *-tests.el))
```

aplaice:
The recipe doesn't work, since all the elisp files are in `vc/el`, not in the root of the repository. (The inability to build hasn't impacted MELPA, since the repo hasn't updated since the recipe change, so the already existing package could be served.)

Adding `:files ("vc/el/*.el" "doc/*")` to the recipe would fix it.

PR #4972, opened 2017-08-31
Fix vc-fossil recipe

aplaice:
The emacs lisp files are at vc/el, not at the root of the repository. 

The doc files are included since they had been, before the move to the 
emacsorphanage.

See issue #4970.

### Checklist

Please confirm with `x`:


tarsius:
I used to periodically import the library by itself but recently switched to importing the actual history. But I forgot to update the recipe. Thanks!

PR #4975, opened 2017-09-01
Add sqlite3-api recipe

pekingduck:

This package does two things:
1. Provides a function `sqlite3-api-install-dynamic-module` to users to install a dynamic module (from same repo below) that wraps around the SQLite3 C API.
2. Defines two error symbols used by the dynamic module.

https://github.com/pekingduck/emacs-sqlite3-api

I am the maintainer.

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/sqlite3-api
+++ b/recipes/sqlite3-api
@@ -1,0 +1,1 @@
+(sqlite3-api :fetcher github :repo "pekingduck/emacs-sqlite3-api")


pekingduck:
Figured out a way to call (defconst ...) in C so this recipe is not needed anymore.

purcell:
Cool, thanks.

PR #4976, opened 2017-09-01
Add esxml-query to esxml

wasamasa:

esxml is a package for taking apart XML, manipulating it and generating XML.  esxml-query is a new addition to it that allows extracting contents in a jQuery-like fashion with CSS selectors.  I'd like to add its file to the recipe so that other packages can make use of this feature. 

https://github.com/tali713/esxml

I'm a contributor with commit access to esxml and wrote esxml-query.  Previously I've only contributed bugfixes.

**None needed**

### Checklist

Please confirm with `x`:

`M-x checkdoc` hangs here.  I'm not sure whether the remaining points are relevant, considering this PR is about adding a file to an existing package, but I'll address them if needed.


purcell:
Thanks!

PR #4979, opened 2017-09-03
Add recipe for mu4e-query-fragments

wavexx:

`mu4e-query-fragments` allows to define query snippets ("fragments") that
can be used in regular `mu4e` searches or bookmars. Fragments can be used to
define complex filters to apply in existing searches, or supplant bookmarks
entirely. Fragments compose properly with regular mu4e/xapian operators, and
can be arbitrarily nested.

https://github.com/wavexx/mu4e-query-fragments.el

Author

### Checklist

Please confirm with `x`:

package-lint complains that I use mu4e/qf as opposed to mu4e-query-fragments as the prefix, however I'm using mu4e/[suffix] to avoid namespace clashes with mu4e symbols, as well to keep it reasonably editable.

The same strategy is used with multiple-cursors (where mc/[ext] is used).

### DIFF
--- a/recipes/mu4e-query-fragments
+++ b/recipes/mu4e-query-fragments
@@ -1,0 +1,2 @@
+(mu4e-query-fragments :repo "wavexx/mu4e-query-fragments.el"
+                      :fetcher github)


purcell:
Thanks! Quick feedback:
- `mu4e` is not a well-formed package dependency: `package-lint` should have warned about this. In fact `mu4e` is not installable as a package, so you should not list it.
- Re. the `mu4e/qf` prefix, just use `mu4e-query-fragments` to match the package name. `mc/` set a bad example, and we don't accept non-standard prefixes these days, since we want to promote the elisp community standards.

wavexx:
On Fri, Sep 08 2017, Steve Purcell wrote:
> - `mu4e` is not a well-formed package dependency: `package-lint`
> should have warned about this. In fact `mu4e` is not installable as a
> package, so you should not list it.

Ok, didn't notice, as I made the mu4e package locally for myself. Sorry
about this.

> Re. the `mu4e/qf` prefix, just use `mu4e-query-fragments` to match the
> package name. `mc/` set a bad example, and we don't accept
> non-standard prefixes these days, since we want to promote the elisp
> community standards.

Is there some "standard" shortcut for long package names that want to be
descriptive?

Prefixing everything with mu4e-query-fragments makes me want to just
change the package name to mu4e-qf, honestly.

wavexx:
I changed the prefix to `mu4e-qf`. If that's not enough, I will rename the package itself.

purcell:
> Prefixing everything with mu4e-query-fragments makes me want to just change the package name to mu4e-qf, honestly.

I actually think the prefix is okay as `mu4e-query-fragments`, particularly for a little package like this, but if you stick with `mu4e-qf`, then yes, it would make sense to rename the package IMO.

wavexx:
On Sat, Sep 09 2017, Steve Purcell wrote:
>> Prefixing everything with mu4e-query-fragments makes me want to just
>> change the package name to mu4e-qf, honestly.
>
> I actually think the prefix is okay as `mu4e-query-fragments`,
> particularly for a little package like this, but if you stick with
> `mu4e-qf`, then yes, it would make sense to rename the package IMO.

Having variables names such as
mu4e-query-fragments-append-to-new-queries is not something I'd want to
do, honestly.

But I also dislike "mu4e-qf" as a package name in terms of
discoverability.

I actually have three more extensions to mu4e that I'd like to publish.
query-fragments is the smallest.

One of them would be "mu4e-jump-to-list", which internally I shortened
to mu4e-jtl-*. Similarly, I would have variables of the form
mu4e-jtl-max-age-threshold, which would become a juggernaut such as
mu4e-jump-to-list-max-age-threshold.

There's a real visual clutter with such long names.
But I'd rather sacrifice the prefix than the informative part of
variable name.

wavexx:
On Sat, Sep 09 2017, Yuri D'Elia wrote:
> On Sat, Sep 09 2017, Steve Purcell wrote:
>>> Prefixing everything with mu4e-query-fragments makes me want to just
>>> change the package name to mu4e-qf, honestly.
>>
>> I actually think the prefix is okay as `mu4e-query-fragments`,
>> particularly for a little package like this, but if you stick with
>> `mu4e-qf`, then yes, it would make sense to rename the package IMO.

Oh well, I went in and renamed the prefix to mu4e-query-fragments.
package-lint now doesn't complain anymore, although there was some eye
watering involved.

wavexx:
@purcell anything missing?

purcell:
Great - merging now. Thanks for your patience. :-)

PR #4982, opened 2017-09-04
Add evil-ledger

atheriel:

`evil-ledger` provides text objects and motions (in the `evil` sense) for working with transactions in a `ledger-mode` buffer, making it easy to select/copy/delete/change them as a unit when `evil-ledger-mode` is active. It also provides some conveniences for navigation, alignment, and sorting.

https://github.com/atheriel/evil-ledger

None needed.

### Checklist

### DIFF
--- a/recipes/evil-ledger
+++ b/recipes/evil-ledger
@@ -1,0 +1,1 @@
+(evil-ledger :repo "atheriel/evil-ledger" :fetcher github)


purcell:
Thanks! Mostly looks great, but `with-eval-after-load` is meant more for end-user configs, and should generally not be used in packages. Are you sure you need to wrap those forms like that for them to work that way?

Also, you should likely autoload `evil-ledger-mode`, so that users can add it to `ledger-mode-hook` without needing to explicitly `require` this file first.

atheriel:
Nice catch on the `autoload`. I had intended to use `with-eval-after-load` to provide "optional" functionality, although it turns out that the appropriate keymap is loaded by `ledger-mode.el` anyway, so the form wasn't actually accomplishing anything -- it's been removed.

But I do have a general question: what's the idiomatic way to include optional support for other packages (e.g. `helm`)? I'd seen some advice that `with-eval-after-load` was the way to go, but is that not your view?

purcell:
> But I do have a general question: what's the idiomatic way to include optional support for other packages (e.g. helm)? I'd seen some advice that with-eval-after-load was the way to go, but is that not your view?

The best way is to provide that as a separate little package with proper dependencies. It's pretty trivial to do that these days. Any other approach has downsides.

Merging this recipe now, in any case.

atheriel:
Thanks!

PR #4984, opened 2017-09-05
Update perl6-mode repo location

zakame:
Per https://github.com/perl6/perl6-mode/issues/11 and https://github.com/hinrik/perl6-mode/issues/20 perl6-mode will now be maintained by the Perl6 community, and so needs an update on MELPA to
use the new location.

Many thanks to @treyharris for initiating the move (https://github.com/perl6/perl6-mode/issues/9,) and to @hinrik for creating this major mode! :tada:

### DIFF
--- a/recipes/perl6-mode
+++ b/recipes/perl6-mode
@@ -1,1 +1,1 @@
-(perl6-mode :fetcher github :repo "hinrik/perl6-mode")
+(perl6-mode :fetcher github :repo "perl6/perl6-mode")
--- a/recipes/perl6-mode
+++ b/recipes/perl6-mode
@@ -1,1 +1,1 @@
-(perl6-mode :fetcher github :repo "perl6/perl6-mode")
+(perl6-mode :fetcher github :repo "perl6/perl6-mode" (:exclude "nqp-mode.el"))


purcell:
Thanks! This should be straightforward, but the repo contents are not the same shape, since the new location has [nqp-mode.el](https://github.com/perl6/perl6-mode/blob/master/nqp-mode.el), which would also get bundled into the package. That file isn't obviously related, has a different name prefix, and invalid headers - what's the plan there? From my point of view it looks like it should be deleted.

zakame:
Thanks @purcell ! @tbrowder might know more about nqp-mode.el as he added that in, though a casual check for diffs with perl6-mode.el suggests no changes, so we could probably remove it or spin it off to another repo.

tbrowder:
@zakame, I believe the nqp mode should stay since it is so closely related to Perl 6.   Making it a separate repo would, I think, create more maintenance problems. Also, a typical Perl 6 dev will be diving into both Perl 6 and nqp code often at the same time.  Dividing the two might make sense for elisp experts, but I have not seen many around here.

zakame:
Thanks @tbrowder.

I think the current concern though is that nqp-mode is *exactly the same* as perl6-mode (as a `diff -Naur` would show,) so perhaps make one a derivative mode of the other, say, perl6-mode derives or is a superset of nqp-mode, via `define-derived-mode` macro; I don't have problems bundling both modes in as long as there's no duplicates.

@purcell for context, nqp-mode aims to provide support for [Not Quite Perl](https://github.com/perl6/nqp), a Perl 6-like environment used in VMs such as the Rakudo Perl6 compiler to further implement Perl 6.

tbrowder:
On Thu, Sep 21, 2017 at 21:40 Zak B. Elep <notifications@github.com> wrote:

> Thanks @tbrowder <https://github.com/tbrowder>.
>
> I think the current concern though is that nqp-mode is *exactly the same*
> as perl6-mode
>

Zak, you and the others are a lot more knowledgeable about elisp than I so
do what is best.  The only differences in the two modes are the regexes
which define the subs and methods, etc. for the indexes which are very
important for my use. As long as a file in my editor identifies the file
type and the indexes are correct, then I am happy!

Best,

-Tom

purcell:
Yep, I've no big problem with including `nqp-mode` in this package, in theory, but in its current form that file will mess things up when `package.el` generates autoloads for all the files in the package: it says, for example, that `perl6-mode` can be autoloaded from `nqp-mode.el`.

So *if* you wanted to include that file, all of its identifiers and headers should be updated so that they are named `nqp-*` rather than `perl6-*`. The alternative is that we simply exclude that file from the recipe. In fact, I'll just do that now in addition to merging this PR.

zakame:
Thanks @purcell @tbrowder !

PR #4985, opened 2017-09-05
Add recipe for nov.el

wasamasa:

nov.el provides a major mode for reading EPUB documents in Emacs.

https://github.com/wasamasa/nov.el

I'm the author.  [My previous PR](https://github.com/melpa/melpa/pull/4976) is a prerequisite for the package.

**None needed**

### Checklist

Please confirm with `x`:

### Notes

Checkdoc insists I document double-dashed functions, doesn't like the argument order at times and incorrectly identifies passive speech.

### DIFF
--- a/recipes/nov
+++ b/recipes/nov
@@ -1,0 +1,1 @@
+(nov :fetcher github :repo "wasamasa/nov.el")


purcell:
Nice work, thanks. I feel like you should add at least one `;;;###autoload`...

wasamasa:
Good catch, thanks!

PR #4986, opened 2017-09-05
Add missing treemacs file.

Alexander-Miller:
Not a new package. This will simply add [a missing python script](https://github.com/Alexander-Miller/treemacs/blob/master/treemacs-dirs-to-collapse.py) to the package built that I didn't know is not caught by :defaults.


purcell:
Yup, `:defaults` is quite conservative. Merging now, thanks.

PR #4987, opened 2017-09-06
Add org babel functions for CFEngine 3

nickanderson:
The code can be found here: https://github.com/nickanderson/ob-cfengine3

### DIFF
--- a/recipes/ob-cfengine3
+++ b/recipes/ob-cfengine3
@@ -1,0 +1,1 @@
+(ob-cfengine3 :repo "nickanderson/ob-cfengine3" :fetcher github)


nickanderson:
Made fixups for package-lint, executed recipie with result:

```
(ob-cfengine3 .
              [(20170906 831)
               nil "Org Babel functions for CFEngine 3" single
               ((:authors
                 ("Nick Anderson" . "nick@cmdln.org"))
                (:maintainer "Nick Anderson" . "nick@cmdln.org")
                (:url . "https://github.com/nickanderson/ob-cfengine3"))])
```

purcell:
Thanks! Few little issues here:
- No licence info. Please use the standard boilerplate, as inserted by `M-x auto-insert`. (Try that in an empty file.)
- You're concatenating filenames into commands without using `shell-quote-argument`, so you're going to have a bad time with paths that contain spaces or quotes.
- Generally you shouldn't use the full path of an executable by default: instead, rely on Emacs having its PATH set properly. So try `cf-agent` instead of `/var/cfengine/bin/cf-agent`.
- `(delete-file tempfile)` should probably be in an `unwind-protect` form, to ensure it always gets called.
- The autoloaded `eval-after-load` block is not idiomatic. Instead, require `ob` at the top, then simply modify the variables directly. See [this](https://github.com/corpix/ob-blockdiag.el/blob/master/ob-blockdiag.el) and [this](https://github.com/stig/ob-applescript.el/blob/master/ob-applescript.el) for examples.

Hope that helps!

nickanderson:
Hey, thanks for the code review!

I will look into these. It may be a while before I can address them.

This was my first package and some of the first elisp i have cobbled together.

nickanderson:
@purcell I tried to address all of your points.

nickanderson:
Thanks!

PR #4989, opened 2017-09-06
Add restore-frame-position

aaronjensen:

Remember and restore initial frame position when restarting Emacs.

https://github.com/aaronjensen/restore-frame-position

maintainer

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/restore-frame-position
+++ b/recipes/restore-frame-position
@@ -1,0 +1,1 @@
+(restore-frame-position :fetcher github :repo "aaronjensen/restore-frame-position")


purcell:
I think the built-in `desktop-save-mode` does this, no?

aaronjensen:
@purcell I didn't know about this mode, thanks! That said, `desktop-save-mode` appears to do too much, at least for my setup. Even when disabling all of the buffer saving, it still breaks my mode line (spaceline) and perspectives when I try to use it.

I can add a note to the readme saying why one may want to use `restore-frame-position` instead of `desktop-save-mode` if you think that would be useful.

kaushalmodi:
I use `desktop-save-mode` to restore my frame. I would suggest filing an emacs bug report explaining the issue you are seeing and how you can start from emacs -Q install only the packages related to this issue and list the steps to recreate the issue.

aaronjensen:
@kaushalmodi what settings do you use? I'll see if I can track down what my issue is. Given that this package is probably redundant, I'll go ahead and close it out.

kaushalmodi:
I just have `desktop-restore-frames` at its default non-nil value. Another important thing for my emacsclient workflow is to load `desktop` after a second after my `init.el` finishes loading. Then I do `(desktop-save-mode 1)` , followed by `(desktop-read)`. My full desktop setup is [here][1].

[1]: https://github.com/kaushalmodi/.emacs.d/blob/master/setup-files/setup-desktop.el

aaronjensen:
@kaushalmodi Thank you for sharing your setup. Seeing your setup and the various workarounds/comments in it, it makes me feel like I'll probably just continue using my package which does one thing and does it well. I'd rather not take the time to try and pare down what `desktop-save-mode` does and work around its issues with my existing setup.

I'll keep my package available on github. It'd be nice to have it in MELPA, but I understand if it wouldn't be acceptable. /cc @purcell

aaronjensen:
I was able to get this working for myself with:

```elisp
  (setq desktop-globals-to-save '()
        desktop-files-not-to-save ".*"
        desktop-buffers-not-to-save ".*"
        desktop-save t)
  (when (window-system)
    (desktop-save-mode)
    (desktop-read))
```

Thanks for the pointers @kaushalmodi and @purcell

purcell:
Nice!

edkolev:
Do you guys happen to know if `desktop-save-mode` can skip saving/restoring the color theme?

If it's not possible, I think this package might be useful.

kaushalmodi:
`desktop-save-mode` does not save the theme for me.

edkolev:
For me, using the configuration by aaronjensen pasted above, it saves stuff such as:
```
...
(foreground-color . "#655370")
(background-color . "#fbf8ef")
(background-mode . light)
(cursor-color . "#100a14")
...
```
in the .emacs.desktop file.

Seems like this is produced by `(frameset-save ...)` from frameset.el though, not by desktop.el.

PR #4990, opened 2017-09-06
Recipe: Add org-web-tools

alphapapa:

This file contains library functions and commands useful for retrieving web page content and processing it into Org-mode content.

For example, you can copy a URL to the clipboard or kill-ring, then run a command that downloads the page, isolates the “readable” content with eww-readable, converts it to Org-mode content with Pandoc, and displays it in an Org-mode buffer. Another command does all of that but inserts it as an Org entry instead of displaying it in a new buffer.

https://github.com/alphapapa/org-web-tools

Author and maintainer.

### Checklist

Please confirm with `x`:

Thanks.

### DIFF
--- a/recipes/org-web-tools
+++ b/recipes/org-web-tools
@@ -1,0 +1,1 @@
+(org-web-tools :fetcher github :repo "alphapapa/org-web-tools")


purcell:
Thanks!

PR #4992, opened 2017-09-06
Add recipe: rg-dired-icons

rgemulla:

Dired minor mode that displays file and directory icons for Windows. Also
provides utility functions for extracting, displaying, and caching icons.

https://github.com/rgemulla/rg-dired-icons

Author and maintainer.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/rg-dired-icons
+++ b/recipes/rg-dired-icons
@@ -1,0 +1,3 @@
+(rg-dired-icons
+ :repo "rgemulla/rg-dired-icons"
+ :fetcher github)


kaushalmodi:
The `rg` prefix of the package make it sound it had to do something with [`ripgrep`][1]. I was confused what "dired icons" had to do with that `rg`, until I realized that those are probably your initials.

[1]: https://github.com/BurntSushi/ripgrep

rgemulla:
True. I am using the prefix for namespacing, but I agree that it is confusing.

The simplest (and perhaps best) option might be to simply remove the prefix and use ```dired-icons``` instead. There is another package called ```dired-icon``` in MELPA, however, which focuses on GTK3 and has a different underlying architecture (e.g., in the way icon sizes are selected and handled). 

If this similarity in naming is fine, please let me know and I'll go ahead and rename things.

kaushalmodi:
May I suggest a name like `dired-icons-win` or something like that that describes your package uniquely. I cannot try the package though as I don't use Emacs on Windows.

rgemulla:
Although that may not mean much, I am using the package since quite a while on multiple computers without experiencing any problems.

I've actually designed the package in a way that makes it relatively easy to add support for other window managers. I'll document how to do this and probably add GTK3 support sometime soon myself.

For this reason, my preference would be to go with ```dired-icons```. If that's not acceptable, I may switch to ```dired-icons-win```, even though I expect that this naming becomes confusing later on.

kaushalmodi:
Just FYI, I am just a casual onlooker who came across this PR.

> I am using the package since quite a while on multiple computers without experiencing any problems.

I don't follow.. the package README says: "Dired minor mode that displays file and directory icons for Windows. " and also a Windows software *ResourceHacker* is listed as a pre-req. Do you mean to say that you use this package on non-Windows machines too?

> I've actually designed the package in a way that makes it relatively easy to add support for other window managers. I'll document how to do this and probably add GTK3 support sometime soon myself.

Have you considered collaborating with devs of other dired-icon packages like [`dired-icon`][1] and [`all-the-icons-dired`][2]?

> For this reason, my preference would be to go with dired-icons. If that's not acceptable, 

I would find it confusing to see both `dired-icon` and `dired-icons` packages listed in the Package manager. I would allow for the Melpa maintainers and others to also comment on that.

> I may switch to dired-icons-win, 

I suggested that as I thought this package is for Windows OS exclusive.

> even though I expect that this naming becomes confusing later on.

This brings us back to the point about collaborating with existing dired icon packages. At the very least if would be useful to explain in your package README how it differs from the other 2 packages if you publish this as a separate package.

Again, I am just a casual bystander who keeps tab on new packages, and thus my curiosity about "rg-" in the package name made me initiate this conversation.

[1]: https://gitlab.com/xuhdev/dired-icon
[2]: https://github.com/jtbm37/all-the-icons-dired

rgemulla:
Well, you do raise some good points. Let's see what the Melpa maintainers say.

For now, I added some remarks about related packages and other minor revisions to the documentation of this package, restructured the package by separating platform-dependent and platform-independent parts, and added documentation on what needs to be done to add support for a new platform.

purcell:
I agree with the feedback about confusing package names, possibly duplicate purpose, and that `rg-` is a confusing prefix. The best option for users would be a patch to the existing `dired-icon` package so that one package can work for people on multiple platforms. It's a waste for lots of people to write their own partial versions of the same package! Would this be feasible?

rgemulla:
Yes, that's feasible in principle.

One potential issue may be that changes to ```dired-icon``` to support multiple platforms may break compatibility with anything built on top (if there are such packages). The behaviour of the two packages is also different in that ```dired-icon``` uses a fixed resolution and does not resize, wheres ```rg-dired-icons``` additionally supports resizing to the current line height (based on the best-fitting available resolution of the corresponding icon).

I suggest to close this pull request for now. I'll keep the package on Github for anyone interested. Once I find the resources to add additional platforms, I'll get in touch with the author of ```dired-icon```, perhaps trying another pull request.

tarsius:
> I suggest to close this pull request for now. 

Done.

PR #4993, opened 2017-09-06
Add recipe for ox-hugo

kaushalmodi:

This library implements a Markdown back-end compatible with the
Hugo static site generator (https://gohugo.io/) for Org exporter.
This includes the generation of the Hugo post front-matter in TOML
or YAML.

https://github.com/kaushalmodi/ox-hugo

Author

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/ox-hugo
+++ b/recipes/ox-hugo
@@ -1,0 +1,2 @@
+(ox-hugo :fetcher github
+         :repo "kaushalmodi/ox-hugo")


purcell:
Thanks!
- The `:files` spec is unnecessary here, I think.
- `ox-blackfriday` looks like it should be a separate package, maybe? Generally all files in a package would have a common name prefix.

kaushalmodi:
The `ox-blackfriday` is current kept as a "silent" exporter. So it's bindings will not show up the Org Exporter dialog, and unnecessarily clutter the interface. If there's demand for standalone `ox-blackfriday` in future, I can package is that way. But right now, it's almost exclusively used by Hugo. 

I'll remove the `:files` spec. I was just being cautious that the `.dir-locals.el` does not get included.

kaushalmodi:
> ox-blackfriday looks like it should be a separate package, maybe? Generally all files in a package would have a common name prefix.

At present, `ox-blackfriday` is kept as an internal separate library, and so it has its own namespace. Is that a big concern. If I package it separately, I would need to expose it's bindings in the Org Exporter menu, which will introduce clutter, as no one would use them.

kaushalmodi:
> The :files spec is unnecessary here, I think.

I have force pushed an update to this PR.. removing the `:files`.. they were not needed as you said.

Is the other point you raised a blocker to this PR?

purcell:
Should be fine, but definitely consider extracting `ox-blackfriday` as a separate package at some point. :-)

kaushalmodi:
> consider extracting ox-blackfriday as a separate package at some point. :-)

I'll definitely do that when there's need. The package is almost ready to be published if anyone ever needed standalone `ox-blackfriday`.. I'd just need to uncomment [these lines][1]. 

And thanks for merging this recipe.. my first major package :D

[1]: https://github.com/kaushalmodi/ox-hugo/blob/69a6001b00b6fe35404c4220e1e476724719215f/ox-blackfriday.el#L51-L60

PR #4995, opened 2017-09-07
Add replace-with-inflections, inflection aware `query-replace'

knu:

This package provides `query-replace-names-with-inflections`.

https://github.com/knu/replace-names.el

I'm the author.

**None needed**

### Checklist

Please confirm with `x`:

    The main (and only interactive) function does not match (start with) the package name, but I think it's a good package name.

    Internal functions do not (yet) have docstrings, though.

### DIFF
--- a/recipes/replace-with-inflections
+++ b/recipes/replace-with-inflections
@@ -1,0 +1,1 @@
+(replace-with-inflections :repo "knu/replace-with-inflections.el" :fetcher github)


kaushalmodi:
What are *inflections*? Can you add a README with examples and may be a screenshot?

knu:
OK, I will.  It uses existing packages named `inflections` and `string-inflection`, so I thought users who'd be search packages related to the idea of inflection should know that already.  It is an implementation of naming conventions adopted in the Ruby on Rails framework and probably well known in the web developers.

purcell:
Thanks! This looks useful. I feel like the package should be renamed to `replace-with-inflections`. `replace-names` is very generic, whereas `replace-with-inflections` has a clear purpose.

knu:
Thanks for the suggestion.  I don't have any plan for the moment to add a function that is unrelated to inflections, so having the word "inflections" in the name would be reasonable.

knu:
Renamed, added README and documentation and updated this PR.

purcell:
Great, thanks!

purcell:
P.S. https://github.com/eschulte/jump.el/pull/13 is now merged, so you can remove the monkey-patch. I'd also suggest renaming the `query-replace-names-with-inflections` command to `query-replace-with-inflections`, or simply `replace-with-inflections`. And `cl` appears not to be necessary.

knu:
Thanks!

knu:
@purcell Thanks, the first two have been addressed.  Regarding `'cl`, it was for using `letf`.  Should I use `cl-letf` via `'cl-lib` instead?

purcell:
Yeah, that'd be great: just require `cl-lib` and add `(cl-lib "0.5")` to the `Package-Requires` header.

PR #4997, opened 2017-09-09
Receipe: circadian - automatic theme-switching based on daytime

GuidoSchmidt:

Circadian provides a way to automatically switch themes for day and night and features two themes: hemera-theme and nyx-theme.

https://github.com/GuidoSchmidt/circadian.el

Creater and maintainer 🙂 

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/circadian
+++ b/recipes/circadian
@@ -1,0 +1,1 @@
+(circadian :fetcher github :repo "GuidoSchmidt/circadian.el")
--- a/recipes/hemera-theme
+++ b/recipes/hemera-theme
@@ -1,0 +1,1 @@
+(hemera-theme :fetcher github :repo "GuidoSchmidt/emacs-hemera-theme")
--- a/recipes/nyx-theme
+++ b/recipes/nyx-theme
@@ -1,0 +1,1 @@
+(nyx-theme :fetcher github :repo "GuidoSchmidt/emacs-nyx-theme")


purcell:
Thanks! I would use something like this myself. Quick comments:
- Please package each theme separately, since they don't belong hidden in another package.
- `(when (not x) ..)` is better written `(unless x ...)`
- Prefer the `-p` to `?` for predicate functions: it's more idiomatic in elisp.
- The `linum`-related code doesn't really belong in this package
- The use of hooks to trigger the theme changes is weird, particularly the use of `after-change-major-mode-hook`. Instead, I'd expect to see timers.
- This would all be best wrapped up as a global minor mode can be toggled.
- Why only two themes? I could imagine using three or more. Maybe just define everything with an `alist` of `(hour . theme-name)`, and perform a lookup from top to bottom based on the current hour.
- You can avoid the kind of code in `circadian-load-theme-disable-old-theme` by setting themes with `(custom-set-variables `(custom-enabled-themes '(,name)))`, where `name` is a variable containing the theme name.

Hope that helps!

GuidoSchmidt:
I'm still beginner at Elisp and still learning a lot, therefore much thanks for the nice and detailed feedback! 🙂 👍 I will try to implement your suggestions.

GuidoSchmidt:
@purcell I've mostly re-implemented the package thanks to your great suggestions, timed functions work great - thanks for pointing me into that direction. Themes have been moved to their own packages.

purcell:
Nice! Merging now. Welcome to MELPA!

PR #4999, opened 2017-09-10
Suggest using the latest version of package-lint

alphapapa:
It seems like package PRs are often submitted with the package-lint box checked, but with problems that should have been detected by it.  Maybe they were on an old version of package-lint.  I don't upgrade my packages very often, so I'm probably guilty of this myself.


purcell:
Yes! Thanks. :-)

PR #5001, opened 2017-09-10
Restore yatex since v1.8 now satisfies GPL-compatibility

takaxp:

YaTeX is an intelligent, acquisitive and integrated package which reduces your efforts of composing LaTeX source on Emacs.

http://www.yatex.org/hgrepos/yatex

An enthusiastic user (who consulting with the maintainer)

The maintainer selected The 2-Clause BSD License for this package from the latest version(1.8) released on 2017-09-09. Now the previous recipe should be re-registered into MELPA repository.

### Checklist

Please confirm with `x`:

- [-] I've used the latest version of [package-lint](https://github.com/purcell/package-lint) to check for packaging issues, and addressed its feedback
  - I did it. Actually, it reports many errors. However, this package has been registered and distributed without any problems for years until the GPL compatibility issue happened. So I would like to recommend that the recipe for this package would be re-registered and distributed from MELPA as it is. This package is very popular and matured. Additionally, I believe it will take long time to strictly meet the current requirements for MELPA registration. This will cause many existing users to lose opportunities to install this package from MELPA, actually it has already happened in the past few weeks.
- [-] `M-x checkdoc` is happy with my docstrings
  - There are some missing elements. However, I propose this package can be distributed as it is for the same reason that I described above.

### DIFF
--- a/recipes/yatex
+++ b/recipes/yatex
@@ -1,0 +1,4 @@
+(yatex
+ :fetcher hg
+ :url "https://www.yatex.org/hgrepos/yatex/"
+ :version-regexp "yatex-\\(.*\\)")


purcell:
Thanks. I agree that there's a lot of work needed to tidy up `yatex`, but that it is still worth re-importing it. Would you mind at least changing the `yatex.el` file header so that it states the license details please? The current header is quite messy and non-standard, so  best way to do this would be to create an empty `yatex.el` and run `M-x auto-insert`, then copy in the code and details. This way you'll get a proper file structure. :-)

takaxp:
Hi purcell. Many thanks to you for understanding the current situation on yatex package and also providing us kind and practical instruction. I'll contact the maintainer to change the yatex.el appropriately, and report back to here ASAP.

takaxp:
Hi purcell. I have received a quick reaction from the maintainer. The modified yatex.el shows the licence information explicitly as [1]. At this stage, we applied minimal changes on yatex package to re-importing it immediately. The maintainer is now pretty busy but the remaining format issues will be addressed and continuously resolved in the future.

[1] https://www.yatex.org/hgrepos/yatex/hgweb.cgi/rev/9c857ce72a03

purcell:
Thanks!

Issue #5002, opened 2017-09-10
New release of caml


Chris00:
The files in the _caml_ package have been updated since 2015.  Please do a new release.

aplaice:
Has the official/canonical source moved from the [svn repository](https://caml.inria.fr/svn/ocaml/trunk/emacs/) to [github](https://github.com/ocaml/ocaml/tree/trunk/emacs)? The MELPA caml ["recipe"](https://github.com/melpa/melpa/blob/master/recipes/caml) currently still points to svn, which hasn't been updated since 2015 — MELPA packages are updated automatically based on changes in the source repository.

Chris00:
I think it is better to use the Github repository.

purcell:
Hey @Chris00, note that `caml` is in danger or being removed from MELPA due to invalid package formatting (see #5010). Perhaps you'd like to follow up upstream to get those problems fixed?

Chris00:
@purcell Thanks for letting me know.  Would these header be fine https://github.com/Chris00/ocaml/blob/melpa/emacs/caml.el ?  (If yes, I'll submit a PR to the main repo.)

purcell:
Yes, that looks great!

Chris00:
Here is the [PR](https://github.com/ocaml/ocaml/pull/1346).

Issue #453, opened 2013-01-13
how do we get travis to consider the current build a failure?


milkypostman:
seems to allow things to pass even though there is part that is failing...

https://travis-ci.org/milkypostman/melpa

Specifically this build: https://travis-ci.org/milkypostman/melpa/builds/4121263

PR #5003, opened 2017-09-11
Update location of ocaml repository for caml recipe

aplaice:

https://github.com/ocaml/ocaml/tree/trunk/emacs

See #5002.

It seems that OCaml is now mainly using github, rather than their own svn repository, and an upstream contributor thinks that fetching from github is better.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/caml
+++ b/recipes/caml
@@ -1,1 +1,3 @@
-(caml :fetcher svn :url "https://caml.inria.fr/svn/ocaml/trunk/emacs/")
+(caml :fetcher github
+ :repo "ocaml/ocaml"
+ :files ("emacs/*.el"))


aplaice:
Aside (possibly should be a separate issue): why does Travis occasionally "succeed" despite not having built the relevant package or even tried building it?

It has occurred for [this build](https://travis-ci.org/melpa/melpa/builds/274191089) and in [the past](https://travis-ci.org/melpa/melpa/builds/258229856). In the latter case, the merging of the pull request resulted in the password-store package no longer being updated. (This was a case where Travis was actually needed, since the problem appears to be that older versions of Ubuntu don't have the appropriate root certificates, which is not something that the person sending the pull request could have been aware of.)

Possibly relevant (very old and closed) issue: #453.

In the specific case of this pull request, the relevant lines from the Travis build log are [1051-1052](https://travis-ci.org/melpa/melpa/builds/274191089#L1051):

    Building recipes touched in commits 3acf5514346d82389e4b98fa6dc27e9edacf5e61...7fceea40d54b7d671c06892390410d0c7001d417

    fatal: bad object 4928e8df7fd6657199e7f2a339fc5f1c07a786d8

The error is presumably generated [here in run-travis-ci.sh](https://github.com/melpa/melpa/blob/master/run-travis-ci.sh#L16).

I think that the error is spurious (the object does not exist and is not referred to by any of the changed tree objects), but since I do not understand why it's occurring, I cannot say for sure. Also, obviously the building of the package has not been tested by Travis (though the package does build locally).

PR #5007, opened 2017-09-14
Add recipe for evil-expat

edkolev:

Adds evil ex commands, ones that a vim user might be accustomed to.

https://github.com/edkolev/evil-expat

Maintainer.

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/evil-expat
+++ b/recipes/evil-expat
@@ -1,0 +1,1 @@
+(evil-expat :repo "edkolev/evil-expat" :fetcher github)


tarsius:
@purcell I think you forgot about this one. I would do the initial review myself, but I know to little about `evil` - for example I don't know whether the lack of autoloads is an issue for extensions to it.

edkolev:
I don't like the lack of autoloads myself, however, evil doesn't support autoloading its `ex` commands. I've brought this up here https://github.com/emacs-evil/evil/issues/926 but we haven't reached a solution mainly because I haven't been able to suggest a proper solution.

purcell:
Looks good to me, code-wise, but I have no idea what it does, and I'm not sure that any potential users would be able to figure it out or recognise that this is a package which might be helpful to them -- the package description is `Evil ex commands` and there's no `Commentary` to speak of. Can I suggest you add a little more information before we merge this, please?

ogdenwebb:
I'm evil and ex vim user. Though I appreciate @edkolev's help for Emacs/evil community, but I can't understand some parts of this package. 

Firstly, diff-orig, remove, rename and reverse commands are really useful. Also I'm wondering why don't you wanna to add them directly into evil package? 

But the rest of commands, hm. For example, what is a point to  add `:colorscheme` ex commands, when evil already can call Emacs interactive(i.e. `M-x`) functions through evil ex? With `:load-theme` you have to press  12 keys(if <Shift> counted) and with `:colorscheme` requires from user 13 keys - as you can see, no benefits. I prefer to call `load-theme` and choose theme from ivy/helm/etc minibuffer dialog instead.

Other new commands even more confusing for more: evil provides the simple way to define an ex command via `define-ex-define-cmd`, so there's no reason to add extra aliases for the default Emacs interactive functions, in my opinion.

The remaining commands require corresponding packages and two questions appear:
1. Not all users(me too, at least tmux) are using these tools. I think will be great to define them as optional ex commands. I mean something like `(evil-expat-ext-cmd '(magit tmux ...))` 
2. Even though I'm using magit, I bind its commands via <leader> key, like Spacemacs does. So optional ex commands is a good thing.

p.s. As I said, i'm happy with evil-lion and evil-goggles, but specifically this package is a bit strange.
p.p.s. Would be awesome if you start think about vanilla vim U map(i.e. undo line).

edkolev:
@purcell thank you for looking into this!

Regarding the documentation - sure, I'll extend it, both the Commentary and the README.md. I'll write here when I do so.

Regarding autoload-ing - [this PR](https://github.com/emacs-evil/evil/pull/949) was just merged in evil, which means that all the functions provided by evil-expat will be autoloaded by using autoload cookies; this is the [relevant commit](https://github.com/edkolev/evil-expat/commit/3d31a5e831cf2b251cdf796ab24e20ac546cae4e) in evil-expat.

@ogdenwebb evil-expat is meant to provide ex commands, which a vim user is likely to be used to. I agree that the `:colorscheme` command might be a good candidate for evil itself, but the others, like `:rename` and `:remove` - they aren't even in vim, they are provided by [vim-eunuch](https://github.com/tpope/vim-eunuch), hence they aren't a good suite for evil itself. BTW you can mark a line/region (with `V` in evil-mode for example) and do `M-x undo-tree-undo` to undo just in this line.

ogdenwebb:
@edkolev well, I understand your goal now, but don't forget my commentary about commands, which require external packages.

edkolev:
Hey, I finally updated the Commentary and Usage sections - here they can be viewed https://github.com/edkolev/evil-expat/blob/master/evil-expat.el#L28

@purcell please let me know if I should further improve the docs or anything else.

edkolev:
@ogdenwebb I've created an issue to track you request https://github.com/edkolev/evil-expat/issues/1

Any further suggestions for improvement are welcome!

purcell:
Merged now - thanks for your patience.

edkolev:
Thank you @purcell for your feedback

PR #5011, opened 2017-09-16
Remove org-mac-iCal and org-mac-link recipes.

joewreschnig:
These recipes use an insecure plain “git://” protocol (issue #3004).
They are also available in the “org-plus-contrib” package in
org-mode’s own ELPA, for users who want to explicitly add an insecure
archive or verify orgmode.org’s self-signed certificate themselves.


purcell:
I believe I added these myself, and don't personally want to use `org-plus-contrib` instead. I guess the root problem here is that the org folks don't have any momentum behind their packaging. :-(

PR #5018, opened 2017-09-24
Update recipes for paper-theme and forecast.

cadadr:
I've moved my public Elisp stuff into a single repository for
convenience (a repo per project was too much as all of them are
kind-of small and it's tedious to maintain one for each).  This commit
updates my recipes to fetch relevant files from the new
github.com:cadadr/elisp repo.

These packages were already on Melpa.

https://github.com/cadadr/elisp

I am the maintainer of both projects

I don't know if it poses a problem but I've removed the release tags in order to switch to using the master branch as the latest stable version of each script.

### Checklist

Please confirm with `x`:

(I'm unsure about the non-checked ones, but the actual elisp files themselves are unchanged, and they were on Melpa since some time already, so I guess there should not be any problems merging.  I'll be happy to fix anything wrong with my patch/packages otherwise.  Sorry for the inconvenience.)

### DIFF
--- a/recipes/forecast
+++ b/recipes/forecast
@@ -1,1 +1,1 @@
-(forecast :fetcher github :repo "cadadr/forecast.el")
+(forecast :fetcher github :repo "cadadr/elisp" :files ("forecast.el"))
--- a/recipes/paper-theme
+++ b/recipes/paper-theme
@@ -1,1 +1,1 @@
-(paper-theme :fetcher github :repo "cadadr/paper-theme")
+(paper-theme :fetcher github :repo "cadadr/elisp" :files ("paper-theme.el"))


purcell:
Thanks. Mostly looks good, but I double-checked and there are two little issues that need fixing, please:

- `bsdpkg.el` is missing a mandatory `Package-Requires` headers, since it uses `cl-lib`: `package-lint` can advise you there.
- `pass-listing` has a malformed `Package-Requires` headers: the `.` in `((password-store . "0.1"))` should be removed. Additionally, the use of `defvar-local` requires you depend on `(emacs "24.3")`.

cadadr:
On 2017-09-24 22:54 +07, Steve Purcell <notifications@github.com> wrote:
> Thanks. Mostly looks good, but I double-checked and there are two little issues that need fixing, please:
>
> - `bsdpkg.el` is missing a mandatory `Package-Requires` headers, since it uses `cl-lib`: `package-lint` can advise you there.
> - `pass-listing` has a malformed `Package-Requires` headers: the `.`
> in `((password-store . "0.1"))` should be removed. Additionally, the
> use of `defvar-local` requires you depend on `(emacs "24.3")`.

I've updated the said packages according to package-lint, but if I'm not
mistaken, these files are not supposed to be on Melpa anyways, i.e., the
recipes should only include forecast.el and paper-theme.el.

tarsius:
> I've moved my public Elisp stuff into a single repository for convenience (a repo per project was too much as all of them are kind-of small and it's tedious to maintain one for each).

A drawback of that is that people who want to install one of your packages by cloning the repository and adding it to the `load-path` now have to install all of them. One can do that manually or using either [`borg`](https://github.com/emacscollective/borg) or [`straight`](https://github.com/raxod502/straight.el) (two alternative package managers that recently gained some popularity).

It also makes it impossible to mirror your unrelated packages *as* separate packages. Well actually it is possible, but the repositoryies on the [Emacsmirror](https://emacsmirror.net) won't contain the actual history, just "downloaded a new snapshot" commits.

The general trend is to move *away* from bundling unrelated things into one all-my-elisp repository and I would like to ask you to reconsider.

What is tedious about maintaining a repository per package? Once the repositories exist both locally and on github, and `init.el` contains the necessary `(add-to-list 'load-path ...)`s it seems to me it basically boils down to having to `cd` occasionally.

tarsius:
> I don't know if it poses a problem but I've removed the release tags in order to switch to using the master branch as the latest stable version of each script.

That means that Melpa-Stable users will be forever stuck on the last tagged versions.

cadadr:
As I said, I mostly have little pieces of elisp for various things, and
I do want to share some of them with the community as they are generic
enough to possibly be useful to others.  Having a repository for each
one is tedious in that as I only modify these stuff ever so often, and
when they’re in one tree each I tend to lose context, and not publish
some stuff because of that.

WRT removing tags, well, these scripts are edited seldom enought that
about each release has one or two commits, so it becomes meaningless.  I
will keep updating the Version: headers in the scripts themselves
though.

tarsius:
Understood. Well, I had to try :wink:

cadadr:
Thanks!

PR #5021, opened 2017-09-26
change unstable branch of ensime

fommil:
We have split out our "unstable" release branch (`2.0`) and our dev branch (`3.0`, will break all the things). So changing melpa to point to `2.0`.

### DIFF
--- a/recipes/ensime
+++ b/recipes/ensime
@@ -1,2 +1,3 @@
 (ensime :fetcher github
+    :branch "2.0"
 	:repo "ensime/ensime-emacs")


purcell:
So you have two stable branches, with different levels of stability? Generally we prefer to track the development version in MELPA, which still allows the distribution of more stable releases via tagging, and therefore MELPA Stable. For this reason, `:branch` in recipes is deprecated.

fommil:
We have stable (tagged) and unstable (2.0) and completely-unusable-rewrite (3.0).

How would you handle the case where there is no `master` branch?

BTW, 2.0 is our default branch on github... if you default to the default then I can close this PR as that's exactly what we'd want :smile:

purcell:
> if you default to the default then I can close this PR as that's exactly what we'd want 😄

Yeah, pretty sure that's what's happening, so we're probably all set here.

fommil:
niiiiice :smile:

PR #5022, opened 2017-09-28
Add recipe for counsel-etags

redguardtoo:

Fast and complete ctags/etags solution using ivy-completion

https://github.com/redguardtoo/counsel-etags

I am the maintainer.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/counsel-etags
+++ b/recipes/counsel-etags
@@ -1,0 +1,1 @@
+(counsel-etags :fetcher github :repo "redguardtoo/counsel-etags")


purcell:
Thanks. Here are a few of the issues I noticed in a quick review:

- This is extremely dependent on `counsel`, to the extent that it should probably be called `counsel-mctags`.
- There are many instances where you're building command strings without escaping filenames that might contain spaces or quotes.
- You should use `when` instead of `if` wherever possible
- Don't use `let` to introduce `nil` values that you then set with `setq`: there are almost always neater approaches, e.g.
```el
(defun mctags-guess-program (name)
  "Guess executable path from its NAME on Windows."
  (or (when (eq system-type 'windows-nt)
        (let ((path (concat "\\\\cygwin64\\\\bin\\\\" name ".exe")))
          (cond
           ((file-executable-p (concat "c:" path)))
           ((file-executable-p (concat "d:" path)))
           ((file-executable-p (concat "e:" path))))))
      name))
```
- Don't `require` features from inside a function. Instead, use `(require 'xref nil t)` at the top level, then check with `fboundp` that `xref-pulse-momentarily` is bound.
- You should `(require 'cl-lib)`

redguardtoo:
Hi, @purcell,
Thanks for the feedback. I've updated the code as you required except the project name.

I prefer using `mctags` because I plan to use both `ido` and `counsel` in the future. Even `counsel` might be slow on my low spec computers. 

My package is actually not heavily dependent on `counsel`. It only use one API `ivy-read`. Other functions like `occur` are addons of `ivy-read` (they will only be triggered after calling `ivy-read`). So they can be easily disabled.

purcell:
Thew best approach would be to use standard functions like `completing-read` so that users can use their preferred completion framework. `helm` and `ivy` will generally work automatically in that case: there's no need to restrict users to your favourite completion mechanism.

redguardtoo:
I have no plan to use helm because it's slower than ido. Besides, there are already two ctags related packages: `heml-tags.el` and `helm-etags-plugs.el` for people who prefer helm.

One major reason I develop this package is my program is much faster than helm packages.

I will add an extra wrapper named `mctags-completing-read` like https://github.com/technomancy/find-file-in-project/blob/dcc0801c4781d76efd9ac0d86bab4084b4b93663/find-file-in-project.el#L530

That solution requires minimum setup from users (toggle a boolean global flag).

redguardtoo:
Hi, @purcell , I change the project name from `mctags` to `counsel-etags`, please review.

purcell:
Thanks, and sorry for the delay. :-)

redguardtoo:
Cool. Thank you.

PR #5023, opened 2017-09-28
Recipe for projectile-trailblazer added

micdahl:

The package adds a minor mode for finding files in Rails projects which use Trailblazer.
It namely adds support for finding cells, contracts, operations and views. 

https://github.com/micdahl/projectile-trailblazer

I am the maintainer, even if the package is heavy inspired by projectile-rails.

*none needed*

### Additional info

`package-lint` shows 

> 109:0: error: Global minor modes must `:require' their defining file (i.e. ":require 'projectile-trailblazer"), to support the customization variable of the same name.

 But I don't see, how to fix that. If I add `:require 'projectile-trailblazer`, I get errors when running my unit and integration tests. If I leave it as it is, everything works fine (as far as I see).

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/projectile-trailblazer
+++ b/recipes/projectile-trailblazer
@@ -1,0 +1,2 @@
+(projectile-trailblazer :repo "micdahl/projectile-trailblazer"
+                  :fetcher github)


purcell:
> If I add :require 'projectile-trailblazer, I get errors when running my unit and integration tests. If I leave it as it is, everything works fine (as far as I see).

Weird. Not sure what's going on there, but it'll almost certainly be a quirk of your testing setup.

In any case, merging this now - thanks for your patience.

PR #5026, opened 2017-09-29
Open source links in new tab

edkolev:
This PR adds `target="_blank"` to Source links. This is more intuitive in my opinion since Source links are external to melpa.org.

Thanks

### DIFF
--- a/html/js/melpa.js
+++ b/html/js/melpa.js
@@ -330,7 +330,7 @@
               m("td.recipe",
                 m("a", {href: p.recipeURL}, glyphicon('cutlery'))),
               m("td.source",
-                p.sourceURL ? m("a", {href: p.sourceURL}, p.source) : p.source),
+                p.sourceURL ? m("a", {href: p.sourceURL, target: "_blank"}, p.source) : p.source),
               m("td", [p.downloads.toLocaleString()])
             ]);
           }))


edkolev:
To be clear, this is the link this PR changes:

screen shot 2017-09-29 at 14 27 11

purcell:
Sorry, I don't agree, and this isn't good practice these days, so I'm going to decline this. (Additionally, fyi, such `target` options are generally unsafe without also specifying `rel="noopener"`.)

purcell:
(Better might be to indicate which are external links.)

edkolev:
OK, thanks for consider it.

I'm used to having external links open in new tabs in any website, hence this PR.

edkolev:
Opening external links in another tab because (quoting myself) "I'm used to having external links open in new tabs" isn't a good enough reason, I agree.

However, since melpa.org is js-based, this scenario seems painfully inconvenient to me:
1. search for `light theme`
2. 34 themes match the criteria, click on any of them
3. click on any theme's Source link (that's where they usually have a screenshot)
4. click the browser's back button, you're back on melpa.org and your search is gone, 3700+ package a listed

Maybe opening in a new tab isn't the best solution here, I'm not sure what is. Would be nice to have this scenario addressed in some way, be it with `target=_blank` or not.

purcell:
You're quite right. I've made the website work a bit better in that respect, in 1fa150a. :-)

edkolev:
Much better, thank you :)

purcell:
Thanks for prodding me to fix it. It had been annoying me for ages, but I lacked the motivation to jump in and work on the javascript.

PR #5027, opened 2017-09-29
Add recipe for basic-mode

dykstrom:

Package basic-mode is a major mode for editing BASIC source code

https://github.com/dykstrom/basic-mode

I am the creator and maintainer of the package

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/basic-mode
+++ b/recipes/basic-mode
@@ -1,0 +1,4 @@
+(basic-mode
+ :repo "dykstrom/basic-mode"
+ :fetcher github
+ :files ("src/*.el"))


purcell:
Thanks! Quick feedback:
- You can use `syntax-ppss` to do [this](https://github.com/dykstrom/basic-mode/blob/1dae06e2b09bd3903643f640badb769f1d50b04a/src/basic-mode.el#L110-L114) much more efficiently.
- IIRC, it's possible to tell `font-lock` to not be case-sensitive w.r.t. keywords etc., which I think would let you tidy things up.
- Instead of listing the (incomplete) bindings list [here](https://github.com/dykstrom/basic-mode/blob/1dae06e2b09bd3903643f640badb769f1d50b04a/src/basic-mode.el#L110-L114), just put `\{basic-mode-map}` in the docstring so that Emacs will do it for you automatically.

None of these are a blocker, so I'm merging now, but they're worth addressing IMO.

dykstrom:
Hi, 

thanks for your comments! Those were things I didn't know, but I'll try to address them in a future version of basic-mode some time soon.

BR, Johan

purcell:
Specifically, `(save-match-data (nth 8 (syntax-ppss)))` returns a non-`nil` value if `point` is in a string or comment, which is pretty convenient. :-)

dykstrom:
I have tried that out, but I experienced one problem. If the point is on the very first character of a string or comment (for elisp that would be the " or ; character), the expression above returns nil. Do you have any good solution for that?

purcell:
> I have tried that out, but I experienced one problem. If the point is on the very first character of a string or comment (for elisp that would be the " or ; character), the expression above returns nil. Do you have any good solution for that?

Ah, yes: what you want is the following instead:
```el
(let ((ppss (syntax-ppss)))
    (or (car (setq ppss (nthcdr 3 ppss)))
        (car (setq ppss (cdr ppss)))
        (nth 3 ppss)))
```

PR #5028, opened 2017-09-29
Add jsonnet-mode recipe

mgyucht:
This package adds support for Jsonnet in Emacs. It provides syntax highlighting, indentation, and some useful commands like jump to definition and render a Jsonnet document into a buffer. It's rather beta, but I want to get feedback on how I could improve the library.

https://github.com/mgyucht/jsonnet-mode

I'm the maintainer of jsonnet-mode.

N/A

### Checklist

Please confirm with `x`:

 - There are two docstrings that it doesn't like, but I think that what I currently have here is fine.


purcell:
Thanks. Quick feedback:

- You're missing a `defgroup` form for the `jsonnet` you're referencing in your `defcustom`s
- [This regexp](https://github.com/mgyucht/jsonnet-mode/blob/8eb547f9c4facadb142c1f0e47235c7dc51b8fc9/jsonnet-mode.el#L56) is a bit crazy: why not just use `regexp-opt`, which can produce comparable regexps from a readable list of strings.
- The [M-. binding](https://github.com/mgyucht/jsonnet-mode/blob/8eb547f9c4facadb142c1f0e47235c7dc51b8fc9/jsonnet-mode.el#L392) isn't idiomatic. It usually means "jump to thing at point", which would be closer to the `jsonnet-find-function-at-point` function than what you've bound it to. It should also work in such a way that `M-,` returns to the previous position, which often requires a bit of extra work: check out other modes which support this.
- Using `prog-mode` requires that you depend on `(emacs "24")`. `package-lint` should have mentioned that.
- This recipe will bundle `jsonnet-flycheck`, but that's bad for a couple of reasons. There's no dependency on `flycheck`, so byte-compilation will fail. But anyway, it's not idiomatic to ship major modes with not-always-applicable extras that have opinionated dependencies: a major mode should focus on font lock etc. So I'd suggest adding package headers (with appropriate dependencies) to `jsonnet-flycheck.el` and submitting two recipes rather than one. Alternatively, you might consider submitting a PR to flycheck which adds this checker.

Hope that helps!

mgyucht:
Thank you so much for your thorough review @purcell! I'm very new to Emacs Lisp so I'm still trying to understand not only how to write Emacs Lisp but also what conventions are expected by the community. I've addressed all of your comments. I wasn't able to figure out how to implement the mark pushing and popping, so I've instead just bound that command to C-c C-f. Let me know if you have any other concerns or comments!

purcell:
Cool, merged now. Re. the pushing/popping, check out the code in https://github.com/purcell/elisp-slime-nav/blob/master/elisp-slime-nav.el for an example. In short, you'd generally store the current position with code like this before jumping:
```el
(if (fboundp 'xref-push-marker-stack)
          (xref-push-marker-stack)
        (with-no-warnings
          (ring-insert find-tag-marker-ring (point-marker))))
```
and then you'd bind `M-,` to `'pop-tag-mark` for jumping back.

mgyucht:
Excellent, thank you so much! Really excited to share this package around!

PR #5029, opened 2017-09-30
Add recipe for company-plsense

CeleritasCelery:

company backend for Perl5 via [plsense](https://github.com/aki2o/emacs-plsense)

https://github.com/CeleritasCelery/company-plsense

Author

none needed

### Checklist

### DIFF
--- a/recipes/company-plsense
+++ b/recipes/company-plsense
@@ -1,0 +1,3 @@
+(company-plsense
+ :fetcher github
+ :repo "CeleritasCelery/company-plsense")


purcell:
Thanks! Today I learned about `tq` - very neat. Quick feedback:

- You're missing a dependency declaration for `cl-lib`. `package-lint` should have pointed that out...
- [Here](https://github.com/CeleritasCelery/company-plsense/blob/c388cb4d88e11886c022f0449c5fa8fa378e18ba/company-plsense.el#L52) you're saying that the customisation group is its own parent.
- You can drop the `:group` keys from the `defcustom` items, since they will default to the last-defined group above.
- Is [this](https://github.com/CeleritasCelery/company-plsense/blob/c388cb4d88e11886c022f0449c5fa8fa378e18ba/company-plsense.el#L218) safe for filenames containing spaces or quote characters? I feel like you'd want to use `shell-quote-argument` or similar here.

CeleritasCelery:
I was under the impression that `cl-lib` was built-in and that's why I didn't include it. Also `package-lint` was clean. Either way I will address your feedback and post a new comment. Thanks!

CeleritasCelery:
Alright @purcell I fixed all the issues you brought up except the one below

> Is this safe for filenames containing spaces or quote characters?

I am glad you pointed it out, because it forced me to double check. When we create a process with `plsense --interactive` it starts a it's own REPL that doesn't have a problem with spaces, quotes, `$`'s, etc in file names. So the answer is yes.

purcell:
> I was under the impression that cl-lib was built-in and that's why I didn't include it. Also package-lint was clean.

Turned out that's a bug in `package-lint` - thanks for discovering it!

purcell:
Super, thanks - merged!

purcell:
And... welcome to MELPA. :-)

PR #5030, opened 2017-09-30
Add ‘apt-sources-list’ recipe

joewreschnig:

A major mode for editing APT’s `.list` files.

The `/etc/apt/sources.list` file and other files in `/etc/apt/sources.list.d` tell APT, found on Debian-based systems and others, where to find packages for installation.

This mode is derived from `apt-sources-mode` distributed in the `debian-el` Debian package, but has many differences. Among them:
- The load hook has been removed. Use `with-eval-after-load`.
- What is properly called a “suite” is no longer called a “distribution.”
- Most keys have changed. This is for better mnemonics (`C-c C-s` to change the suite), or for better Emacs integration (`C-M-n` or your `forward-list` equivalent to move between source lines).
- Editing functions may be called more easily from Emacs Lisp.

https://git.korewanetadesu.com/apt-sources-list.git

Maintainer of this, long-time user of the `debian-el` version.

I emailed the Debian maintainer of `apt-sources-mode` a while back but did not receive an answer. The "official" mode is increasingly bad for me without option support, and it's generally not maintained (for example, it was never updated to add the new Debian release).

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/apt-sources-list
+++ b/recipes/apt-sources-list
@@ -1,0 +1,2 @@
+(apt-sources-list :fetcher git
+                  :url "https://git.korewanetadesu.com/apt-sources-list.git")


purcell:
Thanks, and sorry for the delay.

PR #5042, opened 2017-10-06
Add imake recipe

tarsius:
### Prelude

In the past I have sometimes just pushed recipes for my own packages instead of going through the usual review, but I am not doing that in this case because you might consider `imake` to be one that falls into the *overspecialized, just for personal use* category. But first...

This package provides the command `imake`, which prompts for
a `make` target and runs it in the current directory.

This is an opinionated command suitable for simple Makefiles
such as those that can be found in the repositories of some
Emacs packages.  The make targets to be offered as completion
candidates have to be documented like so:

```make
help:
        $(info make lisp   - generate byte-code and autoloads)
        $(info make clean  - remove generated files)
```

More precisely, a `help` target containing lines that match
the regexp `"^\t$(info make \\([^)]*\\))"` is expected.

### Interlude

So it is a bit opinionated, but not that much - having a `help` target makes sense, and using the `info` function too.

https://github.com/tarsius/imake

Author.

:wine_glass: 

### Checklist

  See https://github.com/purcell/package-lint/issues/93.
  Well mostly. It wants me to add some typos to example configuration.
- [x] I've built ~~and installed~~ the package using the instructions in [CONTRIBUTING.md](https://github.com/melpa/melpa/blob/master/CONTRIBUTING.md)
  And it actually caught a stupid pasto.

### DIFF
--- a/recipes/imake
+++ b/recipes/imake
@@ -1,0 +1,1 @@
+(imake :fetcher github :repo "tarsius/imake")


purcell:
It's very narrow and specific, but arguably potentially useful to others, so I'm happy to merge. You should probably `autoload` the `imake` command, though.

tarsius:
Ups.

tarsius:
I just amended the fix to `Initial import` (I sometimes do such horribly things in the early days of a package (though usually I have not published them by that time yet)). Will that cause issues for Melpa? Especially considering that only the hash changed, both author and committer dates are exactly as before.

purcell:
No, that should be okay.

PR #5045, opened 2017-10-07
Add recipe for js-auto-format-mode

ybiquitous:

Emacs minor mode for auto-formatting JavaScript code. After saving JavaScript file, its' code will be formatted automatically.

By default, this package uses [ESLint](https://eslint.org/), but it can customize any formatters such as [Prettier](https://github.com/prettier/prettier).

https://github.com/ybiquitous/js-auto-format-mode

Author and maintainer.

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/js-auto-format-mode
+++ b/recipes/js-auto-format-mode
@@ -1,0 +1,1 @@
+(js-auto-format-mode :fetcher github :repo "ybiquitous/js-auto-format-mode")


purcell:
Thanks! Quick feedback:
- Since you're using the GPL, please use the standard boilerplate throughout, such that the license summary is embedded in the file itself. (Try `M-x auto-insert` in an empty `.el` file to see what I mean.)
- You should add a `defgroup` form to declare the `js-auto-format` customisation group and to place it in the overall customisation hierarchy by setting an appropriate parent group.
- Building shell commands [like this](https://github.com/ybiquitous/js-auto-format-mode/blob/a9edd9ad3aca88cf8f5da681dc3977bd541cdd8e/js-auto-format-mode.el#L53-L58) isn't safe, since your quoting will break if filenames contain certain characters, e.g. quotes (and, generally, spaces, backslashes etc.). Use `shell-quote-argument` to perform escaping/quoting for you.
- Rather than add special handling for the node_modules path, which is likely to lead to surprises, it would be preferable to keep things simple: let the user explicitly set their command to `npx prettier` if they like, or they can use [add-node-modules-path](https://github.com/codesuki/add-node-modules-path). I like that this module is simpler and more general than some of the related packages already in MELPA, so I feel like you should embrace the generality!

Hope that helps. :-)

ybiquitous:
@purcell Thank you so much for detailed reviews! 😄 

I applied your all helpful feedbacks, then many diffs has been produced... 😓 
But I would be grateful if you would review code again! 🙏 

For details of changes, please see below:
https://github.com/ybiquitous/js-auto-format-mode/compare/a9edd9a...master

purcell:
Looks good, thanks! Merging now. Welcome to MELPA!

ybiquitous:
Thanks a lot! 😆

PR #5048, opened 2017-10-07
Add recipe for zel.

rudolfochrist:

`zel` is much like `recentf`. It provides a list of most used files, but based on _frecency_.
_Frecency_ takes how frequent a file has been used into consideration. 

https://github.com/rudolfochrist/zel

Author

### Checklist

### DIFF
--- a/recipes/zel
+++ b/recipes/zel
@@ -1,0 +1,1 @@
+(zel :repo "rudolfochrist/zel" :fetcher github)


purcell:
Thanks. This looks good, and I'll go ahead and merge it now. A couple of notes along the way:
- Providing version functions like `zel-version` is a little pointless these days: the numbers are essentially meaningless when distributed as packages, and have to by synchronised manually with tags. If you version-tag the file, then MELPA Stable will automatically add a corresponding `Package-Version` header.
- IMO you should provide a function which returns the full sorted file list, so that it can easily be used programmatically in a completion context, e.g. in the same way that Ivy and IDO's "virtual buffers" work.

rudolfochrist:
Thanks for merging this in and the quick code review :) 

- I guess you're right about version functions. I thought that this would help debugging and people could paste the output of those functions into bug tickets.
- You're again absolutely right. I already have `zel--frecent-file-paths` and `zel--frecent-file-paths-with-core` and will make them _public_. Thank you very much for spotting this.

PR #5052, opened 2017-10-09
add crystal-mode to recipes

brantou:

A minimal crystal mode for emacs, based on ruby-mode (of course)

https://github.com/crystal-lang-tools/emacs-crystal-mode

An enthusiastic user

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/emacs-crystal-mode
+++ b/recipes/emacs-crystal-mode
@@ -1,3 +1,3 @@
 (emacs-crystal-mode :repo "crystal-lang-tools/emacs-crystal-mode"
                :fetcher github
-               :files ("crystal-mode.el" "README.md"))
+               :files ("crystal-mode.el"))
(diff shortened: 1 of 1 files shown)


purcell:
Thanks! Quick feedback:
- You've missed some of the feedback from `package-lint`, which you indicated you had run. Please follow up on that.
- You've got duplicate `Author` headers etc. due to reproducing the ruby-mode headers. There should be only one set.
- `flycheck-crystal.el` will be built into this package by default, but it has different dependencies. You should add full headers & license etc. to that file, and then provide two separate recipes, each with a `:files` spec.

Hope that helps!

brantou:
@purcell Thanks! I fix your feedback, please check！

purcell:
You've still missed some `package-lint` feedback... :-)

brantou:
@purcell So sorry, I missed Package-Requires. This time I confirm that all the feedback from **package-lint** has been fixed， please check！

purcell:
Great, thanks! Sent you an additional small fix in https://github.com/crystal-lang-tools/emacs-crystal-mode/pull/5.

PR #5059, opened 2017-10-12
Add tickscript-mode

msherry:
Adds a major mode called tickscript-mode, for working with TICKscript
files (https://docs.influxdata.com/kapacitor/v1.3/tick/), which are used with
InfluxDB and Kapacitor.

https://github.com/msherry/tickscript-mode

I am the maintainer.

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/tickscript-mode
+++ b/recipes/tickscript-mode
@@ -1,0 +1,1 @@
+(tickscript-mode :fetcher github :repo "msherry/tickscript-mode")


purcell:
Thanks! You might consider adding `;; Package-Requires: ((emacs "24"))`, since you're using `prog-mode`, so this technically wouldn't work with an Emacs 23.4 that had `package.el` installed -- that scenario is increasingly unlikely, but you never know.

PR #5060, opened 2017-10-13
Add recipe for cwl-mode

tom-tan:

This is a major mode for editing [Common Workflow Language](http://www.commonwl.org).

https://github.com/tom-tan/cwl-mode

I am a maintainer of the package.

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/cwl-mode
+++ b/recipes/cwl-mode
@@ -1,0 +1,1 @@
+(cwl-mode :repo "tom-tan/cwl-mode" :fetcher github)


purcell:
Thanks. Quick feedback:
- [This line](https://github.com/tom-tan/cwl-mode/blob/a62e038fd7cb7c5a0ee3946e5ddae41af96fcc5e/cwl-mode.el#L84) doesn't do anything, I think.
- The flycheck portion is a bit odd and out of place. First of all, it requires that anyone who uses this major mode must have flycheck installed, which condition is generally not good practice. And secondly, flycheck is enabled by a confusingly-named function `cwl-mode-setup`, which then only enables other checkers. In this case I would probably remove [this line](https://github.com/tom-tan/cwl-mode/blob/a62e038fd7cb7c5a0ee3946e5ddae41af96fcc5e/cwl-mode.el#L92), drop `flycheck` from the `Package-Requires` list, and convert the setup function to an `eval-after-load 'flycheck` block.

tom-tan:
Thank you for the feedback!

> - This line doesn't do anything, I think.

I removed it.

> In this case I would probably remove this line, drop flycheck from the Package-Requires list, and convert the setup function to an eval-after-load 'flycheck block.

I removed flycheck from Package-Requires and used `with-eval-after-load` instead of `cwl-mode-setup`.
I also updated the document for it.

purcell:
Thanks. `with-eval-after-load` is only available in very recent Emacs versions: you should either use `eval-after-load` instead (and quote the contained form) or specific an Emacs version dependency in the `Package-Requires` header.

purcell:
P.S. package-lint would have told you this. :-)

tom-tan:
Thank you for your advice.

I added `(emacs "24.4")` for Package-Requires.

Also I installed flycheck-package and now I can sure my extension is automatically checked by package-lint. :)

purcell:
Great, thanks.

tom-tan:
Thanks!

PR #5062, opened 2017-10-15
Remove org-chinese-utils

tumashu:


purcell:
Thanks

tarsius:
@tumashu Am I right to assume that like `ox-latex-chinese` this is an obsoleted project and one should use org2ctex instead?

tumashu:
@tarsius please wait two month, after ox-latex-chinese users move to org2ctex, i will remove ox-latex-chinese

PR #5064, opened 2017-10-16
Add recipe for quick-shell-keybind

eyeinsky:

Interactively make a keybinding to run a sequence of commands in a
shell buffer.

<https://github.com/eyeinsky/quick-shell-keybind>

I'm the author.

None needed

### Checklist

Please confirm with x:

-   [x] The package is released under a GPL-Compatible Free Software License.
-   [X] I've read [CONTRIBUTING.md](https://github.com/melpa/melpa/blob/master/CONTRIBUTING.md)
-   [X] I've used the latest version of package-lint to check for packaging issues, and addressed its feedback
-   [X] My elisp byte-compiles cleanly (needed to add group to defcustom)
-   [X] M-x checkdoc is happy with my docstrings
-   [X] I've built and installed the package using the instructions in [CONTRIBUTING.md](https://github.com/melpa/melpa/blob/master/CONTRIBUTING.md)

### DIFF
--- a/recipes/quick-shell-keybind
+++ b/recipes/quick-shell-keybind
@@ -1,0 +1,3 @@
+(quick-shell-keybind
+ :fetcher github
+ :repo "eyeinsky/quick-shell-keybind")


purcell:
Looks good. Quick feedback:
- Don't just add extra custom vars into the `'comint` group: make your own group with `defgroup`, and place it in an appropriate parent group (which probably isn't comint, to be honest).
- Add named args to `quick-shell-keybind` and use `interactive` to call `read-key-sequence` etc. for you. This is more idiomatic, makes the function usable programmatically, and will eliminate some boilerplate.
- You should add an `;;;###autoload` for that command too.

eyeinsky:
I've now pushed the fixes: https://github.com/eyeinsky/quick-shell-keybind/commit/86a19f09262c1b442f783680d2dc947c9bdfd3cf, https://github.com/eyeinsky/quick-shell-keybind/commit/7a68c44fa4a4c1aade0da0fb1a726f7ccf52f5d7, https://github.com/eyeinsky/quick-shell-keybind/commit/5a5b0750eacaac8e82e3855c66d20bf2a38645e5.

Let me know if something could be better still! I.e I couldn't find a way to use interactive code for the key, since the other two arguments need lisp expressions to get them. Also, not sure if `tools` is a good parent group.

I will also squash the fixups when they are good enough!

purcell:
Looks good to me! :-) Merged.

purcell:
(Also: welcome to MELPA!)

eyeinsky:
@purcell Just one more question: from now on if I would like to have code review then it's up to me to find a volunteer? (Since I'm quite new to elisp then even the current pull request taught me a lot by prompting me to read up on what `interactive` actually does and how it works together with the function arguments.)

purcell:
Yeah, you'd have to find a volunteer, sorry, but if you want to @ me on your own commits or PRs then I'll look if I have time. I hear that the Emacs IRC channel is helpful too, and the Emacs subreddit is pretty active.

eyeinsky:
Got it!

PR #5065, opened 2017-10-16
add go-gen-test recipe

s-kostyaev:

Generating go tests from emacs.

https://github.com/s-kostyaev/go-gen-test

I am the author.

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/go-gen-test
+++ b/recipes/go-gen-test
@@ -1,1 +1,1 @@
-(go-gen-test :repo "s-kostyaev/go-gen-test" :fetcher github)
+(go-gen-test :repo "s-kostyaev/go-gen-test" :fetcher github :files (:defaults "go-gen-test.el"))
--- a/recipes/go-gen-test
+++ b/recipes/go-gen-test
@@ -1,0 +1,1 @@
+(go-gen-test :repo "s-kostyaev/go-gen-test" :fetcher github)


s-kostyaev:
How to prevent `(:exclude .dir-locals.el test.el tests.el *-test.el *-tests.el)` ?

purcell:
Thanks!

- The `:files` spec is unnecessary here, so please remove it.
- Use `(interactive "r")` to more easily get the region extents in `go-gen-test-dwim`
- The way to decide whether the user wants to use the region is with `use-region-p`, not `region-active-p`.
- Don't [build commands with `format` and string concatenation](https://github.com/s-kostyaev/go-gen-test/blob/c1e7e5ce99f7bf66cc8bfef818b4437f5e7406cb/go-gen-test.el#L122-L124) unless you're also using `shell-quote-argument` -- if `buffer-file-name` contains spaces or quotes here, you're in trouble.

Looks good otherwise. :-)

s-kostyaev:
> The :files spec is unnecessary here, so please remove it.
Then this project can't build because of filename is `go-gen-test.el` which one matched by `*-test.el`
Other things I will fix soon. Thanks for review.

purcell:
> Then this project can't build because of filename is go-gen-test.el

Doh, right. Sorry! Merging now.

PR #5066, opened 2017-10-17
Add lsp-ocaml recipe

anmonteiro:

OCaml/Reason support for lsp-mode using [ocaml-language-server](https://github.com/freebroccolo/ocaml-language-server).

https://github.com/emacs-lsp/lsp-ocaml

I'm the maintainer.

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/lsp-ocaml
+++ b/recipes/lsp-ocaml
@@ -1,0 +1,1 @@
+(lsp-ocaml :repo "emacs-lsp/lsp-ocaml" :fetcher github)


purcell:
Thanks.
- Don't check `(featurep 'projectile)`. Instead check `projectile-mode` with `bound-and-true-p`: you shouldn't use projectile unless the user has enabled it.
- You're using `seq-some`, but that's not available in every Emacs > 24. Either specify a higher Emacs version, or declare a package dependency on the version of `seq` that is available in GNU ELPA.

anmonteiro:
@purcell thanks for the review. Just addressed your feedback in https://github.com/emacs-lsp/lsp-ocaml/commit/aa4240ef24b9eea578361c9d88b52b0138a925bf

purcell:
Awesome, thanks!

PR #5067, opened 2017-10-17
Add org-section-numbers recipe

marcowahl:

Display natural section numbers for orgees [=Org subtrees].

This PR is due to one user who proposed this package could  be added to melpa.

https://github.com/marcowahl/org-section-numbers

 maintainer,contributor,user

 **None needed** 

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/org-section-numbers
+++ b/recipes/org-section-numbers
@@ -1,0 +1,1 @@
+(org-section-numbers :repo "marcowahl/org-section-numbers" :fetcher github)


alphapapa:
There are a couple of significant issues:

1.  Special (global) variables are used without being defined.  Every function appears to be made a closure by putting them all in a single, top-level `let`.
2.  Commands that create new headings in Org are not hooked into, so after the numbers are toggled on, new headings will not have them.

This is a very cool tool, but I think it needs some work before it should be published on MELPA.

marcowahl:
To 1.: Actually I thought the closure asserts that the variables can only be accessed by the functions defined within the `let`.
Am I missing something?  (This could easily be the case, I guess.)

marcowahl:
To 2.: Okay, I think I understand.  Up to now the package is supposed to be a temporary overlay feature.  I don't know how hard it is to realize a reliable dynamic update of the numbering.

Maybe I try some day.  Anyone feel free to jump in any time.


Thanks for your feedback alphapapa!

alphapapa:
> To 1.: Actually I thought the closure asserts that the variables can only be accessed by the functions defined within the let.  Am I missing something? (This could easily be the case, I guess.)

To be honest, I'm not sure how this would generally work:
```el
(let ((a 1))
  (defun fn1 () (incf a))
  (defun fn2 () (decf a)))
```
It seems like both functions should get their own copies of the variable, which would defeat the intended purpose.  And I have no idea how it would function with regard to garbage collection.

Anyway, the idiomatic way is, e.g.:
```el
(defvar a 1 "The counter.")
(defun fn1 () (incf a))
(defun fn2 () (decf a))
```
This also allows other code to access the variable.

> To 2.: Okay, I think I understand. Up to now the package is supposed to be a temporary overlay feature. I don't know how hard it is to realize a reliable dynamic update of the numbering.

It wouldn't be hard to advise a few key Org functions, but that wouldn't handle all cases, e.g. when simply typing `* foo` by hand.

It would probably not be *too* hard to add it to Org directly.  You could probably look at the plain list code for some ideas, since it would operate in a similar fashion to numbered/lettered plain lists, but on headings instead.

Have you ever presented this idea on the org-mode mailing list?  It seems like an interesting feature, especially since it would present a live view of how exported document headings are numbered.  I think there would be interest in the feature.

marcowahl:
To 1.: As far as I understand your example
```
(let ((a 1))
  (defun fn1 () (incf a))
  (defun fn2 () (decf a)))
```
is a reasonable way to realize one integer that can be incremented and decremented.

To 2.: I presented the feature to the list and some expressed their interest (and some found weaknesses). I think this feature is not too important for anyone.

FWIW I proposed package org-section-numbers to MELPA since one user said he'd like to see the package at MELPA.

If I find some muse I try to evolve the program.

I think we can close this PR now.


Thanks for your energy and time!

marcowahl:
Ah, I just realize that I can close this PR.

PR #5068, opened 2017-10-17
Adding ja-network to elpa repository

jamiguet:
### Provides hooks for handling intermittent network connectivity

Defines network-up-hook and network-down-hook which get invoked when no specified interfaces are add up on the host. It also provides a have-network-p which returns nil if none of the listed interfaces are up. 

The package uses customised variables to enable the same configuration to run on several machines.

https://github.com/jamiguet/ja-network

**None needed** 

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/ja-network
+++ b/recipes/ja-network
@@ -1,0 +1,1 @@
+(ja-network :repo "jamiguet/ja-network" :fetcher github)


purcell:
Thanks. Quick feedback:

- You're defining a number of symbols (e.g. network-up-hook) without the proper prefix for your package. `package-lint` should have warned you about this.
- Use `let` for [this code](eac5567e2a47be1272ed73b19c43025a457fa416) - defining variables inside a block isn't a thing in elisp. :-)
- You've got hooks, and you've got a way of enabling functionality... what you've got is a global minor mode, so turn it into one, e.g. `network-monitor-mode`. If you believe this is useful to other people, why publish with the `ja-` prefix?
- [Typos](https://github.com/jamiguet/ja-network/blob/bc027a993020e1b3771950d3a784686d21737447/ja-network.el#L48)
- Your `Commentary` section is malformatted: too many `;;;`.
- Use something like `cl-find` [here](https://github.com/jamiguet/ja-network/blob/bc027a993020e1b3771950d3a784686d21737447/ja-network.el#L69-L75)
- `network-interface-list` isn't a function in my Emacs... where is that coming from?

Hope that helps!x

jamiguet:
Thanks a lot for your feedback its my first contribution so it is very welcome.
I reply here to the simple comments.
I will re-factor the code to turn it into a global-minor-mode. So please bear with me as it may take a little bit of time to do so.

*  package-lint-20171006.1846 reports No issues except the snap-shot version for org.
* Using let already in the first fix
* I read about them, I agree its much cleaner. I will re-factor to user a global minor, please bear with me
* Typos, searched and destroyed ;)
* Commentary fixed, also a better suited commentary text.
* Using cl-find as advised
* network-interface-list is a built-in function in `C source code'. As per what apropos says on Emacs 24.5.1 (x86_64-pc-linux-gnu, GTK+ Version 3.22.7) of 2017-02-10 on lcy01-09, modified by Debian


Thanks for the feedback once again I'll get cracking to fix it.

purcell:
> network-interface-list is a built-in function in `C source code'.

Ah yes, found it now - sorry for the noise! :-)

jamiguet:
As part of packaging the functionality as a global-minor-mode it will change name to network-mode.
Closing pull request.

purcell:
`network-mode` is a bit vague. Can I suggest `network-watch` instead? `-mode` suffixes are generally only advisable in package names when a major mode is provided.

jamiguet:
Will adjust accordingly.
Thanks for the quick feedback.

PR #5072, opened 2017-10-18
Add recipe for docker popups

olymk2:

Docker command wrapper which exposes functionality via magit poups

https://github.com/olymk2/emacs-docker-compose

Maintainer

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/docker-popups
+++ b/recipes/docker-popups
@@ -1,0 +1,1 @@
+(docker-popups :fetcher github :repo "olymk2/emacs-docker-compose")


olymk2:
Let me know if there are any glaring issues, not packaged a multi file project so bit unsure about ert test files and other files related to the project but not required to make things run.

purcell:
Thanks! Quick feedback:
- You've got package dependencies listed in individual files, e.g. `dc-helm`, but none for the package overall (according to `dc-popups-pkg.el`). Nobody could install this package on a fresh system and have all its dependencies automatically installed.
- The package is `dc-popups`, but there are many files which are not named `dc-popups*`: they should *all* follow that pattern.
- In particular, ert-core.el & co definitely don't belong in the package, but they will be included by the provided recipe. Consider moving test code into a `tests` subdirectory
- Some code just uses the `dc-` prefix for arbitrary names, e.g. `dc-str-addresses`: this does not match the package prefix, and is too short.
- Opinionated dependencies such as `helm` shouldn't be built into the package. Provide the `helm` portion as a separate package which depends on the core package.

Hope that helps!

olymk2:
Okay thanks for the feed back, is there some info on the -pkg file format nd what should be in that file or a good example i can use ?

I will update based on your comments

purcell:
> is there some info on the -pkg file format nd what should be in that file or a good example i can use ?

I'd suggest not even using a `-pkg` file because MELPA doesn't need it: just put the correct package description, keywords, `Commentary:` section, `Package-Requires:` header etc in the `.el` file whose name matches the package name.

olymk2:
I have done most of those changes, the one left todo is remove helm but I am unsure what to replace it with, I am basically after a way to let the user select from a list of options I found ido but I am not sure if that's replacing one opinionated dependency for another.

is there like a built in I can use to provide a list of options for a user to select from, or at least an auto complete text entry box ?

Silex:
Ah interesting.

@olymk2: I'm the author of https://github.com/Silex/docker.el, I had a docker-compose branch brewing but never had time to finish it (https://github.com/Silex/docker.el/pull/39). Maybe we'd collaborate?

olymk2:
@Silex happy to help collaborate, mint is not about docker file its all about making an interface like magit to start up my current projects containers, or to send a command to a specific container or even rebuild containers after editing the containers.

so to launch containers for my current project i can do [SPC] d u u to bring up all my compose containers if that makes sense.

if you think this maybe useful we can put this into one project, i will give your a try shortly this was also done partly as a learning experience :)

Silex:
@olymk2: ah, ok then it's in the same spirit as my docker-compose branch, except you actually had time to work on it :-)

My package is about managing images/containers/volumes/networks from emacs (it uses magit-popups and tabulated-list), and `docker-machine`, `docker-compose` and `docker-stack` were the logical next steps but I got swamped at work.

Let's discuss this at https://github.com/Silex/docker.el/pull/39 to avoid polluting this issue if you are ok.

olymk2:
@Silex sure will continue in your issue

@purcell i realise not ido is built into emacs so i will use that and allow overriding

purcell:
> i realise not ido is built into emacs so i will use that and allow overriding

Better still, just use the built-in `completing-read` -- every popular completion library has a way to override that. Choosing any other default is going to annoy someone. :-)

olymk2:
@purcell thanks for that changed to use completing-read which looks the same as ido anyway :)

purcell:
Great! But then `dc-popups-helm` should have its own recipe, and depend on `dc-popups` (which then would not have to depend directly on `helm`.

olymk2:
ah yes i will separate that out hoppefuly this weekend if i get a chance

tarsius:
Friendly ping!

PR #5073, opened 2017-10-19
Add jetbrains recipe

zonuexe:

Emacs-JetBrains IDE bridge

https://github.com/emacs-php/jetbrains.el

I am the author.

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/jetbrains
+++ b/recipes/jetbrains
@@ -1,0 +1,1 @@
+(jetbrains :fetcher github :repo "emacs-php/jetbrains.el")


purcell:
Thanks.
- Your code says "this is not part of gnu emacs", but the licence text talks about the terms of "gnu emacs". I think it should say "this software" instead? Please follow the standard elisp boilerplate, e.g. that inserted by `M-x auto-insert`.
- The parent group `'tool` should be `'tools`, I think.
- Use `make-variable-buffer-local` [here](https://github.com/emacs-php/jetbrains.el/blob/fe14e984152121cb49bf3366e8c222f7861002f6/jetbrains.el#L107) - `make-local-variable` does something different. Alternatively, require at least Emacs 24.3 and then use `defvar-local`.
- You're using `cl-` functions but you don't require `cl-lib` or Emacs >= 24.3. `package-lint` should have warned you about this - did you really run it?

Hope that helps!

zonuexe:
@purcell Thank you for your review, and I fixed them.

I regularly use `flycheck-package`, but unfortunately it seems to miss this pattern.
`package-lint-current-buffer` is the same as that and reports "No issues found."

purcell:
`flycheck-package` generally only runs `package-lint` if it finds some signs that the current file is a package, so they don't always match up. I'll dig into why `package-lint` wasn't working for you - thanks!

PR #5081, opened 2017-10-24
Add recipe for flycheck-nimsuggest

yuutayamada:

This is a flycheck backend for Nim language. It uses nimsuggest (kinda IDE feature support tool) in 
backend. (there is a flycheck backend for Nim called flycheck-nim, but it uses nim's compiler command
`nim check`)

https://github.com/yuutayamada/flycheck-nimsuggest

I'm this package's maintainer and also nim-mode's

currently flycheck-nimsuggest.el is included in nim-mode repository, but I will
change the dependencies as soon as I saw the MELPA's change.

The main reason of such change is that for less dependencies on the nim-mode repository and
from Emacs 26, somehow flymake is re-written, so I would include its backend instead of flycheck-nimsuggest. (FYI: https://lists.gnu.org/archive/html/emacs-devel/2017-09/msg00953.html)

### Checklist

Please confirm with `x`:

Thank you for reviewing!

EDIT: add a link for new flymake and correct wording

### DIFF
--- a/recipes/flycheck-nimsuggest
+++ b/recipes/flycheck-nimsuggest
@@ -1,0 +1,1 @@
+(flycheck-nimsuggest :fetcher github :repo "yuutayamada/flycheck-nimsuggest")


purcell:
Thanks. Mostly looks great, but 2 points of feedback:

- You should add `(cl-lib "0.6")` to your `Package-Requires`, or instead depend on `(emacs "24.3")`
- Don't autoload [this code](https://github.com/yuutayamada/flycheck-nimsuggest/blob/43604c47f8f446e939cccb6a61b54c62b75fea3f/flycheck-nimsuggest.el#L120-L133) - it has the effect of auto-defining the checker when the package is installed, which is bad practice. Instead, users should either call your `flycheck-nimsuggest-setup` function as needed, or add their own `(with-eval-after-load 'flycheck (require 'flycheck-nimsuggest))` form.

yuutayamada:
Thank you for the feedback. I've fixed both.

purcell:
Super, thanks - merged!

PR #5084, opened 2017-10-26
Add recipe for eacl

redguardtoo:

eacl means Emacs auto complete line(s). It provides several commands to auto complete line(s) by grepping current project.

https://github.com/redguardtoo/eacl

I'm the only maintainer.

None need

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/eacl
+++ b/recipes/eacl
@@ -1,0 +1,1 @@
+(eacl :fetcher github :repo "redguardtoo/eacl")


purcell:
Thanks. Feedback:
- What's going on in `eacl-encode`? There are all sorts of ways that function can break. I think you should instead be using `regexp-quote` and then `shell-quote-argument`.
- Once again there's filename interpolation into strings using `'%s'`: files/dirs with quotes will break this. Use `shell-quote-argument`.
- There's already `grep-find-ignored-files` and `grep-find-ignored-directories` -- please don't reinvent your own list.
- Since you're using `ivy`, you could re-use some of what's in `counsel`: it has `counsel-grep`/`counsel-grep-function`/`counsel-ag-function` etc. Counsel also has code for finding the current project, so there's no need to reinvent that too.

Hope that helps!

redguardtoo:
I rewrote eacl-encode to use `regexp-quote`, shell-quote-argument is not necessary because inside `eacl-encode` I replace the quote characters with more generic `.` Apply the `shell-quote-argument` on encoded string will mess up the regex.

There is no file/dirs used in grep cli because I just ask grep to grep current directory.

`eacl` need support emacs v24 but `grep-find-ignored-files` and `grep-find-ignored-directories` are only defined in 25+ (I tested with Emacs 24.3, 24.4, 24.5)

I read the code of counsel/ivy. I used `grep` for only specific purpose (creating matching multi-line candidates). So my grep program setup **is totally differently from `counsel`'s assumption. For example, I setup the grep to output only matched string, without filename, without line number. The candidates are separated by NULL characters instead of line feeds**. 

The only project related variable I found in counsel is `counsel-git-grep-projects-alist`, It's used by `counsel-git-grep` only. Currently I work on mulitple projects managed by git or perforce. So it's better I don't modify `counsel-git-grep-project-alist`.

purcell:
> shell-quote-argument is not necessary because inside eacl-encode I replace the quote characters with more generic .

Those are 2 different things: first you should build a regexp using regexp syntax, then `shell-quote-argument` will add extra `\` where necessary. Right now your code is adding extra `\` manually because you know the regexp will be inserted into a command string, and this is fundamentally an error-prone approach.

> eacl need support Emacs v24 but grep-find-ignored-files and grep-find-ignored-directories are only defined in 25+ (I tested with Emacs 24.3, 24.4, 24.5)

Not true:
```
ELISP> (emacs-version)
"GNU Emacs 23.3.1 (x86_64-apple-darwin, NS apple-appkit-1038.35)
 of 2011-03-10 on black.porkrind.org"

ELISP> (require 'grep)
grep

ELISP> grep-find-ignored-files
(".#*" "*.o" "*~" "*.bin" "*.lbin" "*.so" "*.a" "*.ln" "*.blg" "*.bbl" "*.elc" "*.lof" ...)

ELISP> grep-find-ignored-directories
("SCCS" "RCS" "CVS" "MCVS" ".svn" ".git" ".hg" ".bzr" "_MTN" "_darcs" "{arch}")
```

So those variables are available, and you should use them.

Thanks for the other details, that makes more sense now. For finding project roots, consider making the function that defines this configurable, so that Projectile users can customise the method used.

redguardtoo:
Thanks. I fixed issues you mentioned.

purcell:
Nice! Merged.

PR #5085, opened 2017-10-26
Add recipe for 'iter2' package.

doublep:

`iter2` is a fully compatible reimplementation of built-in `generator`
package.  It provides `iter2-defun` and `iter2-lambda` forms that can
be used instead of `iter-defun` and `iter-lambda`.  All other
functions and macros (e.g. `iter-yield`, `iter-next`) are
intentionally not duplicated: just use the ones from the original
package.

The main reasons for writing this was that 1) built-in can be extremely slow to convert long functions (~10 seconds for one 900 line megafunction) and 2) Emacs upstream is unresponsive to bug reports.

https://github.com/doublep/iter2

I'm the only author and maintainer.

**None needed.**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/iter2
+++ b/recipes/iter2
@@ -1,0 +1,2 @@
+(iter2 :repo "doublep/iter2"
+       :fetcher github)


purcell:
Thanks. [That's quite a macro](https://github.com/doublep/iter2/blob/master/iter2.el#L185-L547)...! :-)

PR #5086, opened 2017-10-26
Add recipe for network-watch

jamiguet:

Package provides global minor mode for dealing with intermittent networking.

https://github.com/jamiguet/network-watch

Maintainer

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/ja-network
+++ b/recipes/ja-network
@@ -1,1 +1,0 @@
-(ja-network :repo "jamiguet/ja-network" :fetcher github)
--- a/recipes/ja-network
+++ b/recipes/ja-network
@@ -1,0 +1,1 @@
+(ja-network :repo "jamiguet/ja-network" :fetcher github)
--- a/recipes/network-watch
+++ b/recipes/network-watch
@@ -1,0 +1,1 @@
+(network-watch :repo "jamiguet/network-watch" :fetcher github)


purcell:
This needs some work still, sorry.
- You say above you're run `package-lint`, but I don't think you have: you've got a bunch of symbols which don't begin with `network-watch-`, which they all must. So use `network-watch-up-hook` etc.
- There's no point altering `after-init-hook`: what if someone loads this library after their Emacs is already started? You've missed my point about the global minor mode: when the mode is enabled, it should start watching the network and invoking the up/down hooks as necessary, and when the mode is disabled it should stop.
- Mapping interfaces per machine is unnecessary. This library doesn't need to know about the machine name to do its job, so it shouldn't.
- Don't call `customize-save-variable` directly: you should never decide for the user that their `custom.el` should be updated.
- Don't use `defvar` to [declare local variables]: you're acually declaring a global variable there. Use `let`, in general. In this particular case there are functions like and `mapcar`, `append` you can use to even avoid needing a local variable. I recommend reading "Practical Common Lisp" (available free online) to learn lisp style.
- Lots of typos: "intermitent", "refersh reate" etc.
- Based on its docstring `network-watch-p` should be called `network-watch-active-p`.
- `network-watch` is a very unclear variable name.
- Don't place the trailing `)` on its own line: that's bad lisp style. :-)

Hope that helps!

jamiguet:
Thanks for the feedback I'll get cracking on those changes and further
education.

Hooks appear to not be reported by package lint I tweak it further.

Thanks again

jamiguet:
Apologies for the delay, life got on the way of hacking.

I believe all the comments have been addressed.
Thank you for your feedback I have learned quite a bit in the process.

Let me know if any other elements need improving.

purcell:
Thanks. Sent you a bunch of fixes in https://github.com/jamiguet/network-watch/pull/1.

purcell:
Merged now - thanks!

PR #5087, opened 2017-10-27
Fix nix-mode

leezu:
Only nix-mode.el was included. But nix-mode relies also on other files such as nix-format.el which are all part of the repository.

### DIFF
--- a/recipes/nix-mode
+++ b/recipes/nix-mode
@@ -1,2 +1,1 @@
-(nix-mode :repo "NixOS/nix-mode" :fetcher github
-          :files ("nix-mode.el"))
+(nix-mode :repo "NixOS/nix-mode" :fetcher github)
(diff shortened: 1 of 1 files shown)


purcell:
Thanks.
- `nix-mode` doesn't rely on `nix-format`: it soft-requires it
- The other files in the repo have various non-listed package depdencies, e.g. on `company`, `mmm-mode`, and should therefore be packaged separately.

It would be acceptable to instead change this recipe so that the `:files` spec uses the `:defaults` but `:excludes` both `nix-company.el` and `nix-mode-mmm.el`. Would you like to make that change? (See the README and examples in other recipes.)

leezu:
Thanks @purcell ! I added an exclude statement for "nix-company.el" and "nix-mode-mmm.el". Please let me know in case of any problems.

purcell:
Looks good, thanks.

PR #5091, opened 2017-10-30
Add  recipe for David's theme

crankyadmin:

Package provides David's theme

https://github.com/crankyadmin/davids-theme

maintainer

**None needed** 

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/davids-theme
+++ b/recipes/davids-theme
@@ -1,0 +1,1 @@
+(davids-theme :fetcher github :repo "crankyadmin/davids-theme")


purcell:
Thanks! Quick feedback:

- The description line is duplicated. Keep the second line (since "Emacs" is redundant) and indent it left.
- It's bogus to call `display-color-cells` at load-time: what if the current window is in a terminal frame at that point? This is what the various face `class` options are for.
- The options in your `Local Variables` block are personal preferences and as a rule of thumb therefore belong in your config rather than in the `.el` file, e.g. you can do [this sort of thing for `rainbow-mode`](https://github.com/purcell/emacs.d/blob/83525d66fa195901e8881a62ffb485d212a39bed/lisp/init-lisp.el#L232-L244).

PR #5092, opened 2017-10-30
Add recipe for fish-completion

Ambrevar:

Add fish-based completion to Emacs and allow Eshell to fall back on it.

https://github.com/Ambrevar/emacs-fish-completion

I am the author.

I am not sure about what the best practice is when it comes to functions setting external global variables, in this case `eshell-default-completion-function`.
Have a look at `fish-completion-eshell-toggle`: I suspect there is a much more idiomatic / cleaner way to do this.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/fish-completion
+++ b/recipes/fish-completion
@@ -1,0 +1,3 @@
+(fish-completion
+ :repo "Ambrevar/emacs-fish-completion"
+ :fetcher github)


purcell:
Thanks! Quick notes:
- Don't use `%s` to interpolate strings directly into command strings: they might contain quotes etc. Use `shell-quote-argument` to escape them for you first.
- When you've got functions for `fish-completion-eshell-toggle-globally` and `fish-completion-eshell-toggle`, what you really want is a local and a global minor mode. :-)
- No need to mention `Emacs` in the description line of the package: all elisp is for Emacs, so this is redundant.

Hope that helps!

Ambrevar:
1. Fixed, thanks for the catch.
2. I wasn't sure whether declaring a minor mode wasn't too overkill for a single variable but it turns out it makes everything simpler.  Thanks!
3. By "Emacs" I actually meant `pcomplete`.  Fixed.

purcell:
Great! Minor modes often simplify things a lot.

Merged now - welcome to MELPA!

Ambrevar:
Thanks!

PR #5094, opened 2017-11-01
Add recipe for company-eshell-autosuggest.

dieggsy:

Provides history suggestions in the style of fish-shell.

See [here](http://whyarethingsthewaytheyare.com/fishlike-autosuggestions-in-eshell/) for visual example.

https://github.com/dieggsy/company-eshell-autosuggest

Maintainer

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/company-eshell-autosuggest
+++ b/recipes/company-eshell-autosuggest
@@ -1,0 +1,1 @@
+(company-eshell-autosuggest :fetcher github :repo "dieggsy/company-eshell-autosuggest")


purcell:
Thanks!

This needs `company`, but doesn't declare the necessary package dependency, and doesn't mention it in the package description. Also, since it's a `company` plugin, it would be better to retain the `company-` prefix for the package name and its symbols.

Ambrevar:
I think the name should not contain `company`. What we want here is autosuggestion in Eshell: company is an implementation detail. Should company get deprecated in favour of X, you'd switch to X but the package would remain the same.

purcell:
I disagree: users have to enable `company` to use this package, therefore it doesn't simply use `company` as an implementation details, it actually requires the user to install, use and enable `company`, otherwise they can't use it.

But in any case, nothing about this completion needs to be specific to any particular completion framework: you can simply provide a `completion-at-point` function that the user can add into their `completion-at-point-functions` list, and then it will work seamlessly with almost any completion framework, or with stock Emacs.

dieggsy:
@purcell Point taken on the extent that company is required, I'll consider renaming.

And well, this isn't really a conventional completion framework - it only provides one candidate per any unique input, and the whole purpose of the package is to display the suggestion on the command line like fish shell does. Providing this single candidate through `completion-at-point` might be sort of useless or limited use because of this. I'd say it at least depends on a front-end like `autocomplete` or `company-mode` provide.

In any case, I'll add the company dependency as soon as I get the chance.

purcell:
I think the `completion-at-point-functions` can be defined quite flexibly, so that multiple functions can contribute completions, so you might still be able to get the effect you want. It's certainly the most widely usable solution, and with few moving parts.

dieggsy:
I'm not entirely convinced - as I understand it, `completion-at-point-functions` offer completion on the call of a command or press of a key. This package attempts to automatically suggest input as you type in the command line. The automatic part is pretty key, as we're emulating the fish shell. The reason I chose company for this is because it already has the idle timer functionality built-in (as well as the front-end) such that it will do just that with only a couple more functions - on some idle delay, automatically provide the last similar command. I think the more "general purpose" solution for this would actually require writing my own idle timer code to provide that completion as the user types.

dieggsy:
Package has been renamed.

dieggsy:
Oh, the company dependency was added too btw.

purcell:
Yep, looks good. Thanks for your patience. :-)

PR #5095, opened 2017-11-01
Add recipe for mu4e-jump-to-list

wavexx:

``mu4e-jump-to-list`` allows to select and view mailing lists automatically using existing ``List-ID`` headers in your mu database. Requires mu4e (http://www.djcbsoftware.nl/code/mu/).

https://github.com/wavexx/mu4e-jump-to-list.el

Author.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/mu4e-jump-to-list
+++ b/recipes/mu4e-jump-to-list
@@ -1,0 +1,2 @@
+(mu4e-jump-to-list :repo "wavexx/mu4e-jump-to-list.el"
+                   :fetcher github)


purcell:
Thanks. Merging now.

BTW, rather than binding keys in another package's mode map directly, it would be preferable for you to define a tiny minor mode with its own mao.

wavexx:
On Sat, Nov 04 2017, Steve Purcell wrote:
> BTW, rather than binding keys in another package's mode map directly,
> it would be preferable for you to define a tiny minor mode with its
> own mao.

Normally yes, I do this. It felt overkill here as I'd have to introduce
a global minor mode (I add the binding in two minor modes) _or_ avoid
the bindings entirely.

I'm not too worried about this, this was already posted in the mu4e ML.
I'll see the feedback. I'm more prone to remove the bindings than
introduce a minor mode.

Thanks.

PR #5096, opened 2017-11-01
Add indent-info recipe

terlar:

Show indentation information in status bar.

https://github.com/terlar/indent-info.el

Author.

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/indent-info
+++ b/recipes/indent-info
@@ -1,0 +1,1 @@
+(indent-info :fetcher github :repo "terlar/indent-info.el")


terlar:
Unfortunately I could not check the box for `M-x checkdock` as it complains about the usage of `indent-info-mode' in the documentation string, as it wants to be more specific. It wants to be prefixed with function, command, variable, option or symbol. But I am not sure what mode counts as.

purcell:
Thanks - looks handy! Quick feedback:
- Please remove the `:files` spec from the recipe, which is unnecessary here.
- [Changing mode-line-position at the top level of the file](https://github.com/terlar/indent-info.el/blob/28045d16e192b61e1ef4af7626278d89592f9ca6/indent-info.el#L117-L121) seems inappropriate. Presumably it should happen when the mode is enabled/disabled instead.
- [redraw-frame](https://github.com/terlar/indent-info.el/blob/28045d16e192b61e1ef4af7626278d89592f9ca6/indent-info.el#L142) should probably be `force-mode-line-update` instead
- [`(if (not x) ...)`](https://github.com/terlar/indent-info.el/blob/28045d16e192b61e1ef4af7626278d89592f9ca6/indent-info.el#L170-L171) is better written `(unless x ...)`

terlar:
Thank you for the feedback. I have fixed the things you mentioned in the feedback. Found a better way to add the indent info to `mode-line-position` and added it inside the mode toggle part, I hope it is the right way.

purcell:
Looks great! Two more small points (though I'll merge now anyway):
- `indent-info-mode-enable` shouldn't be autoloaded: users shouldn't be calling it directly anyway, when they can just use `indent-info-mode` directly instead.
- You should remove the mode line entry when disabling the minor mode, so your `when` should be an `if`.

Welcome to MELPA!

terlar:
Thank you, I have addressed those issues.
Glad to finally do some contribution to the community.

PR #5097, opened 2017-11-02
Add emojify-logos recipe

mxgoldstein:

This package adds logo icons for various programming languages and tools to emojify.el

https://github.com/mxgoldstein/emojify-logos

The maintainer

**None needed**
### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/emojify-logos
+++ b/recipes/emojify-logos
@@ -1,0 +1,1 @@
+(emojify-logos :fetcher github :repo "mxgoldstein/emojify-logos" :files (:defaults "logos"))


purcell:
Thanks! Much of the code here is about downloading the logos, but you should really just include them in the package and avoid all that hassle. We'll rebuild fresh packages when the logos change.

mxgoldstein:
Should I just copy the files directly from the package every time it's run?

purcell:
Yes: `:files (:defaults "logo")` will cause the package to contain the same directory structure as your repository, so you can just use `load-file-name` to find the `logo` directory in both cases.

mxgoldstein:
Adapted it.

purcell:
Thanks, merged now - sorry for the slow response: work has been busy!

PR #5100, opened 2017-11-04
Axiom-environment repository changed to git

pdo:
The axiom-environment repository has changed from mercurial to git, but still uses the same account on bitbucket.  This affects packages: axiom-environment, ob-axiom & company-axiom.

Thanks,
Paul


purcell:
Great, thanks.

PR #5101, opened 2017-11-04
feat: add ivy-feedwrangler

asimpson:
Provides an ivy interface for the https://feedwrangler.net RSS service.

https://github.com/asimpson/ivy-feedwrangler

Maintainer

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/ivy-feedwrangler
+++ b/recipes/ivy-feedwrangler
@@ -1,0 +1,1 @@
+(ivy-feedwrangler :fetcher github :repo "asimpson/ivy-feedwrangler")


purcell:
Thanks! Sorry for the slow response - busy with work right now. Quick feedback:

- `;; Package-Requires: (ivy)` is not a valid format in older Emacsen: please add the specific minimum version string. 
- [This code](https://github.com/asimpson/ivy-feedwrangler/blob/ebeda2482fdbc42ca9e2aebb6f388e56cf0928e9/ivy-feedwrangler.el#L36-L40) isn't idiomatic: don't create nil local vars then set them, only to throw them away. Use `let` instead.
- Avoid putting closing `)` on their own lines: they belong with their friends.
- Add an `;;;###autoload` comment above `(defun ivy-feedwrangler()` so that people don't need to `require` the package in order to call that command.
- `(equal (length feed) 0)` should be written `(null feed)`.
- Either the indentation is wrong [here](https://github.com/asimpson/ivy-feedwrangler/blob/ebeda2482fdbc42ca9e2aebb6f388e56cf0928e9/ivy-feedwrangler.el#L69) or that `let` expression has no effect on the ensuing code. I haven't counted parens to check which is the case. I recommend the use of `aggressive-indent-mode` when editing emacs lisp.
- Don't shell out to `curl` to make HTTP requests: either use the built-in `url` library, or the `requests` package.

Hope that helps!

asimpson:
@purcell No worries on the response time! The feedback you provided was so helpful for someone new to elisp and emacs package development. I really appreciate the level of detail you went into! 👏 

I've pushed up a commit which I think addresses all these comments. Thanks again for melpa, it's an incredible service!

purcell:
Looks good - merged. Welcome to MELPA!

PR #5103, opened 2017-11-06
Add recipe for elfeed-protocol

fasheng:

Provide extra protocols to make like ownCloud News works with elfeed 

https://github.com/fasheng/elfeed-protocol

maintainer

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/elfeed-protocol
+++ b/recipes/elfeed-protocol
@@ -1,0 +1,3 @@
+(elfeed-protocol
+  :repo "fasheng/elfeed-protocol"
+  :fetcher github)


purcell:
Thanks! Sorry for the delay in getting to this. Quick feedback:
- The `elfeed-protocols` function should be called `elfeed-protocol-list`, so that the symbol prefixes are uniform.
- The `elfeed-owncloud` file seems like it should be its own package, since in its current form it includes a lot of top-level symbols that don't start with `elfeed-protocol-`. Otherwise, perhaps it should be called `elfeed-protocol-owncloud.el`, and the symbols in it renamed accordingly.
- You're using `nadvice`, so you need to depend on a specific Emacs version: `package-lint` should have mentioned that.

Hope that helps!

fasheng:
Thanks for the advices, all fixed now, please take a review again.

purcell:
Great, thanks!

PR #5105, opened 2017-11-08
Add commit-msg-prefix

kidd:

Helper functions to hook onto git-commit-mode and provide a list of past commits where you can pick one from the list (different completion options (ivy/helm/ido)) and it inserts part of the text of the selected one. 

More info in:
http://puntoblogspot.blogspot.com.es/2017/07/announcing-commit-msg-prefix.html

https://github.com/kidd/commit-msg-prefix

It's the first time uploading the package

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/commit-msg-prefix
+++ b/recipes/commit-msg-prefix
@@ -1,1 +1,0 @@
-(commit-msg-prefix :repo "kidd/commit-msg-prefix" :fetcher github)
--- a/recipes/commit-msg-prefix
+++ b/recipes/commit-msg-prefix
@@ -1,0 +1,1 @@
+(commit-msg-prefix :repo "kidd/commit-msg-prefix" :fetcher github)
--- a/recipes/git-msg-prefix
+++ b/recipes/git-msg-prefix
@@ -1,0 +1,1 @@
+(git-msg-prefix :repo "kidd/git-msg-prefix" :fetcher github)


purcell:
Thanks! Sorry for the slow response. Quick feedback:
- The default [here](https://github.com/kidd/commit-msg-prefix/blob/e318cc509e0593cd21374203dd7a01920397df31/commit-msg-prefix.el#L54) should be `completing-read`. Generally you shouldn't need any code in packages like this that is specific to completion systems, because all the popular completion systems override `completing-read` according to the user's preferences. (The `helm` source is also therefore out-of-place IMO.)
- How would one customise this? There should be a `defgroup`, and at least one of the `defvar`s should be a `defcustom`.
- This really only supports `git`, so `git` should perhaps be included in the package name, or at least in the package description line.
- Don't refer people to the README for usage instructions: the code should be self-documenting once installed, in the Emacs tradition. :-) Perhaps add more details to the commentary.

Hope that helps!

kidd:
Thanks @purcell  for the feedback!

I'll try to address all the points and ping back

* Regarding completion, I don't believe on the different frameworks overriding the completing-read. I use ido and helm for different kinds of tasks, but I don't want only one framework for all my completions. Still, point taken, I'll see what I can do there.

 * I'm adding a defgroup and changing defvars to defcustom (never think of that, comming from common-lisp :p )

* I'll put something git-related in the name although my idea was that it could evolve to multiple vcs's. I guess if it evolves, there's time for a new package and deprecate this one.

* Removing references to README file.

tarsius:
Friendly ping!

kidd:
I did all the modifications and I'm currently testing the renaming of the whole package to git-msg-prefix. If this works I'll ping you later this week for another review and maybe merge :).

Thanks for the reminder!

purcell:
Nice, thanks.

PR #5106, opened 2017-11-09
Add recipe for helm-exwm

Ambrevar:

Helm for EXWM, the Emacs X Window Manager

https://github.com/emacs-helm/helm-exwm

I'm the author.

 **None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/helm-exwm
+++ b/recipes/helm-exwm
@@ -1,0 +1,3 @@
+(helm-exwm
+ :repo "emacs-helm/helm-exwm"
+ :fetcher github)


purcell:
Thanks! `C-c o` is reserved for end-user customisation, but I see that `helm` uses it elsewhere, which is a shame. Merging now.

Ambrevar:
Thanks!

`C-c o` is only used within the Helm buffer, so that would only be an issue in terms of convention assuming the user would rework Helm's bindings.  I can't change that though, this issue should be filed upstream.

purcell:
Agreed. In theory, `C-c LETTER` belongs to the user in any keymap, but in practice it's fine to follow along with helm's convention here. :-)

PR #5110, opened 2017-11-10
Add term-keys

CyberShadow:

This package allows configuring Emacs and a supported terminal emulator to handle keyboard input involving any combination of keys and modifiers.

https://github.com/CyberShadow/term-keys

Creator and maintainer.

None needed.

### Checklist

Please confirm with `x`:

- [x] I've used the latest version of [package-lint](https://github.com/purcell/package-lint) to check for packaging issues, and addressed ~~its~~ *applicable* feedback

### DIFF
--- a/recipes/term-keys
+++ b/recipes/term-keys
@@ -1,0 +1,1 @@
+(term-keys :repo "CyberShadow/term-keys" :fetcher github)


purcell:
Thanks - this looks useful!

Which feedback from `package-lint` did you decide was not applicable, and why? I notice, for example, that you're still using `/` in symbol names, which doesn't match the elisp guidelines.

CyberShadow:
Thanks!

> I notice, for example, that you're still using / in symbol names, which doesn't match the elisp guidelines.

Yes, that's the main one. I disagree with the guideline, both subjectively and objectively, and would like to use something other than `-` for the package namespace delimiter for my own projects.

package-lint also complained about missing package.el headers on the supplementary files, which I understood weren't necessary.

purcell:
> package-lint also complained about missing package.el headers on the supplementary files, which I understood weren't necessary.

Yep, that's correct, so no issues there.

> Yes, that's the main one. I disagree with the guideline, both subjectively and objectively, and would like to use something other than - for the package namespace delimiter for my own projects.

While there are a few popular packages which use different conventions as a historical accident, we're not currently accepting new packages that don't fit with the [elisp conventions](https://www.gnu.org/s/emacs/manual/html_node/elisp/Coding-Conventions.html), sorry.

It's okay to write personal code in a different style, but when sharing with the community in the form of a package, the code should follow the community norms. It's ultimately better to favour consistency over personal style, even for authors, because it makes it easier for people to read and contribute to each other's projects.

CyberShadow:
> While there are a few popular packages which use different conventions as a historical accident, we're not currently accepting new packages that don't fit with the elisp conventions, sorry.

That's very surprising.

I understand position regarding consistency, and I am happy to follow existing conventions when contributing to other projects. However, for the code where I am the creator and main author, I am loath to change it, as not only do I think the change would be a net negative, but the knowledge that I have been forced to make a change throughout the code that I disagree with would lead to me being less motivated to continue working on this project, as well as future Emacs Lisp packages that would be otherwise suitable for submitting to MELPA.

I'd like to question whether rejecting packages over a disagreement over code style is truly in the interests of users of MELPA, and how many of MELPA's users are aware that useful packages can be rejected over such minor issues.

Are you sure this rule is really in the community's best interests?

purcell:
> Are you sure this rule is really in the community's best interests?

I'm not sure how we would measure this either way, but I know from experience of reviewing every package that appears in MELPA that authors learn (and copy/paste) best practices from the source code of other packages, so every package formatted according to the author's style preferences rather than the elisp standards works subtly against us maintainers and the community as a whole. My personal view is that elisp published as packages is meant to be community property, and it should be formatted and documented as such, just as it is licensed as such.

In the specific case of `/`, note that even the widely-used `yasnippet` eventually changed its `yas/` prefix to `yas-`, recognising that this was a better (more standard) choice for the community. I'd encourage you to reconsider.

CyberShadow:
> I'm not sure how we would measure this either way, but I know from experience of reviewing every package that appears in MELPA that authors learn (and copy/paste) best practices from the source code of other packages, so every package formatted according to the author's style preferences rather than the elisp standards works subtly against us maintainers and the community as a whole.

I understand.

I listed some subjective reasons why I'm loath to perform the change. Here is why I think that the current style guideline is not optimal from (hopefully) a more objective perspective:

1. It makes identifiers much less readable when the package name or prefix consists of more than one word. When a word that's part of an identifier is a verb or adjective, it can no longer be clear what it's referring to.

2. With a character dedicated for the purpose of delimiting the namespace from the rest of the identifier, it becomes trivial (both visually, and programmatically) to distinguish which part of the identifier is the namespace, and which actually describes the identifier's function.

3. As a consequence of the above, it encourages using a namespace prefix that doesn't correspond to the package name. If all symbols were consistently prefixed with the package name (plus a distinct delimiter), that would guarantee that no symbols would ever collide.

4. Considering the wealth of characters that's available for Emacs Lisp identifiers, it seems so wasteful to restrict all word delimiters to one character, and even more asinine to actually enforce this restriction unto others.

5. Looking at some past MELPA submissions (as well as some existing packages in MELPA), I see that using a distinct namespace delimiting character is a popular choice among package authors. I am not the first to attempt to plead allowing the use of another character for delimiting the package namespace.

Considering the arguments above, I am not convinced that enforcing consistency in this case is a goal with a positive net result. I think the MELPA team has a number of options at their disposal regarding this matter:

- Whether and to which extent MELPA enforces the Emacs Lisp Coding Conventions is MELPA's choice. I think MELPA's decision to enforce them unconditionally is an odd and highly unusual one, as I'm not aware of any programming language package repository which enforces a particular code style, even for languages that have published an official style guide.

- Especially on subjective matters, zero-tolerance policies and unconditional restrictions can do more harm than good. As the quote goes, "a foolish consistency is the hobgoblin of little minds". I think such issues should be judged on a case-by-case basis. Will enforcing the convention make the code more readable and accessible, or the opposite?

- The coding conventions were written a long time ago, predating MELPA by many years. As programming practices evolve and improve, many conventions become obsolete or harmful. If the coding conventions were written today, they likely wouldn't be the same as they are now.

- Finally, the coding conventions are likely not set in stone. If they have become outdated, then perhaps they ought to be updated to be more useful today.

> My personal view is that elisp published as packages is meant to be community property, and it should be formatted and documented as such, just as it is licensed as such.

Hmm. Perhaps I have misunderstood MELPA's purpose.

I think that's not what you actually meant, but one way to interpret your statement above is that authors should publish packages on MELPA once they are no longer interested in using, developing, or maintaining them, and would like the community to take over further maintenance. Thus, "community property" would mean that the original author is no longer any more special than any other Emacs user, and thus their further changes should be reviewed by the community as those coming from any other Emacs user. This is certainly not the relationship I have with `term-keys` right now.

Your statement would be more applicable for packages submitted for inclusion with Emacs itself, since this process involves assigning copyright to the Free Software Foundation, and therefore the code literally becomes the (intellectual) property of the FSF. Incidentally, packages that are distributed with Emacs and whose development occurs in the Emacs repository are likely to have much fewer contributions and much slower development pace, due to being tied to the Emacs release cycle and all contributors being required to go through the copyright assignment bureaucracy routine.

I think I understand what you really meant by that statement, and though I empathize with the general sentiment, I think it is not representative of how packages continue to be developed after their submission to MELPA. Of all packages in MELPA, how many have received more (>50%) contributions in code from the community than from their original authors? Not very many, I wager.

In either case, my goal for submitting `term-keys` to MELPA is only to make the package easily installable and updatable via `package.el`, nothing more, nothing less. I have initially written `term-keys` to solve a personal problem, but afterwards, seeing that no similar satisfactory solutions exist, I decided to spend some time to add support for other terminals (which I do not use personally) and document everything. I've submitted it to MELPA in the hope that making the package more discoverable and providing an easy way to install it would help prevent others with the same problem to waste time solving it on their own. However, I do not agree with making the change you request, even though it would settle this discussion and make `term-keys` available in MELPA, because not only do I dislike the change subjectively and think that it would result in a net loss for `term-keys` objectively, but also because I think MELPA's policy of unconditionally enforcing code style is a net loss for the Emacs community as a whole.

> In the specific case of `/`, note that even the widely-used `yasnippet` eventually changed its `yas/` prefix to `yas-`, recognising that this was a better (more standard) choice for the community.

I understand that this change occurred mainly due to pressure from the MELPA team, and the same flawed philosophy discussed here, so I'm not sure if that precedent adds much weight as an argument.

> I'd encourage you to reconsider.

Thank you for having this conversation. In truth, I have predicted that it would occur before I submitted the package to MELPA. For a while, I was torn between either throwing in the towel one way, and performing the change and probably staying away from Emacs Lisp development for a while, or throwing it the other way and abandon any attempt of submitting my Emacs packages to MELPA, both of which would mean that I don't have to spend any more time worrying about this matter. However, as far as I can see from the information currently available to me, and barring any new arguments, I feel like there is room for improvement in MELPA's policy on this matter, and improving the policy would be the optimal outcome in this situation.

Would MELPA consider allowing `/` as the namespace delimiter, but only if the namespace is the same as package name?

milkypostman:
i would just like to point out a few things and ultimately leave the decision to @purcell who puts in so much time with MELPA and improving the packages that get submitted to MELPA that he deserves the right. but i digress...

Emacs packages on MELPA are really no different than Elisp packages in the core Emacs release. the only thing that stops most packages from being part of Emacs is signing over copyright to FSF. by releasing your package you are making a contribution to the Emacs project and based on your statement above you should follow the standards or that project.

What surprises me is that you care about sharing code to help others solve a problem, but you don't care about making that code easier for them to read and change based on a convention that they are accustomed to by using Emacs.

if you strongly disagree with the standard then why not put the effort you used in this PR into trying to convince the Emacs community (emacs-devel@ mailing list) to update the coding standards to be more flexible? why pick a fight and try to convince people who are only putting time and money into trying to improve the community?

By accepting the package we make more work for ourselves by allowing an exception and having that hang over us as we try to have other packages adhere to the Emacs lisp standard conventions. I guess we can always say "If term-keys jumped off a cliff would you?" to those future convention breakers...

By rejecting the package we have this as leverage for future submissions.

In the end whether we accept the package to MELPA or not, we have really accomplished nothing here.

CyberShadow:
Thanks for weighting in, @milkypostman!

> Emacs packages on MELPA are really no different than Elisp packages in the core Emacs release. the only thing that stops most packages from being part of Emacs is signing over copyright to FSF. by releasing your package you are making a contribution to the Emacs project and based on your statement above you should follow the standards or that project.

That's an interesting point of view, and I can now see your perspective (with MELPA being not too different from GNU ELPA technically), however, I'm not sure that many MELPA users or package contributors share that view. In this perspective, MELPA is very different from other projects and programming languages' package repositories, especially considering that the code continues to be developed and under the exclusive control of the package's author.

If someone were to submit a package, address all change requests, get the package included, then or introduce new code or other changes which would not have passed the MELPA inclusion review, there is little MELPA can do about it, other than perhaps removing the package (and causing a problem to MELPA users who had already installed it), or forking the package from the last "good" version. Considering this, does a policy of unconditionally enforcing restrictions during review really make sense?

> What surprises me is that you care about sharing code to help others solve a problem, but you don't care about making that code easier for them to read and change based on a convention that they are accustomed to by using Emacs.

Sorry, that does not accurately describe the situation. I do care about making the code easier to read and change, and my concern for the code's readability is one of the reasons why I think changing the delimiters is a bad idea. What I am convinced of and what we seem to disagree with, is that the convention would be making the code *worse*, as I have argued above.

> if you strongly disagree with the standard then why not put the effort you used in this PR into trying to convince the Emacs community (emacs-devel@ mailing list) to update the coding standards to be more flexible? why pick a fight and try to convince people who are only putting time and money into trying to improve the community?

As mentioned above, that's certainly one direction that we can take, but it would be fruitless if consensus can't be reached at least with the MELPA team.

tarsius:
>> Emacs packages on MELPA are really no different than Elisp packages in the core Emacs release. the only thing that stops most packages from being part of Emacs is signing over copyright to FSF. by releasing your package you are making a contribution to the Emacs project and based on your statement above you should follow the standards or that project.

> That's an interesting point of view, and I can now see your perspective (with MELPA being not too different from GNU ELPA technically), however, I'm not sure that many MELPA users or package contributors share that view. In this perspective, MELPA is very different from other projects and programming languages' package repositories, especially considering that the code continues to be developed and under the exclusive control of the package's author.

You certainly don't have to share that view - I am not sure I do. But what is more important is that the maintainers of MELPA have decided to enforce some (somewhat controversial) conventions. *That* is not different from other package repositories, some of those also enforce certain conventions.

Someone else could create another ELPA archive that is less strict. Doing that might actually be a good idea.

I see no point leaving this open. (And also don't really want to get involved in the discussion. I am not the one making these policy decisions but I agree in this case (and most others).)

I suggest you use `package-build.el` to create releases that can be installed using `package-install-file` or even create an ELPA archive and update the documentation accordingly.

tarsius:
> Feel free to send packages rejected for similar reasons my way.

I'll try to remember that.

PR #5112, opened 2017-11-10
Add latexdiff.el package

galaunay:

Add the possibility to use [latexdiff](https://github.com/ftilmann/latexdiff) in Emacs.
This allows to make diff pdfs between several versions of one same tex file. 

https://github.com/galaunay/latexdiff.el

Maintainer

**None needed** 

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/latexdiff
+++ b/recipes/latexdiff
@@ -1,0 +1,3 @@
+(latexdiff
+ :fetcher github
+ :repo "galaunay/latexdiff.el")


purcell:
Thanks! Quick feedback:

- Don't [insert filenames into command strings using `%s`](https://github.com/galaunay/latexdiff.el/blob/master/latexdiff.el#L203-L204) without first calling `shell-quote-argument` on them. What if they contain spaces or quotes? The same feedback applies to other parts of the code.
- Relatedly, what if one of those commands separated by `;` fails? Prefer `&&`, or pass `-e` to `bash`. `bash` shouldn't be necessary anyway, though: `sh` should be fine. The `GLOBIGNORE` part looks redundant. In any case, you can delete and copy files using Emacs functions rather than writing little shell scripts into strings. :-)
- You can probably arrange to run commands in a temp directory so that your cleanup of generated/temporary files is easier, and less prone to accidentally deleting user's files: I noticed your comment about that.
- The helm block is out of place, because it causes `helm` to be loaded even if the user does not want it to be, simply if it is available on their system: the block that starts `(when (featurep 'helm) (require 'helm)` should instead use `eval-after-load`, and the `autoload`s should eb removed.

 Hope that helps!

galaunay:
Thanks a lot for the feedback !
It did helped a lot.
I have enhanced the code, more specifically regarding your propositions:

> Don't insert filenames into command strings using %s without first calling shell-quote-argument on them. What if they contain spaces or quotes? The same feedback applies to other parts of the code.

Done (good to know by the way, thanks).

> Relatedly, what if one of those commands separated by ; fails? Prefer &&, or pass -e to bash. bash shouldn't be necessary anyway, though: sh should be fine. The GLOBIGNORE part looks redundant. In any case, you can delete and copy files using Emacs functions rather than writing little shell scripts into strings. :-)

I replaced a lot of sh scripts by Emacs functions, this gets rid of the 'GLOBIGNORE' and '&&' problems.
I also replaced bash by sh.

>  You can probably arrange to run commands in a temp directory so that your cleanup of generated/temporary files is easier, and less prone to accidentally deleting user's files: I noticed your comment about that.

All the file are now produced in directories with a specific name beginning with 'diff-'.
I removed the `latexdiff-clean` function, as those folders can be easily removed by classic means.

>  The helm block is out of place, because it causes helm to be loaded even if the user does not want it to be, simply if it is available on their system: the block that starts (when (featurep 'helm) (require 'helm) should instead use eval-after-load, and the autoloads should eb removed.

I put all the helm related functions into an `with-eval-after-load` statement.
These functions are now only loaded if helm is already loaded.

purcell:
Great!

PR #5114, opened 2017-11-12
Add recipe for abgaben.el

akoehn:

abgaben.el helps with correcting and grading assignments you receive by mail.
It automates saving attachments, linking them in org-mode files and composing replies
once you are done annotating the files (probably using pdf-tools).

https://gitlab.com/akoehn/abgaben

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/abgaben
+++ b/recipes/abgaben
@@ -1,0 +1,1 @@
+(abgaben :repo "akoehn/abgaben" :fetcher gitlab)


Alexander-Miller:
The package link is giving me a 404. I can't find the repository on your profile either. Did you delete it or make it private?

akoehn:
whoops, wrote github instead of gitlab...  Fixed.

Alexander-Miller:
* `(defcustom abgaben-root-folder "/home/arne/lehre/2017-gwv/abgaben/` - that should be set to some sane default, like $HOME/abgaben. Same for abgaben-org-file
 * Too much German in the code base. Not just the customization vars, but also text being inserted, like line 261.
 * Looks like you're using mu4e code. If so you should add it as a dependency and require it.
 * The workflow looks fairly nontrivial. A nicely formatted readme with pictures and examples is much more publikumswirksam than just the comment header.

akoehn:
Thanks for the reply, I'll look into those parts.  I forgot that my personal defaults are still in there.

mu4e is not a package; how should I add a dependency?

I'm working on a readme with gifs etc.

Alexander-Miller:
I'm asking about mu4e because I see you use functions `mu4e~view-get-attach`. and `mu4e-view-save-attachment-single`. Where are they defined?

akoehn:
They are part of mu4e, but mu4e is not installable via package.el.  I can add a `(require 'mu4e)` (and have done so), but can't add mu4e to the dependency list of my package.

Other packages extending mu4e also don't depend on mu4e.

Alexander-Miller:
Ah, I didn't know that mu4e is not on melpa. In that case all you need to do is `require` it.

akoehn:
I have updated the code base:

 - generic defaults for defvars
 - removed German strings
 - `(require 'mu4e)`

I'll add a README later, as time permits.

akoehn:
There is now a README with videos explaining the workflow.

purcell:
Thanks! Quick feedback:
- `emacs` isn't a great parent group for the `defgroup`. Maybe `applications` or `comm`?
- Typo: `worklfow`
- `completing-read` prompt strings should end with a space
- I think the correct English translation of Abgabe would be "submission". While I support the name of the package overall, I would advise using "submission" in symbol names, apart from the prefix -- e.g. `abgaben-save-abgabe` would be `abgaben-save-submission`, `abgaben--construct-abgabenbody` would be `abgaben--construct-submission-body` etc.

Hope that helps!

purcell:
Also, please move those lonely `)` where they belong, with their friends on the lines above. :-)

akoehn:
@purcell, thanks for your remarks! I've updated the package accordingly.

purcell:
Awesome - merged now! Welcome to MELPA. :-)

PR #5115, opened 2017-11-12
Add new file directories for treemacs.

Alexander-Miller:

This is just an update to the existing treemacs packages to prepare for some file structure refactoring. I like to separate my code into different files, and by now I've near 20 of them, with yet more to come. Time to (eventually) clean up a little. I want to put my elisp and python code into separate directories instead of stuffing everything into root. It's supposed to look [like this](https://github.com/Alexander-Miller/treemacs/tree/files/src).
Of course the package should still be built while the change isn't made yet, so all the old files and exclusions are kept in the recipes for now. I'll open a second PR to remove them once the refactoring is made.

I am the maintainer.

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/treemacs
+++ b/recipes/treemacs
@@ -1,4 +1,12 @@
 (treemacs
  :fetcher github
  :repo "Alexander-Miller/treemacs"
- :files (:defaults "icons" "treemacs-dirs-to-collapse.py" (:exclude "treemacs-evil.el" "treemacs-projectile.el")))
+ :files (:defaults
+         "icons"
+         "treemacs-dirs-to-collapse.py"
+         "src/elisp/treemacs*.el"
+         "src/scripts/treemacs*.py"
+         (:exclude "src/elisp/treemacs-evil.el"
+                   "src/elisp/treemacs-projectile.el"
+                   "treemacs-evil.el"
+                   "treemacs-projectile.el")))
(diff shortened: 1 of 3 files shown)


purcell:
I'd personally just put the python in a subdir, and leave the `.el` source at the top level. That'd save a bit of recipe clunkiness too. Is that an option?

Alexander-Miller:
Not really. Stashing the python away won't really reduce clutter and my expectations for more scripts have been heavily dialed back - turns out synchronously querying file system contents or parsing thousands of git status entries in elisp is faster than using an asynchronous python or shell script.
So it's really all about putting away all those .el files.

purcell:
> So it's really all about putting away all those .el files.

I guess I really just don't get it: the whole point of that repo is to hold `.el` files, and there's not much else there. So I don't see the benefit of hiding them away in a subdir. :-)

But anyway, I'll go ahead and merge.

Alexander-Miller:
> But anyway, I'll go ahead and merge.

Thanks.

>  So I don't see the benefit of hiding them away in a subdir. :-)

There are a few advantages for me. 

 * A completely flat file structure isn't the best use-case for treemacs, which I always use to work on the repo. A few extra dirs make for much better navigation with treemacs' goto parent/neighbour button functions.
 * My own working copy is quite a bit more cluttered than what you see on github. I have directories for testing git parsing, icon display and performance, and 3 or more various scripts that I play with.
 * Professionally I do java, so I guess this is just what I'm used to.

purcell:
:+1:

PR #5116, opened 2017-11-12
Add dotenv-mode

preetpalS:

A (simple) major mode for *.env* files which are typically used for loading environment variables into an application. *.env* files are typically loaded automatically by an added library such as one of the following:
- [Ruby: bkeepers/dotenv](https://github.com/bkeepers/dotenv/blob/master/lib/dotenv/parser.rb)
- [JavaScript: motdotla/dotenv](https://github.com/motdotla/dotenv)
- [PHP: vlucas/phpdotenv](https://github.com/vlucas/phpdotenv)
- [Golang: joho/godotenv](https://github.com/joho/godotenv)

https://github.com/preetpalS/emacs-dotenv-mode

I am the original author and maintainer.

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/dotenv-mode
+++ b/recipes/dotenv-mode
@@ -1,0 +1,3 @@
+(dotenv-mode
+  :repo "preetpalS/emacs-dotenv-mode"
+  :fetcher github)


preetpalS:
I primarily packaged this major mode for Melpa because it had gotten some stars after I had posted it as a [gist](https://gist.github.com/preetpalS/54acec3f166393f1d9e55380e1df7364). I improved the mode subsequently to support some more syntax although not all of the parsers in the listed *.env* loading libraries support everything that is currently being highlighted by this library (for example, I believe [motdotla/dotenv](https://github.com/motdotla/dotenv) does not support variable expansion by default (so `$` variables which are highlighted by this major-mode are not actually expanded when using that library)).

purcell:
Thanks. Quick feedback:

- The first line says "Major mode for .lock files"... :-/
- Please provide a short `Commentary` section which summarises the purpose and basic usage of the package.
- Why would `:group 'lisp` be the appropriate parent customisation group for this mode?
- Is `"\\.env.*\\'"` really the pattern you want? I feel like the second dot should also be escaped.
- The `dotenv-mode-disable-wildcard-env-filetype-support` var is bogus, and will not be respected if set after loading this file: just remove it and only set undeniably-correct associations in `auto-mode-alist`. The user can add more associations if they want to.
- Rather than `("#.*" . font-lock-comment-face)`, just add a syntax table entry.

preetpalS:
Thanks for the feedback, I appreciate it. I've addressed each point in the latest revision of the package (I've also re-tested with *checkdoc*, *byte compilation*, the *package-lint* package, and the *flycheck-package* package) and responded to each of your points below:

**The first line says "Major mode for .lock files"... :-/**

Fixed. :), Sorry about this, I should have caught that, I used the source code from [emacs-lockfile-mode](https://github.com/preetpalS/emacs-lockfile-mode) (that was already submitted to Melpa) as a template (for the license and basic comment structure).

**Please provide a short Commentary section which summarises the purpose and basic usage of the package.**

Done.

**Why would :group 'lisp be the appropriate parent customisation group for this mode?**

Fixed. I agree with this assessment, I checked out the customization groups in some of the programming language mode source files in Emacs itself, and they all seem to be using the ```'languages``` group.

**Is "\\.env.*\\'" really the pattern you want? I feel like the second dot should also be escaped.**

I basically wanted a pattern to match extensions such as the following:
- `.env`
- `.env.test`
- `.env.local`

That pattern matches them but it also matches patterns like ```.envc``` which is undesirable (but I wasn't really concerned about that). Anyways, `"\\.env\\..*\\'"` is a more appropriate pattern that I am now suggesting in the documentation of the major mode for those that want to associate this major-mode with more filetypes.

**The dotenv-mode-disable-wildcard-env-filetype-support var is bogus, and will not be respected if set after loading this file: just remove it and only set undeniably-correct associations in auto-mode-alist. The user can add more associations if they want to.**

I agree, I made the necessary changes to follow your suggestion.

**Rather than ("#.*" . font-lock-comment-face), just add a syntax table entry.**

I started using a syntax-table for the major mode (I was not familiar with syntax tables TBH).

purcell:
Thanks, looks great - merged!

I *think* that the `:abbrev-table nil` and `:syntax-table dotenv-mode-syntax-table` clauses are unnecessary: the former is redundant and the latter is set automatically by `define-derived-mode` as long as the syntax table variable has been set. Might want to test it, though. :-)

preetpalS:
Thanks for the merge :).

You do not need the `:abbrev-table nil` clause, but if it's missing you will get a warning (`dotenv-mode.el:55:Warning: Reference to unbound symbol: dotenv-mode-abbrev-table`) when you use **elint** to lint the file. I'd rather keep it there to prevent **elint** from having any warnings (even if I am only preventing false positives).

Also you are right about the `:syntax-table dotenv-mode-syntax-table` clause not being necessary if you define the `syntax-table` variable (like with `(setq-local syntax-table dotenv-mode-syntax-table)`) but I am probably going to keep using the clause for now as its more concise and equivalent.

purcell:
Interesting point about the `elint` rule. I'm not sure that it's a sensible check, but great that you're using tools like that. :-)

> Also you are right about the :syntax-table dotenv-mode-syntax-table clause not being necessary if you define the syntax-table variable (like with (setq-local syntax-table dotenv-mode-syntax-table)) but I am probably going to keep using the clause for now as its more concise and equivalent.

No, I mean you really don't need to mention it at all in the `define-derived-mode` form, because `define-derived-mode` generates code which automatically sets the syntax table to `dotenv-mode-syntax-table` if that var is bound -- which it is, in your case. You can see this by using `pp-macroexpand-last-sexp` or similar to expand the `define-derived-mode` form.

preetpalS:
Tools like *elint* and [package-lint](https://github.com/purcell/package-lint) are awesome because they allow you to shift your attention to problems that you actually want to solve.

You are right about not needing that clause. `pp-macroexpand-last-sexp` is incredibly useful (I wish I had known about it earlier as it's incredibly enlightening to say the least when trying to understand macros). Using `pp-macroexpand-last-sexp` is much easier than (although not a replace for) reading the source of a macro IMHO.

Interestingly, you could still decide on using the `:syntax-table dotenv-mode-syntax-table` clause but for another reason, **less generated code** (roughly 20 less generated lines of code and probably better performance).

Expanded `define-derived-mode` macro with `:syntax-table dotenv-mode-syntax-table` clause:
```elisp
(progn
  (defvar dotenv-mode-hook nil "Hook run after entering .env mode.\nNo problems result if this variable is not bound.\n`add-hook' automatically binds it.  (This is true for all hook variables.)")
  (unless
      (boundp 'dotenv-mode-map)
    (put 'dotenv-mode-map 'definition-name 'dotenv-mode))
  (with-no-warnings
    (defvar dotenv-mode-map
      (make-sparse-keymap)))
  (unless
      (get 'dotenv-mode-map 'variable-documentation)
    (put 'dotenv-mode-map 'variable-documentation
         (purecopy "Keymap for `dotenv-mode'.")))
  nil nil
  (put 'dotenv-mode 'derived-mode-parent 'prog-mode)
  nil
  (defun dotenv-mode nil "Major mode for `.env' files.\n\nIn addition to any hooks its parent mode `prog-mode' might have run,\nthis mode runs the hook `dotenv-mode-hook', as the final step\nduring initialization.\n\n\\{dotenv-mode-map}"
         (interactive)
         (delay-mode-hooks
           (prog-mode)
           (setq major-mode 'dotenv-mode)
           (setq mode-name ".env")
           (progn
             (if
                 (get 'prog-mode 'mode-class)
                 (put 'dotenv-mode 'mode-class
                      (get 'prog-mode 'mode-class)))
             (unless
                 (keymap-parent dotenv-mode-map)
               (set-keymap-parent dotenv-mode-map
                                  (current-local-map)))
             nil nil)
           (use-local-map dotenv-mode-map)
           (set-syntax-table dotenv-mode-syntax-table)
           nil
           (setq-local font-lock-defaults
                       '(dotenv-mode-highlights)))
         (run-mode-hooks 'dotenv-mode-hook)))
```

Expanded `define-derived-mode` macro without `:syntax-table dotenv-mode-syntax-table` clause:
```elisp
(progn
  (defvar dotenv-mode-hook nil "Hook run after entering .env mode.\nNo problems result if this variable is not bound.\n`add-hook' automatically binds it.  (This is true for all hook variables.)")
  (unless
      (boundp 'dotenv-mode-map)
    (put 'dotenv-mode-map 'definition-name 'dotenv-mode))
  (with-no-warnings
    (defvar dotenv-mode-map
      (make-sparse-keymap)))
  (unless
      (get 'dotenv-mode-map 'variable-documentation)
    (put 'dotenv-mode-map 'variable-documentation
         (purecopy "Keymap for `dotenv-mode'.")))
  (progn
    (unless
        (boundp 'dotenv-mode-syntax-table)
      (put 'dotenv-mode-syntax-table 'definition-name 'dotenv-mode))
    (defvar dotenv-mode-syntax-table
      (make-syntax-table))
    (unless
        (get 'dotenv-mode-syntax-table 'variable-documentation)
      (put 'dotenv-mode-syntax-table 'variable-documentation
           (purecopy "Syntax table for `dotenv-mode'."))))
  nil
  (put 'dotenv-mode 'derived-mode-parent 'prog-mode)
  nil
  (defun dotenv-mode nil "Major mode for `.env' files.\n\nIn addition to any hooks its parent mode `prog-mode' might have run,\nthis mode runs the hook `dotenv-mode-hook', as the final step\nduring initialization.\n\n\\{dotenv-mode-map}"
         (interactive)
         (delay-mode-hooks
           (prog-mode)
           (setq major-mode 'dotenv-mode)
           (setq mode-name ".env")
           (progn
             (if
                 (get 'prog-mode 'mode-class)
                 (put 'dotenv-mode 'mode-class
                      (get 'prog-mode 'mode-class)))
             (unless
                 (keymap-parent dotenv-mode-map)
               (set-keymap-parent dotenv-mode-map
                                  (current-local-map)))
             (let
                 ((parent
                   (char-table-parent dotenv-mode-syntax-table)))
               (unless
                   (and parent
                        (not
                         (eq parent
                             (standard-syntax-table))))
                 (set-char-table-parent dotenv-mode-syntax-table
                                        (syntax-table))))
             nil)
           (use-local-map dotenv-mode-map)
           (set-syntax-table dotenv-mode-syntax-table)
           nil
           (setq-local font-lock-defaults
                       '(dotenv-mode-highlights)))
         (run-mode-hooks 'dotenv-mode-hook)))
```

PR #5120, opened 2017-11-14
Update imgur recipe

marsam:

Hi:
I want to promote my imgur.el instead of the currenlty available one. Mostly because the current imgur.el uses a [deprecated api since march 2016](https://blog.imgur.com/2016/01/11/imgur-api-version-3/), and I think is no longer maintained.

Feel free to close, if you consider it unnecesary.

https://github.com/emacs-pe/imgur.el

maintainer

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/imgur
+++ b/recipes/imgur
@@ -1,1 +1,1 @@
-(imgur :fetcher github :repo "myuhe/imgur.el")
+(imgur :fetcher github :repo "emacs-pe/imgur.el")


dotemacs:
Your package name @marsam, is the same as the package you're replacing.

What are the rules about this, package name "shadowing"?

tarsius:
Well, I would say the rule is "It all depends on the contents of 'Relevant communications with the upstream package maintainer'", which in this case unfortunately was left out :wink:

marsam:
Hi, I only realized that there was already a imgur.el package after I cloned the melpa repository, sorry I've incorrectly hurried and made this pull request. I'm going to contact with the current package maintainer.

ecraven:
I'd love to see this in melpa, just tried it, and it works well ;)

PR #5123, opened 2017-11-17
Add recipe for org-noter

weirdNox:
This is an alternative to the original Interleave package: It shows a PDF and an Org notes buffer side by side so you can take notes associated with a page using concise shortcuts.

Note: As this provides the same functionality as Interleave, these functions are also prefixed with interleave- or interleave--, so they will collide with the original package.

https://github.com/weirdNox/alt-interleave

I am the maintainer

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/alt-interleave
+++ b/recipes/alt-interleave
@@ -1,0 +1,1 @@
+(alt-interleave :fetcher github :repo "weirdNox/alt-interleave")
--- a/recipes/alt-interleave
+++ b/recipes/alt-interleave
@@ -1,1 +1,0 @@
-(alt-interleave :fetcher github :repo "weirdNox/alt-interleave")
--- a/recipes/noter
+++ b/recipes/noter
@@ -1,0 +1,1 @@
+(noter :fetcher github :repo "weirdNox/noter")
--- a/recipes/noter
+++ b/recipes/noter
@@ -1,1 +1,0 @@
-(noter :fetcher github :repo "weirdNox/noter")
--- a/recipes/org-noter
+++ b/recipes/org-noter
@@ -1,0 +1,1 @@
+(org-noter :fetcher github :repo "weirdNox/org-noter")


purcell:
Hi! Thanks for this first MELPA submssion.

> Note: As this provides the same functionality as Interleave, these functions are also prefixed with interleave- or interleave--, so they will collide with the original package.

Sorry, that's not going to work for distribution here: if this is a distinct package, it should get its own unique symbol prefix, matching the package name.

Best option would be to work with the `interleave` author to extend that package such that it doesn't need to be cloned. Have you investigated that option? We're not keen to host arbitrary forks of packages that differ only slightly, because it's a bad experience for the package-using community.

weirdNox:
Hello there!

I've talked with Sebastian (`interleave` maintainer), but we didn't even discuss merging because, under the hood, it is completely different from the original (in terms of the Org API it uses, and even some of the functionality it provides); it would take a long time and almost for no benefit. I started by contributing to it but, as I had more ideas (which would cause a lot more friction than a single bugfix), I started making my own.

Anyway, I understand your rationale and I only made this package so I could annotate PDFs for studying - as such, I don't mind just having it as part of my dotfiles :)

Thank you for maintaining MELPA, it rocks!

tarsius:
In that case it might be best to pick a different package name / prefix.

weirdNox:
Hello there! I'm sorry about the delay, I spent these last weeks studying for college. I was thinking of changing both the name and the prefix of the package to `noter`. What do you think about it? Would it be accepted?

Also, I take the opportunity to thank you (@tarsius) for the amazing magit package! I use it for everything git related, it is awesome!

purcell:
Yep, `noter` would be fine. :-)

weirdNox:
Hello there! 

I've changed everything in my package to be under the prefix `noter-` and updated the pull request reflecting these changes.

Happy holidays!

alphapapa:
@weirdNox I haven't used either package yet, but yours sounds like a significant improvement, and I hope to try it sometime.  So thanks for contributing it!

Have you considered including `org-` in the name?  `noter` doesn't indicate it has anything to do with org-mode or annotating documents.  Something like `org-annotator` (which seems available, from a quick googling) would be more descriptive.

weirdNox:
`org-annotator` seems fine to me! I'll wait for the final decision from @purcell before making any changes though, so I don't have to fix it again if he decides against, but thanks for the idea :)

purcell:
Either naming is fine as far as I am concerned, as long as all the symbol names match up with the package name. It might be nice to advertise the `org` connection, I guess, but I don't really have a preference. So just let me know what you decide.

weirdNox:
I decided to go with `org-noter`, because, even though it isn't in MELPA, I found some packages named `org-annotate` and things like that, and having `org` in name is nice indeed.

Thanks everybody!

purcell:
Thanks. Sent you https://github.com/weirdNox/org-noter/pull/1, and then we're good to merge. :-)

weirdNox:
Done!

purcell:
Merged - thanks! Welcome to MELPA. :-)

PR #5124, opened 2017-11-17
Add recipe for evil-collection

jojojames:

This package provides a collection of keymap bindings for `evil-mode' for various modes.

Users can pick and choose which sets of bindings they want to enable or disable.

https://github.com/jojojames/evil-collection

Author/Maintainer

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/evil-collection
+++ b/recipes/evil-collection
@@ -1,0 +1,1 @@
+(evil-collection :fetcher github :repo "jojojames/evil-collection")


purcell:
Thanks, and sorry for the slow response - busy time of year!

I can see how this collection could be useful to people. To be well-packaged, though, there should be an overall namespace shared by all the defined symbols, and you've got lots of modules whose file, feature and symbol names could easily end up clashing with other packages. You should therefore rename all the auxilliary files (and the symbols they define) to use the `evil-collection-` prefix.

Ambrevar:
On the one hand it's obviously best practice.

On the other hand, while I understand the problem I don't really like using `evil-collection-` as a namespace for each binding set.  Let's discuss some points and see if we can come up with something better.

Every file is independent and could be a package by itself thus the current namespacing makes a lot of sense.  If we decide to merge some existing third-party bindings into an `evil-collection-` namespace , it would mean we would have to change its namespace (e.g. `evil-foo` to `evil-collection-foo`).

Ideally, we should not expect any competiting package which would clash with the current namespacing, e.g. there is little point for anyone to make a competing evil-help.
A counter example happened once: evil-org and org-evil.  I think it's the counter-example that proves the point: the competition in that area does not bring much to the community and we would probably be better off with just one package.

Besides, it makes little sense to install two competing package at the same time.

The only reason we've put all files under one repo is a matter of convenience for the user: so that they pull only one package from MELPA.  Is there any way to achieve that without changing the namespace?

Last, the `evil-collection-foo-` namespace looks kind of ugly.

purcell:
> On the one hand it's obviously best practice.

Yep!

> On the other hand, while I understand the problem I don't really like using evil-collection- as a namespace for each binding set. Let's discuss some points and see if we can come up with something better.

Well, yes, you'd naturally prefer the short names, because even if there's a name clash elsewhere, it's _your_ code you care about. But this is about giving each author their own namespace which they can reasonably own. 

If someone else submits an `evil-ag` package to MELPA, then it will likely get accepted, and your package will probably break. Or maybe theirs will, but just in certain circumstances: and that's then your fault, not theirs! :-)

> Every file is independent and could be a package by itself thus the current namespacing makes a lot of sense. If we decide to merge some existing third-party bindings into an evil-collection- namespace , it would mean we would have to change its namespace (e.g. evil-foo to evil-collection-foo).

Yes - if code becomes part of your package, it must use the same prefix.

> Ideally, we should not expect any competiting package which would clash with the current namespacing, e.g. there is little point for anyone to make a competing evil-help.

How do you know that? `evil-` + `any-other-package-name` is such a broad naming convention that you cannot assume you can just use it inside your package.

> A counter example happened once: evil-org and org-evil. I think it's the counter-example that proves the point: the competition in that area does not bring much to the community and we would probably be better off with just one package.

I don't think that's relevant here, sorry.

> Besides, it makes little sense to install two competing package at the same time.

That is no argument at all. Emacs packages must be safe for this eventuality. Even the Emacs standard library contains competing packages that do not break each other.

> The only reason we've put all files under one repo is a matter of convenience for the user: so that they pull only one package from MELPA. Is there any way to achieve that without changing the namespace?
> 
> Last, the evil-collection-foo- namespace looks kind of ugly.

Well, this is really the heart of your objection. You can always choose a shorter prefix and call the whole package "evilcoll", "keyvil" etc. instead! :-)

jojojames:
@purcell Is 'ec' (not sure if there's already one) unreasonable to take? Smartparens uses 'sp'.

evilcoll or whatever works for me @Ambrevar. I think following best practices is ideal.

purcell:
`ec` would be okay if you call the package `ec`, but I'd suggest using a more distinctive package name & prefix. We have six times as many packages in MELPA as there are possible two-character prefixes. :-) `smartparens` and `multiple-cursors` wouldn't get approved in their current form if they were submitted now, so they're not a valid precedent.

Ambrevar:
Interesting namespace suggestions, but I think we are better off sticking to `evil-` something: After all, that's what users will be looking for when they browse the package list.

As for my objections: @purcell I agree with all your points when it comes to a single package.  But evil-collection, as the name implies, is a bit different.  My question is: is there any way of handling "meta" packages in MELPA?
(I don't think so, but it's worth asking anyways.)

Technically we could have have every binding released as a seperate package and `evil-collection` depend on all of them, so that the user would only have to install it.  That would clutter the MELPA repo a little bit, but lead to much nicer name-spacing for the bindings.  What do you think?

The heart of the issue is, as always, the Emacs does not have proper name-spacing support with named imports.
If there is no way around this, then I suggest we go ahead with the initial suggestion, i.e. the `evil-collection-` namespace.

jojojames:
@Ambrevar I agree with your points but if we had to use a namespace per package, lets go with a shorter one? The package can still be named evil-collection but the namespace underneath can be something shorter.

ecol, evol, etc.

Ambrevar:
I think the namespace should being with "evil".

purcell:
> Technically we could have have every binding released as a seperate package and evil-collection depend on all of them, so that the user would only have to install it. That would clutter the MELPA repo a little bit, but lead to much nicer name-spacing for the bindings. What do you think?

There's no benefit to that unless the packages would be usable on their own: if that's what you're suggesting, then it might be an option, but I'd suggest that you probably don't want to try and sync up a bunch of recipes with files in your repo anyway. And I'd rather not have a big pile of tiny related-but-not-obviously-related packages in MELPA, because I don't think it helps users.

> The heart of the issue is, as always, the Emacs does not have proper name-spacing support with named imports.

Agreed. A lot of little frustrations in the emacs package world are related to this.

> If there is no way around this, then I suggest we go ahead with the initial suggestion, i.e. the evil-collection- namespace.

That's still the best path IMO, or `evilcol-`, `evilkey-`.

Ambrevar:
> There's no benefit to that unless the packages would be usable on their own:

They are! :)

> And I'd rather not have a big pile of tiny related-but-not-obviously-related packages in MELPA

Well, they are obviously related, aren't they?

That being said, I agree, it would not be pretty to have dozens of evil-foo in MELPA.

jojojames:
Made the namespace changes @purcell.

purcell:
This looks great now - simple, consistent, unambiguous. Thanks!

Merging now... apologies for the slow turnaround. December has been brutal.

jojojames:
Awesome! Thanks @purcell!

> apologies for the slow turnaround. December has been brutal.

Holidays can be rough. :)

PR #5126, opened 2017-11-17
Add recipe for auto-semicolon.

jcs090218:

This is a simple package that will automatically add semicolon
at the end of the line.  This package is only helpful if you
code programming language that need semicolon at the end of
each operation.

https://github.com/jcs090218/auto-semicolon

I am both creator and maintainer.

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/auto-semicolon
+++ b/recipes/auto-semicolon
@@ -1,0 +1,1 @@
+(auto-semicolon :repo "jcs090218/auto-semicolon" :fetcher github)


purcell:
Thanks! Sorry for the slow response - I'm extremely busy right now. Quick feedback:
- You can (and should) remove the `-pkg.el` file, because MELPA will happily read package metadata from the main `.el` file. (Don't forget to replace "a simple package" with something like "Minor mode which inserts semicolons for you".)
- Your `do-list` searches could be written more elegantly with `cl-some`, passing `auto-semicolon-current-char-equal-p` as the predicate.
- Please don't put closing parens on their own lines.
- When there is no `else` form, you should use `when` or `unless` in place of `if`.

But most importantly, is this package different from [smart-semicolon](https://github.com/iquiw/smart-semicolon)? We prefer not to host multiple very similar packages.

jcs090218:
Okay, thank you for the feedback. I didn't know there is already another similar package exists. I will just close this pull request!

PR #5127, opened 2017-11-18
Add recipe for lsp-vue

cryptomaniac512:

Vue.js support for lsp-mode using [vls](https://github.com/vuejs/vetur/tree/master/server)

https://github.com/emacs-lsp/lsp-vue

I am the original author and maintainer.

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/lsp-vue
+++ b/recipes/lsp-vue
@@ -1,0 +1,1 @@
+(lsp-vue :repo "emacs-lsp/lsp-vue" :fetcher github)


purcell:
Thanks, and sorry for the slow response - I'm currently extremely busy! Some quick feedback:

- You shouldn't normally need to `(require 'vue-mode)` inside a function, when the whole module depends on that file. It should be sufficient to `;;;###autoload` the `lsp-define-stdio-client` form and the `lsp-vue-mmm-enable` function, assuming that `lsp-mode` has proper support for deferring those autoloads.
- Please inline some documentation in the `Commentary` section: the elisp file should be self-documenting as far as possible.
- The license is in a separate file which will not be included in the package, so a `License: MIT` header would be appreciated.

Hope that helps!

purcell:
Oh, also: the `:files` form in the recipe is unnecessary here and should be removed.

cryptomaniac512:
Thank you for help!

> when the whole module depends on that file.

Not all module depends on this file, only `lsp-vue-mmm-enable` function.

purcell:
Okay, great - makes sense.

PR #5128, opened 2017-11-19
Add molecule package

drym3r:

This is a molecule wrapper, the ansible testing framework.

https://git.daemons.it/drymer/molecule.el

I'm the developer.

None.

### Checklist

Please confirm with `x`:

- [  ] `M-x checkdoc` is happy with my docstring:
  `"Keymap for all ‘molecule-mode’ functions."` It says it should be something like: `"Keymap for function ‘molecule-mode’."` but that makes zero sense, syntactically speaking.

### DIFF
--- a/recipes/molecule
+++ b/recipes/molecule
@@ -1,1 +1,1 @@
-(nikola :fetcher git :url "https://git.daemons.it/drymer/molecule.el/")
+(molecule :fetcher git :url "https://git.daemons.it/drymer/molecule.el/")
--- a/recipes/molecule
+++ b/recipes/molecule
@@ -1,0 +1,1 @@
+(nikola :fetcher git :url "https://git.daemons.it/drymer/molecule.el/")


purcell:
Thanks! That repo doesn't seem to be available: I'm getting a 502.

drym3r:
It's working right now, so maybe it was a temporal error.

purcell:
Thanks, appears to be working now. Quick feedback:
- You're using `ivy-read`, but you don't declare a package dependency on it. However, don't do that: just use `completing-read` instead, which will get overridden magically by `ivy`, `helm` etc. as preferred by the user. (Also, why explicitly say `Choose between "role" or "scenario"` if you're using a completing read anyway? It will show the options to the user.)
- Concatenating command strings will cause problems when paths contain spaces or quotes etc. Use `shell-quote-argument`.
- It's an antipattern in lisp (and many/most languages, really) to initialise variables to null and then imperatively set them later. So everywhere that you have `let` introducing a bunch of variables, then lots of code which `setq`'s them, just set them directly in `let` instead. Use `let*` if necessary. You can also nest `let` expressions. The code will become shorter and safer. For example:
```el
  (let ((debug))
    ;; Set debug
    (if (eq molecule-debug-v t)
	(setq debug "--debug ")
      (setq debug ""))
    ...
```
becomes
```el
  (let ((debug (if molecule-debug-v "--debug " " ")))
    ...
```
(P.S. any non-`nil` value is true in lisp, so never check for `t` explicitly.)

- Don't [`defalias` to a lambda](https://git.daemons.it/drymer/molecule.el/src/master/molecule.el#L162). Just call `defun` instead!
- `" ADoc"` isn't a great lighter for the mode. I suggest `" Mol"` instead.
- `:group 'molecule.el` and `:require 'molecule.el` should be `:group 'molecule` and `:require 'molecule` respectively.
- `C-x m` is bound by default in Emacs to `compose-mail`, so this choice is not ideal, and slightly hostile.

Hope that helps!

drym3r:
Thanki you very much for the feedback! As you can see, I've programmed little with elisp. I'll apply your suggestions.

tarsius:
Friendly ping!

drym3r:
I've changed most of what you suggest, as I said thank you very much for your suggestions!

In the molecule--wrapper function I still have some variables initialized to null, but I still don't know how to handle them. I'll try it in a next version, I guess.

purcell:
Thanks. Looks good, and I'll merge this now, but an extra note or two:
- `(funcall 'molecule--wrapper "test")` should just be written `(molecule--wrapper "test")`. :-)
- ` (no shit, Sherlock)` - this isn't really helpful to users.
- Many of your comments are distractingly redundant, because the commented code is already clear.
- Using `completing-read`, it's possible for people to enter freeform (or nil) responses, which would break your code. Consider using the `require-match` parameter to that function.

drym3r:
Thanks for your comments, I'll take a further look.

tarsius:
@drym3r Do you still maintain this package? Where is its repository located nowadays?

tarsius:
And I have the same questions about `nikola`.

mattiasb:
@tarsius Looks like they are [here](https://gitlab.com/drymerisnothere/molecule-el) and [here](https://gitlab.com/drymerisnothere/nikola-el). It looks like the git.daemons.it domain does a 301 forward to https://gitlab.com/drymerisnothere/ :

```
curl -v https://git.daemons.it/
*   Trying 185.101.95.207:443...
* Connected to git.daemons.it (185.101.95.207) port 443 (#0)
* ALPN, offering h2
* ALPN, offering http/1.1
* successfully set certificate verify locations:
*   CAfile: /etc/pki/tls/certs/ca-bundle.crt
  CApath: none
* TLSv1.3 (OUT), TLS handshake, Client hello (1):
* TLSv1.3 (IN), TLS handshake, Server hello (2):
* TLSv1.2 (IN), TLS handshake, Certificate (11):
* TLSv1.2 (IN), TLS handshake, Server key exchange (12):
* TLSv1.2 (IN), TLS handshake, Server finished (14):
* TLSv1.2 (OUT), TLS handshake, Client key exchange (16):
* TLSv1.2 (OUT), TLS change cipher, Change cipher spec (1):
* TLSv1.2 (OUT), TLS handshake, Finished (20):
* TLSv1.2 (IN), TLS handshake, Finished (20):
* SSL connection using TLSv1.2 / ECDHE-RSA-AES256-GCM-SHA384
* ALPN, server accepted to use http/1.1
* Server certificate:
*  subject: CN=daemons.it
*  start date: Aug 23 10:15:09 2020 GMT
*  expire date: Nov 21 10:15:09 2020 GMT
*  subjectAltName: host "git.daemons.it" matched cert's "git.daemons.it"
*  issuer: C=US; O=Let's Encrypt; CN=Let's Encrypt Authority X3
*  SSL certificate verify ok.
> GET / HTTP/1.1
> Host: git.daemons.it
> User-Agent: curl/7.69.1
> Accept: */*
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 301 Moved Permanently
< Server: nginx/1.15.12
< Date: Sun, 18 Oct 2020 22:36:55 GMT
< Content-Type: text/html
< Content-Length: 170
< Connection: keep-alive
< Location: https://gitlab.com/drymerisnothere
< 
<html>
<head><title>301 Moved Permanently</title></head>
<body>
<center><h1>301 Moved Permanently</h1></center>
<hr><center>nginx/1.15.12</center>
</body>
</html>
* Connection #0 to host git.daemons.it left intact
```

tarsius:
> drymer is not here

I kinda took that literally.

drym3r:
Hey! Sorry for taking so long to answer. Yes, I migrated my stuff to that server. I do still maintain them. Both packages are on them. Do you want me to send a PR with the updated URLs?

tarsius:
Yes, please do that. Thanks!

PR #5129, opened 2017-11-19
Upgrade nikola repository url

drym3r:

This package is actually accepted, I just changed the domain.


purcell:
Thanks.

PR #5130, opened 2017-11-19
Add recipe for smart-jump

jojojames:

Allows users to set up multiple GoToDefinition/FindReferences functions to be used in a fallback fashion. If those functions/strategies don't work, DumbJump will be used as the final fallback mechanism.

https://github.com/jojojames/smart-jump

Author/Maintainer

None

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/smart-jump
+++ b/recipes/smart-jump
@@ -1,0 +1,1 @@
+(smart-jump :fetcher github :repo "jojojames/smart-jump")


purcell:
Thanks! Sorry for the slow response - I'm extremely busy right now.
 
This is a neat idea, and I think that `xref` has made a mistake in not supporting chains of successive go-to-definition functions. You might consider bringing that up on the emacs dev list, because fixing this at the source would be much cleaner than bolting on an `xref` wrapper like this.

Some other points of feedback:

- Use `derived-mode-hook-name` and `derived-mode-map-name` [here](https://github.com/jojojames/smart-jump/blob/1be5b4b32f93fa8ac831b19345fc7ad7efa206d3/smart-jump.el#L276)
- The text "upon registering `smart-jump'" is unclear. What does "registering" the package mean? Is it when a mode is enabled.
- Docstrings refer to `GoToDefinition`, which is confusing due to the spurious camel case.
- You're reinventing keymaps with [these definitions](https://github.com/jojojames/smart-jump/blob/1be5b4b32f93fa8ac831b19345fc7ad7efa206d3/smart-jump.el#L66) - this whole thing should be a minor mode which the user can enable. It will then have its own keymap, into which you can put these bindings.
- Why is there special handling for `smart-jump-simple-jump-function`? Should you just have a customisable list of handlers which would contain one for `dumb-jump` by default?

jojojames:
> Thanks! Sorry for the slow response - I'm extremely busy right now.

Thanks, no problem.

> This is a neat idea, and I think that xref has made a mistake in not supporting chains of successive go-to-definition functions. You might consider bringing that up on the emacs dev list, because fixing this at the source would be much cleaner than bolting on an xref wrapper like this.

I can check with emacs-devel in parallel but I'm a little wary of the time frame for using something like this if it were part of core.
Also, just wanted to clarify, that, spiritually, this is meant to be a dumb-jump wrapper. xref- just happened to be another convenient fallback.


>    Use derived-mode-hook-name and derived-mode-map-name here

Thanks, will do.


>    The text "upon registering `smart-jump'" is unclear. What does "registering" the package mean? Is it when a mode is enabled.

Thanks, I'll look into clarifying that.

>    Docstrings refer to GoToDefinition, which is confusing due to the spurious camel case.

I'll look into twweaking that. If you had suggestions, I'm all ears too.


>    You're reinventing keymaps with these definitions - this whole thing should be a minor mode which the user can enable. It will then have its own keymap, into which you can put these bindings.

I wanted to leverage the minor/major modes keymap instead to get out of the game of keymap collisions between minor modes.
Looking at dumb-jump, that one seems to provide 2 functions users can bind themselves instead of a minor mode.

For this package, I'm providing the service of rebinding the keys for the user. Moving it to a minor mode might make that service ineffective if they have to figure out how to give precedence to my mode.

>    Why is there special handling for smart-jump-simple-jump-function? Should you just have a customisable list of handlers which would contain one for dumb-jump by default?

I believe I did that for unit testing initially but looking at it more closely, I think you may be right.

purcell:
> I can check with emacs-devel in parallel but I'm a little wary of the time frame for using something like this if it were part of core.
> Also, just wanted to clarify, that, spiritually, this is meant to be a dumb-jump wrapper. xref- just happened to be another convenient fallback.

Yep, fair enough.

> For this package, I'm providing the service of rebinding the keys for the user. Moving it to a minor mode might make that service ineffective if they have to figure out how to give precedence to my mode.

Agree it's possible, if unlikely, but that's also an edge case you could quickly describe in the commentary and then the code could be written in a much more standard and straightforward way, which I would consider a big win for everyone.

jojojames:
> Agree it's possible, if unlikely, but that's also an edge case you could quickly describe in the commentary and then the code could be written in a much more standard and straightforward way, which I would consider a big win for everyone.

I'm not sure it's an unlikely edge case. Most minor modes will overwrite the "M-." / "M-," / "M-?" keys with their own mappings which will conflict with other minor-modes.

I've rewrote the code to use a minor mode and almost every single mode I entered gave precedence to the original minor mode (cider, ggtags, alchemist, etc) instead of smart-jump-mode.

I am assuming a user that uses this package and 'registers' a smart-jump would want smart-jump to override these keys instead of using the underlying mode-map's keys. 

I'm open to suggestions but it seems doing it with a minor-mode would be a frustrating experience for the user (either rewriting all the keybindings per mode or rejiggering the minor-mode-map-alist/list).

purcell:
All good - let's see how the code fares in wider use. :-)

PR #5131, opened 2017-11-20
Add recipe for go-tag

brantou:

Modify field tags for golang struct fields.

https://github.com/brantou/emacs-go-tag

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/go-tag
+++ b/recipes/go-tag
@@ -1,3 +1,3 @@
 (go-tag
  :repo "brantou/emacs-go-tag"
- :fetcher github :files ("go-tag.el"))
+ :fetcher github)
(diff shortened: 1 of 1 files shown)


purcell:
Thanks! Sorry for the slow reply - I'm currently extremely busy.

- Please remove the `:files` spec from the recipe. It is not needed here, because the default works fine.
- You haven't included a package dependency on `go-mode`.

brantou:
Thanks! 
- I have removed the **:file** spec
- A function from go-mode is called, the name of the function is **go--apply-rcs-patch**.

purcell:
> A function from go-mode is called, the name of the function is go--apply-rcs-patch.

Yes! So where is the `;; Package-Requires` entry for `go-mode`? That's my question! :-)

brantou:
@purcell I'm sorry, it's my fault. I have added **Go-mode** to **Package-Requires**，please review again.

purcell:
Great, thanks!

PR #5132, opened 2017-11-22
Add recipe for elcord

Mstrodl:

elcord allows you to use rich presence for [Discord](https://discordapp.com) to show off emacs! It shows the name of the current buffer and the line you're on out of all the lines in the 

https://github.com/Mstrodl/elcord

I'm the developer

**None needed.**

### Checklist

Please confirm with `x`:


purcell:
Hi! Thanks for this, and sorry for the slow response - I'm currently extremely busy.

Quite a few little things to fix up here before this can be merged. This is code that has been hacked until it works for you, but it won't work reliably for other people for a number of reasons.

- You're modifying `debug-on-error` globally, which you must never do.
- You're modifying `print-escape-newlines` globally, which will likely break other code. Instead, `let`-bind it to the value you want while you're calling functions which are affected by that setting.
- `system-type` is a symbol, not a string, so [this](https://github.com/Mstrodl/elcord/blob/7bfcaf16ac7d323b8728af7b55ce8119ee1f7cde/elcord.el#L122) will not work.
- Instead of `(if (not cond) (progn ...))`, just use `(unless cond ...)`. Always prefer `when` and `unless` to `if` when there is only one branch.
- Use `setq` instead of `setf`, generally. The latter might not even work in newer
- `;;;` is for sections: please look at the file structure produced by using `M-x auto-insert` in an empty elisp file, and follow that structure.
- You're sharing this for other people to use it, but there's no `Commentary` section and little documentation: elisp should be self-documenting as far as possible.
- Either include the license text in the head boilerplate, or include a `;; License: MIT` header.
- Don't use the old deprecated `cl` -- use `cl-lib` instead. (You'd need to depend on `cl-lib` or Emacs 24.3 for that, which `package-lint` will tell you.)
- The `provide` form goes between the rest of the code and before the closing comment, not before code. (It's the form which says, "now this feature is loaded".)
- You're making the code connect as soon as the library is loaded, which is the wrong choice: instead, provide a function which starts the connection, then the user can call it as necessary.
- You're watching every single command in every single buffer, and then counting lines, which can be really expensive in large buffers. Instead, this all seems to be about sharing a single buffer (though it's hard to understand what I could do with the package), so probably you should organise this code as a minor mode instead, so that you can toggle it on/off and associate its code and variables with a specific buffer.

And in general, use `flycheck` and `flycheck-package` and fix all the warnings you find.

Hope that helps!

tarsius:
Friendly ping!

Mstrodl:
Sorry about all the wait, I thought I posted something here. I did make a bunch of the changes you asked for but I'm really not sure how I'd go about making a minor mode for this.

Zulu-Inuoe:
Hello! I actually have a PR active for a fork of this library here:

https://github.com/melpa/melpa/pull/5227

I wasn't aware there was already an issue open for adding the original version.
@Mstrodl you can see my fork [here](https://github.com/Zulu-Inuoe/elcord). I've made elcord into a minor mode, changed it to be timer based rather than every keystroke, and implemented a workaround for Windows.

purcell:
Ah, sorry, I went ahead and merged #5227. I'd rather our recipe point at the most official version, though - @Zulu-Inuoe, would it make sense to submit your changes to @Mstrodl's repo as a pull request? Or perhaps @Mstrodl would like to add @Zulu-Inuoe as a collaborator on that repo.

PR #5137, opened 2017-11-23
Add scp

tszg:

Use the SCP command to transfer files with the remote server

https://github.com/tszg/emacs-scp

I'm the developer

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/scp
+++ b/recipes/scp
@@ -1,0 +1,1 @@
+(scp :repo "tszg/emacs-scp" :fetcher github)
--- a/recipes/sftp
+++ b/recipes/sftp
@@ -1,1 +1,0 @@
-(sftp :repo "tszg/emacs-sftp" :fetcher github)
--- a/recipes/sftp
+++ b/recipes/sftp
@@ -1,0 +1,1 @@
+(sftp :repo "tszg/emacs-sftp" :fetcher github)


purcell:
Thanks! This is probably not the best package name, since `sftp` is a specific program, and this package doesn't use it, which could lead to a lot of confusion.

tszg:
This package USES SCP. Can I change it to SCP

tszg:
Thanks @purcell  for the feedback!
This package uses the scp program, and I have changed the package name to scp.

purcell:
Thanks!

- It seems dangerous to [change the global value of `enable-local-variables`](https://github.com/tszg/emacs-scp/blob/f551fa9ca6549c4db85abcca815a24848a335e5e/scp.el#L37). That whole `scp-locals-init` function seems like it shouldn't need to exist.
- Don't use `org-mode` here: use `special-mode`, which will also bind `q` for you.
- Also, [this line](https://github.com/tszg/emacs-scp/blob/f551fa9ca6549c4db85abcca815a24848a335e5e/scp.el#L33) will modify the keymap every time you run `scp`. Put that line at the top level after the mode declaration instead.
- Don't use [replace-regexp-in-string](https://github.com/tszg/emacs-scp/blob/f551fa9ca6549c4db85abcca815a24848a335e5e/scp.el#L56) when the args are not regexes. Bad things can happen! Either use a different function, or escape the regexps first - there's `regexp-quote` for that.
- Variable names should be `remote-path` rather than `remote_path`.
- When you [build strings with `%s`](https://github.com/tszg/emacs-scp/blob/f551fa9ca6549c4db85abcca815a24848a335e5e/scp.el#L72) you will have a bad time - use `shell-quote-argument` to make interpolated values safe inside command strings.

tszg:
Thanks, I will fix it as soon as possible

tszg:
Thanks for @purcell valuable comments, I made most of the changes, please review.

purcell:
Looks good - welcome to MELPA!

alphapapa:
Just a thought: it would generally be better to use `rsync` than `scp`, especially when transmitting directories or multiple files.

tszg:
Thank you @alphapapa , this is a good idea!

PR #5138, opened 2017-11-23
Add recipe for total-lines-mode

hinrik:

Provides a minor mode that maintains a buffer-local `total-lines` variable, e.g. for use in a mode-line.

https://github.com/hinrik/total-lines

I'm the creator.

### Checklist

### DIFF
--- a/recipes/total-lines
+++ b/recipes/total-lines
@@ -1,0 +1,1 @@
+(total-lines :repo "hinrik/total-lines" :fetcher github)


purcell:
Thanks for this - will go ahead and merge now. Sorry for the slow response!

I suspect that this might get slow in bigger buffers.

hinrik:
For a file several hundred MBs in size, I saw no slowdown upon opening the file (when it counts all the lines to get the inital count). Maybe with a file 1GB in size or more you'd start to notice, or when using a very old hard drive. All subsequent work by the mode is proportional to the size of edits, so the file size doesn't affect it.

purcell:
Nice! :-)

PR #5139, opened 2017-11-24
Add recipe for flycheck-lilypond

hinrik:

Provides syntax checking for LilyPond files.

https://github.com/hinrik/flycheck-lilypond

I'm the creator.

### Checklist

### DIFF
--- a/recipes/flycheck-lilypond
+++ b/recipes/flycheck-lilypond
@@ -1,0 +1,1 @@
+(flycheck-lilypond :fetcher github :repo "hinrik/flycheck-lilypond")


purcell:
Looks good, thanks. It wasn't obvious from the `Commentary` section how people should use it, so you should add docs there. Probably suggest something like `(eval-after-load 'flycheck '(require 'flycheck-lilypond))`, for those not using `use-package`.

purcell:
Also, gah, the `LilyPond-` symbol prefix upstream offends my tastes. :-)

hinrik:
Will do. And yeah, that prefix isn't the best. :P

PR #5143, opened 2017-11-25
Add recipe for difflib.

dieggsy:

Provides helpers for computing deltas between sequences. It's a port of [python's difflib](https://github.com/python/cpython/blob/master/Lib/difflib.py) to emacs-lisp that, in the words of the original library, provides "human friendly" diff tools.

https://github.com/dieggsy/difflib.el

Maintainer

**None needed**

### Checklist

Please confirm with `x`:

- [X] `M-x checkdoc` is happy with my docstrings **(mostly)**

### DIFF
--- a/recipes/difflib
+++ b/recipes/difflib
@@ -1,0 +1,1 @@
+(difflib :fetcher github :repo "dieggsy/difflib.el")


purcell:
Thanks - great idea! Quick feedback:

- `difflib-pythonic-strings` is a bad name for the corresponding variable.
- Does it make sense to autoload [difflib-sequence-matcher](https://github.com/dieggsy/difflib.el/blob/d374dbd2a03b44a9852989b72ea585bc0b1f91e6/difflib.el#L50)? And does that work? i.e. does it allow people to write sequence matchers elsewhere such that `difflib` is not actually loaded until one of its methods is called? If not, you should probably just have users `require` this library and do away with the autoload.
- I'd simply write the body of [this function](https://github.com/dieggsy/difflib.el/blob/d374dbd2a03b44a9852989b72ea585bc0b1f91e6/difflib.el#L706-L707) like `(string-match-p (pat "\s*$") line)`.

dieggsy:
- Changed to `difflib-handle-chars-as-strings`, which is more consistent with the docstring.
- It does work, (the object constructor and a couple other things like type checker get autoloaded) and I think it makes sense since you'd only ever used one of `difflib-sequence-matcher`'s methods on an instance of `difflib-sequence-matcher` itself, so making the sequence-matcher class accessible through an autoload would let people play with the class/methods in their init if they want without needing a `require`.
- Good catch, changed `difflib-is-character-junk-p` and `difflib-is-line-junk-p` to follow the simpler syntax, assuming you meant "or".

purcell:
Excellent. And yes, I forgot the `or`, sorry! Merging now. :-)

PR #5146, opened 2017-11-26
add recipe for ninrod/evil-string-inflection package

ninrod:

provides an evil operator to cycle text objects through various case formats: kebab, camel, snake, UPPER.

https://github.com/ninrod/evil-string-inflection

I'm the author

**none needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/evil-string-inflection
+++ b/recipes/evil-string-inflection
@@ -1,0 +1,1 @@
+(evil-string-inflection :repo "ninrod/evil-string-inflection" :fetcher github)


recipes/snakecamelfy
```diff
@@ -0,0 +1 @@
+(exato :repo "ninrod/snakecamelfy" :fetcher github)
```
tarsius: `exato` has to be replaced.

ninrod: :man_facepalming:

ninrod: Ok, done!

ninrod:
@tarsius, should be fine now.

tarsius:

By the way you should also do that fixing an issue in your recipe submission. So instead of adding a second commit "fix copy/paste leftover", you should just <kbd>c e</kbd> to amend and <kbd>P - f p</kbd> to force push. @purcell that's what you would prefer, right?

ninrod:
@tarsius and @purcell, I've implemented the rebase and squashed the commits. but it seems that there's an issue in travis regarding the build [path](https://travis-ci.org/melpa/melpa/builds/308348271#L1036).

tarsius:
Turns out my fix was not complete... still working on it.

tarsius:
Please try again it should work now.

ninrod:
@tarsius, do you need anything else from me regarding this pull?

tarsius:
No, the tests pass, which means the things that I broke are no longer broken and also that nothing that could be detected by the tests is broken in your submission :wink:

purcell:
Great! Same comments here as in #5149, re. package naming and key customisation.

purcell:
Also, you might consider using the [string-inflection package](`https://github.com/akicho8/string-inflection/blob/e644c8bf2a5e36ddadcfe4de657588c00fd368dd/string-inflection.el) to avoid reinventing this code.

ninrod:
> Also, you might consider using the string-inflection package to avoid reinventing this code.

This was a game changer. I had to rewrite and rename the whole package, got more functionality than I ever hoped and in the process end up with just 4 lines of code. Talk about reuse. 👍 

thank you very much @purcell. I'd appreciate a new review.

PR #5148, opened 2017-11-27
Add sailfish-scratchbox recipe

vityafx:
This recipe provides an easy way to interact with the sailfish sdk from inside the emacs.

https://github.com/vityafx/sailfish-scratchbox.el

I am the maintainer of this project.

None needed.

### Checklist

### DIFF
--- a/recipes/sailfish-scratchbox
+++ b/recipes/sailfish-scratchbox
@@ -1,0 +1,1 @@
+(sailfish-scratchbox :repo "vityafx/sailfish-scratchbox.el" :fetcher github)


tarsius:
Could you please rebase this onto `master` and then force push? (You branched of `master` during the approximately 10 minutes when it had the bug that causes Travis to fail.)

vityafx:
Yes, of course.

vityafx:
@tarsius I don't quite understand where this error comes from? What it was unable to load?

tarsius:
There was a second related issue, that's fixed too now. Please do the rebase dance again.

tarsius:
Hm....

tarsius:
I am going to push some additional commits onto this branch for testing purposes. I'll remove that again when I am done.

tarsius:
I have pushed another fix to `master` and rebase your branch onto that.

tarsius:
And now for something completely different:

* Please release the package under "GPL v3 or later" by using the standard permission statement (which you can copy from a built-in library). Only packages that are released under a license that is compatible with that of Emacs can be added to Melpa.
* Don't put closing parentheses on their own lines.

vityafx:
@tarsius May I license it under apache 2 license? It is GPL compatible as I can see.

tarsius:
Yes, that is okay too. Please put the respective permission statement and also add the line `;; License: Apache-2.0`.

vityafx:
Done. Check it again please.

tarsius:
Oh, I didn't see the license file before. MIT is also okay.

vityafx:
@tarsius cmon! I wanted it to be MIT... So I am moving it back to MIT, lol.

tarsius:
I would just have reverted the commit to avoid a license flip-flop so early in a projects history. (Of course that also teaches the importance of atomic commits. "Change license & move paretheses from their own lines" does two completely unrelated things, which makes reverting just one of them harder.)

> Review please.

I can confirm that this uses a valid license and that the closing parentheses are no longer on separate lines :wink: Steve might have some other suggestions also.

vityafx:
@tarsius lol. Yes, you are right about atomic commits and so on :) I will do this now.

vityafx:
@tarsius I have rewritten the history by pushing commits with `force`. Their are atomic now and there is not move to `Apache2` license in the history. Is there anything else I need to do?

tarsius:
Looks good to me now (and the license already was good before :roll_eyes:) but Steve might have additional suggestions.

vityafx:
@tarsius okay, thank you for spending time reviewing my code anyway.

purcell:
Basically looks good, but you're manually quoting command arguments with `concat` and `'`, so I suspect there are various things that might explode if the interpolated values contain spaces, backslashes, quotes etc: use `shell-quote-argument` instead. Should be ready to merge when that's fixed. :-)

vityafx:
@purcell Could you suggest me doing anything else please? The reason I did that is because the `sdk` is usually an alias to the script which runs scratchbox. According to the manual of Sailfish HADK it must be set in the `~/.profile` but emacs's internal `shell-command` function does not read it and simply can't found the alias. I googled and found that solution which you saw. I'd do this better if I knew.

vityafx:
@purcell I have done what you have advised to me. However, is this the only way to execute a shell command which reads the `.profile`/`.bashrc` files before running anything?

purcell:
Merged now! Thanks, and welcome to MELPA!

purcell:
> However, is this the only way to execute a shell command which reads the .profile/.bashrc files before running anything?

Yes, you probably need the `-i` flag for `bash` to guarantee that certain startup files will be read. There isn't an easier way afaik.

vityafx:
@purcell okay :) thank you for everything very much!

PR #5149, opened 2017-11-28
add recipe for ninrod/evil-replace-with-char

ninrod:

This package provides an evil operator, =zx= to replace all chars of a text object with a char.

https://github.com/ninrod/evil-replace-with-char

I'm the author.

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/evil-replace-with-char
+++ b/recipes/evil-replace-with-char
@@ -1,0 +1,1 @@
+(evil-replace-with-char :repo "ninrod/evil-replace-with-char" :fetcher github)


tarsius:

ninrod:
@tarsius, thanks! I've implemented the rebase.

tarsius:
Turns out the issue still isn't fixed... I'll keep you posted.

ninrod:
hum... seems that the error persists. here's the output of `make recipes/replace-with-char`:

```sh
melpa $ make recipes/replace-with-char
 • Building recipe replace-with-char ...
/usr/bin/timeout -k 60 600 emacs --no-site-file --batch -L package-build -l package-build --eval "(let ((package-build-stable nil) (package-bu
ild-write-melpa-badge-images t) (package-build-archive-dir (expand-file-name \"./packages\" package-build--melpa-base))) (package-build-archiv
e 'replace-with-char))"                                                                                                                      
Cannot open load file: É um diretório, /home/ninrod/code/emacs/melpa/package-build
make: [Makefile:92: recipes/replace-with-char] Error 255 (ignored)
ls: não foi possível acessar './packages/replace-with-char-*': Arquivo ou diretório não encontrado
 ✓ Wrote  
 Sleeping for 0 ...
sleep 0

melpa $ 
```sh

ninrod:
Hi @tarsius, the tests seem to be running now. Is there anything else I can do?

tarsius:
I only occasionally review submitted recipes and their packages. I am not currently in the mood to do it now.

purcell:
Thanks for this. Quick feedback:
- The custom group is [apparently its own child](https://github.com/ninrod/replace-with-char/blob/dbe4516fc23a93b501e02d51d893d3747e489ba9/replace-with-char.el#L46) :-/
- Call this package `evil-replace-with-char`, please. There's no indication otherwise that this is for Evil users.
- The customisation isn't going to work, because if someone sets the key after this package has been loaded, it will never take effect. Customisable keymaps are the only reliable and consistently-unsurprising way to allow users to customise keys. Instead, just tell users to `(define-key evil-normal-state-map "zx" 'evil-operator-replace-with-char)`.
- I expect that `evil-define-operator` supports autoloading, so you might want to add an `;;;###autoload` marker comment for the definition.

ninrod:
Hey thanks for the tips @purcell! You're absolutelly right. I'll implement those.

ninrod:
@purcell, I've implemented all your recommendations. ❤️

purcell:
Great - thanks for your patience! Merged now.

PR #5151, opened 2017-11-28
Add bart-mode

mschuldt:

   Display real-time departure information for the Bay Area Rapid Transit (BART) system.

   https://github.com/mschuldt/bart-mode

  I am the author

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/bart-mode
+++ b/recipes/bart-mode
@@ -1,0 +1,3 @@
+(bart-mode :fetcher github
+           :repo "mschuldt/bart-mode")
+


tarsius:

mschuldt:
rebased

mschuldt:
Is there a problem with this?

tarsius:
I don't usually perform the actual review and @purcell probably just overlooked this.

purcell:
Yep, this fell between the cracks, sorry: December & Jan have been crazy for me. Some feedback, then:
- Firstly, nice work!
- Don't hard-code completion to use `ido`. Instead, use `completing-read`, which will be overridden automatically by the user's preferred completion system.
- Use `special-mode` as the parent mode rather than `fundamental-mode`.
- You shouldn't need to set the local keymap [here](https://github.com/mschuldt/bart-mode/blob/f857a9e54801bd057ee28c938f8f08e2ca18eb17/bart-mode.el#L297), because `define-derived-mode` will do it *if* the mode name matches up with the map name -- which you could easily arrange: it's not necessary for the mode name to be `bart--mode` instead of `bart-mode`, and indeed that choice is unusual.
- Setting explicit colours and other face attributes is bad practice. You should instead define customisable faces with dark/light background variants, or (better) that `:inherit` built-in faces by default (e.g. `header`, `warning`, `success`) so that they automatically fit in with users' themes. You should particularly avoid setting font sizes and weights.

Hope that helps!

purcell:
Also:
- `cl-lib` has `cl-caddar`, so you don't need to roll your own.
- Don't wrangle windows explicitly for your own preferences: you'll just annoy somebody else. Just use `display-buffer`, which allows customisation of window selection.

mschuldt:
@purcell Thank you very much for you review, I believe I have addressed all the issues.

I've created a new option to enable the window wrangling in the way I like it, but by default it is disabled and uses display-buffer instead

purcell:
Great, merged! Welcome to MELPA. :-)

PR #5152, opened 2017-11-28
Add recipe for counsel-world-clock

kchenphy:

Displaying world clock using Ivy.

https://github.com/kchenphy/counsel-world-clock

maintainer

NA
### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/counsel-world-clock
+++ b/recipes/counsel-world-clock
@@ -1,0 +1,1 @@
+(counsel-world-clock :repo "kchenphy/counsel-world-clock" :fetcher github)


tarsius:

kchenphy:
rebased

purcell:
Thanks, this is neat! Note that the hard-coded list of zones will not necessarily be available on every system, and they will also change over time and get out of sync with your code. I'm not aware of a standard way to get these from Emacs, but it's something to think about: we should arguably have a `tzdata` package or similar to avoid these issues.

I'll go ahead and merge this, but you should add an `;;;###autoload` comment just before the `counsel-world-clock` function.

Welcome to MELPA!

kchenphy:
Thanks @purcell ! I agree with you that an `tzdata` library is in need.

PR #5153, opened 2017-11-28
Add package celestial-mode-line.

ecraven:

Show lunar phase and sunrise/-set time in modeline

https://github.com/ecraven/celestial-mode-line

I am the author.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/celestial-mode-line
+++ b/recipes/celestial-mode-line
@@ -1,0 +1,1 @@
+(celestial-mode-line :fetcher github :repo "ecraven/celestial-mode-line")


purcell:
Looks good. Couple of thoughts:
- Figure out what to do about that [commented-out defvar](https://github.com/ecraven/celestial-mode-line/blob/master/celestial-mode-line.el#L42). Either you need it or you don't! Perhaps put those non-unicode examples in the `Commentary` instead.
- Why the [sit-for](https://github.com/ecraven/celestial-mode-line/blob/master/celestial-mode-line.el#L158)? If it's related to getting the modeline change to appear, consider calling `force-mode-line-update` judiciously.
- You should probably include an `;;;###autoload` cookie comment for the mode definition, so users can enable this without first explicitly loading the package.

ecraven:
Thanks, I've addressed all of those points:
- The commented-out defvar has moved into the Commentary.
- I've removed `sit-for', the data is not urgent, so even if it isn't displayed right away, nothing bad happens.
- I've included the autoload cookie.
- I've also increased the time for the timer to 60 seconds instead of 20, it really isn't that important that the numbers change (they only change two to three times a day anyway..).
Thanks for looking through it so thoroughly, I hope it's useful to someone, I like seeing the data ;)

purcell:
Great, thanks!

PR #5154, opened 2017-11-28
Specify package-build.el extension for -l flag

labdsf:
I keep getting file not found error without that


tarsius:
Thanks! That fixed building on Melpa itself too.

tarsius:
Why does it not work when the extension is removed? It would be nice to load `package-build.elc` if available.

PR #5155, opened 2017-11-28
Add (or split off) evil-org-agenda

Somelauw:
Thanks to the good work of @Ambrevar, evil bindings have been created for org-agenda.
It's currently on https://github.com/Somelauw/evil-org-mode, but can be used on its own.
The question is whether it should be put in a different recipe or can stay in evil-org.
This pull request puts evil-org-agenda in its own recipe, so it can be installed without evil-org.el
Personally, I'm fine with bundling the packages together, but from the README it seems they should be separated.

Integrate emacs-evil with org-agenda

https://github.com/Somelauw/evil-org-mode

Maintainer

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/evil-org-agenda
+++ b/recipes/evil-org-agenda
@@ -1,0 +1,3 @@
+(evil-org :repo "Somelauw/evil-org-mode"
+          :fetcher github
+          :files ("evil-org-agenda.el"))


Ambrevar:
While modularity is a great idea, in this case I would find it surprising if there would ever be a demand for org-agenda in Evil mode and org without it.

Having two separate packages is a tiny little more work on the user end.  Maybe we would be better off making this completely seemless.

At the beginning I was against bundling all bindings together in evil-collection, I wanted split packages.  Considering where evil-collection is now, I understand this would have been a mistake:  having all bindings together is so much simpler.  And the users are still free to choose whatever bindings they want to activate.

recipes/evil-org-agenda
```diff
@@ -0,0 +1,3 @@
+(evil-org :repo "Somelauw/evil-org-mode"
```
tarsius: This must say `(evil-org-agenda ...)`.

tarsius:
Closing this because the author of the package(s) has spoken out against the proposed change.

PR #5156, opened 2017-11-29
Add recipe for helm-z

yynozk:

Show z directory list with helm.el support.

https://github.com/yynozk/helm-z

maintainer

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/helm-z
+++ b/recipes/helm-z
@@ -1,0 +1,1 @@
+(helm-z :fetcher github :repo "yynozk/helm-z")


purcell:
Thanks! Quick feedback:

[This](https://github.com/yynozk/helm-z/blob/6319c7ec678a59b73d2e7d54d548bfdad1c88a73/helm-z.el#L35) will break horribly when the directory contains spaces. Use `shell-quote-argument`.

In general, note that this will not work reliably unless the shell is invoked by Emacs in such a way that the user's shell config files are sourced in order to change the definition of `cd`, so just because it works for you does not mean it will work for all users. For example, some shell config files are only executed for login or interactive shells. To make this failure more obvious, you might like to change the `"cd"` command so that it fails if `z` is not enabled, e.g. by testing for the `$Z_DATA` variable first.

yynozk:
@purcell Thank you for your feedback! I've implemented your recommendations.

purcell:
Great, thanks. Welcome to MELPA!

PR #5158, opened 2017-11-29
Add package no-emoji

ecraven:

Replace emoji by :emoji-name:

https://github.com/ecraven/no-emoji

I am the author.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/no-emoji
+++ b/recipes/no-emoji
@@ -1,0 +1,1 @@
+(no-emoji :fetcher github :repo "ecraven/no-emoji")


purcell:
Thanks, looks good. Quick feedback:

- You should `;;;###autoload` the minor mode
- Don't define a face with only one colour: it is certain to look wrong on many/most users' screens. Instead, best practice is to define a face which `:inherit`s a built-in face, and then the user can customise it freely, but it will work with any theme if uncustomised.
- The code replaces the display table unconditionally, so will clash with other libraries which rely on display tables, e.g. my `page-break-lines` mode. It would be better to create a new display table only if there is not one already set.

ecraven:
Thanks for the thorough review!

I hope I have addressed all of these points:
- Autoload cookie added.
- The face now inherits from `dired-header'.
- I've taken the liberty to look at your code, and doing the same thing. I update an existing display table (though this will overwrite other modes' display-table settings if they conflict).
I've also refactored into a new function `no-emoji-displayable-unicode-name', so this can be re-defun'd if the actual format (:name:) is not what the user wishes to see.

Thanks for taking the time to review this ;)

purcell:
Great - thanks for your patience! Merged now.

PR #5160, opened 2017-11-29
Add a recipe for smalltalk-mode

Wilfred:

smalltalk-mode is part of GNU Smalltalk, and it provides basic syntax
highlighting for Smalltalk programs.

http://smalltalk.gnu.org/

I'm an enthusiastic user.

All done.

### Checklist

### DIFF
--- a/recipes/smalltalk-mode
+++ b/recipes/smalltalk-mode
@@ -1,0 +1,3 @@
+(smalltalk-mode
+ :url "git://git.sv.gnu.org/smalltalk.git" :fetcher git
+ :files ("smalltalk-mode.el"))


Wilfred:
Woops, looks like it still need a package version.

purcell:
Thanks! Is there an `https` SCM URL available, please? We strongly prefer secure transports. Also, if you could dig up a link directly to the `.el` file that would be appreciated, so that I don't need to check out the whole repo locally in order to look it over.

Wilfred:
You can view the .el file on Savannah: http://git.savannah.gnu.org/cgit/smalltalk.git/tree/smalltalk-mode.el or the GitHub mirror: https://github.com/gnu-smalltalk/smalltalk/blob/master/smalltalk-mode.el

There is an HTTPS URL, but the Savannah docs say it's slower: http://savannah.gnu.org/maintenance/UsingGit/ -- would you like me to update the PR to use that?

purcell:
> There is an HTTPS URL, but the Savannah docs say it's slower: http://savannah.gnu.org/maintenance/UsingGit/ -- would you like me to update the PR to use that?

Yes please.

purcell:
To the specifics of the code (thanks for the direct link!):

- There are a lot of `C-c letter` bindings, which are reserved for end users - the elisp guide says that great trouble was taken to eliminate those bindings from the core libraries, and stresses that libraries should not use them. Any way we can get this updated upstream? Should be an easy sell.
- The major mode function should be rewritten to use `define-derived-mode`, which has long been available, and would eliminate lots of the boilerplate in that function.

Looks good otherwise. :-)

purcell:
^ ping @Wilfred

tarsius:
Friendly ping!

Wilfred:
Sorry, this fell off my backlog. I'll follow up upstream to get this shipshape :)

DamienCassou:
ping :-)

Wilfred:
To confirm I've understood correctly:

```
!        smalltalk-bang
:        smalltalk-colon
C-M-a    smalltalk-begin-of-defun
C-M-b    smalltalk-backward-sexp
C-M-f    smalltalk-forward-sexp
C-c C-a  smalltalk-begin-of-defun
C-c C-b  smalltalk-backward-sexp
C-c C-e  smalltalk-end-of-defun
C-c C-f  smalltalk-forward-sexp
C-c C-n  smalltalk-goto-next-keyword
C-c C-p  smalltalk-goto-previous-keyword
C-c d    smalltalk-doit
C-c f    smalltalk-filein-buffer
C-c m    gst
C-c p    smalltalk-print
C-c q    smalltalk-quit
C-c s    smalltalk-snapshot
C-c t c  smalltalk-class-template
C-c t i  smalltalk-instance-template
C-c t p  smalltalk-private-template
C-j      smalltalk-newline-and-indent
```

Of these, `C-c C-a` etc is fine, correct? I just need to find a good new prefix for `C-c d`, `C-c t` and so on, then ask upstream nicely to take my patch. I'm not sure what would be a good prefix here.

purcell:
Yes, exactly. You can share with upstream the link to the [keybinding conventions doc](https://www.gnu.org/software/emacs/manual/html_node/elisp/Key-Binding-Conventions.html), which talks about how it was a lot of work to eliminate thes bindings in the Emacs core libraries. I don't have a great alternative prefix suggestion, sorry.

DamienCassou:
Should we close this PR for now?

PR #5162, opened 2017-11-30
Add recipe for yoficator

labdsf:

Package allows to interactively replace letter «е» with letter «ё» in Russian texts. See [Wikipedia](https://en.wikipedia.org/wiki/Yoficator) for more info on this type of programs.

https://gitlab.com/link2xt/yoficator

I have just created this package on GitLab. It is based on existing code from http://python.anabar.ru/yo.htm I am going to maintain it, so this repo is now official, at least for my fork (named yoficator, not yo).

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/yoficator
+++ b/recipes/yoficator
@@ -1,0 +1,1 @@
+(yoficator :fetcher gitlab :repo "link2xt/yoficator" :files ("yoficator.el" "yo.txt"))
--- a/recipes/yoficator
+++ b/recipes/yoficator
@@ -1,1 +1,1 @@
-(yoficator :fetcher gitlab :repo "link2xt/yoficator" :files ("yoficator.el" "yo.txt"))
+(yoficator :fetcher gitlab :repo "link2xt/yoficator" :files (:defaults "yo.txt"))


purcell:
Great, thanks!

alphapapa:
Hi,

A few things I noticed:

* https://gitlab.com/link2xt/yoficator/blob/master/yoficator.el#L57  Instead of making a buffer and killing it, you can use `with-temp-buffer`.
* https://gitlab.com/link2xt/yoficator/blob/master/yoficator.el#L80  The docstring for `save-restriction` says:

> Note: if you are using both ‘save-excursion’ and ‘save-restriction’, use ‘save-excursion’ outermost:

* https://gitlab.com/link2xt/yoficator/blob/master/yoficator.el#L84  The `concat`/`mapconcat` could be replaced with this, which will also create a more optimized regexp:
```
(rx-to-string `(optional (or ,@yoficator-cutting-strings)))
```
* https://gitlab.com/link2xt/yoficator/blob/master/yoficator.el#L99  `when-let*` and/or `if-let*` could be used to simplify the `if`, `set`, `when`, `if` chain.

purcell:
> (rx-to-string `(optional (or ,@yoficator-cutting-strings)))

Or just `regexp-opt`.

PR #5165, opened 2017-12-02
Add recipe for eterm-256color.

dieggsy:

Adds support for customizable 256 colors in ansi-term and term.

https://github.com/dieggsy/eterm-256color

Maintainer

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/eterm-256color
+++ b/recipes/eterm-256color
@@ -1,0 +1,1 @@
+(eterm-256color :fetcher github :repo "dieggsy/eterm-256color" :files (:defaults "eterm-256color.ti"))


dieggsy:
Needs a bit more work.

dieggsy:
So there's a terminfo file that needs to be compiled in this package - it currently does a slightly hackish "check if the compiled result exists, and if not compile" but I think this is suboptimal - it only really needs to happen once, and the current method makes the very first usage of this package slightly unintuitive (term needs to be restarted). 

 I feel like it would be better to compile this with a build specification on package install (like `pdf-tools` has), though i'm not familiar with what the options are, and the README says to submit an issue if more complex options are needed.

purcell:
Cool. You should probably make the initial setup (compilation) step a bit more interactive, so that you can show a message to the user about needing to restart term. Use a `y-or-n-p` to confirm compilation, for example.

dieggsy:
Good idea, working on it. Thanks!

PR #5168, opened 2017-12-04
added: auto-sudoedit

ncaq:

auto sudo edit by tramp when find not writable file.

<https://github.com/ncaq/auto-sudoedit>

maintainer

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/auto-sudoedit
+++ b/recipes/auto-sudoedit
@@ -1,0 +1,1 @@
+(auto-sudoedit :repo "ncaq/auto-sudoedit" :fetcher github)


purcell:
Thanks - this looks useful!

- You should not include the `.elc` and `-autoloads.el` files in your repository. `package.el` will generate them locally when the package is installed, and your versions will not be compatible with all Emacs versions.
- The autoloads enable this functionality as soon as the package is installed, which is not allowed. Instead, provide either an autoloaded function which the user can call to enable the behaviour, or (better) provide an autoloaded global minor mode which can be toggled on/off.
- Just use `unless` [here](https://github.com/ncaq/auto-sudoedit/blob/0d13aed9ab39853a7b1bfd49a92816887ca2c7f5/auto-sudoedit.el#L41-L44).

purcell:
Also, thanks for your patience - I've had a busy month and am a long way behind with my MELPA work.

ncaq:
Thank you for writing reviews and comments while you are busy.
I am sorry I overlooked the rudimentary rule of not including binary.
I modified the repository according to your comments.
I have defined a minor mode and enabled it by default, is there no problem with this rule?
Please review again.

purcell:
Looks good, but yes - you shouldn't auto-enable the minor mode. Think of it like this: installable packages are an extension of the Emacs standard library, and it should be possible to have lots of packages installed without them changing the behaviour of the system. If every package in the Emacs standard library were auto-enabled, Emacs wouldn't work very well. :-)

ncaq:
Thank you for pointing out.

~~~el
(auto-sudoedit-mode 1)
~~~

Changed to a format that requires to enable minor mode as above.

purcell:
Great, thanks - merged!

PR #5170, opened 2017-12-04
added: symbolword-mode

ncaq:

modify word split.
symbolword-mode determine word split for symbol.
for instance hyphen.

<https://github.com/ncaq/symbolword-mode>

maintainer

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/symbolword-mode
+++ b/recipes/symbolword-mode
@@ -1,0 +1,1 @@
+(symbolword-mode :repo "ncaq/symbolword-mode" :fetcher github)


purcell:
Thanks! Quick feedback:

- Is this different from `subword-mode`? It looks similar, and I can't easily understand what it does from the Commentary, README or docstrings, so I'm not sure that any potential users would recognise that this package might be helpful to them. Perhaps some examples would be helpful?
- What does "Grab keys" mean [here](https://github.com/ncaq/symbolword-mode/blob/master/symbolword-mode.el#L149)? Should that instead be the short description of the mode?
- You should probably add and `;;;###autoload` comment for the `easy-mmode-define-minor-mode` form.
- The `.elc` files should be removed from the repo.

tarsius:
Friendly ping!

ncaq:
pong.

I was busy.
I will work on it.

tarsius:
This is from last year... I will close it soon.

ncaq:
I remember it.
English is too difficult and I am struggling to write an explanation.

ncaq:
It is very difficult to write a README in English.
It is different from subword-mode but it is difficult to explain it in English…

tarsius:
> It is very difficult to write a README in English.

We have to be careful not to exclude anybody just because they don't know English well enough to write documentation in that language. So I am merging this.

But of course the lack of English documentation means that this package won't find many users who do not understand Japanese. I recommend you do the following:

1. Ask for someone to translate the documentation for you. Add a note to the readme and ask on some Japanese Emacs forums.
2. As @purcell recommended above, add *examples* ... in Lisp.

Additionally you should remove `symbolword-mode.elc`. Such files should never be tracked in Git.

ncaq:
thanks.
I will delete elc file.
least.

PR #5172, opened 2017-12-04
Add recipe for exwm-surf

ecraven:
Interface for Surf (surf.suckless.org) under exwm

https://github.com/ecraven/exwm-surf

I am the maintainer

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/exwm-surf
+++ b/recipes/exwm-surf
@@ -1,0 +1,1 @@
+(exwm-surf :fetcher github :repo "ecraven/exwm-surf")


purcell:

- You're modifying the global version of `post-command-hook`, which is generally not what you want. Same for the minibuffer hooks: see if you can set them locally instead, or `let`-bind modified versions of the hooks while calling `read-string` etc.
- Your local keymap handling (and possibly some other book-keeping) would be simpler if you defined a local minor mode which you could toggle as necessary.

In any case, I'll go ahead and merge this in its current state.

PR #5179, opened 2017-12-05
Add recipe for counsel-notmuch

fuxialexander:

Search emails in Notmuch asynchronously with Ivy

https://github.com/fuxialexander/counsel-notmuch

I'm the maintainer.

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/counsel-notmuch
+++ b/recipes/counsel-notmuch
@@ -1,0 +1,2 @@
+(counsel-notmuch :fetcher github
+              :repo "fuxialexander/counsel-notmuch")


purcell:
Thanks, and sorry about the delay. Some feedback:
- There's no `defcustom` form for the `counsel-notmuch` group - you should add one!
- [Here](https://github.com/fuxialexander/counsel-notmuch/blob/5bea938befa8512be5a08d0bbcbc7b75a29d1911/counsel-notmuch.el#L55) you should use `shell-quote-argument` to make the search string safe for insertion.
- Don't hard-code the default path to `notmuch`. Just use `"notmuch"` as the default, and it will work for most people out-of-the-box.
- The use of `swiper-cleanup` [here](https://github.com/fuxialexander/counsel-notmuch/blob/5bea938befa8512be5a08d0bbcbc7b75a29d1911/counsel-notmuch.el#L115) is a bit strange, and feels like it might be brittle, even if that function (currently) does what you need.

fuxialexander:
Thank you for your comments! It's very valuable for a elisp newbie like me 😁.  I've modified the code accordingly.

purcell:
Great, merged - thanks!

PR #5183, opened 2017-12-06
use-package recipe improvements

nickmccurdy:
Building on #5167, #5173, and #5181, I wanted to make use-package's file list easier to maintain (since there are multiple packages in the same repository) while adding some missing documentation files.

A use-package declaration for simplifying your .emacs

https://github.com/jwiegley/use-package

Contributor

Some discussion with @jwiegley in https://gitter.im/use-package/Lobby.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/use-package
+++ b/recipes/use-package
@@ -1,11 +1,11 @@
 (use-package
     :fetcher github
     :repo "jwiegley/use-package"
-    :files (:defaults
+    :files ("use-package*.el"
+            "use-package*.texi"
             "LICENSE"
             "*.md"
             (:exclude
-             "bind-key.el"
-             "bind-chord.el"
              "use-package-chords.el"
-             "use-package-ensure-system-package.el")))
+             "use-package-ensure-system-package.el"
+             "use-package-tests.el")))
--- a/recipes/use-package
+++ b/recipes/use-package
@@ -2,6 +2,8 @@
     :fetcher github
     :repo "jwiegley/use-package"
     :files ("use-package*.el"
+            "dir"
+            "use-package*.info"
             "use-package*.texi"
             "LICENSE"
             "*.md"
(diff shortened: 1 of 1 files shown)


nickmccurdy:
I tested this now, but even after adding `dir` and `"use-package*.info`, `use-package.info` is still not included. Can someone please help me figure this out?

recipes/use-package
```diff
@@ -1,7 +1,13 @@
...
-    :files (:defaults (:exclude "bind-key.el"
-                                "bind-chord.el"
-                                "use-package-chords.el"
-                                "use-package-ensure-system-package.el")))
+    :files ("use-package*.el"
+            "dir"
+            "use-package*.info"
+            "use-package*.texi"
+            "LICENSE"
+            "*.md"
```
zonuexe: Generally, `LICENSE` and `*.md` files etc are not included.

Default files spec is here:
https://github.com/melpa/melpa/blob/62271788f20919a741acb1e0477a823d300e55ee/package-build/package-build.el#L138-L143

nickmccurdy: Thanks I'm aware, but @jwiegley requested that we add them anyway in case someone wants to cd into `.emacs.d/elpa` and read it before we finish the texinfo docs. Is that ok or do we need to stick with the convention of excluding `LICENSE` and `*.md`?

jwiegley: If there's a standard of not including them, then let's just not include them; even more incentive to get the manual done.

nickmccurdy: Removed.

tarsius:
> I tested this now, but even after adding `dir` and `"use-package*.info`, `use-package.info` is still not included. Can someone please help me figure this out?

Why not just use `use-package.info` instead?

purcell:
There is neither a `dir` nor a `use-package.info` in the upstream repo, so `use-package.texi` or `*.texi` should be sufficient.

tarsius:
The current recipe gets the job done.

PR #5185, opened 2017-12-09
Added a recipe for helm-selected.el

takaxp:

This package is an helm extension for selected.el. selected.el provides a combined keymap with associated commands for the selected region like a popup menu when you right-click on a selected word in your document. The menu, however, will not be displayed in an emacs buffer. So this package provides a visible menu with command candidates by supporting of helm interface.

https://github.com/takaxp/helm-selected

the maintainer

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/helm-selected
+++ b/recipes/helm-selected
@@ -1,0 +1,1 @@
+(helm-selected :fetcher github :repo "takaxp/helm-selected")
--- a/recipes/yatex
+++ b/recipes/yatex
@@ -1,0 +1,4 @@
+(yatex
+ :fetcher hg
+ :url "https://www.yatex.org/hgrepos/yatex/"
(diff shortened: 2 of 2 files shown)


takaxp:
I don't know why the Travis get so angry...
The log message says "Cask could not be bootstrapped. Try again later, or report an issue at https://github.com/cask/cask/issues".
Any thoughts what I should do next?

purcell:
Thanks, looks good. These days you should be using `cl-lib` instead of the deprecated `cl` library. It's generally just a matter of changing the `require` line and using `cl-loop` instead of `loop` etc.

purcell:
Also, sorry for the slow turnaround.

takaxp:
Thank you for the practical comment! I'll change it soon.
And I don't think this merge is slow :) thanks!

PR #5187, opened 2017-12-11
Add evil-stateful

terlar:
An Emacs minor mode that is intended as an extension with convenience functions
for [evil-mode](https://github.com/emacs-evil/evil). It adds the ability to attach functions to major-mode state
changes. For example if you are using markdown-mode you might want to hide
markup on `evil-normal-state` entry, but show markup on `evil-insert-state`
entry.

https://github.com/terlar/evil-stateful

Maintainer

None needed

### Checklist

Please confirm with `x`:

Not sure how to satisfy the disambiguate `evil-stateful-mode` warning as I am not sure what the mode counts as, e.g. function/command/variable/option/symbol.

### DIFF
--- a/recipes/evil-stateful
+++ b/recipes/evil-stateful
@@ -1,0 +1,1 @@
+(evil-stateful :fetcher github :repo "terlar/evil-stateful")


purcell:
It is a bit weird to make a package which provides an idiosyncratic wrapper for the hooks that `evil` already exposes, so I am pretty much -1 on this. You're just reinventing hooks, and users would be better served by just using the hooks directly. Or am I missing something? (In any case, you're modifying the global evil hooks in a local minor mode, so separate buffers will interfere with each other unless fixed.)

terlar:
Thank you for the feedback and perhaps I have been disillusioned by trying to extract common patterns from my own config. The "problem" that is being solved here is to provide a simple way to add mode-specific state-change hooks.

This can of course be achieved by having:

```elisp
(defun markdown-mode-on-insert-state-entry ()
  (markdown-toggle-markup-hiding 0))
(defun markdown-mode-on-normal-state-entry ()
  (markdown-toggle-markup-hiding 1))

(add-hook 'markdown-mode-hook
          (lambda ()
            (add-hook 'evil-insert-state-entry-hook 'markdown-mode-on-insert-state-entry nil t)
            (add-hook 'evil-normal-state-entry-hook 'markdown-mode-on-normal-state-entry nil t)))
```

When it comes to the implementation I guess it should not be a minor mode after all. As it relies on only having one hook bound to the state-entries and inside the function check if there is a buffer-local specified function to run for that hook.

I was hoping this would enable more usage of the pattern to toggle functionality based on the evil state for modes which I have found to be really useful myself. But perhaps as it is now, it does not provide enough value in itself.

purcell:
Yeah, I think it's probably over-extraction of patterns. The posted snippet is quite concise and idiomatic, and if you defined that `lambda` as a named function, it would be trivial to add it to `gfm-mode-hook` too. At that point, you really aren't saving much code using this library. I guess that if I were in your position, I'd probably make a helper like this (untested) for my local config:

```el
(defun on-evil-state (state f)
  "Return a function that will add `f' to `evil-insert-state-entry-hook' locally when called.
  (lambda () (add-hook (intern (concat "evil-" (symbol-name state) "-state-entry-hook")) f nil t)))
```
And then your config could look a bit like:
```
(dolist (hook '(markdown-mode-hook gfm-mode-hook))
  (add-hook hook (on-evil-state 'insert 'markdown-mode-on-insert-state-entry))
  (add-hook hook (on-evil-state 'normal 'markdown-mode-on-normal-state-entry)))
```
which is a bit more composable than an ad-hoc DSL.

For now, I'm going to decline this package, then. Thanks for the discussion!

PR #5188, opened 2017-12-11
added rum-mode

avelino:
rum language is a free software environment for statistical computing
writing in golang

 Emacs major mode for editing Rum 

https://github.com/rumlang/rum-mode

@rumlang authors 

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/rum-mode
+++ b/recipes/rum-mode
@@ -1,0 +1,3 @@
+(rum-mode
+ :repo "rumlang/rum-mode"
+ :fetcher github)


purcell:
Thanks, and sorry for the slow response. Quick feedback:

- `cl` is deprecated, but unused here, so you can/should remove the `require`. If you need something like `cl` in future, use `cl-lib`.
- You can use `regexp-opt` to tidy up the long regexp with all the backslashes.
- Don't use `$` to terminate the filename regexp: it should be `\\'`, which matches the end of input.

Hope that helps!

avelino:
@purcell thanks feedback (Improvements is always valid), review pls 😄

purcell:
Thanks. The last point of feedback is still outstanding, and you should also add an `;;;###autoload` for your `define-derived-mode` form. (Also, as noted, `cl`/`cl-lib` isn't actually _needed_ here, so you can just drop the import/dependency.)

tarsius:
Friendly ping!

avelino:
Sorry delay
`Don't use $ to terminate the filename regexp: it should be \\', which matches the end of input.`
I couldn't do it, can you help me?

tarsius:
This doesn't work?

```diff
diff --git a/rum-mode.el b/rum-mode.el
index 095739a..a3a241f 100644
--- a/rum-mode.el
+++ b/rum-mode.el
@@ -72,7 +72,7 @@ (define-derived-mode rum-mode lisp-mode "rum"
   (set (make-local-variable 'font-lock-defaults) '(rum-font-lock-keywords)))
 
 ;;;###autoload
-(add-to-list 'auto-mode-alist '("\\.rum$" . rum-mode))
+(add-to-list 'auto-mode-alist '("\\.rum\\'" . rum-mode))
 
 (provide 'rum-mode)
 ;;; rum-mode.el ends here
```

avelino:
Sorry 😄 , send to master

tarsius:
> you should also add an `;;;###autoload` for your `define-derived-mode` form. (Also, as noted, `cl`/`cl-lib` isn't actually _needed_ here, so you can just drop the import/dependency.)

Please do those two things too.

avelino:
update guys

purcell:
Looks good - merged, thanks. Feel free to add an `;;;###autoload` cookie comment for the `auto-mode-alist` entry.

PR #5190, opened 2017-12-12
Add spiral recipe

volrath:

SPIRAL is a Clojure IDE for Emacs, based on UNREPL.

https://github.com/Unrepl/spiral

Author & Maintainer

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/spiral
+++ b/recipes/spiral
@@ -1,0 +1,10 @@
+(spiral :repo "unrepl/spiral"
+        :fetcher github
+        :files (:defaults
+                "blob.clj"
+                "tools"
+                "parseclj/parseclj.el"
+                "parseclj/parseedn.el"
+                "parseclj/parseclj-parser.el"
+                "parseclj/parseclj-lex.el"
+                "parseclj/parseclj-ast.el"))


purcell:

- Have you tested this locally? The files in `parseclj` will be copied to the root of the package, given the current recipe, and it's not obvious that that was your intent.
- The `Emacs` in `;;; unrepl.el --- Emacs UNREPL Client` is redundant because all Emacs packages are for Emacs. I suggest `Clojure UNREPL Client` instead, so that the package will appear in searches for Clojure.

Looks good otherwise!

volrath:
@purcell 

> Thanks, and sorry for the slow response.

No problem at all, thank you for taking the time to review!

> Have you tested this locally? The files in parseclj will be copied to the root of the package, given the current recipe, and it's not obvious that that was your intent.

Yes I have.  My intention with this is to be able to require my own copy of `parseclj` from within the `unrepl-*.el` files,  The best way I found to do this was to add it to the root of the package since this would mean that `parseclj` would already be in the `load-path`.

This is a temporary measure while we get `parseclj` [ready for its release](https://github.com/lambdaisland/parseclj/issues/8), but as right now, unrepl.el's internal copy of `parseclj` has everything it needs.  The idea of having it as a submodule (instead of just copying the files directly in the root project) is so that I can commit and PR fixes/changes that I find along the way.

I also thought on using `require`'s second parameter, something like `(require 'parseclj "parseclj/parseclj.el")` but that would also mean that I would have to change `parseclj`'s internal requires.

Maybe I'm missing a better way to accomplish what I want?

> The Emacs in ;;; unrepl.el --- Emacs UNREPL Client is redundant because all Emacs packages are for Emacs. I suggest Clojure UNREPL Client instead, so that the package will appear in searches for Clojure.

Makes complete sense! [I updated it](https://github.com/Unrepl/unrepl.el/commit/516ace167a2d9cebb00ce072c366d9a565eb71fb).

Thanks!

volrath:
@purcell I've been thinking on renaming this project. Is there a protocol to follow to do such thing? or maybe it's better that I close this issue and PR a new one with the new name?

purcell:
> @purcell I've been thinking on renaming this project. Is there a protocol to follow to do such thing? or maybe it's better that I close this issue and PR a new one with the new name?

If you want to do that, just rename the file, push a modified recipe here, then ping to confirm it's done. But `unrepl` seems like the ideal name, no?

volrath:
Ok I will.

unrepl.el is ok, but I think it's better for this project to follow @bbatsov's [advice](https://github.com/Unrepl/unrepl.el/issues/4).  It helps to avoid confusion between the protocol and the Emacs IDE.

purcell:
Fair enough: I hadn't really thought about the potential for confusion when a significant fraction of the users of `x` are using it via `x.el`, because that's rarely the case. I can see it might well apply here though. :-)

volrath:
@purcell I finished the renaming to SPIRAL, and fixed the recipe.

I'm not sure why the build failed though, it seems it couldn't contact MELPA. I'm not able to re-run it =/

bbatsov:
I've seen this error in the past https://github.com/cask/cask/issues/373 I don't think that's anything related to your recipe.

purcell:
Cool, looks good. Re. the `parseclj` path, the current solution makes sense: if you weren't actively planning to make it a separate package then I'd be concerned about those files not being namespaced under `spiral-`, so definitely go ahead and package them up when you get chance. :-)

volrath:
Great! thanks @purcell!

tarsius:
> This is a temporary measure while we get parseclj ready for its release,

Do you have any estimate on how long that will take? Bundling third-party libraries is always problematic, especially when patched.

Also, you depend on `highlight` which is a package from the Emacswiki. We are no longer updating packages from there for security reasons and will eventually remove the frozen versions too. You should probably investigate whether you can drop this dependency.

volrath:
hey @tarsius, thanks for the feedback.

> Do you have any estimate on how long that will take? Bundling third-party libraries is always problematic, especially when patched.

tbh, I'm not sure. I'm just a collaborator in that project, not the author or maintainer, so I really can't say when it'll be ready to ship.  I can say that I know there's still some work to do on the Clojure parsing side of things, but on the EDN parsing (what spiral uses it for,) it is basically complete.  Good thing is I'm shipping `parseclj`'s `master` branch as is, most of the time.  I've shipped a few patched versions but @plexus, the author, is very quick to incorporate those patches into the code base.

> Also, you depend on highlight which is a package from the Emacswiki. We are no longer updating packages from there for security reasons and will eventually remove the frozen versions too. You should probably investigate now whether you can drop this dependency.

Good to know! I'll look into it, I'm almost positive I can drop that dependency.

tarsius:
> I'm shipping `parseclj`'s `master` branch as is, most of the time.  I've shipped a few patched versions but @plexus, the author, is very quick to incorporate those patches into the code base.

To me that sounds like you could remove the submodule and instead depend on a separately packaged `parseclj`. Occasionally having to wait a few days for your contributions to be merged seems acceptable.

volrath:
> Occasionally having to wait a few days for your contributions to be merged seems acceptable.

Agreed, although this is not the reason why I chose this path.

`parseclj` is not yet packaged in MELPA or any other Emacs repo, and AFAIK there's no way `package.el` would resolve packages by git refs as versions, but I don't know much about the internals of `package.el` or MELPA so I'm not sure if I'm missing something..?

Anyway, happy to take pointers into the right direction =)

tarsius:
> `parseclj` is not yet packaged in MELPA or any other Emacs repo

You don't have to be the maintainer to submit a recipe.

> and AFAIK there's no way `package.el` would resolve packages by git refs as versions

That's true, but sadly a common and therefore tolerated situation. The consequence would be that `spiral` would not be installable using just Melpa-Stable. It would however be installable using just Melpa (non-stable), or Melpa-Stable and Melpa. In the former case both `spiral` and `parseclj` would be snapshots, in the latter case it would be possible to install at least the stable version of `spiral` by pinning it to the stable archive.

A related issue would be that you could not explicitly specify the required `parseclj` version and would have to specify a snapshot date instead. But that again is a common problem that needs a generic solution. You *could* just depend on a specific snapshot, but then you should also remember to stop doing so once a `parseclj` release becomes available. Once you do that the issue would return, but it might be less problematic then because breaking changes will probably be less common once a `parseclj` change has been made.

(For Magit I actually depend on snapshots of its dependencies, but when making Magit a release I replace the snapshots with the appropriate releases and revert that in the first commit after the Magit release again. Not sure that's worth the effort for less popular packages. I only do it for Magit but not any of my other packages.)

bbatsov:
I agree that it's probably time for a `parseclj` recipe to be added to MELPA. I don't see anything bad that might come out of this.

plexus:
I would very much appreciate parseclj getting added to MELPA, provided that there are people willing to walk it through the process. I don't have much time to invest in it right now. I also found the experience needlessly unpleasant last time, so I don't have much motivation to kick that process off this time.

One consideration might be that we move parseclj to the clojure-emacs github org first. I was holding off hoping to finish it first before drawing more attention to it (see [backlog](https://github.com/lambdaisland/parseclj/issues/8)), but it doesn't look like I'll be working on it in the immediate future, so maybe it's better to just move it. Maybe that way we can actually find some contributors to finish that backlog.

volrath:
> I would very much appreciate parseclj getting added to MELPA, provided that there are people willing to walk it through the process.

Then this settles it.  My main concern with submitting a recipe was to respect @plexus' decision on when to go public.  I can do the work to put `parseclj` in a MELPA-acceptable stage, I've already started with that and I think we're more than half way there.

> One consideration might be that we move parseclj to the clojure-emacs github org first.

I second this.  Maybe @bbatsov can help?

bbatsov:
> I second this. Maybe @bbatsov can help?

Of course. @plexus is already a member of `clojure-emacs`, so all he needs to do is initiate the transfer and that's done.

plexus:
Yup, I'll do it today or in the coming days.

On Jan 8, 2018 07:15, "Bozhidar Batsov" <notifications@github.com> wrote:

I second this. Maybe @bbatsov <https://github.com/bbatsov> can help?

Of course. @plexus <https://github.com/plexus> is already a member of
clojure-emacs, so all he needs to do is initiate the transfer and that's
done.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/melpa/melpa/pull/5190#issuecomment-355890436>, or mute
the thread

plexus:
I initiated the transfer. Thanks @volrath and @bbatsov!

purcell:
> I also found the experience needlessly unpleasant last time, so I don't have much motivation to kick that process off this time.

Sorry to hear that, @plexus.

PR #5192, opened 2017-12-14
Add multi-run

sagarjha:
multi-run 

A minor mode for managing multiple emacs eshell terminals and running commands on them

https://github.com/sagarjha/multi-run

maintainer

** None needed **

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/multi-run
+++ b/recipes/multi-run
@@ -1,0 +1,1 @@
+(multi-run :repo "sagarjha/multi-run" :fetcher github)


purcell:
Thanks. Quick feedback:

- This is not an idiomatically-formatted elisp file. Please start with the boilerplate inserted by `M-x auto-insert` in a fresh `.el` file, and fill it in carefully. Note that `(provide ...)` goes after the rest of the code, not before it.
- This code has definitely not passed `package-lint`, despite the ticked checkbox above.
- `mr-` is not an acceptable prefix for symbol names: use `multi-run-` throughout.
- `window-layout` is not a standard emacs library: do you need to declare a package dependency?
- `mr-term` should be a symbol, not a string. Apart from it being more idiomatic in this case, it will also let you use `eq` instead of `string=` everywhere. Also look at `pcase`.
- Use `let` to introduce local variables. Statements like [this](https://github.com/sagarjha/multi-run/blob/c0f44f7fbc89cb3423ae4710cd3640bcce9bf75a/multi-run.el#L73) bind new global variables.
- `window-batch` isn't even getting used [here](https://github.com/sagarjha/multi-run/blob/c0f44f7fbc89cb3423ae4710cd3640bcce9bf75a/multi-run.el#L90-L91).

sagarjha:
Hi Steve,
Thanks for the comments. I have addressed all of them and pushed to the `multi-run` repository.
1. Generated the headers etc. using `auto-insert`. The `provide` statement is after the code now.
2. I don't know why I checked the `package-lint` box when I definitely did not test against it. I have verified now that the `package-lint` test passes.
3. I am using the `multi-run-` prefix throughout the code.
4. Yes, I need `window-layout` for my package. I added (window-layout "1.4") in the `Package-Requires` header in the file. I was searching for anything to add in the `melpa` recipe, but couldn't find anything related to dependency.
5. `mr-term (multi-run-term)` is now a symbol. I am using `pcase` instead of `string=` and have added the dependency on `emacs 24` as a result.
6. I am using `let` instead of `setq` wherever possible (a lot of places). I let `setq` statements stay where using `let` complicated the code logic.
7. `window-batch` is getting used. It is an optional argument to the function `multi-run-configure-terminals`. In the function, it is set to 5 if the user does not provide it. Then, it is passed to the helper function `multi-run-make-internal-recipe` as an argument and that function makes extensive use of it.

purcell:
Thanks. Sent you a couple of small fixes in https://github.com/sagarjha/multi-run/pull/1.

There are still a number of issues with the code which you should address:

- You're declaring functions inside other functions, which still just declares them globally. So don't do that, and make sure you've got the package prefix for _all_ functions.
- You're still creating and using global variables with no package prefix, e.g. `sym-vec`.
- Prefer `(unless (...)` to `(when (not (...)`
- Don't use `boundp` for your own variables. It's not good practice to determine program logic based on whether or not magic variables are bound. Just declare them directly at the top level.

sagarjha:
Hi Steve,
Sorry for the delay. I was on vacation.
I have accepted your suggestions starting with the pull request you created. I have also taken care of the issues you point out in the above comment:
1. I am not sure if there's any issue with declaring functions inside other functions. I might be using some variables of the outer function inside the inner function without passing them to the inner function. Let me know if you still think that it is better to define functions globally.
2. I have declared variables like `sym-vec` and others globally with the prefix `multi-run-` added. For some other global variables, I have made them locally bound by using let.
3. Using `(unless` instead of `(when (not ()` everywhere in the code.
4. I have removed all occurrences of `boundp` and declared the corresponding variables at the top level.

Thanks very much for your comments. I am new to contributing code to emacs, so your suggestions are very useful. Let me know if there are any more edits to make. Otherwise, I would like you to add the package to the repository.

purcell:
> I am not sure if there's any issue with declaring functions inside other functions. I might be using some variables of the outer function inside the inner function without passing them to the inner function. Let me know if you still think that it is better to define functions globally.

When you use `defun`, you *are* defining (or, then, redefining) functions globally anyway. So this is confusing and unreliable. What you want instead is to define the helper functions at the top level with explicit argument lists, rather than relying on closed-over variables from the defining scope). Occasionally it can also be helpful to locally bind helper functions, using `cl-letf` etc.

sagarjha:
Thanks! I did move the internal helper functions to the top-level. Doing this simplified some of the logic in the `multi-run-configure-terminals` as I was able to get rid of setq statements and use let* instead. As a result, I don't need to declare some of the variables at the global level.

sagarjha:
Thanks a lot! I am excited!

PR #5193, opened 2017-12-14
Update fetcher for centered-cursor-mode

hinrik:
The author recently uploaded the mode to Github, and has added a bugfix that isn't in the EmacsWiki version.


tarsius:
https://github.com/andre-r/centered-cursor-mode

tarsius:
I have migrated both of @andre-r's packages.

andre-r:
Thanks, do I have to consider something, like `Package-Version`?

tarsius:
Melpa does not require it, but other `package.el` related things do. It is up to you I guess.

I recommend you run `package-lint`. It will tell you what I have just told you, but possibly also some other things.

PR #5196, opened 2017-12-15
Added a recipe for log-mode

ellisvelo:

Log-mode is a minor mode used for finding and navigating errors within a
buffer or a log file.

https://bitbucket.org/ellisvelo/log-mode

I am the maintainer. 

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/log-mode
+++ b/recipes/log-mode
@@ -1,1 +1,0 @@
-(log-mode :fetcher bitbucket :repo "ellisvelo/log-mode")
--- a/recipes/log-mode
+++ b/recipes/log-mode
@@ -1,0 +1,1 @@
+(log-mode :fetcher bitbucket :repo "ellisvelo/log-mode")
--- a/recipes/lognav-mode
+++ b/recipes/lognav-mode
@@ -1,0 +1,1 @@
+(lognav-mode :fetcher bitbucket :repo "ellisvelo/lognav-mode")


purcell:
Thanks! This looks great, but it's an extremely general name for such a new and specific package, so I'd suggest a renaming to `lognav-mode` or similar. Other feedback:

- The package is presumably useful for navigating messages that are not only exceptions, and are sometimes even success messages, so the var name `log-exception-regexp` seems too narrow. Probably just `log-regexp` (=> `lognav-regexp`) would be better.
- That variable is a primary part of the package's configuration: make it a defcustom.
- Re. `log-overlay?` - `log-overlay-p` would be more idiomatic in elisp.
- Unconditionally enabling `auto-revert-tail-mode` is very specific to your workflow. I suggest you omit it, and then users can still add it to their `lognav-mode-hook` if they want.
- Consider `cl-remove-if` or similar [here](https://bitbucket.org/ellisvelo/log-mode/src/41c9aee709f15f739d0edef2333dbe5c5cf11e9b/log-mode.el?at=default&fileviewer=file-view-default#log-mode.el-125).
- The `auto-mode-alist` entry might also be quite presumptuous, but in any case its regexp should perhaps be written `"\\.log\\(\\.[0-9]+\\)?\\'"` instead.

Hope that helps!

ellisvelo:
Hello Steve,

I've renamed the mode to lognav-mode as you had suggested. I've also implemented your recommendations including the removal of auto-revert-tail-mode. It wasn't so much about my workflow as I was thinking of simplifying the configuration for others that are not as familiar with emacs. Thanks for the help!

purcell:
Great, merged! Thanks for your patience.

PR #5198, opened 2017-12-18
Add cyanide

mciocchi:

Features:

    nearly instant project aware search via helm-ag

    hooks to run arbitrary elisp at project load time

    easy project specific configuration via .cy.el dotfiles

    a dead-simple means for users to define project lifecycle tasks (compile, test, run, etc.)

    a way to optionally "pop" into arbitrary buffer and window configurations, especially at project load time

    an extensible API for advanced users to work with projects and artifacts, implemented in EIEIO CLOS

https://github.com/mciocchi/cyanide

Sole maintainer and contributor.

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/cyanide
+++ b/recipes/cyanide
@@ -1,0 +1,4 @@
+(cyanide
+ :repo "mciocchi/cyanide"
+ :fetcher github
+ :files ("*.el" "lib" "Cask" "README.md" "COPYING"))


purcell:
Thanks! I love that you're tackling a big project, and thanks for submitting it to us.

The checkbox above confirms that you've run `package-lint`, but I can immediately see there's some feedback it would have given you that has not been addressed, namely the use of the reserved `C-c c` sequence as a key binding prefix. Please can you try again?

More feedback:
- Don't explicitly require files by path. It's not idiomatic to put sub-libraries in sub-directories, as [here](https://github.com/mciocchi/cyanide/blob/aa32c35870be07f94a0767ae454f43c50088e025/cyanide.el#L62)
- You've got a trailing space in your [mode lighter](https://github.com/mciocchi/cyanide/blob/aa32c35870be07f94a0767ae454f43c50088e025/cyanide.el#L80) which should be removed.
- Don't [enable the mode just because the feature has been loaded](https://github.com/mciocchi/cyanide/blob/aa32c35870be07f94a0767ae454f43c50088e025/cyanide.el#L88). Instead, add `;;;###autoload` comments for the modes.
- Same point applies to [cyanide-main](https://github.com/mciocchi/cyanide/blob/master/lib/cyanide-main.el#L16): trying to defer loading of internal and external libraries is not necessary if you set up your source files idiomatically. Those `require` statements should be at the top level, and without explicit paths.
- You're going to have a bad time if someone tries to call `cyanide-sha1sum` with an argument like `"';sudo rm -rf;echo '"`. Use `shell-quote-argument` religiously. Why would you need to provide crypto functions in this package? Seems totally orthogonal to the purpose of the package, and likely a recipe for trouble.
- Update: I eventually found more crypto code, also using `%s` and shelling out to `gpg`. Emacs has built-in support for encrypting/decrypting files with GPG, so you're reinventing the wheel [here](https://github.com/mciocchi/cyanide/blob/aa32c35870be07f94a0767ae454f43c50088e025/lib/models/cyanide-project.el#L129).
- Why would someone write `(cyanide-return-if-true 'string= a b 'foo)` instead of simply `(and (string= a b) 'foo)`?
- [These functions](https://github.com/mciocchi/cyanide/blob/aa32c35870be07f94a0767ae454f43c50088e025/lib/controller/cyanide-misc-utils.el#L46) are declared in the global namespace, and almost certainly have equivalents in a standard library. `cl-lib` provides `cl-filter`, for example.
- You use `helm`, and there's no package dependency on it.
- What's with the [Python docstrings](https://github.com/mciocchi/cyanide/blob/aa32c35870be07f94a0767ae454f43c50088e025/lib/controller/cyanide-helm-wrapper.el#L22)?

I'm going to stop reviewing here, sorry. The structure really isn't idiomatic for elisp, though, and the code is extremely hard to follow. I can't accept the package in its current form because it needs a lot of work IMO.

mciocchi:
Hi Steve:

Thank you so much for your feedback. I will address each of these points one by one, and then resubmit.

Thanks,
MC

purcell:
Cool. I feel like I was probably a bit grouchy yesterday, so sorry if the feedback was a bit harsh: I don't feel like I provided much useful guidance. My main suggestion, then, would be to look at some larger multi-file elisp projects (e.g. `magit`) to see how such codebases are organised. There's not a huge amount of code in `cyanide`, so a good default is to start with everything in one file (!), organised into sections using `^L`, and then when particular sections become very large, _maybe_ split them out into files. Starting up-front with the artificial M/V/C split is going to lead you away from an easy-to-understand structure in this case, even if it's helpful in some other settings. Feel free to ping me here with questions if you have any.

mciocchi:
No Steve, it's totally fine. I'm used to code reviewing and people coming right out and saying what's wrong with my code.

Your feedback has been invaluable. A lot of the idiosyncrasies that you noticed in the code can be chalked up to the fact that I have had a really hard time finding actual living humans who know elisp and can give me feedback. Over the course of the three years during which the CyanIDE project has been active, I've had to teach myself elisp by reading snippets of other people's code. Some of what I copied was new and good, and some of it was old and funky. Hence the python-like docstrings.

The java-style nesting can be chalked up to the fact that I was coding a lot of java back in 2015 when I originally laid out the project directory structure! I have since learned that the elisp community considers this an antipattern, and I'll fix it.

In fact, in a way, CyanIDE was born out of the need to quickly full-project-search the highly nested legacy code which I was working on at the time. It had thousands of files! Still, it's not the most helpful pattern, I know.

Regarding enabling the mode at load time- good catch. Yes, once I learned what I was doing, this began to bother me, and I had always intended to fix it. I can do that before I resubmit. The other complicated thing which you can see in CyanIDE is how my code changed as I learned elisp, over time. Some of the high level fragments like that were written really early in the project, before I knew what I was doing  ^.^

I'll have to look into shell-quote-argument. I didn't know that existed.  ;p The cyanide-crypto features are used in automatic project initialize/import/export functions which are 80% complete and out on a branch right now. I wanted the ability to quickly export projects to external media, but I wanted them to be protected in case the external media got lost.

Lastly, regarding package-lint: it did give me some output, but I wasn't really sure what some of the messages meant, and how to make them go away! I will come back to it and perhaps ask some questions before I resubmit.

As you can probably tell, CyanIDE includes a bunch of features which I was looking around for when I was a beginner, but couldn't find. I had to write them myself. As a beginner, there were so many basic project management tasks which I struggled with. I almost quit emacs because other (lesser) editors provided those features out of the box.

I've been driven to keep working on this project for all these years because emacs is such an awesome editor/language/OS/whatever you want to call it. I really want emacs and lisp to stay alive so that the next generation of programmers can appreciate them. I really think that if the community wants emacs to stay alive, we will need to provide tools that make it more approachable for beginners. I believe that it is vital that we keep attracting new users so that our community doesn't keep shrinking until we go the way of the dinosaurs.

I will end my rant and manifesto regarding the purpose of CyanIDE here  ;p  I hope you don't mind, but I will probably have to ask you some questions about a few of the issues you listed. I will work through them and get back to you. Likewise, I have become something of an expert on EIEIO/CLOS and a few other features of elisp since I started this project, so if you ever need code review or anything, I'd be happy to return the favor.

MC

PR #5201, opened 2017-12-19
Add tronesque

aurelienbottazini:

Add tronesque which is an emacs color-theme

https://github.com/aurelienbottazini/tronesque

Maintainer

 **None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/tronesque
+++ b/recipes/tronesque
@@ -1,2 +1,1 @@
-
 (tronesque :fetcher github :repo "aurelienbottazini/tronesque")
--- a/recipes/tronesque
+++ b/recipes/tronesque
@@ -1,0 +1,2 @@
+
+(tronesque :fetcher github :repo "aurelienbottazini/tronesque")


tarsius:
You have to set `:files` according to the documentation.

> [x] I've built and installed the package using the instructions in CONTRIBUTING.md

Really?

purcell:
The theme code is not correct in any case, since it [tries to decide at load-time](https://github.com/aurelienbottazini/tronesque/blob/master/themes/tronesque-theme.el#L42) what colours to use based on whether the current frame is a GUI frame or a terminal frame. This means, for example, that it will not work if loaded by an emacs daemon and used in a GUI frame spawned by that process. The correct way to define the theme is with terminal classes.

Additionally, there's an [unprefixed helper function](https://github.com/aurelienbottazini/tronesque/blob/5387b7346d53e04014ed5a7a6bf35245c0ad597e/themes/tronesque-theme.el#L367) which  `package-lint` would have complained about had it _actually_ been run.

Closing this for now, then, sorry.

PR #5202, opened 2017-12-19
Add recipe for dimmer

gonewest818:

This module provides a subtle visual indication which window is
currently active by dimming the faces on the others.  It does this
nondestructively, and computes the dimmed faces dynamically such
that your overall color scheme is shown in a muted form without
requiring you to define the "dim" versions of every face.

The percentage of dimming is user configurable.  In addition, for
users of "light" themes there is a dimmer-invert flag that adjusts
the faces brighter (toward white, rather than toward black).

Unlike other modules with similar goals, this implementation
does *not* change the color of the background in any way. It only
adjusts foregrounds.

https://github.com/gonewest818/dimmer.el

I am the maintainer.

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/dimmer
+++ b/recipes/dimmer
@@ -1,0 +1,1 @@
+(dimmer :fetcher github :repo "gonewest818/dimmer.el")


manuel-uberti:
This looks great!

gonewest818:
Thank you! Hope you give it a try!

manuel-uberti:
I did, actually. It only seems to dim other windows when point is in the minibuffer. Can anyone else confirm?

gonewest818:
Can you open an issue for that? ==> https://github.com/gonewest818/dimmer.el

manuel-uberti:
Done > https://github.com/gonewest818/dimmer.el/issues/3

gonewest818:
Thank you.

purcell:
Thanks! Quick feedback:
- Don't [shim built-in functions](https://github.com/gonewest818/dimmer.el/blob/master/dimmer.el#L82-L108) using their original names. The [elisp conventions](https://www.gnu.org/s/emacs/manual/html_node/elisp/Coding-Conventions.html) mention this. Aren't those functions available in Emacs 25+ anyway? If not, consider bumping the dependency instead of shimming.
- You should add a `defgroup` to define the `dimmer` group and place it in the global custom group hierarchy.
- `dimmer-invert` should be unnecessary: you can infer it by looking at the `default` face and seeing whether the background is brighter than the foreground.

gonewest818:
Roger that. Thanks for the feedback.

gonewest818:
Ok, done. All feedback has been addressed.

purcell:
Awesome, thanks.

BTW, while I'm not personally a `dash`, fan, it provides an `-if-let`/`-if-let*`. The whole history of the built-in counterparts is a bit silly.

PR #5203, opened 2017-12-20
Added recipe for go-imports.

yasushi-saito:
https://github.com/yasushi-saito/go-imports

go-imports inserts a package import statement, given the package name.

https://github.com/yasushi-saito/go-imports

Maintainer

**None needed** 

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/bazel-mode
+++ b/recipes/bazel-mode
@@ -1,0 +1,1 @@
+(bazel-mode :repo "codesuki/bazel-mode" :fetcher github)
--- a/recipes/buckwalter
+++ b/recipes/buckwalter
@@ -1,0 +1,1 @@
+(buckwalter :fetcher github :repo "joehakimrahme/buckwalter.el")
--- a/recipes/company-eshell-autosuggest
+++ b/recipes/company-eshell-autosuggest
@@ -1,1 +1,0 @@
-(company-eshell-autosuggest :fetcher github :repo "dieggsy/company-eshell-autosuggest")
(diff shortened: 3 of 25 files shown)


purcell:
Thanks! Quick feedback:
- This code is not building shell commands safely. For example, what if `root` contains a space, quotes or worse [here](https://github.com/yasushi-saito/go-imports/blob/aae96b81b429c9ca16c926a03508b01f41c754e0/go-imports.el#L127)? Both `root` and `perl-script` should be escaped with `shell-quote-argument`, not simply with `'`.
- Consider providing `;;;###autoload` comments for the main interactive functions so that the user doesn't need to `require` the library.

yasushi-saito:
Thanks. I changed the call-process part so that it invoke perl directly, w/o find&grep and such. This should avoid the problem you mentioned. Also added autoload magic lines.

purcell:
Great! Looks good.
- Don't use `ido-completing-read` directly: just use `completing-read`. Then the behaviour will be overridden automatically by the user's preferred completion mechanism (helm, ivy, ido etc.)
- Rather than putting the perl script in a string and then writing it to a temporary file, just add it to the repository, then you can include it in the package recipe (with `:files (:defaults "go-imports.pl")`. To find its location in the code, you can capture the results of `load-file-name` in a `defconst`.

yasushi-saito:
Thanks for the review. I applied all your comments.

purcell:
Thanks. Added in 4118ebf0, since this PR had become a little cluttered. :-)

PR #5205, opened 2017-12-21
Add recipe for bazel-mode

codesuki:

Supports basic syntax highlighting and formatting for Bazel files.

https://github.com/codesuki/bazel-mode

Maintainer

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/bazel-mode
+++ b/recipes/bazel-mode
@@ -1,0 +1,1 @@
+(bazel-mode :repo "codesuki/bazel-mode" :fetcher github)


codesuki:
`C-c C-c` in the recipe shows me the following error.
```
Install new package? (y or n) y
package-version-join: Wrong type argument: arrayp, nil
```

```
(bazel-mode .
            [(20171221 1514)
             ((emacs
               (24 3)))
             "A major mode for editing Bazel files" single
             ((:commit . "bf41f73c0df7a414c55fc9ac07573300f3161258")
              (:keywords "languages" "bazel")
              (:authors
               ("Neri Marschik"))
              (:maintainer "Neri Marschik")
              (:url . "https://github.com/codesuki/bazel-mode"))])
```

purcell:
Thanks. The error you posted looks odd, sorry, but the build was happy. Your pasted output includes a dependency on Emacs 24.3, which is required by your code, but that package dependency is missing in the latest `master` version of the code, so I'm not sure what's going on there. Everything looks fine apart from that.

codesuki:
I removed that since it was a leftover from code I removed before. So I tried if it works if I simplify the version scheme to be similar to my other Emacs packages. 

Anyhow, I tried to build other recipes locally and they all gave me the same message. So I think it's something wrong with my emacs. 

If it works on the ci it's probably ok.

purcell:
> I removed that since it was a leftover from code I removed before. So I tried if it works if I simplify the version scheme to be similar to my other Emacs packages.
> 
> Anyhow, I tried to build other recipes locally and they all gave me the same message. So I think it's something wrong with my emacs.
> 
> If it works on the ci it's probably ok.

Sorry, perhaps I wasn't clear. Your code uses `setq-local`, so you _must_ declare a dependency on `(emacs "24.3")` in your `Package-Requires` header. The CI doesn't check for that.

codesuki:
Oh sorry. Didn't get that. I'll change it back.

codesuki:
I added back the dependency.

purcell:
Great, thanks - merged!

codesuki:
Thanks!

odanoburu:
@codesuki I'm having this same problem you (are? were?) having:
```
package-version-join: Wrong type argument: arrayp, nil
```
did you solve it eventually, or did you guys just merge because CI showed everything was fine?

thanks for the info!

codesuki:
We just merged and it seems to have worked out nicely. I am pretty sure it has something to do with my local emacs installation that I cannot get that to work.

PR #5208, opened 2017-12-23
Create rsw-elisp

rswgnu:

Interactively evaluate regions, preceding variable definitions and quoted sexpressions

https://github.com/rswgnu/rsw-elisp

Author

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/rsw-elisp
+++ b/recipes/rsw-elisp
@@ -1,0 +1,3 @@
+(rsw-elisp
+  :repo "rswgnu/rsw-elisp"
+  :fetcher github)


purcell:
Thanks! Quick feedback:

- With `rsw-elisp-enable`, `rsw-elisp-disable`, `rsw-elisp-toggle` and `rsw-elisp-p`, plus all the rebinding of commands, you are simply reinventing minor modes in an ad-hoc way. Simply provide a minor mode which people can add to their `emacs-lisp-mode-hook`, and place your overrides in the keymap for that minor mode. You can use a global/local minor mode pair to handle both the `M-:` override and the elisp-mode-specific behaviour.
- Changing unrelated bindings in `minibuffer-local-map` is inappropriate: don't do that. Put it in your personal config if you must.

rswgnu:
On Sat, Dec 23, 2017 at 9:08 PM, Steve Purcell <notifications@github.com>
wrote:

> Thanks! Quick feedback:
>

​Thanks for the feedback.  Even better would be if you try it and send your
experience using it as feedback so we get a sense of what you find you like
and dislike in usage rather than just the implementation constructs.​

>
>    - With rsw-elisp-enable, rsw-elisp-disable, rsw-elisp-toggle and
>    rsw-elisp-p, plus all the rebinding of commands, you are simply
>    reinventing minor modes in an ad-hoc way. Simply provide a minor mode which
>    people can add to their emacs-lisp-mode-hook, and place your overrides
>    in the keymap for that minor mode.
>
> Let's flesh this out some more.  I want the improvements to M-: and C-x
C-e available in every buffer via global key changes, so they can be used
at any time.  Minor modes apply per buffer, local changes.  What advantage
(other than a potential visual indicator in the mode-line) would this
bring?  It certainly seems much like a minor mode at first glance but maybe
that is not the case.

>
>    - Yo
>    ​​
>    u can use a global/local minor mode pair to handle both the M-:
>    override and the elisp-mode-specific behaviour.
>
> If you have to enable a global-minor-mode to get the global key overrides
and there is no context where the local minor mode is really desirable
other than as a sub-part of the global-minor-mode, what will the user
gain?  I am not arguing for or against this, just posing the question and
looking for convincing reasons.

>
>    - ​​
>    Ch
>    ​​
>    anging unrelated bindings in minibuffer-local-map is inappropriate:
>    don't do that. Put it in your personal config if you must.
>    ​​
>
> ​There is a flag variable to disable that, if not desired.  I figured it
was better for people to try it by default.  Let's see what the general
reaction is first.
I agree it should be limited to use with the commands affected by this
file, though, rather than the minibuffer in general, so that could be a
future feature change.

>

purcell:
> Thanks for the feedback.  Even better would be if you try it and send your
> experience using it as feedback so we get a sense of what you find you like
> and dislike in usage rather than just the implementation constructs.​

I'm not in a position to try out all the submitted packages, sorry, but the implementation constructs are still relevant.

>     - With rsw-elisp-enable, rsw-elisp-disable, rsw-elisp-toggle and
>     rsw-elisp-p, plus all the rebinding of commands, you are simply
>     reinventing minor modes in an ad-hoc way. Simply provide a minor mode which
>     people can add to their emacs-lisp-mode-hook, and place your overrides
>     in the keymap for that minor mode.

> Let's flesh this out some more.  I want the improvements to M-: and C-x
> C-e available in every buffer via global key changes, so they can be used
> at any time.  Minor modes apply per buffer, local changes.  What advantage
> (other than a potential visual indicator in the mode-line) would this
> bring?  It certainly seems much like a minor mode at first glance but maybe
> that is not the case.

Minor modes have keymaps which can apply globally or locally, and can be toggled on/off, so that's exactly what that code does. (The visual indicator - "lighter" - is relatively irrelevant here: it's optional anyway.)

>     - You can use a global/local minor mode pair to handle both the M-:
>     override and the elisp-mode-specific behaviour.

> If you have to enable a global-minor-mode to get the global key overrides
> and there is no context where the local minor mode is really desirable
> other than as a sub-part of the global-minor-mode, what will the user
> gain?  I am not arguing for or against this, just posing the question and
> looking for convincing reasons.

A "globalised minor mode" can have its own keymap, as well as decide whether to enable its local minor mode counterpart in specific buffers. I think that's exactly what you want. So you'd have a local minor mode which would provide `C-x C-e` bindings etc., and a globalized version of it defined with `define-globalized-minor-mode` which would define `M-:` etc. in its own map, and also enable the local minor mode in elisp buffers. This is a great outcome for users because they can also simply use the local minor mode if that's the part they're interested in - this would be the case for me, for example. And in providing two keymaps, you're providing standard ways for users to modify keybindings. If someone wants to skip your `M-:` override, they can simply customize that binding to `nil` in the map for your global mode: no special flag variables or options required.


>     - ​​
>     Changing unrelated bindings in minibuffer-local-map is inappropriate:
>     don't do that. Put it in your personal config if you must.
> ​There is a flag variable to disable that, if not desired.  I figured it
> was better for people to try it by default.  Let's see what the general
> reaction is first.
> I agree it should be limited to use with the commands affected by this
> file, though, rather than the minibuffer in general, so that could be a
> future feature change.


That's not a "wait and see" or future thing IMO -- you're changing a global behaviour here for all minibuffer commands, which isn't correct.

rswgnu:
On Sat, Dec 23, 2017 at 10:14 PM, Steve Purcell <notifications@github.com>
wrote:

> ​​
> Thanks for the feedback. Even better would be if you try it and send your
> ​​
> experience using it as feedback so we get a sense of what you find you like
> ​​
> and dislike in usage rather than just the implementation constructs.​
>
> I'm not
> ​​
> in a position to try out all the submitted packages, sorry, but the
> implementation constructs are still relevant.
>
> ​Yes, I realized after responding that you were sending comments as a
MELPA maintainer, not a user, per se.
​


> Minor modes have keyma
> ​​
> ps which can apply globally or locally, and can be toggled on/off, so
> that's exactly what that code does. (The visual indicator - "lighter" - is
> relatively irrelevant here: it's optional anyway.)
>
> ​I see.  The main use you are advocating is easily customizable keys.
​


> - You can use a
> ​​
>  global/local minor mode pair to handle both the M-
> ​​
> :
> override and th
> ​​
> e elisp-mode-specific behaviour.
> ​​
> ​​
>
> A "globalised minor mode" can have its own keymap, as well as decide
> whether to enable its local minor mode counterpart in specific buffers.
> ​  ​
> And in providing two keymaps, you're providing standard ways for users to
> modify keybindings.
>
> ​Yes, it might take a while given the holidays but I'll change it.  I
evolved HyControl, the window and frame manager, into a minor mode and
worked out well.

> Changing unrelated bindings in minibuffer-local-map is inappropriate:
> ​​
> ​​
> don't do that. Put it in your personal config if you must.
> ​​
>
> ​​
> ​There is a flag variable to disable that, if not desired. I figured it
> ​​
>
> ​​
> was better for people to try it by default. Let's see what the gen
> ​​
> eral
> ​​
> reaction is first.
> ​​
> I agree it should be limited to use with the commands affected by this
> ​​
> file, though, rather than the minibuffer in general, so that could be a
> ​​
> future feature change.
>
> That's
> ​​
> not a "wait and see" or future thing IMO -- you're changing a global
> behaviour here for all minibuffer commands, which isn't correct.
>
> ​I meant see how people feel about the key bindings when interactively
evaluating Elisp, not the implementation limitation that exists right now
that makes them apply to all minibuffer commands.

Thanks for the clear feedback.  It is helpful.  Happy holidays.

Bob
​

rswgnu:
On Sat, Dec 23, 2017 at 10:55 PM, Robert Weiner <rsw@gnu.org> wrote:

> On Sat, Dec 23, 2017 at 10:14 PM, Steve Purcell <notifications@github.com>
> wrote:
>
>> ​​
>> I agree it should be limited to use with the commands affected by this
>> ​​
>> file, though, rather than the minibuffer in general, so that could be a
>> ​​
>> future feature change.
>>
>> ​I meant see how people feel about the key bindings when interactively
> evaluating Elisp, not the implementation limitation that exists right now
> that makes them apply to all minibuffer commands.
>

​This has been resolved with version 1.0.4; the helper commands differ from
standard behavior only when called from rsw-elisp-eval-expression and only
in very limited contexts where they can be helpful.  Any user that does not
want them, can disable them with one setting.

Bob

DamienCassou:
Any update on that one?

rswgnu:
On Thu, Mar 1, 2018 at 9:06 AM, Damien Cassou <notifications@github.com>
wrote:

> Any update on that one?
>

​I did as much work on this as I am able to at this time and stopped work
on it at the end of last year.  It is easy to use and as non-intrusive as I
felt appropriate, though I did not get to making it a regular minor mode as
suggested.  It is pretty well documented and does what it intends to.

Regards,

Bob
​

tarsius:
@rswgnu Do you not want to use a minor-mode or did you just not get to it?
@purcell Is it okay to merge this despite the lack of a minor-mode?

rswgnu:
> On Mar 21, 2018, at 7:54 PM, Jonas Bernoulli <notifications@github.com> wrote:
> 
> @rswgnu Do you not want to use a minor-mode or did you just not get to it?
> 
Have no time to work on it much further at this point.

Bob

purcell:
Okay, closing for now then, sorry.

PR #5212, opened 2017-12-27
Add recipe for interactive-align

mkcms:
Interactive version of align-regexp command.
This package provides `ia-interactive-align` command, which will align the selected
region automatically as the regexp characters are typed in the minibuffer.

https://github.com/mkcms/interactive-align

I am the maintainer.

Package-lint reports many "symbol doesn't start with package prefix" error - I use "ia-" prefix, and it wants "interactive-align-".

### Checklist

Please confirm with `x`:

- [] `M-x checkdoc` is happy with my docstrings

### DIFF
--- a/recipes/ialign
+++ b/recipes/ialign
@@ -1,0 +1,2 @@
+(ialign :fetcher github
+	:repo "mkcms/interactive-align")


purcell:
> Package-lint reports many "symbol doesn't start with package prefix" error - I use "ia-" prefix, and it wants "interactive-align-".

Yes, you should use `interactive-align-` please. If you want a shorter prefix, consider renaming the package to something like `ialign`, so that the prefix is `ialign-`.

Looks good otherwise.

mkcms:
I renamed the file and changed prefix to `ialign` and now package-lint shows no errors.

purcell:
Nice, thanks!

I suggest you simply call your main interactive `ialign`. You can use `define-obsolete-function-alias` to alias the old name.

PR #5213, opened 2017-12-29
Add recipe for TeXfrag

TobiasZawada:
TeXfrag can be used for rendering equations in sx-question-mode buffers.
sx.el is on melpa therefore also TeXfrag should be on melpa.

Make AUCTeX preview available for non-LaTeX buffers.

https://github.com/TobiasZawada/TeXfrag

maintainer

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/TeXfrag
+++ b/recipes/TeXfrag
@@ -1,0 +1,1 @@
+(TeXfrag :repo "TobiasZawada/TeXfrag" :fetcher github)


purcell:
Thanks, but that repository just contains a renamed version of `flycheck.el`, and is therefore not close to being ready for distribution via MELPA. It looks like you made a mistake while trying to fix up some `package-lint` warnings.

If you'd like to try again, just comment here once things are fixed.

I'll note in advance, though, that capitalisation in package names and elisp identifiers is not idiomatic, so you should rename your package (and its symbols) to use `texfrag` instead of `TeXfrag`.

TobiasZawada:
@purcell 
Thanks.
I've renamed TeXfrag to [texfrag](https://github.com/TobiasZawada/texfrag/blob/master/texfrag.el).
I've run checkdoc, package-lint, changed the recipe in [melpa](https://github.com/TobiasZawada/melpa/tree/patch-1) and rebuilt for testing. Everything looks fine.
Could we try again?

TobiasZawada:
@purcell 

I read the following comment further above in this discussion:

> Add more commits by pushing to the patch-1 branch on TobiasZawada/melpa.

I did exactly that. The result is the pull request [tree/patch-1](https://github.com/melpa/melpa/pull/5214). Maybe, I did something wrong again:-(.

It's the first time I commit something to melpa. My other package [html-check-frag.el](https://github.com/melpa/melpa/blob/269acf093c3aa508479bf532a4e07c9c6d568c72/recipes/html-check-frag) was committed by Wilfred two years ago.

tarsius:
Where can we get the soft dependency `html+` from. If appropriate, please add a recipe for that.

TobiasZawada:
@tarsius Thanks for the comment. I will remove the dependency on html+. The package texfrag emerged from a private package because it is helpful for sx and html+ is still for my private usage only. I just have to organize things a bit different at my place. That is the reason why it is not done yet. Currently it should do no harm at all if html+ is not available (I took care about that and tested it).

Is it sufficient for updating the package if I push to master at [texfrag](https://github.com/TobiasZawada/texfrag)?

tarsius:
Yes. Thanks!

PR #5215, opened 2017-12-29
Add recipe for rocktl

stevenremot:

Rocktl is a simple task runner. It allows you to define shell commands associated to directories, and to run them quickly.

https://github.com/stevenremot/emacs-rocktl

I am the maintainer of the package

I ran package-lint against the main entry point of my package (rockt.el). I don't know if that's the proper way to check multi-file packages.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/rocktl
+++ b/recipes/rocktl
@@ -1,0 +1,1 @@
+(rocktl :repo "stevenremot/emacs-rocktl" :fetcher github)


alphapapa:
Hi,

This looks interesting.  How does it compare with existing packages that do this kind of thing?  It seems like lots of people are submitting their own do-things-in-directories packages lately.  It might be a good idea to collaborate instead of having lots of similar packages in MELPA.

stevenremot:
Hi,

That's a good idea for sure! When I searched for this kind of package I couldn't find anything. Do you have some examples in head?

alphapapa:
make-it-so and firestarter come to mind as similar (not exactly the same).  There was also one more similar to yours submitted recently, but I forget the name.

stevenremot:
I had a look at make-it-so and firestarter. make-it-so really focuses on transforming files (and is pretty neat by the way), so it's hard to tell if it can merge properly with rocktl or not.

firestarter is pretty different in my opinion, even if it could plug to rocktl easily.

I searched a bit the one you couldn't remember, is it https://github.com/mciocchi/cyanide ? Because this one looks a lot like rocktl for sure! I will try it a bit to see if I can contribute to it instead of pushing a new package.

Thanks for the curation, I'm used to the melpa review improving my package's quality, but I'm glad to see it going to the level of the whole ecosystem 👍

alphapapa:
> I searched a bit the one you couldn't remember, is it https://github.com/mciocchi/cyanide ? Because this one looks a lot like rocktl for sure! I will try it a bit to see if I can contribute to it instead of pushing a new package.

No, that's not the one I was thinking of.  I hadn't even seen that package before.  :)

> Thanks for the curation, I'm used to the melpa review improving my package's quality, but I'm glad to see it going to the level of the whole ecosystem

Steve is the one who deserves most of the credit here, along with the other MELPA maintainers.  I'm just an observer who tries to help a little now and then, hopefully without overstepping my bounds.  You should ignore me and listen to him.  ;)

purcell:
Great conversation here. Sorry for not having been available earlier.

I can see what you're going for here, and I agree with @alphapapa that there's perhaps room for a package which is more general than the packages mentioned above, but I'd say the approach you've chosen isn't the most natural for Emacs.

Specifically, by adding functions for loading, creating, saving and managing command definitions, you're basically reinventing the emacs custom variable system, but in a less proven, flexible and understood way -- you could instead have a custom variable containing a list of commands, and then users could either customise it, load its settings from a file, initialise it in their Emacs start-up files, or even override it on a per-project basis using Directory Local Variables. I feel like if you embraced this approach, very little code would be left.

For these reasons, I'm not keen to merge the package in its current form, sorry.

alphapapa:
> Should I close the PR as it will take time to write all of this?

Yes, please reopen when your new package is ready.  :)

stevenremot:
Done, see you later!

PR #5217, opened 2017-12-31
How to remove the dev repo

ShalokShalom:
Here, I add a short description how to remove the 'bleeding edge' repository.

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,6 +85,10 @@ you prefer to only receive updates for tagged releases, use
   then by default you will get the *development* versions of packages
   and not the stable ones.
 
+* In order to remove the development repo, remove this line:
+  ```  (add-to-list 'package-archives (cons "melpa" url) t))
+  ```
+
 * You will probably want to remove all packages and then reinstall
   them. Any packages you already have installed from MELPA will never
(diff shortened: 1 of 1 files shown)


purcell:
Thanks, but I think it's already _fairly_ clear from the docs, and we don't use the term `development repo` to describe the default MELPA archive anyway, so I think this change would lead to additional confusion.

ShalokShalom:
There is absolutely nothing in the README, while the homepage describes it. 
Plus, that one speaks about "bleeding-edge" packages, that in confuses. 
And third, you indeed mention 

>  If you leave the original MELPA server in your package-archives then by default you will get the development versions of packages and not the stable ones, because the development versions are higher.

ShalokShalom:
Ah, you commited a new version of the README. That one looks fine. 😄

purcell:
> Ah, you commited a new version of the README. That one looks fine.

Yep, forgot to mention that in a comment here - thanks for the prod.

ShalokShalom:
prod.? :)

milkypostman:
a poke. like, "thanks for reminding me"

ShalokShalom:
Ah, I see. :)

PR #5216, opened 2017-12-30
specify "gh-pages" branch since "master" does not exist

emacs18:
Default branch is "master", but https://github.com/google/styleguide.git
git repo does not have such branch.

PR #5219, opened 2018-01-01
use local branch if remote repoo is not accessible

emacs18:
Without internet connection, "git remote show origin" cannot be used to determine the branch name.
In such cases use the default branch name of the local clone rather than using "master" which may 
not be correct branch name for some packages such as google-c-style.

### DIFF
--- a/package-build/package-build.el
+++ b/package-build/package-build.el
@@ -311,8 +311,11 @@ Returns the package version as a string."
 
 (defun package-build--git-repo (dir)
   "Get the current git repo for DIR."
-  (package-build--run-process-match
-   "Fetch URL: \\(.*\\)" dir "git" "remote" "show" "-n" "origin"))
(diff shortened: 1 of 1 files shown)


emacs18:
See also https://github.com/melpa/melpa/pull/5216 which has further details on how this PR came about.

package-build/package-build.el
```diff
@@ -311,8 +311,11 @@ Returns the package version as a string."
 
 (defun package-build--git-repo (dir)
   "Get the current git repo for DIR."
-  (package-build--run-process-match
-   "Fetch URL: \\(.*\\)" dir "git" "remote" "show" "-n" "origin"))
+  (or (ignore-errors
```
milkypostman: lets not do this as it's unnecessary.

milkypostman:
must be submitted to the package-build repo.

PR #5220, opened 2018-01-01
Add recipe for flycheck-xcode

jojojames:

Flycheck extension for xcode projects. If the project has an associated .xcodeproj file, it is an xcode project and will use the same linting (warnings/errors/info) settings Xcode uses.

https://github.com/jojojames/flycheck-xcode

Author/Maintainer

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/flycheck-xcode
+++ b/recipes/flycheck-xcode
@@ -1,0 +1,1 @@
+(flycheck-xcode :fetcher github :repo "jojojames/flycheck-xcode")


jojojames:
@purcell 

Somewhat related question. 

At some point I'd like to add flymake support. Would that go to a separate package? Most of the code should be very similar and it'd be a maintenance burden to support both packages (probably too much to even write the backend in the first place). What's the recommended approach when wanting to support both flycheck and flymake? I'd be happy to change the name (xcode-check?) if that meant I could add both flymake and flycheck code to the same file.

I'm planning to add a few more flycheck extensions soon.

purcell:
Thanks, and sorry for the slow turnaround. Re. the current code, you shouldn't [shim built-in functions/macros](https://github.com/jojojames/flycheck-xcode/blob/master/flycheck-xcode.el#L39-L45) - the [elisp conventions](https://www.gnu.org/s/emacs/manual/html_node/elisp/Coding-Conventions.html) specifically request this. But in any case, I'll go ahead and merge this.

Re. flymake support, you wouldn't include that in this package, because installing it would force the installation of `flycheck`, but assuming that you'd be supporting the `flymake` versions in Emacs 25.1 and up, you could do things the other way around, ie. make `flycheck-xcode` depend on `flymake-xcode`, and put the shared code there.

jojojames:
@purcell 

> Thanks, and sorry for the slow turnaround.

No problem, thanks for taking a look.

> Re. the current code, you shouldn't shim built-in functions/macros - the elisp conventions specifically request this. But in any case, I'll go ahead and merge this.

Is this the relevant section?

"It is a bad idea to define aliases for the Emacs primitives. Normally you should use the standard names instead. The case where an alias may be useful is where it facilitates backwards compatibility or portability."

As far as I understand, there's no way to when-let* in emacs 25 but when-let will throw a deprecation error so the defalias is in the spirit of backwards compatibility.
If there's a better way to do it, I will switch to that.

> Re. flymake support, you wouldn't include that in this package, because installing it would force the installation of flycheck, but assuming that you'd be supporting the flymake versions in Emacs 25.1 and up, you could do things the other way around, ie. make flycheck-xcode depend on flymake-xcode, and put the shared code there.

Thanks, I'm planning on waiting for Emacs 26 to be released which contains the flymake refactor.

purcell:
> Is this the relevant section?

Yes - the point is not about whether to alias things, it's more about whether to alias them as the *original* names: see the next bullet point after the one you quoted.

> Thanks, I'm planning on waiting for Emacs 26 to be released which contains the flymake refactor.

Cool. Then you might find that your `flymake` code depends on a newer Emacs than your `flycheck` code, in which case you might prefer not to have to force flycheck users to have Emacs 26. Happy to discuss options nearer the time: it'd be fine for both `fly*` packages to depend on a small common library, too.

jojojames:
> Yes - the point is not about whether to alias things, it's more about whether to alias them as the original names: see the next bullet point after the one you quoted.

Thanks, I stand corrected.

Just so we're on the same page, does this look to be the correct change?

```
(eval-and-compile
  (if (version< emacs-version "26")
      (with-no-warnings
        (defalias 'flycheck-xcode-if-let* #'if-let)
        (defalias 'flycheck-xcode-when-let* #'when-let)
        (function-put #'flycheck-xcode-if-let* 'lisp-indent-function 2)
        (function-put #'flycheck-xcode-when-let* 'lisp-indent-function 1))
    (defalias 'flycheck-xcode-if-let* #'if-let*)
    (defalias 'flycheck-xcode-when-let* #'when-let*)))
```

> Cool. Then you might find that your flymake code depends on a newer Emacs than your flycheck code, in which case you might prefer not to have to force flycheck users to have Emacs 26. Happy to discuss options nearer the time: it'd be fine for both fly* packages to depend on a small common library, too.

Sounds good, I'll wait until then to raise the issue.

purcell:
> Just so we're on the same page, does this look to be the correct change?

Yep, exactly - it's a bit clunky, but it's safer.

jojojames:
Thanks for the advice!

PR #5222, opened 2018-01-02
Add play-crystal.el

veelenga:

Emacs to [play.crystal-lang.org](https://play.crystal-lang.org/#/crystal) integration.

play.crystal-lang.org is a web resource to submit/run/share [Crystal](https://crystal-lang.org/) code.

This package allows you to use this resource without exiting your favorite Emacs.

Features:
* Allows to fetch code into Emacs buffers from play.crystal-lang.org
* Allows to submit code to play.crystal-lang.org directly from Emacs
* Allows to browse play.crystal-lang.org

Fixes https://github.com/veelenga/play-crystal.el/issues/2

/cc @brantou

https://github.com/veelenga/play-crystal.el

Owner

**None needed**

### Checklist

### DIFF
--- a/recipes/play-crystal
+++ b/recipes/play-crystal
@@ -1,0 +1,1 @@
+(play-crystal :repo "veelenga/play-crystal.el" :fetcher github)


purcell:
Thanks! Quick feedback:

- Don't use Markdown in the Commentary, please.
- Require `cl-lib` instead of the now-deprecated `cl` library: the `cl-` functions and macros you're using from from `cl-lib` anyway. When you do this, you would generally also drop the `eval-when-compile`.
- You should add at least one `;;;###autoload` for top-level interactive functions that the user might call to start using this library, so that they don't need to `require` it first.

veelenga:
Thanks for you feedback. I fixed it.

> Don't use Markdown in the Commentary, please.

The idea was to maintain one version of doc and autogenerate readme using [make-readme-markdown](https://github.com/mgalgs/make-readme-markdown). What are the drawbacks of this?

purcell:
> The idea was to maintain one version of doc and autogenerate readme using make-readme-markdown. What are the drawbacks of this?

Emacs has its own markup for docstrings, including `Commentary`, e.g. Using backtick then `'` to surround symbols. But it's not a huge deal.

I'll go ahead and merge this.

You seem to be interested in having actual version numbers, e.g. `0.1.1`: I suggest you leave them out of the file itself, and instead apply git version tags. MELPA Stable will then build corresponding packages with those versions, and the version number will be placed in the `Package-Version` header of those files. (Also, it's error-prone to put version numbers directly in the file: there are already both 0.1.1 and 0.1.0 in the current `master` version of `play-crystal.el`.)

Welcome to MELPA!

PR #5224, opened 2018-01-04
[add] css-autoprefixer

kkweon:
Add CSS autoprefix.

Make the following CSS
```css
div {
  display:flex;
}
```
into

```css
div {
  display:-webkit-box;
  display:-ms-flexbox;
  display:flex;
}
```

https://github.com/kkweon/emacs-css-autoprefixer

I am the owner of the package

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/css-autoprefixer
+++ b/recipes/css-autoprefixer
@@ -1,0 +1,1 @@
+(css-autoprefixer :fetcher github :repo "kkweon/emacs-css-autoprefixer")


purcell:
Thanks for your patience. Couple of little issues in the code:

- Files with spaces or other special shell characters in their name will break your command construction. Use `shell-quote-argument`.
- Run the command before you delete the region its output will replace, so that if the command fails, the content will not have been lost.

kkweon:
@purcell Thanks for the feedback. I fixed it as below.

- I added `shell-quote-argument` for the filenames.
- I added a fail check test by reading the exit code from `call-process` If there is an error, it will display the error message in a separate buffer instead of erasing it.

purcell:
Great, thanks - merged. Welcome to MELPA!

PR #5226, opened 2018-01-05
Add inf-crystal recipe

brantou:

Run a Inferior-Crystal process in a buffer.

https://github.com/brantou/inf-crystal.el

 **None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/inf-crystal
+++ b/recipes/inf-crystal
@@ -1,0 +1,1 @@
+(inf-crystal :repo "brantou/inf-crystal" :fetcher github)


purcell:
Thanks. This all looks good apart from the `;;;###autoload (add-hook 'crystal-mode-hook 'inf-crystal-minor-mode)` line -- you shouldn't automatically enable the library just because it is installed, because that should be the user's choice. Packages should work like built-in Emacs libraries, and if all the built-in libraries auto-enabled themselves, we'd all get very annoyed! :-)

brantou:
Thanks. Automatic loading is really awkward, I have deleted it.https://github.com/brantou/inf-crystal.el/commit/dd5c85e621976ea09b602182a15396e3b510ec63

purcell:
Great, merged.

PR #5227, opened 2018-01-05
Add elcord

Zulu-Inuoe:

Communicate with locally installed Discord to display Rich Presence information.

https://github.com/Zulu-Inuoe/elcord

I've forked the original package [found here](https://github.com/Mstrodl/elcord) and heavily expanded on it to make it add Windows support, along with making the package more stable and robust.

I've found no way to contact the original author, but I've kept their authorship and copyright claims in the source and license.

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/elcord
+++ b/recipes/elcord
@@ -1,0 +1,6 @@
+(elcord
+ :fetcher github
+ :repo "Zulu-Inuoe/elcord"
+ :files
+ (:defaults
+  "stdpipe.ps1"))


purcell:
Thanks, particularly for your patience - I'm currently very busy! A few quick notes:

- Don't bother wrapping those `require`s in `eval-when-compile` -- I think that's misplaced here.
- You should include in the `Commentary` some information about how people would use this package, since Emacs packages should be self-documenting as far as possible.
- Relatedly, if there are one or more commands which a user might commonly use to launch the code in this package, consider adding `;;;###autoload` marker comments.

Hope that helps!

Zulu-Inuoe:
No, thank you for taking the time to review this!

A question regarding the `eval-when-compile`: The reason I have the `require` forms wrapped as such, is because I use macros from both those modules in my code. I may be misunderstanding the compilation process in elisp, but wouldn't those macro definitions need to be in place for my code to compile?

I've (hopefully) addressed the two other notes in this commit: 
https://github.com/Zulu-Inuoe/elcord/commit/343a2bd3780a99275074e9c19199fe6d11547227

I'd be greatly thankful if you could verify that's correct. Once I get an OK, should I update the release tag to the latest commit, or should I make a new release and update the recipe to point to that?

It's a relatively straightforward package, so I documented enabling it, what information is displayed, and an option to change the information displayed.
I wasn't sure how much detail to go into with the workaround on Windows, the way icons are queried, etc. I'm not opposed to having it there, but I didn't want to overdo it, as these are things a typical user wouldn't need to know.
Though I do go into greater detail in the readme.

Once again, thanks!

purcell:
> A question regarding the eval-when-compile: The reason I have the require forms wrapped as such, is because I use macros from both those modules in my code. I may be misunderstanding the compilation process in elisp, but wouldn't those macro definitions need to be in place for my code to compile?

The `require` is still evaluated at compilation time, so the macros will still be in place. [Here's more info](https://stackoverflow.com/questions/2904555/what-is-the-purpose-of-the-emacs-function-eval-and-compile). `cl.el` was a mess in this regard, but `cl-lib` is fine.

Re. the windows workaround, I'd suggest not defining `elcord--stdpipe-path` conditionally: you can simply note that it is only used on Windows.

purcell:
P.S. Yes, the other fixes are correct.

Zulu-Inuoe:
Alright. I've made the corresponding changes in https://github.com/Zulu-Inuoe/elcord/commit/91c665fd832ef3b79c3eb810b7a6b08979a352cd
Please let me know if there is anything else I should do.

Thank you for your help!

purcell:
Nice, thanks - merged! Welcome to MELPA. :-)

PR #5228, opened 2018-01-06
Add library.el

cadadr:
Library.el is a PDF library manager.  Presently it only displays a
concatenated tabular list of PDF files from many locations, and allows
to open the files or view metadata from that list.  But improvements
are planned.

Library.el displays a combined list of PDF files from a user-configured `library-path` in tabular format, where title, author and path are shown for each file.  Sorting by columns and viewing the file or the metadata is possible.  Some more functionality, especially for manipulation of metadata and moving files around, is planned.

I'm aware that the package name is kind-of too generic, if that's a problem I can change it to sth. like "librarian" to avoid any confusions.

https://github.com/cadadr/elisp/blob/master/library.el

I am the maintainer.

None needed.

### Checklist

(Some extra notes here: package-lint says pdf-tools is not installable, but installation test works.  Except, pdf-tools installation is problematic, the epdfinfo program in that package does not get built and installed correctly.)

Please confirm with `x`:


cadadr:
I've updated the checklist, it seemed as if it was incomplete on the PRs page because I forgot to remove spaces between the brackets editing the comment in the github website. Sorry about that.

alphapapa:
Hi,

This looks neat, thanks.  Have you considered using a more descriptive name?  `library` sounds like it could do many different things, e.g. my first thought was that it was related to Emacs packages or development libraries.  Especially since it's specific to PDF files, something like `pdf-library` would be more helpful to users.

cadadr:
You're welcome :) Yes, I agree that the name is too overbroad.  But I'd
rather not use a name that indicates a specific file format (PDF),
as---even though it's only PDFs ATM, and the documentation says so
too---it can grow to include a bunch more file types, e.g. postscript,
e-books, HTML (say, as in web saves for example); possibilites that I
plan to explore as I myself use the program.  A couple alternatives I
have in mind are (i) bibliothek, the german word for library, seemingly
familiar to anglophones worldwide, and (ii) baghdad, for the house of
wisdom that was there during the islamic golden age.  Would any of these
be apt for use as the new package name?

alphapapa:
Well, IMO "bibliothek" would be better than "library" since it's not a rather vague, generic English word that is also software jargon.  And I guess it would also be better than "baghdad" because of all the recent history surrounding Iraq.  My two cents.

cadadr:
Yeah that's quite reasonable with Baghdad.  I'll go with bibliothek then, update the patch and the source repo as soon as I can.  Thanks.

cadadr:
I have made the suggested changes, except adding autoloads, because I wanted to ask first: Currently the only interactive function that's meant to be used directly is `bibliothek`, the rest are not useful outside the buffer `bibliothek` generated, thus kind-of private to the package.  If it'd still be better to make those interactive functions autoloadable anyways, I have no problem marking them as such.

alphapapa:
> I have made the suggested changes, except adding autoloads, because I wanted to ask first: Currently the only interactive function that's meant to be used directly is bibliothek, the rest are not useful outside the buffer bibliothek generated, thus kind-of private to the package. If it'd still be better to make those interactive functions autoloadable anyways, I have no problem marking them as such.

No, I think you're right that only commands that serve as the "entrance" into the package need to be autoloaded.  I should have looked more carefully before saying that.  :)

cadadr:
No problem at all! Thanks a lot for your feedback.

purcell:
Thanks! Sorry for the slow turnaround. Quick notes:

- [Here](https://github.com/cadadr/elisp/blob/8925ea85143c4c91589213fbacb275e81968d76d/bibliothek.el#L86) you should match end-of-string `\\'` rather than end-of-line `$`.
- The [parent custom group](https://github.com/cadadr/elisp/blob/8925ea85143c4c91589213fbacb275e81968d76d/bibliothek.el#L73) should probably be `'applications` rather than `'emacs`.
- [bibliothek-path](https://github.com/cadadr/elisp/blob/8925ea85143c4c91589213fbacb275e81968d76d/bibliothek.el#L76) should be a `defcustom`.

I'll go ahead and merge this, after removing the `:files` spec from the recipe: it's redundant, because the default is fine here, and therefore preferable.

cadadr:
You're welcome, and no problem at all!  Thanks a lot for merging the
patch, and for the effort you put in to maintaining this integral part
of the Emacs ecosystem.

I have addressed the three issues you've pointed to, thanks.  I'll
consider the point about separate repos too.

tarsius:
> I have addressed the three issues you've pointed to, thanks. I'll consider the point about separate repos too.

I would like to second that request. This would allow me to properly mirror the three packages from your repository on the [Emacsmirror](https://github.com/emacsmirror) without having to choose between:

1. Mirroring each package individually by periodically importing the respective file and committing it with the commit message "updates". I.e. without having to discard the history.
2. Randomly pick one of the packages as *the main package* and make the other libraries/packages part of that package even though they are not actually related.

tarsius:
Related #5018.

PR #5229, opened 2018-01-07
add recipe goldendict

stardiviner:

A simple wrapper of "goldendict" command as dictionary.

https://github.com/stardiviner/goldendict.el

maintainer

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/goldendict
+++ b/recipes/goldendict
@@ -1,0 +1,1 @@
+(goldendict :fetcher github :repo "stardiviner/goldendict.el")


alphapapa:
Hi there,

This looks neat.  However, please be careful with calling shells with user input: https://github.com/stardiviner/goldendict.el/blob/master/goldendict.el#L42

You should at least escape it with `shell-quote-argument`, but you don't actually need to run it as a shell command at all.  Instead you can call the process directly.  Here's a macro that I use:
```el
(defmacro call-process-with-args (process &rest args)
  "Return results of running PROCESS with ARGS."
  (declare (indent defun))
  `(with-temp-buffer
     (unless (= 0 (call-process ,process nil t nil
                                ,@args))
       (user-error ,(concat process " failed")))
     (buffer-substring-no-properties (point-min) (point-max))))
```
Used like:
```el
(defun get-day-of-week (string)
  "Parse STRING with the GNU `date' command and return day-of-week as string."
  (call-process-with-args
    "date" "-d" string "+%a"))
```

stardiviner:
I don't know how to use your code in my case. Because it does not show result in buffer. It popup a GUI popup window. So use shell command is fine I think.

alphapapa:
> it does not show result in buffer. It popup a GUI popup window.

Ah, my mistake.  

> Thanks for your suggestion, I updated my code by applying shell-quote-argument on word.

Did you forget to push?  I don't see that change in your repo.  :)

However, you should probably still use `call-process`.  The docstring for `shell-command` says:
```
In Elisp, you will often be better served by calling ‘call-process’ or
‘start-process’ directly, since it offers more control and does not impose
the use of a shell (with its need to quote arguments).
```
Maybe I'm making too big a deal out of it, but as far as I know, it's generally safer and better to avoid using shell commands whenever possible.

stardiviner:
@alphapapa Aha, Thanks, I see. I have not check out the docstring of `call-process`. I take your point. Will update code soon. Might network delay or push failed. I will check my code. Thanks.

stardiviner:
Now, updated. Check it out. Thanks for reviewing.

alphapapa:
Cool, although now you don't need to use `shell-quote-argument` on `word`.  :)

stardiviner:
Done.

purcell:
Great, thanks.

PR #5230, opened 2018-01-07
Add pipenv

pwalsh:

[Pipenv](http://pipenv.readthedocs.io/en/latest/) porcelain for Emacs.

https://github.com/pwalsh/pipenv.el

Maintainer.

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/pipenv
+++ b/recipes/pipenv
@@ -1,0 +1,1 @@
+(pipenv :repo "pwalsh/pipenv.el" :fetcher github)


pwalsh:
Thanks for the feedback @alphapapa 

I'll try to add some screenshots and more explanation in the coming days.

pwalsh:
The [README](https://github.com/pwalsh/pipenv.el/blob/master/README.org) now has fairly comprehensive documentation.

Compro-Prasad:
This seems cool. Can't wait to check it out.

purcell:
Thanks! Quick feedback:
- Don't expand paths [using executable-fine](https://github.com/pwalsh/pipenv.el/blob/f956a72552d07c4e612609dce34875dba84fe2c2/pipenv.el#L43) - `pipenv` would need to be on the path either way, so if you don't expand it, the path lookup will be done later at runtime, not when the package is initially loaded. Various setups can cause paths to be updated long after init, sometimes even on a per-buffer basis.
- [`inside Emacs`](https://github.com/pwalsh/pipenv.el/blob/f956a72552d07c4e612609dce34875dba84fe2c2/pipenv.el#L34) is redundant.
- `(if x y nil)` [here](https://github.com/pwalsh/pipenv.el/blob/f956a72552d07c4e612609dce34875dba84fe2c2/pipenv.el#L306) is equivalent to `(when x y)`
- [This](https://github.com/pwalsh/pipenv.el/blob/f956a72552d07c4e612609dce34875dba84fe2c2/pipenv.el#L331-L333) could simply be written `(locate-dominating-file "." "Pipfile")`, though for the sake of tidiness you want `default-directory`, not `"."` there.
- `M-p` is a super-common keybinding, so you should choose a different default, e.g. `C-c C-p`.
- It's inappropriate to enable `pipenv-mode` in `python-mode` [automatically upon installation](https://github.com/pwalsh/pipenv.el/blob/f956a72552d07c4e612609dce34875dba84fe2c2/pipenv.el#L409-L410). Like Emacs' standard libraries, installed packages should be available for use, but not enabled: imagine if all of Emacs' libraries were on by default. :-)
- This [arbitrary sleep](https://github.com/pwalsh/pipenv.el/blob/f956a72552d07c4e612609dce34875dba84fe2c2/pipenv.el#L285) looks suspicious... what are you trying to do? There's almost certainly a different and robust way to achieve it.

Hope that helps!

pwalsh:
Thanks for the review @purcell 

> Don't expand paths using executable-fine - pipenv would need to be on the path either way, so if you don't expand it, the path lookup will be done later at runtime, not when the package is initially loaded. Various setups can cause paths to be updated long after init, sometimes even on a per-buffer basis.

https://github.com/pwalsh/pipenv.el/commit/4e93d351ea920d373d6b676b46b50ca9c9d42176

> inside Emacs is redundant.

https://github.com/pwalsh/pipenv.el/commit/20179ce1ecc6252dfbb6bf37a0745a6edcfda627

> (if x y nil) here is equivalent to (when x y)

https://github.com/pwalsh/pipenv.el/commit/d69e38f81a95b47cccd7d0036570d71ed8b39fb1

> This could simply be written (locate-dominating-file "." "Pipfile"), though for the sake of tidiness you want default-directory, not "." there.

https://github.com/pwalsh/pipenv.el/commit/738c8496308f9456f81b46655b931482a2781e6e

(Nice tip! Thanks.)

> It's inappropriate to enable pipenv-mode in python-mode automatically upon installation. Like Emacs' standard libraries, installed packages should be available for use, but not enabled: imagine if all of Emacs' libraries were on by default. :-)

https://github.com/pwalsh/pipenv.el/commit/06e51ae97f2718f8876603dbf320af28eea37186

> This arbitrary sleep looks suspicious... what are you trying to do? There's almost certainly a different and robust way to achieve it.

https://github.com/pwalsh/pipenv.el/commit/36eaa63f1ae00e7b5bd785c4e964f3d1355ec1fa

> M-p is a super-common keybinding, so you should choose a different default, e.g. C-c C-p.

https://github.com/pwalsh/pipenv.el/commit/2af4f010e908e8e5dfe81278b2232c0580cfd424

ninrod:
New method of learning ELisp: read @purcell reviews on melpa 
🤓

purcell:
> New method of learning ELisp: read @purcell reviews on melpa

LOL.

Merging now - nice work @pwalsh, and thanks for your patience. Welcome to MELPA!

PR #5231, opened 2018-01-07
Adding "windswap" recipe.

amnn:

An Emacs library for swapping buffers between neighbouring windows in a frame, in the style of `windmove`.

https://github.com/amnn/windswap

I am the creator/maintainer.

**None needed**

### Checklist

Please confirm with `x`:

- [x] Built and installed with `package-install-from-file` ~~I've built and installed the package using the instructions in [CONTRIBUTING.md](https://github.com/melpa/melpa/blob/master/CONTRIBUTING.md)~~

### DIFF
--- a/recipes/windswap
+++ b/recipes/windswap
@@ -1,0 +1,1 @@
+(windswap :repo "amnn/windswap" :fetcher github)


mattiasb:
This sounds really similar to [buffer-move](https://github.com/lukhas/buffer-move).

amnn:
@mattiasb, you're right... it's exactly that. Creating a package with identical functionality was probably not the best way to search MELPA. I did search first, honest :wink: I was searching for packages that swapped windows, instead of moved buffers, alas. Thanks for pointing it out to me!

mattiasb:
@amnn Yeah Melpa is big and naming is hard, not surprised you missed it. :) Nice work regardless!

PR #5232, opened 2018-01-07
Add org-wild-notifier

akhramov:

This package provides iCal-style notifications for org-agenda, configurable via agenda properties.

https://github.com/akhramov/org-wild-notifier.el

The author

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/org-wild-notifier
+++ b/recipes/org-wild-notifier
@@ -1,0 +1,1 @@
+(org-wild-notifier :fetcher github :repo "akhramov/org-wild-notifier.el")


alphapapa:
Hi,

This looks cool!  Here are a few things you might want to change:

* At least your first three `defvar`s should probably be `defcustom`s, and you probably want to `defgroup` a group too.
* I like how you've made use of `dash` and `pcase`.  One function you may have missed: instead of using `(remove nil ...` you can use `-non-nil`.  :)

Looks like you did a great job!

akhramov:
Hi, thanks for the tips. I addressed all of them except the last one in  [3c667f96482](https://github.com/akhramov/org-wild-notifier.el/commit/3c667f96482a8c83a83dceacc2cafefd4c23afec). 

It's a tough question. Rationale against specifying Org as a dependency, apart from the fact that it's already bundled with GNU Emacs, is that org isn't hosted by MELPA.

So, if MELPA is the only archive it won't be possible to build the package.

alphapapa:
AFAIK that only affects installing packages.  So e.g. if you depend on Org 9.x but a user still has Org 8.x (whether bundled or not), it will warn them if they try to install your package.  I declare dependencies on Org in all of my Org-related packages, and it works fine.

akhramov:
Sorry, I meant installation.

It looks like package installation via `package-install-file` fails if I specify org as a dependency (please see [travis build results](https://travis-ci.org/akhramov/org-wild-notifier.el), [code adding the dependency](https://github.com/akhramov/org-wild-notifier.el/pull/1/files)).

Could you please point me, what am I doing wrong?

alphapapa:
I'm guessing that's because Org is not included in the Emacs archive that you're extracting in the Travis build.  An alternative is to use EVM, which is used in the MELPA build.  I also use it here: https://github.com/alphapapa/helm-org-rifle/blob/master/.travis.yml

akhramov:
Thank you very much for the example. Unfortunately, no, even with Cask /  EVM I'm having the same error:
[travis](https://travis-ci.org/akhramov/org-wild-notifier.el/builds/326704343) / [source](https://github.com/akhramov/org-wild-notifier.el/pull/1/files).

Can we try adding org as  a dependency to `helm-org-rifle` to see if CI would be green?

alphapapa:
Sorry, I should have pointed to https://github.com/alphapapa/org-super-agenda instead of helm-org-rifle.  org-super-agenda does declare a dependency on Org 9.0, and it works with its Travis config (the test is failing due to an unrelated unit test later in the process).

I don't know why it would still fail, but I haven't looked very deeply into how this part of the package system and Travis work--I mainly copied from MELPA's and maybe package-lint's Travis configs.  Does it work if you change it to Org 9.0?

If you can't get it working easily, I'm not sure it's important enough to spend too much time on it.  It's obvious that the package is Org-related, so unless it requires something in very recent versions of Org (and since you're setting it to Org 8, I'm guessing it doesn't), I'm not sure it matters.

akhramov:
Long story short, I had to manage deps by cask, and run tests using cask exec.

Thank you for your patience and assistance, everything should be good now.

purcell:
Thanks. Sorry for the slow turnaround. I'll go ahead and merge this, but you *really* need to put info in the .el file about how to use the code: there's nothing in there which mentions using `org-wild-notifier-start`. And on that topic, since you've got "start" and "stop" you might like to consider hiding those and providing a minor mode instead, which would be more idiomatic and amenable to customisation.

Welcome to MELPA!

akhramov:
@purcell , nothing to sorry about. Thank you for the review and for your suggestions. I'll implement them right away.

Thanks!

PR #5233, opened 2018-01-08
Original repo for Leuven-theme doesn't provide release tags.

bzindovic:
Original repo for Leuven-theme doesn't provide release tags so I've forked original repo and added a tag. I've added this recipe for a stable release of this excellent theme as a temporal solution, until original author adds them upstream. I've contacted original author and suggested adding tags upstream (see the issue: https://github.com/fniessen/emacs-leuven-theme/issues/50) but he hasn't responded yet.

This package installs tagged version of Leuven-theme, intended for MELPA Stable users.

https://github.com/bzindovic/melpa

I am an enthusiast user and would like to add the Leuven-theme to MELPA Stable.

https://github.com/fniessen/emacs-leuven-theme/issues/50

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/leuven-theme-stable
+++ b/recipes/leuven-theme-stable
@@ -1,0 +1,3 @@
+(leuven-theme-stable
+ :fetcher github
+ :repo "bzindovic/emacs-leuven-theme")


milkypostman:
We need to have the original repo tag this. Unless they agree to have you maintain it I am not comfortable switching to some arbitrary fork.

bzindovic:
Original author,although active, hasn't responded to my issue posted almost two months ago. I suppose that this is not unusual and will wait for his response.

With kind regards.

fniessen:
(Late in the flow, but...) What should I do to make this happen?  Just add a Git tag?  Which one?  Tried to find doc, but not yet found.
(I did well find this: https://github.com/melpa/melpa/issues/6656. Quid?)

milkypostman:
If you tag a version string it will work. For example: `1.0.0`

I'd have to look at the code but I also think things like `v1.0` also works.

milkypostman:
Here is the current relevant doc:

> :version-regexp is a regular expression for extracting a version-string from the repository tags. The default matches typical version tags such as 1.0, R16 or v4.3.5, so you should not override it unless necessary. For an unusual tag like "OTP-18.1.5", we might add :version-regexp "[^0-9]*\\(.*\\)" to strip the "OTP-" prefix. The captured portion of the regexp must be parseable by Emacs' version-to-list function.

fniessen:
DONE -- Thanks a lot!

PR #5234, opened 2018-01-10
Add recipe for the Plaster emacs extension.

Shinmera:

This package allows convenient managing of pastes on a remote Plaster paste service instance.

https://github.com/shirakumo/plaster

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/plaster
+++ b/recipes/plaster
@@ -1,0 +1,3 @@
+(plaster :repo "shirakumo/plaster"
+         :fetcher github
+         :files ("plaster.el"))
--- a/recipes/plaster
+++ b/recipes/plaster
@@ -1,3 +1,2 @@
 (plaster :repo "shirakumo/plaster"
-         :fetcher github
-         :files ("plaster.el"))
+         :fetcher github)


alphapapa:
This looks really cool, thanks for sharing.

Just curious, why use `(with-output-to-string (princ...` instead of, e.g. `(format "%s"` or `(format "%S"`?

Shinmera:
I come from Common Lisp, where you have `princ-to-string`, so I didn't even think of `format` for a second.

Thanks for the hints, I've updated it accordingly.

alphapapa:
> I come from Common Lisp, where you have princ-to-string, so I didn't even think of format for a second.

Ah, cool, I thought either that was it, or you knew something I didn't and maybe it was better to do it that way.  I'm coming from the other direction: started with Elisp and interested in doing more in CL in the future.  I've been looking at your projects and they look very interesting.  Your output is prolific!  With your knowledge and background, you should consider becoming an Emacs core dev.  :)

I also see that you're the Lisp Treehouse video series guy.  Very cool.  I've seen those posts in my /r/lisp feed and have been meaning to check them out.

BTW, I see you or your "collective" wrote a new chat system, Lichat?  Looks pretty cool from what I can tell (there doesn't seem to be one home page for the whole project, so it's a bit hard to get a handle on it).  What do you think of Matrix?  I've been doing some work with Jay Kamat on the Emacs Matrix client started by Ryan Rix.  It would be cool if Matrix used s-expressions instead of JSON, but it seems like a pretty good system with a bright future.  Feel free to check out our client if you're interested.

Shinmera:
Thanks for the praise. I'm not sure I have the time to contribute to Emacs, unfortunately.

I'd be glad to have more people around during the streams!

Right, haven't gotten around to making a fancy website for Lichat. I haven't looked into Matrix much, so I don't really have any thoughts on it. Lichat was created primarily to provide a chat system that could be accessed without any account hassle. It also just gave me a nice excuse to see if I could come up with a simpler protocol than IRC.

purcell:
Thanks, this looks good. Sorry for the slow turnaround. Just 2 little points:

- You should add `;;;###autoload` cookie comments for the minor mode and any other commands a user might use as an entry point to this library.
- Please remove the `:files` from the recipe, since it's redundant here.

purcell:
> Please remove the :files from the recipe, since it's redundant here.

Never mind, I've done this myself and will now merge the recipe - welcome to MELPA!

Shinmera:
Thanks for the hint on autoload and for merging things in!

PR #5235, opened 2018-01-12
Add recipes/cquery

MaskRay:

emacs-cquery is a client for cquery, a low-latency language server supporting multi-million line C++ code-bases, powered by libclang.

It leverages lsp-mode, but also provides some cquery extensions to LSP:

* semantic highlighting
* inactive region (e.g. a `#if false` region)
* cross references: `$cquery/base $cquery/callers $cquery/derived $cquery/vars`

 Thus the name does not start with `lsp-`.

https://github.com/cquery-project/emacs-cquery

Contributor

cquery.el @topisani

cquery @jacobdufault

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/cquery
+++ b/recipes/cquery
@@ -1,0 +1,1 @@
+(cquery :repo "cquery-project/emacs-cquery" :fetcher github)


purcell:
Thanks. Quick feedback:
- Don't define a face that has only `foreground` or `background`. Either `:inherit` a built-in face by default (preferred), or define complete face variants for both light and dark backgrounds.
- If exposing variables as custom variables, then `"."` is not sufficient documentation for users to figure out how they should use them. :-)  Same goes for all the other functions with that docstring. It doesn't take long to write a short description.
- Would it make sense to add an `;;;#autoload` cookie for the `lsp-cquery-enable` function?

MaskRay:
Changed to

```elisp
(defface cquery-inactive-region-face
  '((t :inherit shadow))
  "The face used to mark inactive regions."
  :group 'cquery)

(defface cquery-code-lens-face
  '((t :inherit shadow))
  "The face used for code lens overlays."
  :group 'cquery)
```

Language server protocol code lens is still a brand new idea to users. I'm not certain how to pick the best default color.

@topisani

purcell:
> Language server protocol code lens is still a brand new idea to many users. I'm not certain how to pick the best default color.

True. Hopefully themes will add support over time.



> lsp-cquery-enable is a function generated by the macro (lsp-define-stdio-client ....), so there is no entity to add ;;;#autoload

It's still possible to add autoload comments, actually: you just have to include the signature, e.g. `;;;###autoload (autoload 'lsp-cquery-enable "cquery")`

Anyway, I'll go ahead and merge this now - thanks!

PR #5239, opened 2018-01-14
schlau-compile added

flajann2:
This is basically a clone of smart-compile with extra functionality.

https://github.com/flajann2/elisp

Maintainer.

This package has been properly renamed and tested.
### Checklist

Please confirm with `x`:

- [x ] I've used the latest version of [package-lint](https://github.com/purcell/package-lint) to check for packaging issues, and addressed its feedback
- [x ] I've built and installed the package using the instructions in [CONTRIBUTING.md](https://github.com/melpa/melpa/blob/master/CONTRIBUTING.md)

### DIFF
--- a/recipes/schlau-compile
+++ b/recipes/schlau-compile
@@ -1,0 +1,4 @@
+(schlau-compile
+  :fetcher github
+  :repo "flajann2/elisp"
+  :files ("schlau-compile.el"))


purcell:
Thanks! Quick feedback:
- This should live in its own repo so that it can be version-tagged independently for inclusion in MELPA Stable, and so that it wouldn't need the `:files` spec to protect against future neighbouring changes in the shared repo.
- We prefer not to include shallow forks in MELPA where possible, and yours hasn't diverged much from smart-compile: have you tried submitting your enhancements to @zenitani so that they could reach a larger audience?

flajann2:
I have not, and I plan to do more enhancements and the like.So I expect the divergence to grow even more eventually as I would like to take Emacs "to the next level".

Would it make sense to put it in in a seperate branch instead of a seperate repo? You are talking about the plugin as opposed to the melpa file that references it?

purcell:
> Would it make sense to put it in in a seperate branch instead of a seperate repo? You are talking about the plugin as opposed to the melpa file that references it?

No, not a separate branch. It's generally tidier and better for MELPA if each library lives in its own distinct repo, not in a location shared with other elisp.

> I have not, and I plan to do more enhancements and the like.So I expect the divergence to grow even more eventually as I would like to take Emacs "to the next level".

If so, it would still be ideal to collaborate with @zenitani if possible: you could both agree to nominate your actively-developed fork as the one distributed via MELPA, for example, then existing `smart-compile` users would start to receive your enhancements. I'd prefer that the two of you try to discuss this briefly than to rush into adding the renamed version.

DamienCassou:
What is the status?

flajann2:
I have not had a chance to get to this yet. All you say about this makes sense. It will take me some time to deal with it.

tarsius:
I am closing this for now, sorry. When you are ready, then ask us to reopen this again.

PR #5240, opened 2018-01-15
Add pynt

ebanner:

Provide commands for generating and interacting with jupyter notebooks from python code via the [Emacs IPython Notebook](http://millejoh.github.io/emacs-ipython-notebook/) package.

https://github.com/ebanner/pynt

Author

**None needed**

### Checklist

### DIFF
--- a/recipes/pynt
+++ b/recipes/pynt
@@ -1,0 +1,3 @@
+(pynt :repo "ebanner/pynt" 
+      :fetcher github
+      :files ("pynt/*"))


alphapapa:
Hi,

A little feedback:

* I'm guessing you should use `defcustom` for `pynt-epc-port`.

purcell:
Thanks! I sent some fixes in https://github.com/ebanner/pynt/pull/20, but there are a couple more issues, some of which echo @alphapapa's notes:
- Don't [set the local values of the `print-*` vars at the top level](https://github.com/ebanner/pynt/blob/4ea3ba303dbac4ec11e98815e26c0ca16db22484/pynt/pynt.el#L110-L112) - this will only affect the buffer in which the file is first loaded, and is generally nonsense. If you really need these settings (and you shouldn't, except very rarely), you should `let`-bind them locally while running the tiny piece of code that requires those settings.
- Don't perform logic [based on buffer names](https://github.com/ebanner/pynt/blob/4ea3ba303dbac4ec11e98815e26c0ca16db22484/pynt/pynt.el#L122). Buffers can be arbitrarily renamed, and not every buffer is backed by a file. If you're managing the buffer, you can set its associated filename, or note a local variable in that buffer which you can consult later.
- `buffer-string` could likely replace `pynt-get-buffer-string`.
- Add a package dependency on `seq`, ideally whichever version is in GNU ELPA, and a `require` form.
- Should [this](https://github.com/ebanner/pynt/blob/4ea3ba303dbac4ec11e98815e26c0ca16db22484/pynt/pynt.el#L279) use your `pynt-log` function instead of `message`? Probably.
- `lexical-let` came from `cl`, which my PR removes, but it's obsolete anyway: declare `lexical-binding` for your file instead. (Try running `M-x auto-insert` in an empty `.el` file to see how.)

Hope that helps!

ebanner:
Thanks for the feedback and attempts to make my emacs package truly great! I really appreciate it.

First I will respond to @alphapapa's comments:

> I'm guessing you should use defcustom for pynt-epc-port, as well as a few other vars that are currently using defvar.

I've [made `pynt-epc-port` customizable](https://github.com/ebanner/pynt/blob/62777a08ab54294e72622a63407120576a307045/pynt/pynt.el#L39) as well as [`pynt-verbose`](https://github.com/ebanner/pynt/blob/62777a08ab54294e72622a63407120576a307045/pynt/pynt.el#L68). It is not clear to me that any other variables should be customizable with the exception of appending to `pynt-init-code-template` which would be nice but I have other features which are currently higher priority. Please let me know if I have missed any.

> You may need to do (require 'seq).
> I see a call to deferred:$ but it's not required nor declared as a dependency.

I ran a `make sandbox RECIPE=pynt` and everything worked. I assumed that if I needed to declare additional dependencies I would get "package not found errors" at this state. How I am to figure out which dependencies I need to add and which ones I can leave out (FYI I [added deferred as a dependency](https://github.com/ebanner/pynt/blob/62777a08ab54294e72622a63407120576a307045/pynt/pynt.el#L7) anyway)

> It's better to use lexical-binding in the package header than to use lexical-let.

I have [declared lexical bindings at the top](https://github.com/ebanner/pynt/blob/62777a08ab54294e72622a63407120576a307045/pynt/pynt.el#L1) and [replaced the only occurrence of `lexical-let` with a `let`](https://github.com/ebanner/pynt/blob/62777a08ab54294e72622a63407120576a307045/pynt/pynt.el#L238)

Now onto @purcell's comments:

> Don't set the local values of the print-* vars at the top level - this will only affect the buffer in which the file is first loaded, and is generally nonsense. If you really need these settings (and you shouldn't, except very rarely), you should let-bind them locally while running the tiny piece of code that requires those settings.

The inclusion of these lines is an artifact of early days when I was getting emacs freezing issues of printing out EIN data structures which are deeply nested and recursive [[1]](https://github.com/millejoh/emacs-ipython-notebook/issues/256#issuecomment-350771393) [[2]](https://github.com/millejoh/emacs-ipython-notebook/issues/257#issuecomment-351303434) [[3]](https://www.reddit.com/r/emacs/comments/7il6w7/callinteractively_slow_but_mx_fast/dsn5igf/?context=0). These lines were recommended as a fix and they seemed to solve the problem at the time but I'm not even so sure that they are necessary anymore. I will remove them and do some testing to make sure it doesn't break anything.

> Don't perform logic based on buffer names. Buffers can be arbitrarily renamed, and not every buffer is backed by a file. If you're managing the buffer, you can set its associated filename, or note a local variable in that buffer which you can consult later.

I take this recommendation to heart and will make the change. Even when writing this code I was thinking to myself "I know this won't work under such and such situation" but it's a matter of writing something that "works" and then moving on. You are absolutely right and I can see it being a bug that will easily arise.

> buffer-string could likely replace pynt-get-buffer-string.

This one I'm not so sure about. `buffer-string` returns the buffer as a string with tons of font properties. `buffer-substring-no-properties` solves this issue but does not have a "grab the entire buffer" option. Hence the existence of `pynt-get-buffer-string`. I would love to be wrong and would be happy to replace it with a function that already exists if you know of one.

> Add a package dependency on seq, ideally whichever version is in GNU ELPA, and a require form.

My question is the same as is for @alphapapa - how do I know which packages I need to add as a dependency or `require`? I didn't need to do this in my `make sandbox`. Does this mean `make sandbox` should not automatically install these packages? I will add these dependencies and `require`s but to minimize the work on y'all's end it would be great if `make sandbox` threw an error if these packages are not included/`require`d.

> Should this use your pynt-log function instead of message? Probably.

Yes, perhaps. Though everything with `pynt-log` is for debugging (basically stuff I only want to see). Currently it's "necessary" for the user to see this message so they know which iteration they are on in the generated jupyter notebook. So it's not obvious to me that this should be in `pynt-log`.

What I could do to be more explicit is rename `pynt-log` to `pynt-debug` and then make a new `pynt-log` which can be silenced upon configuration (but its default value is to be verbose). Do you recommend this?

> lexical-let came from cl, which my PR removes, but it's obsolete anyway: declare lexical-binding for your file instead. (Try running M-x auto-insert in an empty .el file to see how.)

I [added lexial bindings](https://github.com/ebanner/pynt/commit/7350ab2bbc844261ef5c7d48203f911749ab6666#diff-599a3408ed2397b5ada7d680b7b21f0dR1) but did not run `M-x auto-insert` on an empty `.el` file. I'll make sure to do that.

alphapapa:
You need to declare a dependency on any package that's not included in Emacs, and you need to `require` all of those, as well as any built-in package that isn't loaded by default.

You can use `(buffer-substring-no-properties (point-min) (point-max))` if you must remove the text properties.

purcell:
Looking good!

> This one I'm not so sure about. buffer-string returns the buffer as a string with tons of font properties. buffer-substring-no-properties solves this issue but does not have a "grab the entire buffer" option. Hence the existence of pynt-get-buffer-string.

Generally the properties don't make any difference when the string is taken and used elsewhere.

> What I could do to be more explicit is rename pynt-log to pynt-debug and then make a new pynt-log which can be silenced upon configuration (but its default value is to be verbose). Do you recommend this?

Sure, whatever works.

> I added lexial bindings but did not run M-x auto-insert on an empty .el file. I'll make sure to do that.

Not really necessary: you got the right format anyway. A handy command to know, though.

ebanner:
Hi @purcell,

> Generally the properties don't make any difference when the string is taken and used elsewhere.

For a file with the following contents

```python
print('Hello World!')
```

The output produced by`(message "%s" (buffer-string))` produces

```
#("print('Hello World!')
" 0 5 (fontified t face font-lock-keyword-face) 5 6 (fontified t face (rainbow-delimiters-depth-1-face)) 6 7 (fontified t syntax-table (15) face font-lock-string-face) 7 18 (fontified t face font-lock-string-face) 18 19 (fontified t face font-lock-string-face) 19 20 (fontified t syntax-table (15) face font-lock-string-face) 20 21 (fontified t face (rainbow-delimiters-depth-1-face)) 21 22 (fontified t))
```

Immediately after calling `pynt-get-buffer-string` I pass it over to an external python process over RPC which parses the code. I only want to send over `print("Hello World!")`. Is this a good example not to use `(buffer-string)`? Or am I missing an easy way to strip off the font information?

Thanks again.

purcell:
> The output produced by(message "%s" (buffer-string)) produces

I see that with `%S`, which calls "print" internally, but not with `%s`, which is just string interpolation. Note that the *result* of the `(message ...)` expression is a string which contains properties, but the *output* (in the `*Messages*` buffer) contains no properties. So I think you're safe with `buffer-string` here.

ebanner:
The code sample I pasted was copy/pasted from the `*Messages*` buffer.

I'll try switching out `pynt-get-buffer-string` with `buffer-string` and print it out on the python end. That will be the ultimate test.

purcell:
Just checked this and the issues above seem fixed, so going ahead and merging this now - welcome to MELPA!

PR #5242, opened 2018-01-16
Add playground

akirak:

Run Emacs on a sandboxed $HOME

https://github.com/akirak/emacs-playground

Maintainer

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/play
+++ b/recipes/play
@@ -1,0 +1,1 @@
+(play :fetcher github :repo "akirak/play.el" :files ("play.el"))
--- a/recipes/play
+++ b/recipes/play
@@ -1,1 +1,0 @@
-(play :fetcher github :repo "akirak/play.el" :files ("play.el"))
--- a/recipes/playground
+++ b/recipes/playground
@@ -1,0 +1,1 @@
+(playground :fetcher github :repo "akirak/emacs-playground")


alphapapa:
Hi,

Have you considered naming the package something like `playground` or `sandbox`?  `play` is quite vague, and the first thing that came to mind for me was an audio utility (because there's a `play` utility in SoX).

akirak:
@alphapapa Thank you for your comments. 

I don't mind changing the package name to `playground` or something, but I will have to change the names of all functions and variables (and even update the README), as their namespace is going to change from `play-` to `playground-`. Could you give me a little  time until I send you a new PR? 

Installing using curl is not a necessity. It's just an option, as this packages doesn't require any dependencies. I've understood your point, so I'll delete the instruction after this package is added to MELPA. 

As for the Helm support, it would be better to provide a dedicated Helm function, because there are multiple (local and remote) sources for sandboxes. However, I won't work on it for now, as it will require additional dependency on Helm, which seems an overkill for this trivial stuff.

alphapapa:
> I don't mind changing the package name to playground or something, but I will have to change the names of all functions and variables (and even update the README), as their namespace is going to change from play- to playground-

Yeah, generally renaming an Emacs package is very easy, because doing a simple `M-% RET play- RET playground- RET` will take care of it.  :)

> Could you give me a little time until I send you a new PR?

I'm not MELPA staff, I'm just an observer who tries to help by giving feedback sometimes.  In any case, it's your package, so you can take all the time you want.

> As for the Helm support, it would be better to provide a dedicated Helm function, because there are multiple (local and remote) sources for sandboxes. However, I won't work on it for now, as it will require additional dependency on Helm, which seems an overkill for this trivial stuff.

I see, good point.  Helm's multi-source support is indeed very useful.

purcell:
Thanks! Quick notes:
- Will [this](https://github.com/akirak/emacs-playground/blob/e4dd48574601ea9ab8e706a736ad248ca2e37e32/playground.el#L42) work on Windows? Isn't there a built-in function/variable you could use to determine the home directory?
- [`(when (null repo)`](https://github.com/akirak/emacs-playground/blob/e4dd48574601ea9ab8e706a736ad248ca2e37e32/playground.el#L236) is equivalent to `(unless repo`.
- [This line](https://github.com/akirak/emacs-playground/blob/e4dd48574601ea9ab8e706a736ad248ca2e37e32/playground.el#L317) will break horribly if the home directory has a space in it. Use `shell-quote-argument` whenever you build commands as strings.

Hope that helps!

akirak:
@purcell Thank you for helpful comments. 

> Will this work on Windows? Isn't there a built-in function/variable you could use to determine the home directory?

Windows was not in mind at the time of writing this package, because I don't have a (real) Windows machine. However, your idea can help some people who would wish to use playground no Windows. 

I can test Playground on [Windows VM images](https://developer.microsoft.com/en-us/microsoft-edge/tools/vms/) and/or my 15-year-old (!) Windows XP license. It will take some time, but I don't mind working on it, given that it is a one-shot task. The problem is that there seem to be multiple variants of [Emacs for Windows](https://www.emacswiki.org/emacs/EmacsForWindows), and I don't know how they work. Could you elaborate me on how to get the original home directory (looked up by Emacs) on Windows? 

> This line will break horribly if the home directory has a space in it. Use shell-quote-argument whenever you build commands as strings.

I have never imagined a situation where a login name contains a space. I will fix this, but I will have to set up a new test, which can take some time.

purcell:
Sent you https://github.com/akirak/emacs-playground/pull/5.

purcell:
Merged now - welcome to MELPA!

PR #5248, opened 2018-01-19
Add recipes/anki-editor

louietan:

A utility that helps you create Anki cards in Org-mode.

https://github.com/louietan/anki-editor

Author and maintainer

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/anki-editor
+++ b/recipes/anki-editor
@@ -1,0 +1,1 @@
+(anki-editor :fetcher github :repo "louietan/anki-editor")


alphapapa:
Hi,

A little feedback:

* You should define a group with `defgroup`, and then you can omit `:group 'anki-editor` on the `defcustom`s.

louietan:
Thanks for the detailed feedback, I have fixed some of these issues, please review the changes and let me know if they are OK.

1. [X] Define a group with `defgroup`. **Fixed**.
2. [ ] Consider using `a-list` to deal with alists.
3. [X] Use `push` instead of `add-to-list`. **Fixed**.
4. [X] Quote shell arguments. **Fixed**.
5. [X] Use `let-alist` to bind cdrs of alist. **Fixed**.
6. [X] `(require 'seq)`. **Replaced the sole use of `seq-each` with `dolist`**.
7. [X] Use `defvar` instead of `setq`. **Fixed**.
8. [x] Define a minor mode. **Fixed**.
9. [x] Think about `anki-editor-anki-connect-upgrade`. **I added a
       confirmation step that tells the user what this command is
       for and what it is going to do and asks if they really want to
       continue**.

alphapapa:
Hi,

Thanks, looks good.  Couple of items I noticed:

* AFAIK `C-c a` is the recommended prefix for Org agenda commands, and those are used across buffers, so you probably shouldn't use it for your package.

* > Think about anki-editor-anki-connect-upgrade. **I added a confirmation step that tells the user what this command is for and what it is going to do and asks if they really want to continue.**

Looks good to me.  I'm guessing Steve will approve of the confirmation, but it's his call.

purcell:
> C-c a is the recommended prefix for Org agenda commands, and those are used across buffers, so you probably shouldn't use it for your package.

It's also not a legal keybinding for packages to provide - see https://www.gnu.org/software/emacs/manual/html_node/elisp/Key-Binding-Conventions.html - so it must not be used. `package-lint` should have warned about this.

Everything looks good otherwise - nice work! (And thanks for the great feedback @alphapapa!)

louietan:
Hi, Steve! I can't come up with any better key sequences right now, so I have removed the default key bindings as well as the minor mode, and give users the freedom to define whatever key bindings they like.

purcell:
Good plan. Okay, merging now - welcome to MELPA!

louietan:
So excited to hear about this! Big thanks to all of you!

PR #5249, opened 2018-01-20
Add new recipe md4rd (mode for reddit)

ahungry:

Enables browsing reddit.com via Emacs.

https://github.com/ahungry/redditor-mode

Maintainer and sole author.

None needed

### Checklist

Please confirm with `x`:

There is one caveat to package-lint, I am choosing to use a non-standard separator at the moment, opting for a Common Lisp style (`package-abbrev:private-call` and `package-abbrev::public-call`) over package-lint's favored Elisp style (`package--private` and `package-public`), as I think it makes the code much more readable.

If it's a big deal, I could update for the interactive functions and customs (although really, the only interactive one is `M-x redditor-mode`).

### DIFF
--- a/recipes/md4rd
+++ b/recipes/md4rd
@@ -1,0 +1,4 @@
+(md4rd
+ :repo "ahungry/md4rd"
+ :files ("md4rd.el")
+ :fetcher github)
--- a/recipes/md4rd
+++ b/recipes/md4rd
@@ -1,4 +1,3 @@
 (md4rd
  :repo "ahungry/md4rd"
- :files ("md4rd.el")
  :fetcher github)


alphapapa:
> There is one caveat to package-lint, I am choosing to use a non-standard separator at the moment, opting for a Common Lisp style (package-abbrev:private-call and package-abbrev::public-call) over package-lint's favored Elisp style (package--private and package-public), as I think it makes the code much more readable.

Personally I also prefer more distinct package name separators, but MELPA is strict about this requirement, so you'll need to change them.  That's why the rule was added to `package-lint`, to save us from having to say this in every recipe PR.  :P

ahungry:
thanks so much for the feedback. changes coming tonight.

ahungry:
All changes 1 to 12 have been made except for refactoring the small hash table use to use ht lib (maybe I will incorporate later).

I renamed it from redditor-mode to just redditor for brevity as well, since a redditor-mode will likely be a small part of what is a larger 'redditor' package.

Let me know what you think now!

ahungry:
All changes 1 to 12 have been made except for refactoring the small hash table use to use ht lib (maybe I will incorporate later).

I renamed it from redditor-mode to just redditor for brevity as well, since a redditor-mode will likely be a small part of what is a larger 'redditor' package.

Let me know what you think now!

alphapapa:
Hi again,

Thanks, looks good.  Few other things:

* Why not use `defun` here instead of `defvar` with a lambda?
```el
(defvar redditor--parentfn
  (lambda (name)
    (unless (equal 'thread name)
      (let ((parent-id
             (alist-get
              'parent_id
              (cl-find-if
               (lambda (comment)
                 (equal name (alist-get 'name comment)))
               redditor--comments-composite))))
        (if parent-id parent-id 'thread)))))
```

* You're using `redditor--comment-url` as a global variable.  It's largely a matter of style and preference, so I'm not telling you how to write your code, but you might consider some refactoring to pass it as an argument instead.  That would likely make it easier to do things in the future like loading multiple subs or comments in parallel.  (Of course, that's not relevant to having the package in MELPA.)

* You should probably tidy up the keywords a bit more:

```el
;; Keywords: ahungry reddit browse news
```

Your GitHub username isn't really an Emacs package keyword.  ;)  `browse` is pretty vague as well.  `reddit` and `news` are pretty good.  See the command `finder-by-keyword` for a list of the official package keywords.

ahungry:
I have a few packages on melpa/Gnu elpa and tag with ahungry so people can find packages I have written or maintain and have a certain expectation of quality - I think its a worthwhile tag.

The comment url global is fine as it only loads that page as a button press event.

The lambda as a var was a more direct refactor to the inline lambda in the hierarchy samples.

Ill squash again tonight, sure.

purcell:
Thanks! Looks good, and I'll merge this now, but there's no mode here ("mode" means something very specific), so you should remove `"mode"` [here](https://github.com/ahungry/md4rd/blob/050f2fb66478e4d5465fe872f3cb30e411f429ca/md4rd.el#L332) and remove the [md4rd-mode](https://github.com/ahungry/md4rd/blob/050f2fb66478e4d5465fe872f3cb30e411f429ca/md4rd.el#L338) function.

alphapapa:
@ahungry Curious, why the name change?  `redditor` was descriptive and unique.  `md4rd` looks like something for Markdown files.

ahungry:
@purcell I am going to add mode specific binds, working on oauth integration for upvotes/downvotes now, then the mode specific binds to invoke them to follow

ahungry:
@alphapapa I agree, but according to the reddit API usage guide (which I had to read after getting to the OAuth authenticated stuff), you can't have 'reddit' (in any derivative) without a 'for' preceding it, so 'mode for reddit' works (but is way too long to include as a prefix for every function/variable), hence md4rd, which can read as mode-for-red or something I guess...

https://www.reddit.com/wiki/api

Since it will have the keyword 'reddit' (and the only such package on MELPA that does) I think users will still find it in due time quite easily

alphapapa:
> You may use "for reddit" or "a client for reddit" in the title of your app. You may not use "reddit" without "for" preceding it. You may not use the word "official" in the title, keywords or description in any way that implies the app was developed by Reddit, Inc.

Haha, I think I would call it `unofficial-for-reddit` then.  ;)

PR #5250, opened 2018-01-20
Add elcontext

rollacaster:

Create context specific actions

https://github.com/rollacaster/elcontext

Maintainer and sole author.

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/elcontext
+++ b/recipes/elcontext
@@ -1,0 +1,2 @@
+(elcontext :repo "rollacaster/elcontext"
+           :fetcher github)


alphapapa:
I recommend changing the package name to something more descriptive, like `geofence` or `geofence-actions`.  As it is, "context" is quite ambiguous.  When I saw the description, "Create context specific actions," I thought it meant something like directory-specific actions, or actions depending on where point is in a buffer.

rollacaster:
Hi alphapapa!

Thank you for your feedback. I see your point about context being ambiguous but I do not want to restrict the package just to the geofence feature. E.g. currently time is another context that can be configured.

Your first thoughts for `elcontext` are great ideas for new features. I'll try to add a new context option for folders (which can be combined with geofences and timespans).

Oh and thanks a lot for maintaining `melpa` adding my first package here was really convenient thanks to the docs and `package-lint`

alphapapa:
Hi Thomas,

> Thank you for your feedback. I see your point about context being ambiguous but I do not want to restrict the package just to the geofence feature. E.g. currently time is another context that can be configured.

> Your first thoughts for elcontext are great ideas for new features. I'll try to add a new context option for folders (which can be combined with geofences and timespans).

Ah, I see!  So it's sort of like an Emacs version of the Llama app for Android.  Or an Emacs version of Cron.  Very interesting.

> Oh and thanks a lot for maintaining melpa adding my first package here was really convenient thanks to the docs and package-lint

I'm not a MELPA maintainer, just an occasional contributor who tries to help reduce the workload of Steve and the other MELPA maintainers by reviewing packages now and then.  They deserve all the credit, and especially Steve, for all the time he puts into MELPA!

purcell:
Thanks! This reminds me of the (very handy) ControlPlane app for OS X. And note that my package `osx-location` might save any special handling of `CoreLocationCLI`. Further feedback:
- Don't define partial faces, e.g. just foregrounds etc. Instead, either define `light` and `dark` variants or (much better) just `:inherit` built-in faces by default, e.g. `comment` or `success` etc.
- I think you just want `load-file` [here](https://github.com/rollacaster/elcontext/blob/a6c8e061263b50f9130b6a612213677d790c2504/elcontext.el#L191).
- `(when (not (null x) ...)` is just `(when x` [here](https://github.com/rollacaster/elcontext/blob/a6c8e061263b50f9130b6a612213677d790c2504/elcontext-time.el#L168).

Hope that helps!

rollacaster:
> I see from the readme that it requires Mac OS. Do you plan to support other OSes?

AFAIK there's no standard way for other OSes to get GPS coordinates. I think I will deactivate the geofence feature for any other OS so that at least the other features can be used.

> And note that my package osx-location might save any special handling of CoreLocationCLI

Thanks this will make my code simpler!

> I think you just want `load-file`

`load-file` does not return the loaded file but only `t`. Or did I misunderstood sth?


Thanks in general for all your feedback! I'm almost done putting it all in.

purcell:
> load-file does not return the loaded file but only t. Or did I misunderstood sth?

Hah, good point. It's quite common for libraries to save their state by writing an expression which - when evaluated - will restore the state. The Emacs custom system, and `desktop-save` are examples of this. And in that case, a simple `load-file` is all that's required. I guess either way works, but maybe consider the alternative - there's probably a reason why those libraries favour that approach. :-/

alphapapa:
`recentf` is another example: although it uses `format` and `insert` rather than `prin1` to save data (it originally dates from 1999, so maybe there are historical reasons), it also uses `load-file`.

rollacaster:
I found `saveplace` and `smex` which solve it similar to me. I am more in favour of keeping it because having the side-effect (of setting the data which was read) within the code seems cleaner to me than having it in the data I store. But it's a small implementation detail and I am fine changing it, too.

purcell:
Merged now, thanks for your patience. Welcome to MELPA!

PR #5251, opened 2018-01-20
Add s3ed

mattusifer:

S3ed provides access to files stored on Amazon S3 using normal dired and filesystem commands.

https://github.com/mattusifer/s3ed

Maintainer

N/A

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/s3ed
+++ b/recipes/s3ed
@@ -1,0 +1,1 @@
+(s3ed :fetcher github :repo "mattusifer/s3ed")
--- a/recipes/tramps3
+++ b/recipes/tramps3
@@ -1,1 +1,0 @@
-(tramps3 :fetcher github :repo "mattusifer/tramps3")
--- a/recipes/tramps3
+++ b/recipes/tramps3
@@ -1,0 +1,1 @@
+(tramps3 :fetcher github :repo "mattusifer/tramps3")


purcell:
Thanks! This could be very useful. Quick feedback:
- Your code requires Emacs 24.4, but you don't have a corresponding `Package-Requires` header to declare that dependency. (The explicit check is probably overkill, btw.)
- You also use `dash` and `seq` elsewhere but don't declare dependencies on either.
- `is-tramps3-mode-active` is wrongly-named: it should be `tramps3-mode-is-active` or similar, to keep a consistent prefix. But in any case, that function should be redundant: minor modes contain a correspondingly-named variable which you can simply inspect, e.g. `tramps3-mode`.
- Did you want/need your minor mode to be a global one instead of local?
- Tramp supports pluggable methods, so I wonder whether all of this couldn't simply be a regular tramp extension? That would be much nicer IMO, and avoid the tricky dired special handling.

mattusifer:
Thanks for the feedback! I've addressed the first 4 bullet points. I hadn't considered making it a global minor mode, but I tested that out and I actually prefer it this way - thanks for pointing out that option. Making it global helps ensure that the local data doesn't get out of sync with s3. 

As for making this a regular tramp extension - that was my original intent with this library, and I agree it would be much nicer. I haven't found any documentation on how to write a custom TRAMP method, and the examples I've found (tramp-hdfs, tramp-sh.el, tramp-smb.el, tramp-adb.el) don't have to deal with some of the irregularities of s3 (no concept of a remote shell, no hostname/username/password, etc). To summarize - it is still something I'm actively trying to figuring out, but it looks like it's going to take me a while.

Let me know if there is any other feedback.

purcell:
Nice! I guess my only remaining feedback, then, is that people are going to find this expecting it to work with tramp, and instead it'll do odd things with dired. Would it be worth calling it something like `s3dired` or `s3ed` for now?

mattusifer:
That's a fair point, I like `s3ed`. Thanks for the suggestion. I've renamed everything in the repo and the repo itself, and I updated the recipe here.

purcell:
Great - merged. Thanks, and welcome to MELPA!

PR #5252, opened 2018-01-21
Add recipe for 'extmap' package.

doublep:
This will be used by the next version of `datetime` package to store timezone and locale databases.

`extmap` is a very simple package that lets you build a *read-only*,
*constant* database that maps Elisp symbols to almost arbitrary Elisp
objects.  The idea is to avoid preloading all data to memory and only
retrieve it when needed.

This package doesn’t use any external programs, making it a suitable
dependency for smaller libraries.

https://github.com/doublep/extmap

I'm the only author and maintainer.

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/datetime
+++ b/recipes/datetime
@@ -1,2 +1,3 @@
 (datetime :repo "doublep/datetime"
-          :fetcher github)
+          :fetcher github
+          :files (:defaults "*.extmap"))
--- a/recipes/extmap
+++ b/recipes/extmap
@@ -1,0 +1,2 @@
+(extmap :repo "doublep/extmap"
+        :fetcher github)


doublep:
I included the patch for 'datetime' recipe here, to avoid spamming with pull requests.

purcell:
> I included the patch for 'datetime' recipe here, to avoid spamming with pull requests.

Separate requests are better, tbh, particularly because issues with one package might hold up merging of a recipe for another.

Anyhow, thanks for your patience. Some quick feedback:

### datetime

- Please use `:defaults` in place of `"*.el"` in the recipe's `:files` spec.
- You might like to autoload some of the common entry points.
- Your useful commentary text is incorrectly placed inside the `Code:` section, where tools won't see it. Also, consider starting that section by briefly introducing provided functions that users will immediately recognise as useful. Diving into the formats first is distracting IMO.

### extmap

Looks good. Same feedback re. autoloads and usage examples in the commentary.

doublep:
> Separate requests are better, tbh, particularly because issues with one package might hold up merging of a recipe for another.

I just thought that you are busy because of the long response time. Besides, `datetime` is already in Melpa, I just need to update its recipe and the update is directly related to `extmap`.

> datetime
> Please use :defaults in place of "*.el" in the recipe's :files spec.

Done.  I just squashed and force-pushed a new commit to keep the history shorter.

> You might like to autoload some of the common entry points.

I think that autoloads should be reserved to user-visible things and core functionality.  A programmer should rather explicitly `(require ...)` the libraries he needs.

> Your useful commentary text is incorrectly placed inside the Code: section, where tools won't see it.

It's more of a comment about internals. Up until now `datetime` has not been generally useful anyway, just a utility for `logview`. I will expand the placeholder comment when I add more functionality (I already have it locally, but haven't pushed, since it depends on `extmap` which is not yet available from Melpa).

> Also, consider starting that section by briefly introducing provided functions that users will immediately recognise as useful. Diving into the formats first is distracting IMO.

Yes, will do that in the next version, once recipe is updated so that I can push `*.extmap` files.

> extmap
> Looks good. Same feedback re. autoloads and usage examples in the commentary.

I improved package commentary in commit 235a983.  Didn't add any autoloads, because again, I feel developers should just `(require 'extmap)`.

purcell:
Cool, that all makes sense - thanks! Merging now.

PR #5255, opened 2018-01-21
Add recipe for elgrep.

TobiasZawada:

Elisp grep of files.
Easy usage through user menu generated by command `elgrep-menu`.
Can recurse into sub-directories.
Search function can be customized.
`*elgrep*` buffer can be edited after calling `elgrep-edit`

https://github.com/TobiasZawada/elgrep

Author and maintainer

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/elgrep
+++ b/recipes/elgrep
@@ -1,0 +1,1 @@
+(elgrep :repo "TobiasZawada/elgrep" :fetcher github)


purcell:
Thanks. `package-lint` must have given you a bunch of feedback on this, so I'll let you run it again before I comment here. :-)

TobiasZawada:
I did run `elint-current-buffer` on `elgrep.el` and it run cleanly. But after your comment I guess that is not sufficient? Good. I'll clone from my own repository and try to run `package-lint` as you suggest.

purcell:
Yep, `elint` != `package-lint`, though I'm sure `elint` is also helpful. :-)

TobiasZawada:
Thanks for the comments.
package-lint, checkdoc, and make recipes/elgrep run cleanly now. Everything committed in elgrep/master.

TobiasZawada:
Do you also have something like `suggests` for the soft package dependencies?

TobiasZawada:
FYI: The motivation to put elgrep finally on melpa: https://emacs.stackexchange.com/questions/38240/removing-a-latex-command-by-matching-braces/38244#38244

purcell:
Thanks. Some quick notes:
- Filename patterns should end with `"\\'"` (end of string) rather than `"$"` (end of line). You have many occurrences of this.
- All [those `w-` variables](https://github.com/TobiasZawada/elgrep/blob/fbfa8dc7a2b790e4d5fe3c276f5e59eb51f1a1d8/elgrep.el#L115-L125) need to be prefixed with the same `elgrep-` package prefix.
- `cBeg` and `cEnd` shouldn't be camel-case in lisp.
- The dependency on Emacs 25.1.50 should probably be on 25.1. It's *very* unusual to depend on a specific patch release, and therefore unlikely to be correct.
- There's a lot of `caar` and `cadar` etc. You might consider using alists or plists instead, which often makes code easier to understand.
- [Here](https://github.com/TobiasZawada/elgrep/blob/fbfa8dc7a2b790e4d5fe3c276f5e59eb51f1a1d8/elgrep.el#L486) you might prefer to use `[remap save-buffer]` rather than an explicit `(kbd "C-x C-s")`, so that it still works if the user has changed their default bindings.
- All the [keymap wrangling](https://github.com/TobiasZawada/elgrep/blob/fbfa8dc7a2b790e4d5fe3c276f5e59eb51f1a1d8/elgrep.el#L485) for `elgrep-edit-mode` and [checking for `compilation-mode`](https://github.com/TobiasZawada/elgrep/blob/fbfa8dc7a2b790e4d5fe3c276f5e59eb51f1a1d8/elgrep.el#L505) etc. is very unusual, and makes me wonder if you're doing an easy thing the hard way.

Hope that helps!

TobiasZawada:
@purcell 
Thanks for your valuable comments. I've added a checkmark to each of your comments and added my comments as sublists where appropriate.

-   [X] Filename patterns should end with "\\\\'" (end of string) rather than "$" (end of line). You have many occurrences of this.
    -   You are so terrifying right! I've just create a file with name "crazy\n" under linux and it works!
-   [X] All those w- variables need to be prefixed with the same elgrep- package prefix.
    -   Done even if the \`defvar' forms just silence the byte compiler. The variables are only set buffer-locally in the elgrep-menu buffer.
-   [X] cBeg and cEnd shouldn't be camel-case in lisp.
-   [X] The dependency on Emacs 25.1.50 should probably be on 25.1. It's very unusual to depend on a specific patch release, and therefore unlikely to be correct.
-   [ ] There's a lot of caar and cadar etc. You might consider using alists or plists instead, which often makes code easier to understand.
    -   The need for `caar` and `cadar` mainly comes from the use of `compilation--loc->file-struct` which is part of "compile.el".
    -   The structure of the data returned by elgrep is described in the doc string of `elgrep`.
        It already exploits a property list where it makes sense.
    -   The only other point where \`caar' was used was an iteration **on** a list via \`cl-loop'.
        I have avoided that one usage by optimizing the loop. (I've tested that change. It works as expected.)
-   [X] Here you might prefer to use [remap save-buffer] rather than an explicit (kbd "C-x C-s"), so that it still works if the user has changed their default bindings.
-   [ ] All the keymap wrangling for elgrep-edit-mode and checking for compilation-mode etc. is very unusual, and makes me wonder if you're doing an easy thing the hard way.
    Thank you for that advice. I'll keep it in mind and I'll change that part if I find a better solution.

TobiasZawada:
@purcell FYI:
1. All issues except the `caar`/`cadar` issue are resolved. The `caar`/`cadar` issue is implied by the external library `compilation.el`.
2. All checks (package-lint, byte-compile, checkdoc, make recipes/elgrep have been completed successfully.

Best regards,
Tobias Zawada

DamienCassou:
I'm in favor of merging that one.

purcell:
Great, added - thanks!

TobiasZawada:
Thanks for the great revision work!

PR #5258, opened 2018-01-23
Add parseclj recipe

volrath:

Clojure / EDN parser

https://github.com/clojure-emacs/parseclj

Contributor

Besides discussion in #5190, **None Needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/parseclj
+++ b/recipes/parseclj
@@ -1,0 +1,1 @@
+(parseclj :repo "clojure-emacs/parseclj" :fetcher github)


purcell:
Thanks!
- You might like to autoload `parseclj-parse-clojure`, and possibly other symbols.
- The repo contains `parseedn.el`: this should be `parseclj-edn.el` in order to fit with the rest of the package.

Looks good otherwise!

volrath:
Hey @purcell,  thanks for your review!

Regarding autoload: `parseclj` is intended to be used as an aux library in other packages, none of its parse functions provide any useful output on their own, they produce an AST as a list that needs to be processed by any other packages using it.  For this reason I think it's better to have people explicitly require `parseclj` when they need to use one of the parsers in their code.  I didn't find any elisp style guide / conventions guideline that talks about good practices regarding autoload, that's why I am assuming the idea of autoloading is handy when you want to provide a function that does something on its own (like a command).  Please let me know if I'm missing something and I'll be happy to correct it.

Regarding `parseedn.el`, we decided to use this name because it has two big differences from `parseclj`: 1. It parses EDN, not Clojure, and 2. It doesn't return ASTs like `parseclj`, but an elisp representation of the parsed EDN, similar to [edn.el](https://github.com/expez/edn.el) -- So it kind of feels like a different package rather than a `parseclj` "sub package".

Having said that, I understand this is not a common practice for Emacs packages, and I wouldn't mind renaming it.  I would like to get @plexus' input on this before proceeding.

purcell:
> For this reason I think it's better to have people explicitly require parseclj when they need to use one of the parsers in their code.

Yep, that's absolutely fine, then.



> So it (`parseedn`) kind of feels like a different package rather than a parseclj "sub package".

Yep, I'd agree with that. Either extracting it as a separate package or renaming to `parseclj-edn` would be fine with me.

plexus:
`parseedn` is intended to be a separate package, hence the name.

volrath:
@plexus should I remove it from `parseclj` and create another repo in `clojure-emacs` or would you like to do that yourself?

purcell:
Please remember to ping us here when done so we can merge. :-)

volrath:
@purcell will do, I just have to find some time, I've been swamped lately. Hopefully over this week I'll add a recipe for `parseedn` as a separate package and ping you :)

volrath:
@purcell `parseclj` and `parseedn` are finally split into two packages.

I already double checked all the contribution checklist and everything's ok.  This package should be good to go.

After `parseclj` lands on MELPA, I'll add a `parseedn` recipe (https://github.com/clojure-emacs/parseedn)

DamienCassou:
We could probably merge this one.

purcell:
Super, thanks all.

plexus:
Thanks a lot @volrath, @purcell, @DamienCassou. I'm very excited to have this available to a wider audience!

volrath:
Thanks @purcell and everyone for taking the time :)

PR #5263, opened 2018-01-26
Add posframe

tumashu:

Pop a  child-frame at point

https://github.com/tumashu/posframe

maintainer

**None needed** 

### Checklist

Please confirm with `x`:

- [-] I've built and installed the package using the instructions in [CONTRIBUTING.md](https://github.com/melpa/melpa/blob/master/CONTRIBUTING.md)

### DIFF
--- a/recipes/posframe
+++ b/recipes/posframe
@@ -1,0 +1,1 @@
+(posframe :fetcher github :repo "tumashu/posframe")


tumashu:
@purcell  Is it merge the PR quicker, for two packages of mine are waiting it into melpa :-)

purcell:
Looks good, but I don't really understand the comment about compatibility. If this requires Emacs 26, then why not make that a package dependency?

tumashu:
use pyim as example,  pyim can work well with emacs24, if we let posframe require emacs26, pyim will not installable in emacs24,  is we do not let pyim hard depend posframe, i had let user manual install posframe

purcell:
Following that logic, everyone would list optional dependencies as mandatory (for convenient installation), and every depended-upon package would do its compatibility checks at runtime. That would be terrible for the package ecosystem.

The answer here is surely to make `posframe` depend on the Emacs version it needs, and to *not* add a package dependency on `posframe` to `pyim`. You can still add code in `pyim` which allows it to support `posframe` if available/configured.

tumashu:
Maybe you are right, fixed.

purcell:
Super, thank you! Merged.

purcell:
I appreciate your support! :tada:

PR #5266, opened 2018-01-26
Add new recipe call-graph

beacoder:
Add new recipe call-graph

Library to generate call graph for cpp functions.

https://github.com/beacoder/call-graph

maintainer

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/call-graph
+++ b/recipes/call-graph
@@ -1,0 +1,1 @@
+(call-graph :repo "beacoder/call-graph" :fetcher github)


beacoder:
@purcell ,
hi, sorry to bother you, but is there anything wrong with my commit, why no-body gives any comments on my commit ?

purcell:
Sorry, in the last couple of months it has taken me up to two weeks to review new packages. I’ll hopefully get time to catch up again and review this one at the weekend.

TobiasZawada:
@purcell Please do not let you push! You are doing a very good job with reviewing the melpa packages and it is clear that melpa is not your number one priority -- family as well as your actual job come first -- and code review takes time! Everybody in the queue is politely waiting... But, it is nice to hear that you are still alive;-)).

beacoder:
@TobiasZawada jesus christ, how did this "still alive" come from. I hope everything is ok with your guys.
@purcell , I'm just curious about why no comments on my commit, no hard feelings, and please do take your time, I'm not in a hurry.
thanks for you guys for doing such a fantastic job for bringing this powerful and convenient tool melph to make our life easier.

purcell:
LOL. :-)

Nice job with this library, @beacoder.

- [Here](https://github.com/beacoder/call-graph/blob/a7942124005cb7fd268593b27bea846291572ef1/call-graph.el#L182) you should use `shell-quote-argument` to ensure that the `%s` will always be safe. Better safe than sorry.
- You might like to look at the built-in library `queue.el`. :-)
- The [code to disable `linum-mode` and `nlinum-mode`](https://github.com/beacoder/call-graph/blob/a7942124005cb7fd268593b27bea846291572ef1/call-graph.el#L373-L376) is weirdly specific, and doesn't belong in this file: put it in your own config instead. (What if every major mode tried to predict what would look good to every user?)
- Setting `mode-name` is redundant: `define-derived-mode` generates code to do that for you.
- Setting `list-buffers-directory` also looks out of place.
- You should use `kebab-case` not `camelCase` throughout, e.g. [here](https://github.com/beacoder/call-graph/blob/a7942124005cb7fd268593b27bea846291572ef1/call-graph.el#L317-L322).
- The [window-configuration](https://github.com/beacoder/call-graph/blob/a7942124005cb7fd268593b27bea846291572ef1/call-graph.el#L287) hack shouldn't be rigidly enforced here: you can use libraries like `fullframe` in your config to add this behaviour if desired. Just do one thing well!

Hope that helps!

beacoder:
@purcell, thank you so so so ... much for these great suggestions, I will go through these one by one carefully.

beacoder:
@purcell , thanks again for your valuable advices, and I have updated code to fix all the issues you mentioned in the previous comment, except this one:
Here you should use shell-quote-argument to ensure that the %s will always be safe. Better safe than sorry.
Since I'm not quite familiar with the usage of shell-quote-argument, could you please be more specific as how should I change the code to use the shell-quote-argument ?
Thanks very much.

TobiasZawada:
I am sure Steve just meant:

    (format "global -a --result=grep -r %s | grep -E \"\\.(cpp|cc):\"" (shell-quote-argument function))

Note that you should allow for the customization of the tagging system and the `grep` program. Or better just search the result of the tag search by elisp. (Contrary to `diff` it is very easy to replace `grep` by elisp.)

beacoder:
@TobiasZawada ,thanks for answering my question, also I will take your advice into consideration in the future.
@purcell , I have fixed all the issues you mentioned before, could you please help to check again, thank a lot.

purcell:
Thanks for your patience - looks good, so merging now. Welcome to MELPA!

beacoder:
@purcell thank you very much.

PR #5267, opened 2018-01-27
Add prog-fill recipe

ahungry:

Breaks long line calls (chained methods or huge arg lists) into multi-line somewhat intelligently.

https://github.com/ahungry/prog-fill

Owner

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/prog-fill
+++ b/recipes/prog-fill
@@ -1,0 +1,3 @@
+(prog-fill
+ :repo "ahungry/prog-fill"
+ :fetcher github)


purcell:
Thanks, this looks good! Just one note: you must not [auto-enable](https://github.com/ahungry/prog-fill/blob/7e20b9636ec2247e54318dbac570062fb65c34c3/prog-fill.el#L140-L143) this package upon installation. That has to be the user's choice: like all the built-in Emacs packages, it should be possible for a user to have this installed but not activated.

ahungry:
Sure, Ill update with a fix to the autoload snippet

ahungry:
@purcell updated to remove the autoload key binding

mattiasb:
This looks great!

One thing I (as a potential future user) would want is for the settings variables to use `defcustom` and `defgroup`. I find it much easier to customize a mode inside `M-x customize` than having to reach for my `init.el`. :)

purcell:
Great - merged, thanks!

ahungry:
@mattiasb thanks!  Updated to use defcustom over defvar

mattiasb:
Yay! :)

PR #5008, opened 2017-09-15
Remove all Emacswiki packages

tarsius:
### Update 2018-01-24 - The Finale

Removal of Emacswiki packages has long been requested and planned, yet we've all hung on because the convenience of useful code that's insecurely managed and distributed has consistently exceeded the motivation to mitigate the risks: prominent Emacswiki authors have not taken action, while library and starter kit authors and end users (myself included) have cheerfully continued to depend on Emacswiki-sourced packages.

MELPA stopped building/updating all these packages a while ago, with the intention of removing them once we had a plan to minimise disruption. In the meantime, we continued to serve up our last-built versions. We'd have preferred to publicly sunset our continued hosting of them, but having accidentally deleted those historic (probably-outdated) packages yesterday, we now won't be restoring them. We apologise for the disruption that the sudden removal has caused (and will continue to cause), but this would likely have been the case even if we'd announced a sunset date weeks in advance, and we hope that you'll understand our decision.

In hindsight, MELPA should probably never have distributed Emacswiki-sourced packages. When MELPA was starting off, incorporating those packages helped MELPA's growth, to both the community's benefit and its cost. Thanks to everyone who pushed us to resolve this!

Now that we all know better, let's get on with fixing things and making the package ecosystem better and safer. If Emacswiki authors move their code to a MELPA-supported SCM, we can re-add those packages to MELPA. Some authors won't be willing to do that, so the community will have to decide how much it values that code, and how to get it distributed.

Whatever steps are necessary along the way, we - the MELPA maintainers - will be happy to advise.

-@purcell

---

### Update 2018-01-24

All packages were accidentally deleted, including non-Emacswiki packages. Those are in the process of being restored, but for the Emacswiki packages that is not currently possible. That probably means that we make this final soon, but still waiting for Steve to weight in. The discussion about this begins [further down](https://github.com/melpa/melpa/pull/5008#issuecomment-360098939).

---
----

### Update 2017-09-16

**Status** Packages from the Emacswiki are still available from Melpa, but they are no longer being updates, because https://github.com/melpa/package-build/pull/9 has been merged into Melpa.

**TODO** Here is a list of ongoing efforts to get packages of the Emacswiki:

- ~~#5034 Getting Drew to commit to (a) Git repository/-ies~~ **Unlikely, we've tried for a decade.**
- [ ] #5020 Moving rubikitch's packages to github **No response so far. Sent an email.**
- [x] #2342 Deprecate all emacswiki packages. **Mostly replaced by this issue, except:**
[...]

PR #5268, opened 2018-01-27
Point deleted recipe files from Emacswiki to ahungry rehost

ahungry:
I think you're going to give a lot of users a lot of grief by deleting almost 200 packages in a single PR.

Since MELPA has poor "within Emacs" communication mechanisms (maybe it should do something like magit did for their fundraiser for relaying important news), even someone like myself who is somewhat involved in the elisp content creation community/emacs develop/MELPA had no idea this was coming.

Since I use a few of these deleted packages, I'd like to see them *not* disappear from package-list. :)

This should allow them to be managed in a safe manner, since they're essentially all abandoned (and you've been trying to reach the authors for half a year+ I see).

### DIFF
--- a/recipes/ac-dabbrev
+++ b/recipes/ac-dabbrev
@@ -1,0 +1,4 @@
+(ac-dabbrev
+ :fetcher github
+ :repo "ahungry/emacswiki-mirror"
+ :files ("ac-dabbrev.el"))
--- a/recipes/anchored-transpose
+++ b/recipes/anchored-transpose
@@ -1,0 +1,4 @@
+(anchored-transpose
+ :fetcher github
+ :repo "ahungry/emacswiki-mirror"
+ :files ("anchored-transpose.el"))
--- a/recipes/aok
+++ b/recipes/aok
@@ -1,0 +1,4 @@
+(aok
+ :fetcher github
+ :repo "ahungry/emacswiki-mirror"
+ :files ("aok.el"))
--- a/recipes/apropos-fn+var
+++ b/recipes/apropos-fn+var
@@ -1,0 +1,4 @@
+(apropos-fn+var
+ :fetcher github
+ :repo "ahungry/emacswiki-mirror"
+ :files ("apropos-fn+var.el"))
--- a/recipes/apu
+++ b/recipes/apu
@@ -1,0 +1,4 @@
+(apu
+ :fetcher github
+ :repo "ahungry/emacswiki-mirror"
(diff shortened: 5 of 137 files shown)


purcell:
If we do something like this to soften things following the sudden events this week (see #5008), then who's going to push the community to pick up the libraries for active maintenance? Would you review and incorporate updates? What if updates submitted to you diverge from those in the Emacswiki? And for how long would you do this?

I feel like this ship has sailed, and I'm honestly not keen to simply work around this issue en masse like this, sorry. If you just use a few of these packages, why not pick only those ones up, give them individual repos, and plan to be their maintainer? I suspect the vast majority of the packages will be missed by nobody who also uses MELPA.

tarsius:
Alternatively you could create a separate Elpa archive using the Melpa code but different recipes. If you do that then you should consider getting the packages from the official repository, https://github.com/emacsmirror/emacswiki.org, or better yet from the package specific repositories (https://github.com/emacsmirror/PACKAGE). Doing the latter would make it easier to do the review I think.

But that would be a huge amount of work if done properly, so just proxy maintaining a handful of package seems like a more feasible approach.

ahungry:
Was the removal to close a security hole or to push for active maintainers? 

There are many other unmaintained packages on Melpa, but for most packahes they need no ongoing updates (theyre a snapshot in time of something usable).

I would be happy to maintain and shepherd questions that come in related to these packages in the future (gotta love the GPL), as well as clean up bugs users have.

i would not be willing to go through and refactor to current Melpa package coding standards.

If this solution is not desirable, at the least Ill probably fork and rename the 10 or so in this list that I use and submit individually to Melpa.

milkypostman:
The removal was to close the security issue. 

I'd suggest that you just download the packages from the wiki and commit them to your config repo. I did this long before the removal because I don't trust the elisp from emacswiki unless I've had a look. 

I am sorry about having these things disappear and our lack of communication but this is just part of living in the open source world and I did the best possible to communicate recent problems. There are many different channels for communicating and it is not easy to find all the users. 

As for the response to this pr I'm in agreement with Purcell here. I don't think there will be a huge backlash about this.

purcell:
> If this solution is not desirable, at the least Ill probably fork and rename the 10 or so in this list that I use and submit individually to Melpa.

This would be the best outcome here, I think. No need to rename them, though: we'll be happy to consider your repos the canonical ones for MELPA recipe purposes.

PR #5270, opened 2018-01-27
Add recipe for ibuffer-sidebar

jojojames:

A sidebar for Ibuffer (similar to dired-sidebar).

https://github.com/jojojames/ibuffer-sidebar

Author/maintainer

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/ibuffer-sidebar
+++ b/recipes/ibuffer-sidebar
@@ -1,0 +1,1 @@
+(ibuffer-sidebar :fetcher github :repo "jojojames/ibuffer-sidebar")


purcell:
Nice work! Quick notes:

- `development` isn't the right parent group for the `ibuffer-sidebar` customisation group. Non-developers would find this useful too!
- In the definition of `ibuffer-sidebar-mode`, the body will get run both when the mode is enabled and disabled. The body of a minor mode definition for `the-minor-mode` generally contains an `(if the-minor-mode (... code to enable ...) (... code to disable ...))`.
- Instead of manually toggling `buffer-read-only`, you should generally `let`-bind `inhibit-read-only` while performing modifications.
- [Adding advice in a minor mode](https://github.com/jojojames/ibuffer-sidebar/blob/7f897598f1b0cb035b26fae5d4291e28572e04b9/ibuffer-sidebar.el#L162) is incorrect, since you're modifying the global behaviour of that function. If you need advice that is mode-specific, define it at the top level, then make your advice function a no-op if the minor mode is unset.

Hope that helps!

jojojames:
>    development isn't the right parent group for the ibuffer-sidebar customisation group. Non-developers would find this useful too!

Thanks, I switched to whatever ibuffer's group was.

>    Instead of manually toggling buffer-read-only, you should generally let-bind inhibit-read-only while performing modifications.

Thanks, done.

>    Adding advice in a minor mode is incorrect, since you're modifying the global behaviour of that function. If you need advice that is mode-specific, define it at the top level, then make your advice function a no-op if the minor mode is unset.

Hmnn, I think I captured what you have here, open to any corrections.


>    In the definition of ibuffer-sidebar-mode, the body will get run both when the mode is enabled and disabled. The body of a minor mode definition for the-minor-mode generally contains an (if the-minor-mode (... code to enable ...) (... code to disable ...)).

I intend ibuffer-sidebar-mode to be used from its setup function and not a mode that toggles on/off in a particular buffer. I think reversing the settings is a little tricky because capturing the initial state of ibuffer might be tricky. Open to any advice here.

Thanks!

purcell:
> I intend ibuffer-sidebar-mode to be used from its setup function and not a mode that toggles on/off in a particular buffer. I think reversing the settings is a little tricky because capturing the initial state of ibuffer might be tricky. Open to any advice here.

Sure. Calling it a minor mode is probably a misnomer then. At the very least you should move the existing minor mode body code into an `if` expression as suggested above, even if you then simply trigger an error if the mode gets disabled. Thinking about it, though, you might actually consider turning this into a major mode which is derived from `ibuffer-mode` -- I feel like that is a more logical fit, and would clean up the code a fair bit.

I'll merge this in the meantime, but there's still a bit of work to be done IMO. :-)

PR #5271, opened 2018-01-27
add lcr recipe

jyp:

We call a lightweight coroutine (or `lcr' for short) a function
which does not return its result directly, but instead passes it to
an extra *continuation* argument (often called 'continue' or
'cont').  Codinig with explicit contination arguments is a
well-known technique, called continuation-passing style (CPS).

CPS allows inversion of control to take place.  Indeed the
continuation may be installed as a callback, rather than being
called directly.  In general, any CPS function may yield control,
and thus implement a lightweight form of concurrency.

This module
provides:

- marcros which can translate direct-style code into cps
- a library of lcr's to read from processes, wait some time, etc.

Why use this module, instead of Emacs' built-in concurrency support?

- for better control over context switch and/or scheduling
- for versions of Emacs which do not provide concurrency

https://github.com/jyp/lcr

I am author (and maintainer) of the package.

N/A

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/lcr
+++ b/recipes/lcr
@@ -1,0 +1,2 @@
+(lcr :repo "jyp/lcr"
+     :fetcher github)


purcell:
Hah, I was reading the email notification for this and thinking "I wonder if that would be helpful for Danté", and it took me a moment to realise the connection. Looks good: merging now. :-)

PR #5273, opened 2018-01-27
Add recipe for git-attr

arnested:

Expose gitattributes to emacs buffers

I.e. putting files marked as `linguist-generated=true` into `read-only-mode` on visiting.

https://github.com/arnested/emacs-git-attr

Author/maintainer

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/git-attr
+++ b/recipes/git-attr
@@ -1,0 +1,1 @@
+(git-attr :fetcher github :repo "arnested/emacs-git-attr")


purcell:
Thanks. I'll go ahead and merge this, but a couple of quick notes:

- Typo: `attributtes`
- There's no need to call `executable-find` to locate `git`: if you can find `git` on the path with `executable-find`, then `call-process` will also be able to find it. :-)
- You might like to autoload one or two of these functions.

arnested:
Thank you, @purcell.

* I fixed the typo.
* I use the `executable-find` to make sure there actually is a `git` installed. Otherwise it will fail hard. I could wrap the call in `ignore-errors` -- bit I'm not sure what is more efficient: trying to launch an executable that is not there or search the PATH for the executable.
* I have autoloaded two of them now. My first thought was that it would be sufficient to have "higher level libraries" like my `git-attr-linguist` do autoloads but you are probably right and this makes the functionality more accessible in general.

purcell:
> use the executable-find to make sure there actually is a git installed. Otherwise it will fail hard. I could wrap the call in ignore-errors -- but I'm not sure what is more efficient: trying to launch an executable that is not there or search the PATH for the executable.

Ah, I see. Yes, probably checking with `executabl-find` is the best choice, then. :-)

Merging now - thanks!

PR #5274, opened 2018-01-28
Add panda-theme recipe

jamiecollinson:

`panda` is a dark color theme based on the color scheme from http://panda.siamak.work.

https://github.com/jamiecollinson/emacs-panda-theme

Maintainer.

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/panda-theme
+++ b/recipes/panda-theme
@@ -1,0 +1,1 @@
+(panda-theme :repo "jamiecollinson/emacs-panda-theme" :fetcher github)


jamiecollinson:
This is my first time contributing so wanted to mention that `package-lint` did turn up the issue

```
1:1: error: There is no (provide 'panda-theme) form.
```

From looking at other themes on melpa (e.g. [nord](https://github.com/arcticicestudio/nord-emacs/blob/develop/nord-theme.el)) this seems to always be the case and is simply a linting issue, but I'm happy to correct if I've missed something.

purcell:
Thanks for this submission!

> 1:1: error: There is no (provide 'panda-theme) form.

Yep, it's not strictly necessary, but it's also harmless to have both a `provide-theme` and a `provide` form.

Quick notes, then:
- If you've added the `Package-Version` header to appease `package-lint`, maybe just put `0` in it: MELPA will add a full one anyway. (The existing value is not a valid version number anyway, since it contains a spurious "t".)
- `Created 2018-01-11` is a terrible theme name. :-) This name shows up in the `customize-themes` menu, so perhaps consider something more Panda-related.

Looks good otherwise!

jamiecollinson:
Thanks for the notes @purcell. I didn't realise the `deftheme` docstring is user visible! New to emacs-lisp and thought `customize-create-theme` had done something sensible by default.

I've fixed the issues in https://github.com/jamiecollinson/emacs-panda-theme/commit/548b06f3e37f017fa3996015016fb0e565b84d93, let me know if I missed anything.

purcell:
Great - merged now! Welcome to MELPA. :tada:

PR #5275, opened 2018-01-28
Remove an old file from emacsql-sqlite's recipe

skeeto:
The file emacsql-system.el was removed from the emacsql-sqlite.el package, so it is no longer needed in the recipe.

https://github.com/skeeto/emacsql/commit/62d39157370219a1680265fa593f90ccd51457da


tarsius:
When was that? I don't remember that file at all.

skeeto:
I removed that file immediately before making this pull request, so "old 
file" might be misleading. Its purpose was just to generate a system 
tuple (i.e. "x86_64-linux"), but that's no longer needed.

https://github.com/skeeto/emacsql/blob/76489e0a10a46620df94e4aa42f92914c0849a1e/emacsql-system.el

PR #5277, opened 2018-01-30
add helm-rg recipe

cosmicexplorer:

[`helm-rg`](https://github.com/cosmicexplorer/helm-rg) exposes a [`helm`](https://github.com/emacs-helm/helm) interface to [`ripgrep`](https://github.com/BurntSushi/ripgrep). `ripgrep` is invoked as an asynchronous process, updating the search results as you type.

https://github.com/cosmicexplorer/helm-rg

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/helm-rg
+++ b/recipes/helm-rg
@@ -1,0 +1,1 @@
+(helm-rg :fetcher github :repo "cosmicexplorer/helm-rg")
--- a/recipes/rg3
+++ b/recipes/rg3
@@ -1,0 +1,1 @@
+(rg3 :fetcher github :repo "cosmicexplorer/rg3")
--- a/recipes/rg3
+++ b/recipes/rg3
@@ -1,1 +1,0 @@
-(rg3 :fetcher github :repo "cosmicexplorer/rg3")


purcell:
Thanks! Quick feedback:
- The recipe will include `update-commentary.el`. If that's a tool, put it in a subdirectory of your repo.
- Since this is helm-specific, why not call it "helm-rg"?
- The `M-d` and `M-g` key bindings are a bit random, and since they have common meanings, I'd expect they might be confusing. (Disclaimer: I'm not familiar with helm conventions, though.)
- The helm source is saved in a private var, but isn't it possible to expose it for general use, ie. not necessarily via the `rg3` wrapper command?
- `cl-mapc` can just be `mapc`, which is a standard built-in.

cosmicexplorer:
Wow, thanks a lot (seriously) for the very helpful comments!

> The recipe will include update-commentary.el. If that's a tool, put it in a subdirectory of your repo.

Whoops! Moved to a new [`gen-commentary/` subdirectory](https://github.com/cosmicexplorer/helm-rg/tree/master/gen-commentary).

> Since this is helm-specific, why not call it "helm-rg"?

Thanks a lot -- I hadn't considered that and it makes a lot of sense. I have no allegiance to the previous name, so I've renamed it to `helm-rg` (also reflected in the code and the PR title/post content).

> The M-d and M-g key bindings are a bit random, and since they have common meanings, I'd expect they might be confusing. (Disclaimer: I'm not familiar with helm conventions, though.)

Neither am I! This seems perfectly reasonable -- to clarify, are there obvious, specific keybindings you have in mind for either changing the current directory or the file glob pattern? If so, I would want to use those, otherwise I'll search around a bit to get a better idea of how to map these keys.

> The helm source is saved in a private var, but isn't it possible to expose it for general use, ie. not necessarily via the rg3 wrapper command?

Absolutely -- I hadn't considered that use case at all. I've changed the name of `rg3--process-source` into `helm-rg-process-source`. Is that what you meant?

> cl-mapc can just be mapc, which is a standard built-in.

Didn't know that! Fixed.

purcell:
> Absolutely -- I hadn't considered that use case at all. I've changed the name of rg3--process-source into helm-rg-process-source. Is that what you meant?

Yep.

> are there obvious, specific keybindings you have in mind for either changing the current directory or the file glob pattern? If so, I would want to use those, otherwise I'll search around a bit to get a better idea of how to map these keys.

Not really. Take a look around and see if there's a similar package you could crib from. :-)

PR #5282, opened 2018-02-01
Create feebleline

tautologyclub:

feebleline removes the modeline and replaces it with a slimmer proxy version that gets continuously written to *Minibuf-0* whenever no one else is using it. The modeline of course gets restored when feebleline-mode is toggled. 

https://github.com/tautologyclub/feebleline

I am the author and maintainer of the package. There are two skilled contributors in total, plus myself makes it three contributors in total.

feebleline has a high probability of conflicting with fancy modeline customizing packages -- I haven't tested this. But then again, if feebleline appeals to you, then fancy modelines probably do not.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/feebleline
+++ b/recipes/feebleline
@@ -1,1 +1,1 @@
-(feebleline :fetcher github :repo "tautologyclub/feebleline" :files ("feebleline.el"))  
+(feebleline :fetcher github :repo "tautologyclub/feebleline")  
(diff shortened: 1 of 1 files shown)


purcell:
Thanks! Fun idea. :-)

Quick notes:
- Please remove the `:files` spec from the recipe: its not needed.
- Don't use `/` in symbol names: convention would be to use the `feebleline--` prefix for private symbols.
- You should add an `;;;###autoload` cookie comment for `feebleline-mode`.

tautologyclub:
Fixed to the best of my ability!

purcell:
Nice! Merged. Welcome to MELPA. :tada:

PR #5286, opened 2018-02-04
add python-pytest package

wbolster:

python pytest runner inside comint, with various "dwim" interactive commands.

more modern, more features, and more opinionated, and more ‘works out of the box’ than other python testing related emacs packages.

https://github.com/wbolster/emacs-python-pytest

author/maintainer

none needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/python-pytest
+++ b/recipes/python-pytest
@@ -1,0 +1,3 @@
+(python-pytest
+ :fetcher github
+ :repo "wbolster/emacs-python-pytest")


wbolster:
is there anything i can do help move this forward? i've been getting questions why this is not yet on melpa. :)

purcell:
> is there anything i can do to help move this forward?

Just extremely busy, sorry. Thanks for your patience. Quick notes, then:

- Looks useful! Needs more info in the `Commentary` about how to get started.
- There's lots of "for emacs", which can be dropped. That's almost always redundant -- all elisp packages are for Emacs!
- Why [reimplement shell-quote-argument](https://github.com/wbolster/emacs-python-pytest/blob/master/python-pytest.el#L340-L344)? Seems risky. What about backslashes, for example?
- Generally enabling a major mode will [kill all local variables](https://github.com/wbolster/emacs-python-pytest/blob/master/python-pytest.el#L324) for you - maybe that's the case here?

Looks good otherwise. Hope that helps!

wbolster:
hey @purcell, thanks for the review. much appreciated.

> Looks useful! Needs more info in the Commentary about how to get started.

added some quick notes (without repeating the readme): https://github.com/wbolster/emacs-python-pytest/commit/a7ac61ebdd2e360c5f3f4cf237fb83546c3fe36f

> There's lots of "for emacs", which can be dropped. 

removed lots (4 😉) mentions of emacs: https://github.com/wbolster/emacs-python-pytest/commit/7a26d904c87314b4831eaa72bca3d955fc5c0706

> Generally enabling a major mode will kill all local variables for you

i was under the impression that what would only happen when switching to another major mode, but apparently calling `(some-major-mode)` will kill local vars regardless of the current `major-mode` (i checked). removed the superfluous call: https://github.com/wbolster/emacs-python-pytest/commit/5effd7594f777c21ed09379eb17df15a4ebf36e5

> Why reimplement shell-quote-argument? Seems risky. What about backslashes, for example?

this is done for a good reason. let me explain. (sorry for the verbosity.)

it's common to invoke `pytest` with the `-k` flag to filter which tests get executed, e.g. `pytest -k foo`. this flag takes an expression, so it could also be `pytest -k 'foo or bar'`, which could also be written as `pytest -k foo\ or\ bar`, which is equivalent but less readable.

the `python-pytest-function-dwim` command will ‘do what i mean for the current function’. it does a few things, but most relevant here is that for non-test functions, e.g. `def myfunc(): ...`, it will invoke `pytest` with `-k myfunc`. (see readme for details.)

building upon this, `python-pytest.el` performs a slightly magic but very useful trick when a `-k` filter is specified, and this ‘dwim’ function is used: it combines the user-specified `-k` filter value with the automatically generated one. for example, user-specified `-k 'foo or bar'` and auto-generated `-k myfunc` become `-k 'myfunc and (foo or bar)'`. this is the expected ‘dwim’ behaviour. (this is very useful to filter large parametrized tests that would normally run repeatedly with many inputs; this filtering is an effective way to test/fix/test/fix/test a single failing corner case of a function without repeating all other inputs all the time.)

one of the main features of `python-pytest.el` is that it is very easy to tweak the generated `pytest` command line in the minibuffer, execute that command, and then repeat that `pytest` command line many times until all tests pass.

and here is why the usefulness of the custom shell quoting becomes clear: manually editing a command line like
`pytest --x --y --z tests/test_foo.py -k 'myfunc and (foo or bar)'`
is easy and user-friendly. on the other hand, editing a command that looks like this is painful:
`pytest --x --y --z tests/test_foo.py -k myfunc\ and\ \(foo\ or\ bar\)`

regarding your ‘risky’: from the bash man page:

> There are three quoting mechanisms: the escape character, single quotes, and double quotes.

the first is what `(shell-quote-argument)` uses, and this quickly becomes unreadable, as shown above. the second is what `(python-pytest-shell-quote)` uses. quoting the the bash man page again:

> Enclosing characters in single quotes preserves the literal value of each character within the quotes.  A single quote may not occur between single quotes, even when preceded by a backslash. 

in other words: no escaping needed, but a single quote cannot occur. the way to deal with a single quote occurring in the string (very uncommon in practice luckily) is to split the string into three concatenated parts: 

- the part before the `'`, quoted using single quotes
- the `'` itself, quoted using double (!) quotes
- the part after the `'`, quoted using single quotes

so `here's an example` becomes three concatenated quoted strings: `'here'`, then `"'"`,  followed by `'s an example'`. and that's exactly what `(python-pytest-shell-quote)` does:

``` elisp
(defun python-pytest--shell-quote (s)
  "Quote S for use in a shell command. Like `shell-quote-argument', but prettier."
  (if (s-equals-p s (shell-quote-argument s))
      s
    (format "'%s'" (s-replace "'" "'\"'\"'" s))))
```

note that my custom implementation uses `(shell-quote-argument)` to detect whether any escaping is needed at all, so it won't quote unless necessary.

end result: correctly quoted strings, yet readable and hence easily editable as a command line in the minibuffer! victory!

ok. enough. :)

that said, i addressed all your remarks, and explained the why of the shell quoting approach... time to hit that friendly looking merge button, i'd say. :)

thanks again for the review, learning new things every day!

purcell:
Great explanation, thanks for taking the time. All looks great now, so that big friendly green merging is indeed going to get clicked right now. :-)

PR #5288, opened 2018-02-05
Add attrap

jyp:

Attrap provides a command to attempt to fix the flycheck error at point.
Attrap currently comes with builtin fixers for
`haskell-dante` and `emacs-lisp`. Support for other flycheck
checkers can be added dynamically (see `attrap.el`).

https://github.com/jyp/attrap

Author and maintainer

N/A

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/attrap
+++ b/recipes/attrap
@@ -1,0 +1,2 @@
+(attrap :repo "jyp/attrap"
+        :fetcher github)


purcell:
This is neat. You've got a `defcustom` in here that's in the `dante` group which seems like an oversight. I'd also suggest that the `'attrap-fixers` approach is inferior to (and much less idiomatic than) a customisable alist mapping from modes to fixer functions, because what you *really* want to do to find fixers is to iterate across all available modes with fixes and then check with `derived-mode-p` to see if the current major mode is _based on_ that major mode.

purcell:
Sorry for the slow reply, btw. Quite slammed currently.

purcell:
> However the fixers are not attached to major modes, but to flycheck checkers. This is quite necessary as there can be several tools per mode. So I do not think that derived-mode-p applies.

Ah right. I'm thinking more about the general case of such a library, even distinct from Flycheck. I was planning to write such a think at the end of last year but it didn't work out. :-)

Anyhow, merging now!

PR #5289, opened 2018-02-06
Add bmx-mode package.

josteink:

`bmx-mode` is a minor-mode which complements Emacs' built-in `bat-mode` to provide more IDE-like features.

https://github.com/josteink/bmx-mode

Author.

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/bmx-mode
+++ b/recipes/bmx-mode
@@ -1,0 +1,2 @@
+(bmx-mode :repo "josteink/bmx-mode" :fetcher github)
+


josteink:
Anyone want to take a look at this one yet?

cc: @purcell  @tarsius

purcell:
Thanks.

- You're missing package dependency declarations for `dash` and `s`.
- You need to provide a `defgroup` to define the `bmx-group` customisation group and place it in an appropriate location in the customisation hierarchy.
- The `Commentary` says `This minor-mode is cross-platform and does not depend on any native
 platform tools.` but that is all the information which will be displayed on the MELPA package page and via `?` in the `M-x list-packages` menu, so you should instead include a description of what this package does, and why people might use it, otherwise nobody will have any reason to try it out. (I still have no idea what it does, to be honest.)
- Don't [autoload the `company-backends` change](https://github.com/josteink/bmx-mode/blob/1fa02dfb3e3421d66969a9a21bda08152068ca4b/bmx-mode.el#L445): this might execute before `company-backends` has been defined, so you'll block the normal `company-backends` default entries. Instead, make this change at the top level when the file is loaded, or change the local value of `company-backends` when the mode is initialised.
- Also, don't autoload the `bat-mode-hook` change -- packages should not auto-enable themselves, because it should be possible for a user to install a package but not use it. Instead, tell users how to enable this minor mode as desired.
- Use `defvar` rather than `setq` to define the `bmx-mode-map` variable.

josteink:
Thanks for the constructive feedback! These should all be amended in the latest revision.

> (I still have no idea what it does, to be honest.)

I guess I've put a lot more effort into describing that in the [README-file](https://github.com/josteink/bmx-mode/blob/master/README.md) in the github repo than in the elisp-file itself.

I've updated the commentary to be more informative. Let me know if you still think it needs more work. (Are there any limits to how much you can enter here? Recommendations? Conventions?)

 > Don't autoload

These are reasonable default-settings for people wanting to use this mode. I've moved these settings to a dedicated function and recommended the invocation of this in the README as can be seen in [this commit](https://github.com/josteink/bmx-mode/commit/f7640431c7605db69896a4cd504fe888582f6905). (Should this also go in the commentary?)

Let me know if you still think there's more which needs to be tidied up or made clearer.

purcell:
> I've updated the commentary to be more informative. Let me know if you still think it needs more work. (Are there any limits to how much you can enter here? Recommendations? Conventions?)

It's good to provide a more detailed overview, and some simple "getting started" usage instructions (which can assume the elisp is already installed locally).

> These are reasonable default-settings for people wanting to use this mode. I've moved these settings to a dedicated function and recommended the invocation of this in the README as can be seen in this commit. (Should this also go in the commentary?)

Yeah, definitely worth mentioning. And you can feel free to `autoload` the `bmx-mode-setup-defaults` function for the user's convenience. :-)

Merging now, thanks!

josteink:
Awesome. Thanks!

PR #5291, opened 2018-02-06
admb-mode and tmb-mode

arnima-github:

`admb-mode` and `tmb-mode` are major modes for developing statistical models with the free software platforms AD Model Builder (http://admb-project.org) and Template Model Builder (http://tmb-project.org).

https://github.com/admb-project/admb/tree/master/contrib/emacs
https://github.com/kaskr/adcomp/tree/master/emacs

Creator and maintainer of both.

### Relevant communications with the upstream package maintainer.

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/admb
+++ b/recipes/admb
@@ -1,0 +1,3 @@
+(admb :fetcher github
+      :repo "admb-project/admb"
+      :files ("contrib/emacs/*"))
--- a/recipes/tmb
+++ b/recipes/tmb
@@ -1,0 +1,3 @@
+(tmb :fetcher github
+     :repo "kaskr/adcomp"
+     :files ("emacs/*"))


purcell:
Thanks! Quick feedback, first for `admb`:

- The [version variable](https://github.com/admb-project/admb/blob/7219ff4e1f348383c603b342125c5f8e480b0cea/contrib/emacs/admb.el#L9) breaks the file header so it can't be parsed. Better to not include it anyway, usually: MELPA will insert a version, which for MELPA Stable will be the last version tag on the repo.
- It's not appropriate to [change `PATH` for the user](https://github.com/admb-project/admb/blob/7219ff4e1f348383c603b342125c5f8e480b0cea/contrib/emacs/admb.el#L101-L102) as a side effect of them loading the library
- The `\n` literals in the multi-line docstrings are unusual and undesirable: just put in an extra literal line break.
- The keybindings are all over the place, ranging from function keys, M-up/M-down etc., M-return. You should follow the [emacs lisp conventions](https://www.gnu.org/software/emacs/manual/html_node/elisp/Coding-Conventions.html) in assigning these. Hard-coding tab to `indent-relative` shouldn't normally be done either: instead, set `indent-line-function` for your mode.
- Don't cram [multiple statement forms](https://github.com/admb-project/admb/blob/7219ff4e1f348383c603b342125c5f8e480b0cea/contrib/emacs/admb.el#L409) onto one line.
- When matching patterns against filenames, use `\\'` to match the end of the string, so that `.barrel` doesn't match `\\.bar` etc.
- Leave empty lines [between functions](https://github.com/admb-project/admb/blob/7219ff4e1f348383c603b342125c5f8e480b0cea/contrib/emacs/admb.el#L491) - really, the code is very unidiomatically formatted, and therefore harder to follow and review than it should be.
- Don't [define `prog-mode`](https://github.com/admb-project/admb/blob/7219ff4e1f348383c603b342125c5f8e480b0cea/contrib/emacs/admb.el#L701) if it's missing - instead, name `(emacs "24")` as a package dependency.
- Don't [assume that the buffer name is a file name](https://github.com/admb-project/admb/blob/7219ff4e1f348383c603b342125c5f8e480b0cea/contrib/emacs/admb.el#L462) - it might not be. See `(buffer-file-name)`. Also, don't ever `concat` filenames into commands without escaping them with `shell-quote-argument` -- they might contain spaces or quotes.

Re. `tmb`, I mostly have much of the same feedback.

Hope that helps!

DamienCassou:
Any news?

arnima-github:
Thank you Steve, for a thorough and well-thought review. I'll make revisions and get back to you.

DamienCassou:
Should we close this for now?

purcell:
Yes, closing for now due to lack of response. Feel free to re-open later.

PR #5292, opened 2018-02-06
conllu-mode recipe

odanoburu:

this major mode provides simple utilities for editing and viewing [CoNLL-U](http://universaldependencies.org/format.html) files, which are used in the natural language processing field. (a sample `.conllu` file can be found [here](https://github.com/own-pt/bosque-UD/blob/master/documents/CF1.conllu).)

https://github.com/odanoburu/conllu-mode

I'm the author and current maintainer of the package.

none needed

### Checklist

as an elisp beginner, I appreciate help with the two unmarked tasks!

  > I can install manually by using `package-install-file` on the `.tar` file generated by `make recipes/conllu-mode`, but installing directly after doing `C-c` on the recipe buffer throws the error `package-version-join: Wrong type argument: arrayp, nil`

### DIFF
--- a/recipes/conllu-mode
+++ b/recipes/conllu-mode
@@ -1,0 +1,3 @@
+(conllu-mode
+ :fetcher github
+ :repo "odanoburu/conllu-mode.git")
--- a/recipes/conllu-mode
+++ b/recipes/conllu-mode
@@ -1,3 +1,3 @@
 (conllu-mode
  :fetcher github
- :repo "odanoburu/conllu-mode.git")
+ :repo "odanoburu/conllu-mode")


unhammer:
>  My elisp byte-compiles cleanly 

They should compile without error from a "plain" Emacs.

From your repo, do `emacs  -Q .` – this should open the directory in dired. Then `% g el$ RET` to mark all the `.el` files and `B` to byte-compile them. Then fix any warnings.

unhammer:
Oh, and you may have to `(require)` or `(load)` any dependencies first from that emacs -Q

odanoburu:
@unhammer thank you for the tip! it compiles fine here.

there were some warnings that I took care of, and some that are ok to be left alone (unused variables in a destructuring-bind).

do you have any idea of what could be causing the problem in the unmarked bullet? I tried using edebugging the offending function, to no avail...

unhammer:
Seems like others have something similar: https://github.com/melpa/melpa/pull/5205#issuecomment-353267596 (does it also happen when you try other recipes?)

odanoburu:
@unhammer indeed, I do! I tried three recipes, and they gave the same error..

odanoburu:
yeah, it seems its a local problem, but should work fine when merged, as per https://github.com/melpa/melpa/pull/5205#issuecomment-365447375

purcell:
Thanks, looks good! Just a couple of quick notes:

- `C-c LETTER` bindings are [reserved for end user customisation](https://www.gnu.org/software/emacs/manual/html_node/elisp/Key-Binding-Conventions.html) so you shouldn't [use them](https://github.com/odanoburu/conllu-mode/blob/d99f0c0354c66d65c18ab1cf59f0242e667463c8/conllu-mode.el#L61-L62).
- `truncate-lines` and `whitespace-style` are user-level preferences, which it's not generally appropriate to [override](https://github.com/odanoburu/conllu-mode/blob/d99f0c0354c66d65c18ab1cf59f0242e667463c8/conllu-mode.el#L109-L110)

unhammer:
If people have already got those C-c bindings in their muscle memory, you can do things like `(defun modeprefix-set-up-C-c-bindings () …)` and have in the README a note about 

> if you want these bindings, you can add-hook …  modeprefix-set-up-C-c-bindings)
> and you might also want to add-hook toggle-truncate-lines" etc.

odanoburu:
@purcell thanks for the feedback!

- I'll change the `C-c LETTER` bindings, I'm sorry!

- in this case I'd think it is appropriate to override these settings locally: the file format this mode intends to help editing is kind of idiosyncratic... it is fundamental that the user is able to see the TAB marks, and having Emacs not truncate the lines makes it very easy to confuse the tab-separated fields. is it okay if I leave it as is? (I could also add a hook as @unhammer kindly suggested -- thank you! --, but I really don't think anyone would want like to have it disabled, so I'd prefer to have it working out-of-the-box..)

purcell:
Cool, makes sense. Thanks for your patience - merging ow.

odanoburu:
@purcell thank you! as a first-timer, I know had to learn a few new things...

PR #5296, opened 2018-02-10
Update nordless-theme upstream

lthms:
I have changed my github nickname recently, and created some kind of “parking account” to prevent someone to create a “rogue” repository lethom/nordless-theme.el. That being said, I am pretty sure updating the upstream of nordless-theme (my only package on Melpa) is a good idea :).

I want to apologise for the inconvenience. This will not happen again!

Have a great week-end.

### DIFF
--- a/recipes/nordless-theme
+++ b/recipes/nordless-theme
@@ -1,4 +1,4 @@
 (nordless-theme
-  :repo "lethom/nordless-theme.el"
+  :repo "lthms/nordless-theme.el"
   :files ("nordless-theme.el")
   :fetcher github)


tarsius:
> I have changed my github nickname recently, and created some kind of “parking account” to prevent someone to create a “rogue” repository lethom/nordless-theme.el.

Did you *rename* your account and then create a "new" account using the old name? I thought the latter was not possible - at least not for a while.

tarsius:
Anyway, since it says "Moved to @lthms" and "Joined GitHub on February 10, 2018" on https://github.com/lethom, I am fairly sure @lethom didn't delete all his repositories and you then took the opportunity to hijack one of his packages. It appears you are you. 😉

It does however appear that it is possible to take over the old name after an account was renamed, which is a problem. ☹️

lthms:
> It does however appear that it is possible to take over the old name after an account was renamed, which is a problem. ☹️

It is possible yes, and this is why I decided to create a new account. For now, the repository I own are still redirected (from lethom to lthms), but I am pretty sure if I create a repository with the same name, but the new account, the new repository “wins”.

Anyway, as I said, this will not happen again. Thank you for the quick merge!

tarsius:
> Anyway, as I said, this will not happen again.

Np. If you want to rename your account then that's perfectly find. And if you later decide to rename it again that's fine too - just don't make it a monthly thing. The security issue isn't your fault - it's github's. Thanks for noticing the issue and taking precautions by the way.

PR #5297, opened 2018-02-10
Add echange package recipe

kirill-gerasimenko:

Package provides 2 useful functions:

- opens email messages from MS Exchange in locally installed Outlook (or in default browser) by internet message id. Internet message id can be captured as a part of the link for org-mode
- fetches and saves calendar data from MS Exchange for 2 days to make it available as part of org-mode agenda.

https://github.com/kirill-gerasimenko/echange

I'm creator and maintainer.

### Checklist

Please confirm with `x`:

  I don't know how to check that, still new to Emacs and learning things. I'm willing to do that. Could you please point me to the way to do this? 
  Except of private methods which I don't see the reason to comment now (and manual https://www.gnu.org/software/emacs/manual/html_node/elisp/Documentation-Tips.html#Documentation-Tips says that only functions and variables inteded for users aka public should have docstrings

### DIFF
--- a/recipes/echange
+++ b/recipes/echange
@@ -1,0 +1,6 @@
+(echange
+ :fetcher github
+ :repo "kirill-gerasimenko/echange"
+ :files ("bin/echange.el"
+         "bin/echange.bat"
+         "bin/echange.jar"))


purcell:
Thanks for this!

> My elisp byte-compiles cleanly
> I don't know how to check that, still new to Emacs and learning things. I'm willing to do that. Could you please point me to the way to do this?

Sure - `M-x byte-compile-file`. Or just enable `flycheck` and also install `flycheck-package`.

A few quick notes separately from that:
- Every package dependency should be in the form `(package "ver")`, since the [bare symbol format](https://github.com/kirill-gerasimenko/echange/blob/353854a870df0d1a7abfd4d9568ca3de5d2310e2/src/elisp/echange.el#L10) you've used for some of these you aren't supported by all `package.el` versions.
- Don't change the [global values of another package's variables](https://github.com/kirill-gerasimenko/echange/blob/353854a870df0d1a7abfd4d9568ca3de5d2310e2/src/elisp/echange.el#L43-L45) - if you need certain values, then `let`-bind them for the scope of the code that needs them.
- The `defvar`s [here](https://github.com/kirill-gerasimenko/echange/blob/353854a870df0d1a7abfd4d9568ca3de5d2310e2/src/elisp/echange.el#L47) should probably be `defcustom`s instead, with a containing `defgroup`.
- `lexical-let` is obsolete, and when you've got `lexical-binding: t` enabled then you can simply use `let` instead. So add `-*- lexical-binding: t -*-` at the end of the package's first header line.

Hope that helps!

tarsius:
@kirill-gerasimenko ping?

purcell:
> I'm not sure though regarding lexical-let. This was the only way for the closure to work for me. I've spent lot's of time on getting this to work, but I'll have another round on this.

Maybe because your `(setq lexical-binding t)` has no effect: you need to set it in the first line of the package, by appending `  -*- lexical-binding: t -*-` to the first line of the file. Then if you re-open and then load the `.el` file, you should find that every lambda defined therein can use lexical binding.

tarsius:
Let us know when things are ready.

PR #5300, opened 2018-02-11
Add helm-lines

torgeir:

A helm interface for completing by lines already typed elsewhere in your git project.

https://github.com/torgeir/helm-lines.el

Maintainer

**None needed**

### Checklist

### DIFF
--- a/recipes/helm-lines
+++ b/recipes/helm-lines
@@ -1,0 +1,1 @@
+(helm-lines :repo "torgeir/helm-lines.el" :fetcher github)


purcell:
Thanks! Quick notes:

- [This line](https://github.com/torgeir/helm-lines.el/blob/bdb2a46e01d16bfa26de439d5b498fc7e60b6fcc/helm-lines.el#L44) could presumably just use `eolp` instead of doing regexp matches.
- Consider `string-empty-p` [here](https://github.com/torgeir/helm-lines.el/blob/bdb2a46e01d16bfa26de439d5b498fc7e60b6fcc/helm-lines.el#L73).
- `(when (not x) ..)` [here](https://github.com/torgeir/helm-lines.el/blob/bdb2a46e01d16bfa26de439d5b498fc7e60b6fcc/helm-lines.el#L104) can just be `(unless x ...)`
- Neither the package description nor `Commentary` mention that `ag` and `git` are required, but both seem pretty important.
- Why implement your own version of the built-in `vc-git-root` function? Relatedly, I'd suggest providing a `defcustom` like `helm-lines-project-root-function`, which you could set to `vc-git-root` by default, but could also be set to `projectile-root` by Projectile users.

Hope that helps!

torgeir:
The feedback, as always! 👏

I’ll have a go!

torgeir:
[4 out of the 5 have been fixed](https://github.com/torgeir/helm-lines.el/issues/1), the last one is still on my list.

purcell:
> 4 out of the 5 have been fixed, the last one is still on my list.

Here you go: https://github.com/torgeir/helm-lines.el/pull/2 :-)

tarsius:
@torgeir Please review and merge that pr soon. I intend to close this pr soon.

tarsius:
Actually, I am closing now and you can take as much time as you need. Please let us know when things are ready so that we can reopen this pr.

torgeir:
I completely forgot about this one! Will do.

torgeir:
@tarsius fixed!

torgeir:
Anything more I can do to make this happen?

DamienCassou:
I think you just have to wait for another round of review.

DamienCassou:
Some more feedback:

- the docstring of `-action` should have `indent` instead of `intent`. I suggest replacing "Insert..." with "Replace current line with..."
- usually, we keep `map` variables public to tell the user it can be modified. I would rename `helm-lines--map` to `helm-lines-map`.
- why is the docstring of `-async-shell-command` referencing `NAME`?
- to trim characters from a string, you can use `string-trim` if you want to
- don't wrap argument references within quotes in docstrings.
- in `--candidates`, I don't understand why you mix `format` and `concat`. I would either only use `format` with a literal first argument, or only use `concat` and pass the shell-quoted `query` and `root` as arguments.

DamienCassou:
I'm merging this now, above feedback is not mandatory.

torgeir:
Thanks, I'll have a go!

PR #5305, opened 2018-02-13
Added line-up-words

diml:

This package provides emacs integration for the `line-up-words` tool that aligns words in a region. For instance, selecting lines 2-10 in the following code and pressing `M-x line-up-words` align the code as shown below:

```ocaml
type package =
  { name : string
  ; dir : Path.t
  ; version : string
  ; description : string
  ; archives : Path.t list Mode.Dict.t
  ; plugins : Path.t list Mode.Dict.t
  ; jsoo_runtime : string list
  ; requires : t list
  ; ppx_runtime_deps : t list
  }
```

```ocaml
type package =
  { name             : string
  ; dir              : Path.t
  ; version          : string
  ; description      : string
  ; archives         : Path.t list Mode.Dict.t
  ; plugins          : Path.t list Mode.Dict.t
  ; jsoo_runtime     : string list
  ; requires         : t list
  ; ppx_runtime_deps : t list
  }
```

https://github.com/janestreet/line-up-words

This package is developed by [Jane Street](https://www.janestreet.com/). I'm a developer at Jane Street and I'm a member the team that takes care of publishing our open source packages on github.

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/line-up-words
+++ b/recipes/line-up-words
@@ -1,0 +1,2 @@
+(line-up-words :fetcher github
+               :repo "janestreet/line-up-words")


purcell:
Thanks! A few quick notes:

- You should omit the `:files` spec from the recipe, since the default is fine here.
- Does the underlying tool do anything that's fundamentally smarter than Emacs' own `align-regexp` function, which has a wide range of general and language-specific rules (see `align-rules-list`? (Similarly, in the example above, `M-x align-regexp` then `:` appears to give the same result.)
- The executable name should probably be a `defcustom`, associated with a new `defgroup`.
- The `Commentary` section refers people to the github README, so package users will see nothing about the package on its MELPA page or in the `list-packages` menu. Better to inline some docs there, in the self-documenting Emacs spirit. Also, there's no information provided about how to install the associated tool, which will deter people from using this.

Hope that helps!

diml:
Thanks for your comments! I didn't know about `align-regexp`, I'll ask to get more details about it

ninrod:
Hi @diml. Also see [evil-lion](https://github.com/edkolev/evil-lion). has the same idea, but applies to the [evil](https://github.com/emacs-evil/evil) ecosystem.

diml:
Hi @ninrod, thanks I didn't know about evil-lion either.

/cc @yminsky, who just told me that he has been using `align-regexp` when programming outside of work and he doesn't find it as good as `line-up-words`

yminsky:
align-regexp only aligns a single regexp, but this one can handle more complicated cases. For example, it will turn this:

``` ocaml
val foo : int -> float option
val snooble : string -> int option
```

into this.

``` ocaml
val foo     : int    -> float option
val snooble : string -> int   option
```

I believe it also handles some special cases in OCaml, like this one:

``` ocaml
type t =
  { foo : int -> foop 
  ; peeble : bleep -> bar
  (* this is some : stuff *)
  ; groupple : grimpgrimp -> gromp
  }                          
```

``` ocaml
type t =
  { foo      : int        -> foop 
  ; peeble   : bleep      -> bar  
  (* this is some : stuff *)
  ; groupple : grimpgrimp -> gromp
  }                          
```

Which I think involves some sensitivity to OCaml comments. @mkoconnor would know more about what it does.

mkoconnor:
Yep, that’s basically it. The other difference is that line-up-words will attempt to not change spacing inside quoted strings (although it only looks at the region it’s aligning, so in theory it could be wrong about what’s actually inside a quoted string, although in practice it seems to work well)

purcell:
Thanks all. Yeah, `align-regexp` alone certainly isn't comparable to `line-up-words`, but the `align` machinery in Emacs should be (if tweaked), so if it's not doing a nice job of aligning ocaml code now then an ideal way help Emacs users in the long term would be to create `align` rules, and either publish them separately, include them in `caml-mode` or even submit them back into the default Emacs ruleset.

That said, I understand the value of reformatting executables, and there's no harm in adding this package here, subject to covering the couple of little points in my previous comment.

edkolev:
@diml Purcell mentioned this above, but I think it's worth repeating - this package has no installations instructions for the binary, I doubt many users will be able to use it.

diml:
Sure, I'm going update the package and PR to take into account the review comments

diml:
Ok, I removed the `:files`, added some doc with installation instructions and added a `defcustom` with a group.

I agree that it'd be nicer to not require an external tool, in particular the installation part might look complicated for non OCaml users. It's just that we are more fluent in OCaml than Emacs lisp :)

purcell:
Great - merged, thanks!

diml:
Thanks :)

PR #5306, opened 2018-02-14
Add exercism.el recipe

jdenen:

Provides a minor mode with functions to manage [Exercism](http://exercism.io) practice problems and their solutions. 

https://github.com/jdenen/exercism.el

I am the maintainer.

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/exercism
+++ b/recipes/exercism
@@ -1,0 +1,1 @@
+(exercism :fetcher github :repo "jdenen/exercism.el")


purcell:
Thanks. The `with-eval-after-load` form [here](https://github.com/jdenen/exercism.el/blob/master/exercism.el#L82) doesn't do what you think - firstly, the first argument to the form doesn't name a feature that must be loaded before the block is executed, but in any case you should be running that code when someone starts trying to use the functionality in the library, which could be long after it has been loaded. Try making it a function which gets called from `exercism--run-command` instead, and throw an error if the config is found to be incomplete.

jdenen:
Sorry guys. I will get to this tonight if I can.

tarsius:
Let us know when things are ready.

PR #5307, opened 2018-02-14
Add mpdel

DamienCassou:

MPDel is an Emacs client for Music Player Daemon (MPD), a flexible,
powerful, server-side application for playing music.

https://github.com/DamienCassou/mpdel

I'm the package author and maintainer.

None needed

### Checklist

Please confirm with `x`:

- [ ] I've read [CONTRIBUTING.md](https://github.com/melpa/melpa/blob/master/CONTRIBUTING.md)


purcell:
Thanks! This looks like a nice alternative to the `mpd` support in EMMS, and I look forward to using it myself.

Code looks good. Couple of points about the packaging:

- `libmpdel.el` should either be a separate package (preferred) or be renamed to begin with `mpdel-`.
- Similarly for `ivy-mpdel`, though this looks like a clean add-on which should be a separate package: the core of `mpdel.el` doesn't seem to need Ivy in order to work properly. Somebody should be able to write `helm` code for `mpdel` without needing to have `ivy` installed in their Emacs.

DamienCassou:
Thank you for your review. I though about doing it but was too lazy. I've just updated the PR to only reference libmpdel. I will create a new PR with mpdel when I'm ready (I first need to implement an alternative native version for the feature provided by ivy-mpdel). I've put the libmpdel code into it's own git repository to let me release things separately.

DamienCassou:
I updated the PR with:

- a commit adding libmpdel: the library
- a commit adding mpdel: the main user-interface
- a commit adding ivy-mpdel: the ivy interface

My build process is currently broken because mpdel and ivy-mpdel depend on libmpdel which package-lint can't find in any package repository.

PR #5308, opened 2018-02-14
add rust-mode-reloaded

brotzeit:

It's a fork of https://github.com/rust-lang/rust-mode

https://github.com/brotzeit/rust-mode-reloaded

Unfortunately the original package is unmaintained. I have made several changes and tried to improve the situation.

I guess none needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/rust-mode-reloaded
+++ b/recipes/rust-mode-reloaded
@@ -1,0 +1,3 @@
+(rust-mode-reloaded
+    :repo "brotzeit/rust-mode-reloaded"
+    :fetcher github)


recipes/rust-mode-reloaded
```diff
@@ -0,0 +1,4 @@
+(rust-mode-reloaded
+    :repo "brotzeit/rust-mode-reloaded"
+    :fetcher github
+    :files (:defaults "*.el"))
```
tarsius: Remove that line. `*.el` is part of `:defaults`.

purcell:
> Unfortunately the original package is unmaintained. I have made several changes and tried to improve the situation.

The last commit on Rust Mode was 3 days ago, so it's not clear how it is unmaintained. Can you elaborate please?

Forked code must have files and identifiers which match the package name, so you'd have to rename everything (inside & out) to use `rust-mode-reloaded-` for this submission to be valid.

Even then, though, we'd discourage it - right now there's no indication of what's different in your code, or why there's a fork, so it's not obvious that sharing this code via MELPA would result in more benefit than confusion for potential users.

brotzeit:
> The last commit on Rust Mode was 3 days ago, so it's not clear how it is unmaintained. Can you elaborate please?

Obvious PRs get merged after a while, but many issues don't get answers and there are no improvements except bug fixing.

> Even then, though, we'd discourage it - right now there's no indication of what's different in your code

I have taken care of several feature requests from the issue tracker. I'm not really interested in maintaining a fork, but I'd like to see the mentioned features to be implemented. There's still a lot of stuff I have to fix, but hopefully I can deliver something substantially better soon and it gets merged into the original package. But there's nobody who reviews larger changes and who wants to discuss how things could be improved. So I guess I have to show some results.

If you want I can make a list of the changes I've made. But I can also wait and see if there's anybody who is interested in those changes. I've put my other packages in melpa because people asked me to. That's not the case here.

purcell:
> If you want I can make a list of the changes I've made. But I can also wait and see if there's anybody who is interested in those changes. I've put my other packages in melpa because people asked me to. That's not the case here.

Yeah, I don't really need to see a list, but I guess you just have to decide if you're going to keep maintaining and growing this fork. If so, maybe give it a new name and rename everything accordingly, add some info about your work in the README etc., and we'd definitely be willing to add the package here. Without the consent of the `rust-mode` authors, we wouldn't be able to bring this fork in as the new canonical `rust-mode` package.

In the meantime, then, I'll go ahead and close this, but please keep us posted!

PR #5309, opened 2018-02-14
Remove ac-anaconda recipe.

proofit404:
I've removed `ac-anaconda` package.

* I was broken for at least last 2 years. 
* No significant community interest in it (no active discussions or PRs to change the situation).
* Anyone interested in the auto-complete support can use `jedi.el` package.


purcell:
Makes sense, thanks.

PR #5310, opened 2018-02-15
Add pickle(mode), for cucumber gherkin files

ahungry:

Very small package for fontification of Gherkin (Cucumber) feature files

https://github.com/ahungry/pickle-mode

owner

na

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/pickle
+++ b/recipes/pickle
@@ -1,0 +1,3 @@
+(pickle
+ :repo "ahungry/pickle-mode"
+ :fetcher github)


purcell:
Thanks! Quick notes:

- Please use `define-derived-mode` to declare the major mode. This way you can set a parent mode explicitly, and also delete everything in the body of your current mode declaration (except the font-lock-keywords part) because the macro will generate it for you.
- `define-derived-mode` would also declare the keymap for you, so that you could simply poke the `C-j` binding into it for you, but you shouldn't override that binding anyway - it's a matter of personal preference. (Also, `C-j` is _already_ bound to `newline-and-indent` by default, I believe.)
- `define-derived-mode` declares the major mode hook too, so you can delete even more code. :-)
- Add an `;;;###autoload` for the major mode itself so that users can load it implicitly with `M-x pickle-mode`, or by adding it to `auto-mode-alist` themselves.
- Your docstring says this is a major mode for mercury files?

ahungry:
Thanks @purcell, updated according to your suggestions.  Looks a lot better now!  (Mercury reference was in regards to a leftover remnant from another unpublished mode for Mercury Lang I was working on awhile ago and botched copy/paste).

purcell:
Great, merged. BTW, `font-lock-defaults` is always buffer-local, so you can drop the `make-local-variable` portion and simply set it with `setq`.

PR #5311, opened 2018-02-15
Add lawndart

ahungry:

More current (better) mode for dart files

https://github.com/ahungry/lawndart-mode

owner

na

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/lawndart
+++ b/recipes/lawndart
@@ -1,0 +1,3 @@
+(lawndart
+ :repo "ahungry/lawndart-mode"
+ :fetcher github)


purcell:
Thanks. Some feedback:

- Uses `flycheck`, but doesn't declare that as a package dependency. Major modes shouldn't depend on specific minor modes like `flycheck` anyway. Even if that were okay, it wouldn't be okay for this package to redefine the `dart` flycheck checker.
- Shouldn't things like `indent-line-function` have been set by the initialisation code of the parent mode, `js-mode`? Has much of this code in the `define-derived-mode` form just been copied from `js-mode`?
- You're [autoloading aliases for built-in functions](https://github.com/ahungry/lawndart-mode/blob/master/lawndart.el#L191) here. :-/
- [This autoload](https://github.com/ahungry/lawndart-mode/blob/master/lawndart.el#L198) should be removed because it kidnaps the `dart` interpreter mode: packages shouldn't auto-enable themselves.

ahungry:
Ack, yea, this mode is (for me) a great improvement over the horrid indentation / unmainted mess the current dart-mode is, but I think its too rough around the edges atm (much of it is just a copy/paste of the mode it uses as the parent mode).

Enough to use for local dev, but too bad to publish atm, I'll withdraw this for now.

PR #5313, opened 2018-02-16
add `fuo` recipe

cosven:

Emacs client for [feeluown](github.com/cosven/feeluown-core) daemon.

https://github.com/cosven/emacs-fuo

Maintainer.

require emacs 24.4 or higher version

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/fuo
+++ b/recipes/fuo
@@ -1,0 +1,1 @@
+(fuo :repo "cosven/emacs-fuo" :fetcher github)


purcell:
Thanks! Quick notes:
- [Here](https://github.com/cosven/emacs-fuo/blob/4c21ee5b0c787cdefbc78d15e5d058d1752062ef/fuo.el#L30) you should use `shell-quote-argument` to escape `command` before inserting it into the string with `%s`.
- It shouldn't be necessary to have netcat installed: better to make a network connection with Emacs if possible, which should be easy.
- You can delete [this line](https://github.com/cosven/emacs-fuo/blob/4c21ee5b0c787cdefbc78d15e5d058d1752062ef/fuo.el#L203) and [this line](https://github.com/cosven/emacs-fuo/blob/4c21ee5b0c787cdefbc78d15e5d058d1752062ef/fuo.el#L205), because they are both generated by `define-derived-mode`.

Hope that helps!

cosven:
Thanks for your suggestions. @purcell 

I have updated the code base, which should fix the above problems. See [commit](https://github.com/cosven/emacs-fuo/commit/31a269654bd5f2760076bc74b20c25fc0ad9ef8f)

- 1: problem no more exists.
- 2: use network module instead of netcat
- 3: those two lines have been deleted

purcell:
Great! Merging now, thanks. Welcome to MELPA! :tada:

PR #5316, opened 2018-02-18
Add daemons recipe

cbowdon:

Provides a UI for managing system services.

https://github.com/cbowdon/services-mode

Maintainer

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/daemons
+++ b/recipes/daemons
@@ -1,0 +1,1 @@
+(daemons :fetcher github :repo "cbowdon/daemons.el")


purcell:
Thanks! This looks useful, and I'd personally enjoy seeing or contributing a `brew` backend for us MacOS users in due course. :-)

Before digging into this too much, there's a mismatch because the recipe name is `services-mode`, but your feature is `services`. The two must match. However, there's already [a package called `services`](https://github.com/davep/services.el), and it would be more idiomatic to not use the `-mode` suffix for this new package (even if it weren't confusing). So I'd suggest a different name, unless @davep would like to volunteer to rename his to `servicedb` or similar.

A couple of other little points along the way:
- The `services-do-no-evil` stuff is a distraction that looks like it should simply live in your emacs init file.
- The `-pkg.el` file is unnecessary for MELPA, and we'd advise you remove it. We'll read package metadata from the main `.el` file, ie. the one which matches the package name.

Hope that all helps!

davep:
I don't have an issue with a rename if it's the only sensible solution.
While the package has had that name for a very long time, I doubt there are
many users.

purcell:
Thanks @davep. I think `daemons` would also be a good name - and available - so my preference would probably be to rename the newcomer.

davep:
Agreed.

cbowdon:
Thanks @purcell for your feedback! It's my first package so I really appreciate your help getting it straight.

I'm happy to rename this to `daemons`, it is a good name and I think likely users will find it just as easily as `services`. (Before writing this, I searched for two things: services and daemons). Thank you @davep for offering though.

I've addressed this and your other comments in the package repo, which is now https://github.com/cbowdon/daemons.el. I can rename this PR and force-push a new commit with the updated name so that it's still associated with this discussion, or create a new PR and link to this PR. Which would you prefer?

purcell:
Just go ahead and force-push commits to this PR. :-)

manuel-uberti:
Friendly ping :)

cbowdon:
Woohoo, thank you very much :)

PR #5319, opened 2018-02-20
Moving "Related" to a better place

julien-montmartin:
Hello,

I would like to move my "Related" package to a better place. Nothing changes, except my Bitbucket user account, and a repo with issue tracking.

Best regards, 

Julien

### DIFF
--- a/recipes/related
+++ b/recipes/related
@@ -1,1 +1,1 @@
-(related :fetcher git :url "https://bitbucket.org/lyude/related")
+(related :fetcher git :url "https://bitbucket.org/julien-montmartin/related")


purcell:
This is probably a silly question, but how would we verify that the bitbucket users `lyude` and `julien-montmartin` are the same person?

julien-montmartin:
Hello Steve,

This is a very good question, indeed. :) I wrote you an answer with my old "lyude" Bitbucket account, so you know it's one of my accounts !

https://bitbucket.org/lyude/related/src/73420e62700c2fdf6d9f96a9179b8618864c19ae/HelloSteve.txt

purcell:
Nice - thanks!

PR #5323, opened 2018-02-21
add counsel-css recipe

waymondo:

counsel-css is an ivy-mode backend for css selectors (works for SCSS and LESS too). It also has a hook to integrate its parser into Imenu.

https://github.com/hlissner/emacs-counsel-css

enthusiastic user

none needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/counsel-css
+++ b/recipes/counsel-css
@@ -1,0 +1,1 @@
+(counsel-css :repo "hlissner/emacs-counsel-css" :fetcher github)


manuel-uberti:
Any updates on this one? :)

purcell:
Looks good, and I've merged this now, but hey, @hlissner, you should please add a minimum Emacs version dependency to the `Package-Requires` header. I see you're relying on `subr-x`: `package-lint` can help you see which version you might need to specify. It's best not to rely on transitive dependencies that `counsel` etc. might have.

hlissner:
@purcell And done! 

However, package-lint didn't complain about `subr-x` missing from my Package-Requires header, nor could I figure out what version to use. With `(subr-x "0")` it complains:

```
error     Package subr-x is not installable. (emacs-lisp-package)
warning   Use a properly versioned dependency on "subr-x" if possible. (emacs-lisp-package)
```

In a brief search, I couldn't find another package that depends on subr-x to reference. Any advice?

purcell:
Yeah, `package-lint` will detect the use of many _functions_ from `subr-x`, and suggest a corresponding minimum Emacs version. You're correct that `subr-x` itself shouldn't be declared as a package dependency.

PR #5326, opened 2018-02-22
Various Makefile improvements

tarsius:
Most importantly make it possible to load a `package-build.el` that is not located inside the melpa repository, which is very useful when you are working on that library as I currently am.

This is done by adding a `LOAD_PATH` variable. Also add a `LISP_CONFIG` variable, which makes it possible to set the various path variables used by `package-build`. Finally load `config.mk` if that exists, which provides a convenient way to set the mentioned variables.

And to top if of also some more or less cosmetic changes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 .#*
 /.cask
 /.ecukes-failing-scenarios
+/config.mk
 /download_log.db
 /download_log.json.gz
 /elpa
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,7 @@
 TOP := $(abspath $(dir $(lastword $(MAKEFILE_LIST))))
 
+-include ./config.mk
+
 SHELL         := bash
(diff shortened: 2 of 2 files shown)


purcell:
Thanks - took me a while to get around to looking at this, sorry. Will also follow up with your big `package-build` refactoring. I'm keen to merge both changes, but just wary of debugging any breaking changes that occur only on the build server.

tarsius:
It would have been better to merge this (possibly after rebasing). If there is a bug, then it will be harder now to find it.

purcell:
Oh, I see - github remembers the squash/merge preference, and I routinely squash PRs here to avoid interim commits while contributors make requested changes to PRs.

> If there is a bug, then it will be harder now to find it.

On the flip side, it's easy to revert in case of emergency. Trade-offs, trade-offs! :-)

tarsius:
I figured that's what happened :wink:

PR #5329, opened 2018-02-24
update paperless

atgreen:

Emacs powered PDF filing system.

https://github.com/atgreen/paperless

Author, maintainer

**None needed** 

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/paperless
+++ b/recipes/paperless
@@ -1,3 +1,3 @@
 (paperless :fetcher github
 	   :repo "atgreen/paperless"
-	   :files ("paperless.el"))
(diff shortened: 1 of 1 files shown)


purcell:
Thanks. `org-paperless.el` was only excluded because there was a `:files` spec in the recipe: the default is to include `"*.el"` (but exclude test files). So would you mind adjusting this change in order to instead _remove_ the `:files` spec entirely? Thanks!

purcell:
Never mind - this was trivial to do myself via the github UI. Cheers!

PR #5336, opened 2018-02-26
Update flower recipe repository branch

profitware:

Flower provides simple interface to list tasks (issues) from Jira, TFS, Gitlab and Github. It uses clomacs under the hood for integration with Flower Clojure library.

https://github.com/PositiveTechnologies/flower

Author and maintainer

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/flower
+++ b/recipes/flower
@@ -1,3 +1,4 @@
 (flower :repo "PositiveTechnologies/flower"
+        :branch "emacs"
         :files (:defaults "project.clj")
         :fetcher github)


profitware:
Hi @purcell,

After creating original recipe in #5302 I have realised that Emacs package is updated on each `project.clj` update - I suppose it's not the desired behaviour for me to have package updated on each `flower` Clojure library update (even `-SNAPSHOT` ones). So I decided to point the recipe to a specific `emacs` branch.

Thank you!

tarsius:
> So I decided to point the recipe to a specific emacs branch.

Maybe it would make sense to maintain the elisp in a separate repository altogether?

profitware:
Hi @tarsius,

> in a separate repository altogether

Thank you for your question. I have already considered this possibility before posting the original recipe. There are some limitations in that approach:

1. The [clomacs](https://github.com/clojure-emacs/clomacs) library requires `project.clj` to be in the same package and I have not found MELPA fetcher that can fetch files from two repositories simultaneously.
2. Copying `project.clj` (alone) each time the library gets new enhancements on the Clojure part would decrease maintainability - it would require to update two reposiories instead of one: there is [some Clojure code for Emacs exclusively](https://github.com/PositiveTechnologies/flower/tree/master/flower-utilities).
3. This Clojure code for Emacs may not be split off because it conforms the rules of [version updating](https://github.com/PositiveTechnologies/flower/blob/master/project.clj#L22) and [deploying](https://github.com/PositiveTechnologies/flower/blob/master/project.clj#L23) like other parts of the library do.
4. The authentication by tokens is done by special `~/.credentials.edn` file, so it would decrease user experience to switch between wikis to find out how [authentication in task trackers is done](https://github.com/PositiveTechnologies/flower/wiki/1.2.-Authentication); copying wiki page between two wikis also reduces the maintainability.

So I decided to keep mono repository to maintain Clojure and Elisp code alogether.

Thank you.

purcell:
> So I decided to point the recipe to a specific emacs branch

Yeah, this is really not ideal from our perspective: we want to reliably build your latest elisp into packages, so what if you stop updating the `emacs` branch after a while? MELPA will be stuck forever.

Either `project.clj` is part of the elisp package or not. I suspect it's not logically part of the package, and that the elisp simply requires a minimum version of it. So why not make the elisp fetch the `project.clj` on demand from the repo?

Alternatively, you can keep an appropriate `project-for-elisp.clj` file in the repo next to the `.el` file on `master`, then the recipe can contain something like `:files ("*.el" ("project.clj" "project-for-elisp.clj"))`. Then, when you want to update the `project.clj` which the elisp uses, you just copy `project.clj` to `project-for-elisp.clj` in your repo.

profitware:
You have a point. I have reconsidered the way I was handling unstable versions and found out that having `-SNAPSHOT` versions stay in `master` branch is not a good idea. So I would rather change the way of library development to conform `git flow` with `develop` branch to track current changes and `master` to have stable ones. Thus the library would be divided into several logical parts still staying in mono repository, `project.clj` in the root would serve only for elisp package.

I'm closing this PR for now. Thank you.

PR #5338, opened 2018-02-27
Add flycheck-posframe

alexmurray:

Uses posframe to show flycheck error / warning messages (similar to the existing flycheck-pos-tip etc)

https://github.com/alexmurray/flycheck-posframe

Maintainer

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/flycheck-posframe
+++ b/recipes/flycheck-posframe
@@ -1,0 +1,1 @@
+(flycheck-posframe :fetcher github :repo "alexmurray/flycheck-posframe")


purcell:
Looks good, but you've declared a local minor mode which has global effects, and I'm not sure if that's what you intended. In the current code, you'll be hooking into every command in every buffer, even if `flycheck` is not active in that buffer. If that's necessary, declare `:global t` for your mode, but if not, then consider making this a local minor mode (with a globalized version too, perhaps) so that you can add hooks to local versions of `post-command-hook` etc.

alexmurray:
Thanks for the insightful review - is https://github.com/alexmurray/flycheck-posframe/commit/95de5e23a1da74010ad25c12c4b75cc201315ef4 more like what you mean for avoiding global effects?

purcell:
> Thanks for the insightful review - is alexmurray/flycheck-posframe@95de5e2 more like what you mean for avoiding global effects?

Yeah, I think so, but I'd also maybe expect to see `setq-local` too, so that you could (theoretically) have some buffers using different `flycheck-display-errors-function` values.

alexmurray:
Sure makes sense - thanks again - https://github.com/alexmurray/flycheck-posframe/commit/9ebf295a9af6a93aba8a4687c091df8c5a157a0a

purcell:
Great - and now you can just `kill-local-variable` to unset the overidden vars. :-)

In any case, merged!

alexmurray:
You mean instead of setq-local to revert the changes? Nice, I was not aware of that.

alexmurray:
Actually turns out `flycheck-display-errors-function` has `:risky t` - also I was trying to see if I could set it locally and then unset it but it still doesn't seem to work - flycheck perhaps does some caching internally I think... So setq-local doesn't work as expected - and then neither does kill-local-variable - after disabling flycheck-posframe-mode with the kill-local-variable it still shows via posframe :-/ - even if I try doing:

```
(make-local-variable 'flycheck-display-errors-function)
(setq-local flycheck-display-errors-function #'flycheck-posframe-show-posframe)
```

So am not sure how to proceed - should we go back to the `(setq flycheck-display-errors-function ..)` version?

purcell:
> So am not sure how to proceed - should we go back to the (setq flycheck-display-errors-function ..) version?

Ah, yes, if it's `:risky`, then sure. Ultimately, you're better placed to know what works - I'm just heckling from the sidelines. :-)

PR #5343, opened 2018-02-28
Add recipe for ‘gitlab-ci-mode’

joewreschnig:

gitlab-ci-mode is an Emacs major mode for editing GitLab CI files.
It provides syntax highlighting and completion for keywords and special
variables.

https://gitlab.com/joewreschnig/gitlab-ci-mode.git

Author

### Checklist

Please confirm with `x`:

- [X] I've used the latest version of [package-lint](https://github.com/purcell/package-lint) to check for packaging issues, and *mostly* addressed its feedback (see below)

It's a multi-file package which I'm not 100% sure how to check with `package-lint` since it errors on files without version, etc. headers, and only the main file in the package has those. However, the lack of those is its only complaint.

### DIFF
--- a/recipes/gitlab-ci-mode
+++ b/recipes/gitlab-ci-mode
@@ -1,0 +1,3 @@
+(gitlab-ci-mode
+ :fetcher git
+ :url "https://gitlab.com/joewreschnig/gitlab-ci-mode.git")


purcell:
Thanks. This looks handy.

> It's a multi-file package which I'm not 100% sure how to check with package-lint since it errors on files without version, etc. headers, and only the main file in the package has those. However, the lack of those is its only complaint.

`package-lint` doesn't really handle multiple files. The only real issue I see here is that your [gitlab-ci-mode-flycheck.el](https://gitlab.com/joewreschnig/gitlab-ci-mode/blob/master/gitlab-ci-mode-flycheck.el) won't compile on systems without `flycheck`. In fact, since this file has an additional external dependency package, it should be a separate package itself: in general, major modes should be self-contained and minimal, and when you hypothetically decide to start providing support for revamped `flymake` in new Emacs versions, you don't want to force everyone to also have `flycheck` installed.

joewreschnig:
OK. I've split the Flycheck integration off into a separate package, which I'll submit a PR for after this one is approved.

purcell:
Great - merged!

PR #5347, opened 2018-03-02
Add mpmc-queue

smizoe:

mpmc-queue is an implementation of multiple-producer-multiple-fconsumer queue for emacs >= 26.
This wraps queue structure provided in [queue.el](https://elpa.gnu.org/packages/queue.html) and offers operations useful in multi-threaded programming.

https://github.com/smizoe/mpmc-queue

I'm the author/maintainer.

None needed

### Checklist

Please confirm with `x`:

Regarding the last item, I got an error with message `package-version-join: Wrong type argument: arrayp, nil`, but it seems to be harmless according to https://github.com/melpa/melpa/pull/5205 and several others.

### DIFF
--- a/recipes/mpmc-queue
+++ b/recipes/mpmc-queue
@@ -1,0 +1,3 @@
+(mpmc-queue
+ :repo "smizoe/mpmc-queue"
+ :fetcher github)


purcell:
Thanks, looks good. Merging now. I'd suggest that your `mpmc-queue--emacs-minimal-version` code is too elaborate: if it's necessary (and it probably isn't), you can just check the emacs version at the top level and call `error` if it is too low.

smizoe:
Thanks for the merge and suggestion!

PR #5352, opened 2018-03-04
Add recipe for grep-context

mkcms:

Provides commands to increase context around matches in compilation/grep buffers.

https://github.com/mkcms/grep-context

I'm the author

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/grep-context
+++ b/recipes/grep-context
@@ -1,0 +1,1 @@
+(grep-context :fetcher github :repo "mkcms/grep-context")


purcell:
Looks good, thanks, and sorry for the delay - merging now.

BTW, the `eval-when-compile` wrapper for `cl-lib` is pretty much superfluous: my understanding is that it used to be essential for `cl` but isn't really worthwhile with `cl-lib`.

PR #5355, opened 2018-03-05
Add recipe for org-index.

marcIhm:

A personal adaptive index for org.

https://github.com/marcIhm/org-index

I am the author.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/org-index
+++ b/recipes/org-index
@@ -1,0 +1,2 @@
+(org-index :fetcher github
+	   :repo "marcihm/org-index")


tarsius:
> org-index.el is also in the contrib-directory of org

In such cases it is better to not also add it as a separate package, else two packages - this one and `org-plus-contrib`.

tarsius:
In the past @bzg has said that he prefers if Org extensions are maintained separately instead of being added to `contrib/`, but I don't know how that plan is coming along.

bzg:
Yes, I still have this plan. We will release Org 9.1.8 soon, then 9.2, then remove the `contrib/` directory from Org's repository. The idea is to have `org-contrib.git` living on https://code.orgmode.org.

marcIhm:
Hi, I see the Point; if asked to decide between org-contrib and melpa, I opt for melpa, because it is easier to install from there. So just removed from org-contrib :-( I will add it again, if found useful, once org-contrib has moved.

marcIhm:
Currently org-index is no longer in the head of org (which is okay for bzg and me). However, the latest tag of org (9.1.8) still contains org-index.el. So the situation is somewhat mixed ... 
How to proceed then ? Can this pull-request be accepted or should it wait for the next tag of org (probably 9.2), which will no longer contain org-index ?

bzg:
FWIW, I hope to release Org 9.2 in a month or so.

marcIhm:
Thats not Long :-) Would be fine with waiting until then.

marcIhm:
Yes, made it three subdirs (each with .nosearch). Looks cleaner now :-)

tarsius:
Thanks!

purcell:
Merged now, and I know this is pre-existing code, but I'm just going to say that I think [this function](https://github.com/marcIhm/org-index/blob/master/org-index.el#L2887-L3207) is pretty extreme and could probably be split up. The function says it's written a particular way for performance, but much of its code is for setting up keybindings that could instead be defined in a little minor mode. :-)

marcIhm:
Thanx, for your suggestion !
I am still learning lisp and this sounds like a good improvement.

Am 08.04.2018 um 10:50 schrieb Steve Purcell:

PR #5358, opened 2018-03-08
add package recipe babcore

ArneBab:

babcore provides a basic setup with convenience customizations.

https://bitbucket.org/ArneBab/babcore.el/

Maintainer and main author (mostly due to collecting, curating and minimizing a list of snippets over the past decade).

**None needed**

### Checklist

Please confirm with `x`:

- [x] I've used the latest version of [package-lint](https://github.com/purcell/package-lint) to check for packaging issues, and addressed its feedback -- addressed all errors, but not all warnings (I do not want to restrict the versions of dependencies)

### DIFF
--- a/recipes/babcore
+++ b/recipes/babcore
@@ -1,0 +1,4 @@
+(babcore
+  :fetcher bitbucket
+  :repo "ArneBab/babcore.el"
+  :files (:defaults "babcore.org"))


purcell:
Thanks for taking the time to submit this! We don't generally accept clumps of loosely-related user-level Emacs configuration as packages, because they're unlikely to be useful to an audience wider than the initial user, and will generally need to be loaded in specific ways at start-up time to work as intended. So I'm going to decline this, sorry.

ArneBab:
Hm, ok -- thank you for checking it. I wanted to move babcore to a more easily maintainable place than marmalade (an old version is there -- which still was quite a bit larger -- but I could no longer get that updated), but if it’s not suitable for melpa, then that’s as it is.

PR #5360, opened 2018-03-09
Add duplicate-region

craiglyons:

Duplicate an active region or the current line, above or below the current location, and activate it as a region.

[craiglyons/duplicate-region](https://github.com/craiglyons/duplicate-region)

Author/maintainer

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/duplicate-region
+++ b/recipes/duplicate-region
@@ -1,0 +1,1 @@
+(duplicate-region :repo "craiglyons/duplicate-region" :fetcher github)


purcell:
Thanks! Libraries like this get submitted from time to time, but there are already options like `move-dup` and `duplicate-thing`. I use the former, and I think it does what your library does. Generally we prefer not to add packages which are very close in functionality to pre-existing packages, because it just adds to user confusion -- it would be better if authors collaborated, in many cases.

craiglyons:
Fair point, I didn't see those before. I'll see about adding non-overlapping features to one of those.

purcell:
Great - sorry to turn your submission away. :-)

PR #5363, opened 2018-03-09
Update zig-mode repository

AndreaOrru:
The package is already in MELPA. I just moved the repository to the `zig-lang` organization. Just updating the recipe here.


purcell:
Thanks!

PR #5365, opened 2018-03-10
Alfontzo - Stateful font-configuration per machine/host/os

guidoschmidt:

I use Emacs across different machines across the day
(Macbook at work with Retina displays, Macbook Air at home,
Linux machine at home with an older 1280px display). On each
machine, I prefer a different font size, because each machine 
has a different screen resolution.

Alfontzo provides a font configuration per operating system type and machine — e.g. you work on your Laptop and use Fira Code at size 14, but on your desktop you have a display with a higher resolution and thus want to work with Consolas at size 18. Alfontzo provides `M-x alfontzo-fontface` and `M-x alfontzo-typescale` as a fast and convenient way of setting fontface and size on your current machine and stores them in `custem-set.el` to remember it for the next time Emacs is used on that specific machine.

https://github.com/guidoschmidt/alfontzo

Creator & maintainer

**None needed**

### Checklist

Please confirm with `x`:


purcell:
Thanks! Couldn't this be implemented much more simply by setting the `family` and `size` of the `default` face? Multiple themes can be active at the same time, so you could even accomplish this by loading a secondary theme which sets only those two attributes. In its current form this seems like a lot of machinery. As a general rule, whenever you make something that varies according to the specific machine Emacs is running on, you probably want to lean on the customisation system as heavily as possible.

PR #5366, opened 2018-03-10
Add org-radiobutton

Fuco1:

Add support for radiobuttons (checkboxes with exactly one selected item at all times) to org checklists

https://github.com/Fuco1/org-radiobutton

Author

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/org-radiobutton
+++ b/recipes/org-radiobutton
@@ -1,0 +1,1 @@
+(org-radiobutton :fetcher github :repo "Fuco1/org-radiobutton")


purcell:
Thanks. I'll go ahead and merge this, but I think you should consider having a local and globalised _pair_ of minor modes, rather than just a global mode. It'll eliminate the weird iteration across `(buffer-list)` you're doing in the mode definition body, because the globalised minor mode machinery will handle that for you.

Fuco1:
@purcell I guess you're talking about `define-globalized-minor-mode`?  I was only vaguely aware of that, I'll fix this, thanks for the tip.

purcell:
> I guess you're talking about define-globalized-minor-mode? I was only vaguely aware of that, I'll fix this, thanks for the tip.

Exactly - looks great now! (P.S. `turn-on-org-radiobutton-mode-if-desired` should be `org-radiobutton-maybe-enable` to keep consistent symbol prefixes.)

PR #5367, opened 2018-03-12
ob-clojurescript

statonjr:

Add support for evaluating blocks of ClojureScript code using Org-babel.

https://gitlab.com/statonjr/ob-clojurescript

Author

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/ob-cljs
+++ b/recipes/ob-cljs
@@ -1,1 +1,0 @@
-(ob-cljs :repo "statonjr/ob-cljs :fetcher gitlab)
--- a/recipes/ob-cljs
+++ b/recipes/ob-cljs
@@ -1,0 +1,1 @@
+(ob-cljs :repo "statonjr/ob-cljs :fetcher gitlab)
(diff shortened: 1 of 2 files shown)


purcell:
Thanks! I'll go ahead and merge this, but please note that you should likely fix [this line](https://gitlab.com/statonjr/ob-clojurescript/blob/master/ob-clojurescript.el#L81) to use `shell-quote-argument`, instead of using ad-hoc `\"` to escape the command-line argument string. Safer this way.

Welcome to MELPA! 🎉

tarsius:
Another library by that name already exists (https://github.com/bloat/babel-clojurescript/blob/master/ob-clojurescript.el) but it looks very unfinished and wasn't worked on for six years.

I also had a look at your new library and it looks like you have `(setq-default tab-width 2)`. That doesn't work well for anyone who hasn't done the same, which is almost everyone. If you really want to do that, then you have to add this at the end of every file that you publish:

```lisp
;; Local Variables:
;; tab-width: 2
;; End:
```

Of course Gitlab doesn't understand that, so the file will still look weird there. I would recommend that you use this instead:

```lisp
;; Local Variables:
;; indent-tabs-mode: nil
;; End:
```

And `M-x untabify` the whole buffer in the same commit.

PR #5369, opened 2018-03-15
Add djangonaut recipe.

proofit404:

Minor mode to navigate around Django project. It's like rinari for Rails.

https://github.com/proofit404/djangonaut

Author

**None needed**

### Checklist

### DIFF
--- a/recipes/djangonaut
+++ b/recipes/djangonaut
@@ -1,0 +1,1 @@
+(djangonaut :fetcher github :repo "proofit404/djangonaut")


purcell:
Thanks! Couple of quick notes:

- Please remove "Emacs" from the package description line: it's redundant as all emacs packages and minor modes are for Emacs. `package-lint` should have warned about this.
- `C-c r` is a keybinding [reserved for end-user customisation](https://www.gnu.org/software/emacs/manual/html_node/elisp/Key-Binding-Conventions.html), so you must not use it. You might like to provide a prefix keymap and suggest that users bind it to `C-c r`, though. `package-lint` would also have told you about this issue.

proofit404:
I saw those warnings during package publish preparations. But I ignore them since some popular packages do not follow this practice as well. Sorry for the inconvenience.

All fixed as you advice.

proofit404:
Any news?

purcell:
Yep, looks good, but even though it's now (more easily) configurable you should still change the default from `C-c r`, e.g. to `C-c '`.

I also noticed there's a missing package dependency on `s`, which the code uses.

proofit404:
Done. Thanks for review!

purcell:
Great!

PR #5370, opened 2018-03-17
Add move-2

m-cat:

`move-2` provides more efficient movement commands like `move-2-move-left-twice`.

https://github.com/m-cat/move-2

I am the maintainer.

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/move-2
+++ b/recipes/move-2
@@ -1,0 +1,1 @@
+(move-2 :fetcher github :repo "m-cat/move-2")


purcell:
Thanks. Did you know about `C-2 C-f`, `M-2 M-b`, `M-6 M-n` etc? Maybe read up on "universal arguments" and how they work with common commands like `forward-char`. Since there's a general mechanism for this, I don't think this is a worthwhile standalone package, sorry. It's easy enough to include one or two such functions in an emacs startup file.

Further notes re. the code, since you might find it helpful:
- There's no point aliasing a built-in function with a new name, as in `move-2-move-left`
- `(forward-char 2)` is more efficient than calling your own wrapper for `(forward-char)` twice
- Text like `in the text editor Emacs` is redundant and distracting in elisp code, which can only run in the text editor Emacs. :-)

PR #5372, opened 2018-03-17
Add company-box recipe

sebastiencs:

A front-end for `company-mode`

https://github.com/sebastiencs/company-box

Author

### Checklist

### DIFF
--- a/recipes/company-box
+++ b/recipes/company-box
@@ -1,0 +1,3 @@
+(company-box
+ :fetcher github
+ :repo "sebastiencs/company-box")


purcell:
Thanks. Please run `package-lint` and build the package locally as requested above.

Couple of other points about the code:
- Don't define faces with just `:foreground` or `:background` - try to inherit default (even arbitrary) faces wherever possible, so that existing themes will work smoothly with your code, otherwise define both `dark` and `light` versions of the face.
- "A company front end" doesn't tell potential users why they might use this, so why would the install the package? What's interesting about it? It's worth including just a few words to explain this, since you're publishing it for wider use.
- I think your `~` variables are supposed to be package-private. Use `--` instead, in order to follow existing conventions.
- Your minor mode definition contains only code for turning the mode on. The body of the mode definition should check for the `company-next-mode` variable and handle both turning it on and turning it off.

sebastiencs:
Thanks for the review.

I have rename the package following [this discussion](https://github.com/company-mode/company-mode/issues/756)

> Please run package-lint and build the package locally as requested above.

Done

> Don't define faces with just :foreground or :background - try to inherit default (even arbitrary) faces wherever possible, so that existing themes will work smoothly with your code, otherwise define both dark and light versions of the face.

Done

> "A company front end" doesn't tell potential users why they might use this, so why would the install the package? What's interesting about it? It's worth including just a few words to explain this, since you're publishing it for wider use.

Done

> I think your ~ variables are supposed to be package-private. Use -- instead, in order to follow existing conventions.

I don't like this convention because when a user execute `describe-function` or `describe-variable` and type `company-box-` he will get completions for private and public var/func, which is a bit confusing from a user POV, IMHO.
With `~`, the user will only see the public items.
Is it possible to keep the `~` ?

> Your minor mode definition contains only code for turning the mode on. The body of the mode definition should check for the company-next-mode variable and handle both turning it on and turning it off.

It already contains code to turn off the mode: [here](https://github.com/sebastiencs/company-box/blob/da9a34ed4b60e621f983758aa1a4e17cf84fc584/company-box.el#L609)

dgutov:
> I don't like this convention

Let's not try to invent different conventions in every package. If the behavior of `describe-function` is not to your liking, please report that to Emacs developers.

> It already contains code to turn off the mode: here

What happens if someone calls `(company-box-mode -1)` twice?

sebastiencs:
> If the behavior of describe-function is not to your liking, please report that to Emacs developers.

This has nothing to do with `describe-function`, but about the completions items a user get when writing a variable/function of the package:
For example, if a user wants to change a variable, he will type:
`(setq company-box-[]`
With `--`, the user will also see private variables in the completion list

> What happens if someone calls (company-box-mode -1) twice?

Nothing, is it suppose to do something ?

dgutov:
> but about the completions items a user get when writing a variable/function of the package

That completion table is selected inside `describe-function`. `help--symbol-completion-table`, to be exact.

> Nothing, is it suppose to do something?

OK, here's a smaller question: does `(company-box-mode -1)` properly disable the mode? It should.

sebastiencs:
> That completion table is selected inside describe-function. help--symbol-completion-table, to be exact.

That's good to know

> > It already contains code to turn off the mode: here

>  does (company-box-mode -1) properly disable the mode? 

Smaller answer: yes

dgutov:
> Smaller answer: yes

Indeed, that works fine.

I think you could split `company-box~set-mode` in two functions, though. For readability.

purcell:
> It already contains code to turn off the mode: here

Great!

> With ~, the user will only see the public items.
> Is it possible to keep the ~ ?

No, sorry. These days we insist that authors follow the elisp community standards, since the point of publishing a package is to share it with the community. The standards may not be perfect, but they play a big part in everyone being able to contribute to each others code, so conformity here beats personal preferences.

sebastiencs:
> No, sorry. 


@purcell Alright, it's done

pkulev:
Also you can squash commits, because the package renaming within one PR is not necessary for melpa commit history.

sebastiencs:
Done

purcell:
Merged now - thanks for your patience. :-)

tarsius:
@sebastiencs Are you going to package [icons-in-terminal](https://github.com/sebastiencs/icons-in-terminal) for Melpa too?

sebastiencs:
@tarsius As discussed [here](https://github.com/sebastiencs/icons-in-terminal/issues/8#issuecomment-320904174), I'm not sure. But since there aren't many changes on the repo, it might be a good idea

PR #5374, opened 2018-03-18
Add speeddating recipe

xuchunyang:

Increase date and time at point. For example, there is a date `1999-12-31` and the point is on `31`, Increase it by one will change it into `2000-01-01`.

https://github.com/xuchunyang/emacs-speeddating

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/speeddating
+++ b/recipes/speeddating
@@ -1,0 +1,1 @@
+(speeddating :fetcher github :repo "xuchunyang/emacs-speeddating")


xuchunyang:
I can't built my recipe or any other recipe. I suspect it is because a bug of package-build.el, see https://github.com/melpa/package-build/issues/14. I also notice a obvious bug:

https://github.com/melpa/melpa/blob/191ec704d37de2506f7feeaf16e3bb9ebadebe0f/package-build/package-build.el#L866

should be changed to

    (package-build--archive-file-name (assq (intern name) (package-build-archive-alist)))) 

because the key of the alist is symbol, not string.

tarsius:
Thanks! I have pushed that fix.

xuchunyang:
Thanks. Now I can build my recipe successfully.

purcell:
Nice, thanks!

PR #5376, opened 2018-03-20
Add recipe for citeproc-el

andras-simonyi:

citeproc-el is a library for rendering citations and bibliographies in styles described in the Citation Style Language (CSL), an XML-based, open format to describe the formatting of bibliographic references (see http://citationstyles.org/ for further information on CSL).

https://github.com/andras-simonyi/citeproc-el

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/citeproc
+++ b/recipes/citeproc
@@ -1,0 +1,1 @@
+(citeproc :fetcher github :repo "andras-simonyi/citeproc-el")


purcell:
Thanks! I don't have much feedback on the code itself, which looks great (just little things, like maybe using `-if-let*` [here](https://github.com/andras-simonyi/citeproc-el/blob/master/citeproc-term.el#L105) - returning `nil` explicitly is generally an antipattern).

Regarding the recipe, rather than supplying a `:files` spec in order to exclude your test file, please just rename that file to the more conventional `citeproc-tests.el`, and then the default file pattern will work fine.

andras-simonyi:
Thanks a lot for your feedback -- very much appreciated! I've renamed `citeproc-test-suite.el` to `citeproc-suite-tests.el` and simplified the recipe accordingly. I've also simplified the definition which you referred to by using `-if-let`. 

Maybe this is not a good place to ask, so please disregard if inappropriate, but just out of curiosity: Regarding returning nil explicitly being an antipattern, what would you consider the preferred Emacs Lisp idiom when a function has to return either nil or a non-Boolean object depending on a condition (e.g., a search result)?  I see using `when`, `and`, or `if` with only one argument as the alternatives, but `when` is usually recommended only for side-effects, `and` only for Boolean arguments, and `if` with one argument is typically also frowned upon, which seems to leave `if` with an  explicit `nil` argument the only solution (especially when the `nil` return value is important for the function's use).

purcell:
Great, merged now - sorry for the slow feedback.

Re. your questions:



> Regarding returning nil explicitly being an antipattern, what would you consider the preferred Emacs Lisp idiom when a function has to return either nil or a non-Boolean object depending on a condition (e.g., a search result)? I see using when, and, or if with only one argument as the alternatives, but when is usually recommended only for side-effects, and only for Boolean arguments, and if with one argument is typically also frowned upon, which seems to leave if with an explicit nil argument the only solution (especially when the possibility of a nil return value is important for the function's use).

I'm not sure that those recommendations you're referring to are good ones, and I've never seen them made myself. Embracing the implicit `nil` return value of unsatisfied `when` and `unless` forms leads to neater code, and doing that is definitely idiomatic in elisp.

andras-simonyi:
Thanks a lot, and sorry for responding only now.

> I'm not sure that those recommendations you're referring to are good ones, and I've never seen them made myself. Embracing the implicit nil return value of unsatisfied when and unless forms leads to neater code, and doing that is definitely idiomatic in elisp.

Thanks, I'll try to follow that in the future. Just for the record, the recommendations I referred to were the Emacs Wiki's [When to use if](https://www.emacswiki.org/emacs/WhenToUseIf) page:

> Use ‘when’ or ‘unless’ if the value of the conditional expression is not important, that is, if you are executing the expression only for its side effects.

and Google's (admittedly, Common) [Lisp Style Guide](https://google.github.io/styleguide/lispguide.xml): 

> Note that in Common Lisp, WHEN and UNLESS return NIL when the condition is not met. You may take advantage of it. Nevertheless, you may use an IF to explicitly return NIL if you have a specific reason to insist on the return value. You may similarly include a fall-through clause (t nil) as the last in your COND, or (otherwise nil) as the last in your CASE, to insist on the fact that the value returned by the conditional matters and that such a case is going to be used. You should omit the fall-through clause when the conditional is used for side-effects.

purcell:
Thanks for the links - those are interesting counterpoints.

PR #5379, opened 2018-03-22
recipe for gams-mode.

ShiroTakeda:
Modified the recipe for gams-mode.

This is a major mode for editing GAMS (General Algebraic Modeling System) files.

I just modified recipe file for gams-mode in order to add an additional lisp program (gams-auto-complete.el).

https://github.com/ShiroTakeda/gams-mode

I am a maintainer.

Not require other packages. 

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/gams-mode
+++ b/recipes/gams-mode
@@ -1,5 +1,5 @@
 (gams-mode
  :repo "ShiroTakeda/gams-mode"
  :fetcher github
- :files ("gams-mode.el")
+ :files ("gams-mode.el" "gams-auto-complete.el")
 )


purcell:
This would introduce a new dependency on the `auto-complete` package, which would need to be declared. The correct way to handle this would be to instead distribute `gams-auto-complete` as a separate library which depends on `gams-mode` and `auto-complete`, so I'll close this PR pending that change.

ShiroTakeda:
I understand your suggestion. I will make a new library for gams-auto-complete. 
Thank you.

PR #5380, opened 2018-03-22
Add banner-comment

WJCFerguson:

Trivial Emacs package to format a comment as a banner.  Will reformat an existing banner.

https://github.com/WJCFerguson/banner-comment

I'm the maintainer

**None needed**

### Checklist

### DIFF
--- a/recipes/banner-comment
+++ b/recipes/banner-comment
@@ -1,0 +1,1 @@
+(banner-comment :repo "WJCFerguson/banner-comment" :fetcher github)


purcell:
Thanks, this looks handy!

- There's some `package-lint` feedback you've missed here.
- I suggest you use the standard more verbose header boilerplate which makes the licence explicit.
- Where you've hard-coded 80 as the default width, I suggest you use the value of the `fill-column` variable, which users who prefer longer or shorter lines will already have set according to their preferences.

WJCFerguson:
Good feedback thanks, I'll address those items.  The 80 column thing occurred to me after submitting.

WJCFerguson:
Right, should have resolved those now...

WJCFerguson:
Oh - I just read your suggestion properly for fill-column - instead I made it another variable.  `fill-column` sounds more sensible...  I'll change it to that.

purcell:
Great, merged now. Welcome to MELPA! :tada:

PR #5385, opened 2018-03-24
Remove anything and all packages that depend on it (#3004)

tarsius:
Follow up to https://github.com/melpa/melpa/issues/3004#issuecomment-375912059, https://github.com/melpa/melpa/issues/3004#issuecomment-375915794 (giving Steve a change to intervene).

Anything was long ago superseded by Helm, one of the most popular
packages of all time.  Remaining users of Anything should probably
move to Helm.

Anything is one of only two packages that Melpa still fetches over
an insecure connection.  And that, combined with a maintainer who
is hard to contact, is the reason why we finally pull the plug.

### DIFF
--- a/recipes/ac-cake
+++ b/recipes/ac-cake
@@ -1,1 +1,0 @@
-(ac-cake :repo "k1LoW/emacs-ac-cake" :fetcher github)
--- a/recipes/ac-cake2
+++ b/recipes/ac-cake2
@@ -1,1 +1,0 @@
-(ac-cake2 :repo "k1LoW/emacs-ac-cake2" :fetcher github)
--- a/recipes/anything
+++ b/recipes/anything
@@ -1,4 +1,0 @@
-(anything
- :fetcher git
- :url "http://repo.or.cz/r/anything-config.git"
- :files ("*.el" "contrib/*" "extensions/*.el"))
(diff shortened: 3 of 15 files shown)


purcell:
Yes, I support this. If secure hosting were provided, and users missed this, I'd be willing to reconsider.

However, the inclusion of a couple of `ac-*` packages here seems spurious -- they depend on `auto-complete`, not `anything`.

tarsius:
> However, the inclusion of a couple of `ac-*` packages here seems spurious

That only seems so :wink: 

* `ac-cake` depends on `cake`, which depends on `anything`.
* `ac-cake2` depends on `cake2`, which depends on `anything`.

However it might be possible to remove that dependency from `cake` and `cake2`, either by removing support for `anything` completely or by moving it to optional `anything-cake.el` `anything-cake2.el` libraries. Nowadays we would insist on that anyway before adding a new package (to avoid forcing the authors favorite completion framework on all users of their package (including users who prefer another completion framework)).

@k1LoW would it be feasible to do the above?

@myuhe Likewise could you please also look into removing the dependency on `anything` from `imgur`?

k1LoW:
I talk about `*cake*.el` .

1. Remove `ac-cake` `ac-cake2` `cake` `cake2` from melpa
2. If someone needs `*cake*.el`, do recovery work

Thank you for your great works.

PR #5388, opened 2018-03-25
frame-local recipe

sebastiencs:

Access variables local to a frame

https://github.com/sebastiencs/frame-local

Author

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/frame-local
+++ b/recipes/frame-local
@@ -1,0 +1,3 @@
+(frame-local
+ :fetcher github
+ :repo "sebastiencs/frame-local")


xuchunyang:
@tarsius The build fails due to `package-build` now uses `cl-defmethod` thus requires Emacs 25 and Travis is using Emacs 24.5.

DamienCassou:
This [has been fixed](https://github.com/melpa/package-build/issues/17). Can someone please run the test suite again on Travis?

DamienCassou:
You have some comments written like this:

```emacs
(defun frame-local-set (name value &optional frame)
  "Set the symbol NAME's value to VALUE in the current frame.
If FRAME is non-nil, set the symbol in this frame.
```
I often see similar comments written like this instead:

```emacs
(defun frame-local-set (name value &optional frame)
  "Set the symbol NAME's value to VALUE in FRAME.
If FRAME is nil, set the symbol in current frame.
```

> Similar to `frame-local-set' but NAME doesn't need to be quoted.

should that be "*must not* be quoted" instead?

I find that code a bit weird:

```emacs
(setq frame-local--obarrays (delq frame frame-local--obarrays)
         frame-local--obarrays (delq obarray frame-local--obarrays)))
```

what about merging the 2 assignment into one like this (untested): `(setq frame-local--obarrays (delq obarray (delq frame frame-local--obarrays))))`

tarsius:
> This has been fixed. Can someone please run the test suite again on Travis?

We had to wait a few hours for the Melpa `package-build` package to become available. Turns out there's another issue and we have to wait another few hours. But I locally tested with the Emacs version used by Melpa, so changes are good this fix is enough.

sebastiencs:
@DamienCassou Thanks for the review. It's fixed now

purcell:
Thanks - merged. Sorry for the very slow turnaround on this.

basil-conto:
@sebastiencs Out of curiosity, what is the intended use-case for `frame-local` vs using [frame parameters](https://www.gnu.org/software/emacs/manual/html_node/elisp/Frame-Parameters.html), as suggested here: https://emacs.stackexchange.com/a/40852/15748 ?

sebastiencs:
@basil-conto I was using frame parameters for [sidebar.el](https://github.com/sebastiencs/sidebar.el) but I've found out that when listing +2000 files, I was losing ~1 second to access the frame parameters.
I could probably solve the slowdown another way, but I have prefered to write `frame-local`, which creates an obarray for each frame, I think it is cleaner than using frame parameters, and definitely faster.
Also it doesn't pollute the frame parameters with potentially lots of variables.

If one don't care about theses criterias, I guess that frame parameters are good enough

basil-conto:
@sebastiencs OK, thanks for the explanation. :)

PR #5389, opened 2018-03-26
Add blacken recipe.

proofit404:

Integration with python black tool

https://github.com/proofit404/blackify

author

back-mode does not start with the package prefix. The same way as yapfify and isortify packages do.

### Checklist

Please confirm with `x`:

- [ ] I've used the latest version of [package-lint](https://github.com/purcell/package-lint) to check for packaging issues and addressed its feedback

### DIFF
--- a/recipes/blacken
+++ b/recipes/blacken
@@ -1,0 +1,1 @@
+(blacken :fetcher github :repo "proofit404/blacken")


DamienCassou:
The build has failed because of package-build. The bug [has been fixed](https://github.com/melpa/package-build/issues/17). Can someone please run the test suite again on Travis?

DamienCassou:
Could you please add to the first line of your file that "black" is an external code formatter? You could also add a link to the Commentary section. Before reading your code and searching the web, I stupidly thought "black" was a color :-).

> back-mode does not start with the package prefix. The same way as yapfify and isortify packages do.

is there a reason why you don't follow the convention of always prefixing? Is package-lint happy with your file?

You might want to make the shell command name (`"black"`) a `defcustom` that users can tweak. I guess some users won't have the command in their `PATH`.

If `original-buffer` is [narrowed](https://www.gnu.org/software/emacs/manual/html_node/emacs/Narrowing.html#Narrowing), your code will only send a portion of it to `black` but will then replace the entire content with black's output. This will erase parts of the original buffer's content.

proofit404:
> Could you please add to the first line of your file that "black" is an external code formatter?

Fixed.

> Is there a reason why you don't follow the convention of always prefixing?

Fixed.

> You might want to make the shell command name ("black") a defcustom that users can tweak.

Fixed.

> If original-buffer is narrowed, your code will only send...

I'm not familiar with it and have no idea how to fix it. PR will be appreciated.

proofit404:
I just merged @purcell PR. Tested it locally. I think we are good to go.

purcell:
Yep, thanks!

proofit404:
Great! Thanks!

PR #5391, opened 2018-03-27
Add upbo recipe

shiren:

Upbo is javascript test runner integration for emacs
Currently It support Karma but it will support Jest soon.

https://github.com/shiren/upbo

I wrote this :)

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/upbo
+++ b/recipes/upbo
@@ -1,0 +1,3 @@
+(upbo
+ :fetcher github
+ :repo "shiren/upbo")


purcell:
Thanks! Quick feedback:

- Instead of `upbo-git-root-dir`, just use `vc-git-root`, or even `vc-root-dir`, if you don't need to be `git`-specific.
- You have the word `"Emacs"` in the package description line, and you're using [reserved](https://www.gnu.org/s/emacs/manual/html_node/elisp/Coding-Conventions.html) `C-c u` keybindings -- `package-lint` should have warned you about these issues when you ran it.
- `upbo-mode-hook` calls `upbo-mode` -- that's not what hooks are for. You should just delete that function.
- You don't need to `generate-new-buffer` [here](https://github.com/shiren/upbo/blob/master/upbo.el#L72) - you can just use `(with-current-buffer (get-buffer-create upbo-view-buffer) ...`

Hope that helps.

shiren:
Thank you!
I applied whole your comments :)

purcell:
Thanks. Sent you two further little fixes in https://github.com/shiren/upbo/pull/4 and https://github.com/shiren/upbo/pull/5.

purcell:
Merged now - welcome to MELPA! :tada:

PR #5399, opened 2018-03-29
Added a recipe for `gams-ac' package.

ShiroTakeda:

This package provides auto-complete source for gams-mode ([https://github.com/ShiroTakeda/gams-mode](https://github.com/ShiroTakeda/gams-mode)).

https://github.com/ShiroTakeda/gams-ac

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/gams-ac
+++ b/recipes/gams-ac
@@ -1,0 +1,5 @@
+(gams-ac
+ :repo "ShiroTakeda/gams-ac"
+ :fetcher github
+ :files ("gams-ac.el")
+)
--- a/recipes/gams-ac
+++ b/recipes/gams-ac
@@ -1,5 +1,4 @@
 (gams-ac
  :repo "ShiroTakeda/gams-ac"
  :fetcher github
- :files ("gams-ac.el")
 )


recipes/gams-ac
```diff
@@ -0,0 +1,5 @@
+(gams-ac
+ :repo "ShiroTakeda/gams-ac"
+ :fetcher github
+ :files ("gams-ac.el")
```
DamienCassou: please remove the `:files` line, it is [probably useless](https://github.com/melpa/melpa#recipe-format).

DamienCassou:
Some feedback:

- you might want to add `-*- lexical-binding: t; -*-` at the end of the first line
- please remove the `.` at the end of the last line
- the [the emacs lisp style guide](https://github.com/bbatsov/emacs-lisp-style-guide) recommends

    > [Placing] all trailing parentheses on a single line instead of distinct lines

- I'm not sure your macro makes sense because the resulting code is not much more complex. Additionally, the variables could benefit from a docstring that the macro doesn't support. I recommend inlining the macro-expansion and removing the macro-definition
- I believe `gams-ac-attach-$` is not worth it: I would inline it

ShiroTakeda:
Dear Damien

Thank you very much for useful suggestions and advises.

I want to modify the code.

ShiroTakeda:
Dear Damien

According to your suggestions, I rewrote the entire part of `gams-ac.el`.

I abolished macro and use normal variables and functions instead.

purcell:
Thanks - I think you no longer need to `(require 'cl)`, and you might like to add an `;;;###autoload` cookie comment for the `gams-ac-setup` and `gams-ac-after-init-setup` functions. In any case, this is ready to merge. :-)

ShiroTakeda:
Thank you, Steve.

According to your suggestions, I rewrote the code.

PR #5401, opened 2018-03-30
Add recipe for repeater

xuchunyang:

Repeat a command automatically. For example, if you type <kbd>C-n</kbd> three times, the command will be executed repeatedly, so eventually you will get to the end of the buffer.

https://github.com/xuchunyang/repeater

Maintainer

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/repeater
+++ b/recipes/repeater
@@ -1,0 +1,1 @@
+(repeater :fetcher github :repo "xuchunyang/repeater")


DamienCassou:
Some feedback:
- please run linters specified in CONTRIBUTING.md, you have errors, e.g.,:
    - there is no version number
    - you have missing or problematic docstrings

purcell:
Thanks - this looks okay, but I suggest you make `repeater-mode` a local minor mode, and then use `define-globalized-minor-mode` to make a global version of it. This way, users will be able to toggle the behaviour on a per-buffer or per-emacs-instance basis. (You'd have to modify the buffer-local version of `post-command-hook`, of course.)

PR #5408, opened 2018-04-02
Add omnibox recipe

sebastiencs:

Completion/selection system for Emacs
This package requires #5388 to be merged

https://github.com/sebastiencs/omnibox

Author

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/omnibox
+++ b/recipes/omnibox
@@ -1,0 +1,3 @@
+(omnibox
+  :repo "sebastiencs/omnibox"
+  :fetcher github)


DamienCassou:
When opening the file, my linter tells me there are 56 issues with your file, mostly because of checkdoc. Please fix your file.

purcell:
> When opening the file, my linter tells me there are 56 issues with your file, mostly because of checkdoc. Please fix your file.

Most are on internal functions (`omnibox--*`) so I wouldn't worry too much about that, but obviously a docstring like
```el
(defun omnibox (&rest plist)
  "Omnibox."
```
isn't very helpful. The main reason to write a little documentation is not to keep checkdoc happy, but because you're sharing code with the community, and you'd presumably therefore like them to understand how to use it and extend it. :-)

Otherwise this mostly looks good, and it's certainly a very interesting piece of work. A few thoughts:

- It feels like this could use the recently-published `posframe` package. And there's `ivy-posframe`, which also seems related.
- You shouldn't have _any_ hardcoded colours in the file: instead of `(:background "#1CA0C7" :foreground "black")`, you should define faces that users can theme. By default, `:inherit` built-in faces so that it looks fairly good out of the box on any system.

DamienCassou:
> Most are on internal functions (`omnibox--*`) so I wouldn't worry too
> much about that

my opinion is that it costs more to let warnings around and train your
brain to ignore them than just to fix the docstring of internal
functions. Additionally, chances are high that you will miss important
warnings within a bunch of un-important ones. My goal is always to have
warning-free code and I have tools preventing me from merging any code
with warnings.

-- 
Damien Cassou

purcell:
> my opinion is that it costs more to let warnings around and train your brain to ignore them than just to fix the docstring of internal functions. Additionally, chances are high that you will miss important warnings within a bunch of un-important ones. My goal is always to have warning-free code and I have tools preventing me from merging any code with warnings.

I agree, and I'm the same with my own work, but across the wider community a lack of docstrings on _internal_ functions shouldn't be a barrier to MELPA inclusion IMO.

purcell:
Great - now merged! :tada:

PR #5409, opened 2018-04-03
Remove packages fetched over insecure connections

tarsius:
Melpa used to fetch these packages from the Emacsmirror, which in turn
imported these (and many more) packages by downloading individual
libraries over insecure connections.

The Emacsmirror no longer imports any packages over insecure
connections and because various attempts to get the maintainers to use
https have failed, these packages have been move to the Emacsattic.
Once a package is in the Emacsattic, it is no longer updated.

We have decided to remove insecure packages from Melpa too (#3004).

| Package       | Downloads |
| ------------- | --------- |
| escreen       |     1,616 |
| fm            |       981 |
| hl-sexp       |    25,452 |
| ldap-mode     |       877 |
| list-register |       827 |
| osx-plist     |     1,431 |
| psvn          |     4,116 |
| quack         |    10,675 |
| vkill         |    15,077 |
| x-dict        |       870 |
| xterm-frobs   |     1,067 |
| xterm-title   |       748 |

### DIFF
--- a/recipes/escreen
+++ b/recipes/escreen
@@ -1,1 +1,0 @@
-(escreen :repo "emacsmirror/escreen" :fetcher github)
--- a/recipes/fm
+++ b/recipes/fm
@@ -1,1 +1,0 @@
-(fm :fetcher github :repo "emacsmirror/fm")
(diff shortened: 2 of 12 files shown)


milkypostman:
Looks fine to me.

PR #5411, opened 2018-04-05
Add clush-mode

alexvorobiev:

Emacs mode for interactive clush sessions. clush is a tool from ClusterShell framework (http://clustershell.readthedocs.io/en/latest/index.html).

https://github.com/alexvorobiev/clush-mode

Author

None needed

### Checklist

Please confirm with `x`:


recipes/clush-mode
```diff
@@ -0,0 +1 @@
+(clush-mode :fetcher github :repo "alexvorobiev/clush-mode")
```
DamienCassou: please add a new-line character at the end of the file

DamienCassou:
Please use the linters referenced in the CONTRIBUTING.md file. You will get at least 8 warnings to fix. 

Some more feedback:

- use `set-keymap-parent` and `make-composed-keymap` instead of `nconc`
- you can use `kbd` to specify keys to add to maps
- I think you should not use backquote+quote on non-symbols in docstrings

    > e.g. "-w host" or "-g group"
- why do you use `clush-mode-hook` at the end instead of inlining `clush--initialize` inside `clush-mode`?

purcell:
Good feedback, @DamienCassou. Couple of notes from me:

- Agree about inlining `clush--initialize`: if proper behaviour depends on that code running, it shouldn't be placed in a hook where user customization might remove it.
- It's worth adding an `;;;###autoload` cookie comment for the `clush` command.
- You're using `s` only for `s-join`, so you might consider `(mapconcat 'identity args " ")` instead.

Looks good, though, @alexvorobiev! A quick check with `package-lint` and a couple of little edits and it's ready to go. :-)

tarsius:
I had another look and the first suggestion has not been addressed yet. I didn't look further. I am closing for now. Please let us know when things are ready so that we can reopen and do another review.

PR #5413, opened 2018-04-06
Add intel-hex-mode recipe

mschuldt:

Mode for editing Intel hex files (http://en.wikipedia.org/wiki/Intel_HEX)

https://github.com/mschuldt/intel-hex-mode

Maintainer

This mode was authored by Rubens Ramos <rubensr AT users.sourceforge.net>, I contacted him via email.
He has blessed this Melpa submittal and passed the maintainership on to me.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/intel-hex-mode
+++ b/recipes/intel-hex-mode
@@ -1,0 +1,2 @@
+(intel-hex-mode :fetcher github
+                :repo "mschuldt/intel-hex-mode")


DamienCassou:
Some feedback:

- please use the linters in CONTRIBUTING.md to fix your code. Linters would have tell you:
  - `intex-hex-menu` should be declared before being used
  - the package should declare `easy-menu-add` to avoid a warning. You can try this (untested): `(declare-function easy-menu-add "ext:easy-menu")`. Nevertheless, I think you could just `require` easymenu (which has been in Emacs since version 19, released in 1992 :-)). That way, you could put the menu directly in the variable and avoid the `nil`-check in the mode definition
- you can use `featurep` to detect for the presence of `easymenu` (but again I suggest simply assuming it is there)
- major modes should be defined with `define-derived-mode`, not with `defun`.
- there are two different version numbers in the file: 0.1.3 and 0.1.2
- you may want to use `kbd` to define keys: e.g., `(kbd "C-c C-c") instead of `"\C-c\C-c"`
- the [Emacs Lisp style guide](https://github.com/bbatsov/emacs-lisp-style-guide) recommends placing "all trailing parentheses on a single line instead of distinct lines" (there are some alone parenthesis where the menu content is defined).
- there are some [magit numbers](https://en.wikipedia.org/wiki/Magic_number_(programming)) in `intel-hex-calculate-line-checksum`. Consider giving them names to facilitate code understanding. You can additionally, extract parts of this function to sub-functions: their names will help understanding the main function.
- `(goto-char 0)` should probably be `(goto-char (point-min))` instead as the latter takes care of narrowing. If you really want the whole buffer, you might want to use `save-restriction` and `widen`.
- in `intel-hex-get-segment-base`, you can reverse the two last parameters to `if` to avoid the `progn`. You should get the main code less indented.

PR #5414, opened 2018-04-07
Add hmac recipe

grimnebulin:

Implements a Hash-based Message Authentication Code around Emacs's built-in secure-hash function.

https://github.com/grimnebulin/emacs-hmac

Maintainer.

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/hmac
+++ b/recipes/hmac
@@ -1,0 +1,1 @@
+(hmac :fetcher github :repo "grimnebulin/emacs-hmac")


DamienCassou:
Some feedback:

  - you miss a `;;; Code:` section in the header
  - the first line of `hmac` docstring is more than 80 characters long
  - your `if-let` is broken, it needs an additional pair of parenthesis around the variable bindings (more precisely, you use a deprecated syntax)
- the package documentation could say a little more about why the wrapper is useful and what it does precisely. Please consider contributing to emacs core.

grimnebulin:
I did run the linters.  I don't personally like the `;;; Code:` section because it's perfectly obvious where the code is.  The first line of the `hmac` docstring is 79 characters long, including the newline, and the linter erroneously includes the docstring indentation in its character count.  But since you insist, I've made those changes.

The `if-let` is certainly not broken; it works fine.  I can't find any indication that the syntax is deprecated.  It is explicitly allowed as a special case in my local Emacs, built a few months ago, and also in the current Emacs master.  It is also not deprecated in both the earliest and latest tagged versions of Emacs 25 (as specified in my `Package-Requires` line).  Searching for the terms "emacs if-let deprecated" turns up nothing relevant.

DamienCassou:
Sean McAfee <notifications@github.com> writes:
> […] But since you insist, I've made those changes.


I'm very sorry you felt I insisted. In my mind, I just gave you my
personal opinion. I don't think you should change your code in any way
you don't like: after all, you are going to maintain it, not
me. Moreover, I don't even have the admin rights on Melpa to merge your
PR, so I don't decide on anything.


> The `if-let` is certainly not broken; it works fine.  I can't find any
> indication that the syntax is deprecated.  […]

More information in this discussion:
https://lists.gnu.org/archive/html/emacs-devel/2018-03/msg00321.html


-- 
Damien Cassou

purcell:
Thanks @grimnebulin - looks good, merging now. Sorry for the long delay in getting to this.

PR #5415, opened 2018-04-08
Add nodemcu-mode recipe

andrmuel:

nodemcu-mode is a minor mode that provides a wrapper for the popular nodemcu interaction tools nodemcu-tool and nodemcu-uploader. These are useful to upload Lua files to ESP8266 devices with NodeMCU firmware.

https://github.com/andrmuel/nodemcu-mode

I am the author.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/nodemcu-mode
+++ b/recipes/nodemcu-mode
@@ -1,0 +1,2 @@
+(nodemcu-mode :repo "andrmuel/nodemcu-mode"
+              :fetcher github)


andrmuel:
Hello Damien

Thank you for the feedback

> avoid merge commits in your branch. Instead, rebase your branch on top of latest master
> CONTRIBUTING.md asks that you open one PR for each new recipe. Please do.
> your workflow could probably be simplified if you used a specific branch for each PR instead of using master

The PR only contains one new recipe. However, I forgot to pull before making the new commit, which also caused the merge commit. Should I just open a clean new PR from a separate branch?

> you don't have to specify :package-version for your defcustoms. Especially while your package is still at version 0.x.

OK, removed.

> the defaut value nil for nodemcu-port, nodemcu-baudrate and nodemcu-default-keybindings is not compatible with the specified type. Please add nil as accepted value or please specify a compatible default

Added nil to choice for `nodemcu-port` and `nodemcu-baudrate`. Should I really do this for `nodemcu-default-keybindings` as well? The documentation says boolean is either nil or t.

> in the docstring of the mode, there are 2 consecutive verbs: "allows provides"

Thanks, fixed.

> use sharp-quote to reference functions by name

Fixed.

> you could simplify your code a bit by extracting the pcase nodemcu-backend into a macro

Could you elaborate a bit more how that should look? Would I use the macro just for cases where I only call `nodemcu--run-command` and still use a `pcase` otherwise?

DamienCassou:
> Should I just open a clean new PR from a separate branch?

you don't have to do that if you use `git rebase` to rewrite your commit history so it only contains 1 commit. If  you can't make that work, just create a dedicated branch and open a new PR.

> Should I really do this for `nodemcu-default-keybindings` as well?

no, you are right

> Could you elaborate a bit more how that should look? Would I use the macro just for cases where I only call nodemcu--run-command and still use a pcase otherwise?

I think I would at least wrap the `pcase` that way (or similar):

```emacs
(defun nodemcu-format-device ()
  "Format the NodeMCU device.
This will delete all files, but not the NodeMCU firmware itself."
  (interactive)
  (nodemcu--run-command
    (nodemcu-backend-switch
      (tool "mkfs --noninteractive")
      (uploader "file format"))))
```

Even though I have no authority over what gets accepted into melpa, I believe you really don't have to do that to get your package accepted. It's just a suggestion, please only follow it if you like it.

andrmuel:
I wasn't aware GitHub would be OK with me rewriting the history of the target of an open pull request, but it seems to have worked; thank you for the hint.

As for the macro, I guess it's a good chance to get to understand `defmacro` a bit better ..

Is this what you meant?

```emacs
(defmacro nodemcu--backend-switch (&rest cases)
  `(pcase nodemcu-backend
     ,@(mapcar
        (lambda (c)
          (list (concat "nodemcu-" (symbol-name (car c))) (cadr c)))
        cases)))
```
I can apply that only when all cases are arguments for `nodemcu--run-command` though. Would it make sense to rewrite `nodemcu--run-command` to interpret `nil` as "command not implemented"? Then I could use it for all cases.

DamienCassou:
Andreas Müller <notifications@github.com> writes:
> I wasn't aware GitHub would be OK with me rewriting the history of the target of an open pull request, but it seems to have worked; thank you for the hint.


you are welcome


> Is this what you meant? […]


exactly! Good job.


> I can apply that only when all cases are arguments for
> `nodemcu--run-command` though. Would it make sense to rewrite
> `nodemcu--run-command` to interpret `nil` as "command not
> implemented"? Then I could use it for all cases.


exactly. That way, you don't even have to specify the cases where there
is no command. You could additionally create a new macro using the first
one:

    (defun nodemcu-restart-device ()
      "Restart the NodeMCU device using 'node.restart()' command."
      (interactive)
      (nodemcu--backend-run
        (uploader "node restart")))

Unsupported cases don't have to be written.

-- 
Damien Cassou

andrmuel:
I have now also implemented the second macro and can use the short form as you suggested.

Thanks for all the support!

purcell:
Great - merged now! :-)

PR #5416, opened 2018-04-09
Use a single sed expression in travis build

hjpotter92:
Instead of piping the output from `travis-changed-files` to `grep` and
then another pipe to `sed`; the change achieves the same result using
a single `sed` command.


purcell:
Thanks so much for submitting this change, but it's not quite equivalent, because the revised version would detect `recipes/.dir-locals.el` as a modified recipe if that file had been changed. So I believe the existing version is the correct one. A simple `awk` command could be used to replace the `grep` + `sed` pair, but it's not really a benefit tbh.

PR #5420, opened 2018-04-09
add etop-mode

akacase:

etop allows users to view top and act upon pids via an emacs buffer.

https://github.com/akacase/etop

Maintainer

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/etop-mode
+++ b/recipes/etop-mode
@@ -1,0 +1,1 @@
+(etop-mode :repo "akacase/etop-mode" :fetcher github)


DamienCassou:
Some feedback:

- you might want to explain the difference with the builtin `M-x proced`.
- please run linters on your code (flycheck + flycheck-package). This would have told you about a few errors:
   - missing `Package-requires` line with emacs 24 (because of lexical bindings)
   - first line of docstring for `etop-mode-generate-top-command-function` and `etop-mode-fill-buffer` are not a sentence
    - no docstring for `etop-mode-member-at-least-one`
    - the footer should not have a blank line after the `provide` line
- you might want to add docstrings to your variables
- use `--` after package name prefix for private functions and variables (e.g., `etop-mode--overlay-list`)
- instead of `if COND nil BODY` you can use `when` or `unless`
- I don't understand the `if etop-mode-map` before adding keys to it. It's value should be nil because this is how it is defined the line above. I suggest the more standard following code:

    ```emacs
    (defvar etop-mode-map
      (let ((map (make-sparse-keymap)))
        (define-key map "n" #'etop-mode-next-line)
        ...)))
    ```

akacase:
Thanks for the feedback! I was unaware of M-x proced, so I wrote this. I'm not one for duplication, so I'll close this.

DamienCassou:
- please make first line of docstring for `etop-mode-column-header-regexp` a complete sentence instead of just adding a `.` at the end :-)
- the docstring of `etop-refresh-rate` does not specify the unit (milliseconds, seconds, minutes?)
- please use `define-derived-mode` with parent `special` instead of the home-made function `etop-mode`.

akacase:
I can fix all these issues, I just don't think the package should really exist at this point, since it's duplicating the behavior of M-x proced. I appreciate all the feedback.

DamienCassou:
- in `etop-mode-fill-buffer`, you can pass the result of `get-buffer-create` as first argument of `with-current-buffer` to merge the first two lines
- please let-bind `inhibit-read-only` instead of temporarily setting it with `setq`.

purcell:
> I can fix all these issues, I just don't think the package should really exist at this point, since it's duplicating the behavior of M-x proced. I appreciate all the feedback.

Cool, thanks @akacase!

PR #5423, opened 2018-04-11
Add fontify-face

Fuco1:

Fontify symbols representing faces with that face.

https://github.com/Fuco1/fontify-face

Author

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/fontify-face
+++ b/recipes/fontify-face
@@ -1,0 +1,1 @@
+(fontify-face :fetcher github :repo "Fuco1/fontify-face")


purcell:
Thanks - merging now. You should change the default mode lighter to something textual, though: emoji will just be a pain for many users, and you can change it locally easily enough, particularly if you [define a handy variable for it like this](https://github.com/emacscollective/auto-compile/blob/6ce4255ab9a0b010ef8414c5bd9a6d6d9eea012f/auto-compile.el#L157).

PR #5424, opened 2018-04-12
add recipe for heaven-and-hell

valignatev:

When it's lightly outside, light theme easier on the eyes, and dark theme is better when it's night.
This package allows to quickly switch between dark and light themes. At a single keypress. 

https://github.com/valignatev/heaven-and-hell

I am a creator and maintainer of this package

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/heaven-and-hell
+++ b/recipes/heaven-and-hell
@@ -1,0 +1,5 @@
+(heaven-and-hell
+ :fetcher github
+ :repo "valignatev/heaven-and-hell"
+ :files ("*.el"
+         "README.md"))
--- a/recipes/heaven-and-hell
+++ b/recipes/heaven-and-hell
@@ -1,5 +1,3 @@
 (heaven-and-hell
  :fetcher github
- :repo "valignatev/heaven-and-hell"
- :files ("*.el"
-         "README.md"))
+ :repo "valignatev/heaven-and-hell")


purcell:
Thanks! This looks useful. Sorry for the slow response - here's some quick feedback:

- You should remove `:files` from the recipe. The default is fine, and you shouldn't include README.md inside the Emacs package. Any relevant documentation should go in the `;;; Commentary:` section of the main elisp file.
- `#'wombat` should be `'wombat`. It's not a function.
- `eval` shouldn't be necessary at all.
- "emacs" shouldn't appear in the name of variables or functions you define. It's all emacs code! :-)
- You can use `custom-set-variables` to set and apply the enabled themes more robustly, rather than trying to call `load-theme` with `t` at the right time. Here's some [similar code I've been using for years](https://github.com/purcell/emacs.d/blob/master/lisp/init-themes.el) to do the same thing. Similarly with your reset function, you can just use `custome-set-variables` to set `custom-enabled-themes` to `nil`.
- Emacs supports having multiple themes enabled simultaneously, and your code does not allow for that. Maybe just add/remove the light/dark themes from `custom-enabled-themes` as needed.

valignatev:
Hi @purcell, thanks a LOT! I've learned a big ton about emacs theme system thanks to you. I've tried to improve the quality of `heaven-and-hell` according to your feedback, hope it's good enough now.

P.S. If only npm had that quality bar :smile:

purcell:
Great!

> P.S. If only npm had that quality bar

I don't know if I'd scale to the size of the Javascript community! :-)

purcell:
Anyhow, merged - welcome to MELPA! :tada:

PR #5427, opened 2018-04-15
Add phpstan recipe

zonuexe:

Emacs interface to [PHPStan](https://github.com/phpstan/phpstan), includes checker for Flycheck.

https://github.com/emacs-php/phpstan.el

I'm the author.

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/flycheck-phpstan
+++ b/recipes/flycheck-phpstan
@@ -1,0 +1,2 @@
+(flycheck-phpstan :fetcher github :repo "emacs-php/phpstan.el"
+                  :files ("flycheck-phpstan.el"))
--- a/recipes/phpstan
+++ b/recipes/phpstan
@@ -1,0 +1,1 @@
+(phpstan :fetcher github :repo "emacs-php/phpstan.el")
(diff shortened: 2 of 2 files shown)


purcell:
Thanks! Looks good. A few quick notes:

- You should provide a couple of usage hints in the `Commentary` section, because that will appear on the package page (and because elisp should be self-documenting where possible).
- `(not (null phpstan-replace-path-prefix)` -> `phpstan-replace-path-prefix`
- Don't autoload the [(when (featurep 'flycheck)](https://github.com/emacs-php/phpstan.el/blob/master/phpstan.el#L249) form. It's inappropriate, and won't work reliably anyway. You should really provide a separate `flycheck-phpstan` package which depends properly on `flycheck` and `phpstan`.

zonuexe:
@purcell Thanks for reviewing.  Each point pointed out has modified.

purcell:
That's great! So now what you need to do is add a separate recipe for `phpstan-flycheck`, and add a :files clause to each, e.g. `:files ("phpstan.el")`, because the recipes include `*.el` by default.

zonuexe:
@purcell Thank you, I split these two files as different recipes.

purcell:
Merged now - thanks for your patience!

PR #5428, opened 2018-04-15
Add moom

takaxp:

This package provides a set of commands to control frame position and size.

https://github.com/takaxp/moom

I'm maintainer :)

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/moom
+++ b/recipes/moom
@@ -1,4 +1,1 @@
-(moom :repo "takaxp/moom"
-      :fetcher github
-      :files ("*.el"
-              (:exclude "fig/*.gif")))
+(moom :repo "takaxp/moom" :fetcher github)
--- a/recipes/moom
+++ b/recipes/moom
@@ -1,0 +1,4 @@
+(moom :repo "takaxp/moom"
+      :fetcher github
+      :files ("*.el"
+              (:exclude "fig/*.gif")))


purcell:
Thanks! Quick notes:

- The `:files` spec in the recipe is redundant and should be removed. `*.el` will be included anyway, and `"fig/*.gif"` don't even appear in the repository. Maybe you just copied/pasted this from somewhere else?
- Why `(require 'moom-font nil t)`? The two files will be packaged together, so you could drop the `nil t`.
- Since you're providing a global minor mode, don't suggest that people use `global-set-key`: just tell them to bind keys in `moom-mode-map`.
- You might like to use `read-number` [here](https://github.com/takaxp/moom/blob/374150d588af125394308e710fc6ebadf19a1351/moom.el#L619)

takaxp:
Hi Purcell!
Thank you for your kind comments.

For the first note: I'll change the recipe more simple. 

Basically, I referred the recipe of Helm and let me explain why I did. As you can see, The README.org has many gif files. When I wrote it, the repository actually had the `fig` directory and it held many gif files. The total file size exceeds 180MB!

I run a MELPA repository personally. And I find the cloning process is so slow and working directory exceeds 180MB because of the gif files. I believe this is so bad for the official MELPA repository, so I first tried to use `git submodule` to hide the fig directory from MELPA, but it didn't work well. Secondly, I created another git repository in Github as https://github.com/takaxp/moom-images to just store the huge gif files and link them in README.org under the original repository by referring to `blob/master` in `moom-images`. Now it works fine. The README.org can show all gif images properly and the gif images will not be cloned by MELPA.

If you have another solution to this, please let me know :)

I'll check other comments soon.

Best,
Takaaki

takaxp:
I modified code and README to reflect your comments on `moom-mode-map` and `read-number`. Thanks!

Finally, for the 2nd comment, I intended that some users can use `moom.el` without installing `moom-font.el` by `el-get`,  `use-package`, or downloading directly. So I didn't include `moom-font.el` in `Package-Requires`. But when I byte-compile `moom.el` independently, some warnings were shown. So `nil t` is required to suppress them.

Honestly speaking, I'm still thinking about the dependency between `moom.el` and `moom-font.el`. I need more time to make it more good shape. So please leave this to me :)

Best,
Takaaki

purcell:
That all makes sense, thanks. I'll merge this now, and you can make your own choices about `moom` / `moom-font` later. :-)

takaxp:
Thank you for everything! Good day :)

PR #5431, opened 2018-04-16
Add company-ledger recipe

debanjum:

 Transaction Auto-Completion for Beancount & other Ledger-Likes with Company Mode 

https://github.com/debanjum/company-ledger

Maintainer

**None needed** 

### Checklist

### DIFF
--- a/recipes/company-ledger
+++ b/recipes/company-ledger
@@ -1,0 +1,1 @@
+(company-ledger :fetcher github :repo "debanjum/company-ledger")


purcell:
Thanks! This is interesting to me as a Company and Ledger user, and as one of the `ledger-mode` maintainers.

A few notes, then:

- The completion is just for full postings, so the very broad name `company-ledger` is a little misleading.
- There's actually nothing company-specific about this code as far as I can tell, so you could simply provide a completion-at-point function. This would then magically work with the built-in `company-capf` backend and with `helm` etc. I'd even be happy to include such a completion-at-point function in `ledger-mode` itself.
- The regexes for identifying postings are a little simplistic, and I feel like you might want to use some of the regexes in `ledger-mode` to make this more complete. I've found in my own fiddling with `ledger-mode` that there are lots of quirks in the ledger format.

What do you think?

(P.S. Minor style note: `-backend` shouldn't be included in the backend function's name.)

debanjum:
Thanks for the insightful comments!

- I'm open to changing the name to something less broad. Let me think of something but if you have any ideas let me know
- Using the built-in completion-at-point sounds like a better idea. Let's see if I can make it work without company-mode
- Getting it included into ledger-mode itself would be great! Though providing it as an independent package too would allow using other backends like beancount without depending on ledger-mode to get transaction auto-completion
- Also let me look into the ledger regex's and see what I can pick from it

tarsius:
Any progress on this?

debanjum:
No, wasn't able to replace the package's dependence on company-mode with the vanilla completion-at-point. Hopefully I can have another go at it this weekend.

purcell:
Hey @debanjum, I'll go ahead and close this PR for now, but as/when you have time to fix things up, please feel free to re-open this, or to ask further questions in comments here. I'm also over in the `ledger-mode` repo if you decide you'd prefer to submit a PR directly there instead.

debanjum:
It's been a while but I finally got to working on the feedback for this PR :smile: 
See responses to [this feedback](https://github.com/melpa/melpa/pull/5431#issuecomment-383251973) below:
1. Implementing fuzzy search via Emacs CAPF isn't trivial
    1. The `completion-at-point-function` expects a substring and a candidate list.
    2. It tries to find a maximal prefix match for the substring in the candidate list using [try-completion](https://www.gnu.org/software/emacs/manual/html_node/elisp/Basic-Completion.html#Basic-Completion)
         For example, `(try-completion "hello" '("bye hello")) => nil`
         This model does not work for the fuzzy search implemented by this package.

2. Keeping the package name as company-ledger.
    > The completion is just for full postings, so the very broad name company-ledger is a little misleading.

   The package does more than just posting auto-completion. 
   It handles any paragraph starting with YYYY-MM-DD. 
   This includes accounts, prices and events as well.
   Going forward it'll hopefully do even more!

3. Keeping the regex for completion candidates general instead of transaction specific
    This allows handling other entities apart from just postings as mentioned in previous item
 
4. Removed "-backend" from name of company backend function for company-ledger

*PS: @purcell, how do I reopen this PR? I don't see an option to do that*

riscy:
Here's the latest checklist:

### company-ledger.el ###

- No issues!

- No issues!

package-lint (using version 20200419.406):
```
3 issues found:
42:10: error: You should depend on (emacs "24.3") or the cl-lib package if you need `cl-lib'.
82:3: error: You should depend on (emacs "24.3") or the cl-lib package if you need `cl-case'.
91:6: error: You should depend on (emacs "24.3") or the cl-lib package if you need `cl-remove-if-not'.
```

- company-ledger.el#L57: There is no need to quote lambdas (neither #' nor ')
- company-ledger.el#L65: There is no need to quote lambdas (neither #' nor ')

### Loadability ###

Running #'load on each file:
```
Loading company-ledger.el
Done.
```

### Package ###

- Package-Requires: company "0.8.0"
- Use a LICENSE file that GitHub can detect (e.g. no markup) if possible
- Please add a LICENSE or COPYING file to the repository

debanjum:
Thanks for the [suggestions](https://github.com/melpa/melpa/pull/5431#issuecomment-643670773) @riscy!
Resolved following issues pointed out: 
- [X] Add lexical=binding
- [X] Add `emacs-24` and `cl-lib` to package-requires
- [X] Remove #' from start of lambdas
- [X] Add separate LICENSE file to repository 

Let me know if I missed anything

riscy:
I'll leave this up for a bit longer in case @purcell has more feedback.

In the meantime, regarding the regular expressions -- if you don't want to use any provided in `ledger-mode` (which I'm unfamiliar with), you should probably take a look at them anyway, since they probably have some knowledge baked in.  Here are some quick pointers:
```elisp
"[0-9][0-9][0-9][0-9][\-/][0-9][0-9][\-/][0-9][0-9]"
```
The string will swallow your backslashes, so they won't do anything.  If you want backslashes to appear in the regular expression, you need to double up.  As a result, many nontrivial elisp regular expressions are notoriously ugly and full of backslashes.  Luckily, you don't need to escape the hyphens when they appear at the _beginning_ of the `[character block]`, so you can just remove those backslashes.

Next, if you want to search for n characters in a row, you can use `\\{n\\}`, so this is equivalent to what you have above:
```elisp
"[0-9]\\{4\\}[-/][0-9][0-9][-/][0-9][0-9]"
```

You may also want to look up regular expression syntax for symbol boundaries (`\<`, `\>` I believe) -- for example, what if this string appears somewhere -- <kbd>1928148-12-234982093</kbd> -- should it be a match?

And one small package-lint nag from the latest:

package-lint (using version 20200616.2257):
```
1 issue found:
9:70: warning: An explicit dependency on cl-lib <= 1.0 is not needed on Emacs >= 24.3.
```

debanjum:
Thanks again for the quick replies and the detailed, useful feedback! 

- [X] Fixed the `package-lint` warning (and verified with latest `package-lint` to make sure it doesn't have any more complaints)
- [X] Removed the redundant backslashes from the regex. 
  Didn't add the `\\{4\\}` in the interest of regex readability vs conciseness

  The currently used date time regex is simple and works for the default use case.
  Will update it to handle more complex scenarios as they arise.

riscy:
Okay, good to go.  And welcome to MELPA!

PR #5434, opened 2018-04-16
Add el-fly-indent-mode

jiahaowork:

Indent Emacs Lisp on the fly

https://github.com/jiahaowork/el-fly-indent-mode.el

I am the maintainer.

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/el-fly-indent-mode
+++ b/recipes/el-fly-indent-mode
@@ -1,0 +1,1 @@
+(el-fly-indent-mode :repo "jiahaowork/el-fly-indent-mode.el" :fetcher github)


mattiasb:
How does this compare to [aggressive-indent-mode](https://github.com/Malabarba/aggressive-indent-mode)?

jiahaowork:
@mattiasb Thanks for your reminder. I installed aggressive-indent-mode and tried but I think my package is different from it:

1. You can see on my video demo that mine is keyword aware, so you when you type `(when`, the code below indent as a conditional form. But when you continue to type `(when condition`, it indent as a "then" form.

2. When you type a `(` or a `)`, the code follows will indent accordingly.

aggressive-indent-mode seems not to have these features. I also feel it a bit slower because I use some heuristics to limit the region to indent.
But of course my package only supports Emacs Lisp but aggressive-indent-mode supports more languages.

purcell:
Interesting. I tried this out locally and had to [make some changes](https://github.com/jiahaowork/el-fly-indent-mode.el/pull/1) to get it working conveniently, but then, yes, it's different to `aggressive-indent-mode`. So I'll go ahead and merge it - thanks!

Note that when the code is malformed, the indentation functions can trigger errors, which then interrupts normal editing, so you might like to ignore such errors. (Perhaps I only saw this because I often leave `debug-on-error` enabled, though...)

jiahaowork:
@purcell Thanks! I have merged you commit.

purcell:
Great!

PR #5435, opened 2018-04-17
Add lines-at-once

jiahaowork:

Insert and edit multiple lines at once.

https://github.com/jiahaowork/lines-at-once.el

Maintainer.

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/lines-at-once
+++ b/recipes/lines-at-once
@@ -1,0 +1,1 @@
+(lines-at-once :repo "jiahaowork/lines-at-once.el" :fetcher github)


purcell:
Thanks - this is interesting. I'll go ahead and merge it, but I have a suggestion: rather than requiring the repeat number to be typed into the buffer, just do things the Emacs way and make your expansion command accept a prefix arg. Then you can use a `C-10` or `M-8` prefix to specify the number of repeats. This would be much more natural for most Emacs users, and would also allow users to repeat lines that end with numbers, which your current code will prevent. :-)

jiahaowork:
@purcell Thanks for you comment!
1. For the repeat number thing, yes you are right, I'll try a way to let users customize it. Typing it in the buffer is more natural for me though ^_^ .
2. Actually users CAN repeat lines that end with numbers:
```lines end with numbers 123 2```
will turn into
```
lines end with numbers 123
lines end with numbers 123
```
all the blank at the end of the line will be deleted!

purcell:
> Actually users CAN repeat lines that end with numbers:

Ah yes. Makes sense!

PR #5442, opened 2018-04-21
Add new package netease-music

nicehiro:
You can listen to Netease Music(It's a Chinese music service provider) in Emacs.

https://github.com/nicehiro/netease-music

I'm the maintainer and contributor.

None need.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/netease-music
+++ b/recipes/netease-music
@@ -1,0 +1,1 @@
+(netease-music :repo "nicehiro/netease-music" :fetcher github)


purcell:
Thanks! Looks good. Quick feedback:

- You can delete `for Emacs` [here](https://github.com/nicehiro/netease-music/blob/fd55ceb6d22d3cfc3604f5b1d6dbf3294df40498/netease-music.el#L41) - all Elisp is for Emacs, so this is redundant. :-)
- You need to declare a dependency on the `names` package using a `Package-Requires` header.
- `'music` is not a valid [parent customization group](https://github.com/nicehiro/netease-music/blob/fd55ceb6d22d3cfc3604f5b1d6dbf3294df40498/netease-music.el#L43) - you should use `'multimedia` instead.
- It might make sense to mark one or more of the commands as autoloads. I think that's possible with `names`, but I'm not familiar with the syntax.

nicehiro:
@purcell Thanks for your reply! Your advice is really good and I‘ve accept it.

purcell:
Great, merged! Welcome to MELPA. :tada:

PR #5444, opened 2018-04-21
add recipe proxy-mode

stardiviner:

Toggle proxy (HTTP, Socks, URL.el) inside of Emacs.

https://github.com/stardiviner/proxy-mode

Maintainer

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/proxy-mode
+++ b/recipes/proxy-mode
@@ -1,0 +1,1 @@
+(proxy-mode :fetcher github :repo "stardiviner/proxy-mode")


purcell:
Thanks. I'm sure this will be useful for many people. Quick feedback:

- You can remove `for Emacs` [here](https://github.com/stardiviner/proxy-mode/blob/e62853efd9ff02fe473f8779d48b9566fd70e476/proxy-mode.el#L24) - all elisp is for Emacs, so this is redundant.
- The `proxy-mode-local-only-p` conditionals are probably unnecessary: I think you should define a local minor mode which sets all the variables with `setq-local`, and then use `define-globalized-minor-mode` to create a global version of that mode, which will effectively enable the local minor mode in every buffer.

stardiviner:
Hi, @purcell I updated my code, but I found one problem. `proxy-mode-http-enable` use `(setenv "HTTP_PROXY"  proxy-mode-http-proxy)`. The `setenv` is global. I don't know whether change variable `process-environment` can be buffer locally change `HTTP_PROXY`. And does it work if I change `HTTP_PROXY` in `process-environment`? (I googled, found this variable on Stack Overflow).

BTW, I have an Elisp question, how to distinguish list, alist, plist etc? I use `type-of` on those variable objects, all returns `cons` (like `process-environment). I have Elisp this point. Is there better way?

purcell:
> Hi, @purcell I updated my code, but I found one problem. proxy-mode-http-enable use (setenv "HTTP_PROXY" proxy-mode-http-proxy). The setenv is global. I don't know whether change variable process-environment can be buffer locally change HTTP_PROXY. And does it work if I change HTTP_PROXY in process-environment? (I googled, found this variable on Stack Overflow).

Yes, you can change that variable buffer-locally and it will work fine, but it might be a bit surprising if people modify the environment variables directly. Perhaps the best option is a global mode, but in any case, it shouldn't be necessary to use the `local-only-p` conditionals.

> BTW, I have an Elisp question, how to distinguish list, alist, plist etc? I use type-of on those variable objects, all returns cons (like `process-environment). I have Elisp this point. Is there better way?

They're not really different types, because the contents of `alist`,`plist` etc. just follow certain conventions, and when they are empty, they are all identical. :-)  (`alist`s are particularly convenient because the `let-alist` library has some very useful macros for working with them.)

purcell:
I'll merge this recipe now, though, in any case.

PR #5446, opened 2018-04-22
Add organize-imports-java recipe.

jcs090218:

Mimic Eclipse's Organize Imports functionality.

https://github.com/jcs090218/organize-imports-java

I am the maintainer.

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/com-css-sort
+++ b/recipes/com-css-sort
@@ -1,1 +1,0 @@
-(com-css-sort :repo "jcs090218/com-css-sort" :fetcher github)
--- a/recipes/com-css-sort
+++ b/recipes/com-css-sort
@@ -1,0 +1,1 @@
+(com-css-sort :repo "jcs090218/com-css-sort" :fetcher github)
--- a/recipes/organize-imports-java
+++ b/recipes/organize-imports-java
@@ -1,0 +1,1 @@
+(organize-imports-java :repo "jcs090218@gmail.com" :fetcher github)
--- a/recipes/organize-imports-java
+++ b/recipes/organize-imports-java
@@ -1,1 +1,1 @@
-(organize-imports-java :repo "jcs090218@gmail.com" :fetcher github)
+(organize-imports-java :repo "jcs090218/organize-imports-java" :fetcher github)


purcell:
Thanks! Some quick feedback:

- Don't provide a separate `-pkg.el` file: instead, delete it, and fix up the header of the `.el` file and include any metadata there. Better to have things in just one place.
- The package description in the `-pkg.el` file is too long -- run `package-lint` in the main `.el` to get useful feedback when you've done the step above. You can include longer text in the `Commentary:` section of the header.
- Your code uses functions like `jcs-next-line`, which are not defined anywhere.
- You can write [this sort of pattern](https://github.com/jcs090218/organize-imports-java/blob/master/organize-imports-java.el#L94-L98) more succinctly by using `cl-some` from the `cl-lib` package.
- Don't include [comments in docstrings](https://github.com/jcs090218/organize-imports-java/blob/d9caa45219217e4039e5a9b982e1f59a7b44bcfb/organize-imports-java.el#L134).
- Check the return value of [string-match](https://github.com/jcs090218/organize-imports-java/blob/d9caa45219217e4039e5a9b982e1f59a7b44bcfb/organize-imports-java.el#L155).
- Similarly to above, [cl-duplicates would help here](https://github.com/jcs090218/organize-imports-java/blob/d9caa45219217e4039e5a9b982e1f59a7b44bcfb/organize-imports-java.el#L268).
- Instead of resetting the kill-ring [after modifying it here](https://github.com/jcs090218/organize-imports-java/blob/d9caa45219217e4039e5a9b982e1f59a7b44bcfb/organize-imports-java.el#L470), you can modify it while locally-binding `kill-ring` with a `let` expression.
- Instead of [this](https://github.com/jcs090218/organize-imports-java/blob/d9caa45219217e4039e5a9b982e1f59a7b44bcfb/organize-imports-java.el#L487), consider `(string-match-p "\`^[0-9]\'" c)`.
- Adding `tmp-` to the name of all your local variables makes the code harder to understand, and provides no benefit IMO. Just use clear names.
- Don't use `camelCase` in [argument names](https://github.com/jcs090218/organize-imports-java/blob/d9caa45219217e4039e5a9b982e1f59a7b44bcfb/organize-imports-java.el#L438).
- `progn` is unnecessary [here](https://github.com/jcs090218/organize-imports-java/blob/d9caa45219217e4039e5a9b982e1f59a7b44bcfb/organize-imports-java.el#L418).

Hope that helps!

jcs090218:
Thanks for the feedback. I will like to follow all of your requirements above. I tried to fixed most of the issue you pointed out. I would just have to admit I have issue implementing code you require. 

* You can write this sort of pattern more succinctly by using cl-some from the cl-lib package.
* Check the return value of string-match.
* Similarly to above, cl-duplicates would help here.

Can you explain a bit more so I can fix my code ASAP? Thank you!

purcell:
> Check the return value of string-match. (Do you mean defense programming?)

Yes. You'd typically write:

```el
(when (string-match ...)
  (do-something-with (match-string 1) (match-string 2)))
```

>     Similarly to above, cl-duplicates would help here. (Do you mean cl-remove-duplicates? I had tested this method actually take more performance and I have no idea why...)

Yes, `cl-remove-duplicates`. Interesting that it's noticeably slower for you. Another alternative is `seq-uniq` from the `seq` package.

purcell:
(There's also the built-in `delete-dups` function, which I think is very close to your implementation.)

jcs090218:
Are there other issues? Just wondering.

purcell:
Thanks for your patience. This mostly looks good now - I sent you an important dependency fix in https://github.com/jcs090218/organize-imports-java/pull/1, but I'm assuming you'll merge that, so I'll go ahead and merge this recipe in the meantime. :-)

jcs090218:
@purcell Thank you very much! I just merge your PR.

purcell:
Great!

PR #5448, opened 2018-04-22
Add new package counsel-recentp

pepone:

The package filters the recent files list created by `recentf-mode` and display the recently used git repositories using ivy, then allow to jump to magit status buffer for the selected repository

https://github.com/pepone/counsel-recentp

I'm the author

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/counsel-recentp
+++ b/recipes/counsel-recentp
@@ -1,0 +1,5 @@
+(counsel-recentp :fetcher github
+                 :repo "pepone/counsel-recentp"
+                 :files ("counsel-recentp.el"
+                         "README.md"
+                         "LICENSE"))
--- a/recipes/counsel-recentp
+++ b/recipes/counsel-recentp
@@ -1,5 +1,2 @@
 (counsel-recentp :fetcher github
-                 :repo "pepone/counsel-recentp"
-                 :files ("counsel-recentp.el"
-                         "README.md"
-                         "LICENSE"))
+                 :repo "pepone/counsel-recentp")


DamienCassou:
Some feedback:

1. in the recipe, please don't override `:files`, the default is good enough for you
2. please run linters specified in CONTRIBUTING.md on counsel-recentp.el. They would have told you about:
  - the first line is wrong: it should start with 3 `;`, the file must have the `.el` extension, and the summary line is too long
  - some functions the code use are undeclared: `ivy-read`, `with-ivy-window`, `magit-status-internal`: because the entry-point of your package (`counsel-recentp`) has an autoload, I would just move the `require` lines to the top of the file
3. please add `-*- lexical-binding: t; -*-` at the end of the first line
4. I find naming important. Could you please rename the variable `D` to something more explicit (such as `directory`)?
5. Elisp doctrings usually don't talk abount `True` and prefer `non-nil`. For example, I would have phrased the first docstring: "Non-nil if DIRECTORY has a .git repository."
6. Instead of `concat` on files, please use `expand-file-name`
7. you might want to use `magit-git-repo-p` instead of `counsel-recentp-git-repository-p`
8. To me, "find out" does not say much about what the function is doing and returning. Could you please rephrase? Something like: "Return the git repository containing DIRECTORY, nil if none."
9. `(if COND nil ELSE)` can be rewritten `(unless COND ELSE)` or `(when (not COND) ELSE)`
10. you may want to extract the `(file-name-directory (directory-file-name...))` part into its own function to make its purpose clearer and to avoid code duplication.
11. I think that `counsel-recentp-find-git-repository` is the same as `(locate-dominating-file directory #'counsel-recentp-git-repository-p)` which is probably the same as `(locate-dominating-file directory #'magit-git-repo-p)`
12. if you don't know about it, I recommend you to have a look at [projectile](https://projectile.readthedocs.io/en/latest/) and [counsel-projectile](https://github.com/ericdanan/counsel-projectile) which implement the same as counsel-recentp plus much more.

pepone:
Thanks for the feedback @DamienCassou 

> in the recipe, please don't override :files, the default is good enough for you
Fixed

> please run linters specified in CONTRIBUTING.md on counsel-recentp.el. They would have told you about:

I did but the two errors were a bit puzzling to me, fixed both.

> please add -*- lexical-binding: t; -*- at the end of the first line

done

> I think that counsel-recentp-find-git-repository is the same as (locate-dominating-file directory #'counsel-recentp-git-repository-p) which is probably the same as (locate-dominating-file directory #'magit-git-repo-p)

Yes I didn't knew any this, much simpler now.

> if you don't know about it, I recommend you to have a look at projectile and counsel-projectile which implement the same as counsel-recentp plus much more.

I will check it out, for now I was just trying to get something simple working

Thanks for taking the time to explain thing with so much detail

purcell:
> Yes I didn't knew any this, much simpler now.

Or even `(vc-git-root)`.  :-)

A couple of quick notes from me:

- You shouldn't need to [enable recentf-mode](https://github.com/pepone/counsel-recentp/blob/13f08c5b7cf40cbe4e07fe3c9dc6dccb08cb95ff/counsel-recentp.el#L51) every time you run this command.
- With big lists of files, this will perform a huge number of file/directory lookups.
- Hard-coding `magit-display-buffer-function` to `'magit-display-buffer-fullframe-status-v1` is a very personal choice that removes any control from users. Instead, just make the "open directory" function customisable, and set it to `#'magit-status` by default. 
- `recentp` as a package name and ivy prompt doesn't make much sense IMO. What's a "p"? Perhaps `counsel-magit-recent`? Though note that if you set `magit-repository-directories`, then `C-u M-x magit-status` will give you something very similar to this package.

DamienCassou:
I'm closing this for now. Please reopen when you are ready.

PR #5456, opened 2018-04-28
Support for load-env-vars package

diasjorge:

It lets you load a file with environment variable declarations into your current environment inside emacs

https://github.com/diasjorge/emacs-load-env-vars

Maintainer

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/load-env-vars
+++ b/recipes/load-env-vars
@@ -1,0 +1,1 @@
+(load-env-vars :fetcher github :repo "diasjorge/emacs-load-env-vars")


purcell:
Thanks. Quick feedback:

- `filePath` should be `file-path`
- The prompt `"fEnvironment variables file:"` should end with a space
- You should autoload `load-env-vars-from-file` and `load-env-vars-set-env-vars`
- Consider calling `load-env-vars-from-file` simply `load-env-vars`
- The regexp is too simplistic, and doesn't allow for escaped quote characters. Indeed, such variables will then be silently ignored, which will be hard to debug. :-)
- If you make your functions operate on buffers rather than files internally, they'll be easier to debug.

Generally I'd suggest users who want this functionality should prefer the [direnv](https://github.com/wbolster/emacs-direnv) package.

diasjorge:
Hi, thanks for the code review. I added your suggestions in this branch here: https://github.com/diasjorge/emacs-load-env-vars/tree/improvements

I'll improve the regexp and add a few more cases.
I also added some tests for the package.

One thing I didn't understand is what do you mean by this? 

> If you make your functions operate on buffers rather than files internally, they'll be easier to debug.

purcell:
> One thing I didn't understand is what do you mean by this?
> 
>   >  If you make your functions operate on buffers rather than files internally, they'll be easier to debug.

If you make `load-env-vars-extract-env-vars` operate on a buffer instead of a string, then you can easily test it interactively in an example buffer, and use `M-x re-builder` to experiment with your regexp. You can still wrap it to operate on a file from `load-env-vars-from-file`. That's all. :-)

diasjorge:
Hi, so I've made many improvements to the package, thanks for all your suggestions.

Working with buffers simplified a lot of the code, and I ported a much better regexp from the dotenv gem.

One difference regading direnv is that it will read the .envrc file from the current directory. In my case I have many env files for different environments and load them when I need them.

Do you think this is ready to be merged?

purcell:
Yep, thanks - merged now. Welcome to MELPA! :tada:

PR #5457, opened 2018-04-29
Added rebinder recipe

darkstego:

The package allows rebinding of prefix keys, and allows safe overriding of the prefixes.
So for example if anyone wants to safely move C-c to another key, this package allows
them to do it.

https://github.com/darkstego/rebinder.el

I am the owner and maintainer of the package

**None needed** 

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/rebinder
+++ b/recipes/rebinder
@@ -1,0 +1,1 @@
+(rebinder :repo "darkstego/rebinder.el" :fetcher github)


DamienCassou:
I don't fully understand the purpose of the package nor its implementation. I guess someone else should review the package. Here is some feedback:

- The first word of the `rebinder-minor-mode-key-binding` should probably be "Get" instead of "Gets".
- Some functions could be marked private by separating the package name from the rest of the function name with two dashes. E.g., `rebinder--minor-mode-key-binding`
- Please run checkdoc on the buffer and fix the typos. E.g., "propogating" and "inheretence" 
- key-binding (multiple) inheritance can be done through something like this:

    (set-keymap-parent child (make-composed-keymap parent1-map parent2-map))

- prefix function names with `#'` instead of just `'` to tell the compiler this is a function reference. E.g., `(mapcar #'rebinder-toggle-ctrl map)`

- I would rewrite part of `rebinder-toggle-ctrl` that way to make it a bit clearer (at least to me):

    (let* ((mods (event-modifiers item))
	   (key (event-basic-type item))
           (new-mods (if (member 'control mods)
                         (remove 'control mods)
                       (append mods '(control)))))
      (event-convert-list (append new-mods (list key))))

- you could create a global minor mode that would add `rebinder-override` to `after-change-major-mode-hook` to avoid changing the system when rebinder.el is loaded.

DamienCassou:
I'm closing for now. Please reopen when you are ready.

PR #5458, opened 2018-04-29
Add tff package

gizmomogwai:

The package provides an easy way to switch between files that belong together.
e.g. cpp + hpp
or source-file and accompanying test

https://github.com/gizmomogwai/tff

maintainer?

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/tff
+++ b/recipes/tff
@@ -1,0 +1,1 @@
+(tff :fetcher github :repo "gizmomogwai/tff" :files ("tff.el"))


purcell:
Thanks. `projectile` has this built-in (see `projectile-other-file-alist` and the `projectile-toggle-between-implementation-and-test` command), and there's also a standard Emacs command called `ff-find-other-file`. So I suspect this package is redundant.

gizmomogwai:
Thanks for the reference to those other packages ...

PR #5460, opened 2018-05-03
Add merlin-eldoc recipe

Khady:

Eldoc backend for OCaml and Reasonml. Based on merlin.

https://github.com/khady/merlin-eldoc

I am the creator and maintainer of the package.

None needed

### Checklist

### DIFF
--- a/recipes/merlin-eldoc
+++ b/recipes/merlin-eldoc
@@ -1,0 +1,1 @@
+(merlin-eldoc :fetcher github :repo "khady/merlin-eldoc")


DamienCassou:
Some feedback:

- you can use `…` (1 character) instead of `...` for the `truncate-marker` if you want to
- the docstring of `merlin-eldoc-max-lines` starts with "The final result". What about "The default value..."?
- some docstrings start with the word "Check". I find that a bit confusing because it's not clear what the function actually returns. What about starting with something like "Return non-nil if" instead? Same note for the docstrings starting with `Search`. To me, the verbs you use indicate how the function is implemented but I thing a docstring should instead explain what the function does for the user.
- instead of `(if expr t nil)` you can typically just write `expr` as any non-nil value acts as truthy value.
- instead of `(if (string-match ...) ...)` I would probably use `(string-match-p ...)` to avoid changing the global match-data
- in `-fontify`, the first line produces the string `"<no information>"` which is immediately thrown away and the rest of the function keeps on executing.
- it seems that `-doc-error-messages` is used as a list of regular expressions. You may want to specify that in the variable docstring. Moreover, the `*` at the end of third element will match many quotes, not any character (use `".*"` for that)

DamienCassou:
- please `require` cl-lib and prefix all cl functions (e.g., `find-if`) with `cl-`.
- you may want to sharp-quote function names to let the compiler know that the symbol is a function. E.g.,: `(find-if #'merlin--overlay-pending-error errors)` instead of `(find-if 'merlin...)`
- `-adjust-max-len` could probably be simplified like this (untested):

    ```emacs
    (defun merlin-eldoc--adjust-max-len ()
      "Adjust max len variables based on eldoc settings."
      (setq-local 
       merlin-eldoc--max-lines-type
       (cond
        ((not eldoc-echo-area-use-multiline-p) 1)
        ((or
          (> merlin-eldoc-max-lines-type merlin-eldoc-max-lines)
          (and (integerp merlin-eldoc-max-lines-doc)
               (> merlin-eldoc-max-lines-doc merlin-eldoc-max-lines))) merlin-eldoc-max-lines))))
    ```

- you want to replace `(if cond then)` with `(when cond then)` as it reduces indentation of then `then` part and tell the reader that there is no `else` part to expect.

Khady:
Thanks for the review! This is the first time I am writing some emacs lisp and a package for emacs, so this is very valuable.

I addressed most of your feedback.

> the docstring of merlin-eldoc-max-lines starts with "The final result". What about "The default value..."?

What I'm trying to describe is that the final string displayed by eldoc will not depend only on `merlin-eldoc-max-lines`. The doc is now a bit different. I hope it's clearer.

> -adjust-max-len could probably be simplified like this (untested):

I am keeping a `when` based version for now, I find it clearer. In the long term I'd like to replace this with an assoc list passed to the functions using those configuration value. To avoid global values. It would allow me to write some tests too.

Is it a problem that absolutely needs to be fixed now?

The fixes are in https://github.com/Khady/merlin-eldoc/commit/dabd075c6401266ed2ba2aee21da8f16d8e4e7e7 and https://github.com/Khady/merlin-eldoc/commit/8b1d72fd8b6b5da0b00df9760e5ea584b1ac535c.

Thanks again!

DamienCassou:
Louis Roché <notifications@github.com> writes:
> […] Is it a problem that absolutely needs to be fixed now?


Certainly not. I'm just saying what I think would improve the code
base. You are free to do whatever you want with your source
code. Moreover, I'm not the one deciding which packages get accepted
into Melpa.

-- 
Damien Cassou

tarsius:
> Moreover, I'm not the one deciding which packages get accepted into Melpa.

Maybe @purcell should grant you the right to do so...

purcell:
Hi - thanks for your patience, @Khady. This looks really good now, but there's one remaining blocker: you haven't declared a dependency on Merlin via the `Package-Requires` header. Once that's done, I'm happy for any of us to merge this recipe.

Khady:
@purcell I didn't put merlin in the `Package-Requires` header because merlin is not supposed to be installed from emacs the vast majority of the time. Instead it is installed from opam or (reason-cli for reason users). It comes with a folder to add to emcs' load path. Procedure is described here: https://github.com/ocaml/merlin/wiki/emacs-from-scratch

But if you think it's realy better to have merlin in `Package-Requires`, I will add it.

Thank you :)

purcell:
Ah. We've got a package for `merlin`, so presumably it works okay for most users? If so, it would definitely be better to declare the dependency. If it's generally going to break things for users to install `merlin` from MELPA, then that's something we'd like to know about, I guess. Ideal would be that that package (merlin) would correctly detect whether it was compatible with any merlin-related executables or files installed by opam.

Khady:
merlin from MELPA doesn't stop people from using merlin from opam. So I added the requirement.

However merlin from MELPA uses the beta branch of merlin, which doesn't match a published version of merlin. So I'm a bit surprised. I asked the author about it.

Khady:
merlin authors expect people not to use the version from MELPA. The beta branch is a trick to try to limit the number of breakages for people who are using the MELPA version anyway.

Please let me know if it's ok with the current version (merlin 3.0 in Package-requires field) or if I should change something. I don't care much, as people who follow the installation procedure advertised by merlin won't be affected anyway, and it should still work for people using the MELPA version.

purcell:
Sounds good - merged! Welcome to MELPA. :tada:

PR #5461, opened 2018-05-03
add the nosearch file to yasnippet-snippets

AndreaCrotti:
Small change to include the `.nosearch` file, which if not included it's a bit of an issue.

I'm not 100% sure this is the right thing to do, but definitively that file is stripped out, and in this way it would not being stripped out, see this issue for reference:
https://github.com/AndreaCrotti/yasnippet-snippets/issues/273


purcell:
Makes sense, thanks.

PR #5464, opened 2018-05-04
Add com-css-sort recipe.

jcs090218:

Common way of sorting the CSS attributes.

https://github.com/jcs090218/com-css-sort

I am the maintainer.

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/com-css-sort
+++ b/recipes/com-css-sort
@@ -1,0 +1,1 @@
+(com-css-sort :repo "jcs090218/com-css-sort" :fetcher github)


DamienCassou:
Some feedback:

- in the README, you might want to use `kbd` to specify easier to read keybindings: e.g., `(define-key css-mode-map (kbd "C-k s") 'com-css-sort-attributes-block)`
- prefix function names with #' instead of just ' to tell the compiler this is a function reference. E.g., `(define-key css-mode-map "\C-ks" #'com-css-sort-attributes-block)`
- instead of an animated GIF that makes it hard to compare before/after, I suggest having 2 side-by-side screenshots
- "screen shot" is typically spelled "screenshot" instead
- the "ccs" extension of sorted keys is a bit confusing regarding the "css" files to be sorted. I suggest finding another extension unless this one is already well-known and used by other projects
- the package description "a simple package" should rapidly describe what the package does
- you don't need `cl-lib` in `Package-Requires` because `cl-lib` is included in Emacs 24.3 and your package requires 24.4.
- please prefix all cl functions with `cl-`. E.g., `cl-rotatef` instead of `rotatef`.
- please `require` subr-x at the beginning of the file, you need it for `string-trim`
- you might want to define `com-css-sort-sort-type` as a `defcustom` instead of a `defvar` and change the type from number to symbol (e.g., `'type-sort` and `'lexicographic-sort`)
- please consider `defcustom` for all user-configurable variables
- to detect the root of a project, you might to use `(project-current)` instead of your custom code
- you can usually replace `(equal anything t) by just `anything`. 
- we usually write docstrings like "Return non-nil if ..." instead of "Return true if ..."
- the package has very long functions, please consider decomposing them in smaller ones. This will make your code easier to understand step by step, easier to review and also easier to test
- the package seems to implement a sorting algorithm. You can probably just use `sort` with a custom predicate. Please ask if you have troubles doing that (and explain exactly what you want to do).
- I have no time to read the whole file, I'm sorry

jcs090218:
Okay, I fixed all of them except one point
* the package seems to implement a sorting algorithm. You can probably just use sort with a custom predicate. Please ask if you have troubles doing that (and explain exactly what you want to do).

The idea of sorting algorithm is pretty simple. I sort attributes depends on the given list. The order of the attribute sorted is the same order as the list given/pass in.

DamienCassou:
> The idea of sorting algorithm is pretty simple. I sort attributes depends on the given list. The order of the attribute sorted is the same order as the list given/pass in.

exactly. I'm quite sure you can use `sort` for that. Here is something close I think:

```emacs
(defun my-sort (to-sort sorted)
  "Return a copy of TO-SORT with elements sorted according to SORTED."
  (sort
   (seq-copy to-sort)
   (lambda (elem1 elem2)
     (< (seq-position sorted elem1)
        (seq-position sorted elem2)))))

(my-sort '(4 f a 3 b 1 5) '(1 2 3 4 5 6 a b c d e f))
```

jcs090218:
@DamienCassou Thank you for your feedback. But I tried the 'sort' function and tried several predicate. Weird thing is the logic is correct but it does not give me the same result as my own implementation...

DamienCassou:
Jen-Chieh Shen <notifications@github.com> writes:
> @DamienCassou Thank you for your feedback. But I tried the 'sort'
> function and tried several predicate. Weird thing is the logic is
> correct but it does not give me the same result as my own
> implementation...

are the results wrong or just different? If wrong, we can try to fix the
predicate together. 

-- 
Damien Cassou

jcs090218:
@DamienCassou I tried your code and the code below. It pop out `Wrong type argument: number-or-marker-p, nil`. In my original code logic, I figure out the each attribute's position in the sorted list (Line 315-337). Then I use bubble sort with position index/id which I just found to sort the attribute list (Line 339-376). I just realized that I am actually using the 'pair' data structure here. I tried several predicate, but I am really bad at lisp right now.... (I am still trying though)

```
(setq return-line-list
      (sort (seq-copy return-line-list)
            (lambda (elem1 elem2)
               (< (cl-position elem1 real-sort-list :test 'string=)
                   (cl-position elem2 real-sort-list :test 'string=)))))
    
```

DamienCassou:
The `defcustom`s following a `defgroup` don't have to specify the `:group` because the group is inferred

I'm merging your PR now. Thank you.

PR #5466, opened 2018-05-04
Add some of my small packages

matthewbauer:

- em-dired: treat Eshell buffers like Dired where each ```cd``` brings up a new buffer
- set-defaults: provides an alternative to ```custom-set-variables``` for Emacs distributions where ```standard-value``` is set to your new value
- nethack: provides a major mode for playing nethack in term buffers
- macho-mode: an equivalent to elf-mode for Mach-O executables (available on macOS)

https://github.com/matthewbauer/bauer

Maintainer

None needed

### Notes

- ```set-defaults.el``` defines two functions unprefixed: ```set-envs``` and ```set-paths```. Not sure if that is a deal breaker. They are very useful in my config.
- ```set-defaults.el``` also has to add advice around ```custom-declare-variable``` so that it avoids overwriting config values when a package is loaded. I could make it run through a function but currently it's small enough to work for me.

Anyway, if ```set-defaults.el``` is a deal breaker, I can drop it for now and include it in another PR.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/em-dired
+++ b/recipes/em-dired
@@ -1,1 +1,1 @@
-(nix-update :repo "matthewbauer/bauer" :fetcher github :files ("site-lisp/em-dired.el"))
+(em-dired :repo "matthewbauer/bauer" :fetcher github :files ("site-lisp/em-dired.el"))
--- a/recipes/em-dired
+++ b/recipes/em-dired
@@ -1,0 +1,1 @@
+(nix-update :repo "matthewbauer/bauer" :fetcher github :files ("site-lisp/em-dired.el"))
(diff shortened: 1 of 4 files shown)


tarsius:
Each of these recipes claims it is for the `nix-update` package, whose recipe you probably copy-pasted.

tarsius:
Also please consider creating a separate repository for each of your packages. Having completely unrelated things in a single repository is inconvenient for people who would like to contribute to only one of them or who install a package by cloning its repository.

DamienCassou:
I'm closing this for now. Please reopen when you are ready.

PR #5468, opened 2018-05-06
Add powerthesaurus recipe.

SavchenkoValeriy:
This is my first Emacs package, so any comments are highly appreciated.

This package is an integration with [powerthesaurus.org](https://www.powerthesaurus.org). It helps to look up a word in powerthesaurus and either replace or insert selected option in the buffer (depending on the current selection).

https://github.com/SavchenkoValeriy/emacs-powerthesaurus

### Checklist

- [x] `M-x checkdoc` is mostly happy with my docstrings, it has one false positive (mixes a noun with a verb)

### DIFF
--- a/recipes/powerthesaurus
+++ b/recipes/powerthesaurus
@@ -1,0 +1,3 @@
+(powerthesaurus
+ :repo "SavchenkoValeriy/emacs-powerthesaurus"
+ :fetcher github)


DamienCassou:
Please run linters specified in CONTRIBUTING.md. They would have told you about 46 issues.

SavchenkoValeriy:
Sorry, I guess I forgot to update package-lint. I fixed all issues in: [e54a2fe](https://github.com/SavchenkoValeriy/emacs-powerthesaurus/commit/e54a2feac64cf7b622aef401be6e743af61b9f44)

DamienCassou:
Thanks for taking care of this. I like your coding style: small well-documented functions. Thanks. Some non-mandatory feedback now:

- why do you use `s-format` instead of just `format`. It looks like `s-format` makes the expression looks complex even though what you want seems simple: `(format "https://www.powerthesaurus.org/%s/synonyms" word)`
- pay attention that request.el is un-maintained. As you don't use much of its features, you might be better served by `url-retrieve` (I can help if you want to do that but don't know how).
- you can probably remove dependency dash.el by replacing `-rpartial` with a lambda and `-filter` with `seq-filter` or `cl-remove`.
- You can replace `(lambda (x) x)` with `#'identity`.
- you could pass `synonym` directly to `replace-with` and `insert-word` so you don't have to duplicate the call to `pick-synonym`. 
- `replace-with` and `insert-word` are so simple that you can probably just delete them and `choose-callback`  and put the equivalent code in the callback to `request :success`.
- passing `t` to `completing-read` in `pick-synonym` might not be a good idea: it prevents the user from selecting something else that what's on the list. Imagine that, by reading the list of synonyms, the user decides to select a slightly modified version of one of the choices.
- having a field `word` in a struct `word` makes code reading harder than necessary. What is the difference between a word and a word's word? I suggest calling the field "name" or "string".
- instead of `(lambda (x) (word-word x)))`, you can write `#'word-word`. You have the same pattern with `dom-text`
- if you don't need the field `rating`, you might as well delete it
- in `find-good-match`, if you only care about the first non-nil element, you can use `seq-find` or `cl-find-if-not`.

Good job.

I'm merging this now. Feel free to take into account whatever you want.

SavchenkoValeriy:
@DamienCassou, thank you for your great feedback.
I fixed most of the issues that you commented on.

I indeed don't know much about `url-retrieve` and any help is appreciated. My only concern is the readability of this function.

PR #5469, opened 2018-05-08
Add a recipe for dired-recent

Vifon:

A history of paths visited with Emacs dired.

https://github.com/vifon/dired-recent.el

Maintainer.

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/dired-recent
+++ b/recipes/dired-recent
@@ -1,0 +1,3 @@
+(dired-recent
+ :fetcher github
+ :repo "vifon/dired-recent.el")


DamienCassou:
Some feedback:

  - you need a Homepacke or URL header
  - there should be two spaces after periods in comments
  - Global minor modes must `:require` their defining file
  - you miss the `dash` dependency for `-take` (I suggest rewriting your code not to require `dash`)

Vifon:
I missed `flycheck-package`, thanks. Which linter did you use for the homepage, `:require` and the dependencies? Neither `flycheck-package` nor `package-lint` I used before report this.

DamienCassou:
Instead of re-implementing recentf, you might want to customize it so it remembers directories too. For quite a long time, I had the following in my init.el:

```emacs
(defun recentf-track-dired-buffers ()
      "I want the dired buffers to be tracked by recentf"
      (let ((cur-dir-no-slash (substring ; removes trailing slash
                               (expand-file-name default-directory)
                               0 -1)))
        (when (and (file-directory-p cur-dir-no-slash)
                   ;; because of recentf limitations,
                   ;; - we can't store the filesystem root
                   (not (zerop (length cur-dir-no-slash)))
                   ;; - we can't store a TRAMP root
                   (not (string-equal ":" (substring cur-dir-no-slash -1)))
                   ;; And I prefer not storing TRAMP files
                   (or (not (featurep 'tramp))
                       (not (tramp-tramp-file-p cur-dir-no-slash))))
          ;; recentf does not play well with file ending with a slash
          (recentf-add-file cur-dir-no-slash))))
    (add-hook 'dired-mode-hook 'recentf-track-dired-buffers t)
```

DamienCassou:
Wojciech Siewierski <notifications@github.com> writes:
> Thanks for all the hints!


you are welcome. I think it's important we help each other.


> Do I still need `:group` in my minor mode?


I don't think so. The documentation of define-minor mode says:

    :group GROUP Custom group name to use in all generated defcustom forms.
		 Defaults to MODE without the possible trailing "-mode".
		 Don't use this default group name unless you have written a
		 defgroup to define that group properly.

-- 
Damien Cassou

Vifon:
I think that's all.

DamienCassou:
Thanks

Issue #5449, opened 2018-04-23
Recipes for Debian emacs-goodies-el?


DarwinAwardWinner:
There are a couple of useful tools in the Debian package emacs-goodies-el that might be of interest outside Debian systems. For example, R packages use the Debian control file format, which has a major mode in this package. Hence, it would be nice to have MELPA packages for some of these elisp files. I don't know the details, but at some point I guess Debian migrated to git, which means that there is now an official SCM repo for this package here: https://salsa.debian.org/debian/emacs-goodies-el. Hence, the criteria for making a MELPA recipe are technicaly satisfied.

However, things might be complicated by the fact that the debian package build process involves applying a series of patches (see the `debian/patches` subdirectory of the repo). I don't know if these patches are fixes specific to making these packages work in Debian, whether they are always meant to be applied, or whether it is a mix of the two. In any case, I don't think there's a mechanism for applying patches in the MELPA build process.

So, do you think it's worth looking into making MELPA packages for some of the files in emacs-goodies-el, or is it not practical?

purcell:
Interesting. If there are well-formed elisp packages there which have no other upstream, this would be an acceptable source for them. But yes, we wouldn't be able to apply patches, so you'd have to look at what happens to specific files. Where patches are getting applied by the debian build process, presumably those patches are debian-specific and/or there is still an upstream which can't/won't accept those patches directly.

DarwinAwardWinner:
So I got some more time to look into it and the main set of scripts that are potentially useful and have no other upstream are the ones in [dpkg-dev-el](https://salsa.debian.org/debian/emacs-goodies-el/tree/master/elisp/dpkg-dev-el). There don't appear to be any patches affecting these files, as one would expect since this is already the upstream source for them. I've written recipes for the ones I want, and it's as trivial as you would hope: https://github.com/DarwinAwardWinner/dotemacs/blob/master/config.org#dpkg-dev-el

If you'd like, I can submit a pull request with a recipe for each of the scripts in this directory of the repo.

purcell:
> If you'd like, I can submit a pull request with a recipe for each of the scripts in this directory of the repo.

Yup, sounds good, thanks.

DarwinAwardWinner:
Ok, package-lint has some complaints, which may or may not be worth the effort to address, since I have no idea how receptive the Debian maintainers will be toward changes whose only purpose is to accommodate a non-dpkg packaging system. 

[...]

PR #5471, opened 2018-05-08
Add readme-debian recipe

DarwinAwardWinner:

a simple mode for README.Debian files 

https://salsa.debian.org/debian/emacs-goodies-el/

User

None needed

### Checklist

Please confirm with `x`:

Checkdoc has lots of complaints, but I don't know how interested they are in accepting doc fixes to a 15-year-old elisp file.

### DIFF
--- a/recipes/readme-debian
+++ b/recipes/readme-debian
@@ -1,0 +1,2 @@
+(readme-debian :fetcher git :url "https://salsa.debian.org/debian/emacs-goodies-el.git"
+               :files ("elisp/dpkg-dev-el/readme-debian.el"))


purcell:
Here's where I get a little -1 on the state of these packages, since this file will not load unless `debian-changelog-mode` can be loaded, and that's a separate package that this does not declare a dependency on. I suspect lining up declared package dependencies between these files is going to be hard work.

An alternate approach, then, would be to turn all of these files into a single package. They're not consistently prefixed, but they're all inter-related, so I think we could overlook this. We could do this in one of two ways:

- Get the maintainer to add a `debian-maint-pkg.el` file with a `(define-package ...)` form, and make a recipe for `debian-maint` that includes the elisp with a wildcard `:files` expression
- Add a wildcard to the `debian-changelog-mode` recipe such that all the elisp gets bundled sneakily into that package.

What do you think?

purcell:
> In that case, the package name should probably be something like "dpkg-dev-el", unless the package name also needs to be a feature provided by the package.

It doesn't need to match a feature necessarily, but if it doesn't, there'd have to be a `-pkg.el` file matching the package name.

DarwinAwardWinner:
Ok, then I will have a look at the four files and figure out which one makes the most sense to be the "master" recipe that will include the other 3.

purcell:
> Ok, then I will have a look at the four files and figure out which one makes the most sense to be the "master" recipe that will include the other 3.

Great, sounds like the best plan in this situation, unless you can get the maintainer to accept a `-pkg.el` file.

DamienCassou:
What is the status of this PR. Should we close it?

DarwinAwardWinner:
I've been busy, but I'm planning to get to this over the weekend.

aplaice:
An additional issue, orthogonal to the bundling of partially related files, is that since the tags used by upstream are of the form `debian/X.Y(-Z)` (e.g. `debian/38.0` or `debian/26.1-1 ` ([see here](https://salsa.debian.org/emacsen-team/emacs-goodies-el/tags))), we'd need to set a `:version-regexp` in order for the package to be available on Stable MELPA (which is always nice). In this case, `:version-regexp "debian/\\(.*\\)"` would, I think, work (e.g. `(debian-changelog-mode :fetcher [...] :files [...] :url [...] :version-regexp "debian/\\(.*\\)"`)

tarsius:
See https://github.com/melpa/melpa/issues/5449#issuecomment-429630913.

PR #5474, opened 2018-05-08
Added counsel-codesearch recipe.

abingham:

Adds a counsel interface to codesearch.

https://github.com/abingham/emacs-counsel-codesearch

maintainer

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/counsel-codesearch
+++ b/recipes/counsel-codesearch
@@ -1,0 +1,3 @@
+(counsel-codesearch :fetcher github
+                    :repo "abingham/emacs-codesearch"
+                    :files ("counsel-codesearch.el"))
--- a/recipes/counsel-codesearch
+++ b/recipes/counsel-codesearch
@@ -1,3 +1,2 @@
 (counsel-codesearch :fetcher github
-                    :repo "abingham/emacs-codesearch"
-                    :files ("counsel-codesearch.el"))
+                    :repo "abingham/emacs-counsel-codesearch")


DamienCassou:
Some feedback:

- you might want to separate your packages into several git repositories. This will let you version them separately (melpa uses git tags). This would also simplify the README.
- you might want to use `auto-insert` to get a slightly more standard header
- why do you have `'(integer)` as type instead of just `'integer`?
- you can use `rx` to specify easy-to-maintain regular expressions (it's just an information, your regexp is already quite simple)
- instead of `setenv` which changes the environment for the whole Emacs, you might want to pass variable at the beginning of the command: e.g., `(counsel--async-command (format "CSEARCHINDEX=%s %s -n %s" ...)`
- you could use `save-match-data` around `string-math` and `match-string` to avoid changing the global match-data
- you might want to `widen` the buffer after `find-file` and before `goto-char` because the file might already be opened in a narrowed buffer.
- prefix function names with #' instead of just ' to tell the compiler this is a function reference. E.g., `(ivy-read ... :action #'counsel-codesearch--handle-selection)`

abingham:
Thanks for the review. I'll do most or all of this and update the PR when it's ready.

> instead of `setenv` which changes the environment for the whole Emacs

My understanding was that `setenv` only modifies `process-environment`, which I've bound in my `let`. Is this not the case?

abingham:
@DamienCassou I think it's ready for you to look at again. Thanks!

DamienCassou:
Thanks for your work @abingham. Some more feedback:

- to make your file a proper package, please include `Url`, `Package-requires`, and `Version` headers.
- the `provide` should only appear once, at the end of the file
- you need to `require` ivy and add ivy to `Package-requires` too. Same for `counsel`
- you need to add `codesearch` to `Package-requires`

When you are done, make sure you follow the CONTRIBUTING.md file again to build your recipe and to check that the generated-package loads fine in a sandboxed emacs.

> My understanding was that setenv only modifies process-environment, which I've bound in my let.

you are right. You just taught me something! Thanks

abingham:
I've updated it with your latest suggestions.

DamienCassou:
Thanks

PR #5477, opened 2018-05-11
Add see-mode recipe

marcelino-m:

Edit literal string  in a separate buffer similar to org-src-mode

https://github.com/marcelino-m/see-mode

Maintainer

### Checklist

### DIFF
--- a/recipes/see-mode
+++ b/recipes/see-mode
@@ -1,0 +1,1 @@
+(see-mode :repo "marcelino-m/see-mode" :fetcher github)


manuel-uberti:
Is there any significant difference between this package and [string-edit](https://github.com/magnars/string-edit.el)?

marcelino-m:
Probably not.

purcell:
Yeah, it feels like using `string-edit` and modifying `string-edit-at-point-hook` could get you most of the way to the functionality of `see`, so I'm going to decline this package in the interest of minimising duplication in MELPA - sorry! Thanks so much for taking the time to submit it.

marcelino-m:
-  `string-edit` seems to be not being maintained (last commit was 2 year ago) and only support js-mode and it's capable to edit `html` snipet/string only.
-  In Melpa there are many similar packages, maybe it's a bit late to worry about duplicity  and 
   maybe you should let the users to decide at the end.
- Who will implement what is missing in `string-edit` and see-mode already offers?, you?.
- Finally, please! publishes objective rules for acceptance or rejected packages to not think that your decision is just a whim.

**For that's reason I ask you to reconsider your decision.** and let the **users** to decide

purcell:
> and only support js-mode and it's capable to edit html snipet/string only.

Okay, it would have been more helpful if you wrote the above comment earlier, rather than simply writing that `see-mode` is "probably not" different to `string-edit`. :-)

> In Melpa there are many similar packages, maybe it's a bit late to worry about duplicity and
> maybe you should let the users to decide at the end.

Alternatively, maybe we should have started earlier.

> Finally, please! publishes objective rules for acceptance or rejected packages to not think that your decision is just a whim.

Yes, this is a good point. I'm trying to make time to write a maintainer's guide, which would cover these sorts of things. I can appreciate it might seem capricious sometimes - it'll never be 100% objective, though.

Anyhow, in this case I'm fine to re-open and merge this, so I'll do that now.

Thanks!

marcelino-m:
Thanks @purcell  I will be more careful the next time.

PR #5481, opened 2018-05-13
Add phpactor package

zonuexe:

PHP refactoring and introspection tool.
[Phpactor](http://phpactor.github.io/phpactor/) and this package is in development, but several functions are working.

https://github.com/emacs-php/phpactor.el

I'm the author.

None needed.

### Checklist

Please confirm with `x`:


purcell:
Looks good - merging now.

Note that `company-phpactor` doesn't need to be a Company backend at all - why not just provide a `completion-at-point` function within `phpactor.el` itself? Company will then use that via the `company-capf` backend, and completion will also work with Helm and Ivy.

PR #5482, opened 2018-05-13
Transfer php-auto-yasnippets into emacs-php

zonuexe:
[`php-auto-yasnippets`](https://github.com/emacs-php/php-auto-yasnippets) was transferred to [emacs-php](https://github.com/emacs-php).

refs https://github.com/emacs-php/php-suite/issues/1


tarsius:
Thanks!

PR #5487, opened 2018-05-15
Added Wakib-keys recipe

darkstego:

This package provides new basic keybindings for Emacs. 
This covers movement, selection, deletion and CUA keys, as well
as the common prefixes used by emacs.
Its main advantages are

- Better Ergonomics
- Proper CUA support
- Easy to learn
- Access to all Emacs features

https://github.com/darkstego/wakib-keys

I am the owner and maintainer of the package

**None Needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/wakib-keys
+++ b/recipes/wakib-keys
@@ -1,0 +1,3 @@
+(wakib-keys
+ :repo "darkstego/wakib-keys"
+ :fetcher github)


purcell:
Thanks for the submission and your patience! Quick feedback:

- The symbols are all called `wakib-*`, not `wakib-keys-*`, but they should match. Perhaps rename the file, feature and recipe to simply `wakib`?
- You might like to add an autoload cookie comment for the global minor mode definition.
- Turning off the minor mode should restore all keybindings to their default Emacs state, but loading this file will irreversibly change `isearch-mode-map` -- that's not a blocker, but I'd suggest it's worth thinking about how to avoid this.

Looks good otherwise, and great that you're doing this via a proper minor mode, and not just globally changing all the keybindings.

darkstego:
Thanks for the response.

- I did notice the symbol naming might be an issue, but I decided to keep it this way because this minor mode is actually part of a larger [wakib project](https://github.com/darkstego/wakib.d). Which is basically a starter kit that uses these key bindings. Many of the functions that I felt might be useful to others I added to this minor mode so it becomes available to anyone who want to roll their own emacs.d. Initially this minor mode was called wakib to fit the naming scheme, but I felt wakib-keys was a far better name to indicate what this minor mode does, and wakib- was distinct enough to not cause conflict.. 

- I added the autoload cookie comment

- I actually gave the isearch-mode-map a lot of thought. On one hand, I could add these to the starter kit like I did other modules and that would make everything cleaner, but on the other hand, I could not think of a case where a user would want these keybinding and not want to edit isearch in this manner. And isearch is such an essential part of general cursor movement in emacs. It was a clean code vs. user experience tradeoff. I will move it to the starter kit for the time being.

purcell:
Makes sense, thanks: merging now. Welcome to MELPA. :tada:

PR #5488, opened 2018-05-16
Adds exwm-edit recipe

agzam:

Edit mode for EXWM https://github.com/ch11ng/exwm

https://github.com/agzam/exwm-edit

I am the author

None needed 

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/exwm-edit
+++ b/recipes/exwm-edit
@@ -1,0 +1,3 @@
+(exwm-edit
+ :repo "agzam/exwm-edit"
+ :fetcher github)


purcell:
Thanks for the submission. Sorry about the delay. Quick feedback:

- Where does `gpastel` come from? It's not bundled into my Emacs, and there's no `Package-Requires` entry for it.
- Prefer not to [enable the mode as a side-effect of loading the file](https://github.com/agzam/exwm-edit/blob/0b1be717d28e5b87b46fe8e891b31c3e39b908dd/exwm-edit.el#L129).

Looks good otherwise.

agzam:
@purcell Indeed it seems gpastel is not needed. I have removed the dependency. The package works fine without it.

tarsius:
And the mode activation is gone too.

PR #5491, opened 2018-05-18
Add info-beamer

dakra:

This package provides utilities for working with [info-beamer](https://info-beamer.com/opensource).

https://github.com/dakra/info-beamer.el

Maintainer

None needed

### Checklist

### DIFF
--- a/recipes/info-beamer
+++ b/recipes/info-beamer
@@ -1,0 +1,1 @@
+(info-beamer :fetcher github :repo "dakra/info-beamer.el")


dakra:
I have the `osc` package in "Package requires" [here](https://github.com/dakra/info-beamer.el/blob/master/info-beamer.el#L7) but it's actually optional.
You only need it [here](https://github.com/dakra/info-beamer.el/blob/master/info-beamer.el#L169) if you want to send osc packages to your info-beamer node.

Should I remove that dependency or is there a way to
declare optional dependencies?

purcell:
Sorry for the slow turnaround. This looks good, so I'll go ahead and merge it now.



> I have the osc package in "Package requires" here but it's actually optional.
> You only need it here if you want to send osc packages to your info-beamer node.
> 
> Should I remove that dependency or is there a way to
> declare optional dependencies?

There's no way to declare optional dependencies: maybe just drop the dependency and add a note to the `Commentary` about `osc`.

PR #5494, opened 2018-05-19
Add rubocopfmt package

jimeh:

Emacs minor-mode to format Ruby code with RuboCop on save.

https://github.com/jimeh/rubocopfmt.el

I'm am the author and maintainer of the package.

None needed

### Checklist

### Notes

I built the package with `M-x package-build-current-recipe` instead of running `make recipes/rubocopfmt`. Via `make` I kept getting an error about `cl-lib`:

```
$ make recipes/rubocopfmt
 • Building package rubocopfmt ...
Cannot open load file: cl-lib
make: [recipes/rubocopfmt] Error 255 (ignored)
```

However, I got the same error when trying to build `go-mode`:

```
$ make recipes/go-mode
 • Building package go-mode ...
Cannot open load file: cl-lib
make: [recipes/go-mode] Error 255 (ignored)
```

So I assume the issue is not related to my `rubocopfmt` package. Specially since I managed to build and install the `rubocopfmt` package successfully with `M-x package-build-current-recipe`.

### DIFF
--- a/recipes/rubocopfmt
+++ b/recipes/rubocopfmt
@@ -1,0 +1,1 @@
+(rubocopfmt :fetcher github :repo "jimeh/rubocopfmt.el")


purcell:
Thanks! Quick notes:

- Consider using `pcase` here, since the `if` conds are mutually exclusive: https://github.com/jimeh/rubocopfmt.el/blob/34c69c9c923d0da223f7569a6ecc842095adcf85/rubocopfmt.el#L205-L215
- You shouldn't need to check the major mode [here](https://github.com/jimeh/rubocopfmt.el/blob/34c69c9c923d0da223f7569a6ecc842095adcf85/rubocopfmt.el#L281) because you only add that function to the local version of `before-save-hook`.

In any case, this can be merged in the meantime. :-)

PR #5495, opened 2018-05-20
add recipe company-nginx

stardiviner:

Add Nginx directive keywords to company-mode keywords support.

https://github.com/stardiviner/company-nginx

maintainer

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/company-nginx
+++ b/recipes/company-nginx
@@ -1,0 +1,1 @@
+(company-nginx :fetcher github :repo "stardiviner/company-nginx")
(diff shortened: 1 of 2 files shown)


purcell:
This package auto-enables itself, but it shouldn't: the user should be in control of this, so you should remove the `;;;###autoload`ed form, add an `;;;###autoload` for `company-nginx-keywords` and suggest to the user how/when they might like to call it.

stardiviner:
@purcell I see. Thanks

PR #5499, opened 2018-05-22
Add recipe use-ttf.

jcs090218:

use-ttf is an Emacs package for you to use your own .ttf file in Emacs.

https://github.com/jcs090218/use-ttf

I am the maintainer.

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/use-ttf
+++ b/recipes/use-ttf
@@ -1,0 +1,1 @@
+(use-ttf :repo "jcs090218/use-ttf" :fetcher github)


alphapapa:
Hi,

A couple of things I noticed:

1.  You're building shell commands with `concat`.  This isn't safe.
2.  On Windows, you're directly modifying the Registry.  I'm not sure that's a good idea.  Is there a better, safer way to install fonts?
3.  `use-ttf` isn't a very helpful name.  Something like `font-auto-installer` would be more descriptive.

jcs090218:
@alphapapa Hello!

For your questions:
1. I did a bit of google search and I still do not know why... Would you like to explain it why to me? Thank you!
2. Am I aware of, at least the solution is not on top of my head. Apparently, I am not the BATCH master. I would like to do more research/google search if I found any other solution and of course I would update my package as soon as I get the other way out doing the same thing. Furthermore, any solution from anyone are welcome!
3. Thanks for the advice, I always having trouble naming stuff. I will consider about this!

jcs090218:
Thank you for you explanation. I will consider all this in my next push! :D

jcs090218:
Thanks for the feedback. I think I fixed most of the issues.

purcell:
Thanks.

PR #5502, opened 2018-05-23
Switched URL for jabber from git.code.sf.net to github.com

catap:

I've just switched URL for jabber from git.code.sf.net to github.com because
last one looks like passed away.

At github.com I used @legoscia branch that looks like as official place.

I guess Magnus confirm it ;)

P.S. Magnus, hey, how are you? :)

https://github.com/legoscia/emacs-jabber

Old time ago yes, now nope, but I feel that current maintainer can confirm that it should be changed.

None needed

### Checklist

Please confirm with `x`:


purcell:
Makes sense - thanks!

PR #5510, opened 2018-05-24
Add tfsmacs to MELPA.

sebasmonia:

tfsmacs is a package to work with Team Foundation Server from Emacs. This package deals only with the source control aspect, not PBIs, sprints, builds, or any other thing contained in TFS.

https://github.com/sebasmonia/tfsmacs

I'm the maintainer.

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/tfsmacs
+++ b/recipes/tfsmacs
@@ -1,0 +1,1 @@
+(tfsmacs :repo "sebasmonia/tfsmacs" :fetcher github)


purcell:

This looks great overall, but the `tfs-` prefix for the symbols should match the package and feature name, ie. `tfsmacs-`. It looks like you originally called this `tfs`, in which case the prefix would have matched.

You might also like to provide a prefix keymap that users could bind to `C-c t`, in order to easily get all the suggested global keybindings.

sebasmonia:
Thanks for the feedback!
I'll update the names and documentation and re-submit. It will probably take a few days as I'm away from home :)

sebasmonia:
For the key map I'm reading about define-prefix-command, will add this too.

sebasmonia:
Added a prefix keymap and renamed (hopefully) ALL references to the old name.

Appreciate the code review, this is new territory for me :)

DamienCassou:
Some more feedback:
- could you please squash the commits into just one?
- the file header is very unusual. You might consider using M-x auto-insert to get something standard
- you might want to use `(executable-find "tf.cmd")` as default value for tfsmacs-cmd if that works
- there is a lint warning on the first `define-key` line. I don't understand why though. @purcell?
- instead of concatenating strings for the buffer name, Emacs lisp code usually uses `format`
- `string-empty-p` will tell you if a string is empty, no need to compare it to ""
- I don't understand why you add spaces at the end of the results of -get-collection-parameter and -get-login-parameter
- similarly, I don't get why these functions return the empty string. You could just return nil (the default if you use `when`/`unless` instead of `if`) and filter out the nils at the end
- you could use derived-mode-p instead of testing the value of major-mode
- instead of extracting the first caracter of a string with substring to compare it with string-equal, you could use `string-prefix-p`
- in -write-file-to-temp-directory, the code concatenates a list of strings that have been created by exploding a big string (by `process-lines` in -process-command-sync). This seems both costly and useless.
- I don't have time to read the rest of the file

I think this PR could be merged as soon as the commits are squashed.

sebasmonia:
After fighting a lot with git (thanks heavens for Magit!) I managed to merge the commits into a single one.

I will transfer the feedback into an issue in the tfsmacst repo, thank you!

sebasmonia:
Thanks @DamienCassou for the amazing feedback. I made most of the changes except two which I moved to new, separate issues. More details in [the tfsmacs issue](https://github.com/sebasmonia/tfsmacs/issues/1).

tarsius:
@DamienCassou You should use `Rebase and merge` instead of `Create a merge commit`. Also in cases like this where the commit message contains a lot information about things that went wrong on the first attempt (and which was preserved due to a `git commit --squash` without the message being sanitized), it would be better if you stripped that useless information when you "merge" into `master`.

> After fighting a lot with git (thanks heavens for Magit!) I managed to merge the commits into a single one.

I am guessing that you did `r i` and then `f` on all but the first commit. Instead of `f` you should have `s` on all commits and then removed the obsolete information when presented with the concatenations of the old commit messages.

sebasmonia:
Hi Jonas,

I did a mix of command line and Magit commands since until very recently I was more of an hg guy.
So it sort of worked but not sure of the details.

All I can confirm is that I edited the last commit message after the squash.

Will keep the advice in mind for next time!
________________________________

tarsius:
I don't think one can set a default, just which methods are allowed and which are not. But I believe the last used method is sticky per user and repository.

purcell:
Yeah, I typically squash and merge commits in PRs, because it's sometimes kinder for contributors, and the setting is indeed sticky. It's never bad to teach people about rebasing and squashing, though. :-)

PR #5512, opened 2018-05-26
Add recipe for hjertnes/mb.el to MELPA

hjertnes:

My package adds four methods for posting content to micropub enpoints. 

https://github.com/hjertnes/mp.el

Maintainer

No problem

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/mp
+++ b/recipes/mp
@@ -1,0 +1,1 @@
+(mp :repo "hjertnes/mp.el" :fetcher github :files ("mp.el"))


purcell:
Thanks! Sorry it's taken a few days to respond. A few points of feedback, then:

- `mp` really isn't an ideal name: two-character package names are discouraged. Why not simply `micropub`?
- When there's no `else` form, you should prefer `when` to `if` - it makes the code easier to read.
- Empty lines between functions are conventional.
- The `:files` spec in the recipe is unnecessary, and should be removed: the default includes `*.el`, so is fine here.
- The phrase "a simple package" in the package description is redundant, and should be removed.
- The code refers to `micropub-endpoint` and `micropub-token`, but those variables are not declared. They should likely be `defcustom` variables, placed in an appropriate `defgroup`.
- Where you have `(read-string "Link: ")`, you should instead use the `interactive` form to handle this input. ie. add a `link` function variable, and write `(interactive "Link: ")` instead of `(interactive)`.

Hope that helps!

DamienCassou:

PR #5513, opened 2018-05-27
Add recipe for digitalocean api

olymk2:
Commands to interact with digitalocean api v2 in particular starting stoping and creating droplets from with in emacs.

https://github.com/olymk2/emacs-digitalocean

I am the author

None Needed

### Checklist

Please confirm with `x`:

The code byte compiles cleanly with one exception, an unused lexical binding this is given to me from the widgets library but I do not use is there a way to silence that ?

### DIFF
--- a/recipes/digitalocean
+++ b/recipes/digitalocean
@@ -1,0 +1,3 @@
+(digitalocean
+ :repo "olymk2/emacs-digitalocean"
+ :fetcher github)


purcell:
Thanks! This looks useful. A few points of review, mostly stylistic:

- Your use of conditionals doesn't really embrace the nature of lisp. For example,
```el
(if (= (length value) 0)
     nil
   (list value))
```
is just `(when value (list value))`, or - slightly uglier - `(unless (null values) (list value))`. (Note also that `null` is always better than testing `length` against zero, which requires walking the entire linked list, which might be long.)
- Also, prefer `(unless ...)` and `(when ...)` for cases with only one branch, e.g. `(if (not (boundp 'digitalocean-token)) (error "User variable digitalocean-token not set"))` => `(unless (boundp 'digitalocean-token) (error "User variable digitalocean-token not set"))`
- You should `(require 'cl-lib)` explicitly
- Leave empty lines [between functions](https://github.com/olymk2/emacs-digitalocean/blob/master/digitalocean.el#L400)
- You'll probably want to add at least one `;;;###autoload`, for whichever commands a user might be likely to want to use without first explicitly loading the package.

Hope that helps!

olymk2:
Thanks for those tips, should all be addressed now

PR #5515, opened 2018-05-28
Add issue-link

sshaw:

Get the link to JIRA/Tracker/GitHub etc... bugs, features & issues. Turn their IDs into buttons.

https://github.com/sshaw/issue-link

Developer/Maintainer

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/issue-link
+++ b/recipes/issue-link
@@ -1,0 +1,3 @@
+(issue-link
+ :fetcher github
+ :repo "sshaw/issue-link")


dakra:
Looks very similar to `bug-reference-mode` which is built-in to Emacs. Maybe you should describe the differences if your package does something else/better?

sshaw:
> Looks very similar to bug-reference-mode which is built-in to Emacs. 

I hadn't heard of this, but seems like I can leverage it and drop a dependency. 

> Maybe you should describe the differences...

Yes, will do. Adding a link to the buffer is %33 of what it does.

purcell:
Yep, I'm also not clear what this provides over and above `bug-reference-mode`, and we strongly prefer to not host multiple very similar packages where possible, to avoid confusion - it's better to have authors collaborate on established packages unless they're a disaster. :-)

DamienCassou:

PR #5521, opened 2018-06-01
adding logpad.el

dertuxmalwieder:

This minor mode transfers the logging mechanisms of Windows Notepad to GNU Emacs.

https://bitbucket.org/tux_/logpad.el

I am the maintainer of both this one and its predecessor "logpad.vim".

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/logpad
+++ b/recipes/logpad
@@ -1,0 +1,2 @@
+(logpad :fetcher bitbucket
+        :repo "tux_/logpad.el")


DamienCassou:
Some feedback for King Crimson :-):

- the variable `add-newline` has a default value of nil which is described as not adding newlines. This seems counter-intuitive with respect to the variable name
- usually, the user should decide when to activate minor modes: you should probably remove the `add-hook` at the end and write in the README how to activate logpad. In principle, only loading an Emacs lisp file should not change Emacs' behavior
- you should probably provide a global version of the minor mode that would do the magic in every text file starting with .LOG.

dertuxmalwieder:
Thank you for your feedback.

> the variable `add-newline` has a default value of nil which is described as not adding newlines. This seems counter-intuitive with respect to the variable name

If set to "t", it will add additional newlines, but Notepad's default behavior is to not do that. Which variable name would you suggest instead?

> you should probably provide a global version of the minor mode that would do the magic in every text file starting with .LOG.

That's what I thought my code would do. Sorry, this is my first try to write something useful for Emacs. Mind to explain the difference?

dertuxmalwieder:
- [x] Cleaned the `defcustom`s
- [x] Added `-*- lexical-binding: t; -*-`
- [x] Made `(logpad-mode)` entirely optional

DamienCassou:
> If set to "t", it will add additional newlines, but Notepad's default behavior is to not do that. Which variable name would you suggest instead?

I would change the docstring instead of the name. Something like: "When non-nil additional newlines are added for look-and-feel reasons. Set this to nil if you want the original Notepad behavior of not having newlines."

You may want to add an autoload cookie `;;;autoload` before the minor mode definition so users can enable it without first having to require the package.

I'm merging this.

dertuxmalwieder:
Will be considered for an update, thank you!

PR #5522, opened 2018-06-02
added org-listcruncher

dfeich:
org-listcruncher is a planning tool that allows the conversion of an Org mode list to an Org table (a list of lists).  The table can be used by other Org tables or Org code blocks for further calculations

https://github.com/dfeich/org-listcruncher

I am the author and maintainer

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/org-listcruncher
+++ b/recipes/org-listcruncher
@@ -1,0 +1,3 @@
+(org-listcruncher
+ :fetcher github
+ :repo "dfeich/org-listcruncher")


purcell:
Thanks! You're missing a dependency on (and `require` for) the `seq` package, but everything looks good otherwise.

dfeich:
Thanks for spotting that. Just fixed it and committed.

DamienCassou:
Some feedback:

- the dependency to `helm` seems useless
- I guess you could remove `cl-lib` from `Package-requires` as it is included in the Emacs version the package requires (24.4)
- I think you should avoid using `:version` in `defgroup`. The documentation says it means 

    > that the variable was first introduced, or its default value was changed, in Emacs version VERSION.

- usually docstrings are on their own line(s). Please separate `org-listcruncher-parse-fn`'s docstring from the next argument
- similarly, in `defgroup` and `defcustom`, Emacs lisp developers tend to put each key/value pair on its own line. `org-listcruncher-consolidate-fn` has both `:group` and `:type` on the same line
- you don't have to add parenthesis around a `defcustom` type: `:type 'function` is good enough
- if you want to follow standard Emacs Lisp conventions, please read https://github.com/bbatsov/emacs-lisp-style-guide. This includes not putting closing parenthesis alone on a line or after a whitespace
- you may want to define your regular expressions with rx (part of Emacs)

dfeich:
thanks for this very thorough feedback. I resolved the issues:

- the dependency to helm seems useless

  You are completely right. This was a silly copy/paste error.
    
- I guess you could remove cl-lib from Package-requires as it is included in the Emacs version the package requires (24.4)

  Ok. Done
  
- please add -*- lexical-binding: t; -*- at the end of the first line

  Done
  
- I think you should avoid using :version in defgroup. The documentation says it means
   that the variable was first introduced, or its default value was changed, in Emacs version VERSION.

  removed. I thought that one is expected to set this.

- usually docstrings are on their own line(s). Please separate org-listcruncher-parse-fn's docstring from the next argument

  Done
  
- similarly, in defgroup and defcustom, Emacs lisp developers tend to
  put each key/value pair on its own
  line. org-listcruncher-consolidate-fn has both :group and :type on
  the same line

  fixed

- you don't have to add parenthesis around a defcustom type: :type 'function is good enough

  Done
  
- if you want to follow standard Emacs Lisp conventions, please read
  https://github.com/bbatsov/emacs-lisp-style-guide. This includes not
  putting closing parenthesis alone on a line or after a whitespace

  Ok. Need to get accustomed to this. Usually while programming I like to see the
  correctly indented closing bracket for visibility... there you see my old heritage
  as a C/C++ programmer.

- you may want to define your regular expressions with rx (part of Emacs)

  I would like to leave them as they are. But rx looks interesting. I just never
  yet investigated deeply.

DamienCassou:
I merged your PR after adding a new line character at the end of the recipe, rephrasing the commit message and rebasing the PR on top of master.

Is that the right process @purcell?

DamienCassou:
@dfeich next time, could you please create a dedicated branch on your remote instead of using `master`? It will simplify the reviewer's work. Please ask if unsure, I will help.

dfeich:
@DamienCassou  Ok. I think I understand the motivation for branching. I wanted anyhow to study a bit the newer magit features for such workflows. And: Thanks again for the work that you guys from MELPA are doing for the community.

purcell:
> I merged your PR after adding a new line character at the end of the recipe, rephrasing the commit message and rebasing the PR on top of master.
> 
> Is that the right process @purcell?

Yep, that's fine: I'm okay with anything that optimises for our time. I often just squash pull requests on merge, but if they're a bit messy it's a good opportunity to share tips with a contributor.

PR #5523, opened 2018-06-03
Add recipe project-abbrev.

jcs090218:

Customize your own abbreviation expansion in the project.

https://github.com/jcs090218/project-abbrev

I am the maintainer.

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/custom-abbrev
+++ b/recipes/custom-abbrev
@@ -1,1 +1,0 @@
-(custom-abbrev :repo "jcs090218/custom-abbrev" :fetcher github)
--- a/recipes/custom-abbrev
+++ b/recipes/custom-abbrev
@@ -1,0 +1,1 @@
+(custom-abbrev :repo "jcs090218/custom-abbrev" :fetcher github)
--- a/recipes/project-abbrev
+++ b/recipes/project-abbrev
@@ -1,0 +1,1 @@
+(project-abbrev :repo "jcs090218/project-abbrev" :fetcher github)


alphapapa:
Aren't there other packages that do this already and integrate into existing Emacs commands?  For example, here's a file from Oleh Krehel's config that adds abbreviations for Java: https://github.com/abo-abo/oremacs/blob/github/etc/abbrev_defs

jcs090218:
Indeed, I believe there are other packages did this already. But I don't think they all managing it the same way. I would like to have all my abbreviation configuration in the separate file (not bind with Emacs) and in the separate project (Each project each abbrev config). Which I think is good because I don't need to reopen Emacs, plus mine package is cross language. It could be any thing by just having the file in the project root directory.

DamienCassou:
I agree with @alphapapa that there seems to be quite a bit of duplication in the existing packages already. In Melpa, we try to avoid adding packages that we don't believe add enough value compared to what's already in there. The similar packages that come to my mind are:

- abbrev.el included in Emacs (if this is not exactly what you want, I think you could at least base your package on the provided features, e.g., `make-abbrev-table` and `define-abbrev`)
- yasnippet (very powerful and well-known)

jcs090218:
I think the idea of this package is one config file per project and customize abbreviations without restarting Emacs. If I use abbrev.el I do not think I will able to keep the core idea of this package? Can I? In the mean time I will do more research on this.

DamienCassou:
Jen-Chieh Shen <notifications@github.com> writes:
> I think the idea of this package is one config file per project and
> customize abbreviations without restarting Emacs. If I use abbrev.el I
> do not think I will able to keep the core idea of this package? Can I?
> In the mean time I will do more research on this.

I think you could use abbrev for that. Each time a new buffer with
custom-abbrev-mode enabled opens:

1. get the custom-abbrev configuration file from the corresponding
project directory. Use the `project-current` function.

2. create a buffer-local variable and initialize it with
`make-abbrev-table`. Fill the variable by calling `define-abbrev` for
each element of the configuration file

You may want to change the package name to reflect the "project" part of
it. Something like project-abbrev?

-- 
Damien Cassou

jcs090218:
I do not think it will be necessary to create a mode for this package. I know package `emmet-mode` for editing web related files, but I think they store it in the Emacs buffer? I would just like to keep the package simple to read the `custom-abbrev.config` file and do the job. If anyone who do not want this package to be enable in a project can just either ignore the key or just do not  create config file. 

1. I do use the `project-current` function for searching config file on line 130.
2. I do not know if that is an appropriate way of doing the job because I am reading a config file then do the swap when the `custom-abbrev-complete-word` function is called, is two steps and no cache (in purpose). Now if I use `make-abbrev-table` and `define-abbrev` in between, would this be an unnecessary step (It became tree steps) and conflict the design of this package? I would just like this package to be simple, easy to use and let the user have all the control to decide the Emacs behaviours.

I think `custom-abbrev` is a better name than `project-abbrev` because the 'customize' part is more important than the 'project' part. And plus I think people always work in some kind of project so I do not think is necessary to brought up the word 'project' in this scenario. I could consider put my package name to something like `project-custom-abbrev`, but I think is too long and weird to naming it this way.

DamienCassou:
What is your opinion about this @tarsius?

tarsius:
I agree that its always a good idea to base new packages on built-in functionality and popular third-party packages if possible.

I don't think that `custom-abbrev` is a good name because that implies that it has something to do with the Custom interface. Given this name I would assume that provides support for defining abbreviations using the Custom interface.

> And plus I think people always work in some kind of project so I do not think is necessary to brought up the word 'project' in this scenario.

I think this can be used as argument in *favor* of using `project-` in the package name. It helps users who are looking for "packages that make it nicer to work with projects" to actually find this package.

DamienCassou:
Jen-Chieh Shen <notifications@github.com> writes:
> Okay, I think I can just name the package name to `project-abbrev`?

sure

-- 
Damien Cassou

jcs090218:
Done, I just change my package name from 'custom-abbrev' to 'project-abbrev'. Please check if there are other issues. Thank you!

jcs090218:
Done merged! Thank you!

PR #5524, opened 2018-06-03
Adds mw-thesaurus

agzam:

Merriam-Webster Thesaurus

This package allows you to lookup a word and display its thesaurus definition in a nice Org-mode format using merriam-webster api.

https://github.com/agzam/mw-thesaurus.el

I am the author

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/mw-thesaurus
+++ b/recipes/mw-thesaurus
@@ -1,0 +1,3 @@
+(mw-thesaurus
+ :repo "agzam/mw-thesaurus.el"
+ :fetcher github)


alphapapa:
I think you need to ` (require 'dash)`.

alphapapa:
BTW, have you seen these other thesaurus packages?

* synonymous
* synosaurus
* www-synonyms

Not that you can't necessarily make your own, but maybe you could collaborate on one of these other packages and improve it with the source and features your package has.  That way users won't be confronted with as many packages to test and decide on to perform a common task.  :)

agzam:
@alphapapa Yeah, I've tried those and the reason I wrote this one is that I wasn't really (for various) reasons satisfied. As a non-native English speaker I do need to use thesaurus a lot. And it ails me to admit there's sadly no package for Emacs that can absolutely work for everyone. I myself currently have to switch between `sdcv` (where I use project Gutenberg's synonyms dict) and this one all the time, because Merriam-Webster is not exactly comprehensive. But sometimes you don't want too many choices (it becomes noise). Sometimes you need to check thesaurus while offline, etc.
Personally for me probably Google's would work the best but I haven't figure out how to use their API (or if that's possible at all).
So as you can see I made this not because I was bored or something. However I agree with you - it would be nice to have a single, nice package for English thesaurus and dictionary.

purcell:
Thanks for this. Quick notes:

- `[remap org-open-at-point]` and `[remap evil-record-macro]` aren't appropriate as keybindings: you should be specific here, because it's quite possible that neither `org` nor `evil` will have been loaded when this file is loaded.
- You should probably `(require 'cl-lib)` for `cl-function`, just to be explicit.

Otherwise looks good.

-Steve

agzam:
@purcell I added require for org and removed evil specific keybinding:

https://github.com/agzam/mw-thesaurus.el/commit/cb0aae4b6f104c033afd79dddf7a982bd2dcdbd8

Please let me know if this would work. And if you spot anything else suspicious. I'm still learning how to  do these things right.  Thank you.

purcell:
> Please let me know if this would work. And if you spot anything else suspicious. I'm still learning how to do these things right. Thank you.

This would cause `org` to be loaded even if you're using the thesaurus to write text or Markdown. Even better, IMO, would be to instead bind the desired key explicitly, e.g. use `(kbd "C-c C-o")` instead of `[remap org-open-at-point]`.

In any case, this isn't a blocker for merging the recipe, which I'll do right now.

alphapapa:
@agzam Please don't remap Org keys in your package.  Just because a user uses your package doesn't mean he wants to change Org bindings.  Now he'll have to map `org-open-at-point` to another key in his config, or re-remap it on top of your package's rebinding.  And `org-open-at-point` is an important command.

Instead, describe your package's commands in the documentation, and suggest a good default bind with a piece of code to do so (I suggest using `general`).

agzam:
> Please don't remap Org keys in your package

it's remapping only for the minor-mode, when you lookup for a word it opens a buffer, sets major mode to Org (because it's a nice looking outliner), enables minor mode and rebinds RET, so user can lookup further (for definitions in the buffer). This does not affect other Org-mode buffers at all.

alphapapa:
Oh, I'm sorry!  :)

PR #5525, opened 2018-06-04
migrate steckerhalter projects to gitlab

steckerhalter:
I moved my stuff away from microsoft :smile:


steckerhalter:
@tarsius thanks!

Issue #5361, opened 2018-03-09
Strategy for huge repositories


purcell:
To build packages, we clone the repositories that contain them. Sometimes the resulting working directories are enormous, e.g.

- Erlang: 460MB
- Mozc: 940MB
- Nemerle: 600MB
- Ansible: 170MB
- Cmake-mode: 170MB
- Fuel: 260MB
- Sekka: 185MB
- Po-mode: 174MB
- Caml: 200MB

This is a nuisance on the server, though arguably the solution may simply be a larger server. I feel it's worth considering removing _some_ of these packages if they're little-used, e.g. Nemerle, Sekka, Po-Mode. And in other cases, some agitation upstream to split Emacs libraries out from massive repositories might be worthwhile.

I've looked previously at sparse checkouts, and checkouts with limited history, but it conflicted with our need to have the latest change times for any files matched by our recipe wildcards (even if they were from very old commits).

tarsius:
> And in other cases, some agitation upstream to split Emacs libraries out from massive repositories might be worthwhile.

Sounds like a good plan (even for the cases mentioned earlier).

jaseemabid:
@purcell 

1. Is it ok if I make a git repo for things like llvm-mode and copy over the code and release it on melpa? Does it _have_ to be official? 
2. Is it acceptable to maintain a git repo for these libs under the melpa org and consider that safe and official? We will have to keep the code in sync, but I wouldn't expect the llvm mode for example to change that often. I could cherry pick the commits and keep it in sync for some of these. 

3. Can we have one official melpa repo with all these _problematic_ emacs modes grouped together? Do we have a 1-1 mapping b/w emacs packages and git repos or is it flexible?

jaseemabid:
@purcell Can you add llvm-mode to the list please? Makes it easier to find this issue with search engines.

tarsius:
Well it's a list of packages that might be removed - not including packages that have already been removed. Also now that you have mentioned it yourself, adding it again is no longer necessary for that purpose.

tarsius:
> Is it ok if I make a git repo for things like llvm-mode and copy over the code and release it on melpa? [...]

The first step is to contact upstream and try to work out something with them. In this case (and many of the others listed above) this is a huge project and the maintainers probably do not care very much about this non-critical part. So you should probably contact the author of the elisp part instead. That person should then setup a separate repository and ask upstream to remove the bundled copy. 

[...]

PR #5526, opened 2018-06-04
futhark-mode.el was moved to a subdirectory.

Athas:
This recipe is already in MELPA, but upstream moved around.

Maintainer.

### Checklist

Please confirm with `x`:


tarsius:
Please also consider moving the elisp to its own repository. Programming language repositories tend to get rather large and as a result getting the elisp out of `contrib/` can get costly. See #5361. It's also inconvenient for users who install the language implementation using their distribution's package manager but would still like to get the elisp directly from the repository (e.g. to contribute to it).

Athas:
Good point.  I moved the mode to a separate repository and updated the recipe.

purcell:
Thanks, merging now, with a further small change to remove the unnecessary `:files` spec from the recipe.

PR #5527, opened 2018-06-04
Move plain-theme to GitLab

yegortimoshenko:
This pull request changes plain-theme fetcher to point to the new origin at https://gitlab.com/yegortimoshenko/plain-theme.


purcell:
Thanks.

PR #5533, opened 2018-06-07
Recipe: Add frame-purpose

alphapapa:

frame-purpose makes it easy to open purpose-specific frames that only show certain buffers, e.g. by buffers’ major mode, their filename or directory, etc, with custom frame/X-window titles, icons, and other frame parameters.

https://github.com/alphapapa/frame-purpose.el

Author and maintainer.

### Checklist

Please confirm with `x`:

Thanks for your work on MELPA.

### DIFF
--- a/recipes/frame-purpose
+++ b/recipes/frame-purpose
@@ -1,0 +1,1 @@
+(frame-purpose :fetcher github :repo "alphapapa/frame-purpose.el")


DamienCassou:
The packages seem a bit different, but similar enough to deserve some kind of comparison: https://github.com/bmag/emacs-purpose.

alphapapa:
I was confused at first, as I didn't remember an `emacs-purpose` package, but it's called `window-purpose` on MELPA...

Anyway, `window-purpose` seems to be about controlling which buffers are displayed in which window:

> Dedicating a window limits which buffers will be displayed in it. There are two types of window dedication: buffer-dedication and purpose-dedication.

It's also got a lot of other features, more than I can grok at a glance.  :)

`frame-purpose` is about controlling which buffers are available for selection in certain frames.  It's much simpler and doesn't have the features that `window-purpose` has, like restoring window configuration, locking windows to certain buffers, etc.  But `window-purpose` doesn't do what `frame-purpose` does, either.

Ultimately they seem complementary, similar to @akirak's https://github.com/akirak/frame-workflow package.

purcell:
Looks good - merged. Sorry for the really slow response. My only little point of feedback would be that you're using four libraries that provide convenient functions for operating on sequences. I personally feel like `dash` and `dash-functional` are unnecessary in the presence of `seq` and `cl-lib`. :-)

alphapapa:
Thanks, Steve.  No problem, I know there is a lot of reviewing to do, and I appreciate your faithfulness in doing so.  I appreciate the feedback, as always, and I'll think about changing that, however I consider `dash` an essential Elisp library, so I'll probably keep using it.  :)

alphapapa:
Ah, since I only used one function from `seq`, `seq-group-by`, I can use `-group-by` instead and then I don't need to `(require 'seq)`.  Thanks.

I realize this is probably the opposite direction from what you were thinking, haha.  ;)

PR #5536, opened 2018-06-07
Add mu4e-conversation

Ambrevar:

Show a complete mu4e thread in a single buffer

https://notabug.org/Ambrevar/mu4e-conversation

I'm the author.

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/mu4e-conversation
+++ b/recipes/mu4e-conversation
@@ -1,0 +1,1 @@
+(mu4e-conversation :fetcher gitlab :repo "ambrevar/mu4e-conversation")


purcell:
Thanks! Quick notes:

- `mu4e` isn't an installable package, so you should drop the `Package-Requires` entry which refers to it.
- Don't arbitrarily set one colour for a face, e.g. `'((t :background "#553333")`. This will be guaranteed to look awful in many themes. Instead, find built-in faces and `:inherit` them wherever possible, e.g. `outline-1` etc.
- Global toggle functions should often be global minor modes, e.g. `mu4e-conversation-toggle-globally`.

Looks good otherwise!

Ambrevar:
Hi Steve!

I've addressed all issues.

Thanks for the review, and thanks for pointing out the outline-* faces,
it's much smarter indeed!  The global toggle is not really needed so
I've removed it altogether.  Hope it's all good now! :)

Cheers!

Ambrevar:
I've just updated the recipe to reflect the move to gitlab.

purcell:
Great, thanks.

PR #5540, opened 2018-06-10
Remove tests that failed to error

tarsius:
Please see the commit messages and https://github.com/melpa/package-build/pull/16#issuecomment-395940916.


Cask
```diff
@@ -1,4 +1,6 @@
 (source "melpa" "https://melpa.org/packages/")
 
 (development
- (depends-on "cl-lib"))
+ (depends-on "cl-lib")
+ (depends-on "ecukes")
```
DamienCassou: the commit seems to add dependencies, not remove them.

tarsius: That's weird :astonished:

tarsius: This was some history manipulation gone wrong. Looks like I tried to move the removal from the preceding commit to this separate commit but ended up not removing it from the other commit and instead reverting it in the new commit. Fixed.

PR #5546, opened 2018-06-10
Add swift-playground-mode

msanders:

 Emacs support for Swift playgrounds.

https://gitlab.com/michael.sanders/swift-playground-mode

Maintainer

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/swift-playground-mode
+++ b/recipes/swift-playground-mode
@@ -1,0 +1,5 @@
+(swift-playground-mode :repo "michael.sanders/swift-playground-mode"
+                       :fetcher gitlab
+                       :files ("swift-playground-mode.el"
+                               "SwiftPlayground.vim/runner.sh"
+                               "SwiftPlayground.vim/PlaygroundRuntime.swift"))


DamienCassou:
1. Please run linters specified in CONTRIBUTING.md. They would have told you about:

    - by default, 2 spaces must follow each dot to separate sentences
    - there are more than 1 argument to `pkg-info-version-info` in the call to `declare-function`
    - "non-nil" is usually preferred over "true" in docstrings
    - the footer references
    - you miss a `(require 'comint)` for `comint-check-proc`

2. as explained in the docstring of `read-only-mode`, it is recommended to temporarily bind `inhibit-read-only`
3. docstrings usually start with imperative verbs: "Close" instead of "Closes" in `close-buffer`
4. both `apply` and `funcall` can take many arguments, you don't have to put them all in a list first. This should work (untested):

    ```emacs
    (apply 'call-process
           (car command-list)
           infile
           destination
           display
           (cdr command-list)))    
    ```

5. `-call-process-with-output` says it returns the exit status but this seems wrong.
6. instead of `…-is-…`, predicate usually ends with `-p`: e.g., `current-buffer-playground-p`
7. usually, closing parenthesis are not on their own lines. You might want to read and follow  https://github.com/bbatsov/emacs-lisp-style-guide

DamienCassou:

msanders:
Sorry for the delay. I've updated this with the suggested changes.

Regarding the docstrings, I've updated `sentence-end-double-space` to be set to nil in a file-local variable, and `checkdoc` is passing now (which should be shown shortly on Travis).

PR #5547, opened 2018-06-11
Geiser repository moved to Gitlab

jaor:
This is an update of an existing recipe (geiser) with the only change of moving its repository fetcher from github to gitlab.  The reason is that I (geiser's maintainer) am going to discontinue Github hosting, making Gitlab the new home of the project's code.


purcell:
Thanks.

PR #5550, opened 2018-06-11
Add doom-modeline recipe

seagle0128:

The modeline package extracted from [DOOM Emacs](https://github.com/hlissner/doom-emacs/tree/master/modules/ui/doom-modeline).

https://github.com/seagle0128/doom-modeline

maintaner

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/doom-modeline
+++ b/recipes/doom-modeline
@@ -1,0 +1,1 @@
+(doom-modeline :repo "seagle0128/doom-modeline" :fetcher github)


seagle0128:
Any update for this PR? Can you please accept it?

DamienCassou:
Sorry for the delay.

1. the `deface`s following a `defgroup` don't have to specify the `:group` because the group is inferred
3. Please run linters specified in [CONTRIBUTING.md](https://github.com/melpa/melpa/blob/master/CONTRIBUTING.md). They would have told you about:
    1. including Emacs in the package description (first line of the file is redundant)
    2. you should use standard keywords
    3. you shouldn't globally change values of other packages' variables because users might have specified something different. If you only want to declare the variables to please the compiler, remove the second argument to `defvar`. If you want to change the value of these variables while doom-modeline executes, let-bind them. If you really want to force new defaults on these variables, do that from a minor mode.
    4. there should be 2 spaces after periods in docstrings
    5. function docstrings should reference every argument (not the case of `doom-modeline--buffer-file-name-relative`)
    6. all variables and functions must have a docstring
5. the `defgroup` makes `doom-modeline` a subgroup of `doom` which may not exist. I suggest referencing a standard group instead
6. please add dash to the list of Package-Requires (even if it's already a requirement of another dependency)

seagle0128:
Thanks for your update!

I've fixed all issues according to your feedback. Please check the latest commit.

Only one question.
> there should be 2 spaces after periods in docstrings

If I add 2 spaces after periods, I got another warning:
`Second line should not have indentation (emacs-lisp-checkdoc)`.

So I have to a one empty in the second line. Is it Okay?

alphapapa:
I think the two-spaces rule doesn't apply across line breaks.

seagle0128:
@alphapapa @DamienCassou  Thanks! I updated the comments again. Please review and approve.

purcell:
Looking ready to merge now, which I'll go ahead and do, but I'd like to point out that you shouldn't ever `:inherit` other faces and *also* override their attributes, e.g. by making them bold, underlined or inverse-video. That could easily look awful, just as if you had invented your own colours. The best approach here is to simply `:inherit` the faces, and leave the user to customise the rest.

seagle0128:
Thanks for your comments, @purcell !
I will improve in the future release.

PR #5551, opened 2018-06-12
Add asana recipe

lmartel:

Quickly create, view and bulk-update [Asana](https://asana.com) tasks.

https://github.com/lmartel/emacs-asana

I'm the maintainer.

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/asana
+++ b/recipes/asana
@@ -1,0 +1,1 @@
+(asana :fetcher github :repo "lmartel/emacs-asana")


purcell:
Thanks! Sorry for the delay getting to this. Quick notes:

- This relies on Helm, which is quite an opinionated dependency that many people will want to avoid. You should at least mention that the package uses Helm in the package description line, and maybe even call the package `helm-asana`.
- `C-c a` is reserved for end users. Instead, you should define a prefix keymap and then tell the user how to bind it to their top-level keybinding of choice (which might then be `C-c a`)
- Generally you should just require `cl-lib` rather than `cl-seq` for portability: you can easily write code that works with older Emacsen by relying on the `cl-lib` package in ELPA, but that one only has a single top-level `cl-lib` feature.
- You will probably want to add a `;;;###autoload` for the `asana` command.

DamienCassou:

PR #5556, opened 2018-06-16
Add affect-vs-effect

grettke:

I always forget the difference between Affect vs. Effect. This package explains the difference

https://github.com/grettke/affect-vs-effect

Maintainer

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/affect-vs-effect
+++ b/recipes/affect-vs-effect
@@ -1,0 +1,1 @@
+(affect-vs-effect :repo "grettke/affect-vs-effect" :fetcher github)


alphapapa:
> Effect is a noun. It is an outcome or result of a verb. For example "Choosing spaces for indentation had a positive effect on her happiness."  There are other definitions for affect and effect and you probably shouldn't use them.

NAK.  "Effect" is also a verb, which means to cause something, e.g.:

>      1. To produce, as a cause or agent; to cause to be.
>         [1913 Webster]
>   
>               So great a body such exploits to effect. --Daniel.
>         [1913 Webster]
>   
>      2. To bring to pass; to execute; to enforce; to achieve; to
>         accomplish.
>         [1913 Webster]

:P 

But, really, a package for this?

grettke:
> NAK. "Effect" is also a verb, which means to cause something, e.g.:
> :P


> But, really, a package for this?

I imagine that there are hundreds and hundreds of Emacsers like me who need a frequent reminder even though we re-read the definition every couple of months.

vspinu:
The source of the proposed definition is missing BTW. Is this your own definition? 

But really, packages like this cause noise. I am also not convinced that affect-vs-effect confusion is that common. I am myself a big specialist in word confucius, but never with these.

milkypostman:
I think we have to pass on this. In general the rule of thumb has been that if a package is very specialized or is just as good as a snippet for a config then it’s probably not a good candidate for MELPA. 

I think the more general version of this would be a dictionary package or a “quick reference” package that could contain more general information covering a broader range of topics. 

Thanks for taking the time to submit this.

grettke:
@alphapapa I hear you. If there were a package for every grammar topic then, that wouldn't be good. Grammar checking belongs in a grammar checker.

@vspinu Yes it is my own definition. 

@milkypostman Gotcha. It is really a snippet. I was looking at [LanguateTool](https://github.com/languagetool-org/languagetool) and found the rule for this :)
If you are curious Languatetool covers it with multiple examples and URLs of references, pretty nice!

Thanks for spending your time helping me understand!

Have a great day.

PR #5557, opened 2018-06-16
Add code-archive

mschuldt:

maintains a git controlled source code archive and linked source code references using org-mode code blocks. Also provides helpers for org-capture.

https://github.com/mschuldt/code-archive

I am the Author

none needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/code-archive
+++ b/recipes/code-archive
@@ -1,0 +1,2 @@
+(code-archive :fetcher github
+              :repo "mschuldt/code-archive")


alphapapa:
Hi,

This looks very interesting!  Quick question:

> The variable code-archive-src-map maps major modes to org block source types. If the mode name does not have the format “<language>-mode” then the mode to language mapping must be added.

This doesn't feel right to me.  Isn't there already a way to do this?  e.g. `auto-mode-alist` maps extensions to major modes, so you could use it as a reverse map from major modes to extensions, and use the extension for the Org source block.

mschuldt:
@alphapapa Thanks for the feedback!

I replaced the call to shell-command-to-string with code-archive--run-git

I don't think using auto-mode-alist will work - the file extension or name does not provide the source type as needed by the org-mode source block, and even if it did, using auto-mode-alist means I'd have to parse it out of the regexp which does not sound fun. auto-mode-alist is also missing entries for some modes such as lisp-interaction-mode.

mschuldt:
Without entries for lisp-interaction-mode and GNUmakefile in auto-mode-alist I don't see it replacing the current method

And how to handle modes which map to many extensions, such as ruby-mode? which one to choose? seems like this makes things more complicated

mschuldt:
The point of code-archive-src-map was to map major-mode names
that are not recognized as org source block types to
valid block types. I've looked at each case again and
the modes names actually all work as source block names.
This might have started because I was mistaken about how the
source block types worked, I can't remember anymore -
this code has been part of my config for too long.

So code-archive-src-map could be deleted without loss of functionality.
I'd like to keep it because it enables mapping from multiple
mode types to the actual source type. This might enable searching
for codeblocks by source type instead of having to enumerate
all the associated major modes.

In practice each mapping is to the parent mode, this could be
done automatically but most modes are derived and I'm not
sure where to stop on the parent chain.

purcell:
Merged now - thanks for your patience.

I'd suggest that you switch to using `cl-lib`, since `cl` is obsolete. Generally this change is just a matter of adding a bunch of `cl-` prefixes to functions/macros you use.

PR #5563, opened 2018-06-19
evil-python-movement: port Neovim's movement in Python to Emacs/Evil

FelipeLema:

Port of Python movement in Neovim to Evil+Emacs.

Inspired by https://stackoverflow.com/a/28284564

https://bitbucket.org/FelipeLema/evil-python-movement.el/src/master/

Creator and maintainer

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/evil-python-movement
+++ b/recipes/evil-python-movement
@@ -1,0 +1,1 @@
+(evil-python-movement :url "https://bitbucket.org/FelipeLema/evil-python-movement.el" :fetcher git)


Alexander-Miller:
A couple of thoughts:

 * There's a thread over at /r/emacs about your package (and emojis in code in general). The general sentiment seems to be that you've committed a crime against nature.
 * I agree, those emojis make the code both unreadable and unwriteable.
 * It also goes against the convention of your package name being the prefix for your functions.
 * You're missing a Package-Requires header.
 * Shouldn't some of those functions or motions be autoloaded?

FelipeLema:
Just submitted changes to address your remarks. I didn't _think_ those motions needed autoload, but I don't mind them having them either.

edkolev:
@FelipeLema you'll need something like this for autoload to work with evil: https://github.com/edkolev/evil-lion/blob/master/evil-lion.el#L75

FelipeLema:
Done, thanks @edkolev

purcell:
Thanks for your patience - some quick notes:

- I think the URL in the recipe should be https://bitbucket.org/FelipeLema/evil-python-movement.el, ie. not include the "/src/master" suffix
- Your symbol prefixes are still not consistently `evil-python-movement-`
- The lines inside the `Commentary` section should begin with `;;`, not `;;;`.

FelipeLema:
Updated according to comments 👍

FelipeLema:
Just updated. Couldn't grep for `\bevil-python-[^m]`

DamienCassou:
  - you miss Homepage or URL header
  - either include standard keywords, or remove the `Keywords:` header

FelipeLema:
Updated per @DamienCassou's comments. Indeed I missed `package-lint`'s comments (I had installed with `flycheck-package`, but not activated it).

Sorry about missing that.

DamienCassou:
1. `top-level-def-regex` and `def-regex` are mostly the same. You might want to extract the common part and use the rx' `regex` construct to embed it into the other two.
2. instead of using the heavyweight `s-match` only as a boolean, you might want to use `string-match-p`.
3. I'm not sure why you can't use `search-forward-regexp` and `search-backward-regexp` instead of implementing your own (e.g., with `forward-to-top-level-def`).
4. In `common-python-movement`, I find it a bit weird that the code constructs a list only to get the last element. You could try to reimplement that with `cl-loop`
5. I suggest you write a few words around your `;; [[` and `;; [m` comment lines in the code. I first thought those comment lines were some special markup for a minor mode or for evil. What about `;; Definition for the [[ keybinding`

PR #5564, opened 2018-06-20
Recipe: Add magit-todos

alphapapa:

This package displays keyword entries from source code comments and Org files in the Magit status buffer. Activating an item jumps to it in its file. 

https://github.com/alphapapa/magit-todos

Author and maintainer.

### Checklist

Please confirm with `x`:

Thanks.

### DIFF
--- a/recipes/magit-todos
+++ b/recipes/magit-todos
@@ -1,0 +1,1 @@
+(magit-todos :fetcher github :repo "alphapapa/magit-todos")


matthew-piziak:
Any updates here?

alphapapa:
Don't worry, it will probably be reviewed soon.  There's a bit of a backlog, and reviewing new packages is mostly done just by Steve.  In the meantime, you can install it using other tools like `straight`, `quelpa`, `el-get`, `borg`, etc.

DamienCassou:
Please run linters on the file. The `Package-Requires` line is incorrect and some docstrings are incomplete.

matthew-piziak:
I've never submitted to MELPA so I don't know how the process works, but I've made a PR to try and fix the linting. Hope it helps! https://github.com/alphapapa/magit-todos/pull/22/files

alphapapa:
@DamienCassou My apologies, when I first submitted the PR there were no package-lint warnings, but I made some changes since then.

tarsius:
> The version of hl-todo I have installed has no version header,

None of my libraries have a version header because that would be incorrect most of the time - it would only be correct for that one commit that actually has been tagged with the same version string. Melpa gets the version from the tag anyway. So you should depend on `1.9.0`.

Of course because all melpa snapshot releases have timestamps as "versions" you cannot really specify the minimal version. For Magit itself I actually depend on a snapshot because it is popular enough for issues to arise every time if I didn't do that (except for tagged commits, those depend on actual releases, which I then revert in the very next commit). But for smaller packages it is not worth the effort.

alphapapa:
Thanks, Jonas.  What I've been doing for version headers is appending `-pre` to it in the untagged commits, so e.g. `1.1.0-pre`.  Then when I tag a version/release, I remove the `-pre`, tag the commit as e.g. `1.1.0`, and then increment the version and append `-pre` again in the next commit, e.g. `1.2.0-pre`.  According to the Emacs version-parsing code, that's a valid specification.  Would that work for Magit?  (Not that I would tell you how to do things. ;)

alphapapa:
@DamienCassou The package should be fit for MELPA now.  Thanks.

purcell:
Looks good, thanks - sorry it's taken a long time for me to get to this.

alphapapa:
No problem, Steve.  I appreciate all you do for MELPA.

PR #5565, opened 2018-06-20
Add org-outline-numbering

andersjohansson:

This package defines a minor mode that displays an outline numbering as overlays on Org mode headlines. The numbering matches how it would appear when exporting the org file.

https://gitlab.com/andersjohansson/org-outline-numbering

I am the maintainer.

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/org-outline-numbering
+++ b/recipes/org-outline-numbering
@@ -1,0 +1,1 @@
+(org-outline-numbering :repo "andersjohansson/org-outline-numbering" :fetcher gitlab)


alphapapa:
Hi, this looks really cool, a feature that Org has needed for a long time!

I wonder about using `org-element-parse-buffer`, because it means parsing the whole buffer, which is slow for large buffers. Of course, it's natural to use it since you're using the org-export code, which is great to reuse, and important to do so to present the same result as exporting.  But I wonder if it could be optimized by either having it only parse headlines, or gathering headlines directly with `re-search-forward` and transforming them into a format that the rest of the numbering code could use.

The end goal, of course, would be to make updating the outline numbers automatic whenever the structure is changed. I was thinking that maybe we could patch Org to add some hooks which would be called whenever outline structure changes, and then your package could hook into it there (or even be integrated directly into Org). 

Thanks for sharing your work!

andersjohansson:
> But I wonder if it could be optimized by either having it only parse headlines, or gathering headlines directly with re-search-forward and transforming them into a format that the rest of the numbering code could use.

Oh, yes I have also thought about this. My first tries were with `(org-outline-parse-buffer 'headline)` (as well as restricting to `greater-element` or `element`) but I found (through lots of trial-and-error) that `org-export--prune-tree` only works on a complete parse tree.

The re-search-forward solution was in the initial code that inspired me to do it this way (https://emacs.stackexchange.com/a/32422) but that didn’t allow as much flexibility and consistency with exports as I wanted.

For my use, (2000 lines, 300KB org-mode document for my thesis) it has been fast enough however (less than a second). I just update it on-demand. But this would of course not suffice for auto-updates whenever the org structure changes.

Perhaps some further changes to org (maybe a custom/simplified `org-export--prune-tree`) to allow this to be done more efficiently would be required. But I really don’t know how. Understanding the org parsing in `org-element` and `ox`  mostly gives me headache, especially because it is so hard to inspect the parsed structures via edebug etc.

alphapapa:
Sounds like you are way ahead of me.  :)

> Understanding the org parsing in org-element and ox mostly gives me headache, especially because it is so hard to inspect the parsed structures via edebug etc.

I sympathize with you here!  org-element is great but hard to follow.  And I quickly tire of waiting for Emacs to stop churning the CPU because of long lines in edebug (sometimes I wonder if that only happens to me).  Sometimes `pp-eval-last-sexp` can help when looking at large lists, but it can also be slow.

alphapapa:
Here's a very ugly function that seems about 3x faster than using `org-element`:
```el
(defun org-outline-numbers ()
  (interactive)
  (save-excursion
    (let* ((positions-levels (progn
                               (goto-char (point-min))
                               (when (org-before-first-heading-p)
                                 (outline-next-heading))
                               (cl-loop while (not (eobp))
                                        collect (cons (point) (org-current-level))
                                        do (outline-next-heading))))
           (tree (cl-loop with current-top-level = 0
                          with current-subtree-numbers
                          with results
                          with previous-level
                          for (position . level) in positions-levels
                          if (= 1 level)
                          do (progn
                               (setq current-subtree-numbers nil)
                               (setq previous-level level)
                               (push (a-list 'heading (save-excursion
                                                        (goto-char position)
                                                        (substring-no-properties
                                                         (org-get-heading t t)))
                                             'position position
                                             'level level
                                             'number (concat (number-to-string (incf current-top-level)) "."))
                                     results))
                          else do (let* ((current-level-number (cond ((= level previous-level)
                                                                      (incf (map-elt current-subtree-numbers level)))
                                                                     ((< level previous-level)
                                                                      (incf (map-elt current-subtree-numbers level)))
                                                                     ((> level previous-level)
                                                                      1)))
                                         text-number)
                                    (setq previous-level level)
                                    (map-put current-subtree-numbers level current-level-number)
                                    (setq text-number (cl-loop for lookup from level downto 1
                                                               for lookedup = (map-elt current-subtree-numbers lookup)
                                                               if lookedup
                                                               collect lookedup into result
                                                               else collect current-top-level into result
                                                               finally return (s-join "." (mapcar #'number-to-string (nreverse result)))))
                                    (push (a-list 'heading (save-excursion
                                                             (goto-char position)
                                                             (substring-no-properties
                                                              (org-get-heading t t)))
                                                  'position position
                                                  'level level
                                                  'number text-number)
                                          results))
                          finally return (nreverse results))))
      (ov-clear)
      (--each tree
        (let-alist it
          (ov (+ .position (1- .level)) (+ .position .level)
              'display .number))))))
```
It needs to be tidied up, but it seems to work.  Feel free to take any of it if you find it useful.

purcell:
Looks good, but you're missing a package dependency on `cl-lib`: you should add `(cl-lib "0.6")` to the `Package-Requires` header.

purcell:
Once done, just ping here, and any maintainer can merge this. Sorry for the slow turnaround. :-)

andersjohansson:
Ok! Added the `cl-lib` dependency. I didn’t think it was needed since it’s part of Emacs, but I guess I still have some things to learn about packaging!

tarsius:
> I didn’t think it was needed since it’s part of Emacs,

It isn't for older Emacs releases, which come with `cl` instead. When using a newer Emacs release, then the `cl-lib` dependency is indeed satisfied by Emacs itself.

PR #5570, opened 2018-06-20
Add a recipe for deadgrep

Wilfred:

Deadgrep is an Emacs package for performing ripgrep searches. Unlike `M-x grep` and ag/rg equivalents, it does not use `compilation-mode`. Instead, it is tailor-made for displaying, navigating
and filtering ripgrep results.

deadgrep_screenshot

https://github.com/Wilfred/deadgrep

I'm the author.

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/deadgrep
+++ b/recipes/deadgrep
@@ -1,0 +1,1 @@
+(deadgrep :repo "Wilfred/deadgrep" :fetcher github)


alphapapa:
This looks really nice!  I wonder about the name though, because it doesn't convey that it uses rg, but suggests that it uses grep, which might cause potential users to overlook it.  Of course naming things is hard. :)

manuel-uberti:
FWIW, I kind of like the dead/rip play here. :)

alphapapa:
Ahh, okay, now I get it.  :)

purcell:
Thanks, merged. Can I suggest using the now-built-in "project-root" function instead of "projectile"?

purcell:
Also, sorry for the slow turnaround.

Wilfred:
Thanks for merging, and please don't apologise for your consistent support of MELPA :)

By the way, I can't see a function called `project-root`. I can only see `project-roots`, which needs a project as an argument. Is this a feature that's only on Emacs trunk?

purcell:
Sorry, I meant `project-current`. This is in `project.el`, which says ";; Copyright (C) 2015-2018 Free Software Foundation, Inc.". Unsure when that function was added, sorry.

purcell:
I'm a bit annoyed that newly-added libraries and functions are not reliably annotated with the version in which they were added these days. I try to keep track of many new libs/functions in order to provide feedback via `package-lint`, but it's a losing battle.

PR #5574, opened 2018-06-23
Add counsel-org-capture-string

akirak:

Quickly capture a line of text using counsel and org-capture-string

https://github.com/akirak/counsel-org-capture-string

I am the maintainer.

**None needed**

### Checklist

Please confirm with `x`:


purcell:
Thanks! You might like to include at least one `;;;###autoload`. Also, you should generally require `cl-lib`, not `cl-macs` -- the latter is not available in the standalone `cl-lib` package, so `cl-` code  is more portable if you prefer `(require 'cl-lib)`.

PR #5576, opened 2018-06-24
Add libgit

tarsius:

This is an experimental module for libgit2 bindings to Emacs, intended to boost the performance of magit.

https://github.com/magit/libegit2

@TheBB is the author and the repository is part of the `magit` "organisation", whose maintainer I am.

https://github.com/magit/magit/issues/2959 covers most of the past discussions.

### Checklist

  - I'll submit a pull-request.
  - I'll submit a pull-request.
- [ ] `M-x checkdoc` is **mostly** happy with my docstrings
  - I'll submit a pull-request
  - I think messages such as `"libgit: building failed with exit code %d"` are okay, or do we have to say `"Libgit: ..."`?
  - `libgit-load` needs a doc-string.
- [x] I've loaded `libgit`, which triggers the module to be build, and that succeeded.
* [x] I fixed a bug in `package-build.el` in the process :wink: 

### Going forward

Lets use this pull-request to discuss what still needs to be done (packaging-wise) before we can start distributing this as a (M)ELPA package.

Installing this package using `package.el` does not complete the installation, the C module also has to be build. If the module isn't available yet, then loading `libgit` triggers the build.

I would like to add this package soon, as long as possible before Magit actually starts depending on it. Provided some users volunteer to try installing it, this should give us some time to iron out as many build issues as possible before many users are affected.

We should probably create a wiki page, which lists the libraries and tools necessary to build the modules and how to install them on various platforms. And encourage testers to help with that.

Some other notes/questions:

* I tried to remove unneeded files but we might be able to remove some more. The tarball is still 5M, down from 20M.
* I kept `gitlib2/{COPYING,.HEADER}` for legal reasons. 
* Why do we bundle `uthash`?

### DIFF
--- a/recipes/libgit
+++ b/recipes/libgit
@@ -1,0 +1,17 @@
+(libgit :fetcher github
+	:repo "magit/libegit2"
+	:files ("CMakeLists.txt"
+		("libgit2" "libgit2/cmake")
+		("libgit2" "libgit2/CMakeLists.txt")
+		("libgit2" "libgit2/COPYING")
+		("libgit2" "libgit2/deps")
+		("libgit2" "libgit2/.HEADER")
+		("libgit2" "libgit2/include")
+		("libgit2" "libgit2/libgit2_clar.supp")
+		("libgit2" "libgit2/libgit2.pc.in")
+		("libgit2" "libgit2/script")
+		("libgit2" "libgit2/src")
+		"libgit.el"
+		"Makefile"
+		"src"
+		"uthash"))


manuel-uberti:
If you need a volunteer, count me in. Just describe the exact steps you need me to test it properly, and I'll report back any problem I encounter (if any).

TheBB:
The testing steps should be something like this:

- Clone this repo (that is, melpa) and check out this PR branch.
- `make recipes/libgit`
- Go to your Emacs and do `package-install-file` with the generated file.
- Do `(require 'libgit)`, and see what happens.

Currently the library should try to build itself. This may take some time. See the `libgit.el` file.

I was hoping to distribute binaries on Github releases so that the user has the option (default or not) to download those when first loading the library. If this solution becomes good enough we won't need to distribute libgit2 sources either, which is what blows up the tarball at the moment.

I used uthash for internal memory management. Of course it's possible to just do the hash table ourselves too.

manuel-uberti:
This is what I get and evaluating `(require 'libgit)` after `package-install-file` runs correctly.
```
CMake Error: The source directory "/home/manuel/.emacs.d/elpa/libgit-0.0.1" does not appear to contain CMakeLists.txt.
Specify --help for usage, or press the help button on the CMake GUI.
```

TheBB:
Thanks. That's weird though, the recipe says to bring that file and I can't reproduce here (it works fine).

manuel-uberti:
The file is present:
```
~/githubs/melpa/working/libgit master 
❯ ls -lah CMakeLists.txt 
Permissions Size User   Date Modified Name
.rw-r--r--   809 manuel 25 giu  9:06  CMakeLists.txt
```

TheBB:
Yes, in the cloned repo, but is it in the tarball?

manuel-uberti:
What tarball? :)

TheBB:
Ah, okay. After the make step you should have a file melpa/packages/libgit-....tar. That's the one you need to pick when doing package-install-file. Sorry I wasn't clear.

manuel-uberti:
I see. Yes, the file it's in the tarball as well. So I install the tarball when prompted from `package-install-file`?

TheBB:
Yep.

purcell:
Great, thanks!

manuel-uberti:
@TheBB @tarsius FYI: I installed it from MELPA with
```
(use-package libgit                     ; Bindings for libgit2
  :ensure t
  :after magit)
```
No problem at all. Compilation of the module started automatically.

PR #5587, opened 2018-06-29
Add word-count

mhatta:

A minor mode which shows the number of characters/words/lines of the current buffer in the modeline.

https://github.com/mhatta/word-count-mode

I assume I'm the current maintainer.

This elisp was originally written by Hiroyuki Komatsu years ago.  After a  long period of inactivity (and the upstream website has long gone), Tomasz Skutnik salvaged it from Wayback Machine and made it work with Emacs 24.  I made it work with Emacs 26.

I contacted with Tomasz via  e-mail, and with his blessing, I prepared for MELPA.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/word-count
+++ b/recipes/word-count
@@ -1,0 +1,1 @@
+(word-count :repo "mhatta/word-count-mode" :fetcher github)


alphapapa:
Unfortunately that code has a lot of cruft from very old Emacs versions, as well as XEmacs.  It reimplements a lot of things that should be handled by built-in functions or libraries on MELPA.  IMHO it needs significant cleaning up and some rewriting to be suitable for publishing on MELPA, but Steve, et al. will make that decision.

mhatta:
@alphapapa Thanks for the comment.  Could you give me some hints on how this can be improved, such as libraries I should consider using?  Also, I can remove XEmacs-related code (which will eliminate most of the remaining warnings, I guess).

alphapapa:
It's hard to know where to begin.  I see that it has its own versions of `defvar` and `defvar-local`, its own version of `replace-regexp-in-string`, it's own list-building and -modifying functions...  I think it would be easier to rewrite it from scratch, considering that it only needs to do two things: 1) get word count, 2) put it in the mode-line.

casouri:
That sounds kinda cool... from a historical perspective.

DamienCassou:
I'm closing this for now. Please reopen if you want to.

PR #5598, opened 2018-07-04
Add fd-dired

yqrashawn:

Alternative for `find-dired`, use `fd` as backend.

https://github.com/yqrashawn/fd-dired

Maintainer

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/fd-dired
+++ b/recipes/fd-dired
@@ -1,0 +1,1 @@
+(fd-dired :fetcher github :repo "yqrashawn/fd-dired")


DamienCassou:
1. the function `fd-dired` is really long. The longer a function the harder it is for a MELPA maintainer to understand what is happening and to give feedback. It is also much harder to point to a specific portion of the code we believe could be improved. Please spend time decomposing your long functions into smaller, well-named and documented, ones.
2. prefix function names with #' instead of just ' to tell the compiler this is a function reference. E.g., `(define-key map "\C-c\C-k" #'kill-find)`

PR #5601, opened 2018-07-05
add recipe for nopaste

anarcat:

nopaste is a helper to send various contents from emacs into online
"paste" services. This is useful to share long code snippets with
other users on IRC or other medium.

https://github.com/anarcat/nopaste

I was briefly a contributor and an enthusiastic user but I was
delegated the maintainership of the package, see https://github.com/avar/nopaste/pull/2#issuecomment-401323781

**None needed**.

### Checklist

### DIFF
--- a/recipes/nopaste
+++ b/recipes/nopaste
@@ -1,0 +1,1 @@
+(nopaste :fetcher github :repo "anarcat/nopaste")


anarcat:
hmm... sorry about the layout, it looks like the `hub` command messed with your issue template. i'll see if i can fix that...

DamienCassou:
Thank you for your proposal. This looks very similar to webpaste though: https://melpa.org/#/webpaste. Could you please contribute to webpaste instead?

MELPA provides a curated set of Emacs Lisp packages, not an exhaustive list of every single Emacs Lisp file ever created. As such, MELPA maintainers tend to only accept packages not duplicating existing packages.

anarcat:
I did not know about this package, it looks interesting!

However it takes a radically different approach than nopaste: it
implements the RPC mechanisms in Emacs itself instead of leveraging the
existing infrastructure (e.g. the "nopaste" commandline binary).

In my specific case (packaging for Debian), we like to default the
pastebin to paste.debian.net, which is not supported by webpaste.el, but
is supported by nopaste. And while I could contribute yet another plugin
to webpaste.el, I am somewhat at the end of my energy on this and I
don't think I will be able to push this much further...

It does feel somewhat frustrating to me to see everyone reimplement the
same protocols over and over again in all sorts of different
languages. Nopaste is only one of them - there's also pastebinit
(written in Python instead of Perl) which does basically the same thing,
and now webpaste.el. All of them struggle to support multiple pastebins
and all fail to support them all. If energies would be spent
contributing on the commandline tools instead, more websites would be
supported...

But I understand where MELPA is coming from. If the difference in
implementation is insufficient to justify inclusion, I guess you can
just close this pull request and move on.

Thanks for the response!

DamienCassou:
I think I understand your frustrations: people duplicating other's work instead of leveraging it. One thing you could do about it is to open an issue at webpaste presenting App::Nopaste and the advantages of using that instead of reimplementing the wheel.

/cc: @milkypostman @purcell @tarsius @syohex what is your opinion?

DamienCassou:
In this absence of answer from other maintainers, I'm closing this ticket. One of them can always reopen if they want to.

anarcat:
i understand, thank you for your patience and clear understanding. :)

Issue #5590, opened 2018-06-30
The sources of spike-theme and sqlite.el have disappeared


aplaice:
### `spike-theme`

[MELPA's recipe](https://github.com/melpa/melpa/blob/master/recipes/spike-theme).

(Broken) [GitHub link](https://github.com/m31271n/spike-theme).

[Emacsmirror link](https://github.com/emacsmirror/spike-theme).

### `sqlite.el`

[MELPA's recipe](https://github.com/melpa/melpa/blob/master/recipes/sqlite).

(Broken) [GitHub link](https://github.com/cnngimenez/sqlite.el).

[Emacsmirror link](https://github.com/emacsmirror/sqlite)

The relevant [emacswiki elisp source page](https://www.emacswiki.org/emacs/sqlite.el) which points at the above broken GitHub link. The [description page](https://www.emacswiki.org/emacs/SQLite-el) also points there.

Looking at [MELPA's metadata](https://melpa.org/archive.json), neither of the packages seems to be depended upon.

tarsius:
Thanks for notifying us. Could you please also contact the upstreams and figure out what has happened? Maybe the packages have moved elsewhere, maybe they have been deprecated, ...

Unless that results in a more appropriate action, I will move these packages from the Emacsmirror to the Emacsattic.

aplaice:
I've sent e-mails to them asking about the situation. I'll provide updates if/when I get replies.

cnngimenez:
Hi!

Following the recommendations of most people, I migrate my repositories to Gitlab. If you cannot find some repositories, it is possible to find them in other hosts (Bitbucket, Savannah, etc.).

sqlite.el con be finded at https://gitlab.com/cnngimenez/sqlite.el

aplaice:
Hi!

Thanks very much for the reply! Moving to GitLab is indeed probably a wise course of action. :) 

Unfortunately, it seems that your repositories are private — when I try to go to [sqlite.el](https://gitlab.com/cnngimenez/sqlite.el) I get the message:

> You need to sign in or sign up before continuing.

It appears that Travis [can't access your repository](https://travis-ci.org/melpa/melpa/builds/401463245) either. Your list of [publicly available personal projects](https://gitlab.com/users/cnngimenez/projects) also seems to be empty.

Sorry for the trouble!

cnngimenez:
Fixed! It should be public now. 

Sorry, didn't realize it was imported as private.

tarsius:
[...]

PR #5605, opened 2018-07-08
sqlite.el repository: fixed repository URL

cnngimenez:
I have fixed the sqlite.el fetcher to point to the gitlab. 

Refer to issue #5590.


purcell:
Thanks!

PR #5612, opened 2018-07-13
Update MELPA to build both unstable and stable.

milkypostman:
- Combine stable and unstable builds into the same rotation so one
builds after the other.

- Move all the services to docker containers,
  - nginx: contains everything for handling http traffic including log
  rotation and ssl certs.
  - syncer: manages syncing all the html files to the latest build
  information.
  - logprocessor: processes the http logs.
  - rsyncd: rsync for melpa repo.


milkypostman:
this change is going to make our lives easier in the long run. the docker change is HUGE and a really big improvement IMO. docker will keep everything running and it's easy to get logs of everything that is happening if we want to try and track down issues.

also, with the new server i have changed to use block storage that is much cheaper and is something we can grow as we need. already the old servers are running out of space all the time. and it's a lot of $$$ to grow them.

html/js/melpa.js
```diff
@@ -524,7 +524,8 @@
   // Changing the appearance of the MELPA Stable page
   //////////////////////////////////////////////////////////////////////////////
 
-  melpa.stable = m.prop(window.location.host === 'stable.melpa.org');
+    melpa.stable = m.prop(window.location.host === 'stable.melpa.org' ||
+                          window.location.host === 'stable-test.melpa.org');
   melpa.archivename = {};
```
Silex: The indendation looks off here

PR #5615, opened 2018-07-13
Update index.html

anharismail:


purcell:
Thanks. Adding the logo isn't as simple as this, sorry - I'll do it directly myself.

PR #5616, opened 2018-07-13
Add flatui-theme-mw.

MetroWind:

A custom theme inspired by http://flatuicolors.com. Has both light and dark variants.

https://github.com/MetroWind/emacs-flatui-theme

Author.

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/flatui-theme-mw
+++ b/recipes/flatui-theme-mw
@@ -1,4 +1,0 @@
-(flatui-theme-mw
- :fetcher github
- :repo "MetroWind/emacs-flatui-theme"
- :branch "master")
--- a/recipes/flatui-theme-mw
+++ b/recipes/flatui-theme-mw
@@ -1,0 +1,4 @@
+(flatui-theme-mw
+ :fetcher github
+ :repo "MetroWind/emacs-flatui-theme"
+ :branch "master")
--- a/recipes/flucui-themes
+++ b/recipes/flucui-themes
@@ -1,0 +1,3 @@
+(flucui-themes
+ :fetcher github
+ :repo "MetroWind/flucui-theme")


purcell:
Thanks, particularly for your patience. Some feedback:

- The naming here is all very inconsistent - I don't get why there's an `-mw` suffix on one file and the package name, but not elsewhere. I'd suggest using `flatui-mw-` as a consistent prefix for all files and symbols, and you should also keep the file and theme names in lower case in order to fit in with elisp conventions.
- You should also `;;;###autoload` [this line](https://github.com/MetroWind/emacs-flatui-theme/blob/65d0dc9ffb2cb92491c1f8898afa86962f9e9945/flatui-theme-mw.el#L23).
- The phrase "for Emacs" is almost always redundant in elisp code, which can only ever be for Emacs. :-)
- [This code](https://github.com/MetroWind/emacs-flatui-theme/blob/65d0dc9ffb2cb92491c1f8898afa86962f9e9945/FlatUI-theme.el#L273) is bogus - the file is not evaluated every time the theme is loaded, so this is not the way to reliably set `ns-appearance`, even if it were appropriate to do so as part of the theme, which it isn't. :-)

Hope that helps!

MetroWind:
> I don't get why there's an -mw suffix on one file and the package name

The reason is that there's already a `flatui-theme` in melpa, which is different from my theme. One option for me is to rename the package "flatui-theme**s**". Do you think it is more acceptable?

milkypostman:
i think flatui-themes is misleading and inconsistent. it should end in
`-theme`. so probably something like `flatui-mw-theme` or `flatmw-theme` or
something completely different.

MetroWind:
Renamed to `flucui-themes`. (Should I edit my previous comments and title to reflect this?) Also addressed other comment from @purcell. But I’m not sure about 

> You should also ;;;###autoload this line.

(That line is a `(add-to-list …)`)
I’m not sure how `autoload` works on a function call…?

purcell:
> I’m not sure how `autoload` works on a function call…?

It puts the code in the package's autoload file. You will want the package's directory adding to the theme load path at startup, so that the themes are available after installation, so you'll have to autoload this to avoid unexpected difficulties for users.

Ping when done, and any of us can then merge this - it all looks good.

PR #5627, opened 2018-07-17
Add counshell

lpenz:

counshell provides an interactive command line where you can navigate the resulting file names and line numbers using ivy.

https://github.com/lpenz/counshell

I'm the author and maintainer

None

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/counshell
+++ b/recipes/counshell
@@ -1,0 +1,2 @@
+(counshell :repo "lpenz/counshell"
+           :fetcher github)


alphapapa:
This is really cool.

But also really dangerous.  As you admit in the readme:

> Be aware that several partial commands will be executed while you type, which might get "interesting" if your command has side-effects. For instance, if the command requires you to type the name of a file that is created, files with partial names can end up being created.

Sorry, but IMO evaluating shell commands character-by-character as the user types them is a *really, really bad idea* and should definitely not be in MELPA.

lpenz:
Oh yeah, the type-the-whole-command function requires some attention.
To be honest, I've been using the functions that have the start of the line already typed and read-only more useful than the write-the-whole-command function. It's more practical, and allows me to give grep/rg/etc. more command-line options - which is what I miss in the standard counsel functions. But you are still going through the shell, and you can just type a `;` and invoke a completely different command - meaning that you still have to be aware that that's how it works.

Thanks for the feedback.

alphapapa:
Alternatively, may I suggest this package be renamed `footgun-mode`.  ;)

lpenz:
Isn't that already an alias for c++-mode? ;)

purcell:
This is delightful and terrifying in equal parts! I'd like to merge it, but I also think that there should be some sort of yes/no acknowledgement, perhaps even the first time you use it, because some users will certainly not realise their commands will be run on every keystroke. Make it a boolean `defcustom`, and if it's not set, prompt the user to confirm, then set it.

P.S. Thanks for your patience.

lpenz:
I've created https://github.com/lpenz/counshell/pull/1
@purcell would you mind taking a look, if that's sort of what you had in mind?

kidd:
I'm no authority here, but I think this package is too dangerous to be in melpa as is. 

Some ideas to make it less scary: 

- Make it opt-in. The user has to set some variable to non-nil. via elisp. not even defcustom.
- blacklist some commands. `mv` and `rm` are the ones that come to my mind.
- Be explicit about how expansion happens. (what would happen if I type `mv * `.) 

Here are some comments on a similar package that appeared in reddit just a few minutes ago (general scare): https://www.reddit.com/r/programming/comments/9qyp15/up_a_tool_for_writing_linux_pipes_with_instant/.

DamienCassou:
We need to take a decision here to not keep the author waiting forever.

tarsius:
Lets not add it.

Silex:
> * blacklist some commands. `mv` and `rm` are the ones that come to my mind.

It'd be a whitelist instead.

purcell:
Looks like there's consensus on not adding this to MELPA, due to the footgun factor. Sorry @lpenz - it's a nice hack!

lpenz:
Thanks guys! It was worth a shot, and I appreciate the feedback.

PR #5628, opened 2018-07-17
Add defkey

nickdrozd:

Provides a novel API for defining key bindings

https://github.com/nickdrozd/defkey

Author, maintainer

None needed

### Checklist

Please confirm with `x`:

#### `package-lint` objects to having a macro called `defkeys` in the package `defkey`.

### DIFF
--- a/recipes/defkey
+++ b/recipes/defkey
@@ -1,0 +1,1 @@
+(defkey :repo "nickdrozd/defkey" :fetcher github)


alphapapa:
This is neat, but TBH, it appears that the only difference between it and `general-def` is the lack of quotation marks and quoted function symbols (neither of which I'm sure is a good thing).  e.g.:

```el
(general-def
  "C-x C-f" #'helm-find-file)
```
Compared to:
```el
(defkeys
  (C-x C-f) helm-find-file)
```
And the necessity to escape certain characters like this seems like a big drawback:
```el
(defkeys
  C-\( this
  M-\; that
  s-\, the-third)
```
Also, what about remapping?  e.g.:
```el
(general-def
  [remap winner-redo] #'winner-redo-repeat
  [remap winner-undo] #'winner-undo-repeat)
```
Can `defkey` do that?

I don't mean to demean your package, but `general` is so well-done and comprehensive, that I'm not sure `defkey` provides any benefit to users, in which case having yet another package for users to decide between may actually be a net negative.

noctuid:
It doesn't seem ideal to me to break/be incompatible with so many possible keys and definitions just to prevent the need to quote strings and (sharp) quote functions. Commands are the most common definition, and while I find the syntax less readable, I don't see a problem with the package as long as it's recommended as specific-purpose definer and not as a replacement for define-key.

nickdrozd:
As for whether this approach is "good", I just came across [this interview](http://www.princeton.edu/~hos/mike/transcripts/weinberger.htm) from 1989 with Peter Weinberger discussing the development of Awk (the name of which comes from Aho, Weinberger, and Kernighan):

> [Aho] was not very enthusiastic about [extensible languages] for the technical reasons were that he was at that time, and I think as everybody is now, much more in favor of table-driven parsers, in particular LR parsers, and if you've compiled a table it's a little hard to adjust the syntax. And I suspect in retrospect he didn't think much of the usefulness of the technique. And I think in retrospect he was probably right. Extensible languages is not something that worked out well. And the places where people are very enthusiastic about tailoring the syntax of [?] for users. There's two of them that I can think of. One is emacs, which is this editor where you can bind keys in weird ways. And the other is with something called Read Macros in Lisp. And the effect of both of those is that I can't enter your world, because each person personalizes his world in some some way they like, and nobody else can cope. So in fact it's not at all clear it was a good idea.

My personal config *is* my world; why should I care if anyone else can "enter" or "cope"? And if such personalizations are forbidden, why bother using Lisp at all?

In any case though, I was amused to see that this debate has been going on for decades in *exactly* the same terms.

noctuid:
I'm not saying that there's anything "wrong" with or "bad" about defkey's approach, just that there are many valid define-key statements that can't be translated to defkey. For your personal config, that may not be an issue, but a package in MELPA is not personal configuration. I disagree with that quote completely too. Like I said, commands are the most common key definition, so some users may see defkey's syntax as better for most keybindings. I just think some confusion could be avoided by listing all limitations in the readme. That was my only suggestion.

milkypostman:
Certainly this is an interesting approach but I would consider this a personal customization over what I would call a package. It would be valuable as a snippet or something for the Emacs Wiki but I don’t think it belongs, in its current form, in MELPA

it is nicely developed with tests so I appreciate the work you’ve done but I ultimately agree with others here.

PR #5630, opened 2018-07-18
Reorganize the CONTRIBUTING file

DamienCassou:
I took the liberty to completely reorganize the CONTRIBUTING file. I moved everything around, deleted what I felt was duplicate, and added a bunch of paragraphs of my own. Please have a careful look as you might not like my additions (I hope you do though :-)).


alphapapa:
Looks very good!

CONTRIBUTING.org
```diff
@@ -0,0 +1,186 @@
...
+** Waiting for reviews and taking feedback into account
+
+MELPA maintainers spend a lot of time reviewing proposed packages and
+also have quite a lot of other non-MELPA-related activities. Please be
+patient as it might take a week (sometimes several) before one starts
+having a look at your pull request. 
+
+You can help MELPA maintainers take care of pull requests much faster
+by paying real attention to the quality of your package (see above for
+some quality checks and links). If you feel for it, you can also take
```
tarsius: "If you were asked to make several changes, then you should explicitly mention everything that you have fixed, and possibly even link to the relevant commits. Once way of doing that is to mention the melpa pull-request in every commit that addresses one of the raised points, using the format `melpa/melpa#N`."

(Double check that `melpa/melpa#N` works.)

CONTRIBUTING.org
```diff
@@ -0,0 +1,186 @@
...
+     no longer accepted.
+
+- Official repository :: Packages should be built from the official
+     package repository.  Forks of the official repository will not be
+     accepted except in extreme circumstances.
+
+- Dedicated SCM repository :: Keep each package in its own SCM (e.g.,
+     git) repository. This makes it possible to have a different
+     version number for each package as MELPA stable looks at the SCM'
+     tags to assign version numbers to recipes.
```
tarsius: In a separate item you should mention how to make a package independent of a certain completion framework, by using `dont-remember-ask-steve`.

You should place that item after this one, because it is an exception to the one-package-per-repo rule: when splitting e.g. `helm-foo` from `foo`, it is perfectly okay (though not required) to keep both libraries in the same repository.

CONTRIBUTING.org
```diff
@@ -0,0 +1,186 @@
...
+
+Packages submitted to MELPA regularly suffer from the same problems
+which delay review by several days or even weeks. Please double check
+this list before submitting your package:
+
+- Please run quality-checking tools specified above (really, do it!).
+- Please add ~-*- lexical-binding: t; -*-~ at the end of the first
+  line of each Emacs Lisp file.
+- A ~defcustom~ or ~defface~ following a ~defgroup~ doesn't have to
+  specify the ~:group~ because the group is inferred.
```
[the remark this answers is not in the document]
milouse: I agree @tarsius. As often tell in the python world: « explicit is better than implicit ».

CONTRIBUTING.org
```diff
@@ -0,0 +1,186 @@
...
+- Please avoid defining a face that both ~:inherit~ another face and
+  also override their attributes (e.g. by making them bold, underlined
+  or inverse-video). The result could be really bad depending on user
+  customizations. The best approach here is to simply ~:inherit~ the
+  faces, and leave the user to customise the rest.
+- Prefix function names with #' instead of just ' to tell the compiler
```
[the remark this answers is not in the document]
milouse: Also add a reference to the [emacs lisp documentation](https://www.gnu.org/software/emacs/manual/html_node/elisp/Anonymous-Functions.html)?

CONTRIBUTING.org
```diff
@@ -0,0 +1,186 @@
...
+the stable package.
+
+*** Fixing typical problems
+
+Packages submitted to MELPA regularly suffer from the same problems
+which delay review by several days or even weeks. Please double check
+this list before submitting your package:
+
+- Please run quality-checking tools specified above (really, do it!).
+- Please add ~-*- lexical-binding: t; -*-~ at the end of the first
```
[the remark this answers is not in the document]
milouse: Documentation is [here](https://www.gnu.org/software/emacs/manual/html_node/elisp/Lexical-Binding.html), but what we lack here is an explanation of why lexical binding is important for our files. I mean: if it is not enabled by default in emacs, it should be for a reason, thus what's the point about blindly enable it everywhere ? :/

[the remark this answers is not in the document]
DamienCassou: > Link to do documentation or some other suitable explanation.

done

> Documentation is here

thank you for the link, I used it.

> but what we lack here is an explanation of why lexical binding is important for our files.

I added a link explaining that

>  I mean: if it is not enabled by default in emacs, it should be for a reason

backward compatibility with old code relying on dynamic scoping.

CONTRIBUTING.org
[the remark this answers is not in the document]
DamienCassou: I removed the item.

CONTRIBUTING.org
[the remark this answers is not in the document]
DamienCassou: > Did Org choke on ~#'~ and ~'~? If not, then I would do that.

yes. I think Org requires some configuration to be able to do this highlighting.

> Also add a reference to the emacs lisp documentation?

done. Thanks for the link.

CONTRIBUTING.org
```diff
@@ -0,0 +1,186 @@
...
+Create a file under the directory specified by
+~package-build-recipes-dir~ (default: ~recipes/~ directory where
+~package-build~ was loaded). If you prefer, the interactive command
+~package-build-create-recipe~ in ~package-build/package-build.el~ will
+guide you through this process. The filename should match the name of
+the package's provided feature.
+
+See the [[file:README.md#recipe-format][recipe format]] section of the README for more information on
+the content of a recipe file. The package name should be the same as
+the filename.
```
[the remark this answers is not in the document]
DamienCassou: > It must.

done

> Maybe mention that exceptions can be made

I mentioned that globally at the beginning of the document

CONTRIBUTING.org
```diff
@@ -0,0 +1,186 @@
...
+#+BEGIN_SRC shell
+EMACS_COMMAND=/path/to/emacs make sandbox INSTALL=<name>
+#+END_SRC
+
+From within Emacs, install and test your package as appropriate. This
+is a useful way to discover missing dependencies.
+
+** Opening a pull request
+
+Create a [[https://github.com/magit/magit/wiki/Dedicated-pull-request-branches][dedicated pull request branch]] in your clone of the [[https://github.com/melpa/melpa][MELPA
```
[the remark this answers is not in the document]
DamienCassou: awesome text, thank you!

CONTRIBUTING.org
DamienCassou: Thanks for the paragraph. I introduced it.

> Double check that melpa/melpa#N works

done. It does

DamienCassou:
I took care of all feedback and added a paragraph about long functions.

CONTRIBUTING.org
```diff
@@ -0,0 +1,213 @@
...
+     and associated packages as well as keeping the system working.
+
+- Recipe authors :: They responsible for submitting good recipes and
+                    maintaining them when the package is changed
+                    (e.g., renamed or deprecated).
+
+- Package author :: They are responsible for writing Emacs Lisp
+                    packages of good quality.
+
+We advise package author to also take the recipe author role when
```
milkypostman: “the package author” or “package authors”

CONTRIBUTING.org
```diff
@@ -0,0 +1,213 @@
...
+possible.
+
+In the following, package and recipe authors will find guidelines to
+help them prepare a package and its recipe for inclusion into
+MELPA. Please follow the guidelines as closely as possible to speed up
+the inclusion process. Guidelines are not strict rules though and they
```
milkypostman: “Guidelines are not strict and evaluated on a case by case basis given proper justification.“

CONTRIBUTING.org
```diff
@@ -0,0 +1,213 @@
...
+required to explain why a guideline was not followed.
+
+** What to know before opening a pull request
+
+New recipe submissions should adhere to the following guidelines:
+
+- One recipe per pull request :: If you have several packages to
+     submit, open several pull requests. If the packages are very
+     closely related and maintained in the same repository (see below
+     for why it might not be a good idea though), you might open only
```
milkypostman: drop “though”

CONTRIBUTING.org
```diff
@@ -0,0 +1,213 @@
...
+     one pull request for them all; in this case, keep adding one
+     recipe per commit though.
```
milkypostman: Drop though

CONTRIBUTING.org
```diff
@@ -0,0 +1,213 @@
...
+
+- Reasonably innovative package :: MELPA provides a curated set of
+     Emacs Lisp packages, not an exhaustive list of every single Emacs
+     Lisp file ever created. As such, MELPA maintainers tend to only
+     accept packages not duplicating existing packages. Please try to
```
milkypostman: By default, MELPA maintainers will reject packages that duplicate functionality provided by existing packages.

CONTRIBUTING.org
```diff
@@ -0,0 +1,213 @@
...
+
+In the following, package and recipe authors will find guidelines to
+help them prepare a package and its recipe for inclusion into
+MELPA. Please follow the guidelines as closely as possible to speed up
+the inclusion process. Guidelines are not strict rules though and they
+can sometimes be bypassed: in these cases, the package author is
+required to explain why a guideline was not followed.
+
+** What to know before opening a pull request
+
```
milkypostman: Let’s add a paragraph here: 

Expect feedback! Part of the review process is having a MELPA maintainer look at your package in order to make sure it will work in MELPA and the Emacs package system. As part of this review MELPA maintainers may provide comments which may or may not be blockers for addition to MELPA.

DamienCassou:
I took care of all feedback I got.

DamienCassou:
@tarsius care to mark my changes as solving your feedback?

anyone else wants to review before this gets merged?

tarsius:
Maybe we should also add something like:

* If you later move the upstream repository, then please try to remember to update the recipe too.

* If you transfer ownership to someone else and both the old and the new location are on Github, then please use Github's repository transfer mechanism to move the repository instead of creating a new repository. This makes it easier for us to verify that the move is legit. Otherwise you should update the README at the old location to say that the repository has moved and provide a link to the new location.

DamienCassou:
> Maybe we should also add something like: […]

done

-- 
Damien Cassou

milkypostman:
Lets commit this if we haven’t yet. We been sitting on it too long.

PR #5631, opened 2018-07-18
new recipe for magit-rbr package

fanatoly:

This package tweaks magit to recognize `git rbr` rebases and use corresponding commands during the magit rebase sequence. This means that when you abort a rebase during a recursive rebase, magit will abort the rbr rather than a particular phase of rbr. This also adds recursive rebase as an option to the rebase popup.

For context git-rbr lives here: https://github.com/dropbox/git-rbr

https://github.com/fanatoly/magit-rbr

Maintainer

### Checklist

Please confirm with `x`:

- [x] I've used the latest version of [package-lint](https://github.com/purcell/package-lint) to check for packaging issues, and addressed its feedback: `package-lint` complains about redefining functions without the `magit-rbr` prefix. This is intentional and necessary for the package to function. Please, let me know whether this is ok for inclusion in MELPA.

### DIFF
--- a/recipes/magit-rbr
+++ b/recipes/magit-rbr
@@ -1,0 +1,1 @@
+(magit-rbr :fetcher github :repo "fanatoly/magit-rbr" :files ("magit-rbr.el"))
--- a/recipes/magit-rbr
+++ b/recipes/magit-rbr
@@ -1,1 +1,1 @@
-(magit-rbr :fetcher github :repo "fanatoly/magit-rbr" :files ("magit-rbr.el"))
+(magit-rbr :fetcher github :repo "fanatoly/magit-rbr")


tarsius:
Sadly I couldn't find any documentation about `git rbr` itself. Could you please locate some useful resource such as a blog post or announcement and link to that in your documentation.

fanatoly:
Thanks for taking a look! The closest I can find to documentation is this: https://github.com/dropbox/git-rbr/issues/3. Unfortunately, the tool lacks public documentation. I can submit a PR with some docs to the project. Is that something that you would consider blocking for inclusion in MELPA?

purcell:
This seems okay to merge in principle, but it looks like the code is redefining internal magit commands, which is really bad practice. You should instead be able to provide new commands and a minor mode, such that in the keymap for the minor mode you would shadow the bindings of the original commands.

tarsius:
I would be willing to merge https://github.com/magit/magit/compare/basic-rbr-support.

tarsius:
The lines in the Commentary section should begin with only two semicolons.

tarsius:
The years in the copyright line are wrong. You don't really need the created keyword. You shouldn't use a timestamp as version. There's usually an empty line before "this file is not part of GNU Emacs".

tarsius:
Break the `magit-define-popup-action` line into two.

tarsius:
`(message "Recursively Rebasing...")` downcase the second word.

tarsius:
> I would be willing to merge https://github.com/magit/magit/compare/basic-rbr-support.

ping @fanatoly 

If I remember correctly you can remove everything but `magit-rbr-rebase-recursive` thanks to my commit, which I will merge once you have adjusted and tested your package.

DamienCassou:

tarsius:
There are still outstanding issue. Mentioned earlier:

> The years in the copyright line are wrong. You don't really need the created keyword. You shouldn't use a timestamp as version. There's usually an empty line before "this file is not part of GNU Emacs".

I would also

* remove the `Created` header keyword, that's not really necessary when you can get the same information from `git log`.
* Remove the `Version` header keyword, and instead create a tag `0.1` once all the issues mentioned here have been addresses. I wouldn't use a timestamp as version identifier. Melpa ignores the `Version` header and consults the latest tag instead.

Most importantly though, and this is why this cannot be merged yet, the license has to be specified, the single line 

```
;; This file is free software...
```

isn't enough. It should be replaced with

```
;; This file is free software: you can redistribute it and/or modify
;; it under the terms of the GNU General Public License as published by
;; the Free Software Foundation, either version 3 of the License, or
;; (at your option) any later version.

;; This file is distributed in the hope that it will be useful,
;; but WITHOUT ANY WARRANTY; without even the implied warranty of
;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
;; GNU General Public License for more details.

;; You should have received a copy of the GNU General Public License
;; along with this file.  If not, see <https://www.gnu.org/licenses/>.
```

tarsius:
Ping @fanatoly

fanatoly:
@tarsius I updated the magit-rbr repo to reflect the comments in this thread. I've also tested the new configuration and things seem to work. If you're ok with the source repo, I can tag v0.1 and this can be merged.

tarsius:
Looks good. Thanks!

tarsius:
Oh, one more thing: Drop the `:files ("magit-rbr.el"` from the recipe. That's not necessary.

tarsius:
I have merged locally and pushed because I was unsure if "squash and merge" does the right thing (i.e. if it does "squash and rebase and fast-forward").

PR #5632, opened 2018-07-19
Add helm-jira

DeX3:

Helm-bindings for interacting with JIRA issues and BitBucket/Stash Pull requests.

https://github.com/DeX3/helm-jira

I'm the creator and maintainer of this package.

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/helm-jira
+++ b/recipes/helm-jira
@@ -1,0 +1,3 @@
+(helm-jira
+ :fetcher github
+ :repo "DeX3/helm-jira")


alphapapa:
Hi,

A little feedback:

In the defcustoms, you don't need `:initialize 'custom-initialize-set` or `:group 'helm-jira`.

noctuid:
> If a defcustom does not specify any :group, the last group defined with defgroup in the same file will be used. This way, most defcustom do not need an explicit :group.

I didn't know this before today. Seems like this would be a good thing to check for in package-lint.

alphapapa:
> Seems like this would be a good thing to check for in package-lint.

Occasionally it's necessary to specify it, but most of the time not.  Perhaps they would welcome a patch over there.  :)

DeX3:
Nice, thanks for the feedback, I updated the package.

purcell:
Thanks, particularly for your patience. This looks good, so merging now. You should consider adding `;;;###autoload` cookies for the various interactive commands that might be user-initiated entry points for the package.

purcell:
And: welcome to MELPA! :tada:

PR #5634, opened 2018-07-20
Added package crystal-playground.

jasonrobot:

Create a basic crystal-lang package to test code. The playground can easily be re-used or deleted.

https://github.com/jasonrobot/crystal-playground

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/crystal-playground
+++ b/recipes/crystal-playground
@@ -1,0 +1,1 @@
+(crystal-playground :fetcher github :repo "jasonrobot/crystal-playground")


purcell:
Thanks, particularly for your patience. Looks great, just a couple of small issues:
- `C-c LETTER` is reserved for end-user customisation, so [this](https://github.com/jasonrobot/crystal-playground/blob/f8f383eabd72a2ae6af73db63531c35d9e458d39/crystal-playground.el#L101) needs to change. `package-lint` should have told you about this when you ran it.
- `in-crystal-playground` doesn't match the package prefix. However, it'd be fine to make an exception for the more-idiomatic `with-crystal-playground`, so I'd suggest renaming that macro.
- When [matching filenames with regexps](https://github.com/jasonrobot/crystal-playground/blob/f8f383eabd72a2ae6af73db63531c35d9e458d39/crystal-playground.el#L209-L210), you should use `\\'`, not `$`, in order to match end-of-string rather than end-of-line.
- [This function](https://github.com/jasonrobot/crystal-playground/blob/f8f383eabd72a2ae6af73db63531c35d9e458d39/crystal-playground.el#L105) can almost certainly be replaced with an invocation of `locate-dominating-file`.
- Don't use `(if (not x) nil y)` or `(if (not x) y nil)` when you can write `(when x y)` and `(unless x y)` instead

Hope that helps!

purcell:
Nice. Merged! :-) Welcome to MELPA. :tada:

PR #5638, opened 2018-07-24
Add org-starter

akirak:

A configuration framework for org-mode

https://github.com/akirak/org-starter

I am the maintainer

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/org-starter
+++ b/recipes/org-starter
@@ -1,0 +1,4 @@
+(org-starter :repo "akirak/org-starter"
+     :fetcher github
+     :files (:defaults
+             (:exclude "counsel-org-starter.el" "helm-org-starter.el")))


casouri:
Looks cool! Why don't you advertise on Reddit? There are more people on Reddit than melpa watch list.

akirak:
Yes, I plan to do so, but I have to write a tutorial on the package. I am developing the package as I use it, so it is not easy to describe the whole feature set.

alphapapa:
Hm, this is interesting.  So basically, IIUC, it moves configuration out of `#+` in-buffer settings in each Org file?

akirak:
It has a lot of features, most of which are not documented in README yet. I have to create a wiki for extensive documentation. To name a few:

- Search a file from `org-starter-load-path` and configure variables with its absolute path.
- Define `org-agenda-files`, `org-refile-targets`, and `org-capture-templates` incrementally as you define files rather than in a centralized place. This will slow down the init time of Emacs, but I don't care much about that. 
- Run `:config` if and only if the file/directory exists. This is analogous to `use-package`.

My [local](https://github.com/akirak/emacs.d/tree/master/local) directory in emacs.d contains examples.

Compro-Prasad:
Is this PR getting any attention?

akirak:
@Compro-Prasad 
It will take some time until it gets merged, but I also have to add more documentation on its usage.

purcell:
Thanks, including for your patience!

Don't shim `if-let*` etc. [here](https://github.com/akirak/org-starter/blob/1343de823633960ddbec9f83628fad7f51eecec9/org-starter.el#L43-L47) - as the [elisp conventions](https://www.gnu.org/software/emacs/manual/html_node/elisp/Coding-Conventions.html) say, `"If a package needs to define an alias or a new function for compatibility with some other version of Emacs, name it with the package prefix, not with the raw name with which it occurs in the other version."`

Otherwise this looks good, so because we've kept you waiting, I'll go ahead and merge this now, and you can fix the above issue separately.

PR #5639, opened 2018-07-26
Update contextual package source

e-user:
I'm the original author of the `contextual` package. It was originally released under copyright of my previous employer - however that company doesn't exist anymore and I've added a new feature since, so I'd like to change the source of the package to my personal GitHub account where the new, active fork lives.

### DIFF
--- a/recipes/contextual
+++ b/recipes/contextual
@@ -1,1 +1,1 @@
-(contextual :fetcher github :repo "lshift-de/contextual")
+(contextual :fetcher github :repo "e-user/contextual")


purcell:
Hmm, did you lose access to the previous github account? It should have been possible to transfer the github repo to the new user. I ask because I need to think about protecting users from packages getting hijacked.

(Also, sorry for the slow response.)

e-user:
Hi @purcell, thanks for picking this up.  

I just did some digging and could indeed recover the account. The ownership of the original repository is now with me again.

purcell:
Cool. What would you like to do with the recipe? If you decide to transfer the repo then everything will continue to work seamlessly, and then updating the recipe is just a nice tidy-up.

e-user:
I already did transfer the repo. Sorry, I should've been more explicit. Please see for yourself, `lshift-de/contextual` now redirects to `e-user/contextual`. Is there anything else that needs to be done with it..?

purcell:
Great - merged, then! Thanks.

PR #5647, opened 2018-07-28
Add Easy Frame

onixie:
Easy Frame provides key bindings to manipulate frames easily.

Easy Frame binds arrow keys [+ modifier keys] for various frame placement commands in an X window environment.

https://github.com/onixie/easy-fwb

I'm the maintainer.

**None needed**

### Checklist

### DIFF
--- a/recipes/easy-frame
+++ b/recipes/easy-frame
@@ -1,0 +1,1 @@
+(easy-frame :repo "onixie/easy-fwb" :fetcher github :files ("easy-frame.el"))


alphapapa:
Hi,

A little feedback:

1.  Loading the package shouldn't do `(add-hook 'window-setup-hook 'easy-frame-mode)`.  That is for users to do in their config.
2.  More documentation is needed.  As it is now, users will likely not understand what the package does, and won't bother to find out.

purcell:
Thanks @onixie - this still needs a little more documentation please. I have no idea what this package is for, but I'm sure it's useful. :-)

DamienCassou:

tarsius:
I am closing this because @onixie has not responded to your suggestions and because I think that these libraries are of limited use to anybody beside their author. In particular I think that the key bindings are too much tailored to the authors preference. Some of the libraries define some commands which could be of use to other users though, provided some documentation is added. I am not against adding this in principal, but that will need some work and so far we got silence instead.

PR #5653, opened 2018-07-29
Add isolate recipe

casouri:

This is a surrouding tool.
It features powerful and easy customization system.
You can create complex regexp rules easily.

https://github.com/casouri/isolate

Author

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/isolate
+++ b/recipes/isolate
@@ -1,0 +1,1 @@
+(isolate :repo "casouri/isolate" :fetcher github)


akirak:
This looks cool. I am a former vim user, and I have wanted an equivalent to [surround](https://github.com/tpope/vim-surround) **without evil**. This package may fill in the need.

casouri:
Talking about surround without evil, there is embrace:

https://github.com/cute-jumper/embrace.el

Althought I’ll quietly say that mine is better.

purcell:
Thanks, and sorry for the delay. Quick notes:

- "flexable" should be "flexible", "surrouding" should be "surrounding". (Maybe try `flyspell-prog-mode`?)
- Why not write `isolate--append` and `isolate--push` as `defsubst` instead of macros? Your macro expand their arguments multiple times, which could lead to multiple evaluation - read the macros section of Practical Common Lisp (free online) to understand what I mean.

In any case, I'll merge this now, but you should address the above points separately.

casouri:
Thanks! I'll fix those issues. I just got my flyspell back to work again recently... I will scan for any other typos.

casouri:
Fixed.

PR #5655, opened 2018-07-29
Add package hack-mode, a major mode for hacklang

jra3:

A new major mode for [hacklang](https://hacklang.org/)

https://github.com/hhvm/hack-mode

Author and maintainer

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/hack-mode
+++ b/recipes/hack-mode
@@ -1,0 +1,2 @@
+(hack-mode :repo "hhvm/hack-mode"
+           :fetcher github)


purcell:
Thanks. Couple of issues here:

- You've got the dependencies a bit muddled up here, because a major mode should not be depending on something opinionated like `lsp-hack`. Let your users install that separately, don't force it on them. Major modes are primarily for indentation and font lock.
- You're trampling the `auto-mode-alist` entry for `.php` whenever this is installed, but that's bad practice unless your package is unambiguously the community-wide default for that extension. Instead, suggest that users add that entry themselves, or simply do it when this feature is loaded.
- Don't [modify your own major mode hook](https://github.com/hhvm/hack-mode/blob/a41009780b7eb940585602a0b0005abb1dd32fb8/hack-mode.el#L639)
- [These old-fashioned lines](https://github.com/hhvm/hack-mode/blob/a41009780b7eb940585602a0b0005abb1dd32fb8/hack-mode.el#L627-L632) should be generated for you by `define-major-mode`, so you should be able to delete them.

Hope that helps!

purcell:
Also, sorry about the slow turnaround.

jra3:
I appreciate the feedback. It was really valuable and pointed out a bunch of problems.

I've made the suggested changes in the source repo

purcell:
Looks great, thanks - merged!

PR #5660, opened 2018-07-30
Add dap-mode recipe

yyoncho:

`dap-mode` is an emacs integration for Debug Adapter Protocol which is the protocol used by VScode/Eclipse Che for integration with the debugger. Together with LSP mode the package aims to provide full IDE experince for Emacs. Currently supports Java debugger.

https://github.com/yyoncho/dap-mode

Author/Maintainer

**None needed** 

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/dap-mode
+++ b/recipes/dap-mode
@@ -1,0 +1,1 @@
+(dap-mode :repo "yyoncho/dap-mode" :fetcher github)


alphapapa:
Hi,

I don't have the expertise to use or review the package, but one suggestion: You probably don't need to autoload so many commands, just the one that would be the entry point into the package.

yyoncho:
@alphapapa Thanks for your comment I will fix it. This is my first package, so feel free to provide any advice.

purcell:
This looks good, but it's more of an interface than a mode as the name implies, so can I please suggest renaming it to `dap`? This would just involve renaming/editing `dap-mode.el` and the recipe.

Also, thanks for your patience.

yyoncho:
@purcell thanks for the feedback I don't mind renaming the file but I would prefer to keep the current name in order to be consistent with the related `lsp-mode` since the code have similar architecture: `dap-mode.el` -> `lsp-mode.el`,  `dap-ui.el` -> `lsp-ui.el`, `dap-java.el` -> `lsp-java.el`, etc. Furthermore, all of the ui related code is in a separate file(`dap-ui.el`) and could potentionally grow into separate package(similar to what has happened to `lsp-mode`). What do you think?

purcell:
Yeah, I think `lsp-mode` is misnamed, so I wouldn't necessarily recommend copying that scheme, but I guess I can live with either name. I'll go ahead and merge now.

purcell:
Done - welcome to MELPA! :tada:

prabhu-madanagopal:
Thanks, can't wait to package-install.

PR #5661, opened 2018-07-31
Add reazon

nickdrozd:

Implementation of *miniKanren* logic programming language

https://github.com/nickdrozd/reazon

Creator, maintainer

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/reazon
+++ b/recipes/reazon
@@ -1,0 +1,1 @@
+(reazon :repo "nickdrozd/reazon" :fetcher github)


purcell:
Thanks, and sorry for the delay. This looks very cool. Quick notes:

- "for Emacs" is redundant and should be deleted. All elisp is for Emacs. :-)
- [This](https://github.com/nickdrozd/reazon/blob/eb15897fd557f765058306d09e8d82168b73dda0/reazon.el#L38-L40) implies you've forgotten to declare a package dependency on `cl-lib`, and you should also just call `cl-gensym` instead of aliasing it as `gensym`.
- You might like to add an `;;;###autoload` or two.

Hope that helps.

nickdrozd:
* `gensym` is a builtin in 26, but not prior versions. This code can (I think) be run in 24 and 25 except for the use of `gensym`, so I added that expression as a compatibility layer. Is there a better way to do that? I don't want to use `cl-gensym`. (This is what I meant with https://github.com/purcell/package-lint/issues/109.)

* I don't understand autoloads, where should they go?

nickdrozd:
If you don't mind, I'd like to keep the header line as it is for now. "miniKanren for $language" seems to be the standard way to label these things in the miniKanren world (see https://github.com/clojure/core.logic/blob/master/project.clj#L2 and https://github.com/miniKanren/Racket-miniKanren).

`package-lint` says "Including "Emacs" in the package description is usually redundant", but in this case I think it isn't :)

purcell:
> gensym is a builtin in 26, but not prior versions. This code can (I think) be run in 24 and 25 except for the use of gensym, so I added that expression as a compatibility layer. Is there a better way to do that? I don't want to use cl-gensym. (This is what I meant with purcell/package-lint#109.)

Aliasing built-ins is disallowed by the [elisp coding conventions](https://www.gnu.org/software/emacs/manual/html_node/elisp/Coding-Conventions.html), so the better way is to simply use `cl-gensym`. Adding a package dependency on `(cl-lib "0.6")` will let this work seamlessly in older Emacsen, so this is optimal.

> 
> I don't understand autoloads, where should they go?

If there are any functions/commands you'd like people to be able to use without having to explicitly load the file first, you can mark them as autoloaded by putting an `;;;###autoload` cookie comment before them. Not really a big deal in the case of a programmatic library like this, I guess.

purcell:
> I'd like to keep the header line as it is for now. "miniKanren for $language" seems to be the standard way to label these things in the miniKanren world

Sure. :-)

nickdrozd:
Okay, I took out the compatibility alias, bumped the version requirement to 26, and added a note in the readme about adding a `gensym` alias in versions 24 and 25. (The note also mentions that versions less than 24 absolutely will not work.)

Aside from an unimportant profiling function, there are no commands in the package, so I don't think autoloads apply. This might change in the future, for instance if a major mode gets added.

purcell:
Super, thanks - merged now. Welcome to MELPA! :tada:

PR #5670, opened 2018-08-03
Add magik-mode recipe

roadrunner1776:

Provides a major mode for editing Smallworld (GIS) Magik files.

https://github.com/roadrunner1776/magik

Revived this old Emacs 19 package since I need it.

**None needed**. Package repository above is mine.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/magik-mode
+++ b/recipes/magik-mode
@@ -1,0 +1,2 @@
+(magik-mode :fetcher github
+	    :repo "roadrunner1776/magik")


purcell:
Thanks, and sorry for the delay in getting to this. Some notes:

- All the files' headers should please follow the standard boilerplate layout, as `M-x auto-insert` would insert into an empty file.
- In particular, there's no mention of a license, so we can't redistribute this code in its current form.
- Replace the now-obsolete `cl` with `cl-lib` when you can, declaring a `Package-Requires` dependency on that package.
- [These vars](https://github.com/roadrunner1776/magik/blob/5c7f10c997842d63ead52b521049d3d19b6909d3/magik.el#L20-L21) seem spurious, not being defined by any required or related package. Maybe they should be properly declared locally with a `magik-` prefix instead?
- Use `define-derived-mode` to declare `magik-mode` rather than doing it the old-fashioned way.
- Since this package declares a major mode, and also since its name is very similar to that of `magit`, this should instead be called `magik-mode`.
- [This `defvar`](https://github.com/roadrunner1776/magik/blob/5c7f10c997842d63ead52b521049d3d19b6909d3/magik.el#L13) should probably be a `require` instead.

Hope that helps!

tarsius:
> Since this package declares a major mode, and also since its name is very similar to that of `magit`, this should instead be called `magik-mode`.

Due to the latter I agree, but the former by itself would not be a good reason by itself. I cannot find any documentation about it but the convention is "just because there already exist packages named `*-mode` and because your package defines a major mode, that's no reason to name the package `*-mode`".

tarsius:
Not sure about `prog-mode`s though -- maybe those are supposed to have `-mode` in their name.

roadrunner1776:
- Changed the headers to the 'standard boilerplate layout' 
- Added a `LICENSE` file
- Renamed the package to `magik-mode`
- Now using `define-derived-mode`

purcell:
Great, merged now! Welcome to MELPA. :tada:

tarsius:
This package has a soft dependency on `mpde-env` but no package in Melpa provides that. @roadrunner1776, do you know what that does and where to get it? If not I would recommend removing all of `(if (featurep 'mpde-boot) ...)`.

roadrunner1776:
@tarsius `mpde` was part of the original package. I've removed the references to it. Thanks! Let me know if you have other suggestions.

tarsius:
`magik-cb.el` now requires `magik-cb-ac` like so:

```emacs-lisp
(eval-after-load 'autocomplete
  '(require 'magik-cb-ac))
```

but that feature is not provided by any library.

@roadrunner1776 Please fix that.

roadrunner1776:
> `magik-cb.el` now requires `magik-cb-ac` l
> but that feature is not provided by any library.
> 

@tarsius That file somehow got corrupted on the git remote. It has been fixed, thanks for noticing!

tarsius:
Thanks but I think you forgot to push the change.

tarsius:
Never mind I just noticed that you fixed it by modifying the latest commit. In such cases I have to intervene manually to start updating the package on the Emacsmirror again and I only do so sporadically. That's why I didn't see the fix.

PR #5671, opened 2018-08-04
Add dtk recipe

dtk01:
Facilitates reading diatheke-accessible material, in Emacs. 

https://github.com/dtk01/dtk

Author and maintainer of the package

**None needed**

### Checklist

x The package is released under a GPL-Compatible Free Software License.
x I've read CONTRIBUTING.md
x I've used the latest version of package-lint to check for packaging issues, and addressed its feedback
x My elisp byte-compiles cleanly
x M-x checkdoc is happy with my docstrings
x I've built and installed the package using the instructions in CONTRIBUTING.md

### DIFF
--- a/recipes/dtk
+++ b/recipes/dtk
@@ -1,0 +1,1 @@
+(dtk :fetcher github :repo "dtk01/dtk")


alphapapa:
Here's a bit of feedback that I may add to (there's a lot of code in the package):

* Instead of your own `dtk-string-trim-whitespace` function, you could use `string-trim` from `subr-x`, which is built-in to Emacs.

purcell:
Thanks, and sorry for the delayed response. Great feedback from @alphapapa above. I'd add the following:

- Instead of an empty `defvar` plus a `setq`, as [here](https://github.com/dtk01/dtk/blob/63d7d77bea922c7f9e4a1b30c4b6657ecb33bddd/dtk.el#L20-L21), you probably want `defconst`. And with a docstring. :-)
- You've marked above that `checkdoc` is happy with your docstrings, but I'm certain it isn't. :-)
- If you pass it the right flags, `completing-read` can [check for you that the input is not empty](https://github.com/dtk01/dtk/blob/63d7d77bea922c7f9e4a1b30c4b6657ecb33bddd/dtk.el#L323-L327).
- You're using functions from `dash`, e.g. [`-let`](https://github.com/dtk01/dtk/blob/63d7d77bea922c7f9e4a1b30c4b6657ecb33bddd/dtk.el#L438), but haven't declared a package dependency on it.
- You should `require` libraries you use, such as `cl-lib`.

Hope that helps!

DamienCassou:
What is the status of the PR?

purcell:
> Are you suggesting something like `(completing-read "choose" '("1" "2" "3") nil t nil nil '(nil))`?

Something like that, though I'm not sure which exact args you need.

I still see a few issues, some of which `package-lint` should really have reported to you when you ran/re-ran it:

- The bare package names `cl-lib dash` in `Package-Requires` should instead be of the form `(cl-lib "0.6")` etc.
- You'll need a newer Emacs than 24 for `subr-x`.
- You're using the `s` library but you don't declare a package dependency on it, or `require` it.
- Instead of the clunky `dtk-books-regexp`, just use `regexp-opt`
- `M-x checkdoc` will help you fix up the few extremely long docstrings.
- When there's no `else` form to an `if`, you can use `when` instead to avoid needing a `progn`.

thomp:
@purcell: Thanks for taking the time to take another look at this. The above issues and suggestions are addressed now. Checkdoc and package-lint seem to be content as well...

DamienCassou:
@purcell do you want to have another look?

purcell:
Looks good - merging now. Thanks for your patience! :-)

PR #5677, opened 2018-08-09
fix ordering in builder Dockerfile

milkypostman:
to satisfy the ordering we move curl to the first in the list.

### DIFF
--- a/docker/builder/Dockerfile
+++ b/docker/builder/Dockerfile
@@ -1,5 +1,5 @@
 FROM debian:unstable
 RUN apt-get update
-RUN DEBIAN_FRONTEND=noninteractive apt-get --yes install emacs25-nox git make mercurial ruby texinfo curl
+RUN DEBIAN_FRONTEND=noninteractive apt-get --yes install curl emacs25-nox git make mercurial ruby texinfo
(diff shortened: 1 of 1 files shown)


Silex:
I'm a fan of multilines lile https://github.com/Silex/docker-emacs/blob/master/templates/ubuntu/18.04/Dockerfile for long lists of dependencies :wink: 

I discovered that alphabetically sorting things where it makes sense is quite a time saver, no need to think where to go.

Silex:
p.s. your commit message should start with `Fix` instead of `fix` :stuck_out_tongue_closed_eyes:

milkypostman:
Then I’d have to hit the shift key! I’d rather save some time.

Silex:
I guess you were joking, but just in case some future readers fall on this, here's some lecture :smiley: 

https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html 
https://github.com/erlang/otp/wiki/writing-good-commit-messages
https://chris.beams.io/posts/git-commit/

milkypostman:
I prescribe to this professionally and should take the same care with this project. 

Although in general I’m not a huge fan of capitalization. 

Good links though.

PR #5678, opened 2018-08-09
add recipe for zeno-theme 

jbharat:

It's a new theme, dark and based on variations of blue colour.

https://github.com/jbharat/zeno-theme

I am the maintainer.

### Checklist

Please confirm with `x`:


purcell:
Thanks, and sorry for the delay. Quick notes:

- Don't use camel or snake case in elisp, like `slantType` or `atlantis_green`, `fullBlack`. Use `kebab-case` instead. And certainly don't use a mix!
- Don't set font families directly, e.g. [here](https://github.com/jbharat/zeno-theme/blob/master/zeno-theme.el#L213). It doesn't work at all well across platforms. Instead, `:inherit` the built-in `'variable-pitch` face if that's the effect you want.

Looks good otherwise. :-)

jbharat:
Thanks. You are right. I will make the  changes as per your comments.

purcell:
Cool, just ping me here when done, and I'll take another look.

jbharat:
I have changed all variables to kebab case and removed font family

purcell:
Much nicer! :-)

Merging now.

PR #5683, opened 2018-08-11
Add idle-highlight-in-visible-buffers-mode

ignacy:

idle-highlight-in-visible-buffers works exactly like idle-highlight but in all visible buffers.

https://github.com/ignacy/idle-highlight-in-visible-buffers-mode

Maintainer

**None needed**

### Checklist

Please confirm with `x`:


purcell:
Thanks, and sorry for the delay. This looks nice, and I'll merge it now, but you should definitely be using the `'highlight` face rather than `'region`. Welcome to MELPA! :tada:

PR #5684, opened 2018-08-11
Add recipe for comb

cyrus-and:

Comb is a native Emacs Lisp solution to search, browse and annotate occurrences of regular expressions in files. The interactive interface allows to perform an exhaustive classification of all the results to rule out false positives and asses proper matches during manual static analysis.

https://github.com/cyrus-and/comb

Author and maintainer.

The package summary is 59 characters long instead of the 50 required by package-lint, but there are plenty of packages that exceed this limit in MELPA. Since it's already quite condensed I would like to keep it as it is, but if it's a problem I can bring it down to 50.

Also package-lint and checkdoc have been run against the main `.el` file only, I don't know what are the best practices here, e.g., the commentary and the license are not present in the other files.

### Checklist

Please confirm with `x`:

- [x] I've used the latest version of [package-lint](https://github.com/purcell/package-lint) to check for packaging issues, and addressed its feedback (**see above**)

### DIFF
--- a/recipes/comb
+++ b/recipes/comb
@@ -1,0 +1,1 @@
+(comb :repo "cyrus-and/comb" :fetcher github)


alphapapa:
Hi,

This looks like a very interesting and useful package! I suspect it might
be able to help MELPA maintainers review new packages.

A few things I noticed:

1. Generally you should run checkdoc and package-lint on all the elisp
files. Some of the issues reported might not be applicable to every file,
but most will be.

2.  I'm curious as to why you implemented your own event and input-handling
loop instead of using a keymap and letting Emacs handle input. Generally it
doesn't seem like a good idea and shouldn't be necessary, but maybe you're
doing something fancy that requires it.  Does your loop properly handle
global key bindings, special key sequences like C-g, etc?  Also, packages
don't generally handle customization of key bindings through defcustom
variables, but with standard keymaps and key-binding commands. Doing it
that way will generally make it harder for users to customize and integrate
your package.
On Aug 11, 2018 10:57 AM, "Andrea Cardaci" <notifications@github.com> wrote:

Brief summary of what the package does

Comb is a native Emacs Lisp solution to search, browse and annotate
occurrences of regular expressions in files. The interactive interface
allows to perform an exhaustive classification of all the results to rule
out false positives and asses proper matches during manual static analysis.
Direct link to the package repository

https://github.com/cyrus-and/comb
Your association with the package

Author and maintainer.
Relevant communications with the upstream package maintainer

The package summary is 59 characters long instead of the 50 required by
package-lint, but there are plenty of packages that exceed this limit in
MELPA. Since it's already quite condensed I would like to keep it as it is,
but if it's a problem I can bring it down to 50.

Also package-lint and checkdoc have been run against the main .el file
only, I don't know what are the best practices here, e.g., the commentary
and the license are not present in the other files.
Checklist

Please confirm with x:

   - The package is released under a GPL-Compatible Free Software License
   <https://www.gnu.org/licenses/license-list.en.html#GPLCompatibleLicenses>
   .
   - I've read CONTRIBUTING.org
   <https://github.com/melpa/melpa/blob/master/CONTRIBUTING.org>
   - I've used the latest version of package-lint
   <https://github.com/purcell/package-lint> to check for packaging issues,
   and addressed its feedback (*see above*)
   - My elisp byte-compiles cleanly
   - M-x checkdoc is happy with my docstrings
   - I've built and installed the package using the instructions in
   CONTRIBUTING.org
   <https://github.com/melpa/melpa/blob/master/CONTRIBUTING.org>

------------------------------
You can view, comment on, or merge this pull request online at:

  https://github.com/melpa/melpa/pull/5684
Commit Summary

   - Add recipe for comb

File Changes

   - *A* recipes/comb
   <https://github.com/melpa/melpa/pull/5684/files#diff-0> (1)

Patch Links:

   - https://github.com/melpa/melpa/pull/5684.patch
   - https://github.com/melpa/melpa/pull/5684.diff

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/melpa/melpa/pull/5684>, or mute the thread

cyrus-and:
> This looks like a very interesting and useful package! I suspect it might
> be able to help MELPA maintainers review new packages.

Thanks. That would be great!

> 1. Generally you should run checkdoc and package-lint on all the elisp
> files. Some of the issues reported might not be applicable to every file,
> but most will be.

OK, then I already did that.

> I'm curious as to why you implemented your own event and input-handling
> loop instead of using a keymap and letting Emacs handle input.

I wanted an experience similar to `ispell` which is implemented using a `read-key` loop.

I tried with the keymap approach, specifically I tried with an empty (sparse) keymap filled only with comb-related keybinding but then it didn't work well when a prompt is presented to the user during the interactive session because said keymap would lack self-insertion keybindings. Of course one could work around that by setting the default keymap as a parent but I didn't want to silently ignore non relevant keybindings and let them perform the default action. Another solution would be to switch back to the global keymap temporarily:

```elisp
(use-global-map global-map)
(read-string "Type something: ")
(use-global-map comb-map)
```

> Does your loop properly handle
> global key bindings, special key sequences like C-g, etc?

No but I can do that I guess, what are the others? I can only think of C-g, in this context.

> Also, packages
> don't generally handle customization of key bindings through defcustom
> variables, but with standard keymaps and key-binding commands. Doing it
> that way will generally make it harder for users to customize and integrate
> your package.

Yeah, I struggled with this one. The fact is that I don't want to expose comb *commands* to the users as they only make sense during the interactive menu (in fact they are private, non-interactive functions), so something like the following is not acceptable in this case.

```elisp
(global-set-key (kbd "C-c n") 'comb--menu-next)
```

Another blocker is that I'd like to generate the help buffer programmatically and using a keymap is not clear to me how to easily do that.

I'd love to hear your thought about this.

noctuid:
> And in doing so users should be able to use simple keybindings like n or p but not defined over the global keymap like what happens out-of-the-box with a minor mode map or using set-transient-map because I don't want that a wrong key performs the default function or exits the interactive session.

Using a minor mode works fine for `git-timemachine-mode`, for example, and I think it would work fine for this package. I don't think you should worry about the user making mistaken keypresses. If the user isn't aware they're in a comb session, that's the issue, and with a readonly buffer, there's less potential for mistakes even in that case. It's perfectly possible that a user might want to use global non-navigation commands during a session, and I think easily allowing for that should be prioritized over preventing unlikely mistakes. Also, I think the buffer should be the same major mode as the original,.

> What are the actual issues of interfering with the echo area in this context?

As soon as something else calls `message`, the prompt disappears. If you use something like lv, the help message won't ever be unintentionally hidden. In my case, I probably need to track down a timer or hook that is messaging and prevent that, but lv is a nice way to prevent such a problem without any downsides (that I'm aware of).

> I'm not an evil user, but I just tried and it works fine apparently. (Apart from not being able to quit the buffers with q...) What's the problem in your case?

There are precedence issues. For example, I have `h` bound to `help-command` in normal state, and that is called instead of the default `h`. It's also not possible to customize comb's keybindings specifically for evil without a keymap. Currently, I'd actually have to disable evil for a comb session, and I don't think that's ideal.

> I thought of that and in fact there's a way to suspend the comb session (z) to look around and do stuff.

I prefer to have some global keybindings available in all buffers; the only time I'm fine with these keybindings being unavailable is in the minibuffer. I wouldn't want to have to use some extra keybinding to suspend a session. I think the loop is okay for really simple scenarios like `query-replace`. Generally, I'd prefer a minor mode, especially once there are a lot more commands and a session might take a while. If you think that mistaken keypresses would be an issue for some reason, there are other potentially other ways you could prevent it. Where do you see mistaken keypresses becoming an issue?

> They shouldn't, that's the point, and if I override the global map I face this problem.

Minor mode keybindings shouldn't have any effect for `read-string` or whatever. Is there some reason you would need to use `use-global-map`?

> Hm, I don't know, if I see an interactive foo-function I (as user) conclude that's legit for me to do M-x foo-function without any further consideration.

This assumption would be incorrect for hundreds of builtin commands and probably thousands when you consider packages. I can call `org-todo` in an emacs lisp buffer from `M-x`, but that doesn't mean that it makes sense to.

Comb seems like a pretty cool package that I might find a lot of good uses for. For me personally, I think using comb as it is currently could be jarring in some cases, but I have a pretty personalized workflow. Take this all with a grain of salt as a lot of users may not care that they can't access global keybindings or that keybindings for comb aren't made in the usual way. They may always go through a comb session in one go without ever using non-comb commands. If you think that this method has advantages that are impossible with a minor-mode, then it may be better to keep it. I think you could probably switch to using a minor mode and for a similar experience.

cyrus-and:
> Using a minor mode works fine for git-timemachine-mode, for example, and I think it would work fine for this package.

Yes, but the main difference here is that it switch to a new buffer, but I get the idea.

> Also, I think the buffer should be the same major mode as the original

What do you mean by that? The buffer mode is currently unchanged.

> As soon as something else calls message, the prompt disappears.

Yes, which anyway never happens right now because any user action *refreshes* the message.

> There are precedence issues. For example, I have h bound to help-command in normal state, and that is called instead of the default h.

I don't get how this can happen... so if you do `M-: RET (read-key-sequence nil) RET h` you don't get "h"? I'm more interested in why [`use-local-map`][1] is not honored in vanilla evil mode:

[1]: https://github.com/cyrus-and/comb/blob/700cada35ec9c71d6dc088c781d10e4b3d012820/comb-common.el#L80-L84

> It's also not possible to customize comb's keybindings specifically for evil without a keymap. Currently, I'd actually have to disable evil for a comb session, and I don't think that's ideal.

You can do that via `M-x customize-group RET comb RET`.

> Where do you see mistaken keypresses becoming an issue?

> Is there some reason you would need to use use-global-map?

I didn't like mixing comb commands with the rest, especially if they are bound to *simple* keys like `p` or `n`, I just don't find it coherent. But I understand the need of seamlessly doing something else. The incoherence here is that users should bind comb commands in such a way that there's no conflict with those existing global keybindings in order to allows them to do *something else*.

---

Anyway, let me play with this idea of the minor mode and do some experiments. I like the purpose of `lv` but I don't like that it randomly disappears, grows or gets replaced by another buffer. I'd need a more solid way to display the dashboard.

Another concern regards the keymap, take for example `git-timemachine-mode` it defines the minor mode keymap like this:

```elisp
:keymap
'(("p" . git-timemachine-show-previous-revision)
  ("n" . git-timemachine-show-next-revision)
  ("g" . git-timemachine-show-nth-revision)
  ("t" . git-timemachine-show-revision-fuzzy)
  ("q" . git-timemachine-quit)
  ("w" . git-timemachine-kill-abbreviated-revision)
  ("W" . git-timemachine-kill-revision)
  ("b" . git-timemachine-blame))
```

Users can then add custom keybindings (I guess) like this:

```elisp
(define-key git-timemachine-mode-map (kbd "x") 'git-timemachine-quit)
```

But what if they want to replace the whole keymap?

Is there a better way than go and set `nil` for every entry?

```elisp
(define-key git-timemachine-mode-map (kbd "q") nil)
```

cyrus-and:
I'm trying a slightly different approach, instead of using a minor mode I simply use custom buffer with a keymap attached to it, reasons are:

- a minor mode cannot be global because I don't want comb keybindings everywhere;

- a minor mode cannot be buffer-local because switching the buffer when jumping to the next occurrence results in a buffer with said minor mode disabled, thus no keybindings;

- the keymap is now exposed a simple variable so a complete override is as simple as (this answers my own question: "But what if they want to replace the whole keymap? Is there a better way than go and set nil for every entry?"):

```elisp
(setq comb-keymap (make-sparse-keymap))
(define-key comb-keymap (kbd "x") 'comb-quit)
(define-key comb-keymap (kbd "?") 'comb-help)
```

So I'm following an approach similar to `git-timemachine-mode` where I visit the buffer I want to show, but I don't select it, instead I copy its content in the comb buffer and apply the proper major mode. (It's possible that using [indirect buffers][1] may be a better solution instead of copying.)

[1]: https://www.gnu.org/software/emacs/manual/html_node/emacs/Indirect-Buffers.html

Finally, instead of using `message` or `lv-message`, I'm trying to put everything in a customized modeline. (Worst case, notes may appear transiently in the echo area if a single line is not enough.)

The result seems neat so far, I'm going to push a new branch soon if any of you is interested in giving it a try.

noctuid:
> What do you mean by that? The buffer mode is currently unchanged.

I just realized that comb opens `#filename#` in fundamental mode for me.

> Yes, which anyway never happens right now because any user action refreshes the message.

The example cases (e.g. message run from timer)I  gave are not user actions; this does happen for me almost immediately.

> I'm more interested in why use-local-map is not honored in vanilla evil mode

The local keymap is usually the major mode keymap. Evil would not work correctly if it let major or minor mode keybindings clobber its keybindings. The `(current-local-map)` is close to the lowest precedence keymap (only next to the global map). Evil's keymaps are in `emulation-mode-map-alists` which has much higher precedence.

> You can do that via M-x customize-group RET comb RET.

What I meant is that you cannot customize the keybindings for evil like you would normally; you just have to disable it.

> I didn't like mixing comb commands with the rest, especially if they are bound to simple keys like p or n, I just don't find it coherent. The incoherence here is that users should bind comb commands in such a way that there's no conflict with those existing global keybindings in order to allows them to do something else.

If you needed to prevent other keybindings from being active, using a higher precedence keymap would be better (especially since simple keys for other packages are usually bound in minor mode maps), but like I said, I don't think you need to worry about this too much. For the global keys I want to keep, with evil it would be very easy to prevent conflicts. For non-evil users in editable buffers, letter keys are likely bound to self insertion, so there generally won't be conflicts/issues (especially in a readonly duplicate buffer). I don't know if comb is meant to be used in special buffers that bind letter keys, but you could potentially bind all letter keys to `ignore` by default for that case if you wanted to.

> I like the purpose of lv but I don't like that it randomly disappears, grows or gets replaced by another buffer. I'd need a more solid way to display the dashboard.

I haven't had these problems. Maybe you should create an issue on the hydra repository and just stick with messaging for now.

> I'm trying a slightly different approach, instead of using a minor mode I simply use custom buffer with a keymap attached to it,

You could just enable the minor mode in the comb buffer or enable it for (copies of) all buffers with matches. Since the user wouldn't actually manually enable the minor mode, I don't think it matters which approach you go with except in terms of keymap precedence.

> I'm trying to put everything in a customized modeline

I think that the modeline may not be able to fit all information well.

cyrus-and:
> Pushed the branch.

Merged into master.

So is there something to be addressed in order for this package to be included in MELPA?

purcell:
Thanks, including for your patience. Great discussion above. I've had a quick skim of the code and have the following few notes:

- Yes, a shorter package description would be helpful.
- Killing [all buffers starting `*Comb`](https://github.com/cyrus-and/comb/blob/c544b6c56e9769c9fc7c7422b7068e63c9a10da3/comb-common.el#L126) might kill unrelated buffers, e.g. `*Combined whatever*`.
- Splitting this related code into multiple files makes the code harder to follow, not easier.  I have learned this lesson many times over in elisp projects, and would suggest you consider putting most of the code in one file.
- Generally you should require `'cl-lib`, not `'cl-macs`. The latter is not available in the standalone `cl-lib` package, and should be considered an implementation detail.

Of the above points, only the first needs action before merging. Hope that helps!

cyrus-and:
Thank you @purcell for the notes (and for catching up with all those PR!).

> Yes, a shorter package description would be helpful.

It's now "Interactive grep tool for manual static analysis" (48 characters).

> Killing all buffers starting *Comb might kill unrelated buffers, e.g. *Combined whatever*.

Good catch, changing the regexp to `"^\\*Comb[*:]"` should be enough, keeping track of the buffers globally might be overkill.

> Splitting this related code into multiple files makes the code harder to follow, not easier. I have learned this lesson many times over in elisp projects, and would suggest you consider putting most of the code in one file.

I really struggled about this one and ultimately decided to split, I didn't find any proper elisp guidelines, but I guess I shouldn't try to force other progamming languages' best practices into elisp. I'll give your suggestion a try.

> Generally you should require 'cl-lib, not 'cl-macs. The latter is not available in the standalone cl-lib package, and should be considered an implementation detail.

Done!

purcell:
Nice, thanks - merged!

Issue #5682, opened 2018-08-10
Regarding non-https, especially on Windows


bradyt:
Regarding commit a46d878 which closed issue #4820.

In that issue, @johnmorrisbeck says,

> For windows you have to install melpa with http only.

This is not true. Options for allowing https via emacs on windows include:

- Using Cygwin emacs (many more things will work out of the box this way anyways, such as `M-x grep`)
- Following instructions at https://www.gnu.org/software/emacs/manual/html_node/emacs-gnutls/Help-For-Users.html
- It appears since emacs 25, gnu.org is providing the dependencies: https://ftp.gnu.org/gnu/emacs/windows/emacs-25/

I don't think melpa should be so tasked to encourage non-https. At best, it should be a comment. The snippet at https://melpa.org/#/getting-started is getting a bit complicated, and new users may likely not read it before copy-pasting. Even the conditioning on emacs < 24 may be just a comment, 23 is pretty old now.

If I were on Windows, and did not have https via emacs, I'd be more likely to download a package from github via the browser, where there should be https.

The snippet from https://melpa.org/#/getting-started:

``` emacs lisp
(require 'package)
(let* ((no-ssl (and (memq system-type '(windows-nt ms-dos))
                    (not (gnutls-available-p))))
       (proto (if no-ssl "http" "https")))
  ;; Comment/uncomment these two lines to enable/disable MELPA and MELPA Stable as desired
  (add-to-list 'package-archives (cons "melpa" (concat proto "://melpa.org/packages/")) t)
  ;;(add-to-list 'package-archives (cons "melpa-stable" (concat proto "://stable.melpa.org/packages/")) t)
  (when (< emacs-major-version 24)
    ;; For important compatibility libraries like cl-lib
    (add-to-list 'package-archives '("gnu" . (concat proto "://elpa.gnu.org/packages/")))))
(package-initialize)
```

milkypostman:
Since I don’t use Windows I can’t say the right solution. Please update the text which you have objection to and submit a PR and we can make the fix from that approach.

bradyt:
For non-Cygwin Emacs:

Before emacs 25, users were simply directed to dump binaries in `bin` from sourceforge, per the manual. Looking through the emacs-devel list, I'm realizing that at version 25, the download servers provided the dependencies as a separate download. As of version 26, there is a download option prebundled with the dependencies included.

From https://github.com/emacs-mirror/emacs/blob/master/nt/INSTALL#L753-L760:

[...]

PR #5685, opened 2018-08-11
Don't encourage http

bradyt:
Propose commit for issue #5682.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,15 +40,12 @@ Enable installation of packages from MELPA by adding an entry to
 
 ```lisp
 (require 'package)
-(let* ((no-ssl (and (memq system-type '(windows-nt ms-dos))
-                    (not (gnutls-available-p))))
-       (proto (if no-ssl "http" "https")))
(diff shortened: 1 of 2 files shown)


purcell:
Someone's going to have to deal with the support load when MELPA then doesn't work for users with an Emacs built without SSL support. This change isn't the way to go IMO, but if you'd like to propose an alternative change which notes that the snippet will fall back to HTTP - or which adds a `warn` call when doing so - then that might be a good compromise.

bradyt:
I think priority is stop making it this easy to use http. The second priority is adding a note for Windows users without Cygwin, with emacs < 26, and with emacs < 25 that didn't see the deps, or emacs < 24 and didn't see the documentation on how to enable gnutls. That should ideally be rare, I hope you are expecting users to research better emacs use. The user this commit seems to be based off is frankly wrong. You are doing harm to anyone that is not doing their research. Make it hard to do the wrong thing.

bradyt:
@purcell, I won't always be able to quickly respond, but if a support load arises, I'd be open to being pinged on this topic, and besides assisting users, discussing/ contributing to how the experience could be improved based on those issues.

PR #5686, opened 2018-08-12
Add org-timeline

Fuco1:

Add graphical view of agenda to agenda buffer.

https://github.com/Fuco1/org-timeline/

Author

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/org-timeline
+++ b/recipes/org-timeline
@@ -1,0 +1,1 @@
+(org-timeline :fetcher github :repo "Fuco1/org-timeline")


alphapapa:
This looks neat! Would you consider renaming it to org-agenda-timeline? That way it would be clear that it's an agenda extension rather than a general timeline view for a buffer.

Fuco1:
@purcell Yea, thanks for the feedback.  I always try to `:inherit` and even annoy people with pull requests doing just that.  I mostly just pulled this out of my config because someone asked me to so I guess I forgot.  Will be fixed!

PR #5689, opened 2018-08-13
add pact-mode recipe

emilypi:

A major mode for editing Pact, a smart contract language, in Emacs. See http://kadena.io/pact.

https://github.com/kadena-io/pact-mode

Maintainer, contributor, colleague of author

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/pact-mode
+++ b/recipes/pact-mode
@@ -1,0 +1,3 @@
+(pact-mode :fetcher github
+           :repo "kadena-io/pact-mode"
+           :files ("pact-mode.el"))


purcell:
Thanks, and sorry for the slow turnaround. Quick notes:

- This package will bundle `pact-flycheck.el`, but that should be a separate package because it has opinionated and heavier-weight dependencies. So this recipe should have `:files ("pact-mode.el")`, there should be a separate `pact-flycheck` recipe with `:files ("pact-flycheck.el")`, and `pact-flycheck.el` should declare package dependencies on `flycheck` and `pact-mode`. Also note that `flycheck-pact` would be a better name, since it better reflects current conventions.
- [This version check](https://github.com/kadena-io/pact-mode/blob/5b334255212abf29890f01055f5abd5c28348efc/pact-mode.el#L89-L90) should be at the top level, or (better) simply removed. You should instead declare a package dependency on Emacs 24.3 - `package-lint` should have mentioned this when you ran it.
- You can drop the `progn` [here](https://github.com/kadena-io/pact-mode/blob/5b334255212abf29890f01055f5abd5c28348efc/pact-mode.el#L188)
- [Version variables](https://github.com/kadena-io/pact-mode/blob/5b334255212abf29890f01055f5abd5c28348efc/pact-mode.el#L37) are generally a bad idea. This one is already out of step with the file header. Just use the file header (and note that MELPA will insert a package version for you there).

Hope that helps!

emilypi:
@purcell Thanks for the feedback Steve! No worries on the turnaround - I'll sort out your suggestions and come back today 😄

emilypi:
@purcell All of your comments have been addressed. I've submitted the `flycheck-pact` recipe in a [separate PR](https://github.com/melpa/melpa/pull/5711).

purcell:
Awesome, merged now. Welcome to MELPA. :tada:

emilypi:
Thanks @purcell! Very stoked :)

PR #5690, opened 2018-08-16
Add Reykjavik-ish theme

jssee:

A dark theme based on Reykjavik-theme 

https://github.com/jssee/reykjavik-ish

maintainer

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/reykjavik-ish
+++ b/recipes/reykjavik-ish
@@ -1,0 +1,1 @@
+(reykjavik-ish :fetcher github :repo "jssee/reykjavik")


purcell:
Thanks, and sorry for the slow response!

You've got a package here with one name, but its contents are a file with a different name. The two have to match, so you must change the recipe to `reykjavik-theme`.

Re. the theme file itself:
- May I suggest that `"Theme"` is not the best package description.
- The `deftheme` form is also missing the description parameter.
- The `Commentary` section is repeated and malformed, and its contents should be line-wrapped.
- The `Copyright` header line is missing a name, and the year is presumably wrong.

Hope that helps!

jssee:
Hi sorry for the late response but the author of the original theme made updates that make this version not necessary. closing this now.

PR #5691, opened 2018-08-16
Add lsp-ruby

gpittarelli:

Implements some Ruby IDE features via [lsp-mode](https://github.com/emacs-lsp/lsp-mode) and [this solargraph gem](https://github.com/castwide/solargraph)

https://github.com/emacs-lsp/lsp-ruby

maintainer

N/A

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/lsp-ruby
+++ b/recipes/lsp-ruby
@@ -1,0 +1,3 @@
+(lsp-ruby
+ :fetcher github
+ :repo "emacs-lsp/lsp-ruby")


alphapapa:
Hi,

I know little about Ruby, but a little about Emacs, so I have this suggestion:
```el
(defun lsp-ruby--render-string (str)
  "Render STR with `ruby-mode' syntax highlighting."
  (ignore-errors
    (with-temp-buffer
      (ruby-mode)
      (insert str)
      (font-lock-ensure)
      (buffer-string))))
```
Instead of using a new temporary buffer and activating `ruby-mode` each time this function is called, you could use a permanent, hidden buffer created when the function is first called, and only activate `ruby-mode` once.  This would likely perform much better, and it would avoid activating whatever hooks the user may have configured for `ruby-mode` every time the function is called.  For example, I do this in https://github.com/alphapapa/helm-org-rifle/blob/master/helm-org-rifle.el#L1530 to avoid activating `org-mode` every time a string is fontified.

yyoncho:
@alphapapa to avoid the hooks you may use  `(delay-mode-hooks (ruby-mode))` .

alphapapa:
@yynozk Thanks, that's good to know, but in this case, I think running `ruby-mode` for every string should still be avoided.  Even if user hooks are delayed, it's still not necessary to do everything the mode does.

yyoncho:
@alphapapa 
```emacs-lisp
(benchmark
 1000
 (with-temp-buffer
   (delay-mode-hooks (ruby-mode))))
```
=> Elapsed time: 0.000112s

purcell:
Good notes above, thanks. I'll merge this now, but yes, @gpittarelli, you should use `delay-mode-hooks` in this code, as above.

PR #5692, opened 2018-08-17
Add lab-themes.

MetroWind:

A custom theme for Emacs

https://github.com/MetroWind/lab-theme

Author

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/lab-themes
+++ b/recipes/lab-themes
@@ -1,4 +1,3 @@
 (lab-themes
  :fetcher github
- :repo "MetroWind/lab-theme"
- :branch "master")
+ :repo "MetroWind/lab-theme")
--- a/recipes/lab-themes
+++ b/recipes/lab-themes
@@ -1,0 +1,4 @@
+(lab-themes
+ :fetcher github
+ :repo "MetroWind/lab-theme"
+ :branch "master")


purcell:
Thanks! A few comments:

- Please remove the `:files` spec from the recipe, since it is redundant.
- Stick to lower case file and theme names, please.
- The package description, `"A carefully contructed custom theme"` is not terribly helpful.
- The theme files say they're based on flatui - is that accurate?
- Setting `ns-appearance` [like this](https://github.com/MetroWind/lab-theme/blob/master/Lab-Dark-theme.el#L301-L306) is an unreliable and undesirable hack, and should be removed.

Hope that helps. Otherwise everything looks good! :-)

DamienCassou:
What is the status of the PR?

purcell:
Great, looks good now, but needs the same `;;;###autoload` for the `custom-theme-load-path` as mentioned in the `flucui-themes` PR.

purcell:
:tada:

purcell:
P.S. You might be interested in a mode I just published which aims to keep the `ns-appearance` frame parameter in sync with the current theme: https://github.com/purcell/ns-auto-titlebar

MetroWind:
> P.S. You might be interested in a mode I just published which aims to keep the `ns-appearance` frame parameter in sync with the current theme: https://github.com/purcell/ns-auto-titlebar

Hah, that’s interesting~ Thanks for the info~

PR #5693, opened 2018-08-18
Add `proc` package.

jmjoy:

Macros for writing human readable elisp code. 

https://github.com/jmjoy/proc

maintainer

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/proc
+++ b/recipes/proc
@@ -1,0 +1,1 @@
+(proc :repo "jmjoy/proc" :fetcher github)


alphapapa:
There is no documentation nor examples of usage.  There is a single macro, `proc`, which has the docstring, "C-style variables definition."  There is essentially nothing here.

purcell:
Hi @jmjoy - thanks for taking the time to submit this to MELPA.

> There is no documentation nor examples of usage. There is a single macro, proc, which has the docstring, "C-style variables definition." There is essentially nothing here.

I agree with this feedback, sorry. The name is very confusing, since "proc" is a common term, and there's no indication of what this package/code should do. The package should be renamed and documented, so I'll go ahead and close this pull request, and you can open another when ready.

PR #5697, opened 2018-08-21
Add recipe for flymake-shellcheck

federicotdn:

This package adds a new backend for Flymake, which uses [ShellCheck](https://github.com/koalaman/shellcheck) to report problems found on bash/sh scripts.

https://github.com/federicotdn/flymake-shellcheck

Author

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/flymake-shellcheck
+++ b/recipes/flymake-shellcheck
@@ -1,0 +1,1 @@
+(flymake-shellcheck :repo "federicotdn/flymake-shellcheck" :fetcher github)


purcell:
Nice!

Does Emacs 25 really have a sufficiently up-to-date `flymake` for this to work? Please double-check that, and adjust the `Package-Requires` dependency if necessary.

federicotdn:
@purcell You're right, my package only works with the new Flymake included in Emacs 26+. I'll update the dependency as you mentioned.

PR #5702, opened 2018-08-23
Need to update my project's files 

ahungry:
In 4.x branch of my org-jira project I introduced a new file to include as part of repo - need the recipe updated so the build doesn't fail for users.


tarsius:
I would have to checkout this branch locally to be sure what is going on and I am not sure what will happen if I do a "rebase and merge" using Github. Please remove the commits that are not about `org-jira` from the pull-request.

ahungry:
Thanks, I'll do that this evening and make sure it's a single commit coming in.  if you just decline this and update yourself, I'm fine with that as well. I just need to include that new file in the fetcher for the users

PR #5704, opened 2018-08-25
Add terminal-focus-reporting.el

veelenga:

Make Emacs play nicely with iTerm2 and tmux.

This plugin restores `focus-in-hook`, `focus-out-hook` functionality. Now Emacs can save when the terminal loses a focus, even if it's inside the tmux.

https://github.com/veelenga/terminal-focus-reporting.el

Maintainer

**None needed** 

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/terminal-focus-reporting
+++ b/recipes/terminal-focus-reporting
@@ -1,0 +1,1 @@
+(terminal-focus-reporting :repo "veelenga/terminal-focus-reporting.el" :fetcher github)


alphapapa:
Hi,

I think your package should use a minor mode to set its hooks and key bindings.  Also, please enable lexical-binding, and run package-lint, etc.

veelenga:
@alphapapa thanks for the feedback. Will do.

Is there a way to run some action if the minor mode disabled? I would need to stop focus reporting if the mode is disabled.

alphapapa:
* Lexical binding must be enabled with the proper file header.  Please see https://github.com/alphapapa/emacs-package-dev-handbook#lexical-binding-1
* I don't know how terminal focus handling works, but I'm guessing that these hooks should be enabled/disabled by the minor mode--that's the point of having a minor mode:
```el
(add-hook 'kill-emacs-hook 'terminal-focus-reporting--deactivate)
(add-hook 'terminal-focus-reporting-mode-hook 'terminal-focus-reporting--toggle)
```
* Nitpick: There are indentation mistakes, so you might want to reindent the buffer.  (`aggressive-indent-mode` is helpful, but a matter of preference.)

veelenga:
Okay, makes sense. Corrected.

alphapapa:
Thanks.  Again, I don't know exactly how terminal focus handling works, but in general, hooks that are added when a minor mode is activated should be removed when the minor mode is disabled, but I see in https://github.com/veelenga/terminal-focus-reporting.el/commit/41d4d0d655e9bb03b1dbce6a5d205f9c21e97a21 that you only add the hook, and nothing is done when the minor mode is disabled.  Is this appropriate?

veelenga:
Hm, actually i was thinking that hooks added by the minor mode are automatically removed if the minor mode is disabled. If this is not a case, how can I remove such a hook?

veelenga:
@alphapapa you are correct, I had a misunderstanding how the minor mode works and was thinking it automatically removes the added hooks once disabled.

Corrected in https://github.com/veelenga/terminal-focus-reporting.el/commit/63eb210da8d2c7cae6a2a5fae86a13956a49e34f

alphapapa:
Thanks, but I'm not sure doing it that way will work.  It's not the convention to use `:after-hook` like that.  Instead, the minor mode variable is tested in the body of `define-minor-mode`, like this:

```el
(define-minor-mode magit-todos-mode
  "Show list of to-do items in Magit status buffer for tracked files in repo."
  :require 'magit-todos
  :group 'magit-todos
  :global t
  (if magit-todos-mode
      (add-hook 'magit-status-mode-hook #'magit-todos--add-to-status-buffer-kill-hook 'append)
    ;; Disable mode
    (remove-hook 'magit-status-mode-hook #'magit-todos--add-to-status-buffer-kill-hook)))
```

veelenga:
Actually, it works for me (both `add-hook` and `remove-hook`). But if you are saying this is not a convention, i am ok to change it: https://github.com/veelenga/terminal-focus-reporting.el/commit/75286c4aac23069f23b73e988a708305d243df4f

alphapapa:
Thanks.

purcell:
Looks great now, thanks for the productive discussions above! Merging...

PR #5705, opened 2018-08-26
Update recipe for org-jira to include additional needed file

ahungry:
This is a revision that includes just the single missing commit/file.


tarsius:
Thanks!

ahungry:
np, thank you!

PR #5708, opened 2018-08-28
Add slow-keys

manuel-uberti:

A minor mode to enforce slow typing and prevent RSI.

Original author is @chrisdone, I hope he'll be happy with this. :)

https://github.com/manuel-uberti/slow-keys

I am the maintainer.

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/slow-keys
+++ b/recipes/slow-keys
@@ -1,0 +1,1 @@
+(slow-keys :fetcher github :repo "manuel-uberti/slow-keys")


purcell:
Merged, thanks. Don't use `bound-and-true-p` [here](https://github.com/manuel-uberti/slow-keys/blob/master/slow-keys.el#L97) - you can just test the variable directly, because it will always be bound: it's defined in the same file as where you're testing it!

PR #5709, opened 2018-08-29
Add highlight-doxygen

Lindydancer:

Highlight Doxygen comments.

In addition to highlighting Doxygen commands and their arguments, the entire comment is highlighted as a block. In addition, code blocks inside Doxygen comments are highlighted as code, according to the language they written in.

https://github.com/Lindydancer/highlight-doxygen

I'm the author.

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/highlight-doxygen
+++ b/recipes/highlight-doxygen
@@ -1,0 +1,1 @@
+(highlight-doxygen :repo "Lindydancer/highlight-doxygen" :fetcher github)


Lindydancer:
Hi!

Steve wrote:

> Thanks! I feel like all the packages you submit are very well organised
> and straightforward to review and merge, which makes me happy. :-)
>

Thanks!

I've spent may years writing Emacs packages, and I always try to make them
both efficient and maintainable -- by write clean code and by providing
suitable test cases. I'm really glad that you appreciate the work I've put
in.


This is no exception, and my only note is that the combination of :inherit
> font-lock-doc-face and :background "grey95" etc. is likely to lead to
> ugly results by default with many themes. I wonder if there's a more robust
> alternative. Anyhow, I'll go ahead and merge now.
>

The idea is to make a block with a slightly different background than the
default, which typically is black-ish or white-ish. By inheriting from the
"doc" face and setting the background color, the effect is that the
foreground color (and other attributes) remain as the theme specified, on
top of a background that is of a slightly different color than the normal
background (grey85 and grey15, respectively).

I've tested this on a handful of themes, and it looked good in all of them.
Of course, should there be a theme where it doesn't look good, the faces my
package provide can be customized to match the theme.

    -- Anders

purcell:
> I've tested this on a handful of themes, and it looked good in all of them. Of course, should there be a theme where it doesn't look good, the faces my package provide can be customized to match the theme.

Yup, agree, I don't think there's a perfect solution. :-)

PR #5710, opened 2018-08-29
Add vdm-mode

peterwvj:

Major mode for the Vienna Development Method 

https://github.com/peterwvj/vdm-mode

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/flycheck-vdm
+++ b/recipes/flycheck-vdm
@@ -1,5 +1,4 @@
 (flycheck-vdm
  :fetcher github
  :repo "peterwvj/vdm-mode"
- :files ("flycheck-vdm.el"
-         "vdm-mode-util.el"))
+ :files ("flycheck-vdm.el"))
--- a/recipes/flycheck-vdm
+++ b/recipes/flycheck-vdm
@@ -1,0 +1,5 @@
+(flycheck-vdm
+ :fetcher github
+ :repo "peterwvj/vdm-mode"
+ :files ("flycheck-vdm.el"
+         "vdm-mode-util.el"))
--- a/recipes/vdm-mode
+++ b/recipes/vdm-mode
@@ -1,5 +1,5 @@
 (vdm-mode
  :fetcher github
  :repo "peterwvj/vdm-mode"
- :files ("*.el"
-         "snippets"))
+ :files ("vdm-mode.el"
+         "vdm-mode-util.el"))
--- a/recipes/vdm-mode
+++ b/recipes/vdm-mode
@@ -1,0 +1,5 @@
+(vdm-mode
+ :fetcher github
+ :repo "peterwvj/vdm-mode"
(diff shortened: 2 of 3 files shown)


purcell:
Thanks! Quick feedback:

- The code all looks nice and tidy! :-)
- Please use `:defaults` in place of `"*.el"` in the recipe's `:files` spec.
- In general, `flycheck` is a very opinionated/divisive package upon which major modes should not depend, so you should package `vdm-mode-flycheck` separately (as "flycheck-vdm") and make it depend on `vdm-mode` and `flycheck`, so that `vdm-mode` itself has lighter dependencies.
- Similarly for `yasnippet`, `vdm-snippets` would be best as a separate package instead of tangling everything up together. Major modes are meant to be primarily about indentation, navigation and font-lock.
- The empty list return values [here](https://github.com/peterwvj/vdm-mode/blob/ba609c501443aefccfc299594b902055943760c1/vdm-mode-util.el#L86-L87) are the default, so you might consider using `when` instead of `if`.

Hope that helps.

peterwvj:
Yes, this helps a lot - thanks a lot for your feedback! :)

Based on your suggestions,

- the `yasnippet` and `flycheck` features are now being packaged separately (see https://github.com/melpa/melpa/pull/5710/commits/430584bb712b85058745d5cf62c707462123745d and https://github.com/melpa/melpa/pull/5710/commits/35d87823738a7a96ab20933bd35717e9fba8b72d) as `flycheck-vdm` and `vdm-snippets` (see https://github.com/peterwvj/vdm-mode/commit/f9ffbf0a40dac260680d78dda5b10956931dd083).

- the `vdm-mode` recipe has been updated accordingly (see https://github.com/melpa/melpa/pull/5710/commits/e2ef40d1331b7cbefbe4a35ca52b0c08e01a47e1).

- `vdm-mode` does not depend on `flycheck-vdm` and `vdm-snippets` or require these packages be installed.

- `vdm-mode-util-find-vdm-files ` has been changed to use `when` instead of `if` as you suggested (see https://github.com/peterwvj/vdm-mode/commit/a5def26447b9ea1e27b3348cb3ac6578bae064ee).

purcell:
Awesome. One last thing - rather than bundle `"vdm-mode-util.el"`, `flycheck-vdm` should just depend on `vdm-mode`.

peterwvj:
Again, thanks for your feedback.

- flycheck-vdm now depends on vdm-mode (see https://github.com/peterwvj/vdm-mode/commit/0c083ee4848ea5d78de7894a4a0722d6630839c9)

- the flycheck-vdm recipe has been updated accordingly (see https://github.com/melpa/melpa/pull/5710/commits/30060274908c0031b242d32ade13fc122ad189dc)

purcell:
Fantastic - merging now!

PR #5712, opened 2018-08-31
Add flymake-racket

jojojames:

Flymake extension for racket.

https://github.com/jojojames/flymake-racket

Maintainer

None needed

### Checklist

Please confirm with `x`:


purcell:
Looks good and I've merged it (thanks for your patience!) but I'm thinking maybe make `async` a dependency if the async behaviour is preferable, then you don't have to do any conditional magic.

PR #5713, opened 2018-08-31
Add flymake-gradle

jojojames:

Flymake extension for gradle build tool.

https://github.com/jojojames/flymake-gradle

Maintainer

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/flymake-gradle
+++ b/recipes/flymake-gradle
@@ -1,0 +1,1 @@
+(flymake-gradle :fetcher github :repo "jojojames/flymake-gradle")


purcell:
Looks good, and I've merged it, but you might consider fixing up those `cond` switches based on `major-mode`. For a start, `cl-case` or (probably better) `pcase` would tidy those up, but more generally it's worth considering `derived-mode-p` instead of testing for exact equality of `major-mode` with a specific symbol.

PR #5716, opened 2018-09-01
Update recipe of ede-php-autoload

stevenremot:
ede-php-autoload has changed owner. This PR updates the recipe accordingly.


purcell:
Thanks, and sorry for the delay.

PR #5719, opened 2018-09-03
Add org-msg recipe

jeremy-compostella:
Signed-off-by: Jeremy Compostella <jeremy.compostella@gmail.com>

OrgMsg is a [GNU/Emacs] module making use of [Org mode] to write and
reply to emails in a Outlook HTML friendly style.

Not that I really like the Outlook email style but in a work
environment dominated by this software, it is good to be able to reply
with the same style.  By default, if the original message is in text
form 'org-msg' keeps it that way and does not activate itself.  It
allows to reply to developer mailing list (or any real technical
people who knows how to handle text email) seamlessly.  If the
original message is in the HTML form, it activates the 'org-msg' mode
on the reply buffer.

*org-msg* is an derivation of [org-mode] in which some functionality
of [message-mode] are imported or replicated. For instance, A
*org-msg* buffer uses the same 'font-lock-keywords' than
[message-mode] or the 'TAB' key while the cursor is in the header
calls the 'bbdb-complete-mail()' function.

For convenience, the original message is quoted below the '--citation
follows this line (read-only)--' marker.  So you can easily refer to
the original message.  However, the entire quoted text is read-only
because *org-msg* does not support modification of the original
content.

Outlook email are really poor compared to what can be achieved by
[org-mode].  I personally use the '#+{begin|end}_quote' to quote part
of the email I am replying to.  I extensively make use of
'#+{begin|end}_src mode' to provide extract of code or example of
changes, I sometimes use [org-babel] to generate an on the fly
sequence diagram (using [plantuml]) or a simple graph (using [dot]).

https://github.com/jeremy-compostella/org-msg

Maintainer

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/org-msg
+++ b/recipes/org-msg
@@ -1,0 +1,2 @@
+(org-msg :repo "jeremy-compostella/org-msg" :fetcher github)
+


alphapapa:
Hi,

A few issues I noticed:

1.  These mode hooks should be enabled/disabled by your minor mode.  Simply loading your package shouldn't modify any hooks.
```el
(add-hook 'gnus-message-setup-hook 'org-msg-post-setup)
(add-hook 'message-send-hook 'org-msg-prepare-to-send)
(add-hook 'org-ctrl-c-ctrl-c-final-hook 'org-msg-ctrl-c-ctrl-c)

(add-to-list 'message-syntax-checks '(invisible-text . disabled))
```
2.  You should enable and test with lexical binding.  See https://github.com/alphapapa/emacs-package-dev-handbook#lexical-binding-1

jeremy-compostella:
@alphapapa I took your two comments into account which leaded to some interesting clean-up changes in the org-msg module.  Let me know if there is anything missing.

alphapapa:
From a quick look, the changes look good.  

One more suggestion: rather than using `local-set-key` in `org-msg-edit-mode`, you should probably define an `org-msg-edit-mode-map` keymap.

jeremy-compostella:
Done.  I also remove the unnecessary :group on the defcustom

jeremy-compostella:
@alphapapa , I think I covered all your comments.  Please let me know if I missed anything.

purcell:
Thanks, and sorry for the long delay - this looks good. Merging now, but I'd like to suggest you stop using the deprecated `cl` and instead use only the new `cl-lib`, which you're already using without `require`-ing, e.g. for `cl-macrolet`. Welcome to MELPA!

jeremy-compostella:
@purcell thanks for merging and and the review.  I took your comment into account (https://github.com/jeremy-compostella/org-msg/commit/b835102f09bade51350f36d1c0f0371ba70ca7b4)

PR #5720, opened 2018-09-03
Add pdfgrep recipe

jeremy-compostella:
Signed-off-by: Jeremy Compostella <jeremy.compostella@gmail.com>

*pdfgrep* is a [GNU/Emacs] module providing `grep' comparable facilities
but for PDF files.  Its usage is similar to the `grep' function: using
the `next-error' function gets you to the next matching page...

The `pdfgrep' program must be installed on your system.

If the [pdf-tools] module is installed and enabled, when a matching page
is displayed, the matching regions are highlighted.

https://github.com/jeremy-compostella/pdfgrep

Maintainer

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/pdfgrep
+++ b/recipes/pdfgrep
@@ -1,0 +1,1 @@
+(pdfgrep :repo "jeremy-compostella/pdfgrep" :fetcher github)


alphapapa:
Hi,

A few issues:

1.  Please enable and test with lexical-binding.
2.  You don't need to specify the `:group` for `defcustom`s that come after the `defgroup`.
3.  You don't need to use `purecopy`.
4.  Loading your package should not add advice like this: `(advice-add 'compilation-goto-locus :after #'pdfgrep-goto-locus)`.  That should be activated by a minor mode, or manually by the user.
5.  There are some minor indentation issues.  `aggressive-indent-mode` helps with this.
6.  For ignored arguments, please prefix them with `_` to help the byte-compiler.
7.  You might want to use `executable-find` to ensure `pdfgrep` is present before trying to run it.
8.  Generally you should escape shell commands before running them.
9.  Your readme would greatly benefit from installation instructions and a screenshot.  :)

alphapapa:
> Regarding "You don't need to specify the :group for defcustoms that come after the defgroup." Could you point me out to documentation, I can't find anything about that.

See the elisp manual for `defcustom`.

> Regarding the advice part, I need to figure out a way. How can I disable an interactive function. Let's assume I turn pdfgrep-mode into a global minor mode. When that mode is disabled, I want the pdfgrep function to not be callable. Is there some best practice for such a use case ?

You can override the function like `(advice-add #'pdfgrep :override #'your-function-here)`.

Silex:
> You don't need to specify the :group for defcustoms that come after the defgroup.

Is this really bad practice tho? All my packages use `:group`, but if it's accepted practice to not use them I might switch.

jeremy-compostella:
Thank you for pointing me out to the elisp manual.  Naively, I was only looking at the Info pages and describe-function :/

I am focusing on melpa/melpa#5719 first and I take into account some of your comment from this pull request too.  I am going to work on the pdfgrep package soon but I usually only have time to work on emacs packages on the week-ends.

purcell:
Looks great, thanks.

alphapapa:
@purcell AFAICT he hasn't addressed the issues I mentioned yet.  Most of them aren't a big deal, but the one about advice and escaping shell commands seem important.

jeremy-compostella:
@alphapapa I have just taken into account most of your comments.

The escape of the command is a bit enigmatic.  I based my work on the Emacs grep module, I don't find any reference to that part.

I appreciate your reviews.

alphapapa:
> The escape of the command is a bit enigmatic. I based my work on the Emacs grep module, I don't find any reference to that part.

In general it's a good idea to `shell-quote-argument` when building shell commands, especially with user input.  Even if it's not so much of a security issue in this case, since the user is typing them in, it can help prevent accidents.

jeremy-compostella:
OK I get your point.  However, I don't think it applies to the pdfgrep module case.  The user isn't supplying any parameter.  He writes his own command like with M-x grep.

PR #5721, opened 2018-09-04
Add counsel-repo

keegancsmith:

Jump to repository using [ivy](https://github.com/abo-abo/swiper).

https://github.com/keegancsmith/counsel-repo

Maintainer

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/counsel-repo
+++ b/recipes/counsel-repo
@@ -1,0 +1,3 @@
+(counsel-repo
+  :repo "keegancsmith/counsel-repo"
+  :fetcher github)


alphapapa:
This is a little unusual.  

1.  The package requires building and installing a Go program, without which the Emacs package won't work
2.  It seems like this is already covered by existing tools, especially https://github.com/ericdanan/counsel-projectile.  What exactly does this package provide over and above that (which doesn't require installing an external utility)?

keegancsmith:
> The package requires building and installing a Go program, without which the Emacs package won't work. That program is simply a faster implementation of find, and there are already faster find-like utilities, like fd. It would be better if your package used existing utilities like that and pointed users to the best one.

The README talks about this, but not in depth https://github.com/keegancsmith/counsel-repo#how-it-works
Both find and fd can't find all project roots quickly, because you can't stop them from recursing into project directories (you are looking for a child `.git`, but you want to stop recursing at the parent). I give an example find command which shells out to test to avoid this, which is faster but still very slow. See the timings difference. Its the difference between adding caching vs just always searching. Additionally I implemented a way to do it with find, but it honestly was so slow I thought it would be a poor user experience to have it at all.

> It seems like this is already covered by existing tools, especially https://github.com/bbatsov/projectile and https://github.com/ericdanan/counsel-projectile. What exactly does this package provide over and above those (which don't require installing an external utility)?

This specifically targets jumping to any repo directory, not just "open" or "saved" project directories. If I am not mistaken projectile's jump to project buffer doesn't work that way. I could be mistaken though.


I can understand the hesitation of adding a tool which has a dependency on an external tool. There are many tools like this in the registry, but maybe I should give this a different name to more clearly communicate it is very closely tied to an external program?

alphapapa:
> Both find and fd can't find all project roots quickly, because you can't stop them from recursing into project directories (you are looking for a child .git, but you want to stop recursing at the parent). 

I see.  That's a clever optimization.

> Additionally I implemented a way to do it with find, but it honestly was so slow I thought it would be a poor user experience to have it at all. However, I can reconsider that decision.

I guess that's up to your judgement as the package author.  :)

For some platforms, a potential alternative for finding repos, although potentially not always up-to-date, would be to use something like `locate -r '/\.git$'`.  That would work without installing extra utilities.  And in some cases, it would be faster than actually searching the filesystem.  So you might consider implementing that as an alternative backend that would work without installing the Go utility.

> This specifically targets jumping to any repo directory, not just "open" or "saved" project directories. 

I see.  I can see how that would be useful for some users.

> I can understand the hesitation of adding a tool which has a dependency on an external tool.  There are many tools like this in the registry, but maybe I should give this a different name to more clearly communicate it is very closely tied to an external program?

I think that might be a good idea, yes.  Note, however, that I am not MELPA staff, so these are only my opinions.

Have you thought about a way to generalize your solution?  i.e. a find-like utility with conditional recursion could probably be useful for other tools.  Maybe you could even patch an existing project like `fd`.

terlar:
Speaking of external tools, my first reaction when I saw this and that I really liked was that this was like an Emacs version of navigating projects like https://github.com/motemen/ghq. One annoyance with projectile is that it only navigates projects that you previously have opened files inside. I often find myself in the situation where I would like to do a project search and haven't visited the project before.

Really looking forward to this tool, no matter the decision for back-end.

keegancsmith:
@terlar Thanks for the link to ghq, I wasn't aware of that project and `ghq list` is perfect for the counsel-repo use-case. `counsel-repo` does order the repositories by `HEAD`'s timestamp, which ghq does not, which is very useful for making the candidates ivy filters by more relevant. But from what I can tell that would likely be a welcome feature upstream and relatively easy to do from the quick read of the code I did.

At this stage, I believe ghq is the more general tool and when I have played with it more I'll likely switch to it. When that time comes I'll likely publish some elisp which integrates with it. As such I'll close this out.

@terlar for now you can follow the instruction's in counsel-repo's README to get the functionality you desire.

alphapapa:
> At this stage, I believe ghq is the more general tool and when I have played with it more I'll likely switch to it. When that time comes I'll likely publish some elisp which integrates with it. As such I'll close this out.

That sounds like it will be a very useful package!

PR #5724, opened 2018-09-06
Update racket-mode :files to support racket/ subdir

greghendershott:
- Handle *.rkt files in a `racket/` subdir in both source and target.

  Note: To smooth the transition, keep the top-level `"*.rkt"` for now,
        so that the recipe works for both the status quo flat file
        layout and the new nested layout.

- Exclude some test and example files.

---

Major mode for Racket lang.

https://github.com/greghendershott/racket-mode

Maintainer

n/a

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/racket-mode
+++ b/recipes/racket-mode
@@ -1,2 +1,10 @@
-(racket-mode :repo "greghendershott/racket-mode" :fetcher github
-             :files ("*.el" "*.rkt"))
+(racket-mode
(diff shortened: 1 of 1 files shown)


purcell:
I sanitised this to avoid some unnecessary bits. Merging now. (I suspect `("racket" "racket/*")` could simply be `"racket"` but I haven't tested it locally to see if the subsequent `:exclude` works that way.)

PR #4307, opened 2016-10-10
Submitting Secretaria.el

shackra:
### Brief summary of what the package does

My attempt of an assistant for you in org-mode. This library contains
utilities that enhance your experience with org-mode.

#### Features
  - Reminders
      - All tasks scheduled or that have a deadline set to "today", but
        have no time of the day specified.
      - The current clocked task, every N minutes (10 by default).
      - In case of Emacs crashing, the task clocked in at the moment so
        you don't forget about fixing that.

### Direct link to the package repository

https://bitbucket.org/shackra/secretaria.el

### Your association with the package

I'm the creator and current maintainer.

### Relevant communications with the upstream package maintainer

None.

### Checklist

- [x] I've read [CONTRIBUTING.md](https://github.com/melpa/melpa/blob/master/CONTRIBUTING.md)
- [x] I've built and installed the package using the instructions in the [README](https://github.com/melpa/melpa/blob/master/README.md)

PR #5730, opened 2018-09-11
Updating Secretaria.el

shackra:

Enable better behavior for task reminders in org-mode.

https://gitlab.com/shackra/secretaria

### Checklist

Please confirm with `x`:

- [ ] I've built and installed the package using the instructions in [CONTRIBUTING.org](https://github.com/melpa/melpa/blob/master/CONTRIBUTING.org) -- I did `make sandbox INSTALL=secretaria` but this installed the last version of the package and not the latest version with the changes and additions

https://github.com/melpa/melpa/pull/4307

### DIFF
--- a/recipes/secretaria
+++ b/recipes/secretaria
@@ -1,1 +1,1 @@
-(secretaria :files ("secretaria-*" "secretaria*") :fetcher gitlab :repo "shackra/secretaria")
+(secretaria :files ("secretaria*") :fetcher gitlab :repo "shackra/secretaria")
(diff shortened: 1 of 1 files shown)


shackra:
Thanks, I'll fix this mess as soon as possible!

shackra:
@purcell I did as you instructed!

purcell:
Super, thanks.

purcell:
Not now, it doesn't. I think you just didn't wait long enough.

shackra:
Indeed, it fixed itself!

purcell:
The build process runs in a several-hour cycle, so it takes a while for changes to take effect.

PR #5731, opened 2018-09-14
Added recipe for lsp-sh

innerout:
This package adds a new client to lsp-mode for
the Sh-Language-Server.

https://github.com/emacs-lsp/lsp-sh

I am a user of lsp-sh.

I have talked with the package maintainer and he agreed [here](https://github.com/emacs-lsp/lsp-sh/issues/1).

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/lsp-sh
+++ b/recipes/lsp-sh
@@ -1,0 +1,1 @@
+(lsp-sh :repo "emacs-lsp/lsp-sh" :fetcher github)


purcell:
Thanks, and sorry for the slow turnaround. Couple of quick notes:
- The `;;; Commentary:` section header belongs below the "Author" header etc., and before the `;;; Code:` section header.
- Since this is for bash and uses `bash-language-server`, shouldn't this all be called `lsp-bash`?

/cc @wilkystyle

innerout:
About the first question it was my mistake as the checkers did not complain so i let the commentary section there.
About the second question lets wait for @wilkystyle .

wilkystyle:
I stuck with `lsp-sh` because the corresponding mode in Emacs is called `sh-mode`. If `bash-language-server` is bothersome in light of that, I'm okay changing it to `sh-language-server`.

purcell:
> I stuck with `lsp-sh` because the corresponding mode in Emacs is called `sh-mode`. If `bash-language-server` is bothersome in light of that, I'm okay changing it to `sh-language-server`.

Yeah, that would probably be the way to go. In any case, I'll go ahead and merge this. Thanks folks.

PR #5732, opened 2018-09-15
Add smart-mode-line-atom-one-dark-theme recipe

daviderestivo:
This package adds a smart-mode-line theme that matches the
atom-one-dark one.

https://github.com/daviderestivo/smart-mode-line-atom-one-dark-theme

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/smart-mode-line-atom-one-dark-theme
+++ b/recipes/smart-mode-line-atom-one-dark-theme
@@ -1,0 +1,1 @@
+(smart-mode-line-atom-one-dark-theme :repo "daviderestivo/smart-mode-line-atom-one-dark-theme" :fetcher github)


recipes/smart-mode-line-atom-one-dark-theme
```diff
@@ -0,0 +1 @@
+(smart-mode-line-atom-one-dark-theme :repo "daviderestivo/smart-mode-line-atom-one-dark-theme" :fetcher github)
```
DamienCassou: please add a new-line character at the end of the first line.

daviderestivo:
> That's a bit of a long package name. Can you come up with something shorter?

What about sml-atom-one-dark-theme ?

recipes/smart-mode-line-atom-one-dark-theme
daviderestivo: Thanks ... I will do it.

daviderestivo:
Hi @tarsius,

In Melpa there is already another theme for sml (smart-mode-line-powerline-theme). Now for consistency reason I guess that my proposed name (sml-atom-one-dark-theme) is not really nice. So either we keep the current name (smart-mode-line-atom-one-dark-theme) or I propose smart-mode-line-aod-theme but I don't really like it.

Regards,
Davide

tarsius:
That's just one other theme, so maybe that could be renamed too?

Before you do anything, lets see what @Malabarba and @purcell have to say.

daviderestivo:
> That's just one other theme, so maybe that could be renamed too?
> Before you do anything, lets see what @Malabarba and @purcell have to say.

The problem is the way sml works. It expect that the package name start with smart-mode-line then the rest of the name is used on the config ... for example:

``` elisp
;; Tell Emacs where is your personal theme directory
(add-to-list 'custom-theme-load-path (expand-file-name "themes"
                                                       user-emacs-directory))
(use-package smart-mode-line
  :config
  (setq sml/theme 'atom-one-dark)
  (sml/setup))
```

Malabarba:
I wouldn't mind doing something to allow for different theme names in sml. But there's one more problem to consider, renaming packages is not something that package.el handles nicely.

daviderestivo:
So, shall I rename the package to smart-mode-line-aod-theme or leave as it is?

purcell:
> The problem is the way sml works. It expect that the package name start with smart-mode-line then the rest of the name is used on the config

Given this, let's merge as-is. Thanks.

daviderestivo:
Thanks you.

PR #5735, opened 2018-09-18
Add metamorph

AdamNiederer:

`mapcar` over a buffer using a similar interface to `replace-regex`. Possibly more higher-order text manipulation in the future (monomorphization, etc).

https://github.com/AdamNiederer/metamorph

I wrote the package

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/metamorph
+++ b/recipes/metamorph
@@ -1,0 +1,1 @@
+ (metamorph :fetcher github :repo "AdamNiederer/metamorph") 


alphapapa:
Hi,

This looks very interesting.  A few points:

* It's good to use `lexical-binding` whenever possible.
* Instead of calculating `search-extent` every time through the loop, you could use a marker.
* `(setq match-index (1+ match-index))` -> `(cl-incf match-index)`.
* I'd suggest making `metamorph-map-region-safe` the default (i.e. remove the `-safe` suffix), and calling the unsafe version `metamorph-map-region-unsafe`.
* It might not be worth "polluting" the obarray with `metamorph--stringify`, since it's just one `if`.  I already have 14 matches for functions containing `stringify` in their names in my Emacs config.
* "replace-string, but crazier" isn't a very helpful description, even if it is accurate.  ;)

purcell:
Agree with @alphapapa, particularly that this is going to be error-prone unless you use `markers` and also take care to avoid excursions and changes to match-data, which could easily result from any expressions you evaluate.

AdamNiederer:
Hey, thanks for all of the feedback!

> - Instead of calculating `search-extent` every time through the loop, you could use a marker.
> - `(setq match-index (1+ match-index))` -> `(cl-incf match-index)`.
> - I'd suggest making `metamorph-map-region-safe` the default (i.e. remove the `-safe` suffix), and calling the unsafe version `metamorph-map-region-unsafe`.
> - "replace-string, but crazier" isn't a very helpful description, even if it is accurate.  ;)
> - [snip] this is going to be error-prone unless you use markers and also take care to avoid excursions and changes to match-data [snip]

Incorporated in https://github.com/AdamNiederer/metamorph/commit/d9dc7037b7eed7b3fe85ea50e91f332e3f831514

> - It's good to use `lexical-binding` whenever possible.

Agreed, but I believe the user-provided transformation uses dynamic binding to access the state of the search (`% %! %i %0`), unfortunately. I tried `(eval transform nil)` without success.

> - It might not be worth "polluting" the obarray with `metamorph--stringify`, since it's just one `if`.  I already have 14 matches for functions containing `stringify` in their names in my Emacs config.

I'd like to keep this in case I decide to do any post-processing for types which aren't handled gracefully by `prin1-to-string` (hash tables, etc). It's also most likely going to be called at least once per user-facing function in this library, as this library is all about pretty-printing the results of a lisp expression into your buffer. I can try to rename it, but I don't think anybody would be tempted to call a private/`--` function.

purcell:
Looks ready to go, thanks.

PR #5737, opened 2018-09-20
Add new recipe for parrot

dp12:

parrot is used to rotate words according to a user-defined dictionary (e.g. "rock" -> "paper" -> "scissors"). As an additional feature, it displays the animated Slack party parrot in the mode line.

https://github.com/dp12/parrot

Maintainer

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/parrot
+++ b/recipes/parrot
@@ -1,0 +1,4 @@
+(parrot
+ :fetcher github
+ :repo "dp12/parrot"
+ :files (:defaults "img"))


purcell:
Thanks! Who doesn't love a good party parrot?

Some quick notes:
- `party-parrot.el` should instead be `parrot-party.el`, since all the `.el` files in your package (and the symbols defined therein) should begin with the package name.
- The recipe can (and should) then simply use `:files (:defaults "img")`.
- Don't use the deprecated `cl` - use `cl-lib` instead.
- Toggling the mode [here](https://github.com/dp12/parrot/blob/db123bff48089fc98ec475e800987b5340917371/party-parrot.el#L53-L54) looks bogus. Do you just need `(force-mode-line-update)` instead?
- The start and end bound regexes have docstrings which talk about matching beginnings and ends of words, so you probably want `\\_<` etc. instead of `[[:space:]]`.

Hope that helps!

dp12:
Thanks for your comments! I've applied the changes you suggested to the parrot repo and the melpa branch with the following notes:
- I flipped the dependency order of `party-parrot.el` and `parrot.el` to make the naming make sense. `parrot.el` is now `parrot-rotate.el`, and `party-parrot.el` is renamed to `parrot.el`
- I updated the docstrings for the start and end bound regexes to better say what they do, because for now, I really do want the user to be able to search for the first invalid character to define a "word". Maybe if I experiment more in the future, I'll change my mind!

dp12:
Sorry, accidentally hit the close button

DamienCassou:
Some initial feedback (I will try to read more of your code later):

1. the `defcustom`s following a `defgroup` don't have to specify the `:group` because the group is inferred
2. Please run linters specified in [CONTRIBUTING.org](https://github.com/melpa/melpa/blob/master/CONTRIBUTING.org). They would have told you about:
  1. there is a circular dependency: parrot-rotate.el depends on `parrot-start-animation` which is defined in parrot.el which depends on `parrot-rotate`. You probably want to use `declare-function`.
  2. `parrot-rotate-word-at-point` binds `pulse-flag` that is not declared or used anywhere
  3. by default, Emacs docstring should have 2 spaces after a period. You can either follow this convention or change the value of `sentence-end-double-space`
3. Usually, docstrings describing booleans are phrased like this: `"Non-nil if\u2026"`. This avoids adding `"This can be t or nil."` at the end of each of your boolean `defcustom`s.
5. The docstring of `parrot-rotate-convert-rotations-to-regexp` could better describe the relationship between the `rotations` parameter and the result. Is the regexp supposed to match any of the element or all of them? You might also want to add that `rotations` contain strings.
6. The Emacs-Lisp style guide recommends avoiding closing parenthesis at the beginning of a line. You have (at least) one in parrot-rotate.el:130 and another one in parrot-rotate.el:145.
7. Regarding the code of `parrot-rotate-convert-rotations-to-regexp`, you might want to use `cl-loop`, `dolist` or `mapcan` instead of the repeated uses of `append` and `mapcar`. Also, I would recommend using `append` inside the lambda instead of a let-bound variable. All-in-all, I would make it look like this (untested):

    ```emacs
    (regexp-opt
      (mapcan
        (lambda (entry)
          (append
            …
            (when (…) (mapcan #'upcase (plist-get entry :rot)))))
        rotations))
    ```

DamienCassou:
Here is the last part of my review for parrot-rotate.el. I will have a look at parrot.el then.

1. The docstring of `parrot-rotate-get-rots-for` references `ROTATIONS` which is not a parameter. Reading the docstring gave me no clue as to what the function is doing.
2. The docstring of `parrot-rotate-next` references `ROTATIONS` which is not a parameter. What is a "reversed rotation list"?
3. `parrot-rotate-next` uses quite a few nested `if` calls with the `THEN` part throwing `error`. To avoid too much indentation, Emacs Lisp code often uses sequences of `when` calls instead:

    ```emacs
    (let (…)
      (when (…)
        (error …))
      …
    ```

4. The function `parrot-rotate-word-at-point` is very long. I would suggest decomposing it into smaller ones which will force you to find names and docstrings to explain what the code is doing.

DamienCassou:
I just forgot to say the most important: I really like the code of your package. I'm really impressed as I would never have guessed it could be your first package. I can't wait to see what will be your next one!

Please be aware that my feedback above is only my own opinion and you are free to follow my recommendations or totally ignore it :-). I think the package is mostly ready for inclusion so don't be afraid by the amount of feedback, it's all optional.

dp12:
Thanks for your detailed and valuable comments, which are super helpful for a new elisper like me! I will try to work in the changes you suggested when I get a chance this week.

dp12:
Okay, I've updated the parrot repo with changes for most of the fixes suggested. 

One thing I didn't change was pulse-flag, which is a special variable defined in the pulse.el package - it prevents the highlight from going away with a timer when the user performs a rotation. 

The `parrot-rotate.el` file specifies lexical binding, but since pulse-flag is defined with defcustom, it is dynamically scoped and can be set with the let statement.

DamienCassou:
In parrot-rotate.el, the call to `declare-function` is broken (please enable flycheck to detect those things yourself). You need to put double-quotes around the filename. Also, it's quite strange that you need this circular dependency at all.

In parrot-rotate.el, you need to declare `pulse-flag` to avoid a compiler warning: `(defvar pulse-flag)` should do.

The rest if for parrot.el.

1. I don't think `(featurep 'parrot-mode)` is going to return anything else than `nil` as the feature is called `parrot` and not `parrot-mode` (this is the name that is `provide`d at the end of the file). 
2. You might want to use `bound-and-true-p` instead of `(and (boundp 'parrot-mode) parrot-mode)`
3. I suspect that milliseconds might be a better unit for `parrot-animation-frame-interval` than seconds. You might want to change that. The type of the `defcustom` could then be `integer` instead of `float`.
4. Even if the docstring of `defvar`s is optional, I suggest adding them to ease understanding.
5. I would probably extract a `parrot-sequence-length` function which takes a parrot type as parameter and return a number (e.g., 10 for the "default" type). This would simplify `parrot-set-parrot-type` quite a bit.
6. Instead of displaying a message if a parrot type is not found, you could use `user-error`. Also you might want to pass `t` to `completing-read` for the `REQUIRE-MATCH` parameter to force the user to select a valid type
7. in `parrot-load-frame`, you might want to use `expand-directory` instead of `concat` and `"/"`. This is more portable. Also, there are 2 very similar `format` calls that you might want to extract into a dedicated function to avoid duplication and to ease code maintenance.

recipes/parrot
```diff
@@ -0,0 +1,4 @@
+(parrot
+ :fetcher github
+ :repo "dp12/parrot"
+ :files (:defaults "img"))
```
DamienCassou: please add a new-line character at the end of the last line

DamienCassou:
Very good job!

PR #5738, opened 2018-09-21
Add hideshowvis

sheijk:

Add little +/- icons in the Emacs fringe for regions which hideshow.el can hide.

https://github.com/sheijk/hideshowvis

Maintainer

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/hideshowvis
+++ b/recipes/hideshowvis
@@ -1,0 +1,1 @@
+(hideshowvis :repo "sheijk/hideshowvis" :fetcher github)


purcell:
Thanks!

This [autoloaded function which defines custom vars when called](https://github.com/sheijk/hideshowvis/blob/614e856620445166a60c00a1c4653dbd59c871c9/hideshowvis.el#L225-L272) makes me think your mental model for autoloads etc. is a bit muddled. What are you trying to achieve?

sheijk:
The autoload cookie is there so calling hideshowvis-symbols will load the lib. Symbols are defined lazily because this functionality is optional (it used to be part of the example config) and used to define functions called hs-.. and IIRC had risk of unintended interactions with hideshow. I think now that I've renamed them all everything but the setq could be moved out of it

sheijk:
I plan to do this the next time I add new functionality to the package so I won't have to do all the testing twice as it doesn't add any functionality. I've added a ticket to not forget it. So feel free to proceed

sheijk:
@tarsius: ping

tarsius:
pong

I don't now what @purcell's intentions were, but I was going to wait until you have taken care of the issue raised by him before merging the recipe.

sheijk:
Hm, this might take a while. Last release before updating this was 5 years ago and there has been no need for any new features since then nor any bug reports and this might stay like this for a couple more years. So I think the cost/benefit for this is pretty bad. It adds risk and takes testing effort for nearly nothing other than the code being a bit more nice and fuzzy

@purcell is your plan to wait? In that case I'll probably cancel the pull request and redo it if I ever make another version

tarsius:
That's a bit of a radical stand on stability. No change is completely risk free, so maybe the requested change actually will break something, but if you spend—I don't know—one hour on testing you can significantly reduce the risk. Maybe the reason there weren't any bug reports during the last five years is that nobody installed it during those years because it was not available through `package.el`. In any case, if something *does* break, then that isn't the end of the world—this isn't a mission critical package.

purcell:
> is your plan to wait?

Yes, but I sent you https://github.com/sheijk/hideshowvis/pull/4. There are more nitpicks with this package, but this is the big one.

> So I think the cost/benefit for this is pretty bad. It adds risk and takes testing effort for nearly nothing other than the code being a bit more nice and fuzzy

When you're sharing code with the community, it matters that it's in decent shape because people will read and copy it. So the community standards are important, and following them is a sign of care for the community. Yes, any little change can break a package, but publishing new versions with MELPA is automatic, so let's not be afraid of making things better!

sheijk:
@tarsius: I guess we have quite different views on stability then. I believe package managers delivering broken code are something which should never happen as users might depend on their tools to get important work done. I've been bitten by this often enough in Emacs that even after moving to melpa-stable I still have to make a backup of my elpa dir before each update. I've had packages either breaking or breaking my config or packages not being on Melpa anymore when I'm installing on a new machine way too often. Many times this is quick to fix but sometimes it isn't. And even a nice-to-have package like this can keep Emacs from loading the config so I think I should not distribute code to anyone which isn't well tested. Also hideshowvis is on Marmalade and (unfortunately) Emacs Wiki and I see a couple thousand mentions of it on Github so I can tell it is getting used and available through package.el

sheijk:
> > is your plan to wait?
> 
> Yes, but I sent you [sheijk/hideshowvis#4](https://github.com/sheijk/hideshowvis/pull/4). (There are some more nitpicks with this package, but this is the big one, and the only thing I'm inclined to raise.)

@purcell Thanks for the pull request. My issue with this is still that I have to spend 1-2 hours of time on testing which I'd prefer not to do twice (once now and again for the next update with actual features)

> When you're sharing code with the community, it matters that it's in decent shape because people will read and copy it.

This is a good argument. That's the way I ended up with this code style, too after all. I will have a look at this as soon as I get around it (should be the next few days) and apply the pull request. But I'd like to hear your other nitpicks, too. The package has plenty of issues but I'm happy to fix anything which doesn't require a rewrite. If I spend the time on testing now I can as well fix other things

purcell:
> But I'd like to hear your other nitpicks, too.

Cool. Primarily:

- Avoid faces using boxes and only foreground or background. Instead `:inherit` standard faces wherever possible so that this automatically looks acceptable with all pre-existing themes.
- Add the global minor mode as noted in the "to-dos", where `hideshowvis-enable` is the "maybe-turn-on" function.
- Rename `hideshowvis-enable` to `hideshowvis-maybe-enable`, and alias the old name with `define-obsolete-function-alias`.
- Typo: "Hiddent" => "Hidden" (I think I fixed that in my PR, actually)
- Rather than "wrapping" `hs-minor-mode`, consider making this a separate minor mode which can be toggled on/off whether `hs-minor-mode` is enabled or not. (Related issue: turning off `hideshowvis-minor-mode` would currently not disable `hs-minor-mode`, which is weird.)

DamienCassou:
I'm closing this one for now as @purcell says he preferred to wait for https://github.com/sheijk/hideshowvis/pull/4 to be merged before merging this PR. Please reopen if something changes.

PR #5741, opened 2018-09-22
Add espy

walseb:

Espy scans a password file for headers with passwords and prompts the user for which headers password to copy to killring.

https://github.com/walseb/espy

Author and maintainer

**None needed** 

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/espy
+++ b/recipes/espy
@@ -1,0 +1,1 @@
+(espy :fetcher github :repo "walseb/espy")


alphapapa:
Hi,

I haven't looked at the code much, but here are a few issues I can think of:

1.  Putting passwords in the kill-ring seems like maybe not a good idea to me, especially because users may have a package configured that saves the kill-ring across sessions, in which case it will be saved to disk in plain-text.
2.  If you're going to use Org files, it seems a shame to break the file format by putting the password on a line immediately after the heading, which prevents property drawers from working.
3.  It would seem like a good idea to put the actual password-fetching code in a library function so other code could get the password directly, essentially making it like a password-storing backend.

walseb:
> Putting passwords in the kill-ring seems like maybe not a good idea to me, especially because users may have a package configured that saves the kill-ring across sessions, in which case it will be saved to disk in plain-text.

I have thought about this but I can't find an alternative. I could do `(insert password)` but that would not work for x windows or the minibuffer if recursive editing is off. I have read `simple-secrets.el` and `pass.el` and they also just copy directly to the killring. There was some package that also erases all passwords from the killring after a set time, I might look into this but right now I think it's better if it's handled by the user and it wouldn't solve the problem you mentioned anyway.

> If you're going to use Org files, it seems a shame to break the file format by putting the password on a line immediately after the heading, which prevents property drawers from working.

You can actually put the password anywhere inside a header, it just picks the first occurrence of a password inside every heading. I updated the screenshot to show this

> It would seem like a good idea to put the actual password-fetching code in a library function so other code could get the password directly, essentially making it like a password-storing backend.

Fixed

walseb:
> Putting passwords in the kill-ring seems like maybe not a good idea to me

I just updated espy to support executing a shell command with the password instead of copying it to the killring. Not sure how to do it though because `(shell-command "echo test | xclip -selection c ")` freezes emacs for me. I also added support for getting a username (i realized this was missing when I started using it myself heh) and changed around the defaults.

alphapapa:
Cool.  :)

purcell:
Thanks. Merging now. Most of those `defvar`s should be `defcustom`s instead, with a new parent `defgroup`. Some are concatenated to make regexps, so they should have `:type 'regexp`.

PR #5742, opened 2018-09-23
merlin: beta branch is outdated, use master

edwintorok:
`beta` branch of merlin is 105 commits behind latest release v3.1.0.
It would be better to fetch master or the latest release instead.

/cc upstream maintainers @let-def @trefis


purcell:
Yep, we always prefer to track `master`.

PR #5743, opened 2018-09-26
Add yaml-imenu, which enhances the imenu support in yaml-mode

knu:

yaml-imenu enhances the imenu support in yaml-mode.

https://github.com/knu/yaml-imenu.el

None

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/yaml-imenu
+++ b/recipes/yaml-imenu
@@ -1,3 +1,3 @@
 (yaml-imenu :repo "knu/yaml-imenu.el"
             :fetcher github
-            :files ("*.el" "*.rb"))
+            :files (:defaults "*.rb"))
--- a/recipes/yaml-imenu
+++ b/recipes/yaml-imenu
@@ -1,0 +1,3 @@
+(yaml-imenu :repo "knu/yaml-imenu.el"
+            :fetcher github
+            :files ("*.el" "*.rb"))


purcell:
Thanks. A couple of notes:

- This [auto-enabling of the package](https://github.com/knu/yaml-imenu.el/blob/b1d36aa4d1abf37c9874f8239172dc8980b34c40/yaml-imenu.el#L110-L113) is inappropriate. It should be possible to install this package but not activate it, just like the packages in the Emacs standard library. Instead provide an autoloaded `yaml-imenu-enable` which users can call from their init file, or add to their `yaml-mode-hook`.
- If a yaml file contains malicious content, is it still safe to open that file using this code? You should make this clear.

knu:
> - This [auto-enabling of the package](https://github.com/knu/yaml-imenu.el/blob/b1d36aa4d1abf37c9874f8239172dc8980b34c40/yaml-imenu.el#L110-L113) is inappropriate. It should be possible to install this package but not activate it, just like the packages in the Emacs standard library. Instead provide an autoloaded `yaml-imenu-enable` which users can call from their init file, or add to their `yaml-mode-hook`.
OK.

> - Don't use `"*.el"` in the recipe: please use `:defaults` instead.
Thanks, that's nice to know.

> - If a yaml file contains malicious content, is it still safe to open that file using this code? You should make this clear.
The parser just parses a document without evaluating it, so it should be safe.

Thanks for your review, I've addressed the above points.

purcell:
Great, thanks - merged now. :tada:

PR #5744, opened 2018-09-26
Add rem

baygeldin:

The main purpose of rem.el is to simplify building text interfaces for Emacs. It provides a bunch of utilities that when combined together allow to structure your code in the MVC way. At the core, what it does is memoizing, but without potential memory leaks. The approach is similar to React/Redux, but simpler and adapted for Emacs reality.

https://github.com/baygeldin/rem.el

Author and maintainer

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/rem
+++ b/recipes/rem
@@ -1,0 +1,1 @@
+(rem :fetcher github :repo "baygeldin/rem.el" :files ("rem.el"))


purcell:
Thanks - interesting idea! Quick notes:

- Please remove the `:files` spec from the recipe in order to use the defaults. If this means moving `rem-example.el` to an `examples` subdir so it is not bundled into the package, this would be preferable.
- You haven't declared dependencies on the packages you're using. They are not built into Emacs 24.4.
- You should `(require 'cl-lib)` explicitly.

Re. your gotcha note

> Make sure to enable lexical binding if you plan on using rem.el, because it relies on it.

perhaps your macros can simply check that lexical-binding is enabled when they are expanded?

DamienCassou:
Closing this PR as there hasn't been any answer from the author.

PR #5745, opened 2018-10-02
Add recipe for go-imenu

brantou:

imenu for go language

https://github.com/brantou/go-imenu.el

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/go-imenu
+++ b/recipes/go-imenu
@@ -1,0 +1,3 @@
+(go-imenu
+ :repo "brantou/go-imenu.el"
+ :fetcher github)


purcell:
Thanks! Some tips:
- [This](https://github.com/brantou/go-imenu.el/blob/5e2c0f28e3da32d2f9b2e83a871c774e640fe1c5/go-imenu.el#L107-L117) could be written as a `pcase` expression on `type`, to avoid repeating the `string-equals` forms.
- [This](https://github.com/brantou/go-imenu.el/blob/5e2c0f28e3da32d2f9b2e83a871c774e640fe1c5/go-imenu.el#L73-L77) is equivalent to just listing the initially-null symbols in the `let` clause, eg. `(let (index-alist index-var-alist ...) ...)`
- [cmd-args](https://github.com/brantou/go-imenu.el/blob/5e2c0f28e3da32d2f9b2e83a871c774e640fe1c5/go-imenu.el#L87) could be initialised in the `let` too
- Use `when` instead of `and` in places [like this](https://github.com/brantou/go-imenu.el/blob/5e2c0f28e3da32d2f9b2e83a871c774e640fe1c5/go-imenu.el#L129)
- This code will break when `buffer-file-name` is nil, e.g. [this line will break](https://github.com/brantou/go-imenu.el/blob/5e2c0f28e3da32d2f9b2e83a871c774e640fe1c5/go-imenu.el#L68) because `(file-exists-p nil)` throws an error

Nonetheless, I'll merge this now, and you can fix the above in your own time.

PR #5750, opened 2018-10-08
Move org-emms to gitlab

jagrg:
I'm moving the project to gitlab.


tarsius:
Please provide links to the old and new repositories so that we don't have to construct the urls ourselves. It makes a difference when dealing with thousands of pull-requests.

jagrg:
You mean here or is there something else I need to do in the PR? The old link is:

https://github.com/jagrg/org-emms

The new link is:

https://gitlab.com/jagrg/org-emms

PR #5752, opened 2018-10-11
Add pacfiles-mode recipe

UndeadKernel:

Major mode to merge configuration updates (e.g., *.pacnew files) left by the *pacman* tool in Arch-based 
 GNU/Linux systems.

https://github.com/UndeadKernel/pacfiles-mode

I'm the package's author and maintainer. 

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/pacfiles-mode
+++ b/recipes/pacfiles-mode
@@ -1,0 +1,1 @@
+(pacfiles-mode :repo "UndeadKernel/pacfiles-mode" :fetcher github)


purcell:
Thanks! Some notes:

- Require `cl-lib` instead of `cl-seq` [here](https://github.com/UndeadKernel/pacfiles-mode/blob/1cf3fdf79317303d382d71e2f92596de2127cdb1/pacfiles-mode.el#L33) -- the standalone version of `cl-lib` doesn't have a separate `cl-seq.el`.
- You should always use `shell-quote-argument` when inserting filenames that might contain spaces or quotes into command strings, such as [here](https://github.com/UndeadKernel/pacfiles-mode/blob/1cf3fdf79317303d382d71e2f92596de2127cdb1/pacfiles-mode.el#L45).
- All of [this code](https://github.com/UndeadKernel/pacfiles-mode/blob/1cf3fdf79317303d382d71e2f92596de2127cdb1/pacfiles-mode.el#L250-L260) relates to user preferences, and you should not put this in the major mode initialisation code, but instead put it in your own init files.

Hope that helps!

UndeadKernel:
Thanks @purcell, all your suggestions were spot on.
I've addressed the changes you suggest and made notes about the changes in the commit description.
Do let me know if you find anything else worth addressing.

purcell:
Great, merged now - thanks! Welcome to MELPA. :tada:

PR #5758, opened 2018-10-14
Add debian-el and dpkg-dev-el

tarsius:
* https://github.com/melpa/melpa/issues/5449
* https://salsa.debian.org/emacsen-team/debian-el
* https://salsa.debian.org/emacsen-team/dpkg-dev-el

@purcell Is there anything you want to see improved in these packages before adding them here? Are you okay with just adding these two packages instead of adding the individual libraries as packages in their own right (and removing `debian-changelog-mode`)?


purcell:
Yeah, works for me. It's a bit messy, but I think getting these packaged is a step forward. I wasn't clear from #5449 if this is ready to go, but please go ahead and merge if so.

tarsius:
There's still some work to be done on `dpkg-dev-el.el`.

PR #5764, opened 2018-10-20
Add recipe for poly-ansible

mavit:

This mode is for editing [YAML files for Ansible](https://docs.ansible.com/ansible/latest/reference_appendices/YAMLSyntax.html) containing embedded [Jinja2 templating](http://jinja.pocoo.org/docs/).

This is a [polymode](https://polymode.github.io/), gluing [`jinja2-mode`](https://github.com/paradoxxxzero/jinja2-mode) into [`yaml-mode`](https://github.com/yoshiki/yaml-mode).

https://gitlab.com/mavit/poly-ansible

I am the author.

None needed.

### Checklist

### DIFF
--- a/recipes/poly-ansible
+++ b/recipes/poly-ansible
@@ -1,0 +1,4 @@
+(poly-ansible
+ :fetcher gitlab
+ :branch "test"
+ :repo "mavit/poly-ansible")
--- a/recipes/poly-ansible
+++ b/recipes/poly-ansible
@@ -1,4 +1,3 @@
 (poly-ansible
  :fetcher gitlab
- :branch "test"
  :repo "mavit/poly-ansible")


mavit:
> Thanks, and sorry for the slow turnaround.

No worries.

>     * As `package-lint` will have noted when you ran it, `pm-inner/jinja2` should be called something looking like `poly-ansible-*`. I don't think there's a polymode-related reason that's not possible here.

Only that this is the naming convention that the [polymode documentation suggests](https://polymode.github.io/defining-polymodes/) and that polymode itself uses.  I would prefer to remain consistent with that if acceptable.

>     * `$` in `auto-mode-alist` regexes should always be `\\'` instead: matches end-of-string, rather than end-of-line.

Fixed.

>     * The first `auto-mode-alist` entry is quite loose due to the `.*`. Is that for nested directories? Otherwise `[^/]*` might be more appropriate.

Indeed, the YAML files will often be found several directories deep.

Thanks.

purcell:
Merged, thanks.



> Only that this is the naming convention that the [polymode documentation suggests](https://polymode.github.io/defining-polymodes/) and that polymode itself uses. I would prefer to remain consistent with that if acceptable.

Sure, but I don't think they're suggesting a convention there, just giving an example. So when possible, I think we should prefer the community conventions. I'll leave that up to you. :-)

PR #5765, opened 2018-10-20
Add recipe for wiki-drill

mtekman:

Use the Wikipedia summaries from jozefg's wiki-summary package and generates org-drill entries from them.

https://github.com/mtekman/wiki-drill.el

Maintainer

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/wiki-drill
+++ b/recipes/wiki-drill
@@ -1,0 +1,1 @@
+(wiki-drill :fetcher github :repo "mtekman/wiki-drill.el")


purcell:
Hi, from just a quick look, there are a bunch of little issues with this code -- as a first step, you should run `package-lint` as requested, and fix the issues it reports, along with  any issues highlighted by `flycheck`.

mtekman:
@purcell thanks for the feedback, I have made a few commits and it now compiles clean, passes the checkdoc, and the package-lint checks.

mtekman:
@purcell thank you so much for the reviews!

I've made the changes you requested, except

> You're using wiki-summary without depending on a package which provides it.

I did not know how to resolve this. I looked for example in other PRs on how to require other  MELPA packages, but my search was fruitless. Can I ask what I should be doing here?

purcell:
Easy! You'd just add `(wiki-summary "0")` to your existing `Package-Requires:` header. Normally you'd use a non-"0" version number, but there doesn't appear to be a stable release number, and "0" is preferable to depending on a specific MELPA snapshot version.

mtekman:
Oh haha nice, I completely missed that!

Okay, so I added the dependency to the requires line, but now I get this error in the linting:

> 8:37: error: Package wiki-summary is not installable.

is this something I can ignore?

DamienCassou:
1. the `defcustom`s following a `defgroup` don't have to specify the `:group` because the group is inferred
2. please add `-*- lexical-binding: t; -*-` at the end of the first line
3. if `wiki-drill--file` should be visible to the user, then it should not have two dashes in its name (`wiki-drill-file`). Two dashes indicate a private element. It it should not be visible to the user, then it should be a `defvar`, not a `defcustom`. Same remark for the other `defcustom`.
3. Please run linters specified in [CONTRIBUTING.org](https://github.com/melpa/melpa/blob/master/CONTRIBUTING.org). They would have told you about that you should define the variables `binding-clozer-mark` and `binding-submit` before using them.
4. Elisp packages don't usually create `defcustom` variables for each keybinding. A mode map is enough: users can add and remove the keybindings they want from any map.
5. You don't need a `progn` to specify the else-part of an `if`: `(if condition true-body false-body1 false-body2…)`
4. The Commentary section could be extended to say what happens after "[prompting the user] for an article to search".
5. You don't have to give a docstring to `defvar` variables, but I recommend doing so to ease maintenance.
6. In the definition of `wiki-drill-clozer-mode`'s keymap, you assign lambda expressions to keys only to add the `interactive` token. I suggest adding the token to the functions directly: `(defun wiki-drill-clozer-brackets () "Surrounds with [[words||hint]]." (interactive) …)`. That way, the keymap becomes easier to understand, `C-h k` works, and users can do `M-x wiki-drill-clozer-brackets` if they want to.
7. The `progn` in `wiki-drill-refile-into-file` is useless.
8. For a user to be able to type `M-x wiki-drill`, the function `wiki-drill` must be marked `(interactive)`.
9. Adding the `;;;###autoload` cookie everywhere is not desired. Only add it to the functions that may be called before `wiki-drill.el` is loaded. For example, the documentation says the user should type `M-x wiki-drill` so `wiki-drill` sounds like a good target for an `;;;###autoload` cookie.

mtekman:
@DamienCassou I've made the changes you requested (and thanks a lot for the very informative review!)

One exception is (5) where I have kept the `defcustom` keymap bindings so that they can also appear in the FlashCard buffer as instructions on what the user should do.

Other than that; linting, checkdoc have revealed no errors to me except that the wiki-summary requirement does not have a stable version, but this I cannot control.

purcell:
> One exception is (5) where I have kept the `defcustom` keymap bindings so that they can also appear in the FlashCard buffer as instructions on what the user should do.

No need. You can just get a string describing the current binding from the keymap, e.g.

```
(key-description (where-is-internal 'your-command-here wiki-drill-clozer-mode t))
```

(Of course, `'your-command-here` requires that you bind proper functions, not `lambda`s into your keymap.)

P.S. You should add a `defgroup` form before your `defcustom`s. :-)

DamienCassou:

PR #5767, opened 2018-10-21
Add lsp-dart

twlz0ne:

Dart support for lsp-mode.

https://github.com/twlz0ne/lsp-dart

Author&Maintainer

**None needed** 

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/lsp-dart
+++ b/recipes/lsp-dart
@@ -1,0 +1,3 @@
+(lsp-dart
+ :fetcher github
+ :repo "twlz0ne/lsp-dart")


purcell:
Thanks. The commentary docs mention `lsp-dart-enable`, but that function isn't defined anywhere that I can see, so you might need to fix the doc.

Looks good otherwise, though, so merging now! :-)

twlz0ne:
@purcell The `lsp-dart-enable` function is generated by `lsp-define-stdio-client`, for example:

```elisp
;; lsp-foo.el
(lsp-define-stdio-client lsp-foo ...)

;; lsp-mode.el
(cl-defmacro lsp-define-stdio-client (name ...
  ...
  (let ((enable-name (intern (format "%s-enable" name))))
    `(progn
       ...
       (defun ,enable-name ()
 ...)
```

https://github.com/emacs-lsp/lsp-mode/blob/7f3971af37cf20fb42fea705951bba1708211497/lsp-mode.el#L167-L171

purcell:
Oh cool, I did wonder if that was the case. Thanks.

PR #5769, opened 2018-10-21
Add elsa

Emiller88:

Elsa is a tool that analyses your elisp code without loading or running it. It can track types and provide helpful hints when things don't match up before you even try to run the code.

https://github.com/emacs-elsa/Elsa

An enthusiastic user

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/elsa
+++ b/recipes/elsa
@@ -1,0 +1,1 @@
+(elsa :fetcher github :repo "emacs-elsa/Elsa")


Fuco1:
This won't work properly because the `bin` directory won't be packaged.  Look at how `buttercup` does it.

Fuco1:
This won't work properly because the `bin` directory won't be packaged.  Look at how `buttercup` does it.

Fuco1:
This won't work properly because the `bin` directory won't be packaged.  Look at how `buttercup` does it.

Fuco1:
This won't work properly because the `bin` directory won't be packaged.  Look at how `buttercup` does it.

Fuco1:
This won't work properly because the `bin` directory won't be packaged.  Look at how `buttercup` does it.

Fuco1:
This won't work properly because the `bin` directory won't be packaged.  Look at how `buttercup` does it.

Fuco1:
This won't work properly because the `bin` directory won't be packaged.  Look at how `buttercup` does it.

Fuco1:
This won't work properly because the `bin` directory won't be packaged.  Look at how `buttercup` does it.

Fuco1:
This won't work properly because the `bin` directory won't be packaged.  Look at how `buttercup` does it.

Fuco1:
This won't work properly because the `bin` directory won't be packaged.  Look at how `buttercup` does it.

Fuco1:
This won't work properly because the `bin` directory won't be packaged.  Look at how `buttercup` does it.

Fuco1:
Sorry for the duplicates.... github was acting crazy in the morning.

vermiculus:
Looking at `bin/elsa`, is it really necessary to run the necessary code? It seems to only be a convenience wrapper around `elsa-run`.

vermiculus:
Is `bin` actually necessary? `bin/elsa` seems to be a very thin wrapper around `elsa-run`.

Fuco1:
@vermiculus There still needs to be some binary that will run the thing.  You can absolutely not run it in your "editing" instance of Emacs.

vermiculus:
In other words, it must be a separate emacs process, yes? I'd be interested to learn how this limitation came to be, but I can leave that as an exercise to myself for a weekend :-) Your word is more than enough here.

If there was a way to kick off a subprocess from within emacs, that would accomplish the same goal, yes?

Fuco1:
You need to have clean namespace for the symbols.  If we are analyzing something in a tainted environment it can not work properly (load paths, separation, repeatable results...).

Imagine you package.el update elsa and run analysis again from the same process but some data is saved in an old format.  A basic user will be completely lost.

vermiculus:
Would an interactive command to start an emacs subprocess with the right arguments be an acceptable way to kick off the checks?

Fuco1:
I save some results of the analysis on the interned symbols.  These persist for the whole session.  So it's not so much about migration than it is about consistency.  You can change the code and re-run the analysis and it can mix up old and new results in some weird ways and give wrong results.

I simply did not want to deal with this.  Running a new process is a super easy way to ensure everything is clean.

vermiculus:
Thanks for the info :-) the first question stands, though

Fuco1:
Yes, it would, but I don't find interactive ways to do testing very appealing.  I always run ert/buttercup from shell to be sure.  How would something like flycheck use an interactive command?  Elsa is a tool mostly ment to be used with something like that (=flycheck), not really "interactively" with yet another custom-made interface.

vermiculus:
It may not be the most effective way to use Elsa, but it could be *a* way to use Elsa. I agree that creating a new ERT-like interface for it would be way too much work (unless the relevant code can be easily re-used), but some Emacs-based entry point would be prudent for something installable over ELPA -- even if that entry point is just `shell-command-to-string` or similar.

purcell:
So what's the conclusion with this, @Fuco1 ? That the recipe makes sense, but only if we add the bin wrapper? Not sure how to proceed here.

purcell:
(Closed accidentally, sorry.)

Fuco1:
@purcell Yes to make it work with Cask which is my preferred method of how people use this we need that bin wrapper.

I don't know if I can push a fix to this branch.

purcell:
Great, thanks folks.

PR #5770, opened 2018-10-22
Add recipe for ob-html-chrome.

nikclayton:

Org-Babel support for rendering HTML to PNG files using Chrome.

This is similar functionality to ob-browser, without the PhantomJS requirement.

https://github.com/nikclayton/ob-html-chrome

Maintainer

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/ob-html-chrome
+++ b/recipes/ob-html-chrome
@@ -1,0 +1,1 @@
+(ob-html-chrome :repo "nikclayton/ob-html-chrome" :fetcher github)


nikclayton:
Ping.

purcell:
Thanks. Some notes:

- The bare package names in `Package-Requires` should be replaced with `(name "version")` pairs, as `package-lint` should have pointed out.
- Generally speaking, don't depend on `org` as a separate package.
- I think there's a docstring missing [here](https://github.com/nikclayton/ob-html-chrome/blob/487e199af0d27b938df695ddcc18dfea31ee64aa/ob-html-chrome.el#L24)

Otherwise looks fine, and can be merged by any maintainer when the above points are addressed. :-)

DamienCassou:

nikclayton:
Thanks for the feedback. Fixed and re-opening. 

BTW, that's not a missing docstring in the defcustom call -- that's the default value for that symbol (the empty string). The docstring is the third argument.

PR #5771, opened 2018-10-23
Add recipe for ox-slack

titaniumbones:
Please see PR for more details.

ox-slack provides a simple org-mode export backend for the slack instant messaging syntax.

https://github.com/titaniumbones/ox-slack

I am the maintainer.

I'm the author, so while some changes may be needed, I'm the one responsible for fixing them :-) 

### Checklist

Please confirm with `x`:

- [x] I've used the latest version of [package-lint](https://github.com/purcell/package-lint) to check for packaging issues, and addressed its feedback **see below**

**Note:** I've followed what I believe is the convention for org-mode exporters: functions are named `org-slack-xxx`, but the package is named `ox-slack`. `package-lint` gives a set of warnings for this, but I believe I'm still following best practice for org-mode extensions.

### DIFF
--- a/recipes/ox-slack
+++ b/recipes/ox-slack
@@ -1,0 +1,1 @@
+(ox-slack :fetcher github :repo "titaniumbones/ox-slack")


purcell:
> I've followed what I believe is the convention for org-mode exporters

Yep, that's fine.

`package-lint` should have told you about the malformatted `Package-Requires` line, though, which should be addressed.

Also, you can remove the [pointless `let*` here](https://github.com/titaniumbones/ox-slack/blob/24ba8c1595b0fead9a4a0619fd98a9c4face0556/ox-slack.el#L69).

Looks good otherwise, so this can be merged when the `Package-Requires` line is fixed up.

titaniumbones:
should be fixed in #1695056

thanks for the feedback! Package-Requires was also missing the initial `:` so `package-lint` didn't pick it up I guess. complained about `gfm` but I don't believe the version number is important there.

purcell:
Thanks. You still need the version number in every `Package-Requires` entry in order for older package.el versions to parse the line correctly. That's why `package-lint` is pedantic about it.

titaniumbones:
should be fixed now I hope! Thx for the patience & sorry for the errors.

purcell:
Super, merged! Welcome to MELPA. :tada:

PR #5773, opened 2018-10-26
Add new recipe for evil-better-visual-line

YourFin:

Provides an improved experience on `evil-[next/previous]-visual-line` (common stand in j/k) in that it falls back to non-visual line movement in visual line and block mode

https://github.com/yourfin/evil-better-visual-line

Maintainer

Got the package linting correctly!

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/evil-better-visual-move
+++ b/recipes/evil-better-visual-move
@@ -1,0 +1,1 @@
+(evil-better-visual-move :fetcher github :repo "YourFin/evil-better-visual-move")


purcell:
Looks good - merging now! Welcome to MELPA. :tada:

tarsius:
I have renamed this to `evil-better-visual-line` because that is the actual name of the package.

tarsius:
I just spend half an hour on this because Github says "You must be logged in to do that." when it should say "You cannot fork this either because it doesn't exist or because it is private repository, which we you don't have access. (In the latter case we would not want to disclose this information to you, so we have to be a bit fague)". So there I was trying to figure out why all those other things that require authentication work but not forking this. :rage:

YourFin:
Sorry about that; I think I put this together rather late at night and forgot I had changed the name.

PR #5774, opened 2018-10-26
sql-presto: Add New Recipe

kat-co:

Allows users to utilize [Presto](https://prestodb.io/) via SQLi.

https://github.com/kat-co/sql-prestodb

It me: the maintainer.

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/sql-presto
+++ b/recipes/sql-presto
@@ -1,4 +1,4 @@
 (sql-presto
  :fetcher github
  :repo "kat-co/sql-prestodb"
- :files ("artifacts/*"))
+ :files ("artifacts/*.el"))
--- a/recipes/sql-presto
+++ b/recipes/sql-presto
@@ -1,0 +1,4 @@
+(sql-presto
+ :fetcher github
+ :repo "kat-co/sql-prestodb"
+ :files ("artifacts/*"))


purcell:
Looks good, and this is cool to see given that I was excited to see Presto being used to add SQL support to Apache Pulsar recently.

I note that you're generating the elisp file using a makefile, but the two are out of sync: the current generated artifact doesn't contain the `;; Package-Requires: ((emacs "24.4"))` line. Is that necessary?

In general your generation would be simpler if you had a more conventional layout in `src/sql-prestodb.el` which included the header boilerplate, with just a `$$COMMENTARY$$` placeholder or similar which the Makefile could replace. Spreading the headers across a separate file is a bit more complicated with no obvious benefit, and it makes it less convenient to work with the source file (e.g. you'd potentially get different behaviour from code evaluated there due to `lexical-binding` not being defined in it).

DamienCassou:

kat-co:
Hey all, thanks for the great review. I apologize for the long delay on making the requested changes. They are pushed up under the 1.0.2 tag:

- I consider generating the comments within the `Makefile` a failed experiment and have moved them into the source code as suggested.
- I've added a `sql-presto` `defgroup` as suggested.
- I've added the `;;;###autoload` cookie as suggested.

Thanks, and I hope this clears things for landing :)

purcell:
That's mostly great - you just need to remove the backslashes in [this line](https://github.com/kat-co/sql-prestodb/blob/f3962ae8d437b19e7a2bb561ed7ee6f9ff27f5e0/artifacts/sql-presto.el#L7). In the meantime I'll re-open this, but please add a quick comment here when fixed so we know we can merge.

kat-co:
Thanks! This fix has been pushed and tagged as 1.0.3.

purcell:
> Thanks! This fix has been pushed and tagged as 1.0.3.

Ah, but it's not yet in `artifacts/sql-presto.el`, so I think perhaps a `make` is needed.

kat-co:
Dangit! Sorry, that's pushed up as 1.0.4 now :|

purcell:
Great - merged now! Welcome to MELPA. :tada:

kat-co:
:tada: 

And thanks for the warm welcome, and great reviews!

PR #5776, opened 2018-10-28
Add recipe for company-phpactor

kermorgant:

company backend using [phpactor](https://github.com/phpactor/phpactor)

https://github.com/emacs-php/phpactor.el

Co-author

 **None needed** 

### Checklist

### DIFF
--- a/recipes/company-phpactor
+++ b/recipes/company-phpactor
@@ -1,0 +1,1 @@
+(company-phpactor :repo "emacs-php/phpactor.el" :fetcher github :files ( "company-phpactor.el"))


purcell:
Thanks. This is missing a `Package-Requires` dependency on `phpactor`, which it `require`s.

kermorgant:
Thanks for the review. Package-Requires dependency on phpactor is now added.

purcell:
Excellent, merging now, thanks. You might like to drop the use of `cl-lib` and its dependency, since `pcase` should be available in Emacs 24.3 as a built-in, and can be used much like `cl-case`.

purcell:
Also: welcome to MELPA! :tada:

PR #5777, opened 2018-10-28
Add recipe for load-bash-alias

daviderestivo:

Convert bash aliases into eshell ones.

https://github.com/daviderestivo/load-bash-alias

**None needed**

### Checklist

### DIFF
--- a/recipes/load-bash-alias
+++ b/recipes/load-bash-alias
@@ -1,0 +1,1 @@
+(load-bash-alias :repo "daviderestivo/load-bash-alias" :fetcher github)
--- a/recipes/load-bash-shell-aliases
+++ b/recipes/load-bash-shell-aliases
@@ -1,1 +1,0 @@
-(load-bash-shell-aliases :repo "daviderestivo/load-bash-shell-aliases" :fetcher github)
--- a/recipes/load-bash-shell-aliases
+++ b/recipes/load-bash-shell-aliases
@@ -1,0 +1,1 @@
+(load-bash-shell-aliases :repo "daviderestivo/load-bash-shell-aliases" :fetcher github)


purcell:
Thanks. Emacs 24.1 doesn't have `seq`, so you'll need another `Package-Requires` entry for that.

The `lbsa/` prefix naming for symbols is not permitted, sorry. Use the prefix `load-bash-shell-aliases-`, or rename the package to `load-bash-alias` if you want a shorter prefix (ie. `load-bash-alias-`).

daviderestivo:
> Thanks. Emacs 24.1 doesn't have `seq`, so you'll need another `Package-Requires` entry for that.
> 
> The `lbsa/` prefix naming for symbols is not permitted, sorry. Use the prefix `load-bash-shell-aliases-`, or rename the package to `load-bash-alias` if you want a shorter prefix (ie. `load-bash-alias-`).

I will update the package as per your suggestion. I have only one small question: is there any special reason why the lbsa/ prefix is not allowed? Any technical or stylistic reason?

purcell:
> I have only one small question: is there any special reason why the lbsa/ prefix is not allowed? Any technical or stylistic reason?

Yes - see the [emacs lisp coding conventions](https://www.gnu.org/software/emacs/manual/html_node/elisp/Coding-Conventions.html). There are existing popular packages which do not follow that rule, but that is an unfortunate historical mistake, and in MELPA we're now insisting that everyone follows the conventions. (One of those popular packages, yasnippet, actually changed from `/` to `-` in its prefix recently because of this!)

daviderestivo:
> Yes - see the [emacs lisp coding conventions](https://www.gnu.org/software/emacs/manual/html_node/elisp/Coding-Conventions.html). There are existing popular packages which do not follow that rule, but that is an unfortunate historical mistake, and in MELPA we're now insisting that everyone follows the conventions. (One of those popular packages, yasnippet, actually changed from `/` to `-` in its prefix recently because of this!)

Thanks a lot for the link. I will update the package accordignly.

purcell:
Looks great - just caught a little issue in https://github.com/daviderestivo/load-bash-alias/pull/1. When that's merged, we're good to go here. Also suggest that you change the `'emacs` parent group for the `defgroup` to `'eshell`.

daviderestivo:
> Looks great - just caught a little issue in [daviderestivo/load-bash-alias#1](https://github.com/daviderestivo/load-bash-alias/pull/1). When that's merged, we're good to go here. Also suggest that you change the `'emacs` parent group for the `defgroup` to `'eshell`.

@purcell I've merged your pull request (... thanks for having spot this silly mistake) and modified the parent group to 'shell.

purcell:
Great - merged now!

PR #5778, opened 2018-10-29
Add rust-auto-use.el

vmalloc:
This package automatically inserts `use` statements in Rust code based on the current symbol.

https://github.com/vmalloc/rust-auto-use.el

I am the developer and current maintainer

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/rust-auto-use
+++ b/recipes/rust-auto-use
@@ -1,0 +1,1 @@
+(rust-auto-use :repo "vmalloc/rust-auto-use.el" :fetcher github :files ( "rust-auto-use.el"))
--- a/recipes/rust-auto-use
+++ b/recipes/rust-auto-use
@@ -1,1 +1,1 @@
-(rust-auto-use :repo "vmalloc/rust-auto-use.el" :fetcher github :files ( "rust-auto-use.el"))
+(rust-auto-use :repo "vmalloc/rust-auto-use.el" :fetcher github)


vmalloc:
I tried to follow the guides as best as I can. This is my first contribution, so I apologize in advance if I missed something. Any feedback is greatly appreciated.

purcell:
Thanks! Notes:
- Please use a more complete boilerplate, as `M-x auto-insert` would insert in an empty `.el` file. Then the headers will be more complete, and the licence text will be explicit.
- Make `rust-auto-use-cache-filename` a `defcustom`, in a `defgroup`.
- Don't use `concat` to build filenames: prefer `expand-file-name`
- Add a `;;;###autoload` cookie comment before the definition of the `rust-auto-use` command.
- When invoking `emacs-lisp-mode`, also use `delay-mode-hooks`, to avoid the temp buffer having `flycheck`, `company` etc. enabled in it.

Hope that helps!

vmalloc:
@purcell thanks for the feedback! Can you clarify a bit about the last point? Do you mean embed a call to that function within the `with-temp-buffer` body?

purcell:
>  Can you clarify a bit about the last point? Do you mean embed a call to that function within the `with-temp-buffer` body?

Instead of `(emacs-lisp-mode)` you'd use `(delay-mode-hooks (emacs-lisp-mode))`.

vmalloc:
@purcell Thanks! Will get to it soon

vmalloc:
@purcell I pushed the changes corresponding to your feedback. I hope I did them correctly -- if not will gladly fix any further problems.

Thanks again!

purcell:
Great. Sent you some tweaks in https://github.com/vmalloc/rust-auto-use.el/pull/1, but this is basically looking good, so I'll merge this now. Thanks for you patience, and welcome to MELPA! :tada:

vmalloc:
Thanks!

PR #5780, opened 2018-10-30
Remove color-theme

tarsius:
For your consideration.

```
 Remove obsolete color-theme

Since Emacs 24.1, released in 2012, Emacs comes with support
for themes, making the older `color-theme' package obsolete
and incompatible.

`color-theme' is one of the package that Melpa gets from the
Emacsorphanage, which I update semi-automatically and not very
often from the upstream non-git/hg repository.

This particular repository was not updated in ten years, which
isn't surprising since it has been obsolete for six of those.

I have stopped updating this mirror now and its recipe could be
updated to fetch "updates" from the Emacsattic instead.  But we
might as well take this opportunity and remove `color-theme'
and all old-school themes.

There are six theme packages whose repositories contain old-school
themes beside modern themes.  There are no old-school-only theme
packages.

The recipes of three of those already limit the Melpa package to
the modern theme.  The other tree, two of which by Steve, didn't
do so until this commit.

I think these packages should also be renamed from `color-theme-*'
to `*-theme(s)'.  I would also recommend that the `color-theme-*'
files be removed from all repositories (or moved to a branch named
"obsolete" or something like that).
```

```
 [WIP] Rename theme packages

!! Due to issues mentioned below, this cannot actually be merged yet.

These packages should be named after the new-style themes they provide
instead of after the old-style themes their repositories still contain.

I would encourage you to also remove the obsolete themes from the
repositories, but that can happen in a second step.

The `solarized-theme.el' library of package `sellout-solarized-theme'
conflicts with the library by the same name of the `solarized-theme'
package.  I think it should be renamed to `sellout-solarized-theme.el'
as is done for `sanityinc-solarized-theme'.
```

### DIFF
--- a/recipes/color-theme-sanityinc-solarized
+++ b/recipes/color-theme-sanityinc-solarized
@@ -1,2 +1,0 @@
-(color-theme-sanityinc-solarized :repo "purcell/color-theme-sanityinc-solarized" :fetcher github)
-
--- a/recipes/color-theme-sanityinc-tomorrow
+++ b/recipes/color-theme-sanityinc-tomorrow
@@ -1,2 +1,0 @@
-(color-theme-sanityinc-tomorrow :repo "purcell/color-theme-sanityinc-tomorrow" :fetcher github)
-
--- a/recipes/color-theme-solarized
+++ b/recipes/color-theme-solarized
@@ -1,1 +1,0 @@
-(color-theme-solarized :repo "sellout/emacs-color-theme-solarized" :fetcher github)
--- a/recipes/sanityinc-solarized-themes
+++ b/recipes/sanityinc-solarized-themes
@@ -1,0 +1,4 @@
+(sanityinc-solarized-themes
+ :fetcher github
+ :repo "purcell/color-theme-sanityinc-solarized"
+ :files (:defaults (:exclude "color-theme-sanityinc-solarized.el")))
--- a/recipes/sanityinc-tomorrow-themes
+++ b/recipes/sanityinc-tomorrow-themes
@@ -1,0 +1,4 @@
+(sanityinc-tomorrow-themes
+ :fetcher github
+ :repo "purcell/color-theme-sanityinc-tomorrow"
+ :files (:defaults (:exclude "color-theme-sanityinc-tomorrow.el")))
--- a/recipes/sellout-solarized-theme
+++ b/recipes/sellout-solarized-theme
@@ -1,0 +1,4 @@
+(sellout-solarized-theme
+ :fetcher github
+ :repo "sellout/emacs-color-theme-solarized"
+ :files (:defaults (:exclude "color-theme-solarized.el")))


milkypostman:
Make it so!

purcell:
Sounds good in general, and this would be in the interest of positive forward progress, but the package is in the 93rd percentile of downloads -- I wonder how much of that is due to confusion. I've read that a bunch of people are still using `color-theme` because of the very wide range of bundled themes available for it.

An alternative to removal would be to deprecate the emacsorphanage version, by making the package description discouraging, and even adding warnings.

tumashu:
color-theme has been ported to new theme framework: https://github.com/emacs-jp/replace-colorthemes

purcell:
@tumashu Thanks, yes - I saw that after writing the above comment!

tarsius:
For now I have just changed the recipe to "update" `color-theme` from the attic instead of the orphanage.

> An alternative to removal would be to deprecate the emacsorphanage version, by making the package description discouraging, and even adding warnings.

Sounds good to me (we can do that even if the package is in the attic).

DamienCassou:

tarsius:
I have added documentation and warnings to the `color-theme` package. Please see the pull-request https://github.com/emacsattic/color-theme/pull/1. If you spot mistakes or have suggestions on how to improve this, then please comment on that pr.

----

Once I have merged that, we should also make changes elsewhere.

* @purcell As mentioned before I think you should rename your theme packages to be named after the name of the deftheme-theme instead of after the color-theme-theme that are being maintained in the same repositories. I would even go as far as removing the old themes completely (or at least move them to another branch).
* The `solarized-theme.el' library of package `sellout-solarized-theme' has to be renamed to `sellout-solarized-theme` because there is already another, vastly more popular `solarized-theme` package.

purcell:
Yeah, I'll have a think about what to do with my packages so that they don't just disappear. I agree, just have to consider the mechanics.

tarsius:
I have pushed the deprecation warning to `color-theme`'s `master` branch now.

I have also removed the commit that removed `color-theme` from this pull-request branch.

> Yeah, I'll have a think about what to do with my packages so that they don't just disappear. I agree, just have to consider the mechanics.

The remaining commit is how I would suggest you do it.

tarsius:
@sellout Please consider renaming your `solarized-theme.el` to `sellout-solarized-theme.el` (and making the necessary changes within). Or if you are not using your implementation anymore, then maybe it should be removed altogether? The implementation from https://github.com/bbatsov/solarized-emacs is vastly more popular.

tarsius:
I have removed @sellout's implementation because he never replied and because his implementation contains a library that conflicts with the @batsov implementation and that uses the name of the library as the package name also and therefore takes precedence.

@purcell I am wondering, why do you maintain your own implementation? What's wrong with @bbatsov's, which is more popular (though [not as much](http://www.melpa.org/#/?q=solar&sort=downloads&asc=false) as I would have expected.

I am also opening a pull-request at https://github.com/altercation/solarized to update https://ethanschoonover.com/solarized to link to @bbatsov implementation.

bbatsov:
> @purcell I am wondering, why do you maintain your own implementation? What's wrong with @bbatsov's, which is more popular (though not as much as I would have expected).

The stats are really interesting indeed! I'm curious what drove those significant download numbers for @purcell's version.

> I am also opening a pull-request at altercation/solarized#392 to update https://ethanschoonover.com/solarized to link to @bbatsov implementation.

Thanks!

purcell:
> @purcell I am wondering, why do you maintain your own implementation? What's wrong with @bbatsov's, which is more popular (though [not as much](http://www.melpa.org/#/?q=solar&sort=downloads&asc=false) as I would have expected).

This was due to shortcomings in the sellout version, IIRC. I contributed to that one for a while, but all the colours needed an overhaul so I made my own version. I did a lot of work around getting sRGB colours correct on MacOS, including patching the Emacs NS code.

> The stats are really interesting indeed! I'm curious what drove those significant download numbers for @purcell's version.

It's included in my config, which has 1700+ forks, so that's probably the reason.

bbatsov:
> I did a lot of work around getting sRGB colours correct on MacOS, including patching the Emacs NS code.

I remember. :-) https://emacsredux.com/blog/2014/01/11/a-peek-at-emacs-24-dot-4-srgb-colours-on-os-x/ 

> It's included in my config, which has 1700+ forks, so that's probably the reason.

Got it. Would you be interested in merging your version of Solarized with mine? I'd be happy to make you a co-maintainer of it.

tarsius:
Closing this. We have decided to not remove `color-theme` for now. I have done the work necessary to warn users not to use it.

> Got it. Would you be interested in merging your version of Solarized with mine? I'd be happy to make you a co-maintainer of it.

I would welcome it if the two of you did that. But that's not really the place to track this.

PR #5786, opened 2018-11-03
add recipe for veri-kompass pkg

AndreaCorallo:

Veri kompass parse a verilog design ad generate a hierarchy bar that let the user navigate the code. 
More over functions to follows wires (drivers and loads) into the design are provided.

https://gitlab.com/koral/veri-kompass

I'm the author/mantainer

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/veri-kompass
+++ b/recipes/veri-kompass
@@ -1,0 +1,1 @@
+(veri-kompass :repo "koral/veri-kompass" :fetcher gitlab)


purcell:
Thanks! Quite a few little issues here:

- In general this seems to be built around `helm`, so you should probably more properly call the package `helm-veri-kompass`. Alternatively, use the standard completion facilities (`completion-at-point-functions`, `completing-read`) so that this package works with any completion system, e.g. `helm`, `ivy`, `ido`.
- You're importing lots of libraries but don't have corresponding `Package-Requires` entries where necessary, e.g. for `helm`.
- `for GNU Emacs` in the package description is redundant and should be removed. `package-lint` probably mentioned this. There are other redundant mentions of `for Emacs` too. All elisp functions and libraries are for Emacs. :-)
- You haven't declared the `veri-kompass-mode` custom group you're mentioning: add a `defgroup`
- The `windmove` bindings [here](https://gitlab.com/koral/veri-kompass/blob/master/veri-kompass.el#L662-669) are out of place: these should go in a user's personal config, not here.
- The `auto-mode-alist` pattern is too broad -- it would match that string anywhere in a path.
- You should recommend that users call `(add-hook 'verilog-mode-hook 'veri-kompass-minor-mode)` themselves, not do it for them when the package is loaded.

Hope that helps!

AndreaCorallo:
Hi,
thanks for the review!
I pushed some commits to address the highlighted points.

Bests
  Andrea

purcell:
Looks great - merged! Welcome to MELPA. :tada:

P.S. I think that `helm` also overrides `completing-read`, so you might be able to do without the code which checks `(fboundp 'helm)`. If not, a better test might be to check if `completing-read-function` is `eq` to `helm-completing-read` (or whatever helm calls its completing read function).

AndreaCorallo:
And that is just amazing!! :) :)

Thanks for your job.

Bests
  Andrea

PS for what I tried helm does override completing-read if you like and it works well infact. 
But like I'm doing I can specify the name of the completion buffer that pops-up and the presented strings keeps their properties so integration is a little better.

PR #5787, opened 2018-11-04
Add zprint-mode

pesterhazy:

Integrates zprint, the pretty-printer for Clojure(Script) files, into Emacs

https://github.com/pesterhazy/zprint-mode.el

I am the package author.

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/zprint-mode
+++ b/recipes/zprint-mode
@@ -1,0 +1,1 @@
+(zprint-mode :repo "pesterhazy/zprint-mode.el" :fetcher github)


tarsius:
Remove `:files ("clojure-mode.el")` because (a) it specifies a file that does not exist and (b) if you don't specify the files then the default set of files will be used, which will work wonderfully here.

pesterhazy:
@tarsius you're right, I forgot to push my latest version.

Only the `.el` file is needed. The rest of the files are only used to update the included bash helper

tarsius:
> Only the `.el` file is needed. The rest of the files are only used to update the included bash helper

They won't be included by default. See `:files` at https://github.com/melpa/melpa/blob/master/README.md#recipe-format.

pesterhazy:
@tarsius makes sense, updated

purcell:
Looks good, thanks: merging now. You should also autoload the main `zprint` command, since people might prefer to use that ad-hoc rather than enable `zprint-mode`.

Welcome to MELPA! :tada:

purcell:
P.S. It's fine to ship a shell script with the package, and doing so would probably be a bit cleaner and simpler than inlining a long string with a build step.

pesterhazy:
@purcell both great points, fixed the first and will address the other later. Thanks for the friendly and educational welcome!

PR #5789, opened 2018-11-05
Add org-pivotal

qhuyduong:

org-pivotal is an Emacs minor mode to extend org-mode with Pivotal Tracker abilities

https://github.com/org-pivotal/org-pivotal

I am the author

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/org-pivotal
+++ b/recipes/org-pivotal
@@ -1,0 +1,3 @@
+(org-pivotal
+ :fetcher github
+ :repo "org-pivotal/org-pivotal")


purcell:
Thanks: this looks useful! Notes:

- You're missing package dependencies on `dash`, `a` etc.
- Don't hard-code the use of `ido-completing-read` etc. Just use `completing-read` so that the user's preferred completion mechanism will be used, e.g. helm, ivy, ido.
- You can remove "utility to" from the package description line -- it's unnecessary

Hope that helps!

qhuyduong:
Hi @purcell , sorry for taking so long to answer!
I have fixed all your comments. Please take a look again.
Thank you!

purcell:
Thanks - still a couple of issues:
- [This docstring](https://github.com/org-pivotal/org-pivotal/blob/d6a003624934ecb30b690535aa335e40dbe0a3ca/org-pivotal.el#L149) doesn't make much sense
- Don't [autoload changes to org-mode-hook](https://github.com/org-pivotal/org-pivotal/blob/d6a003624934ecb30b690535aa335e40dbe0a3ca/org-pivotal.el#L157-L158) -- this should be the user's choice, not the package maintainer's. Note that it should always be possible to have packages installed without them being enabled, so this breaks that rule. :-)

Looks great otherwise!

qhuyduong:
Thank you @purcell. I have just correct them. Please take a look again.

purcell:
Merged now, thanks for your patience, and sorry for the slow turnaround!

PR #5790, opened 2018-11-06
Add emacs-zmq

dzop:

Adds Emacs bindings to the ZeroMQ (http://zeromq.org/) message passing library.

https://github.com/dzop/emacs-zmq

Package mantainer

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/emacs-zmq
+++ b/recipes/emacs-zmq
@@ -1,0 +1,7 @@
+(emacs-zmq
+ :fetcher github
+ :repo "dzop/emacs-zmq"
+ :files (:defaults "Makefile" ("src"
+                               "src/*.c" "src/*.h"
+                               "src/Makefile.am"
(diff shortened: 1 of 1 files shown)


purcell:
Thanks. In the `:files` pattern in the recipe, just use `"src"` instead of `("src" ...)`. I think you want the entire directory.

Looks good otherwise!

purcell:
Great, merged now, thanks! Welcome to MELPA. :tada:

tarsius:
I have renamed the package to just `zmq`.

The name of an Emacs package usually shouldn't contain "emacs"
and its name must match the name of the library, here "zmq.el".

dzop:
@tarsius Thanks for the info.

@purcell Glad to be here!

purcell:
Sorry @tarsius, I missed that the two were unmatched.

PR #5792, opened 2018-11-06
Add git-time-metric

c301:

Plugin for the Emacs editor to be used with the [Git Time Metric](https://github.com/git-time-metric/gtm) platform

https://github.com/c301/gtm-emacs-plugin

I'm the author

**None needed**

### Checklist

### DIFF
--- a/recipes/git-time-metric
+++ b/recipes/git-time-metric
@@ -1,0 +1,2 @@
+(git-time-metric :repo "c301/gtm-emacs-plugin"
+			:fetcher github)


purcell:
Thanks! I think this Emacs package shouldn't have "emacs" in its name (or "plugin", for that matter), or am I misunderstanding something? Why not call it "git-time-metric.el" instead?

c301:
Good point! Name was changed

purcell:
Super, merged now. Thanks for your patience, and welcome to MELPA! :tada:

PR #5794, opened 2018-11-08
Add hasklig-mode

minad:
https://github.com/minad/hasklig-mode

Support for hasklig font ligatures using prettify-symbols

https://github.com/minad/hasklig-mode

Maintainer

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/hasklig-mode
+++ b/recipes/hasklig-mode
@@ -1,0 +1,3 @@
+(hasklig-mode
+ :repo "minad/hasklig-mode"
+ :fetcher github)


minad:
Thank you! I added the missing autoload.

minad:
@purcell I might add support for other fonts at some point. Is it ok for melpa to rename a package in that case or should I just create another package, e.g. ligatures-mode?

purcell:
> I might add support for other fonts at some point. Is it preferable for melpa to rename a package in that case or should I just create another package which supersedes hasklig-mode, e.g. ligatures-mode?

I guess either could work. If you do it soon before there are any users to speak of, then we can just rename the recipe.

minad:
Ok, thank you!

PR #5795, opened 2018-11-08
Add emidje

alan-ghelardi:
[Emidje](https://github.com/nubank/emidje) integrates
[Cider](https://github.com/clojure-emacs/cider) and
[Midje](https://github.com/marick/Midje) in a more pleasant way. It is a nREPL
based replacement for the old midje-mode.

https://github.com/nubank/emidje

I'm the package author and one of the maintainers.

**None needed**

Please confirm with `x`:

- [x] I've built and installed the package using the instructions in
- [CONTRIBUTING.org](https://github.com/melpa/melpa/blob/master/CONTRIBUTING.org)

### DIFF
--- a/recipes/emidje
+++ b/recipes/emidje
@@ -1,4 +1,3 @@
 (emidje :fetcher github
 :repo "nubank/emidje")
-(emidje :fetcher github
-:repo "nubank/emidje")
+
--- a/recipes/emidje
+++ b/recipes/emidje
@@ -1,0 +1,4 @@
+(emidje :fetcher github
+:repo "nubank/emidje")
+(emidje :fetcher github
+:repo "nubank/emidje")


purcell:
Thanks. _Please_ run `package-lint` as requested to save the MELPA maintainers' time. I can immediately see some issues it should report, e.g.

- You're missing a package dependency declaration for `seq`.
- string-suffix-p, thread-last and other functions are only available in newer Emacsen

Additionally:

- Don't define fresh faces with custom colours: instead, `:inherit` built-in faces like `success`, `error`, `warning` etc. This way, your faces will look okay with all existing themes by default.
- Don't autoload the middleware injection with `cider`: it should be possible for users to have this package installed without it being automatically enabled, just like with all the built-in Emacs packages. Instead, provide an autoloaded setup function, or provide users with instructions.
- Similarly, autoload the `emidje-mode` definition, and tell users to add it to their `clojure-mode` or `cider-repl-mode` as desired. You can provide a function to do that conveniently for them if you like.

Hope that helps!

alan-ghelardi:
@purcell, thank you very much for your comments; they were super helpful. I think that I've addressed all pointed issues, except that related to auto-loading the middleware injection. I've followed the same pattern adopted by [clj-refactor.el(https://github.com/clojure-emacs/clj-refactor.el/blob/master/clj-refactor.el#L4084). Since `emidje` is meant to be a `cider` plugin I thought that would be reasonable to use the same approach to inject the middleware in the `cider-jack-in`. Do you think that there are downsides in this strategy from the perspective of good practices in writing Emacs packages? Sorry for the question, but this is my first Emacs package.

purcell:
> I've followed the same pattern adopted by [clj-refactor.el](https://github.com/clojure-emacs/clj-refactor.el/blob/master/clj-refactor.el#L4084).

That is also inadvisable, IMO. (cc @bbatsov)

The general idea is that installation of a package doesn't imply wanting to always use a package. There are all sorts of clever `autoload` tricks that people try to use so that packages "just work" when they're installed, but it's bad practice because it subverts the above idea, and can lead to complex autoloads, unnecessarily slow start-up, and hard-to-debug issues. 

Better to provide an "enable" function that people can call from their init file or add to a hook.

bbatsov:
I agree with @purcell. Implicit injections can be very confusing for users if a package diverges from `cider` and becomes incompatible with it. 

I think the users should request the additional injected dep(s) explicitly.

bbatsov:
I've also noticed this link is broken https://github.com/nubank/midje-nrepl

alan-ghelardi:
@purcell,  thank you very much for clarifying. I think that now I've fixed all raised points. @bbatsov, thank you for the comments too. About the link, I was moving the repo in question from my personal account to the Nubank organization. Now it is there. By the way, I'd like to let you know about those projects once they were officially released, but as you joined the conversation, if you have some time to take a look at them and have comments, I would be more than glade to hear your opinion. We use Midje a lot at Nubank and as we became the official maintaners of it, I decided to write Emidje to improve the integration of Cider and Midje.

purcell:
Great, merged now - thanks!

Welcome to MELPA. :tada:

PR #5797, opened 2018-11-09
Add ox-wk.

w-vi:
`org-mode` publishing backend for exporting org files to Dokuwiki and Creole wiki formats.

https://github.com/w-vi/ox-wk.el

Author and maintainer.

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/ox-wk
+++ b/recipes/ox-wk
@@ -1,0 +1,1 @@
+(ox-wk :fetcher github :repo "w-vi/ox-wk.el")


purcell:
Thanks! Don't use the deprecated `cl` library -- use `cl-lib` instead, or maybe `seq`, depending on what you're using `cl` for. Looks okay otherwise, so I'll merge this now, and you can fix the `cl` issue separately. :-)

w-vi:
Thank you, good point on the `cl` I tend to forget about it, will address it.

PR #5798, opened 2018-11-09
Add alarm-clock

wlemuel:

An alarm clock management tools for Emacs.

https://github.com/wlemuel/alarm-clock

Author and maintainer

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/alarm-clock
+++ b/recipes/alarm-clock
@@ -1,0 +1,1 @@
+(alarm-clock :fetcher github :repo "wlemuel/alarm-clock")


purcell:
Thanks! Looks good, so merging now. Welcome to MELPA! :tada:

P.S. Consider using `when`/`unless` instead of single-branch `if`s. It improves readability, because people don't have to scan parentheses to see if there's an "else" form, and no `progn`s are necessary.

wlemuel:
All the `if` are replaced with `when`, thanks for your advice.

PR #5799, opened 2018-11-09
Update :files in vdm-mode recipe

peterwvj:
I'm currently working on extending vdm-mode and I'm adding additional Elisp files during that process. For this reason, `:files` needs to be updated.

----

Major mode for the Vienna Development Method 

https://github.com/peterwvj/vdm-mode

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/vdm-mode
+++ b/recipes/vdm-mode
@@ -1,5 +1,4 @@
 (vdm-mode
  :fetcher github
  :repo "peterwvj/vdm-mode"
- :files ("vdm-mode.el"
-         "vdm-mode-util.el"))
+ :files ("vdm-mode*.el"))


purcell:
> I'm currently working on extending vdm-mode and I'm adding additional Elisp files during that process. For this reason, `:files` needs to be updated.

The defaults already include `*.el` (and exclude test files) so you should generally just eliminate the `:files` spec altogether.

Other notes:
- Use `eval-after-load` to add the smartparens pair [here](https://github.com/peterwvj/vdm-mode/blob/d8e1ce912f5c771ef372f04afa9b2d5cd037ed3c/vdm-mode.el#L45-L46). The existing code doesn't allow for smartparens being loaded after vdm-mode.
- The dots in the file extension regexes [here](https://github.com/peterwvj/vdm-mode/blob/d8e1ce912f5c771ef372f04afa9b2d5cd037ed3c/vdm-mode.el#L49) should be escaped.
- `flycheck-vdm` should be a separate package, so please provide two recipes in this PR: they would both need `:files` specs in this case. `vdm-mode` should not load `flycheck-vdm` explicitly. You could imagine a `flymake` version of this functionality which users might prefer, especially since the latest flymake has essentially gained a lot of the flycheck functionality.
- Similarly `vdm-snippets` should be a separate package, because otherwise there will be byte compilation errors/warnings upon installation for users who don't have `yasnippet`.

Hope that helps - this all looks very polished!

purcell:
Hah, I just re-read the PR, and realised that this will have the effect of excluding the flycheck and snippets files from vdm-mode, which is ideal. Hopefully the other feedback above will still be useful to you.

peterwvj:
Yes, feedback like this is always very useful and highly appreciated -- many thanks!

Regarding the other issues:

- `vdm-mode` now uses `eval-after-load` to remove the smartparens pair like you suggested
- The actual regex is constructed in the [following line](https://github.com/peterwvj/vdm-mode/blob/master/vdm-mode.el#L50) so the dots are accounted for.

purcell:
>     * `vdm-mode` now uses `eval-after-load` to remove the smartparens pair like you suggested

Nice.

>     * The actual regex is constructed in the [following line](https://github.com/peterwvj/vdm-mode/blob/master/vdm-mode.el#L50) so the dots are accounted for.

Ah yes - forgot that `regexp-opt` quotes the strings. :-)

PR #5800, opened 2018-11-09
Add recipe for path-helper

arouanet:

Set PATH environment variables from configuration files.

https://github.com/arouanet/path-helper

Author and maintainer.

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/path-helper
+++ b/recipes/path-helper
@@ -1,0 +1,1 @@
+(path-helper :fetcher github :repo "arouanet/path-helper")


purcell:
Thanks, this looks useful.

purcell:
Merged now - welcome to MELPA! :tada:

arouanet:
Thanks Steve. I was a bit worried that it would step into the territory of your very useful exec-path-from-shell, but I think there is room for both.

purcell:
Yep, I think so. Your package made me wonder whether I should break my own shell env vars out into files in a `.d` directory, but that wouldn't work for me in practice because of conditional and/or scripted settings like the following:

```
nix_init=~/.nix-profile/etc/profile.d/nix.sh
if [[ -f "$nix_init" ]]; then source "$nix_init"; fi
```

terlar:
Personally I have a quite simple solution that I threw together some time ago, it it works quite well.

Each environment variable is stored as a file named the same as the environment variable I want to set, if it is just text, the value will be stored in the variable, if it is an executable script it will set the value generated by that script. Environment variables can be structured by directories and loaded recursively. So for example I put symlinks in a directory called `auto` and then I can load all those variables on demand on shell launch etc.
```
$ tree -al $EVPATH 
/Users/user/.config/env
├── auto
│   ├── gnupg -> ../gnupg
│   ├── go -> ../go
│   ├── java -> ../java
│   ├── node -> ../node
│   ├── ruby -> ../ruby
│   └── tools -> ../tools
├── PATH
├── gnupg -> ../../src/github.com/terlar/dotfiles/gnupg/.config/env/gnupg
│   ├── GPG_TTY
│   └── SSH_AUTH_SOCK
├── go -> ../../src/github.com/terlar/dotfiles/env/.config/env/go
│   └── GOPATH
├── java -> ../../src/github.com/terlar/dotfiles/java/.config/env/java
│   ├── CLASSPATH
│   └── JAVA_HOME
├── my_company
│   └── TEAM
├── node
│   ├── NODE_VERSIONS
│   └── NPM_CONFIG_PREFIX
├── ruby
│   └── GEM_HOME
└── tools
    ├── BROWSER
    ├── EDITOR
    ├── MANPAGER
    ├── PAGER
    ├── TERMINAL
    └── VISUAL -> EDITOR
```

```
$ ev.sh auto
GPG_TTY=/dev/ttys000
SSH_AUTH_SOCK=/Users/user/.gnupg/S.gpg-agent.ssh
GOPATH=/Users/user/go
CLASSPATH=.:/Users/user/.java/lib/eclipse.jdt.ls/plugins/org.eclipse.equinox.launcher_1.5.200.v20180922-1751.jar
JAVA_HOME=/Library/Java/JavaVirtualMachines/openjdk-11.0.1.jdk/Contents/Home
NODE_VERSIONS=/Users/user/.nodes
NPM_CONFIG_PREFIX=/Users/user/.npm-global
GEM_HOME=/Users/user/.gem/ruby/2.3.0
BROWSER=browser
EDITOR=emacsclient
MANPAGER=most
PAGER=less
TERMINAL=kitty
VISUAL=emacsclient
```

```
$ ev.sh my_company
TEAM=giraffes
$ ev.sh tools/EDITOR
EDITOR=emacsclient
```

```
$ cat $EVPATH/tools/EDITOR
emacsclient
$ cat $EVPATH/java/JAVA_HOME 
#!/bin/sh
/usr/libexec/java_home
```

env var loader/executer:
https://github.com/terlar/dotfiles/blob/3ae2dce14e0f8c4e7d3312b07b98644f6e2a530c/env/.local/bin/ev.sh
macos env sync to ensure GUI apps get these variables as well:
https://github.com/terlar/dotfiles/blob/3ae2dce14e0f8c4e7d3312b07b98644f6e2a530c/env/.local/bin/macos-sync-env

Hope someone finds it useful, sorry for the semi-non-related spam.

PR #5801, opened 2018-11-10
Add recipe for tmux-pane

laishulu:

This package provide integration between emacs window and tmux pane. 

https://github.com/laishulu/emacs-tmux-pane

maintainer

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/tmux-pane
+++ b/recipes/tmux-pane
@@ -1,0 +1,3 @@
+(tmux-pane
+  :fetcher github
+  :repo "laishulu/emacs-tmux-pane")


purcell:
Thanks. Notes:

- This isn't remotely a well-formatted package. Make a new empty `tmux-pane.el` and run `M-x auto-insert`. Then move the code into that.
- Declare a package dependency on `s`.
- Don't provide `defcustom`s for every key and then apply them with `global-set-key` -- this is just a clumsy way to do something standard in Emacs: a keymap! What you should actually do is declare a global minor mode, then set up its default keybindings. Then there's a standard way for users to customise it if they way.

laishulu:
Thanks for the tips. Just need some further instructions on this:

>Declare a package dependency on s and names.

Do you mean declare them on README?

laishulu:
@purcell 
Code updated following your instructions.
please check.

PR #5802, opened 2018-11-10
Allow company-solidity.el in solidity-mode

LefterisJP:
Company-solidity is now fully
[integrated](https://github.com/ssmolkin1/company-solidity#now-fully-integrated-in-solidity-mode-no-longer-independently-maintained)
into solidity-mode. As such the `exclude` rules breaks the melpa package and
should be removed.


LefterisJP:
@purcell Thank you for the feedback. 

Can I have 3 separate packages from the same repository using the exclude rules? I would prefer to not have to make 3 different github repositories.

a13:
>Can I have 3 separate packages from the same repository using the exclude rules?

Yes, take a look at use-package and use-package-ensure-system-packages

LefterisJP:
@purcell @a13 is this more like what you wanted? Also made sure that company-solidity and solidity-flycheck are not autoloaded: https://github.com/ethereum/emacs-solidity/commit/d6c48a1cb64d3c8a825dc0d06c839f2cacd4d289

LefterisJP:
@purcell ping?

PR #5806, opened 2018-11-12
Add multi-magit

luismbo:
I still have issues to deal with in the checklist. Can you help me with the last one? Are the other ones showstoppers?

Multi-magit is a set extensions to Magit for handling multiple git repositories simultaneously.

https://github.com/luismbo/multi-magit

Author.

**None needed**

### Checklist

- [ ] I've used the latest version of [package-lint](https://github.com/purcell/package-lint) to check for packaging issues, and addressed its feedback. (I've addressed most of it, but can't get rid of `692:1: error: "magit-multi-magit-repo-section-map" doesn't start with package's prefix "multi-magit".` because magit expects this variable to start with `magit-`.)
- [ ] My elisp byte-compiles cleanly. (Not yet: `multi-magit.el:252:14:Warning: ‘magit-branch-delete’ is for interactive use only.` don't know how to fix this yet.)
- [ ] `M-x checkdoc` is happy with my docstrings. (Not yet.)
- [ ] I've built and installed the package using the instructions in [CONTRIBUTING.org](https://github.com/melpa/melpa/blob/master/CONTRIBUTING.org) (Building the recipe complains about magit-2.11 not being available. What should I depend on instead?)

### DIFF
--- a/recipes/multi-magit
+++ b/recipes/multi-magit
@@ -1,0 +1,1 @@
+(multi-magit :repo "luismbo/multi-magit" :fetcher github)


tarsius:
I've added a new class slot called `keymap`. The value can be any symbol whose value is a keymap.

I'll release `2.91.0` today (for unrelated reasons) and you should depend on that. I don't know why `magit-2.11` isn't satisfied for you. Seems weird, but may this also fixes that.

`magit-branch-delete` is for interactive use only because most code that might use it probably should just use something like `(magit-call-git "branch" "-D" branch)`. But in this case it is probably okay to use this function; you really want some of the additional behavior that it implements. It might be possible to split up this function and you could then call one of those directly. Or not, I won't rush something here. For now you either have to live with the warning or temporarily silence by changing the value of the symbol property inside an `unwind-protect`.

You should fix

* First line is not a complete sentence
* Lisp symbol `top-level' should appear in quotes
* Argument `repo' should appear (as REPO) in the doc string
* ... and alike ...

but I would ignore

* All variables and subroutines might as well have a documentation string
* (false-positive) Probably "changes" should be imperative "change"

tarsius:
I haven't looked at the implementation much, but from the sound of it this implements a feature that has been requested a few times in the past, so it seems worthwhile doing it.

luismbo:
@tarsius thanks for your comments. Regarding `magit-branch-delete`, I wonder If I can get away with `call-interactively`?

For the record we've discussed multi-magit in the past here: https://github.com/magit/magit/issues/3295 I do have a couple of implementation issues I should run by you some time, and I'll probably use that issue for that purpose.

tarsius:
> Regarding `magit-branch-delete`, I wonder If I can get away with `call-interactively`?

That won't work because then you cannot pass the branch to be deleted.

> I do have a couple of implementation issues I should run by you some time, and I'll probably use that issue for that purpose.

Go ahead.

tarsius:
@luismbo I am closing this now, but let us know when you have made progress.

PR #5808, opened 2018-11-13
Add recipe for hl-fill-column

laishulu:

This package provide modes to highlight fill column.

https://github.com/laishulu/hl-fill-column

the maintainer

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/hl-fill-column
+++ b/recipes/hl-fill-column
@@ -1,0 +1,3 @@
+(hl-fill-column
+  :fetcher github
+  :repo "laishulu/hl-fill-column")


laishulu:
travis-ci failed, but not my problem:

```
The command "sudo apt-get -yq update" failed and exited with 100 during .
```

laishulu:
@purcell 

>"hl-fill-column" would be better?

agree. Done.

>but in my experience none is super reliable

Yes, I tried to use `fci`, but its laggy and comaptibility is not good enough.
So I decided to roll out my own finally.

>consider also adding an ;;;###autoload for the global minor mode.
Done.

purcell:
Great, merged now - thanks!

PR #5813, opened 2018-11-16
Add 'evil-nl-break-undo' recipe to MELPA.

VanLaser:

evil-nl-break-undo is a (very simple) evil-mode add-on that breaks the undo sequence in the insert state, when the buffer is changed over a line boundary, i.e. when at least one line is created or deleted. It does not depend on how the key RET is mapped. The 'boundary' is actually customizable using a regex. It is a (better) equivalent of the following Vim mapping:

```vimscript
inoremap <silent> <cr> <c-g>u<cr>
```

Note: it assumes that `evil-mode` is enabled, and that the variable `evil-want-fine-undo` is `nil` (this is the default in `evil-mode`, but not with Spacemacs).

https://github.com/VanLaser/evil-nl-break-undo

I am the author/maintainer of the package.

None needed

### Checklist

Please confirm with `x`:

- [x ] I've read [CONTRIBUTING.org](https://github.com/melpa/melpa/blob/master/CONTRIBUTING.org)

### DIFF
--- a/recipes/evil-nl-break-undo
+++ b/recipes/evil-nl-break-undo
@@ -1,0 +1,3 @@
+(evil-nl-break-undo
+  :fetcher github
+  :repo "VanLaser/evil-nl-break-undo")


purcell:
Thanks - this looks okay, except that there is no package description in the first line after the `---`. Please can you add that (keep it short please), then we're good to merge. :-)

VanLaser:
Hi, done, hope that's ok (and thanks!)

purcell:
Great, merged now - thanks for your patience.

PR #5816, opened 2018-11-18
Create mint-mode

creatorrr:

Major mode for [mint lang](https://mint-lang.com). Provides:
 - Syntax highlighting
 - Auto format on save using `mint format`

https://github.com/creatorrr/emacs-mint-mode

Author and maintainer

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/mint-mode
+++ b/recipes/mint-mode
@@ -1,0 +1,5 @@
+(mint-mode
+ :fetcher github
+ :url "https://github.com/creatorrr/emacs-mint-mode"
+ :repo "creatorrr/emacs-mint-mode"
+ :files ("mint-mode.el"))
--- a/recipes/mint-mode
+++ b/recipes/mint-mode
@@ -1,6 +1,5 @@
 (mint-mode
  :fetcher github
- :url "https://github.com/creatorrr/emacs-mint-mode"
  :repo "creatorrr/emacs-mint-mode"
  :version-regexp "0\\.4\\..+"
  :files ("mint-mode.el" "tokens"))
--- a/recipes/mint-mode
+++ b/recipes/mint-mode
@@ -2,4 +2,5 @@
  :fetcher github
  :url "https://github.com/creatorrr/emacs-mint-mode"
  :repo "creatorrr/emacs-mint-mode"
- :files ("mint-mode.el"))
+ :version-regexp "0\\.4\\..+"
+ :files ("mint-mode.el" "tokens"))
--- a/recipes/mint-mode
+++ b/recipes/mint-mode
@@ -1,5 +1,5 @@
 (mint-mode
  :fetcher github
  :repo "creatorrr/emacs-mint-mode"
- :version-regexp "0\\.4\\..+"
+ :version-regexp "0\\.5\\..+"
  :files ("mint-mode.el" "tokens"))


purcell:
Thanks! Quick notes:
- The `:version-regexp` in the recipe is unnecessary and should be removed.
- The `:files` spec should ideally read `(:defaults "tokens")`
- Why are the libraries `require`-d at compilation time when they are clearly used at runtime? Those should not be wrapped in `eval-when-compile`, which has probably been cargo-culted from elsewhere. :-)
- Instead of `mapconcat` + `regexp-quote` and `\\|`, use the `regexp-opt` function.
- Don't use your own major mode hook like [this](https://github.com/creatorrr/emacs-mint-mode/blob/2e2568bd58dcf31b520c8a77ccf518c43e8aeb16/mint-mode.el#L270). Instead, provide a `defcustom` called something like "mint-format-on-save", and then add `mint-format-on-save-maybe` to `before-save-hook`. Note that I say `before-save-hook`, while you're currently using `after-save-hook`: `mint-format-file` wrongly assumes that every buffer will have an associated file, which is not true. `mint-format-file` should instead either copy the current buffer's contents to a temporary file, or you can use `call-process-region`/`shell-command-on-region` if `mint format` can read from STDIN.

Hope that helps!

DamienCassou:

lassik:
@creatorrr @riscy Can you re-open this PR? https://github.com/creatorrr/emacs-mint-mode/pull/7 ought to have fixed most of the issues.

riscy:
Reopened and just doing a very quick pass right now.  The previous review from @purcell looks like some of it still applies (e.g. the use of `:version-regexp` in the recipe, which we can just edit here.)

`mint-mode.el` with byte-compile using Emacs 28.1:
```
mint-mode.el:257:12: Warning: the function `ansi-color-apply-on-region' is not
    known to be defined.
mint-mode.el:214:16: Warning: the function `company-doc-buffer' is not known
    to be defined.
```

`mint-mode.el` with package-lint 20220607.738:
```
1 issue found:
217:0: error: "format-all--locate-file" doesn't start with package's prefix "mint".
```

`mint-mode.el` with [melpazoid](https://github.com/riscy/melpazoid):
```
- mint-mode.el#L55: It's safer to sharp-quote function names; use `#'`
- mint-mode.el#L280: It's safer to sharp-quote function names; use `#'`
```

Package and license:
- Prefer the default recipe or `:defaults`, if possible.

riscy:
I'll mark this awaiting-upstream for now, but feel free to reach out.

jgarte:
Hi, I am helping maintain `emacs-mint-mode` for @creatorrr .

I just released version `2.0.0` of `emacs-mint-mode`.

Should I open a new PR for adding `emacs-mint-mode` to `melpa`?

@purcell I can fix the issues that you and others have requested.

jgarte:
>     * Don't use your own major mode hook like [this](https://github.com/creatorrr/emacs-mint-mode/blob/2e2568bd58dcf31b520c8a77ccf518c43e8aeb16/mint-mode.el#L270). Instead, provide a `defcustom` called something like "mint-format-on-save", and then add `mint-format-on-save-maybe` to `before-save-hook`. Note that I say `before-save-hook`, while you're currently using `after-save-hook`: `mint-format-file` wrongly assumes that every buffer will have an associated file, which is not true. `mint-format-file` should instead either copy the current buffer's contents to a temporary file, or you can use `call-process-region`/`shell-command-on-region` if `mint format` can read from STDIN.

format on save has been completely remove in `2.0.0` of `emacs-mint-mode` so this suggestion is now resolved.

jgarte:
>     * Why are the libraries `require`-d at compilation time when they are clearly used at runtime? Those should not be wrapped in `eval-when-compile`, which has probably been cargo-culted from elsewhere. :-)
> 
>     * Instead of `mapconcat` + `regexp-quote` and `\\|`, use the `regexp-opt` function.

The eval-when-compile suggestion should also be resolved now.

jgarte:
Hi, I'm just following up to know if I can open a new PR to fix this one or if there is a way for me to take it over. I'd like to resolve the outstanding issues for including `emacs-mint-mode` in melpa.

jgarte:
After reading [some stackoverflow](https://stackoverflow.com/questions/50781553/how-to-take-ownership-of-a-pull-request) it sounds like the way to go would be for me to open a new PR since this PR is abandoned by the original author who has ownership over it. I'll open a new PR soon and link to this one.

jgarte:
Hi, This PR can be closed. I opened a new one #8257

riscy:
Closing in favor of #8257.

PR #5820, opened 2018-11-20
Add recipe for named-timer.el

DarwinAwardWinner:

`named-timer.el` provides `named-timer-run` and other related functions. `named-timer-run` works like `run-with-timer` but takes an additional argument giving the timer a name. Crucially, creating a named timer automatically cancels any existing timer with that name. This simplifies the pattern of cancelling and restarting a timer to a single line, and ensures that no coding mistake can accidentally leave duplicate timers running.

https://github.com/DarwinAwardWinner/emacs-named-timer

Author

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/named-timer
+++ b/recipes/named-timer
@@ -1,0 +1,1 @@
+(named-timer :fetcher github :repo "DarwinAwardWinner/emacs-named-timer")


DarwinAwardWinner:
Note that package-lint and checkdoc both complain about the advice function for `cancel-timer`, but I consider both spurious, as advice functions have somewhat different standards. (e.g. the function name is the same one that would be used by `define-advice`, but I'm avoiding `define-advice` to maintain compatibility with older versions of Emacs)

alphapapa:
Yeah, I'd recommend against using advice unless absolutely necessary.

purcell:
Nice, thanks!~

PR #5821, opened 2018-11-21
create hsluv recipe

woozong:

This package provides an Elisp implementation of the HSLUV colorspace conversions as created by Alexei Boronine (@boronine), and documented on http://www.hsluv.org/
HSLuv is a human-friendly alternative to HSL.
CIELUV is a color space designed for perceptual uniformity based on human experiments. When accessed by polar coordinates, it becomes functionally similar to HSL with a single problem: its chroma component doesn't fit into a specific range.
HSLuv extends CIELUV with a new saturation component that allows you to span all the available chroma as a neat percentage.

https://github.com/hsluv/hsluv-emacs

Maintainer of the Emacs implementation of HSLuv
Contributor to the hsluv org on github. 

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/hsluv
+++ b/recipes/hsluv
@@ -1,0 +1,3 @@
+(hsluv
+ :fetcher github
+ :repo "hsluv/hsluv-emacs")


purcell:
Thanks, looks good, but you'll need to declare a package dependency on the version of `seq` that's in GNU ELPA. :-)

woozong:
@purcell I assume
`  ;; Package-Requires: ((seq "2.20"))`
would be sufficient? (my first package here ;-))

woozong:
package dependency added to [hsluv.el](https://github.com/hsluv/hsluv-emacs/blob/master/hsluv.el)

purcell:
Great, thanks - merged! Welcome to MELPA. :tada:

woozong:
Oh woohoo :-)

PR #5822, opened 2018-11-22
Warn users when falling back to insecure http

tarsius:
This adds the warning message suggested by Steve
in #5685, which in turn was in response to #5682.


purcell:
Super

PR #5823, opened 2018-11-22
gnus-recent

unhammer:

Maintains a list of recently read Gnus articles, to easily go
back through previously read articles, or optionally pick from a list
shown with ivy.

https://github.com/unhammer/gnus-recent

Author.

 **None needed**

### Checklist

Please confirm with `x`:

I'd prefer ivy to be an optional dependency, but this means I get warnings about missing ivy functions from byte compilation. I don't know if there's a way to avoid that without making ivy a hard requirement.
EDIT: declare-function'd 

The `make recipes/gnus-recent` goes fine, but the `$ EMACS_COMMAND=/usr/bin/emacs make sandbox INSTALL=gnus-recent` complains about emacs 25.3.2 not being available(?).

### DIFF
--- a/recipes/gnus-recent
+++ b/recipes/gnus-recent
@@ -1,2 +1,3 @@
 (gnus-recent :fetcher github
-             :repo "unhammer/gnus-recent")
+             :repo "unhammer/gnus-recent"
+             :files ("gnus-recent.el"))
--- a/recipes/gnus-recent
+++ b/recipes/gnus-recent
@@ -1,0 +1,2 @@
+(gnus-recent :fetcher github
+             :repo "unhammer/gnus-recent")


purcell:
> I'd prefer ivy to be an optional dependency, but this means I get warnings about missing ivy functions from byte compilation. I don't know if there's a way to avoid that without making ivy a hard requirement.

The right way to do that would be separate little package. It's so tiny in this case that it would be okay to just add a `declare-function` clause.

In truth, the ivy code is completely out of place here, and could as easily live in personal configuration. `eval-after-load` is not meant for use in packages, just for personal config, because it causes code to be evaluated in an unpredictable order. So my suggestion would be to rip out those few lines, resulting in a cleaner package.

purcell:
Thanks. I tweaked the recipe to include only `gnus-recent.el`, then, and now it's merged. If you'd like to add a further one for `gnus-recent-ivy`, please send a little PR, and also add the missing entry for `ivy` in the `Package-Requires` header of `gnus-recent-ivy.el`.

PR #5824, opened 2018-11-22
Add exwm-firefox-core

walseb:

This package defines functions that execute hotkeys relevant to firefox so that you can bind your keys inside emacs instead of relying on (unreliable) firefox keybind packages. This is a library so it's not functional without user configuration or the use of an implementation of it like exwm-firefox-evil which i will be submitting here too.

https://github.com/your/awesome_package

I'm the author

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/exwm-firefox-core
+++ b/recipes/exwm-firefox-core
@@ -1,0 +1,1 @@
+(exwm-firefox-core :fetcher github :repo "walseb/exwm-firefox-core")


purcell:
Thanks. 2 points of feedback:

- `exwm-input--fake-key` is apparently an internal function, based on its name, so perhaps you're misusing that package?
- That's a lot of autoloads, all of which take time to evaluate at Emacs startup time. Perhaps cut them down, or suggest that users explicitly load this library with `(eval-after-load 'exwm)` or similar?

walseb:
> `exwm-input--fake-key` is apparently an internal function, based on its name, so perhaps you're misusing that package?


It's only used by `exwm-input-send-next-key` and `exwm-input-send-simulation-key` which only has the parameter `TIMES` (since the user is supposed to type in the keys after running it), so there isn't really any other way to do what `exwm-input--fake-key` does

> That's a lot of autoloads, all of which take time to evaluate at Emacs startup time. Perhaps cut them down


I though about this but I'm not sure if I can without removing functionality. The only solution I can think of is using a giant pcase (something like this `(pcase command ('up-key (up-key-func))...)`). Do you think that would be better?

> or suggest that users explicitly load this library with `(eval-after-load 'exwm)` or similar?


Fixed

PR #5825, opened 2018-11-22
Add exwm-firefox-evil

walseb:

exwm-firefox-evil makes firefox's control scheme modal. It uses vi-like keys and doesn't require any addons. It does this by implementing [exwm-firefox-core](https://github.com/walseb/exwm-firefox-core) which sends fake keys to firefox using exwm.

https://github.com/walseb/exwm-firefox-evil

I'm the author

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/exwm-firefox-evil
+++ b/recipes/exwm-firefox-evil
@@ -1,0 +1,1 @@
+(exwm-firefox-evil :fetcher github :repo "walseb/exwm-firefox-evil")


purcell:
Thanks. Quick notes:

- You need to declare a package dependency on `evil`.
- Is `exwm-input-line-mode-passthrough` buffer-local? If not, it would be strange to change the `exwm` behaviour globally while modifying the evil state buffer-locally.
- The `exwm-firefox-evil-mode-enable` and `exwm-firefox-evil-mode-disable` functions are unidiomatic and should be unnecessary: the code in those functions should be in the definition of `exwm-firefox-evil-mode`, so that that mode can simply be toggled on and off.

Hope that helps!

walseb:
Everything should be fixed now!

purcell:
Great, merged now - thanks for your patience.

PR #5826, opened 2018-11-22
add recipe for slimhtml

balddotcat:

slimhtml is an Emacs org mode export backend.  It is a set of transcoders for common org elements which outputs minimal HTML.  The aim is not to re-invent the wheel over the default org-mode HTML exporter - as it tackles a much bigger, and different problem - but to provide a small set of components for easier customization of HTML output from org.

https://github.com/balddotcat/slimhtml

author

flycheck emits warnings for 'unused lexical arguments' that are from the original org-export function signatures, and are required for functionality

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/slimhtml
+++ b/recipes/slimhtml
@@ -1,0 +1,1 @@
+(slimhtml :fetcher github :repo "balddotcat/slimhtml")


balddotcat:
Thank you for the feedback @alphapapa!
Please see the updated the [README](https://github.com/balddotcat/slimhtml) file, I hope it clarifies my reasoning; I will also update the namespace shortly.

PR #5827, opened 2018-11-22
New recipe: detour

ska2342:

Detour provides two small functions which allow storing the current position in a register, going somewhere else, and then quickly jumping between those positions. The benefit over just using registers for this is speed when bound to easily reachable keys. 

https://github.com/ska2342/detour

Sole author and maintainer.

### Checklist

Please confirm with `x`:

- [x] I've used the version of [package-lint](https://github.com/purcell/package-lint) shipped with Emacs 25.2.2 to check for packaging issues, and addressed its feedback. (Not sure if this is the _latest_ version, though.)

### DIFF
--- a/recipes/detour
+++ b/recipes/detour
@@ -1,0 +1,2 @@
+(detour	:repo "ska2342/detour"
+        :fetcher github)


purcell:
>     * I've used the version of [package-lint](https://github.com/purcell/package-lint) shipped with Emacs 25.2.2 to check for packaging issues, and addressed its feedback. (Not sure if this is the _latest_ version, though.)

Emacs doesn't ship with `package-lint`. :-/

In any case, this looks good, so I'll go ahead and merge it. It's almost too trivial to be a package, but I can see how it would be handy.

PR #5828, opened 2018-11-23
pretty-print archive-contents to make it human readable.

emacs18:
Prior to this change this file is one very long line with more than
a million bytes.

This step takes less than 3 seconds on virtual linux running on my one year old Dell laptop
to process the full 1.3 megabytes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -76,6 +76,7 @@ packages: $(RCPDIR)/*
 packages/archive-contents: .FORCE
 	@echo " • Updating $@ ..."
 	@$(EVAL) '(package-build-dump-archive-contents)'
+	@$(EVAL) '(progn (find-file "packages/archive-contents") (emacs-lisp-mode) (pp-buffer) (save-buffer))'
 
 cleanup:
 	@$(EVAL) '(package-build-cleanup)'


purcell:
It would be better to make `package-build-dump-archive-contents` do this, IMO.

emacs18:
That may well be.  However I have two reasons why I did not.  First I can't test it, since I don't build only few hundred packages, not thousands.  Second it may actually be much slower.  As you may recall I suggested this idea few years ago.  At the time I did modify package-build.el.  However it was much much slower; something like 6 or 7 seconds to process 300  or so packages.  I can only guess as to the root cause of the slow speed.  Perhaps there are other features enabled when package-build.el is in play.   In any case my lack of ability to test is key reason why I did not mess with package-build.
Besides the change in this PR is simple (just one makefile line), fast, and does not modify existing code.   That I consider a positive.

emacs18:
Actually package-build-dump-archive-contents already has pretty-print optional argument!
If I recall correctly, I contributed this code, but you decided not to use it based on my reported 
speed information, i.e., it was too slow.

purcell:
> Actually package-build-dump-archive-contents already has pretty-print optional argument!

Ah yes. The key issue is that the archive contents still need to be rebuilt every time a package is built. To make this really fast for MELPA would be tricky, and it might even be worth apply pseudo-pretty-printing instead, e.g. just printing each package entry on a separate line, but not fully pretty-printing everything.

emacs18:
Thanks.  One line per package is certainly a big improvement from one line for all packages.
However I still like pretty printed version, because it is so much more readable.
I can read the unified diff of archive-contents changes to extract useful information quickly.
I'll keep the change in this PR on my own branch and use it for a while.  Perhaps one day
we can find a way to make pretty printing fast enough.  By the way, what would be fast enough?
Would one second be fast enough?

purcell:
>  By the way, what would be fast enough?
> Would one second be fast enough?

Good question. Yes, probably a second for the full MELPA set would be okay.

PR #5835, opened 2018-11-26
Add recipe for moccur-edit

tigermonkey:

Add recipe for moccur-edit

https://github.com/myuhe/moccur-edit.el

I am a user of moccur-edit

None needed

### Checklist

Please confirm with `x`:

    I haven't fixed those, but I can start
```
8 issues found:

1:1: error: Package should have a Homepage or URL header.
7:1: warning: You should include standard keywords: see the variable `finder-known-keywords'.
139:1: error: "moccur-query-when-buffer-read-only" doesn't start with package's prefix "moccur-edit".
152:1: error: "moccur-mode-edit-set-key" doesn't start with package's prefix "moccur-edit".
171:1: error: "moccur-mode-change-face" doesn't start with package's prefix "moccur-edit".
356:1: error: "current-line" doesn't start with package's prefix "moccur-edit".
360:1: error: "max-line" doesn't start with package's prefix "moccur-edit".
457:33: warning: This key sequence is reserved (see Key Binding Conventions in the Emacs Lisp manual)
```

### DIFF
--- a/recipes/moccur-edit
+++ b/recipes/moccur-edit
@@ -1,0 +1,1 @@
+(moccur-edit :repo "myuhe/moccur-edit.el" :fetcher github)


tigermonkey:
> and it’s unclear to a potential user why he would choose this instead of the many other packages that do multi-file search-and-replace

I wasn't aware that you could edit a normal occur buffer. Nevermind then, I will use normal occur.

PR #5839, opened 2018-11-29
Added reciple for format-table.

functionreturnfunction:

Library which reformats tabular output from many RBDMS systems' command-line tools, with additional support for org tables and json.

https://github.com/functionreturnfunction/format-table

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/format-table
+++ b/recipes/format-table
@@ -1,0 +1,1 @@
+(format-table :fetcher github :repo "functionreturnfunction/format-table")


functionreturnfunction:
@riscy Thanks for the feedback!  There's another component to this that hooks into SQLi/comint to reformat things on the fly; it's still a WIP and if there's any ideas I can ruthlessly steal from `shx' I assure you I will ;-)

I've addressed the issues you listed, except I'm not a fan of using `and' or `or' to do conditional execution of child expressions unless the result is actually being used.  The following are all functionally equivalent, I just believe the intent is more explicit in the first for cases like this:

```elisp
(unless (or nil nil) t)     ;; => t
(and (not nil) (not nil) t) ;; => t
(or nil nil t)              ;; => t
```

functionreturnfunction:
Also I had to walk back the change from `` `format-table-trim-row'`` to `` `string-trim'``, as the optional TRIM-LEFT and TRIM-RIGHT arguments weren't added until Emacs 26.  In future I'll be sure to monitor my feature branches in travis before I merge them out...

PR #5840, opened 2018-11-30
Add ibuffer-project

muffinmad:

Group ibuffer's list by project or default directory

https://github.com/muffinmad/emacs-ibuffer-project

Author

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/ibuffer-project
+++ b/recipes/ibuffer-project
@@ -1,0 +1,1 @@
+(ibuffer-project :fetcher github :repo "muffinmad/emacs-ibuffer-project")


muffinmad:
@riscy Thanks for feedback! I fixed those issues but leave `defvar` near column definition.

purcell:
Thanks, will review this in due course. Looks like this is largely an adaptation of my `ibuffer-projectile.el`. I think our packages will produce some of the same filter/column names, e.g. `ibuffer-make-column-project-relative-file` -- I should probably be the one to rename the columns/filters in `ibuffer-projectile`, even though it risks breaking things for some users.

purcell:
All those choices make sense. Code & recipe look good, so merging now. :-)

PR #5841, opened 2018-12-01
Add recipe for auto-rename-tag

jcs090218:

Automatically rename paired HTML/XML tag. 

https://github.com/jcs090218/auto-rename-tag

I am the maintainer.

None needed.

### Checklist

Please confirm with `x`:


purcell:
Thanks, looks good. It's not obvious from the package name that "tag" means HTML tag, but no big deal, so I'll merge now. You should add an `;;;###autoload` cookie for `auto-rename-tag-mode`, at least, and maybe the globalized version too.

PR #5842, opened 2018-12-02
Add recipe for abs-mode, a major mode for the modeling language Abs

rudi:

A major mode for the modeling language Abs (http://docs.abs-models.org/)

https://github.com/abstools/abs-mode/

Author, maintainer

**None needed**

### Checklist

Please confirm with `x`:

(note: package-lint complains about unversioned dependencies, but these are legal and I don't depend on specific versions for `erlang' and `maude-mode')
(note: two deprecation warnings wrt flymake on emacs 26.1, but I want to support emacs 25 if possible)

### DIFF
--- a/recipes/abs-mode
+++ b/recipes/abs-mode
@@ -1,0 +1,3 @@
+(abs-mode :repo "abstools/abs-mode"
+          :fetcher github
+          :files ("abs-mode.el"))


purcell:
Thanks. Quick feedback:
- Please use versioned dependencies [here](https://github.com/abstools/abs-mode/blob/336719ded3b5320abce3057b8bf932247a636e8a/abs-mode.el#L8), even if you have to use `"0"`. The bare package name format is not understood by older versions of `package.el`, so is to be avoided. `package-lint` should have mentioned this.
- Don't initialize executable paths using `executable-find` like [this](https://github.com/abstools/abs-mode/blob/336719ded3b5320abce3057b8bf932247a636e8a/abs-mode.el#L58) - just use the bare executable name. `executable-find` might lock in a wrong path that is changed later, whereas the bare program name will be looked up using the path that is in effect at the time it is needed.

Looks good otherwise. :-)

rudi:
Many thanks for the review!  Both issues are fixed in e676a02e2b64fe84abaa7360ef4051beb6ea080f

Additionally, 726a1a0f2292779702d905777f1e545033af592f obsoletes `abs-indent`, which was inactive since we switched to cc-mode indenting (existing users have this as a file-local variable so it can't be removed without disrupting their work).

rudi:
Ping - is there something else I should do?  No rush, just want to make sure we're not mutually waiting on each other.
(Should I have removed the "awaiting-upstream" tag somehow?)

purcell:
> Ping - is there something else I should do? No rush, just want to make sure we're not mutually waiting on each other.
> (Should I have removed the "awaiting-upstream" tag somehow?)

Nope, looks good. Merged now - thanks for your patience.

PR #5847, opened 2018-12-04
Add recipe for auto-close-tag

jcs090218:

Automatically add HTML/XML close tag.

https://github.com/jcs090218/auto-close-tag

I am the maintainer.

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/auto-close-tag
+++ b/recipes/auto-close-tag
@@ -1,0 +1,1 @@
+(auto-close-tag :repo "jcs090218/auto-close-tag" :fetcher github)


purcell:
There's a lot of fiddly code here, and I feel like this behaviour must already be supported by other code. For example, the built-in `electric-pair-mode` would presumably make all this easy. Have you looked at that?

jcs090218:
I notice there are many packages that do the similar task. Here are several reasons I made this package..  
* This package is simply ported from <a href="https://marketplace.visualstudio.com/items?itemName=formulahendry.auto-close-tag">auto-close-tag</a>, this is a VSCode extension and I think this package could benefits people who will like to switch their editor from VSCode to Emacs.
* This is not a heavy package, and I do not think there is another package serve exactly the same purpose, which mean the package is unique and do not require them to customize much. (This reason is base on the first reason)
* I have checked the `electric-pair-mode` but I am not quite sure this would do the trick. Despite `electric-pair-mode` could done this, I am not sure this is the better way to achieve this? I was thinking `electric-pari-mode` is more likely for pairing symbol not pairing thing like 'tag'?
* I have compare this package to `web-mode`, and I realize `web-mode` does not have feature like this. Even though, I suggest having more alternatives than just one package conquer the whole web development environment. 
* I personally does not heavily depend on this package for my daily life of coding. I googled and acknowledge there are community likely to have this type of habit on coding the html. People from switching their previous editor to Emacs seem to ask for package like this, and the Emacs community rather giving them the solution, they suggest other alternatives/methods, which I do not think this is a good solution because people are changing their habit from changing their editor. Indeed, many people out there advocate "just use web-mode!", not trying to offense `web-mode`, `web-mode` is a great package. Here I am just saying this could be one of the alternative for what if people do not want to use `web-mode`? Anyway, this package compatible to `web-mode` so this could be an augmentation too.

purcell:
While both `web-mode` and `sgml-mode` support tag auto-closing when the user types `</`, I can see a package like yours might still be helpful, and I agree that `electric-pair` doesn't look suitable.

I think the following code would be fairly equivalent to your current code, no?

```el
(defun auto-close-tag-self-insert (count)
  (interactive "P")
  (self-insert-command (or count 1))
  (unless count
    (sgml-close-tag)))

(defcustom auto-close-tag-mode-map (make-sparse-keymap)
  "Keymap for `auto-close-tag-mode'.")

(define-key auto-close-tag-mode-map (kbd ">") 'auto-close-tag-self-insert)

(define-minor-mode auto-close-tag-mode
  "Automatically close HTML tags when '>' is entered."
  nil
  " >" auto-close-tag-mode-map)

```

jcs090218:
Hi, I am not sure about the `sgml-mode`, I know `web-mode` supports closing tag when hit `</`. The difference is this package support closing tag with `>`. After thinking all this, I reckon I could just contribute to `web-mode` or `sgml-mode` instead of making a whole new and individual package. Thanks for reviewing this package!

purcell:
Great! In the meantime, feel free to test the code above for a while and see. :-)

> I am not sure about the `sgml-mode`, I know `web-mode` supports closing tag 

The trick with sgml-mode is:

`(setq sgml-quick-keys 'close)`

PR #5849, opened 2018-12-06
add literate-elisp recipe

jingtaozf:
a literate programming tool to load emacs lisp codes from org mode file directly

https://github.com/jingtaozf/literate-elisp

author/maintainer

new package

### Checklist

Please confirm with `x`:

- [X ] The package is released under a [GPL-Compatible Free Software License](https://www.gnu.org/licenses/license-list.en.html#GPLCompatibleLicenses).
- [ X] I've read [CONTRIBUTING.org](https://github.com/melpa/melpa/blob/master/CONTRIBUTING.org)
- [ X] I've used the latest version of [package-lint](https://github.com/purcell/package-lint) to check for packaging issues, and addressed its feedback
- [ X] I've built and installed the package using the instructions in [CONTRIBUTING.org](https://github.com/melpa/melpa/blob/master/CONTRIBUTING.org)

### DIFF
--- a/recipes/literate-elisp
+++ b/recipes/literate-elisp
@@ -1,0 +1,1 @@
+(literate-elisp :fetcher github :repo "jingtaozf/literate-elisp")


equwal:
How is this better than using org-mode to tangle?

equwal:
Add `-*- lexical-binding: t; -*-` to the end of the first line, and run `M-x package-lint-current-buffer` in the buffer for your `.el` file from within emacs.

jingtaozf:
fixed,Thanks!

equwal:
Jumping to untangled org source is duplicated here:
https://github.com/melpa/melpa/pull/5844

jingtaozf:
Please note that literate-elisp.el is also generated from literate-elisp.org automatically, to setup this library ,but after that, all source codes can be written and loaded in a org file without tangling to any elisp file.

purcell:
Thanks, interesting. Couple of notes:
- The `literate-elisp.el` file should contain a helpful `Commentary`. The note that it was auto-generated will not look good on the MELPA package page. :-)
- The code formatting is quite idiosyncratic (lots of unexpected newlines) and hard to read

jingtaozf:
I have changed the way tangling lisp codes from org file to keep them unchanged.
So it should be easy to read now.

purcell:
Thanks! Still a couple of little points still to address here, which `package-lint` should  have listed:
- All the identifiers should begin with the `literal-elisp-` prefix
- You're using `cl-lib`, so you'll need to include `(cl-lib "0.6")` in the `Package-Requires` header, or `(emacs "24.3")`.

jingtaozf:
Fixed,thanks!

jingtaozf:
Thanks.:)

PR #5850, opened 2018-12-06
Add ivy-explorer

clemera:

Provide a dynamic file browsing grid using ivy 

https://github.com/clemera/ivy-explorer

I'm the author

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/ivy-explorer
+++ b/recipes/ivy-explorer
@@ -1,0 +1,2 @@
+(ivy-explorer :repo "clemera/ivy-explorer"
+              :fetcher github)


riscy:
This looks cool.  `M-x checkdoc` gives me these:
```
132:1: error: "display-line-numbers" doesn't start with package's prefix "ivy-explorer".
133:1: error: "golden-ratio-mode" doesn't start with package's prefix "ivy-explorer".
267:1: error: "ivy-dispatching-done-columns" doesn't start with package's prefix "ivy-explorer".
429:1: error: "counsel-explorer" doesn't start with package's prefix "ivy-explorer".
```
The first two are compiler pacifier as far as I can tell and you can instead write:
```elisp
(defvar display-line-numbers)
(defvar golden-ratio-mode)
```
The third and fourth I think you should look into.

clemera:
> (defvar display-line-numbers)
(defvar golden-ratio-mode)

I didn't new that, fixed. Thanks!

> 267:1: error: "ivy-dispatching-done-columns" doesn't start with package's prefix "ivy-explorer".

This is because the variable is from `ivy-hydra`, I thought it would be OK to just introduce a defvar with its default value. Anyway I changed it, so the compiler does not complain and `ivy-hydra` does not need to be loaded.

> 429:1: error: "counsel-explorer" doesn't start with package's prefix "ivy-explorer".

This one is intentional and I would like to leave it that way, if possible. Its the counterpart to ivy-explorer. Otherwise I would have to name it ivy-explorer-counsel. But all other counsel commands start with counsel-.

clemera:
> ivy-explorer-separator is more likely to look reasonable in most themes if you :inherit a built-in face (like border)

Thanks that is a good point. I have used the same face definition that `lv` uses because it uses this for separators for the dispatch (if the user has set `lv-use-separator`). I'm not sure which is better, adjust for the theme or adjust to look the same as `lv-use-separator` for now I changed it to inherit from `lv-use-separator`.

> display-line-numbers and golden-ratio-mode should be unnecessary

This is copied from `lv`, ivy-explorer uses a second lv like window for it's display. I guess the reason is that those modes could be global and mess up the `lv` message display buffer?

clemera:
I fallback to border face now, when `lv` is not installed. I thought it's a dependency if `ivy` itself but that isn't the case.

purcell:
> This is copied from `lv`, ivy-explorer uses a second lv like window for its display. I guess the reason is that those modes could be global and mess up the `lv` message display buffer?

Yes, I'd caution against copying-and-pasting code without completely knowing why. I think if `lv` is doing this same thing, then that is also misguided. There are any number of packages/options that might interact with a library, and you can't (and shouldn't) try to defeat them all.

purcell:
I guess I'm not going to insist on it, but I do think it's hard to know where to draw the line -- at a certain point, users have to be responsible for integration issues that result from the combinatorial explosion of their potential choices.

More compelling, I think, is that well-meaning workarounds can cover up bad design choices. For example, it looks like `golden-ratio-mode` already allows for buffer exclusions, ie. a convenient mechanism for users to decide how they want golden ratios to apply to their explorer buffers, so if that's not possible here for some reason, then that's an opportunity to make it possible, or to improve `golden-ratio-mode`. If fix-up code can only be effectively embedded in your code, then that suggests you might need to provide a hook that is run when the buffer is set up, for example.

`display-line-numbers` is built-in, so perhaps less controversial, but again: where to stop? I can immediately think of `show-trailing-whitespace`, that you might also "need" to set to `nil` here.

Food for thought, anyway. :-) Will go ahead and merge in the meantime.

clemera:
Thanks for merging and thanks for sharing your thoughts, they are instructive!

PR #5851, opened 2018-12-06
Add treemacs-icons-dired.

Alexander-Miller:

Small extra package for treemacs (mostly based on all-the-icons-dired) that'll [add treemacs' icons to dired buffers](https://github.com/Alexander-Miller/treemacs/blob/master/screenshots/dired-icons.png).

I have also included changes to the file structure to move the growing number of extra packages like this one from the main elisp directory into their own. For now these extra packages (treemacs-evil, treemacs-projectile and the unfinished treemacs-magit) are copied in both directories. Once this PR is merged I will delete them from /src/elisp.

https://github.com/Alexander-Miller/treemacs/blob/master/src/extra/treemacs-icons-dired.el

Maintainer

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/treemacs
+++ b/recipes/treemacs
@@ -5,5 +5,4 @@
          "icons"
          "src/elisp/treemacs*.el"
          "src/scripts/treemacs*.py"
-         (:exclude "src/elisp/treemacs-evil.el"
-                   "src/elisp/treemacs-projectile.el")))
+         (:exclude "src/extra/*")))
--- a/recipes/treemacs-evil
+++ b/recipes/treemacs-evil
@@ -1,4 +1,4 @@
 (treemacs-evil
  :fetcher github
  :repo "Alexander-Miller/treemacs"
- :files ("src/elisp/treemacs-evil.el"))
+ :files ("src/extra/treemacs-evil.el"))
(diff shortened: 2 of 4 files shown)


riscy:
Just checking this out.  `treemacs-icons-dired.el` requires `(treemacs "0")` and package-lint seems to always at least want a decimal in the version number.  But it looks like treemacs itself is on version 2.3?

Alexander-Miller:
Yes, that was Steve's idea back when treemacs and its companion treemacs-evil were first added to melpa. 

Basically I do not want to put small addons like this one into their own repositories due to their trivial nature (for example treemacs-evil is just a collection of keybinds) and instead keep them inside the main treemacs repo. In addition the addons do not require any specific version of treemacs, they are always supposed to go with the current master, without exceptions.

This is why Steve had advised me to require at least a treemacs version of 0, since the addons do still depend on the main package. If you check treemacs-evil and -projectile you'll see that their setup is the same. So yeah, it's a package-lint warning, but it's all going according to plan.

riscy:
Oh yeah, that makes sense.  I think you can nix the warning if you change `"0"` to `"0.0"`, but of course that's just cosmetic

purcell:
Thanks for this, and for your patience.

PR #5852, opened 2018-12-06
Add org-snooze recipe

xueeinstein:

Inspired by Google Inbox, org-snooze helps you to snooze your code, doc, feed at a given time. 

https://github.com/xueeinstein/org-snooze.el

maintainer.

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/org-snooze
+++ b/recipes/org-snooze
@@ -1,0 +1,3 @@
+(org-snooze
+ :repo "xueeinstein/org-snooze.el"
+ :fetcher github)


DamienCassou:
Thank you for your PR. Good job. Here is some feedback:

1. please add `-*- lexical-binding: t; -*-` at the end of the first line
2. please explain in the README what you mean by "appt reminder"
3. please use `~/.emacs.d/` as base directory for the records file
4. please clarify the docstrings of your variables and functions. For example, what do you mean by "title of org file", what does "time of late today" mean and what is this string going to change for the user.
5. please use `defcustom` instead of `defvar` for variables whose value can be changed by users
6. please run checkdoc to avoid typos in your docstrings (e.g., "propertiese")
7. `org-snooze-left-trim-special-chars` could use `string-trim-left` (same for `org-snooze-trim-*`)
8. `org-snooze-trim-state-chars` could use `string-prefix-p` which doesn't require `save-match-data`. The `dolist` form could have `s` at the end of its first argument so you wouldn't have to add `s` after the `dolist`. The `(if (string-match…` could be replaced with a `when`. The function could be renamed not to end with `-chars` because it's trimming states directly.
9. I suggest you add a `;;;###autoload` cookie before `org-snooze` function so it could be used before `org-snooze` is required.
10. You may extract functions out of `org-snooze`. This will ease code reading and automated testing.
12. Your code might face problem if the snooze file starts with some configuration lines (e.g., for export).
13. `add-to-list` already checks if the new element is already part of the list

xueeinstein:
Hi @DamienCassou, thank you for your review! I updated my code to support new function and followed most of your reviews.  Please have a check.

purcell:
Thanks, including for your patience! This looks useful, and great job fixing things up. Some further notes from me:
- Why not use a single `replace-match` pass [here](https://github.com/xueeinstein/org-snooze.el/blob/5fdf5025563851b695bb0e72811ea6cc48153685/org-snooze.el#L55)? You can combine the list of strings into a single regexp using `regexp-opt`.
- Please place the `lexical-binding` form on the first line, after the package description. Some Emacs versions ignored it if it was not on that line.

Neither of those is a blocker to merging, which I will do now - welcome to MELPA! :tada:

xueeinstein:
Hi Purcell, I updated my code followed your new comments. Thank you for the merge!

PR #5855, opened 2018-12-07
Add recipe for wiki-this

jcs090218:

Simply search in Wikipedia under point.

https://github.com/jcs090218/wiki-this

I am the maintainer.

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/wiki-this
+++ b/recipes/wiki-this
@@ -1,0 +1,1 @@
+(wiki-this :repo "jcs090218/wiki-this" :fetcher github)


riscy:
Suggestions:
- make the language customizable, but not the URL
- use `(interactive "r")` to make region specification more idiomatic
- use `(thing-at-point 'symbol)` so you can use this on e.g. `brother-in-law`
- use `(format ...)` instead of `concat` and various helper functions
- drop `(require 's)` - in fact wikipedia is happy to let you put whitespace in url's
- replace all whitespace , not just `" "`

Roughly something like this:
```elisp
(defvar wiki-this-url "https://%s.wikipedia.org/wiki/%s")
(defcustom wiki-this-language "en" "Language preference." :type 'string)
(defun wiki-this (&optional start end)
  "Open Wikipedia re: the point or from the region from START to END."
  (interactive "r")
  (let ((keyword (if (use-region-p)
                     (buffer-substring start end)
                   (thing-at-point 'symbol))))
    (if keyword
        (browse-url (format wiki-this-url wiki-this-language keyword))
      (message "Nothing to wiki!"))))
```

At that point this becomes quite a small package, albeit useful.  Maybe adding it to the MELPA package manager is overkill, but that's not for me to say.

jcs090218:
@riscy Thank you for those suggestions! It really helps me polishing on lisp and elisp. Indeed, I know this is a small package, and still, this is a small package I often used and I think this could be useful to anyone like me. On the other hand, I think any wiki related actions can be add or expand to this package in the future.

DamienCassou:
Please consider contributing to https://github.com/jozefg/wiki-summary.el instead.

jcs090218:
@DamienCassou I did thought about that, but I do know if that require changing the package name or not. Anyway, I will at least try contact the current maintainer of the package `wiki-summary` and see what he thinks.

alphapapa:
Yes, this package essentially boils down to this function:
```el
(defun wikipedia (term)
  (interactive (list (if (use-region-p)
                         (buffer-substring (region-beginning) (region-end))
                       (read-string "Term: " (thing-at-point 'word)))))
  (browse-url (format "https://wikipedia.org/wiki/%s" (url-hexify-string term))))
```
It should be easy to add that to another package.

jcs090218:
I think I will at least wait until @jozefg get back to me because I do not repeatedly want to open and reopen this PR before seeing a clear direction for this functionality/package. :D

alphapapa:
@jcs090218 What I mean is, we've already established that this shouldn't be a package of its own, because it's essentially just one function.  Even if jozefg doesn't respond, I'm sure you can find some package to contribute it to or another place to publish it, like /r/emacs or something.  So you can close this PR now.

jcs090218:
Okay, since this is so much effort. I’m done.

PR #5856, opened 2018-12-09
Add recipe for ms-python

xhcoding:

Python support for lsp-mode using [Microsoft Python Language Server](https://github.com/Microsoft/python-language-server)

https://github.com/xhcoding/ms-python

I am a maintainer

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/ms-python
+++ b/recipes/ms-python
@@ -1,0 +1,1 @@
+(ms-python :repo "xhcoding/ms-python" :fetcher github)


purcell:
Looks good, and thanks for your patience. The only note I have for you is that the "applications" parent group for the `defgroup` should probably be `languages`, or perhaps even `python`. Merging now - welcome to MELPA! :tada:

PR #5857, opened 2018-12-09
Add recipe for unicode-math-input.el

astoff:

This package includes a more comprehensive variant of the built in TeX input method (includes all symbols listed in this file: http://ctan.org/tex-archive/macros/latex/contrib/unicode-math/unimath-symbols.pdf)

It also includes an Ivy method to browse and insert math symbols.

https://github.com/astoff/unicode-math-input.el

I'm the author

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/unicode-math-input
+++ b/recipes/unicode-math-input
@@ -1,0 +1,1 @@
+(unicode-math-input :fetcher github :repo "astoff/unicode-math-input.el")


astoff:
package-lint complains that the command `unicode-math-browse` is not prefixed with the package's name. Is this unacceptable? I could call it `unicode-math-input-ivy`, but that sounds a bit awkward.

jcs090218:
@astoff I think you do have to prefix it with `unicode-math-input-`, I reckon having the variable's name with prefix is ensure not having define the same variable name multiple times. You could rename the variable with prefix and also make you feel less awkward?

purcell:
Hi, and thanks for your patience!

Yes, the symbol prefix must match the package name, so `unicode-math-browse` could be renamed to `unicode-math-input`.

The `ivy` dependency is quite opinionated and I wonder if it's actually necessary: `completing-read` is overridden by `ivy`/`helm` etc. when the user enables them, and `completing-read` (and Ivy) support providing a hash table of completions. I suspect you may be able to just call `completing-read` with a hash table mapping the input strings to the output characters.

Otherwise, this looks great!

astoff:
Thanks for the feedback! I fixed the namespace issue, and I moved from ivy to a simple `completing-read`. Everything should be fine now.

PR #5858, opened 2018-12-10
Add recipe for coleslaw-mode

equwal:

A mode for editing files used by [kingcon's "coleslaw" site content generator](https://github.com/kingcons/coleslaw) to generate static html content. The interesting bits are:
- Every file requires a header (bound only in `.page` and `.post` files to `M-;`. It won't function otherwise).
- The type of header required changes based on the filetype: `.page` and `.post` files need different headers.
- The user specifies the style of markup in the header, so the major mode needs to be changed accordingly.

I found myself hand writing these and making plenty of errors, while also having to `M-x lisp-mode` or `M-x markdown-mode` and  `M-x markdown-preview-eww` everytime I wanted to edit or make a new page. There are no hard dependencies.

https://github.com/equwal/coleslaw-mode

I am the creator. Coleslaw itself is here: https://github.com/kingcons/coleslaw

I am the upstream package maintainer. Coleslaw itself doesn't require any contact.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/coleslaw
+++ b/recipes/coleslaw
@@ -1,1 +1,2 @@
 (coleslaw :fetcher github :repo "equwal/coleslaw")
+
--- a/recipes/coleslaw
+++ b/recipes/coleslaw
@@ -1,2 +1,1 @@
 (coleslaw :fetcher github :repo "equwal/coleslaw")
-
--- a/recipes/coleslaw
+++ b/recipes/coleslaw
@@ -1,0 +1,1 @@
+(coleslaw :fetcher github :repo "equwal/coleslaw")
--- a/recipes/coleslaw-mode
+++ b/recipes/coleslaw-mode
@@ -1,0 +1,1 @@
+(coleslaw-mode :fetcher github :repo "equwal/coleslaw-mode")
--- a/recipes/coleslaw-mode
+++ b/recipes/coleslaw-mode
@@ -1,1 +1,0 @@
-(coleslaw-mode :fetcher github :repo "equwal/coleslaw-mode")


equwal:
Only unfixed automated complaint is package-lint asking me to depend on Emacs 24 for lexical scope. The packages doesn't actually use lexical scope so this isn't necessary.

equwal:
Thank you for reviewing my project. I made the suggested edits, and refactored the code. I have more planned but hopefully it is sufficient for acceptance. If not, please let me know.

DamienCassou:
Thank you for taking care of the feedback that fast. Here is some more:

- please squash your 4 commits into just one
- in `coleslaw-mode`, I would avoid adding `flyspell-mode` to the hook because users might prefer using something different. Also, you don't have to run `coleslaw-mode-hook` or set a mode map, this is done automatically because you use `define-minor-mode`. Finally, please remove `(setq minor-mode ...)` as there is no such a variable.
- `auto-insert` is a function, why would you set its variable value to `t`?
- I would move the changes to `auto-insert-alist` and the addition of `auto-insert` to `coleslaw-mode-hook` to a `coleslaw-setup` function that the user has to call if he wants the same setup has you.
- please replace the `(add-hook 'coleslaw-mode-hook 'auto-insert)` with an `(auto-insert)` line within the body of `coleslaw-mode`
- either markdown-mode is a dependency of your package or it is not. If it is, you should add it to "Package-Requires". If it is not, you should remove the `require`

equwal:
Wow thanks! Once again I have adopted your recommendations, and I squashed the commits. I'll make sure to branch it for development from now on.

purcell:
Closing due to lack of response -- please feel free to re-open.

equwal:
Thanks for all that feedback. I've come back and hopefully improved it nearer to acceptance, I reapplied here:
https://github.com/melpa/melpa/pull/6232

After some major edits.

PR #5859, opened 2018-12-10
Added the TeaCodeExpand package to use TeaCode with Emacs on macOS.

raguay:

This package allows for the insertion of snippets from the (TeaCode)[https://www.apptorium.com/teacode] for the macOS.

https://github.com/raguay/TeaCodeExpand

I'm the package creator and maintainer. My first one!

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/TeaCodeExpand
+++ b/recipes/TeaCodeExpand
@@ -1,0 +1,3 @@
+(TeaCodeExpand
+   :fetcher GitHub
+   :repo "raguay/TeaCodeExpand")


purcell:
Thanks for the submission, and to add to the comments from @riscy:

- Don't `global-set-key`: the user can do that.
- Add an `;;;###autoload` cookie comment above the command
- `(buffer-file-name)` can return nil, so you couldn't use this command in buffers that aren 't backed by a file: this seems broken.
- You could replace your nested `let` with a `let*` to allow referring to previously-mentioned bindings. Alternatively, use (and depend on) `let-alist` for easy access to `alist` items.
- Why is expansion line-wise? It feels like the command should take a region, and maybe fall back to the current line if `(use-region-p)` returns `nil`.
- There's no need to use the full path to `osascript`: you can drop the `/usr/bin`
- You'll potentially have a Very Bad Time™ if the line contains a string like `';rm -rf /`. You can use `shell-quote-argument` to make command argument strings safe.

Hope that helps.

purcell:
Hey @raguay - I think I closed this accidentally. In any case, I sent you some fixes in https://github.com/raguay/TeaCodeExpand/pull/1, and then I think we'd be good to go, except that the recipe would need to be changed to match the new non-camel-case name.

purcell:
(I'd suggest filing a fresh pull request with a new recipe, and provide a link back to this one.)

PR #5860, opened 2018-12-11
Add right-click-context package

zonuexe:

A customizable context menu, call it from right-click of mouse.

https://github.com/zonuexe/right-click-context

I'm the author.

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/right-click-context
+++ b/recipes/right-click-context
@@ -1,0 +1,4 @@
+(right-click-context
+ :repo "zonuexe/right-click-context"
+ :fetcher github
+ :files ("right-click-context.el"))


DamienCassou:
Thank you for you package. Good job. Some feedback:

1. please add `-*- lexical-binding: t; -*-` at the end of the first line. This will require moving the definitions of `beg` and `end`.
2. prefix function names with #' instead of just ' to tell the compiler this is a function reference. E.g., `(define-key map … #'right-click-context-menu)` and `(right-click-context--process-region beg end 'url-encode-url)`
3. please surround `right-click-context-mode` with quotes in the docstring of `right-click-context-mode-map`
4. please remove the commented menu items in `right-click-context-global-menu-tree` or at least explain there why they are commented
5. use two semi-colons `;;` instead of just one before the `"Rectangle"` comment so it indents properly
6. the `type` of `right-click-context-global-menu-tree` could be
   specified with more details so it allows using the customize interface
7. why do you add `nil t` in the `require` line for `popup`? This package doesn't seem optional
8. why do you require `undo-tree`?
9. `right-click-context--build-menu-for-popup-el` please use longer variable names so people can better understand the source code
10. the error message "Invalid tree" doesn't take into account the recursive calls. So, "Invalid tree. (0-th elm)" could be the very first item of the tree, or the first item of any branch.
11. you can use `defvar-local` instead of `defvar` and `make-variable-buffer-local`
12. in `right-click-context--process-region`, you can use `let*` instead of `let` followed by `setq`. You can also use `when` instead of `if` to get rid of 1 indentation level for the 3 last statementsright-click-context--process-region

zonuexe:
@DamienCassou Thank you for reviewing!
Since this code was written two years ago, there was a change in my thought during that time.

I solved a lot of pointing out.
https://github.com/zonuexe/right-click-context/compare/0.1.0...master

> please add `-*- lexical-binding: t; -*-` at the end of the first line. This will require moving the definitions of `beg` and `end`.

Probably this indication is inaccurate.
Dynamic binding of variables will work even in files declared with `lexical-binding: t;`.
However my idea of short coding has changed, I deleted it because it is also contrary to the Lisp coding convention.

-----

By the way I am your fan as I used your PPA all the time in Ubuntu a few years ago.
I am pleased to receive your detailed review.

DamienCassou:
> By the way I am your fan as I used your PPA all the time in Ubuntu a few years ago.

thank you :-).

purcell:
It'd be nice if `region-convert` were in a separate repo, so that no `:files` spec would be needed in this recipe. It's a completely separate package, right?

zonuexe:
@purcell I see.  I will split the repository this evening.

PR #5862, opened 2018-12-14
Add recipe for agtags

vietor:

Another Emacs frontend to GNU Global. It support [**wgrep**](https://github.com/mhayashi1120/Emacs-wgrep) in **agtags-grep-mode**. Add simple Xref support.

https://github.com/vietor/agtags

Author and maintainer

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/agtags
+++ b/recipes/agtags
@@ -1,0 +1,3 @@
+(agtags
+ :repo "vietor/agtags"
+ :fetcher github)


purcell:
Thanks for your patience. I agree with @riscy's helpful comments above, and have some additional notes:

- You've indicated above that this is licensed compatibly with the GPL, but there's no evidence of that in the repo or file headers. Please make the licence explicit in one or both of those.
- Consider [string-empty-p](https://github.com/vietor/agtags/blob/f4b50361288631e5c27e3fd030fcc6b2dbcae289/agtags.el#L71) here
- As a point of style, [this](https://github.com/vietor/agtags/blob/f4b50361288631e5c27e3fd030fcc6b2dbcae289/agtags.el#L127-L129) would be clearer without binding the local variable "user-input": introducing additional names doesn't always increase code clarity.
- Consider a clearer mode lighter like `" AgTags"` [here](https://github.com/vietor/agtags/blob/f4b50361288631e5c27e3fd030fcc6b2dbcae289/agtags.el#L271) -- users can always abbreviate it locally if they prefer something shorter, and you can even make it customisable.
- As `package-lint` would have pointed out, "emacs" is redundant in the package description "emacs frontend ..." -- all elisp packages are for Emacs. :-)

Hope that helps!

vietor:
Thanks for the review!

 I have corrected problems.

agtags.el:228:1:Warning: Unused lexical argument ‘status’ (you can change this to _status)
> Modified.

package-lint complains about line 102 for me (you can just add a period)
> Add (require 'compile) to try fix it. In my  environment `package-lint` always show 'No issues found'.

It's not a bad idea to add a docstring when you defconst (several places).
> Add more docstring.

You have a few functions that are 'private' (double-dashed), but also interactive ...
> Move that functions to public.

vietor:
Thanks @purcell , i have corrected.

PR #5863, opened 2018-12-14
add flymake-diagnostic-at-point recipe

waymondo:

This provides a handy minor mode to show the diagnostic information at point for newer versions of flymake.

https://github.com/meqif/flymake-diagnostic-at-point

enthusiastic user

none needed

### Checklist

Please confirm with `x`:

connect to https://github.com/meqif/flymake-diagnostic-at-point/issues/3

### DIFF
--- a/recipes/flymake-diagnostic-at-point
+++ b/recipes/flymake-diagnostic-at-point
@@ -1,0 +1,1 @@
+(flymake-diagnostic-at-point :repo "meqif/flymake-diagnostic-at-point" :fetcher github)


purcell:
Thanks. Quick notes:

- You should add a `defgroup` for the `flymake-diagnostic-at-point` custom group so that it is placed appropriately in the customisation hierarchy.
- Prefer a `boundp` or `fboundp` test [here](https://github.com/meqif/flymake-diagnostic-at-point/blob/379616b1c6f5ebeaf08fbe54ae765008a78b3be7/flymake-diagnostic-at-point.el#L115) if possible.
- The `Commentary` doesn't explain how to use this package, and the autoloads are odd, so it's a bit confusing overall. I'd expect that the minor mode would be autoloaded, and that the package would then be enabled by adding `flymake-diagnostic-at-point-mode` to `flymake-mode-hook`.

Nonetheless, I'll go ahead and merge this now, and you can address the above in your own time. :-)

PR #5864, opened 2018-12-16
Add eldoc-box

casouri:

Make eldoc display document messages in childframe

https://github.com/casouri/eldoc-box

Maintainer (wrote the package but copied code from lsp-ui, hence maintainer)

**None needed**

### Checklist

Please confirm with `x`:

- [x] `M-x checkdoc` is happy with my docstrings (except a minor one)

### DIFF
--- a/recipes/eldoc-box
+++ b/recipes/eldoc-box
@@ -1,0 +1,1 @@
+(eldoc-box :repo "casouri/eldoc-box" :fetcher github)


casouri:
Done, does it look ok?

alphapapa:
That looks very cool!  I'd suggest taking the screenshot again with the child frame's background a bit lighter, though, because I had to look very closely to see where the child frame was.

One other thing, it would be helpful if you could clarify the differences between the two functions:

> eldoc-box-hover-mode
> Show documentation upon hover. Note that you need to enable ElDoc mode for this to work.

> eldoc-box-hover-at-point-mode
> This minor mode shows doc at point. Don’t try to enable both minor mode: you can only enable one in the same time.

They both have `hover` in their names, but only one's docstring says `hover`, so I'm not sure which does what.

Also, regarding:

> Don’t try to enable both minor mode: you can only enable one in the same time.

Why not add a check in the modes to disable the opposite mode, so the user can't do it wrong?

casouri:
Thank you for taking your time.

I've changed the screenshot, updated the explanation of the two minor modes, and added safety code in minor mode definitions.

Does the screenshot look ok? It seems distinctive in my eyes but I'm not sure if it's the same for others.

alphapapa:
Yes, the doc frame is easier to distinguish in the screenshot now.

purcell:
Thanks for this, and for your patience. Couple of things stand out to me:

- The code is surprisingly complex, and I'm surprised you can't use something like `posframe` here. Is there a reason that's not possible?
- Lots of typos, e.g.: "childframw", "cildframe"

casouri:
> The code is surprisingly complex

Can't deny it. The main reason is to fix edge cases here and there and try to play nice with eldoc, which is mainly designed for displaying stuff in the echo area. If you look at the [initial version](https://github.com/casouri/eldoc-box/tree/bfda87893d1835fce022b21c40b9fca934ee56f5), you can see it is indeed pretty simple. All the superfluous code boils down to 1) keep eldoc from hiding the childframe when it shouldn't, 2) hide the childframe when it should be hidden.

> I'm surprised you can't use something like posframe here. Is there a reason that's not possible?

Not really. It is just because lsp-ui doesn't use posframe. At first, I just want a `lsp-ui-doc` for eglot because eglot doesn't have one (and won't include one). So I went to `lsp-ui`, copied some code from it and modified the code to work with eldoc (because eglot only have an API through eldoc). It was initially named `eglot-doc`.

Then the author of eglot said it can be used with eldoc in general, instead of eglot. It's a good idea so I changed the name to `eldoc-box` and removed eglot dependency.

The code I copied from `lsp-ui` never had any problem so I didn't bother to replace it with posframe.

> Lots of typos, e.g.: "childframw", "cildframe"

Sorry for that, "childframe" is marked wrong itself in flyspell so I didn't pay much attention to it. They should be fixed in the latest update.

Hope I answered your questions.

purcell:
Makes sense, merged now: thanks for your patience, and sorry for the slow turnaround!

PR #5865, opened 2018-12-17
Add recipe for k8s-mode

TxGVNN:

The major for developing Kubernetes files. The main purpose is to support snippet.

https://github.com/TxGVNN/emacs-k8s-mode

Maintainer

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/k8s-mode
+++ b/recipes/k8s-mode
@@ -1,0 +1,4 @@
+(k8s-mode
+ :repo "TxGVNN/emacs-k8s-mode"
+ :fetcher github
+ :files ("*.el" ("snippets" ("k8s-mode" "snippets/k8s-mode/*"))))


purcell:
Thanks! Just a couple of comments:

- The [yasnippet](https://github.com/TxGVNN/emacs-k8s-mode/blob/5dc9c655c6b618123fc656250798d0294680b831/k8s-mode.el#L65-L68) code is misplaced. Not only will it cause the snippets to be loaded globally every time the mode is enabled locally in a buffer, but it also hard-wires the enabling of `yas-minor-mode`, which is bad behaviour since the user might not want that. Instead, just wrap the `yas-load-directory` in an `(eval-after-load 'yasnippet ...)` form at the top level.
- Consider `push`-ing `k8s-complete-at-point` onto `completion-at-point-functions` rather than replacing the value of that variable entirely: it might contain other helpful completion functions, which the current code will wipe out.

TxGVNN:
Thank you for your review and feedback. It will be really valuable experiences to me. I updated on upstream

>     *It feels like there's an opportunity here to font-lock even more patterns than just apiVersion. Maybe in future versions?

Actually, I see the font-lock has been support by yaml-mode (parent mode) is still good. So I just add `apiVersion` keyword to divice the object sections. WDYT?

>     *You can quiet a couple compiler nags with: (declare-function yas-load-directory "yasnippet") (declare-function yas-minor-mode "yasnippet")
Yep, I updated

>     *In the Commentary: you mention using a magic comment to enable k8s-mode. Is it worth mentioning that it's not feasible to use the convention of adding to auto-mode-alist since k8s config files are yaml files, and that's why the solution is a magic comment?

You're right, I thought I can not add *.yaml to auto-mode-alist. Do you have any good idea?

>     *For extra portability (e.g. someone just copying the .el file around without the snippets -- who knows how people might use this!), you could add a check that the yasnippet directory exists e.g. (when (and (featurep 'yasnippet) (file-directory-p k8s-snip-dir)), or even warn users when it is missing (Emacs will complain, but it isn't very instructive).

I updated

>     * The [yasnippet](https://github.com/TxGVNN/emacs-k8s-mode/blob/5dc9c655c6b618123fc656250798d0294680b831/k8s-mode.el#L65-L68) code is misplaced. Not only will it cause the snippets to be loaded globally every time the mode is enabled locally in a buffer, but it also hard-wires the enabling of `yas-minor-mode`, which is bad behaviour since the user might not want that. Instead, just wrap the `yas-load-directory` in an `(eval-after-load 'yasnippet ...)` form at the top level.

I updated
>
>     * Consider `push`-ing `k8s-complete-at-point` onto `completion-at-point-functions` rather than replacing the value of that variable entirely: it might contain other helpful completion functions, which the current code will wipe out.

Did you mean?
```
(set (make-local-variable 'completion-at-point-functions)
       (add-to-list 'completion-at-point-functions 'k8s-complete-at-point))
```

purcell:
Thanks, I sent you https://github.com/TxGVNN/emacs-k8s-mode/pull/1. When that is merged, I think this PR will also be ready to merge.

Re. the auto-mode-alist question, if there isn't a consistent naming pattern for the k8s yaml files, then your current solution is the best.

purcell:
Great, merged, thanks! :-)

TxGVNN:
Thanks! :-)

PR #5866, opened 2018-12-18
Add with-venv recipe

10sr:

Execute functions with Python virtual environment activated

https://github.com/10sr/with-venv-el

The maintainer

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/with-venv
+++ b/recipes/with-venv
@@ -1,0 +1,1 @@
+(with-venv :repo "10sr/with-venv-el" :fetcher github)


purcell:
Looks good - merging now. Note that your `with-venv-dir` macro is currently susceptible to multiple-expansion of its `dir` variable: you can read about this issue in Practical Common Lisp, available free online. In short, if you pass an expression for `dir` that has a side effect, e.g. a function call which searches the filesystem and returns the directory, then that side effect can be performed multiple times. The solution for this is to use `gensym`.

10sr:
Thanks, I'll look into it!

PR #5867, opened 2018-12-18
Added the TeaCodeExpand package to use TeaCode with Emacs on macOS.

raguay:

This package will take a line of code and expand it using TeaCode on the macOS. TeaCode is a generic snippet expander that works in any code editor and not just Emacs. It also run scripts to fill in the expansion as well. 

https://github.com/raguay/TeaCodeExpand

I created it and I'm maintaining it.

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/TeaCodeExpand
+++ b/recipes/TeaCodeExpand
@@ -1,3 +1,0 @@
-(TeaCodeExpand
-   :fetcher GitHub
-   :repo "raguay/TeaCodeExpand")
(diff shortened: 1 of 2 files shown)


raguay:
I've made the changes asked for except for the error message if a file isn't a part of the buffer. I want to make it user definable as the default language to use for the expanding. But, I'm currently defaulting to "any language" as that is the normal procedure in other editors.

purcell:
> I've made the changes asked for except for the error message if a file isn't a part of the buffer. I want to make it user definable as the default language to use for the expanding. But, I'm currently defaulting to "any language" as that is the normal procedure in other editors

Oh cool, yes, makes sense.

I fixed an issue with the recipe and will merge now. Welcome to MELPA! :tada:

raguay:
Great Thanks. This is my first one! Hopefully I'll be doing more.

PR #5868, opened 2018-12-18
Add lsp-scala

rossabaker:

Integrates [lsp-mode](https://github.com/emacs-lsp/lsp-mode) with the [metals](https://scalameta.org/metals/) Scala language server.

https://github.com/rossabaker/lsp-scala

I am the maintainer.

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/lsp-scala
+++ b/recipes/lsp-scala
@@ -1,0 +1,1 @@
+(lsp-scala :repo "rossabaker/lsp-scala" :fetcher github)


purcell:
Thanks, including for your patience.

This mostly looks great, but the [`with-eval-after-load` form](https://github.com/rossabaker/lsp-scala/blob/9ce966300feaacfbeb414cd3f3b6761e5e8a4ef0/lsp-scala.el#L64) is misplaced, and I believe that when this code is loaded, `lsp` will always have been loaded alread, so you should just put that code at the top level. (`with-eval-after-load` is meant for user configs, and should generally not be found in package code.)

rossabaker:
We dropped support for the legacy version in melpa-stable, which eliminates the `with-eval-after-load`.

purcell:
Merged now, thanks for your patience, and sorry for the slow turnaround! Welcome to MELPA. :tada:

PR #5871, opened 2018-12-19
Add a recipe for ob-html-chrome.

nikclayton:
This is https://github.com/melpa/melpa/pull/5770 with the feedback applied. Unfortunately, there doesn't appear to be an easy way for me to re-open that pull request, so I have had to create this new one.


purcell:
Great, thanks @nikclayton, merging now.

PR #5872, opened 2018-12-19
adding recipe for kubernetes-helm

abrochard:

Provides utility functions to interact with Helm (https://helm.sh/), the kubernetes package manager.

https://github.com/abrochard/kubernetes-helm

Maintainer

**None needed** 

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/kubernetes-helm
+++ b/recipes/kubernetes-helm
@@ -1,0 +1,1 @@
+(kubernetes-helm :repo "abrochard/kubernetes-helm" :fetcher github)


purcell:
Thanks. A couple of quick notes:

- Don't build command lines with `concat` if you can avoid it: prefer Emacs process functions which accept an array of arguments, so that you do not have issues with special characters (including spaces) in file/directory names, for example. At the very least, here you would have to use `shell-quote-argument` to make the `directory` arguments safe.
- Adding `&` to the command and using `shell-command` is hacky: that function is made for interactive use, and here it would be better to use the functionality in `compile.el`, or to call something like `start-process` directly.
- Prefer `(unless x y)` to `(if (not x) y)`: generally, prefer `when` and `unless` to `if` in cases where there is only one branch.
- Typo: "teh"
- "Emacs utility functions and bindings for" can just be "Commands for". ("Emacs" is redundant in any case, since all elisp is for Emacs. :smile:)

Hope that helps!

abrochard:
Hi @purcell , thank you so much for the feedback! As always, it's greatly appreciated.
I'm on vacation right now, but I will tackle these as soon as I'm back. I believe that my (other PR)[https://github.com/melpa/melpa/pull/5873] suffers from the same flaws. I will update that one too.
Wishing you happy holidays!

purcell:
Merged now, thanks for your patience, and sorry for the slow turnaround!

abrochard:
Thank you! No worries!

PR #5873, opened 2018-12-19
adding Kubel recipe

abrochard:

Emacs extension for controlling Kubernetes pods with limited permissions. The existing extension will not be happy if you cannot list namespaces and given the scope of the change required to support that, and the actual scope of what you need to control pods, I propose this smaller extension as a solution.

https://github.com/abrochard/kubel

Maintainer

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/kubel
+++ b/recipes/kubel
@@ -1,0 +1,1 @@
+(kubel :repo "abrochard/kubel" :fetcher github)


tarsius:
@purcell You labeled this `awaiting-upstream` without requesting any changes.

abrochard:
@tarsius It's because I have [another PR](https://github.com/melpa/melpa/pull/5872) with some feedback that I'm currently working on, and I plan to apply some of the changes here too.

abrochard:
@purcell I just pushed changes to the package:

- avoiding concat and using process functions instead
- replaced `if` with only one branch by `when` and `unless`
- checked for typos, the best I could at least.

Thanks!

purcell:
Merged now, thanks @abrochard

abrochard:
Thank you!

PR #5874, opened 2018-12-20
Add a link to the opposite Melpa in the header

danielcompton:
We initially hide the link to prevent flashing changes when the page is loading, then unhide it once we have determined which Melpa we are on.


purcell:
Thanks - merged with some little fixes. Should show up on the live site in the next few hours. :-)

danielcompton:
It looks like both Melpa and Melpa Stable are both linking to Melpa Stable in the header now? I took a look at the JS being served and it seemed to match 
https://github.com/melpa/melpa/commit/378720400eecb7cc94b4aca19c40d1b0440ddd6e?

PR #5875, opened 2018-12-21
Add pack recipe

10sr:

Pack and unpack archive files

https://github.com/10sr/pack-el

The maintainer

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/pack
+++ b/recipes/pack
@@ -1,0 +1,1 @@
+(pack :repo "10sr/pack-el" :fetcher github)


Silex:
What's the difference with dired `Z` key (`dired-do-compress`)?

10sr:
`dired-do-compress` creates *compressed* files, where `pack-dired-dwim` creates *archive* files.
For example, when you mark two files named `a.txt` and `b.txt` in dired buffer and call these functions:

- `dired-do-compress` will create two files named `a.txt.gz` and `b.txt.gz`
- `pack-dired-dwin` will create one file like `a.txt.tgz` (this name is a default: this command will prompt you to input name for the archive file before creating it)

purcell:
Thanks! Nice work, and I'll go ahead and merge it now.

(PS. typo: elemetn)

10sr:
Thanks!

PR #5878, opened 2018-12-21
Recipe: Add yequake

alphapapa:

 This package provides configurable, drop-down Emacs frames, similar to drop-down terminal windows programs, like Yakuake.  Each frame can be customized to display certain buffers in a certain way, at the desired height, width, and opacity.  The idea is to call the =yequake-toggle= command from outside Emacs, using =emacsclient=, by binding a shell command to a global keyboard shortcut in the desktop environment.  Then, with a single keypress, the desired Emacs frame can be toggled on and off, showing the desired buffers.

http://github.com/alphapapa/yequake.el

Author, maintainer, enthusiastic user.

I consult with myself as necessary.

### Checklist

Please confirm with `x`:

Thanks for all your work on MELPA!

### DIFF
--- a/recipes/yequake
+++ b/recipes/yequake
@@ -1,0 +1,1 @@
+(yequake :fetcher github :repo "alphapapa/yequake")


alphapapa:
> Rather than shimming Emacs built-ins without a prefix, which is frowned upon, you should probably just use the dash versions of if-let/if-let*.

Well, I hate to do that, but the way Emacs 26 essentially "broke" `if-let` and `when-let` by renaming them without providing aliases, it seems like the best thing to do.  Otherwise, either Emacs 25 users or Emacs 26 users are going to get compile warnings.  And if every package aliased them to their own prefix, that would really pollute the obarray.  I have over 200 packages installed, and I sure wouldn't want 200 `foo--if-let`'s coming up when I do `C-h f`.

They also don't expand to the same as the `dash` counterparts:

```el
(if-let* ((a 1)
          (b a))
    a
  'nope)
```
Expands to:
```el
(let* ((a (and t 1))
       (b (and a a)))
  (if b
      a
    'nope))
```
While:
```el
(-if-let* ((a 1)
           (b a))
    a
  'nope)
```
expands to:
```el
(let ((a 1))
  (if a
      (let ((b a))
        (if b
            a
          'nope))
    'nope))
```
Of course they're functionally equivalent, but `if-let*` is faster than the `dash` version:
```
#+BEGIN_SRC elisp
  (bench-multi-lexical :times 100000 :ensure-equal t
    :forms (("if-let*" (if-let* ((a 1)
                                 (b a))
                           a
                         'nope))
            ("-if-let*" (-if-let* ((a 1)
                                   (b a))
                            a
                          'nope))))
#+END_SRC

#+RESULTS:
| Form     | x faster than next | Total runtime | # of GCs | Total GC runtime |
|----------+--------------------+---------------+----------+------------------|
| if-let*  | 1.08               |      0.007944 |        0 |                0 |
| -if-let* | slowest            |      0.008597 |        0 |                0 |
```
So, I don't know.  I wish they hadn't changed them without aliasing them back.  What do you think?

alphapapa:
LOL  :)  Thanks, Steve.

PR #5880, opened 2018-12-22
Add esonify recipe

oflatt:
An emacs extension that sonifies your code. Skilled users will begin to be able to recognize some code based solely upon the music it generates. Hear sine waves at different frequencies for lowercase letters, square waves for uppercase letters, and triangle waves for keybindings.

https://github.com/oflatt/esonify

creator/maintainer

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/esonify
+++ b/recipes/esonify
@@ -1,4 +1,4 @@
-(sound-wav
+(esonify
 :repo "oflatt/esonify"
 :fetcher github
-:files ("sounds/*"))
+:files ("*.el" "sounds"))
--- a/recipes/esonify
+++ b/recipes/esonify
@@ -1,0 +1,4 @@
+(sound-wav
+:repo "oflatt/esonify"
+:fetcher github
+:files ("sounds/*"))
--- a/recipes/esonify
+++ b/recipes/esonify
@@ -1,4 +1,4 @@
 (esonify
 :repo "oflatt/esonify"
 :fetcher github
-:files ("*.el" "sounds"))
+:files (:defaults "sounds"))


alphapapa:
Well...that sounds interesting...it's just begging for a demonstration video, though.  :)

oflatt:
Here is a demonstration video.
Audio quality is a bit poor because the recording buffers didn't seem to deal with all the new sounds very well.
<https://www.youtube.com/watch?v=fwBh6FKxzcQ>

oflatt:
esonify now has more documentation and customizable variables.

purcell:
Thanks! Some notes:

- In the recipe, `"*.el"` should be `:defaults`.
- `alphabethash.el` should be named `esonify-alphabethash.el`, if it's even needed -- it looks like the data is already in `esonify.el`
- There's no `defgroup` for the `esonify` customisation group: you should add one, to place it appropriately in the custom hierarchy.
- [This line](https://github.com/oflatt/esonify/blob/7cfea844f61090888c662b4c6fa1adc0713ab637/esonify.el#L42) seems hokey and unnecessary: you can just `(require 'esonify-sound-wav)` if you need it, however...
- `esonify-sound-wav` appears to be a "rebadged" `sound-wav.el`, which looks like an upstream library. It should be relied on directly, and published separately if necessary.
- [Here](https://github.com/oflatt/esonify/blob/master/esonify.el#L110-L119) there's no need to convert the symbol to a string -- just compare symbols, e.g. (eq last-command-event 'up)`. `pcase` is a neat way of writing these expressions, btw.

Hope that helps - shout out if you have questions.

oflatt:
Thank you for all of the advice.

About sound-wav.el- the reason I rebadged sound-wav.el is that, for my purposes, I would like it to use Play() instead of PlaySync() for playing through .net on windows. However, I am not sure this is best for sound-wav.el in general. I have tried to contact the maintainer of sound-wav and submitted an issue.

What's the best thing to do in this situation, in your opinion?

oflatt:
A bit confused about require in elisp- I'm getting
File error: Cannot open load file, No such file or directory, esonify-sound-wav
When trying to replace the line with (require 'esonify-sound-wav)

purcell:
> About sound-wav.el- the reason I rebadged sound-wav.el is that, for my purposes, I would like it to use Play() instead of PlaySync() for playing through .net on windows. However, I am not sure this is best for sound-wav.el in general. I have tried to contact the maintainer of sound-wav and submitted an issue.
> 
> What's the best thing to do in this situation, in your opinion? You said, "published separately if necessary." What does this entail? Another pull request to melpa?

Ah, I see https://github.com/syohex/emacs-sound-wav/issues/6 -- perhaps now that you've got your own patched version, you can submit a pull request to @syohex, and then simply depend on the `sound-wav` package.

If for some reason that's really not possible, then you could include `esonify-sound-wav.el` in the recipe, but you'd need to change all the `sound-wav-*` symbols to be `esonify-sound-wav-`. The first solution would be greatly preferable for everyone, I think.


> A bit confused about require in elisp- I'm getting
> File error: Cannot open load file, No such file or directory, esonify-sound-wav
> When trying to replace the line with (require 'esonify-sound-wav

If you're trying this locally, then yes, this will happen: you'd need to `M-: (push default-directory load-path)` to put the current directory on `load-path`, as would be the case when installed as a package. But also note that you won't be able to `require` the file unless its `provide` form provides a feature with a matching name, which isn't yet the case, as above.

oflatt:
For now, I did the esonify-sound-wav.el solution.
I also submitted the pull request on emacs-sound-wav.

purcell:
> For now, I did the esonify-sound-wav.el solution.

Makes sense. MELPA doesn't pull together package dependencies from all the files, only from the main file (esonify.el), so you'll also need to include the `deferred` and `cl-lib` dependencies for `esonify-sound-wav.el` in the `Package-Requires` line of `esonify.el`. Then I think we'll be good to go. :-)

oflatt:
Done.

oflatt:
Are we waiting on something to merge, or are we good to go?

purcell:
> Are we waiting on something to merge, or are we good to go?

There's still no `defgroup` as mentioned above, and the `defcustom` vars are (by definition) public, and so should use the `esonify-` prefix, not `esonify--`. Other than that, all looks good.

oflatt:
Sorry I missed that one, esonify now has a defgroup.
Thanks for the help.

purcell:
Sweet, merged now - thanks!

alphapapa:
@oflatt You might find this helpful when testing: https://github.com/alphapapa/emacs-package-dev-handbook#byte-compile-and-load-directory

PR #5882, opened 2018-12-23
Add recipe with-os

EricCrosson:

Provide a `use-package`-esque way to conditionally execute sexp's by current operating system.

https://github.com/EricCrosson/with-os

- [X] maintainer
- [X] contributor

None needed

### Checklist

Please confirm with `x`:

### Notes

It's a rough-but-functional initial version.  More experienced elispers might gasp at the contents.  I'm all ears if anybody is kind enough to offer feedback.

Cheers and thanks for all your help, melpa crew!

### DIFF
--- a/recipes/with-os
+++ b/recipes/with-os
@@ -1,0 +1,1 @@
+(with-os :fetcher github :repo "EricCrosson/with-os")


purcell:
I think `use-package` supports platform tests, doesn't it? Were you just aiming to provide an alternative?

As for the code, I think it should be possible to expand the whole thing into a macro: it's kinda strange in the current code that the expansion of `with-os` would simply contain a call to `with-os-process`. It should be possible to simply expand it out into a series of `(when (memq system-type '(linux)) ...)` forms.

(Relatedly, I think perhaps you should support consecutive `:platform` keys so that blocks could be applied to multiple OSes without duplication.)

EricCrosson:
> I think use-package supports platform tests, doesn't it? Were you just aiming to provide an alternative?

Look at that, it sure does! This is news to me, thanks for the heads-up. 
This is a great start, but not all platform-specific code in my .emacs is currently an invocation to `use-package`, and as such I can see a separate use-case for this package.

> As for the code, I think it should be possible to expand the whole thing into a macro

I agree that the entire package could be condensed down to a single macro, it'll be a good exercise for me to get the code to that state.  First I should nail-down the invocation form, as you brought up a great point with your next sentence:

> (Relatedly, I think perhaps you should support consecutive :platform keys so that blocks could be applied to multiple OSes without duplication.)

This is a great note, and is enough to suspend this pull-request for a moment.  To make sure I heard you correctly, are you suggesting that 

```lisp
(with-os
  :darwin
  :linux 
  (message "Hello from a unix-y system")
```

should execute the `message` sexp on both darwin and GNU/Linux systems? 

Would it then be reasonable to expect the following to work?

```lisp
(with-os
  :darwin
  :linux 
  (message "Hello from a unix-y system")
  :darwin
  (message "This part only runs on mac")
  :linux 
  (message "This part only runs on GNU/Linux")
```

EricCrosson:
I think I see what you're saying -- this should all be possible and cleaner with a macro expansion. I'll take a crack at the macro-based implementation. 

Not sure how you want to handle this partially-complete PR -- thanks again for your feedback so far

purcell:
> are you suggesting that

Yes!



> Not sure how you want to handle this partially-complete PR

Happy to keep this open for now. Ping here if you have any questions.

purcell:
Closing now, but feel free to re-open.

PR #5884, opened 2018-12-24
Add arrow

bradyt:

In this package I will attempt to accrue convenient sets of directional commands and bindings. Potentially I may look into quickly swapping out sets of directional actions onto directional keys, such as arrow keys plus modifiers, hjkl, etc.

https://github.com/bradyt/arrow

Author.

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/arrow
+++ b/recipes/arrow
@@ -1,0 +1,1 @@
+(arrow :fetcher github :repo "bradyt/arrow")


alphapapa:
>  I've used the latest version of package-lint to check for packaging issues, and addressed its feedback

bradyt:
> > I've used the latest version of package-lint to check for packaging issues, and addressed its feedback

I'm not certain what you're saying. Are you saying that `package-lint-current-buffer` shows issues for this buffer?

Are you looking at https://github.com/bradyt/arrow/commit/69e8c7a4ecaac2ca69aece1e788d67af1975f00d? I haven't seen any package-lint errors since the second commit.

After `M-x package-lint-current-buffer RET`, I see:

> No issues found

bradyt:
@alphapapa Can you help me understand which commits have problems? I was able to get another compiler error, and fix it, but not a linting error. I haven't seen a linting error since the second commit.

alphapapa:
Well, the first thing I see is that this is what MELPA users will see for your package's description:

> ;; error: Package should have a non-empty ;;; Commentary section.

bradyt:
@alphapapa Apologies for trying to race through that process with the commentary header, and forgetting to update it. I hope I have covered all the points now, in current head.

alphapapa:
Ok, the next thing I see is the description in the first line:

> All my arrow related functions and bindings 

That's not very helpful or descriptive.  It sounds a lot like, "I copied the keybindings from my init file and put them on MELPA."  MELPA generally discourages packages like that.

I see in the full description that you say:

> The package will start out assuredly simple, but we hope to find better ways to quickly assign and swap keys and commands, bind and unbind keys, as the project develops.

Nice use of the "royal we."  ;)  But seriously, I think you should consider waiting until the package develops further before publishing it to MELPA.

Other misc. issues:

* You load `evil` when `arrow` is loaded.  That generally shouldn't happen unless `arrow` is a package only for EVIL users.
* Instead of using `byte-compile-warnings: (not free-vars)`, you can use `defvar`, which is generally the "proper" way.  It's mentioned in the manual somewhere IIRC, maybe where it talks about byte-compiler warnings.

bradyt:
> [...] But seriously, I think you should consider waiting until the package develops further before publishing it to MELPA.

This is fine, I have no issue with waiting until then. We are hopeful I stick with it to such point. ;)

> Other misc. issues:
> 
> * You load `evil` when `arrow` is loaded.  That generally shouldn't happen unless `arrow` is a package only for EVIL users.

The require won't cause failure if evil is not present. I don't see how that block negatively effects evil nor non-evil users.

> * Instead of using `byte-compile-warnings: (not free-vars)`, you can use `defvar`, which is generally the "proper" way.  It's mentioned in the manual somewhere IIRC, maybe where it talks about byte-compiler warnings.

It was not obvious to me why either of these solutions is the ideal. I will have to research. I think I've read someone specifically say that part of manual is not quite right. Maybe to do with lexical being more common now.

alphapapa:
Thanks.  I look forward to seeing how your package develops.  It does seem like the arrow keys are ripe for assigning to unique commands, maybe mode-dependent.  I hadn't thought about it until I saw `arrow`.

PR #5885, opened 2018-12-24
Add recipe for file-header

jcs090218:

Highly customizable self design file header.

https://github.com/jcs090218/file-header

I am the maintainer.

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/file-header
+++ b/recipes/file-header
@@ -1,0 +1,1 @@
+(file-header :repo "jcs090218/file-header" :fetcher github)


alphapapa:
How does this package compare with other tools that do the same thing, e.g. `yasnippet`, `snippet`, `yankpad`, `yatemplate`, `dollaro`, `lice`, `muban`, `skeletor`, as well as built-in Emacs skeletons (see `Info node(autotype) Autotyping`)?

jcs090218:
> How does this package compare with other tools that do the same thing, e.g. `yasnippet`, `snippet`, `yankpad`, `yatemplate`, `dollaro`, `lice`, `muban`, `skeletor`, as well as built-in Emacs tools like skeletons, Tempo, etc. (see `Info node (autotype)Autotyping`)?

Cool, I do not use most of the packages you mentioned, hence evidently that I would not know the differences between those packages and this package. If any of these package could do the same thing I would not be surprised and I will definitely willing to close this PR. Anyway, here are some points that I reckon this package could be unique compare to other packages...
* No need to restart Emacs to take effect from the template. 
* You would not need to edit your Emacs' configuration, you would just need to edit the template file. Sometimes I miss type something in the config file and then when I restart Emacs nothing works, I made this does it this way benefit me make fewer mistakes. 
* This package only does this task, which I think this could benefit people who only want the file header and not any other unrelated functionalities. 
* Highly customizable, you can assign property value to a lisp function. For instance, get-time, you would just need to design a function that returns a string of the time, and this package will do the rest of the work for you. You could even define your own property name, in addition. For any file, any programming language, for anything.

jcs090218:
Okay, I made a little research on `yasnippet` and `auto-insert-mode`. It seems like by combining the two functionalities/packages could achieve pretty much the same thing. And still this is more likeliest the preferences issue, the way `yasnippet` + `auto-insert-mode` manage those templates file are different than this package. Is just like `Emacs` v.s. `Vim`, the editor war. Both are just editor but with just different religion because people have their own preferences. Leading from this, should people stop making new editor because we should all use what they already grant? I would always prefer having the other alternatives rather than just one absolute answer unless they are identical serving the same results and the same preferences. By the way, I personally think `yasnippet` + `auto-insert-mode` are just too laborious for a simple task like to make a file header. And that is way I made this package because I think this brings benefits to people. Well, I would still respect what `melpa`'s author thinks since I am attempting to release my package here.

alphapapa:
> Is just like Emacs v.s. Vim, the editor war. Both are just editor but with just different religion because people have their own preferences. Leading from this, should people stop making new editor because we should all use what they already grant? I would always prefer having the other alternatives rather than just one absolute answer unless they are identical serving the same results and the same preferences.

Of course, you should make whatever packages you want.  But MELPA is a repository of somewhat curated packages.  When there are already at least 10 different packages that do this, including some included with Emacs, it's reasonable to ask what makes your package uniquely valuable before adding it to MELPA.

Having choices is valuable, but there is an associated cost: every time a user looks for a package that does a certain thing, the more choices there are, the more difficult and time-consuming it is to choose one, and the less likely it is the user will find the best one.

I heard an interesting episode of the podcast "Freakonomics" recently: one of the things they mentioned was a study on how people make choices depending on how many are available.  They looked at a free samples table in a grocery store that handed out coupons with the samples.  IIRC, they found that, when there are 3-4 samples, people were much more likely to try a sample and use the coupon to buy one, than when there were tens of samples.

Of course, MELPA isn't a grocery store, and we're not selling anything, and we value choice highly.  But a lot of MELPA users are not experts, and no one, not even the MELPA maintainers, knows the details of every package in the archive.  So when an Emacs user goes looking for a package to do a certain thing, he only has so much time and energy to put into the task, and I think it's reasonable to think about that common experience when adding packages to MELPA.

I don't want to discourage you from writing and publishing your packages.  In this specific case, I question whether this package is uniquely valuable enough, compared to the 10-12 packages that already do the basic task of inserting templates filled with data, to justify adding another choice to MELPA.  There are other ways and places to share and publish your work besides MELPA.  For example, I publish smaller portions of code that aren't enough to justify a package to https://github.com/alphapapa/unpackaged.el, and I have some actual packages on my GitHub account that aren't on MELPA.

Another thing to consider is, what if the existing packages could be improved instead, so that both new and existing users would benefit?  Is there a similar package already on MELPA that you could submit improvements to, rather than publishing a new one?  You said that you don't use them, so I guess you haven't investigated that possibility yet.  I encourage you to consider that, because I know that people love getting PRs to their projects.  Collaboration is fun!

Anyway, I have no objection to adding packages that do nearly anything.  For me, the question comes when adding a package that is very similar to a number of other packages which do virtually the same thing.  If it's significantly better or very different, sure.  But if they only do it in a slightly different way, is it worth adding another choice that users will have to dig through?  

Well, it's not my decision to make, because I'm not a MELPA maintainer.  I just try to help them out by pointing out some potential issues on PRs before they get to them.

jcs090218:
I guess I will just wait for them to review, I am totally fine if they feel this package isn't worthy enough to release on their platform.

purcell:
Agreed, `auto-insert` is the way to go here. This package is not a novel approach, and where possible it's better for the community if we use (and collectively improve) the same tools when possible.

PR #5888, opened 2018-12-27
Replace ':defaults' with '":defaults"' at create recipes json file.

conao3:
This pull request is not a package addition, but a proposal to change MELPA's core file.

I have created a new package manager named [feather.el](https://github.com/conao3/feather.el), and I'm using MELPA's recipe as its recipe. ([feather-recipes](https://github.com/conao3/feather-recipes))

In the process, I found mysterious points in the [JSON file](https://melpa.org/recipes.json) hosted by MELPA and I will propose a fix.

### fix point
Before, `json-encode` mistakenly recognized `:defualts` as a key.
```
(json-encode '(((dyalog-mode :fetcher bitbucket :repo "harsman/dyalog-mode" :files (":defaults" "Emacs.apl")))))
=> [{"dyalog-mode":{"fetcher":"bitbucket","repo":"harsman/dyalog-mode","files":{"defaults":"Emacs.apl"}}}]
```

After, by replacing `:defaults` with `":defaults"`, `json-encode` is now able to recognize it correctly as array.
```
(json-encode '(((dyalog-mode :fetcher bitbucket :repo "harsman/dyalog-mode" :files (":defaults" "Emacs.apl")))))
=> [{"dyalog-mode":{"fetcher":"bitbucket","repo":"harsman/dyalog-mode","files":["//:defaults//","Emacs.apl"]}}]
```

Replaced with `//:defaults//`  because it does not conflict with the user's file, and I will change it if other strings are recommended.

### DIFF
--- a/package-build/package-build.el
+++ b/package-build/package-build.el
@@ -940,15 +940,14 @@ artifacts, and return a list of the up-to-date archive entries."
          (lambda (lst)
            (let (frg)
              (mapcar (lambda (x)
-                       (cond ((eq x :files) (prog1 x (setq frg t)))
-                             (frg (prog1 x
-                                    (setq frg nil)
-                                    (cl-mapcan
-                                     (lambda (y)
-                                       (if (listp y)
-                                           (if (eq (car y) :exclude)
-                                               `((:exclude ,(cdr y))) y)
-                                         (list y))))))
+                       (cond ((eq x :files) (setq frg t) x)
+                             (frg (setq frg nil)
+                                  (cl-mapcan
+                                   (lambda (y)
+                                     (if (listp y)
+                                         (if (eq (car y) :exclude)
+                                             `((:exclude ,(cdr y))) y)
+                                       (list y))) x))
                              (t x))) lst)))))
     (with-temp-file file
       (insert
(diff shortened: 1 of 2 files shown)


purcell:
Thanks for this. I'll have a think about the `package-build.el` changes -- that file comes from a separate repo, which is where the fix would need to be made, if we decide it's needed. The `slime` recipe fixes probably make sense anyway: I'll see if I can cherry-pick those for now.

conao3:
OK, thanks!
I have not been able to accurately grasp the extent of this change, so I think that decision is a matter of course.
But in developing my package, this change is useful, so I'd like to be merge.

Thank you as always!

conao3:
Fix conflict.

purcell:
Here I'm wondering if the ultimate solution is to switch from the keyword `:defaults` to the symbol `'defaults`...

conao3:
Thank you for your reply!

The seemingly strange points are often understood as historical reasons and are required to be covered in operation.
It is too late to change the recipe format, so I think that changing keywords into symbols just before outputting a JSON file is appropriate even considering the amount of work and the result.

It's not a difficult job to convert only the recipe by script, but I think it is difficult to evaluate the impact it has.

conao3:
Sorry, I remove the branch with a mistake. Restore it and reopen PR.

tarsius:
See https://github.com/melpa/package-build/pull/33 for an alternative.

conao3:
Good. Thanks 🎉

PR #5889, opened 2018-12-27
Recipe: Add scrollkeeper

alphapapa:

This package provides scrolling commands and several customization options.  The commands use `pulse` to display a quickly fading guideline, or another chosen style, on the line at which new contents are visible after scrolling.  Also, scrolling can be divided into steps which proceed at the desired speed.  Together, these features help your eyes to keep their place in the buffer while scrolling.

https://github.com/alphapapa/scrollkeeper.el

Author, maintainer, enthusiastic user.

I consult with myself as needed.

### Checklist

Please confirm with `x`:

Thanks for your work on MELPA!

### DIFF
--- a/recipes/scrollkeeper
+++ b/recipes/scrollkeeper
@@ -1,0 +1,1 @@
+(scrollkeeper :fetcher github :repo "alphapapa/scrollkeeper.el")


alphapapa:
> What is the difference between this and https://github.com/Malabarba/beacon ?

Please see the readme file, video demonstration, etc.

Silex:
Ah sorry, missed the README section.

purcell:
Thanks for this. Quick notes, then:
- You'll need a package dependency on `cl-lib`, or on Emacs 24.3.
- When was `pulse.el` added to Emacs? I suspect you might need to depend on a newer Emacs version for that one.
- The faces are a bit Wild West IMO: a better default for the highlighting would be to `:inherit` the `highlight` face. Using the foreground from one face as the background for another is unlikely to work well across themes by default.
- The `thinline` variant won't work in TTY frames, and would always fractionally move the content below it: I wonder if you should stick with the underline instead.
- I really like the `cl-typecase` snippet.

alphapapa:
> You'll need a package dependency on cl-lib, or on Emacs 24.3.

I always forget about that.  I'll just target 25.1.  [*Done.*]

> When was pulse.el added to Emacs? I suspect you might need to depend on a newer Emacs version for that one.

Looks like [2009](https://github.com/emacs-mirror/emacs/commits/master/lisp/cedet/pulse.el) actually.  :)

> The faces are a bit Wild West IMO: a better default for the highlighting would be to :inherit the highlight face. Using the foreground from one face as the background for another is unlikely to work well across themes by default.

Well, the `font-lock-string-face` foreground color is virtually guaranteed to be visible against any theme's background (the theme would be badly broken otherwise), but the problem with doing it this way is that, if the user changes theme, the face won't update its inherited color automatically.

The problem with inheriting from `highlight` is that it is the same color used by, e.g. `hl-line-mode`.  Of course, some users might prefer that, but it may not be a good default, because it can make the line hard to distinguish from the current-line highlight.

So I think I'd like to leave it as-is for now, and hopefully find a better solution for the default in the future.

> The thinline variant won't work in TTY frames, and would always fractionally move the content below it: I wonder if you should stick with the underline instead.

Yes, the thinline does always slightly move the line below.  I haven't been able to find a way to make the line thinner (I wish I could make it 1 pixel high).  It's not the default method, but I left it in there in case some users might prefer it, and in case there's a way to improve it.

Thanks for all your feedback, Steve, as usual!

purcell:
>  but the problem with doing it this way is that, if the user changes theme, the face won't update its inherited color automatically.

Yes, this is the *really* bad thing here, and should really be fixed IMO. Tracking down these sorts of appearance quirks is the worst! If it's a momentary flash, why not just make the default `:inverse-video t`?

In the meantime I'll merge this. Woo!

alphapapa:
> Yes, this is the really bad thing here, and should really be fixed IMO. Tracking down these sorts of appearance quirks is the worst!

That's a good point, I wouldn't want to cause someone to waste their time trying to figure it out.

> If it's a momentary flash, why not just make the default :inverse-video t?

That's a good idea, I'll try that.

Thanks.

alphapapa:
Unfortunately `:inverse-video` has no effect on the `pulse` face, but I found what seems to be a good solution: https://github.com/alphapapa/scrollkeeper.el/commit/fba677f853b07d6354b092bfd31a3b74a0e1f0c6  Thanks!

PR #5890, opened 2018-12-28
Add recipe: use-package-hydra

To1ne:

`use-package-hydra` is the glue between use-package and hydra. It adds
the `:hydra` keyword support to the `(use-package)` macro.

https://gitlab.com/to1ne/use-package-hydra

I'm the original author of the package.

None needed. 

### Checklist

Please confirm with `x`:

#### package-lint

2 issues I could not fix cause of the naming conventions in `use-package` core.

```
79:1: error: "use-package-handler/:hydra" doesn't start with package's prefix "use-package-hydra".
79:1: error: `use-package-handler/:hydra' contains a non-standard separator `/', use hyphens instead (see Elisp Coding Conventions).
```

### DIFF
--- a/recipes/use-package-hydra
+++ b/recipes/use-package-hydra
@@ -1,3 +1,2 @@
 (use-package-hydra :repo "to1ne/use-package-hydra"
-                    :fetcher gitlab
-                    :files ("use-package-hydra.el"))
+                    :fetcher gitlab)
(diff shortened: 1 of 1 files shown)


purcell:
Looks good to me. I'll fix up the recipe to remove `:files`, which isn't needed here, then go ahead and merge it. Welcome to MELPA! :tada:

PR #5891, opened 2018-12-28
add exunit recipe

ananthakumaran:

A minor mode that provides commands to run [ExUnit](https://hexdocs.pm/ex_unit/ExUnit.html) tests. The output is syntax highlighted and stacktraces are navigatable.

https://github.com/ananthakumaran/exunit.el

maintainer

**None needed** 

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/exunit
+++ b/recipes/exunit
@@ -1,0 +1,1 @@
+(exunit :fetcher github :repo "ananthakumaran/exunit.el")


alphapapa:
Hi,

I don't use Elixir, so I can't offer much feedback from that perspective, but I see a lot of functions without docstrings.  Would you mind adding some?  They help other users to understand how your code works and are a significant aid when someone wants to modify it.

ananthakumaran:
Added some more docs, I hope the remaining functions are self-explanatory.

alphapapa:
Curious, why use two lists like this?

```el
(defvar exunit-compilation-error-regexp-alist-alist
  '((elixir-warning "warning: [^\n]*\n +\\([0-9A-Za-z@_./:-]+\\.exs?\\):\\([0-9]+\\)" 1 2 nil 1 1)
    (elixir-error " +\\(\\(?:([0-9A-Za-z_-]*) \\)?[0-9A-Za-z@_./:-]+\\.\\(?:ex\\|exs\\|erl\\)\\):\\([0-9]+\\):?" 1 2 nil 2 1)))
(defvar exunit-compilation-error-regexp-alist
  (mapcar 'car exunit-compilation-error-regexp-alist-alist))
```

ananthakumaran:
```elisp
(defcustom compilation-error-regexp-alist
  (mapcar 'car compilation-error-regexp-alist-alist)
  "Alist that specifies how to match errors in compiler output.
On GNU and Unix, any string is a valid filename, so these
matchers must make some common sense assumptions, which catch
normal cases.  A shorter list will be lighter on resource usage.

Instead of an alist element, you can use a symbol, which is
looked up in `compilation-error-regexp-alist-alist'.  You can see
the predefined symbols and their effects in the file
`etc/compilation.txt' (linked below if you are customizing this).
```
`define-compilation-mode` is doing all the wiring up here

alphapapa:
Yep, I looked that up right after I clicked "Comment," so I deleted mine.  :)  Thanks.

purcell:
Thanks. Quick notes:

- Since you're depending on Emacs 24.3, you can wrap `defvar-local` [here](https://github.com/ananthakumaran/exunit.el/blob/516e71ba6a9e74e64f2a82c124f3a8ffa16743b7/exunit.el#L54-L55). But in any case, you only use it to declare one variable, so just use `defvar-local` plus the  `put` for `permanent-local`.
- The value of a `setq` expression is the variable's new value, so you can delete [this line](https://github.com/ananthakumaran/exunit.el/blob/516e71ba6a9e74e64f2a82c124f3a8ffa16743b7/exunit.el#L71).
- Use `expand-file-name` & co instead of splicing paths together with `/` - it's more portable. When splicing multiple parts, you can either use `f-join` (from the `f` package) or write your own trivial helper using `cl-reduce` and `expand-file-name`:
```el
(defun my-join-paths (&rest args)
  (cl-reduce (lambda (a b) (expand-file-name b a)) args))

(my-join-paths "/some/dir" ".." "subdir") => "/some/subdir"
```
- `exunit-mode` appears unused, and should be removed if so.

Hope that helps!

ananthakumaran:
Thanks for the review, updated now

purcell:
Great, merged - thanks! :tada:

PR #5894, opened 2018-12-30
Remove obsolete lsp-clangd and lsp-python

tarsius:
These packages are no longer needed when using `lsp-mode`'s `master` branch and they also fail to load because they require `lsp-common`, which no longer exists. However when installing from Melpa Stable, then they are probably still required, so we can't really remove them yet.

Maybe there should be a way to remove a package from Melpa but keep it in Melpa Stable?

### DIFF
--- a/recipes/lsp-clangd
+++ b/recipes/lsp-clangd
@@ -1,3 +1,0 @@
-(lsp-clangd
- :fetcher github
- :repo "emacs-lsp/lsp-clangd")
--- a/recipes/lsp-python
+++ b/recipes/lsp-python
@@ -1,1 +1,0 @@
-(lsp-python :repo "emacs-lsp/lsp-python" :fetcher github)


tarsius:
- https://github.com/emacs-lsp/lsp-clangd
- https://github.com/emacs-lsp/lsp-python/issues/33

yyoncho:
You may remove also: lsp-javascript, lsp-ruby, lsp-go, lsp-rust, lsp-php, lsp-sh, lsp-html, lsp-groovy, lsp-ocaml .

tarsius:
@yyoncho We can do that once that doesn't cause breakage for users installing `lsp-mode` from Melpa Stable. When do you plan to release a version of `lsp-mode` that contains all the merged formerly-separate language specific libraries?

Also, why is this being done? The general trend for Emacs packages seems to be to break out such additional libraries into separate packages or even repositories. Why are you doing the opposite? (Not saying it is wrong, but such changes cause the Melpa maintainers work so it would be nice to know why this is being done.

@Trevoke Re #5903. You are going with the Emacs trend and against the `lsp-*` trend.

yyoncho:
@tarsius
I am planning to address one or two issues in this release. I hope to do the release by the end of next week.

> Also, why is this being done? 

A couple of things:
1. We wanted to decrease the configuration complexity based on the user feedback
2. In latest lsp-mode some of the configuration that used to be part of the packages is now provided by lsp-mode in a generic way. As a result, some of the packages were diminished to a single call.
3. It was easier to migrate the packages to the new interface. There was a period in which both lsp-mode interfaces were supported.

Note that the more complex lsp extensions were not absorbed(like lsp-java, cquery) and

dmacvicar:
@yyoncho in #5972 my list is a bit longer than yours. Am I missing something?

* lsp-clangd
* lsp-css
* lsp-dart
* lsp-fortran
* lsp-go
* lsp-html
* lsp-javascript-typescript
* lsp-ocaml
* lsp-php
* lsp-python
* lsp-ruby
* lsp-rust
* lsp-sh
* lsp-typescript
* lsp-vue

yyoncho:
@dmacvicar your list seems to be ok

tarsius:
@yyoncho I see `lsp-mode` v6.0 has been released. Does that mean that all those `lsp-<language>` packages are obsolete now *even when using only Melpa Stable* and that the respective recipes can therefore be removed?

yyoncho:
@tarsius  Yes.

Trevoke:
Such rapid development indicates that maybe the version number for lsp-mode would be below 1.0 if following semver?

As for the emacs/lsp practice: We are interested in providing an Elixir-specific tool. There are some code actions that don't make sense in other languages, and I would like to also provide the servers and a way to choose (different servers for different versions of Erlang).

yyoncho:
@Trevoke 

> Such rapid development indicates that maybe the version number for lsp-mode would be below 1.0 if following semver?

When I became the maintainer the version was 5.0 and I did one big redesign so the only option was to use 6.0 but I guess you are right. At this point, I am considering the API to be stable.

As for the elixir tool I guess it is related to https://github.com/elixir-lsp/lsp-elixir.el, right ? You may ping me on gitter so we can discuss what you are trying to achieve(or move the discussion in the corresponding repo). If I had understood you correctly you may create a language server which runs in parallel with the ElixirLS and provides only code actions.

tarsius:
There recipes and more were removed with #5972.

PR #5897, opened 2019-01-03
Add recipe for the Markless major mode package

Shinmera:

This package implements a major made for the [Markless](https://shirakumo.github.io/markless/) document markup standard.

https://github.com/shirakumo/markless.el

I am the maintainer.

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/markless
+++ b/recipes/markless
@@ -1,0 +1,2 @@
+(markless :repo "shirakumo/markless.el"
+          :fetcher github)


Shinmera:
Thanks for the review! The mentioned issues should be addressed now.

purcell:
Thanks, looks good. Your [auto-mode-alist regexp](https://github.com/Shirakumo/markless.el/blob/master/markless.el#L387) should end with `\\'`, so that it does not match filenames like `*.messy`. Merging now in any case.

PR #5899, opened 2019-01-04
update literate-elisp recipe to include org file as source.

jingtaozf:
a literate programming tool to load emacs lisp codes from org mode file directly

https://github.com/jingtaozf/literate-elisp

author/maintainer

exist package

### Checklist

Please confirm with `x`:

- [X ] The package is released under a [GPL-Compatible Free Software License](https://www.gnu.org/licenses/license-list.en.html#GPLCompatibleLicenses).
- [ X] I've read [CONTRIBUTING.org](https://github.com/melpa/melpa/blob/master/CONTRIBUTING.org)
- [ X] I've used the latest version of [package-lint](https://github.com/purcell/package-lint) to check for packaging issues, and addressed its feedback
- [ X] I've built and installed the package using the instructions in [CONTRIBUTING.org](https://github.com/melpa/melpa/blob/master/CONTRIBUTING.org)

### DIFF
--- a/recipes/literate-elisp
+++ b/recipes/literate-elisp
@@ -1,1 +1,1 @@
-(literate-elisp :fetcher github :repo "jingtaozf/literate-elisp")
+(literate-elisp :fetcher github :repo "jingtaozf/literate-elisp" :files ("*.el" "*.org"))


purcell:
> Maybe it is not necessary to include document files.

Correct, you would normally not include them. Thanks.

PR #5901, opened 2019-01-04
Add recipe for emacs-reveal

lechten:

The software emacs-reveal allows to create HTML presentations (with audio explanations if you wish) with reveal.js from Org mode files in GNU Emacs with a forked version of org-reveal and several reveal.js plugins.  Generated presentations are usable with standard browsers, also mobile and offline.

https://gitlab.com/oer/emacs-reveal

I am the maintainer/creator.

None needed.

### Checklist

Please confirm with `x`:

  It tells me that `org-ref` is not installable, which is incorrect. Also, it claims that I need Emacs 25, which I don't need as explained in comments underneath my `Package-Requires` header.
  It suggests to capitalize "emacs" as part of a shell command. Is there markup to prevent this?

### DIFF
--- a/recipes/emacs-reveal
+++ b/recipes/emacs-reveal
@@ -1,0 +1,5 @@
+(emacs-reveal
+ :repo "oer/emacs-reveal"
+ :branch "melpa"
+ :fetcher gitlab
+ :files (:defaults "org-reveal" "org" "css" "title-slide" "reveal.js" "reveal.js-plugins" "Reveal.js-TOC-Progress" "reveal.js-jump-plugin" "reveal.js-quiz"))
--- a/recipes/emacs-reveal
+++ b/recipes/emacs-reveal
@@ -2,4 +2,4 @@
  :repo "oer/emacs-reveal"
  :branch "melpa"
  :fetcher gitlab
- :files (:defaults "org-reveal" "org" "css" "title-slide" "reveal.js" "reveal.js-plugins" "Reveal.js-TOC-Progress" "reveal.js-jump-plugin" "reveal.js-quiz"))
+ :files (:defaults "README*" "LICENSE*" "org-reveal" "org" "css" "title-slide" "reveal.js" "reveal.js-plugins" "Reveal.js-TOC-Progress" "reveal.js-jump-plugin" "reveal.js-quiz"))


purcell:
Thanks for taking the time to submit this, but this isn't really a well-formed package.

You're using an arbitrarily-named package here to deliver forked versions of files from elsewhere, but using the same names. If this package was installed along with `org-reveal`, which `org-reveal.el` would be loaded? -- it's unpredictable. That's bad, because all packages should be able to co-exist with each other without interfering, just like in the Emacs standard library.

If this package provides something new and novel, it should have its own uniquely-named `.el` file, and the package name should match it. Alternatively, if this is a bundling of improvements to `org-reveal` and/or `reveal.js`, those improvements should be submitted upstream. I'll therefore close this pull request for now, but please feel free to ask questions here if the above isn't clear.

lechten:
Thanks for your review. I understand your comments.

I must admit that I didn't think properly about the conflict with org-reveal. However, that conflict does not seem overly important to me as org-reveal cannot be installed with recent versions of Org any more: https://github.com/yjwen/org-reveal/issues/342

In general, development of org-reveal seems to have stalled. I tried to contribute but did not receive a response from the maintainer:

- https://github.com/yjwen/org-reveal/pull/277
- https://github.com/yjwen/org-reveal/issues/349

So, no, my package does not try to provide something novel compared to org-reveal but an improved version that is still maintained. I went for inclusion on MELPA in response to this issue: https://gitlab.com/oer/emacs-reveal/issues/5

I'd appreciate any advice given the current situation.

gahag:
I believe that, although `org-reveal`'s development is stalled, it still owns the package name, as the development might as well be resumed in the future. I suggest we use a new name for this one, such as `org-revealjs` or anything.

purcell:
> I believe that, although `org-reveal`'s development is stalled, it still owns the package name, as the development might as well be resumed in the future. I suggest we use a new name for this one, such as `org-revealjs` or anything.

Yes, this would be acceptable, however the _best_ solution for everyone would be if @yjwen would be willing to make @lechten (and possibly others) an `org-reveal` co-maintainer -- there are a lot of outstanding pull requests that could be very beneficial to users if someone could go through and handle them. What do you think, @yjwen?

gahag:
I think we should set a deadline for this, as @yjwen is apparently inactive at github for half a year. I suggest we move on with a new package by next monday, making up a week of time for @yjwen to reply before we disconsider his participation.

gahag:
So guys, should we agree to move on with this?

lechten:
This is my suggestion: I create a new package `org-revealjs` (later this week).  That would include my fork of `org-reveal` and also code from `emacs-reveal.el`.

@purcell  I'm not sure about naming.  Files of Org export modules have names starting with "ox-" but their contents are prefixed with "org-" (e.g., org-html-publish-to-html defined in ox-html.el).  Should I stick with that Org convention or go for the usual package convention?

Once that package is settled, we might discuss whether `emacs-reveal` is suitable for MELPA.  I guess that it would contain only little elisp configuration.  The bulk would be reveal.js with plugins.

lechten:
I just opened a pull request for org-revealjs: https://github.com/melpa/melpa/pull/5979

I decided against moving code from emacs-reveal.el to the new package (as Org macros are used to invoke that additional functionality, which is different from ordinary Org export). If org-revealjs gets accepted, I'll update the melpa branch of emacs-reveal and ask again, whether that package can be included.

PR #5906, opened 2019-01-05
fix ac-math recipe

conao3:
I am developing a new package manager named [feather.el](https://github.com/conao3/feather.el), and I'm using MELPA's recipe as its recipe. ([feather-recipes](https://github.com/conao3/feather-recipes))

In the process, I check all the recipes especially for the `:file` option.

If delete specify detailed specifications for the recipe, the recipe JSON file will be smaller and reduce internal process, I think that my package will run faster.
If there is no problem, please merge it.

https://github.com/vspinu/ac-math

volunteer

### Checklist
Please confirm with `x`:

*(As this PR is a modification of existing recipe, check all the checklists below)*

### DIFF
--- a/recipes/ac-math
+++ b/recipes/ac-math
@@ -1,1 +1,1 @@
-(ac-math :repo "vspinu/ac-math" :fetcher github :files ("ac-math.el"))
+(ac-math :fetcher github :repo "vspinu/ac-math")


purcell:
Thanks @conao3 -- this is potentially _very_ helpful, but please can you share your methodology here? I don't want to just blindly merge all these (40+?) pull requests without understanding how you determined which `:files` specs are redundant.

conao3:
Thanks. Those PRs are all manual.
By Elisp below I found a suspicious recipe that file designation does not conform to MELPA rules and actually visited the repository.
I think that this is enough because (the very old) recipe that does not conform to the rule seems to be less than 100 at most.
```
(po (mapcan (lambda (x)
              (let ((val (cdr x)))
                (when (plist-get val :files)
                  `(:package
                    ,(if (string= (plist-get val :fetcher) "git")
                         (plist-get val :url)
                       (format "https://%s.com/%s" (plist-get val :fetcher) (plist-get val :repo)))
                    ,(plist-get val :files)))))
            (let ((escape-fn (lambda (find-str replace-str)
                               (goto-char (point-min))
                               (while (search-forward find-str nil t)
                                 (replace-match replace-str nil t))))
                  (escape-:exclude-fn
                   (lambda (lst)
                     (let ((frg))
                       (mapcar (lambda (x)
                                 (cond ((eq x :files) (setq frg t) x)
                                       (frg (setq frg nil)
                                            (mapcan
                                             (lambda (y)
                                               (if (listp y)
                                                   (if (eq (car y) :exclude)
                                                       `((:exclude ,(cdr y))) y)
                                                 (list y)))
                                             x))
                                       (t x)))
                               lst)))))
              (cl-mapcan
               (lambda (name)
                 (condition-case nil
                     ;; Filter out invalid recipes.
                     (when (with-demoted-errors (package-recipe-lookup name))
                       (with-temp-buffer
                         (insert-file-contents
                          (expand-file-name name package-build-recipes-dir))
                         (funcall escape-fn ":defaults" "\"//:defaults//\"")
                         (list (funcall escape-:exclude-fn
                                        (read (buffer-string))))))))
               (package-recipe-recipes)))))
```

conao3:
Theoretically: The `:file` keyword should be eliminated by the Elisp community's cooperation, which should make the recipe easier and contribute to the weight saving of JSON file.

alphapapa:
> If delete specify detailed specifications for the recipe, the recipe JSON file will be smaller and reduce internal process, I think that my package will run faster.

Have you benchmarked your code to determine whether removing one key/value from the JSON actually makes a difference in your code's performance?

conao3:
I'm sorry I haven't benchmarked not yet.
`feather.el` will download [this file](https://github.com/conao3/feather-recipes/blob/master/recipes/melpa.el) at *Emacs startup* if necessary. If the `:file` keyword value becomes `nil` and the file size gets smaller, `feather.el` can move on to the next task faster.
In Emacs startup, it is desirable (for some users) to speed up microsecond order.

purcell:
Thanks! Thinking about it, the MELPA build actually logs all `:files` specs that seem to be redundant - see [here](https://github.com/melpa/package-build/blob/master/package-build.el#L752). This might be the most reliable way to handle this. It would require cross-referencing the MELPA and MELPA Stable builds, though, because those the repository contents of the versions used for MELPA Stable packages might be very different to the contents used for the regular MELPA packages.

purcell:
To clarify, what I'm proposing is that we look at the MELPA & MELPA Stable build logs and look for packages named in the `:files spec is equivalent to the default` messages. If such a package is absent from MELPA Stable, or if the same message is printed there, then we can remove the build message.

(Somewhat related: I'd also quite like to store the latest build log for each package alongside the package itself, so that it can be displayed via the MELPA web front end.)

alphapapa:
@conao3 Where does https://github.com/conao3/feather-recipes/blob/master/recipes/melpa.el come from?  Are you generating it yourself?

conao3:
@purcell 
Thanks! It sounds nice.
I'm sorry that I did not consider MELPA stable.

And I understand MELPA and MELPA stable build log is the most trusted log.
I think that it will be a help for developers if they can see the log and notice from MELPA web (or another page).

@alphapapa 
Generate this file from `feather-recipes.el` on repository root and [melpa fork](https://github.com/conao3/feather-melpa) every day.

alphapapa:
> Generate this file from `feather-recipes.el` on repository root and [melpa fork](https://github.com/conao3/feather-melpa) every day.

Since you're generating that file yourself, why don't you fix the `:files` in `feather-recipes.el`?

conao3:
I was only thinking about modifying the recipe, not just only my script, but share with the community, was it bad?

alphapapa:
> I was only thinking about modifying the recipe, not just only my script, but share with the community, was it bad?

No, what I mean is, you might be able to fix it on your end first, without modifying anything in other projects.

purcell:
I've closed these pull requests for now, but I appreciate you taking the time to submit them. I will see what I can do (as time allows) to look at the build logs and remove unnecessary `:files` specs from recipes where possible.

conao3:
Thanks, I'm sorry. I hope to contribute to the MELPA and Emacs community another way another time.

purcell:
I'm sure you will! Looking forward to seeing how feather.el works out.

PR #5908, opened 2019-01-05
Remove obsolete flycheck-purescript

marsam:
I no longer maintain this package, and it has grown obsolete

maintainer

None needed

### Checklist

Please confirm with `x`:


purcell:
Thanks, makes sense!

PR #5959, opened 2019-01-05
Add helm-lxc

montag451:

Helm interface to manage LXC containers

https://github.com/montag451/helm-lxc

I'm the maintainer

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/helm-lxc
+++ b/recipes/helm-lxc
@@ -1,0 +1,3 @@
+(helm-lxc
+ :repo "montag451/helm-lxc"
+ :fetcher github)


montag451:
@riscy Thanks for you feedback ! I have updated my repository accordingly. Regarding the license, the text in the source file states that:

> ;; This program is free software; you can redistribute it and/or modify
> ;; it under the terms of the GNU General Public License as published by
> ;; the Free Software Foundation, either version 3 of the License, or
> ;; (at your option) any later version.

So if I understand the text above (I may be wrong, I'm a total newbie regarding this license things :)) the license used is GPL 3 or later no ?

riscy:
You're completely right about the version being in there - sorry about that.  It's subjective, but I think it is still worth adding a copy of the license to your repository (a COPYING file).  After all, your commentary says your users should have received one. :)  It's a small thing, but there is also the additional advantage that GitHub will detect the license file and create an icon for people to look for:
image

purcell:
Thanks, and sorry for the slow turnaround.

One note I have is that you should not [define new faces with only `:foreground`](https://github.com/montag451/helm-lxc/blob/a4e17dda329ec39a3dac5751ddcef1145b3d91c1/helm-lxc.el#L53-L66) -- these will be guaranteed to look bad in many themes. Instead, `:inherit` from built-in faces like `success`, `warning`, `error` etc. by default.

But I'll go ahead and merge this in the meantime. :-)

montag451:
Don't worry about the "slow turnaround", the work you do for the Emacs community is very much appreciated.

Regarding the definition of faces, I will make the changes you requested as soon as possible. 

Thank you again!

PR #5960, opened 2019-01-05
Add flow-js2-mode

razzius:
Add flow-js2-mode

Major mode supporting flow type annotations in JavaScript.

https://github.com/Fuco1/flow-js2-mode

I'm a user

https://github.com/Fuco1/flow-js2-mode/issues/19

### Checklist

### DIFF
--- a/recipes/flow-js2-mode
+++ b/recipes/flow-js2-mode
@@ -1,0 +1,2 @@
+(flow-js2-mode :repo "Fuco1/flow-js2-mode" :fetcher github)
+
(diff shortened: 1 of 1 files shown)


razzius:
Whoops, opened this by mistake, but if it's no trouble I'll keep it open as I complete the remaining checkboxes above.

The remaining work here is to do some linting on the source code since not all the functions have docstrings and there seem to be unused variables.

purcell:
Yep, still some work to be done here upstream, e.g. adding a package dependency on js2-mode.

purcell:
I looked upstream and this seems to have been fixed, though nobody mentioned it here.

PR #5961, opened 2019-01-06
Add recipe for kfg

czipperz:
This essentially reverts commit bc88f9ce

### DIFF
--- a/recipes/kfg
+++ b/recipes/kfg
@@ -1,0 +1,2 @@
+(kfg :fetcher github
+     :repo "czipperz/kfg")


purcell:
The rationale for removing this was that the package has no maintainer, so we'd be wary of just adding back in the recipe which builds it from its temporary location in the `emacsattic` org. Would you like to take over maintainership? That would be as simple as forking the emacsattic version and submitting a MELPA recipe which builds from your fork.

czipperz:
K

purcell:
Thanks, and sorry for the slow turnaround.

PR #5962, opened 2019-01-06
Add org-d20 recipe

spwhitton:

Minor mode for for GMs running tabletop roleplaying games whose rules centre around rolling d20s

https://git.spwhitton.name/org-d20

Author and maintainer

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/org-d20
+++ b/recipes/org-d20
@@ -1,0 +1,1 @@
+(org-d20 :fetcher git :url "https://git.spwhitton.name/org-d20")


spwhitton:
Thank you for the interest and review.

The checkdoc output is just "All variables and subroutines might as well have a documentation string".  This is a recommendation, not an error -- is it really a blocker to inclusion in MELPA?  It strikes me as busywork.  These are all just helper functions which have comments explaining their purposes.

I cannot reproduce that bytecompiler output (Emacs 25.1.1), but I've added types to those defcustoms -- thank you.

riscy:
I'm only a contributor and not a MELPA maintainer, so I can't say whether it's a blocker.  I call them errors because that's what checkdoc categorized them as.

Even a lot of the packages that come with Emacs aren't rigorous about this, but I wouldn't necessarily call it "busy work".  For example, documenting an elisp function like this:
```elisp
;; Convert a signed integer to a string term
(defun org-d20--num-to-term (n)
  (let ((k (if (stringp n) (string-to-number n) n)))
    (if (>= k 0)
        (concat "+" (int-to-string k))
      (int-to-string k))))
```
is unorthodox, and it's easy to turn this into a helpful docstring that interfaces with the emacs help system.

purcell:
Looks good to me now - thanks for your patience. I'll go ahead and merge this.

PR #5963, opened 2019-01-06
text-buffer package

dkrivets:

Simple way to create new buffer does not think about it name.

https://github.com/dkrivets/text-buffer

I am

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/text-buffer
+++ b/recipes/text-buffer
@@ -1,0 +1,1 @@
+(text-buffer :repo "dkrivets/text-buffer" :fetcher github :version-regexp "v0.0.1" :files ("*.el"))


dkrivets:
> This functionality seems to be better covered by the `find-temp-file` package.
Another direction: find-temp-file stores temporary files, but the text buffer is just a buffer with a structured default name and it doesn't require to store file.

alphapapa:
* It would be a trivial addition to `find-temp-file` to, e.g. not write a file when called with a prefix argument, but to just open a buffer by that name.
* There is already a function to generate unique buffer names built-in to Emacs: `generate-new-buffer`.  This package's functionality essentially boils down to wrapping `generate-new-buffer` in a command like:
```el
(defun unique-buffer ()
  "Pop to new, uniquely named buffer."
  (interactive)
  (pop-to-buffer (generate-new-buffer "TEMP")))
```
* Your package is doing some very strange things, e.g. `text-buffer--optimize`'s byte-compiling files for no apparent reason, and using a global minor mode to enforce a key binding rather than simply letting the user bind an interactive command to the key of his choosing.

dkrivets:
Thanks for review.  My general aim is understand how to create emacs package and distribute it. 
text-buffer is my first attempt. And of course it couldn't be a trivial and useful a little.

> * There is already a function to generate unique buffer names built-in to Emacs: `generate-new-buffer`.  This package's functionality essentially boils down to wrapping `generate-new-buffer` in a command like:
> 
> ```emacs-lisp
> (defun unique-buffer ()
>   "Pop to new, uniquely named buffer."
>   (interactive)
>   (pop-to-buffer (generate-new-buffer "TEMP")))
> ```
That example is really does that text-buffer does.
But I'm newbie in emacs to write elisp and know emacs well too much.
And I think like a new has been done 
1) Try to find in google
2) Find in remote package store
That's why I done text-buffer :)  

> * Your package is doing some very strange things, e.g. `text-buffer--optimize`'s byte-compiling files for no apparent reason, and using a global minor mode to enforce a key binding rather than simply letting the user bind an interactive command to the key of his choosing.

Yes, it's my fault. When I made package I thought that self compile was a good thing: user could be decide to speed this package code up or not. I steel don't know how good this idea.
For key-binding: I added 'text-buffer-load-hook as ability and added example of it in readme-file and docstring. :)

alphapapa:
I appreciate your wanting to contribute a package.  I recommend that you start by studying some simple packages that are reasonably popular and modeling your code after them.  Also, study existing resources like the elisp manual, the MELPA packaging guide, https://github.com/bbatsov/emacs-lisp-style-guide, https://github.com/alphapapa/emacs-package-dev-handbook, etc.

dkrivets:
I'm sorry for long absence. How can I change reciept in this PR? It needs to show changes of new version

PR #5964, opened 2019-01-08
Add double-saber

dp12:
double-saber is used to filter out unwanted search output by narrowing or deleting with regular expressions. 

At the same time, it preserves parts of the search output buffer such as the search command.

https://github.com/dp12/double-saber

Maintainer

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/double-saber
+++ b/recipes/double-saber
@@ -1,0 +1,2 @@
+(double-saber :fetcher github :repo "dp12/double-saber")
+
--- a/recipes/parrot
+++ b/recipes/parrot
@@ -1,0 +1,4 @@
+(parrot
+ :fetcher github
+ :repo "dp12/parrot"
+ :files ("parrot.el" "party-parrot.el" "img"))


dp12:
checkdoc gives me a warning that "changes" should probably be imperative "change". 

But I believe checkdoc is in error, because the `undo' function from `simple.el' (part of emacs) has the docstring "Undo some previous changes."

manuel-uberti:
FWIW I gave it a test, and it works as expected. Nice package!

purcell:
Thanks for this, and sorry for the slow turnaround. Quick notes:

- Don't [require `subr-x` only at compile-time](https://github.com/dp12/double-saber/blob/785e5d85e5e6d2feab1468113f423cd4fe98c7e6/double-saber.el#L36) - you're using `string-trim` at runtime, so this makes no sense.
- Your code for tracking and maintaining the read-only state of the buffer is much more complex than it needs to be: all you need to do is let-bind `inhibit-read-only` to `t` while you modify the buffer.
- The [`double-saber-mode-setup`](https://github.com/dp12/double-saber/blob/785e5d85e5e6d2feab1468113f423cd4fe98c7e6/double-saber.el#L148) function is unidiomatic, and you shouldn't be modifying other modes' keymaps. Instead, define a minor mode called `double-saber-mode`, with its own keymap, and then users can enable it automatically or manually as needed, just like any other minor mode.

Otherwise looks good, and useful!

dp12:
Thanks for the suggestions! I think I've addressed the points you mentioned.

dp12:
Thanks again for the comments! I added the check you suggested and and'ed it with the original fboundp check (because flycheck was complaining about not knowing about `undo-tree-redo` and `undo-tree-undo`)

manuel-uberti:
I used the last version @dp12 has on his repository, and it works as expected. Great stuff. :+1:

purcell:
Thanks - merged now, and sorry for the slow turnaround. Welcome to MELPA! :tada:

PR #5966, opened 2019-01-11
Update caml recipe

andschwa:
In https://github.com/ocaml/ocaml/pull/2182 the Emacs package `caml`
providing `caml-mode` was moved from the `ocaml/ocaml` repository to an
independent repository `ocaml/caml-mode`. This commit updates the Melpa
package to point to the new location.

The `caml` package provides an Emacs mode for the OCaml language.

https://github.com/ocaml/caml-mode

Enthusiastic user.

This [comment](https://github.com/ocaml/ocaml/pull/2182#issuecomment-444196492) on the PR which moved the location of the package indicated that Melpa would get updated, but this did not happen.

### Checklist

Please confirm with `x`:

- ~~[ ] I've used the latest version of [package-lint](https://github.com/purcell/package-lint) to check for packaging issues, and addressed its feedback~~
- ~~[ ] My elisp byte-compiles cleanly~~
- [x] `M-x checkdoc` is happy with my docstrings (mostly, given https://github.com/ocaml/caml-mode/pull/3)

I am simply updating the location of the package in accordance to what its maintainers did. This package already exists in Melpa, so I assume a few of the last checklist items are not necessary from me.

### DIFF
--- a/recipes/caml
+++ b/recipes/caml
@@ -1,3 +1,1 @@
-(caml :fetcher github
- :repo "ocaml/ocaml"
- :files ("emacs/*.el"))
+(caml :repo "ocaml/caml-mode" :fetcher github)


andschwa:
@purcell Would you please take a look at this? I checked it with the maintainers. All that happened was that `caml-mode` was officially extracted from one repo to another; it's an existing MELPA package. Thanks!

purcell:
Merged, thankx

andschwa:
Thank you!

PR #5967, opened 2019-01-11
Add text buffer 0.0.3

dkrivets:

Simple way to create new buffer does not think about it name.

https://github.com/dkrivets/text-buffer

I am

None needed

### Checklist

Please confirm with `x`:


alphapapa:
Why did you open a second pull request when there's still https://github.com/melpa/melpa/pull/5963 ?

dkrivets:
I think the first would be canceled: the version of package has been chanded

purcell:
> That's not how MELPA works.

Yep, the recipe is how MELPA will (always) build the latest version of your code, so the recipe generally stays the same while your code is free to change over time.

PR #5968, opened 2019-01-11
Add fullframe-minor-mode recipe

jtkDvlp:

Minor mode for easy use via keybindings of [fullframe](https://github.com/tomterl/fullframe)

https://github.com/jtkDvlp/fullframe-minor-mode

I am the maintainer and contributor and enthusiastic user.

**None needed**

### Checklist

Please confirm with `x`:

I will check them in a few days.

### DIFF
--- a/recipes/fullframe-minor-mode
+++ b/recipes/fullframe-minor-mode
@@ -1,0 +1,2 @@
+(fullframe-minor-mode :fetcher github
+                      :repo "jtkDvlp/fullframe-minor-mode")


alphapapa:
* I'm having a hard time understanding exactly what this package does.  I sort-of understand what `fullframe` does, although its documentation was not easy to understand (not your package, of course).
* It seems to be primarily a couple of keybindings.  MELPA generally discourages packages that simply bind keys, because that should be left up to users.
  * The key `C-q` is a fairly important default key binding.  If you are going to have default key bindings in a package, they shouldn't conflict with ones like that.

purcell:
Thanks, and sorry for the slow response!

I don't think `fullframe` is a good way to do this: it's much more complex than you need, and isn't really designed for this usage because it works with advice.

Instead, users can simply enable the built-in `winner-mode`, then they can just hit `C-x 1` to "enable full frame" and use `winner-undo` to restore the previous window configuration.

If you really want to reproduce the behaviour of `fullframe` as a minor mode, then it would be easy to write code that saves the window configuration to a buffer-local variable then calls `delete-other-windows`.

jtkDvlp:
Sorry for late replay. @purcell thank you for your idea. You are right, so there is no need for this mode.

PR #5969, opened 2019-01-12
Delete tern-django recipe.

proofit404:
`tern-django` was unmaintained for a while. It is an insecure broken package. I decide to remove it since no community interest occurs for a few years.


purcell:
Thanks, makes sense.

PR #5973, opened 2019-01-13
Added a package file for poet-client

wailo:

Emacs client for po.et. A decentralized protocol for content ownership, discovery and monetization in media.

https://github.com/wailo/emacs-poet

Author

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/poet-client
+++ b/recipes/poet-client
@@ -1,0 +1,1 @@
+(poet-client :repo "wailo/emacs-poet" :fetcher github)


alphapapa:
Please tidy up your code to make it easier to review.  e.g. there are a lot of blank lines in strange places, code that isn't indented properly, etc.  You may find `aggressive-indent-mode` helpful.

wailo:
Thanks for the tip. I have reformatted the whole code and removed blank lines

alphapapa:
It's not generally correct to use `defvar` inside a function.  Is there something about the widget library that requires that?

wailo:
Without defvar, I was getting (assignment to free variable) warnings. which indicateas that the variables are shared in the global space, while my intention is to keep the vars in a local scope.
According to the answer in this link: https://emacs.stackexchange.com/questions/24169/how-to-fix-warning-assignment-to-free-local-variable-in-lambda-function, using defvar solves the warning. and the doc in this link https://www.gnu.org/software/emacs/manual/html_node/elisp/Defining-Variables.html states that defvar will bind the variable to local scope if declared without init value. This is why I used defvar. 

Looking at the code again, I decided to get rid of all defvars and replace them with a temporaries in a [let] statement. Changes have been pushed.

purcell:
Looks good - thanks for your patience, merging now. Welcome to MELPA! :tada:

wailo:
Thank you!

PR #5976, opened 2019-01-17
Add recipe for blame-this

pierre-lecocq:

Display a git blame summary of the current line.

https://github.com/pierre-lecocq/blame-this

Author

 **None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/blame-this
+++ b/recipes/blame-this
@@ -1,0 +1,1 @@
+(blame-this :fetcher github :repo "pierre-lecocq/blame-this")


purcell:
Isn't this pretty much the same as [git-messenger](https://github.com/syohex/emacs-git-messenger/blob/83815915eb8c1cb47443ff34bca3fecf7d2edf3a/git-messenger.el)?

alphapapa:
Magit also covers this functionality quite thoroughly.

pierre-lecocq:
@purcell , @alphapapa thanks for pointing me these alternative solutions !
I don't use magit myself, and did not know about git-messenger.

I wanted to make a _very lightweight_ package for this feature as it is an important part of our workflow.
Nevertheless, if you do think it is too redundant, feel free to reject this pull request.

On a side note, thank you very much for your hard work maintaining such a valuable tool for the community!

Pierre.

purcell:
Yep, I think that since this is well covered elsewhere, I'd prefer to decline the package, but thanks for taking the time to submit it!

PR #5977, opened 2019-01-17
Document requirement for GPL compatible licenses

danielcompton:
I was looking at the contributing requirements and noticed that the requirement to have a GPL compatible license was in the PR template, but not in the CONTRIBUTING file. I didn't have any firm details on the background of this, so I left it pretty minimal, but a maintainer might want to add more context to this requirement.


purcell:
Awesome, yes, thank you!

PR #5979, opened 2019-01-18
Add recipe for org-revealjs

lechten:

This package provides a maintained fork of org-reveal. As such, it provides Org export functionality to generate HTML presentations with the presentation framework reveal.js.

https://gitlab.com/oer/org-revealjs

Maintainer

Currently not possible.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/org-revealjs
+++ b/recipes/org-revealjs
@@ -1,0 +1,5 @@
+(org-revealjs
+ :repo "oer/org-revealjs"
+ :branch "melpa"
+ :fetcher gitlab
+ :files (:defaults "LICENSES" "Readme.org" "local.css" "images"))


alphapapa:
If I may, I suggest using a package name that more obviously communicates that it's *de facto* a successor to the `org-reveal` package, rather than simply another implementation.  

As it stands, if I'm a new user, and I'm looking for an Org reveal.js package, I can't tell from the names `org-reveal` and `org-revealjs` which one I should use, which one is newer, etc.  A name could be chosen which would make it more obvious.  Some possibilities:

* `org-reveal2`: Maybe a little ugly with the `2`, but pretty clear that it's a successor.
* `org-reveal-fork`
* `org-reveal-ng`: While perhaps not enough of a rewrite to justify "ng", at least "ng" is somewhat of a traditional suffix for forks of abandoned packages in the FOSS world.
* `org-re-reveal` or `org-rereveal`: A little silly, but at least unique, and maybe it would pique users' curiosity enough to investigate further.

My 2 cents.  Thanks for picking up this package.

lechten:
Many thanks for the suggestion.  I pushed a commit to the melpa branch that extends the Commentary section with a sentence on purpose and history of the package.

I'll sleep over org-re-reveal. I also think about org-reveal-oer for Open Educational Resources...

kidd:
Have you tried to reach out to the maintainer of the old package? I adopted one package from a MIA author, and the process took some time, but it's way better and cleaner for the long run. 

I'd encourage you to try that way instead of creating another package.


- https://github.com/myuhe/org-gcal.el/issues/115
- https://github.com/melpa/melpa/pull/5624

lechten:
Yes, I tried to work with the maintainer:

- https://github.com/yjwen/org-reveal/issues/349
- https://github.com/yjwen/org-reveal/pull/277

See also PR #5901 where @purcell addressed @yjwen without response.

lechten:
Your question certainly wasn't noise. I should have explained that here or provided a pointer right away...

Concerning the name: Among the suggestions, I prefer `org-re-reveal`, though I'm not sure that it really communicates being a successor. A quick poll: Go for `org-re-reveal` or stick with `org-revealjs`?

lechten:
Opened PR #5989 for `org-re-reveal`, closing this.

PR #5980, opened 2019-01-19
add package: equake

emacsomancer:

a drop-down console with tabs in the style of Yakuake for use with eshell/(ansi-)term/shell

https://gitlab.com/emacsomancer/equake

creator/maintainer

**None needed**

### Checklist

Please confirm with `x`:

- [x] My elisp byte-compiles cleanly^†

† There are warnings about `trampolined-function` as a side-effect of using [tco](https://github.com/Wilfred/tco.el).

### DIFF
--- a/recipes/equake
+++ b/recipes/equake
@@ -1,0 +1,1 @@
+(equake :fetcher gitlab :repo "emacsomancer/equake")


emacsomancer:
> I like this - it's working without a hitch on macOS. 

That's great news! Thanks for reporting this. I was hoping this would true since I'm just relying on `frame.el`, and it seems to handle macOS and Windows (so hopefully Equake should work fine on Windows as well). 

> Just passing through with some quick comments:

Many thanks for these. I've fixed all of these as suggested, with the exception of:

> ... the string `"EQUAKE"` is frequently searched for as a means to determine whether a buffer is an equake buffer -- could this be factored out into a predicate like `equake-buffer-p`?

The trouble is that sometimes I need to see if the `current-buffer` is an Equake-associated buffer (i.e. has "EQUAKE[" in its name), but sometimes I need to check if the `current-buffer` is an Equake-associated buffer on a particular monitor/screen, and sometimes I need to check for arbitrary buffers. So I feel like I would really just be replacing `string-match-p` with `equake-buffer-p` which would require nearly the same set of arguments anyway, or else be writing three or four different functions to cover the different cases.

emacsomancer:
@alphapapa Instead of checking buffer/frame names, or for some other purpose? (I decided to try to encode more information directly in the buffers/frames for robustness reasons - to be less fragile against events occurring 'outside of' equake .)

purcell:
- Code says "(setq lexical-binding 't) ; redundant?" -- the answer is "yes"!
- "Percentage (.01-1.0) for width of equake console." -- this should be "fraction", surely?
- The default for `equake-default-sh-command` should not simply be "/bin/bash". I'd suggest preferring `nil`, and then replace that with the runtime value of `shell-file-name` if unset, to avoid user confusion.
- The various `-foreground` and `-background` colour variables are unidiomatic and not amenable to theming. Instead, define faces.
- `equake-key-bindings` suggests strongly that this should be either a global minor mode, or a global+local mode pair. I suspect this might help tidy up some of the other code too: there's indeed a lot of suspicious-looking checking for EQUAKE in buffer names, which you could entirely avoid if those buffers were in a local minor mode that you could simply check for.

Hope that helps - this looks really interesting!

emacsomancer:
@purcell  Thanks!  I've addressed all of the points, and created an Equake minor-mode which has more easily customisable keybindings as well as proper theming. This enabled me to avoid *some* checking of buffer names, but not completely as sometimes it is crucial which monitor/screen a buffer is 'associated' with, and buffer names encode this information.

purcell:
Awesome, thanks - merged now. Welcome to MELPA. :tada:

PR #5983, opened 2019-01-20
Add new recipe named elscreen-tab

aki-s:

- Show the dedicated buffer for `elscreen` (This can be alternaitve to `(setq elscreen-display-tab t)`.).
  - Existing `tab` of `elscreen` use `header-line-format` but using this is not preferable because;
    - The header may already be used for the other purpose (e.g. `which-func-mode`)
    - The same buffer shows the same header to show tabs of `elscreen` (e.g. When the buffer showing `tab`s is split with `C-x 3`.).

https://github.com/aki-s/elscreen-tab

- Are you the maintainer? => Yes
- A contributor? => Yes
- An enthusiastic user? => Yes. I've created this mode and it's almost 2 years past since the first version.
I've fostered this mode by myself. I felt I can make public now.

**None needed**

### Checklist

Please confirm with `x`:

- [x] I've used the latest version of [package-lint](https://github.com/purcell/package-lint) to check for packaging issues, and addressed its feedback `  
<- I want to use : as a separator for namespace..., because - is ambiguous delimiter. -- is for *private* method by contract, so what shall I use...`
- [△] `M-x checkdoc` is happy with my docstrings `  
<- I have omitted comment for private function such as elscreen-tab:--set-position.  
<-  I also did not added keyword function or variable for word quoted with \`', because it is not actually ambiguous. If I add keyword, then sentence becomes redundant or unclear.`

### DIFF
--- a/recipes/elscreen-tab
+++ b/recipes/elscreen-tab
@@ -1,0 +1,1 @@
+(elscreen-tab :fetcher github :repo "aki-s/elscreen-tab")


aki-s:
@purcell Thank you for your review.

I have reflected what you noted.
Could you check again please?

---

>Please use versioned dependencies on cl-lib and seq -- package-lint should have pointed this out. (Although: neither is actually necessary, since you're depending on Emacs 26.)

https://github.com/aki-s/elscreen-tab/pull/1/commits/2c53c523a0ffa7cf1c082a735c278d8f1ba598d5

>Don't use : in symbol names: instead of elscreen-tab:tab-window-parameters, for example, use elscreen-tab-tab-window-parameters. package-lint should also have pointed this out.

https://github.com/aki-s/elscreen-tab/pull/1/commits/bcb71e661e596422d051bc4989d4f06e05b582f0
https://github.com/aki-s/elscreen-tab/pull/1/commits/c023ad3594367bb2680e526a3798409123b679cb

>Instead of specifying colours for your new face, prefer to :inherit a built-in face so that it looks okay be default in all existing themes.

https://github.com/aki-s/elscreen-tab/pull/1/commits/385628ab38c313653c6c59d08838975f908f2255

>There's no need for elscreen-tab:debug-log to be a macro: it can just be a function.

https://github.com/aki-s/elscreen-tab/pull/1/commits/7b51dec9dcf30f56aa23de7e7eeba839b3b1b1e6

>I think the with-eval-after-load form is incorrect, because there's no guarantee that at the time elscreen-persist is loaded, elscreen-tab-mode is enabled.

https://github.com/aki-s/elscreen-tab/pull/1/commits/44f1a5f0ac0e32e7272dd79e2226a8fd235b27fb

aki-s:
@purcell
I'm sorry to poke you, but 20 days have past since I updated my PR.
How did you like it?

purcell:
Thanks, looks good now, so will merge. Sorry it took a while: I'm very busy and have to batch-process these requests every 1-3 weeks. I'm half-way through about 40 (!) pull request reviews today... Welcome to MELPA! :tada:

(P.S. Don't leave all those `)` and `))` on their own lines: they get lonely, and it's not conventional elisp style. :smile:)

PR #5987, opened 2019-01-24
Add recipe for keystore-mode

peterpaul:

This package provides a major mode based on `tabulated-list` for viewing and editing jks and pkcs12 keystores using the `keytool` command.

https://github.com/peterpaul/keystore-mode

I am the author/maintainer of this package.

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/keystore-mode
+++ b/recipes/keystore-mode
@@ -1,0 +1,4 @@
+(keystore-mode
+ :repo "peterpaul/keystore-mode"
+ :fetcher github
+ :files ("*.el"))


purcell:
Thanks! Couple of points:
- You're concatenating strings into arguments that might contain spaces, quotes or other special shell chars. This is a Very Bad Idea™ :-) -- use `shell-quote-argument`.
- `list-keystore` is not named with the correct package prefix: it should probably be `keystore-list`.
- The `details-mode` elisp uses the `seq` package, but the "master" .el file does not declare that package as a dependency.
- Use `defvar` instead of `setq` to introduce variables, e.g. [here](https://github.com/peterpaul/keystore-mode/blob/12d9fafefe8af58c7cf68e73c36eacfb0c08603a/keystore-details-mode.el#L45)
- Don't modify origami's config globally [here](https://github.com/peterpaul/keystore-mode/blob/12d9fafefe8af58c7cf68e73c36eacfb0c08603a/keystore-details-mode.el#L125). You probably need to be setting that variable locally inside specific buffers instead?

Hope that helps!

peterpaul:
Thank you very much for your review.

I'm still quite new to (e)lisp, so I take this as an opportunity to learn :). I already fixed the easy items you raised.

For the [Bobby Tables](https://xkcd.com/327/) problem I need to think a bit about how to solve it. I might have over-refactored this. 

I named `list-keystore` like this because of the following section in the documentation of `tabulated-list-mode`:
> An inheriting mode is usually accompanied by a "list-FOO"
> command (e.g. list-packages, list-processes).

Is `list-keystore` OK then, or do you still prefer `keystore-list`?

purcell:
> Is `list-keystore` OK then, or do you still prefer `keystore-list`?

That advice is misguided and you should use `keystore-list`, please. The Emacs core library authors can (unfortunately, sometimes) largely do whatever they please, but as a general rule packages should keep to their own well-defined namespaces to avoid interfering with each other.

peterpaul:
Thank you for taking the time to explaining that to me. It is highly appreciated. And I'll never again question the output of `package-lint` ;).

Since the function `keystore-list` did already exist, I renamed it to `keystore-visit`.

To fix the command injection problem, I now apply `shell-quote-argument` to all the atoms that form a command (https://github.com/peterpaul/keystore-mode/commit/d1fbf22c9f325bb8b47972ca01b8a3960698344b)

I believe with that I fixed all the issues you raised.

purcell:
Super, merged now - thanks! :-)

PR #5988, opened 2019-01-25
Add language-id recipe

lassik:

*language-id* is a small, focused library that helps other Emacs packages identify the programming languages and markup languages used in Emacs buffers. The main point is that it contains an evolving table of language definitions that doesn't need to be replicated in other packages.

Right now there is only one public function, *language-id-buffer*. It looks at the major mode and other variables and returns the language's GitHub Linguist identifier. We can add support for other kinds of identifiers if there is demand.

This library does not do any statistical text matching to guess the language.

https://github.com/lassik/emacs-language-id

Just wrote it ;-)

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/language-id
+++ b/recipes/language-id
@@ -1,0 +1,1 @@
+(language-id :fetcher github :repo "lassik/emacs-language-id")


purcell:
Thanks! You should probably prefer `derived-mode-p` to `eq` when checking the major mode. And [here](https://github.com/lassik/emacs-language-id/blob/master/language-id.el#L196-L197) you might like to use `cl-find` or similar, with a custom `:test` function.

purcell:
(Merged in the meantime. :-))

lassik:
Thanks for merging and the helpful comments :) I changed it to use `derived-mode-p` as you suggested, and replaced the `cl-dolist`/`cl-return` loop with `cl-some`. It is indeed better style.

purcell:
Great!

PR #5989, opened 2019-01-25
Add recipe for org-re-reveal

lechten:

This package provides a maintained fork of org-reveal, which is no longer maintained. As such, it provides Org export functionality to generate HTML presentations with the presentation framework reveal.js.

See PR #5979 for background and naming, which I'll close in a minute.

https://gitlab.com/oer/org-re-reveal

Maintainer

Currently not possible.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/org-re-reveal
+++ b/recipes/org-re-reveal
@@ -1,0 +1,4 @@
+(org-re-reveal
+ :repo "oer/org-re-reveal"
+ :fetcher gitlab
+ :files (:defaults "LICENSES" "Readme.org" "local.css" "images"))


purcell:
Thanks! I think that a couple more of the `reveal` references should be changed to `re-reveal`: the custom group name and the backend name, specifically. (It's fine to leave the various `:reveal-*` options as-is for compatibility, of course, but the goal is to able to have both `org-reveal` and `org-re-reveal` installed and for them to not interfere with each other.) Looks good otherwise.

lechten:
I changed that. I also changed the export key bindings to lower case to have different ones from `org-reveal`. Besides, I added `htmlize` as dependency as some Org versions (e.g., 9.1.9 shipping with Emacs master) throw an error when exporting `Readme.org` otherwise.

purcell:
Great, thanks - merged now. :-)

lechten:
Many thanks for your feedback and efforts, I really appreciate your amazing work!

PR #5997, opened 2019-02-01
add ammonite-term-repl

zwild:

Scala Ammonite REPL in Emacs term mode

https://github.com/zwild/ammonite-term-repl

I'm the creator of this package

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/ammonite-term-repl
+++ b/recipes/ammonite-term-repl
@@ -1,0 +1,2 @@
+(ammonite-term-repl :fetcher github
+                    :repo "zwild/ammonite-term-repl")


purcell:
Nice and tidy, thanks - will go ahead and merge, but [this line](https://github.com/zwild/ammonite-term-repl/blob/777296e039a3547ab769265a5e2f95f4be0e8ad6/ammonite-term-repl.el#L175) is a mistake, since the hooks will also get called when the mode is disabled. Generally `define-minor-mode` will define the hook var for you, and call it when appropriate.

PR #5998, opened 2019-02-02
adding recipe for mermaid-mode

abrochard:

Major mode for working on mermaid graph files.

https://github.com/abrochard/mermaid-mode

maintainer?

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/mermaid-mode
+++ b/recipes/mermaid-mode
@@ -1,0 +1,1 @@
+(mermaid-mode :repo "abrochard/mermaid-mode" :fetcher github)


purcell:
Thanks! Some notes:
- Use `defconst` to introduce `mermaid-highlights`, [not `setq`](https://github.com/abrochard/mermaid-mode/blob/b9782e75ab9ab1d6a70bd944213c396f7740bf79/mermaid-mode.el#L60).
- Consider converting the two `defvar`s into `defcustom`s, with a containing `defgroup`.
- There's no guarantee that `(buffer-file-name)` returns non-nil, so [this code](https://github.com/abrochard/mermaid-mode/blob/b9782e75ab9ab1d6a70bd944213c396f7740bf79/mermaid-mode.el#L97) will be unreliable. Where possible, don't expect that every buffer in your major mode is associated with a file.
- [This line](https://github.com/abrochard/mermaid-mode/blob/b9782e75ab9ab1d6a70bd944213c396f7740bf79/mermaid-mode.el#L123) is generated for you by `define-derived-mode`, so you should delete it.
- `mermaid-highlights` should be called `mermaid-font-lock-keywords`, which is much more conventional.
- Don't call "open" [here](https://github.com/abrochard/mermaid-mode/blob/b9782e75ab9ab1d6a70bd944213c396f7740bf79/mermaid-mode.el#L105) - instead, use `browse-url-default-browser`. ("open" is only available on MacOS.) Same applies to `mermaid-open-doc`.

Hope that helps.

purcell:
Super slow (sorry) but merged now. Thanks!

abrochard:
No problem! Thank you very much! Onto the next one.

purcell:
> Onto the next one.

Give me some time to keep up... ;-)

PR #5999, opened 2019-02-02
add ob-ammonite

zwild:

org-babel for scala evaluation in ammonite-term-repl

https://github.com/zwild/ob-ammonite

I'm the creator of this package.

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/ob-ammonite
+++ b/recipes/ob-ammonite
@@ -1,0 +1,2 @@
+(ob-ammonite :fetcher github
+             :repo "zwild/ob-ammonite")


zwild:
Hi. Thanks for your reply and tips!
Now I update my codes with all your tips but the last one, which the literal newline `^M` is redundant. So I replace it with an empty string. If I replace it with `\n`, the result returned by `org-babel-execute:amm` will contain a redundant newline, which is very strange.

This is an example. 



Thanks for your time.

purcell:
Great, thanks - merging now.

PR #6000, opened 2019-02-03
Add recipe for dune

Chris00:

This is a mode to edit configuration files for the [dune build system](https://github.com/ocaml/dune).

https://github.com/ocaml/dune

I am a [co-maintainer of the package](https://github.com/ocaml/dune/issues/1710).

None needed.

### Checklist

Please confirm with `x`:

These confirmations depend on the [following PR being merged](https://github.com/ocaml/dune/pull/1799).

### DIFF
--- a/recipes/dune
+++ b/recipes/dune
@@ -1,0 +1,2 @@
+(dune :fetcher github :repo "ocaml/dune"
+      :files ("editor-integration/emacs/*.el"))


tarsius:
Please consider putting the elisp libraries in a separate repository. That makes it easier for people who want to install dune using a system package but install emacs packages by cloning git repositories. (I do that sort of thing using [borg](https://github.com/emacscollective/borg.) A separate repository would also make it easier to mirror the package on the Emacsmirror. And finally it might make it a bit more likely that users contribute to the package.

I think caml-mode used to be part of the ocaml repository but was recently moved out. It would be nice if that could be done across all repositories of the ocaml organisation.

Chris00:
@tarsius I took you points to the main dune developers and they [are against it](https://github.com/ocaml/dune/pull/1799#issuecomment-460101355).

purcell:
Thanks, including for following up about @tarsius' point above.

Quick notes:
- If filenames contain spaces then [this code](https://github.com/ocaml/dune/blob/b8f09237f6b9cab03f2ea83dbf50e1598af2c36f/editor-integration/emacs/dune.el#L391-L401) will likely break. This may be unlikely, but you can make it safe with `shell-quote-argument`.
- The code requires Emacs 24.3, but the header has no `Package-Requires` dependency on that Emacs version. `package-lint` should have pointed this out.

Chris00:
@purcell Thanks for your comments.

>     * This may be unlikely, but you can make it safe with `shell-quote-argument`.

Done in https://github.com/ocaml/dune/pull/1799/commits/690bf79d84a6b3d70ac4b788fb80ece1745c0ef9

>     * The code requires Emacs 24.3

This is correct for the current version of the code but will no longer be the case after the PR https://github.com/ocaml/dune/pull/1799 is merged.

Chris00:
The PR on the Dune side has been merged.

purcell:
Great, thanks! This is now merged. You might also like to arrange to turn `dune-flymake-program` into a `defcustom`.

Chris00:
Thanks.

PR #6001, opened 2019-02-03
Add recipe for org-re-reveal-ref

lechten:

This package extends `org-re-reveal` with support for a bibliography slide based on package `org-ref`.  Thus, `cite`commands of `org-ref` are translated into hyperlinks to the bibliography slide upon export by `org-re-reveal`.  Also, export to PDF via LaTeX with Org's usual export functionality works.

I created a separate package for this as `org-ref` has lots of dependencies, which I do not want to force upon users of `org-re-reveal`.

https://gitlab.com/oer/org-re-reveal-ref

Maintainer

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/org-re-reveal-ref
+++ b/recipes/org-re-reveal-ref
@@ -1,0 +1,4 @@
+(org-re-reveal-ref
+ :repo "oer/org-re-reveal-ref"
+ :fetcher gitlab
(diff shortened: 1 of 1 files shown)


purcell:
Thanks!

PR #6002, opened 2019-02-03
add dired-rmshit

TsarFox:

The interactive function, `dired-rmshit` will mark all files in the current Dired buffer that match one of the patterns specified in `dired-rmshit-shitty-files`. The tool is intended as a simple means for keeping one's home directory tidy -- removing "junk" dotfiles.

https://git.sr.ht/~jakob/dired-rmshit

I am the maintainer.

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/dired-rmjunk
+++ b/recipes/dired-rmjunk
@@ -1,0 +1,1 @@
+(dired-rmjunk :url "https://git.sr.ht/~jakob/dired-rmjunk" :fetcher git)
--- a/recipes/dired-rmshit
+++ b/recipes/dired-rmshit
@@ -1,1 +1,0 @@
-(dired-rmshit :url "https://git.sr.ht/~jakob/dired-rmshit" :fetcher git)
--- a/recipes/dired-rmshit
+++ b/recipes/dired-rmshit
@@ -1,0 +1,1 @@
+(dired-rmshit :url "https://git.sr.ht/~jakob/dired-rmshit" :fetcher git)


TsarFox:
Thanks for the feedback! I've renamed the package; I agree that the crass name is distasteful in a public context, which the original script wasn't really released to. I've also modified the function's behavior to reflect the "pattern" terminology.

I didn't think having defaults for `dired-rmjunk-patterns` would be an issue because the utility only marks files rather than delete them, but now that you mention it, the potential for accidental data loss is something I should be keeping in mind. Do you happen to know if there's a way of specifying multiple defaults that a user can choose from? I'm not particularly familiar with `custom`.

purcell:
Merged now, thanks, and welcome to MELPA! :tada:

> Do you happen to know if there's a way of specifying multiple defaults that a user can choose from? I'm not particularly familiar with `custom`.

Yes, you can have named selectable sets (with `:tag`) while keeping the default empty or minimal. I can't remember the syntax off-hand, but try grepping your installed packages and you'll find some examples.

Alternatively you could use a two-level list scheme like `compile` does with `compilation-error-regexp-alist` and `compilation-error-regexp-alist-alist` (IIRC) so that you have a shorthand for different groups of patterns.

TsarFox:
Thanks for the info, I'll be looking into those possibilities over the coming days. I think I need to build a proper understanding of `defcustom` before I'll be able to wrap my head around selectable defaults

PR #6003, opened 2019-02-03
Add Hyperspace

ieure:

Jumps you to different parts of Emacs or the web.

https://github.com/ieure/hyperspace-el

I'm the package author.

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/hyperspace
+++ b/recipes/hyperspace
@@ -1,0 +1,1 @@
+(hyperspace :fetcher github :repo "ieure/hyperspace-el")


emacsomancer:
(you may want to correct the link. I.e. to https://github.com/ieure/hyperspace-el )

ieure:
>  (you may want to correct the link. I.e. to https://github.com/ieure/hyperspace-el )
>
Thanks, fixed.

purcell:
This is cool. Couple of notes:

- "Search Quick" should be "Search quickly"
- Don't forget to run `checkdoc` - I can see at least one thing it would complain about.
- Does this really byte-compile cleanly? I'm not sure how it could, since the code uses functions from elfeed and mu4e. What I believe you should do is provide this base package without any builtins that require external packages, e.g. those two. Then you can define the separate helpers either in your personal config or add-on packages which have proper dependencies. Otherwise there is no sensible strategy for extending this package other than adding a lot of dangling references to functions in possibly-not-installed other packages.
- Require `'cl-lib` to use `cl-loop` etc.
- You should likely prefer `use-region-p` to `region-active-p` [here](https://github.com/ieure/hyperspace-el/blob/66bdf3081fd7397027d78dc546f8a112bf100d68/hyperspace.el#L204)
- Consider calling the global minor mode `hyperspace-global-mode`.

Hope that helps!

ieure:
> * Don't forget to run `checkdoc` - I can see at least one thing it would complain about.

What are you seeing?  It passes for me on Emacs 25.1.1:

```
Buffer comments and tags:  Ok
Documentation style:       Ok
Message/Query text style:  Ok
Unwanted Spaces:           Ok
```

> Does this really byte-compile cleanly?
>
Does "cleanly" mean "with no warnings," or "with no errors?"  There are warnings caused by the way `pcase` expands which appear unavoidable:

```
hyperspace.el:166:1:Warning: Unused lexical variable ‘kw’
hyperspace.el:166:1:Warning: Unused lexical variable ‘kw’
```

> I'm not sure how it could, since the code uses functions from elfeed and mu4e.
>
You're correct, I hadn't noticed because both those were loaded into my Emacs session.

> What I believe you should do is provide this base package without any builtins that require external packages, e.g. [mu4e and elfeed].
>
I'm not terribly keen on this, but might go that route if I can come up with a way to do it that I think is good.  I don't want manual configuration when you install an addon, but they also shouldn't screw with a user-customizable var, so some other mechanism would be needed.

I'll just pull those out and shove them in the README as an example until then.

> Require `'cl-lib` to use `cl-loop` etc.

Is there a coding standard that requires doing so?  The `cl-loop` macro is autoloaded, so it's not necessary to require the whole lib, nor does it give a byte-compiler warning.  I'd prefer not to load the whole lib if I don't have to.

  The only thing I'm using is `cl-loop`, which is autoloaded, and the byte-compiler doesn't complain that it's not being explicitly `require`d.  I'm not arguing, just curious.

Other items fixed.

purcell:
> What are you seeing

Ah. I was seeing missing trailing "." in at least one docstring, but maybe I'm imagining that checkdoc catches those. No worries then.

> Does "cleanly" mean "with no warnings," or "with no errors?" There are warnings caused by the way `pcase` expands which appear unavoidable:

With no errors: in a clean Emacs, without `mu4e` and `elfeed` being explicitly required, referring to their functions should cause byte compilation issues. (Note here that `flycheck` will run compilation in a fresh background sub-Emacs, which can provide better errors than calling `M-x byte-compile` in your working Emacs.

> I'll just pull those out and shove them in the README as an example until then.

That's a good move for now. Happy to help advise re. alternative plans.

> Is there a coding standard that requires doing so? The `cl-loop` macro is autoloaded, so it's not necessary to require the whole lib, nor does it give a byte-compiler warning. I'd prefer not to load the whole lib if I don't have to.

I think you'll find that the autoload for `cl-loop` causes all of `cl-lib` to be loaded anyway, so there's no difference here in what gets loaded, and it's better to be explicit with your `require`s where possible because you can't guarantee that `cl-loop` was autoloaded in all the previous versions of `cl-lib` that might satisfy your package's dependencies.

ieure:
> > Does "cleanly" mean "with no warnings," or "with no errors?" There are warnings caused by the way `pcase` expands which appear unavoidable:

> With neither, ideally: in a clean Emacs, without `mu4e` and `elfeed` being explicitly required, referring to their functions should cause byte compilation issues. (Note here that `flycheck` will run compilation in a fresh background sub-Emacs, which can provide better errors than calling `M-x byte-compile` in your working Emacs.

This is an improvement, though flycheck complains about the `s` package not being available.  I'll have to poke at it a bit more, but I believe the issues are resolved for now.

> The `pcase` errors might be avoidable by calling the var `_kw` instead of `kw`, though it's an odd case.

That produces a different warning:

`hyperspace.el:161:1:Warning: variable ‘_kw’ not left unused`

This seems like an Emacs bug.  I went ahead and rewrote it without `pcase`, it ended up being more straightforward than I thought it would be, and terser than `pcase` anyhow.

> I think you'll find that the autoload for `cl-loop` causes all of `cl-lib` to be loaded anyway, so there's no difference here in what gets loaded, and it's better to be explicit with your `require`s where possible because you can't guarantee that `cl-loop` was autoloaded in all the previous versions of `cl-lib` that might satisfy your package's dependencies.

Fair, though since `cl-loop` is part of the mu4e stuff, loading would be deferred to the mu4e use, rather than paid for up front.  It's a moot point now, nothing but the one mu4e method used it, it's not needed since that's removed.

purcell:
Great, thanks - merged now.

PR #6004, opened 2019-02-04
Add package company-ml.

fgeorges:

Company backend for MarkLogic function completions.

https://github.com/fgeorges/company-ml

Maintainer

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/company-ml
+++ b/recipes/company-ml
@@ -1,0 +1,4 @@
+(company-ml
+  :repo    "fgeorges/company-ml"
+  :fetcher github
+  :files   ("src/*.el"))


purcell:
Thanks! Some notes:

- `-ml` in the package name would be *much* clearer as `-marklogic`, due to the common use of "ml" to refer to the ML-family languages and Machine Learning.
- The only source code here is elisp, so please consider moving it from the `src` subdirectory to the top-level, which eliminates the need to specify `:files` in the recipe. This is a more conventional structure for small elisp repositories.
- You might find the `regexp-opt` function helpful [here](https://github.com/fgeorges/company-ml/blob/24e6e5449ed4659fd48299a05bd3a8c0db7a12a8/src/company-ml.el#L32)
- "prefix command" means something Emacs-specific, so I think you just mean "prefix" [here](https://github.com/fgeorges/company-ml/blob/24e6e5449ed4659fd48299a05bd3a8c0db7a12a8/src/company-ml.el#L56)
- You're using [at least one function](https://github.com/fgeorges/company-ml/blob/24e6e5449ed4659fd48299a05bd3a8c0db7a12a8/src/company-ml.el#L118) from the library `seq`, but that's not available in Emacs 24.4, nor are you declaring a dependency on the `seq` package in ELPA.
- I think you can use `cl-labels` to avoid all the `funcall` invocations [here](https://github.com/fgeorges/company-ml/blob/24e6e5449ed4659fd48299a05bd3a8c0db7a12a8/src/company-ml-xqy.el#L3482-L3493)

Hope that helps!

purcell:
Closing now due to inactivity, but please feel free to re-open once resolved.

PR #6007, opened 2019-02-06
Add recipe for use-package-company

Foltik:

Adds the keyword `:company` to `use-package`, which simplifies the process of adding 
mode-specific `company` backends.

Examples shown in [README.md](https://github.com/Foltik/use-package-company/blob/master/README.md)

https://github.com/foltik/use-package-company

Maintainer

**None needed**

### Checklist

### DIFF
--- a/recipes/use-package-company
+++ b/recipes/use-package-company
@@ -1,0 +1,2 @@
+(use-package-company :repo "foltik/use-package-company" :fetcher github)
+


purcell:
Thanks! Some notes, firstly based solely on the README:

```el
(unless (member 'company-jedi company-backends)
	  (add-to-list 'company-backends 'company-jedi))
```
is exactly equivalent to
```el
(add-to-list 'company-backends 'company-jedi)
```

I'm not sure you've understood how mode hooks and local/global variables behave. I think you're expecting that your code will add the `company-jedi` backend to `company-backends` locally when the mode's hook is run. That's not the case: every time (say) a Python buffer is opened, `company-jedi` will be added to the global list of `company-backends`.

That's generally fine, because the entries in `company-backends` are expected to behave as no-ops when they are used in unrelated buffers, but in this case you'd only have to add the backend to `company-backends` once, e.g. after `company` and `python-mode` have been loaded.

Some notes about the code itself:

- You're using the `seq` package extensively, but it's not built in to the version of Emacs you require, nor have you added a dependency to the standalone `seq` package in `Package-Requires`.
- The `append-yasnippet` handling is a strange special case, which is a red flag. There are all sorts of generally-useful backends which a user might want to use in all buffers. I think this mechanism is a bit broken given the earlier notes, and in any case I would recommend avoiding making your `use-package` keyword behaviour depend on global custom variables, which the user should rightly expect to be able to set with `customize` (and therefore potentially take effect *after* their `init.el` has been evaluated).
- You can use `derived-mode-hook-name` instead of [this code](https://github.com/Foltik/use-package-company/blob/875f4b70d21ad9564d643a8ea348c5b17f6aa101/use-package-company.el#L78), I think

Hope that helps! Not sure exactly what all that means about the correct design for the package, but happy to discuss here.

Foltik:
For the `append-yasnippet` behavior, I was going off of [this](https://github.com/company-mode/company-mode/blob/edbb3c5354ad7909b4b952c61fd8238fd656b1d0/company-yasnippet.el#L104) in order to get `yasnippet` completion candidates to show alongside those from mode specific backends and avoid shadowing others. I could just remove the global entirely and use `'(backend :with company-yasnippet)` with the keyword directly when needed.

Regarding how the backends are appended to the global `company-backends`, from what I understand instead adding to `(make-local-variable 'company-backends)` would add the backend locally whenever the mode hooks are run, which was the intended behavior.

I'm not exactly sure what versions of Emacs or `seq.el` I should be requiring, as `use-package` itself uses the very same seq functions and also requires Emacs 24.3 and not `seq.el`.

Given fixing these issues and the other small changes you mentioned, do you think this is worth being merged or should I just give up and keep it in my own config?

PR #6010, opened 2019-02-08
Add recipes/pretty-py

simonfxr:
Pretty-Py allows you to format your Python code using a selection of standard tools, currently supports yapf, autopep8 and black.

https://github.com/simonfxr/pretty-py.el

**None needed**

### Checklist

Please confirm with `x`:

- [x] `M-x checkdoc` is happy with my docstrings (except capitalization of emacs inside Package-Requires, can this be fixed?)

### DIFF
--- a/recipes/pretty-py
+++ b/recipes/pretty-py
@@ -1,0 +1,4 @@
+(pretty-py 
+  :repo "simonfxr/pretty-py.el"
+  :fetcher github
+  :files ("pretty-py.el"))


purcell:
Thanks! You've obviously worked a lot on this, and I see that a lot of the code has been cribbed from elsewhere, e.g. the `apply-rcs-patch` stuff which was originally intended to minimise modification of the current buffer and yet is unnecessary when one understands how to help Emacs handle that itself.

Additionally the code has common errors, such as that it unnecessarily modifies the global version of `before-save-hook` [here](https://github.com/simonfxr/pretty-py.el/blob/4f4250eb0e0218fb74bcfd5eded36086d22883aa/pretty-py.el#L398), necessitating the check [here](https://github.com/simonfxr/pretty-py.el/blob/4f4250eb0e0218fb74bcfd5eded36086d22883aa/pretty-py.el#L390).

All of this is why I recently wrote the [`reformatter` library](https://github.com/purcell/reformatter.el), which you can use to extremely easily wrap up command line reformatters into commands and minor modes which are well-behaved and consistent in their usage. Can I suggest that this package would be better if the core of it were one or more `reformatter-define` statements? Givent that your code bundles together several quite different formatters into one package mainly to re-use the bulky core logic, it would even be better to publish a separate small package for each formatter that is not already provided by a pre-existing package. (I see that there's a package for `yapf` already, for example.)

Hope that helps!

simonfxr:
Thank you for the review! This is my first elisp package and you obviously know a lot more about this stuff than me. Your reformatter package seems interesting too bad I didn't know about it.
Do you have some references how emacs handles it, when you replace the complete buffer content?  (E.g. if you change a few lines in a buffer and use the before-save-hook) How is font-locking and other stuff handled? I tried the yapify package before and using the rcs-patch logic it felt subjectively smoother.

purcell:
> Do you have some references how emacs handles it, when you replace the complete buffer content? (E.g. if you change a few lines in a buffer and use the before-save-hook) How is font-locking and other stuff handled? I tried the yapify package before and using the rcs-patch logic it felt subjectively smoother.

It depends how you do it. Specifically, `insert-file-contents` and `replace-buffer-contents` (both implemented in C) do some clever things to minimise changes, and to preserve marks etc. When portions of the buffer are not modified, font-lock remains unchanged for those portions. `replace-buffer-contents` - which is new in Emacs 26.1 - is particularly clever, even including a diff mechanism, so this should make the rcs-patch code completely redundant, but unfortunately the 26.1 release version of this function has an awful bug with multi-byte characters which makes it unusable. In practice, I'd be very surprised if the behaviour of code which uses even `insert-file-contents` is noticeably less smooth than the rcs-patch mechanism.

simonfxr:
Thanks for the explanation! I tested it on 25.2, maybe it was just an impression I got. 
Would it be acceptable if I reuse your reformatter package and keep support for all the formatters in a single package? My plan is for it to be integrated into spacemacs, where I would have to have some formatter switching logic anyway, I would prefer for it be factored out into a single package, someone might find it useful.

NB:
Some formatters support directly outputting a (unified) diff, maybe your reformatters package would be a good place for supporting that. If you are open to it, I could look into how feasible it is to do.

purcell:
> Thanks for the explanation! I tested it on 25.2, maybe it was just an impression I got.
> Would it be acceptable if I reuse your reformatter package and keep support for all the formatters in a single package? My plan is for it to be integrated into spacemacs, where I would have to have some formatter switching logic anyway, I would prefer for it be factored out into a single package, someone might find it useful.

I understand why you might think it better to have everything in a single package, because it provides an easy way to switch between formatters. The thing is, there's already an easy and standard way: using configurable Emacs hooks and minor modes.

With a "single package" approach, if there's a new alternative formatter that someone wants to use in future, they have to either get it added to your package, or have a "next level up" method for switching between that formatter and yours, and *also* maybe configuring yours. I really think it's better for everyone if all the formatters are separate specific packages, and the standard process is used to configure them.

> Some formatters support directly outputting a (unified) diff, maybe your reformatters package would be a good place for supporting that. If you are open to it, I could look into how feasible it is to do.

Yes, that might be something to consider over time. Want to open an issue there?

purcell:
I'm going to go ahead and close this, sorry. For the configuration preference of having a single package which centrally manages switchable formatters, there's [format-all](https://melpa.org/#/format-all), and for the one-package-per-formatter preference there's `reformatter.el`. This falls somewhere in the middle. I don't wish to be discouraging, and look forward to future submissions from you.

- Grüße from a former München resident. :-)

PR #6012, opened 2019-02-09
Add recipe for saw-script

Ptival:

Major mode for the SAWscript language (cf. https://github.com/GaloisInc/saw-script )

https://github.com/GaloisInc/saw-script

I am a developer of SAW and the author of the package @david-christiansen is sitting next to me right now! :-)

We fixed some byte-compiler warnings.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/saw-script
+++ b/recipes/saw-script
@@ -1,3 +1,4 @@
 (saw-script
  :repo "GaloisInc/saw-script"
- :fetcher github)
+ :fetcher github
+ :files ("saw-script.el"))
--- a/recipes/saw-script
+++ b/recipes/saw-script
@@ -1,0 +1,3 @@
+(saw-script
+ :repo "GaloisInc/saw-script"
+ :fetcher github)


Ptival:
cc. @david-christiansen

purcell:
Thanks, mostly looks great. Couple of notes:

- Change the global `compilation-error-regexp-alist-alist` variable at the top level, not [every time a buffer is opened in `saw-script-mode`](https://github.com/GaloisInc/saw-script/blob/2fe89d6c12bd0bacafe9a475664e02694a42a90a/saw-script.el#L389-L403). You probably then want to modify `compilation-error-regexp-alist` buffer-locally, but note that it will not be propagated from the major mode buffer to the compilation buffer started by `saw-script-run-file`.
- Use `\\'`, not `$`, in the filename regexp [here](https://github.com/GaloisInc/saw-script/blob/2fe89d6c12bd0bacafe9a475664e02694a42a90a/saw-script.el#L419)
- [Enabling flycheck automatically](https://github.com/GaloisInc/saw-script/blob/2fe89d6c12bd0bacafe9a475664e02694a42a90a/saw-script.el#L412-L416) is inappropriate for a major mode and should be removed. It's trivial for users to enable this themselves, and the availability of `flycheck` on the current system does not imply that the user wants to use it.  What's more, even depending on `flycheck` here is problematic, as hinted at by the `with-eval-after-load` sneakiness: my suggestion would be to split out `flycheck-saw-script.el`, and then we can publish two packages, each with proper dependencies and clean byte-compilation.

Hope that helps!

purcell:
I dug around and found that this got fixed up, so I've merged it. I'll manually add a second recipe for flycheck-saw-script. Thanks!

purcell:
Wait up... I wonder if I might suggest moving the elisp to a second repo? This one is gigantic, at > 900MB (!), and needing to have two check-outs of it on our server is not good, so I'll have to revert it for now.

Steps for you would be: 1) Create new repo, copy `.el` files across, 2) file quick MELPA PR to change the recipes, starting with reverting 1319ada9, 3) Remove .el files from original repo.

Thanks!

PR #6013, opened 2019-02-09
Add recipe for djinni-mode

danielmartin:

Major mode for editing Djinni files. Djinni is an IDL for generating interfaces in C++, Objective C and Java. 

https://github.com/danielmartin/djinni-mode

I'm the creator and maintainer of this package.

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/djinni-mode
+++ b/recipes/djinni-mode
@@ -1,0 +1,2 @@
+(djinni-mode :repo "danielmartin/djinni-mode"
+             :fetcher github)


danielmartin:
Thanks for the review, @riscy! I've addressed your comments in https://github.com/danielmartin/djinni-mode/commit/f1716ba7dd81285f28c49ea555ff4885ef9f10c4

purcell:
Thanks, mostly looks great. Just a few little comments:
- You should provide a `defgroup` to define the `djinni` custom group you reference in your `defcustom`s
- You're making changes to the global values of `compile-command` and `compilation-error-regexp-alist` etc. every time a buffer is opened. I don't think that's your intention. Instead: a) make your own compilation command, and bind that to `C-c C-c` instead of `compile`, which will leave `compile` usable for other purposes; b) just modify `compilation-error-regexp-alist-alist` globally, and; c) add `djinni` to `compilation-error-regexp-alist` buffer-locally in your compilation buffer. There's a useful `define-compilation-mode` macro.

danielmartin:
Thanks for the review, @purcell ! I've addressed your comments in https://github.com/danielmartin/djinni-mode/commit/d5ba87e8beb4449647d46cb4e2d9cf5d681275a4

purcell:
Looks great, and I'll merge now. Note that you can (and probably should) just modify `compilation-error-regexp-alist-alist` at the top-level, because it's really just a bag of shorthand names for sets of patterns.

PR #6014, opened 2019-02-09
Add freeradius-mode for editing FreeRadius configuration files and un…

VersBinarii:
…lang policy files

Helper mode for FreeRadius server config files and unlang policies.

https://github.com/VersBinarii/freeradius-mode

I am maintainer.

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/freeradius-mode
+++ b/recipes/freeradius-mode
@@ -1,2 +1,1 @@
-(freeradius-mode :fetcher github :repo "VersBinarii/freeradius-mode
-")
+(freeradius-mode :fetcher github :repo "VersBinarii/freeradius-mode")
--- a/recipes/freeradius-mode
+++ b/recipes/freeradius-mode
@@ -1,0 +1,2 @@
+(freeradius-mode :fetcher github :repo "VersBinarii/freeradius-mode
+")


VersBinarii:
Hello and thanks for the feedback. 
Yes, you are correctly identified the config files to check :)
I believe i addressed everything you listed above. I plan to add couple of more features to the package as i pick up more elisp skills ;)

I would like to improve the regex matching and find a way to enable the mode automaticaly for the config files. The problem with this is that the files have no extension to run a regex on.
So i though about 2 approaches:
1. match on the directory, since by convention all config files must reside in raddb directory, and enable the mode for all files within it.
2. force user to activate the mode manually for the file initially and store the name of the file and trigger the mode the next time the file is opened.

Would you be able to advice if either of those approaches is "idiomatic" for emacs?

riscy:
Matching on the directory is probably your best bet here - I would maybe follow `nginx-mode` which does this as `("/nginx/.+\\.conf\\'" . nginx-mode)`.

VersBinarii:
Great. I'll check it out then.

purcell:
Ping! :-)

VersBinarii:
Yes, yes :) 
Thank you for all the feedback. I was just busy with some other stuff last few weeks.
I hope to sort it all out next week.

VersBinarii:
Hello @purcell,
 I believe i addressed most of the items from your feedback. Not sure about my indentation settings, i will need to investigate it a bit more. I will also try to improve the regexes a bit more over next few days.

purcell:
Cool, thanks!

PR #6015, opened 2019-02-10
add receipt for perfect-margin-mode

mpwang:

perfect-margin-mode is a global minor mode which will auto center windows, design for working with maximized frame or fullscreen.

Works good with linum-mode and try to cooperate with minimap-mode,  no impact if use don't install minimap or enable linum/minimap.

https://github.com/mpwang/perfect-margin

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/perfect-margin
+++ b/recipes/perfect-margin
@@ -1,0 +1,1 @@
+(perfect-margin :repo "mpwang/perfect-margin" :fetcher github)


mpwang:
I've fixed the style issues, added hints for compiler and added missing document string. Now `checkdoc` is happy and `byte-compile-file` results in clean output.

This is my first emacs package, although I've used it for many years by myself. 

Thanks for your advices!

purcell:
Thanks - before I comment on details of the code, have you seen `writeroom-mode` and the library `visual-fill-column` it uses? This package feels very similar.

terlar:
There is also `centered-window-mode` and `sublimity`. Just to add to the list. However I would be interested to see how this solves it in another way. As all those packages have their caveats.

mpwang:
This package is not aiming to provide distraction-free writing environment like `writeroom-mode` or `centered-window-mode`, which will keep only one window. 

I created this package because I want to keep a minimap window while keeping the main window in the centered position as if therer is no minimap window. And requirement to use minimap is not mandatory.

I've try `sublimity` before but I prefer to `minimap` as it can display `semantic` information nicely on the minimap window.  

Also I found a way to fix linum-mode [reset margin issue](https://github.com/rnkn/olivetti#known-bugs) when linum-monde is on.

I personally use this package for mainly 2 purposes:
- keep a minimap window on the left, while keeping the main window text centered
- When I use a very wide screen and split window vertically, text on both windows are centered.

The display effect with minimap is not perfect yet, I will work to improve it.

purcell:
Thanks! I took at look at the code and have a few notes:
- [This code with hard-coded rules](https://github.com/mpwang/perfect-margin/blob/d5ffa1ffcf38dc93e0887849a11bc19a96fcde00/perfect-margin.el#L135-L142) should be list of filter functions, each called with the window as its argument, so that it is more likely to work for other people
- [Here](https://github.com/mpwang/perfect-margin/blob/d5ffa1ffcf38dc93e0887849a11bc19a96fcde00/perfect-margin.el#L275-L277) you need a docstring and the `:lighter` for a minor mode must begin with a space. I'd advise not using the copyright symbol: just use something more obvious like `" PMargin"`, put it in a `perfect-margin-mode-lighter` custom variable, and then people (including you) can easily change it if they don't like it.
- The special handling of `linum` makes me think: there's also `nlinum` and `display-line-numbers-mode` (in recent Emacsen). Does this work with those? If not, it's worth noting that.

Hope that helps!

purcell:
Ping! :-)

mpwang:
sorry for not updating for a long time. 😓I plan to make it work with nlinum at the weekend. it's the only remaining issues.

mpwang:
I've refactored the code according to your suggestions, changed the hard-coded conditions and enabled customization on lighter. 

`display-line-numbers-mode` works fine together with this package. 

Spent quite some time on `nlinum-mode`, I've figured out how to make nlinum works with this package, but I can't come out a way without touching its source code (nlinum-format-function).

purcell:
Thanks. Sent you a couple of little fixes in https://github.com/mpwang/perfect-margin/pull/2. Should be fine to merge this once those are merged.

mpwang:
merged. really appreciate your help ^_^

purcell:
Welcome to MELPA! :tada:

PR #6016, opened 2019-02-10
add recipe org-pretty-tags

marcowahl:

Display text or image surrogates for org mode tags.

https://gitlab.com/marcowahl/org-pretty-tags

maintainer, contributor, user

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/org-pretty-tags
+++ b/recipes/org-pretty-tags
@@ -1,3 +1,2 @@
 (org-pretty-tags :repo "marcowahl/org-pretty-tags"
-           :fetcher gitlab
-           :files ("org-pretty-tags.el"))
+           :fetcher gitlab)
--- a/recipes/org-pretty-tags
+++ b/recipes/org-pretty-tags
@@ -1,0 +1,3 @@
+(org-pretty-tags :repo "marcowahl/org-pretty-tags"
+           :fetcher gitlab
+           :files ("org-pretty-tags.el"))


riscy:
And, to add, I like the idea.  I'd love to get them colorized too.

marcowahl:
# Checkdoc output:

{ M-x checkdoc RET } gives me

    Buffer comments and tags:  Not checked
    Documentation style:       Ok
    Message/Query text style:  Ok
    Unwanted Spaces:           Ok

Could you please point to the problematic sections?




# What are these comments all about?  They might confuse future contributors.

    ;; cache for the images
    ;; :PROPERTIES:
    ;; :ID:       fb26c0bc-a69e-4cd2-8b5a-800682d24706
    ;; :foo:      foo
    ;; :END:

The comment-line containing "foo" is trash, thanks for the catch.

The comment-lines starting with ":PROPERTIES:" allow a connection to
the actual source file.  Note that the org-pretty-tags program is
actually a literate program.  The actual source is file
"org-pretty-tags.org" (which is a copy of "readme.org" BTW.)

I tried to make this more clear with the new section "About this
Program" in the org-pretty-tags.org, better documentation in
org-pretty-tags.el and setting org-pretty-tags.el to view-mode at
load.



# Add \`(require 'org)\`

Thanks.



# Add \`(require 'subr-x)\` for \`when-let\`

Thanks.



# \`assert\` comes from \`cl.el\` which is deprecated - use cl-assert and \`(require 'cl-macs)\`

Thanks.



# Why is the group \`org-tags\` and not \`org-pretty-tags\`?

I find it appropriate to find the org-pretty-tags config stuff below
the org-tags config stuff.



# Could you make the minor-mode lighter " pretty tags" customizable?

Sure.  See org-pretty-tags-mode-lighter.


# \`"programm logic error."\` isn't very informative

Okay.  I extended the text some.



# programm should be program

Thanks.



# A couple comments on usage which may or may not be important:

-   If I'm editing an org buffer and I add a \`:music:\` tag, it doesn't
    show up as musical notes until I toggle the mode off and
    on. Intentional?
    -   With the current program structure it's too much CPU waste to get
        the surrogates right for every possible change of the tags.  I
        guess this could be done by hooking into every key-stroke event,
        but I did not check this.  At least when you CHANGE a tag via C-c
        C-c on a heading the tag surrogates get updated.
    
    -   See section "Hints" in file readme.org.

-   If I'm editing a tag that has been prettified (e.g. delete
    characters off the end of \`:imp:\`) my edits don't show up until I
    delete the whole tag.  There's probably a way to fix that.
    -   What do you think about using C-c C-c?

marcowahl:
# And, to add, I like the idea. I'd love to get them colorized too.

Thanks.  I did not think about colorization, yet.  This is potential for a new feature, I guess, maybe your contribution?

riscy:
Having another check, there's still an instance of `assert` instead of `cl-assert`

> Could you please point to the problematic sections?

The only way checkdoc should give you `Not checked` is if the buffer you're running it on has no filename, so you should check and see what's going on there.  Here are the particular issues:

- `org-pretty-tags-delete-overlays` has no docstring
- `org-pretty-tags-refresh-overlays-agenda` has no docstring
- `org-pretty-tags-refresh-overlays-org-mode` has no docstring
- `(error "function does not ...")`'s string argument should be capitalized: `(error "Function ...")`

Once I go through and fix these, re-running checkdoc finally nits about these comments:
```elisp
  "Container for the overlays of org-pretty-tags-mode."
  "Update `org-pretty-tags-image-cache' from list `org-pretty-tags-surrogate-images'."  
```
In the former case it wants quotes around `org-pretty-tags-mode` and disambiguation (is this a function, variable, etc).  In the latter case it wants disambiguation.  These are pedantic, but I usually try to reword my comments to make checkdoc happy so I don't have to see these.

> What do you think about using C-c C-c?

Neat, I didn't know about this function on headings.  I see there's a pointer in your `hints` section -- in that case, this looks fine.  I was thinking about the way latex blocks disappear if you delete part of them, but couldn't easily see how that's accomplished.

My last thought is on using a literate org program for a package -- and keep in mind I'm just a visiting commenter with an opinion.  I think you shouldn't do it:
- Those of us who install your package from MELPA are not necessarily going to have your org source file on hand, but they have to bear the burden of the anchor commentary marking up the code (and having the buffer be in view-mode)
- Packages already have a standard layout that I don't think the org markup/scaffolding adds to
- I'm going to go ahead and blame your issues with checkdoc (and the occasional misindented line) on this workflow; contributors to your work will also have to deal with this idiosyncratic workflow
- I think literate programming with org is akin to Jupyter notebooks -- it's fine for teaching or exploratory work, but not for finished packages

marcowahl:
Looks like there is indeed an issue with checkdoc here.

The issues you pointed out are solved now AFAICS.

As regards to the literate programming I want to stick with it for this package.  I think literate programming is a great idea in general and this little package is an opportunity to see literate programming in practice.

I kicked out the weird looking comments from the elisp result file (and also the switch to view-mode), following your advise.

Thanks

marcowahl:
Thanks for your review.

1. :file spec removed.

1. (require 'cl-lib) is in.

1. " pretty-tags" is the lighter now.

1. I changed org-pretty-tags-mode to be local for Org mode files.  I added a function to set the mode for all Org mode files at once.  No good idea for a global mode came into my brain.

At regeneration, e.g. with key "g", each Org agenda item gets the surrogates of its Org mode heading.

WDYT?

purcell:
Thanks (including to @riscy for the helpful initial review). I'm going to merge this, and it looks really good overall, but my review point above stands:

>     * it would be nicer (and would encourage more robust code) if you could split it into a local and global minor mode pair.

Specifically, you're hand-rolling an unidiomatic [partial version of a real global minor mode here](https://gitlab.com/marcowahl/org-pretty-tags/blob/caf18c06b20a6e5386db8a8ec0e8508d474973ed/org-pretty-tags.el#L238). Just make use `define-globalied-minor-mode`, and you'll get a better, more conventional result!

marcowahl:
Thanks.

I'll try to switch to `define-global-minor-mode`.  I simply somehow was agnostic of the latter function.

PR #6017, opened 2019-02-10
Adding chronometer receipe.

marcelotoledo:

Chronometer is a major mode to help you keep track of time. It has an alarm system and can warn you whenever a certain period of time has passed directly in GNU Emacs.

https://github.com/marcelotoledo/chronometer

I am the Author.

First submission.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/chronometer
+++ b/recipes/chronometer
@@ -1,0 +1,1 @@
+(chronometer :fetcher github :repo "marcelotoledo/chronometer")


marcelotoledo:
@riscy thank you for all the commentaries.

- Checkdoc ok!
- Docstring of `chronometer-help` fixed (thanks for the tip!)
- All other Docstrings fixed.

marcelotoledo:
Thank you so much @purcell for all the great inputs!

Applied all the changes but the last. Comments bellow:

- Why not use not here? ✅ 
- Minor typo in "automatically" here ✅ 
- Use define-derived-mode (with special-mode as the parent) here instead of rolling your own mode function like in the bad old days. ;-) You'll be able to remove these lines if you do that because define-derived-mode handles it for you. ✅ 
- You don't need a progn inside when ✅ 
- Some of the defvars should probably be defcustoms in a :group 'chronometer defined with defgroup. 💬 They are all internal controlled variables, users really shouldn't interfere with it. But thanks for double-checking it.

purcell:
Cool, merged now, thanks. And sorry for the time it's taken. :-)

PR #6022, opened 2019-02-13
Adding helm-taskswitch

bdc34:

Provides an X window and buffer switching with helm.

https://github.com/bdc34/helm-taskswitch

I am the maintainer. This is my first MELPA package. 

Maintainer would like package on MELPA.

### Checklist

Please confirm with `x`:

- [ x] `M-x checkdoc` is happy with my docstrings (except warning about first line summary but I have lexical-binding there. )

### DIFF
--- a/recipes/helm-taskswitch
+++ b/recipes/helm-taskswitch
@@ -1,0 +1,1 @@
+(helm-taskswitch :fetcher github :repo "bdc34/helm-taskswitch")


bdc34:
Thanks for your help. I've made these changes here: [Commit](https://github.com/bdc34/helm-taskswitch/commit/25e8bf2fcdce2afc3c5197399f5192246ff244f1)

purcell:
>     * I think the magic `lexical-binding` comment would also work on the second line

Nope, that should always go on the first line, odd though that is -- there was at least one Emacs version which didn't correctly apply it if it was not on the first line.

bdc34:
Thanks, I've moved the lexical-binding to the first line. [Commit](https://github.com/bdc34/helm-taskswitch/commit/e74ae4a8e4e037d8b719e287da6f08fcae156b9f)

purcell:
I took a closer look and sent you a fix in https://github.com/bdc34/helm-taskswitch/pull/1, but I _think_ I can go ahead and merge now anyway. Welcome to MELPA! :tada:

P.S. You should `:inherit` built-in faces where possible instead of choosing your own colours. This allows your package to look reasonable in various themes.

PR #6024, opened 2019-02-16
Add org-reverse-datetree

akirak:

Reverse date trees for Org mode

https://github.com/akirak/org-reverse-datetree

I am the maintainer

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/org-reverse-datetree
+++ b/recipes/org-reverse-datetree
@@ -1,0 +1,1 @@
+(org-reverse-datetree :fetcher github :repo "akirak/org-reverse-datetree")


noctuid:
Thanks so much for this! I've wanted this functionality for a long time.

akirak:
Thank you, I've fixed those issues.

It seems that use of `cl-macs` is discouraged in the recent versions of Emacs. I encountered a linting error by adding its import.

purcell:
> It seems that use of `cl-macs` is discouraged in the recent versions of Emacs. I encountered a linting error after adding its import, so I didn't add it.

Yes, that's right: it's best to just use `cl-lib`, which is available in both the modern Emacs standard library and the `cl-lib` package on ELPA for older Emacsen.

In any case, this all looks good, so I'll go ahead and merge now. Thanks!

PR #6026, opened 2019-02-16
Add org-recur

m-cat:

This package extends org-mode and org-agenda with support for defining recurring tasks.

https://github.com/m-cat/org-recur

**None needed*

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/org-recur
+++ b/recipes/org-recur
@@ -1,0 +1,1 @@
+(org-recur :fetcher github :repo "m-cat/org-recur")


m-cat:
Awesome, thanks for the feedback! Should all be fixed now.

> Is it worth making `org-recur--weekday-recurrence` customizable (and maybe `org-recur--weekday`) so that users can potentially support different languages?

Great question. I don't know if org-mode supports different languages or if "mon", "wed" etc. are hard-coded. But there's at least one good reason to make these variables customizable: not everyone works or studies on the same days. Some people don't work Wednesdays, others work Saturdays. Instead of making them type `|mon,tue,wed,thu,fri,sat|` for all of their work tasks, I think it makes sense to make this customizable, so I went ahead and did so.

dzop:
This looks cool. I'm wondering if it can be done using [sexp timestamp entries](https://orgmode.org/guide/Timestamps.html) in `org-mode`? For example you can get the `|1,15|` recurrence with `SCHEDULED: <%%(diary-date t '(1 15) 2019)>` (see `C-h f diary-date`). 

I haven't looked at all the features your package has so there may be cases it solves that sexp entries don't.

m-cat:
The biggest shortcoming of that approach, it seems to me, is the complexity of it. You also can't see the timestamp from the org-agenda view. I added it to the Alternatives section of the Readme :)

dzop:
Just to note, you can put a time of day with `SCHEDULED: <%%(diary-date t '(1 15) 2019) 10:00>`

m-cat:
Ah, if you want to do that with org-recur, the equivalent is this: `|1 10:00, 15 10:00|`. So it's a bit more verbose, but on the other hand, you can specify different times of day for each date :) If you don't want the verbosity you can move the time of day outside of org-recur: `|1, 15| 10:00 headline` works just fine, and is what I have been doing. org-agenda will pick up the time in either scenario.

m-cat:
I added a note about that to the [Readme](https://github.com/m-cat/org-recur#time-of-day) ;)

m-cat:
I noticed a bug with the aforementioned syntax actually: the time was being ignored. You could reproduce it with something like `|10:00, 9:00, 11:00|`. I pushed a commit fixing this.

Unfortunately, it seems like there's another bug, though this time I would blame org-mode itself. When the current date is March 1 and you try to reschedule to `1`, it should pick April 1 if the `org-read-date-prefer-future` variable is set (which it is by default). However, `org-read-date-analyze` will reschedule it to the current day, March 1, which is clearly not what you'd ever want. I can fix this by copy/pasting the entirety of `org-read-date-analyze` into org-recur, but I'm wondering if there's a better and more future-proof way to patch this function.

Also note that if you reschedule to `Fri`, `org-read-date-analyze` *will* pick next Friday, and this is clearly what *should* happen, so I declare that this is a bug in org-mode.

purcell:
Thanks! Some notes:

- If you can include some helpful info in the `Commentary` section, that's better than directing people to your README, because that text appears on the MELPA page and in `package.el`'s "?" buffer.
- [faces](https://github.com/m-cat/org-recur/blob/b430cb0db5295e1a7af806df9ab0ca60b2c38348/org-recur.el#L52) doesn't seem like the right parent custom group. :smile:
- Don't forget that the body of your `define-minor-mode` is also called when the mode is turned off. To handle the toggle, use an `if` to check the `org-recur-mode` var in that block.
- Special prize for checking `font-lock-mode` [here](https://github.com/m-cat/org-recur/blob/b430cb0db5295e1a7af806df9ab0ca60b2c38348/org-recur.el#L217) :trophy:

Hope that helps!

m-cat:
Awesome, thanks for the feedback! I think I fixed all of those issues, plus a few more I found having to do with turning on/off the mode in org-agenda. To do that, I added `org-recur-agenda-mode`, which was needed anyway -- the highlighting scheme is different in org-agenda, and users may want to add different bindings to the keymaps (as in the suggested config in the Readme). Speaking of which, I added the missing keymaps 😄 

Hope that all looks good, let me know if you have any more feedback!

purcell:
Great, looking good. Couple of points:
- You're [enabling a different global minor mode](https://github.com/m-cat/org-recur/blob/69e5f2e7bf66d9afa0b807c88451b092b657bad7/org-recur.el#L241) in your local minor mode. That can't be right...? Recommend you remove that.
- `major-mode` isn't normally [a string](https://github.com/m-cat/org-recur/blob/69e5f2e7bf66d9afa0b807c88451b092b657bad7/org-recur.el#L181). In any case, you should generally check major modes using `derived-mode-p`.
- In any case, almost every line in `org-recur-finish` is different according to whether the buffer is an agenda or not. Better to have 2 separate small easy-to-read functions which you call conditionally.
- Consider [`unless` here](https://github.com/m-cat/org-recur/blob/69e5f2e7bf66d9afa0b807c88451b092b657bad7/org-recur.el#L106) and deleting the cond clause [here](https://github.com/m-cat/org-recur/blob/69e5f2e7bf66d9afa0b807c88451b092b657bad7/org-recur.el#L129).

m-cat:
Thanks! Pushed another fix.

> You're enabling a different global minor mode in your local minor mode. That can't be right...? Recommend you remove that.

Hmm, I was getting a warning when loading org-recur: something about `highlight-regexp` requiring `hi-lock-mode`, and to turn on `global-hi-lock-mode` in Emacs 22 and up. But I removed the line and I can't reproduce the warning now. Weird.

Just found the [exact warning](https://github.com/jkitchin/org-ref/issues/16) in Google. But I don't get it anymore. No idea what's going on. I'll admit that I find that warning message incomprehensible.

purcell:
Great, merged now!

PR #6027, opened 2019-02-17
add recipe for lsp-python-ms

andrew-christianson:

This package implements an lsp-mode client for Microsoft's python
languager server.

https://github.com/andrew-christianson/lsp-python-ms

I am a maintainer of the package.  Code initially authored by @cpbotha

**none-needed**
### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/lsp-python-ms
+++ b/recipes/lsp-python-ms
@@ -1,0 +1,1 @@
+(lsp-python-ms :fetcher github :repo "andrew-christianson/lsp-python-ms")


andrew-christianson:
Thanks for the comments, I'll make the suggested changes.

MatthewZMD:
Any updates?

muffinmad:
There are https://github.com/xhcoding/ms-python already on melpa which provides same feature. Maybe it would be better to improve `ms-python` package if it missing some features of this package?

purcell:
> There are https://github.com/xhcoding/ms-python already on melpa which provides same feature. Maybe it would be better to improve `ms-python` package if it missing some features of this package?

Yes, agree. How about it, @xhcoding and @andrew-christianson? What would be the best way to get a single great package which works for (at least) both of you?

cpbotha:
Thank you very much @muffinmad and @purcell for the time you invest in this too.

This was discussed with @xhcoding and @andrew-christianson in this issue: https://github.com/andrew-christianson/lsp-python-ms/issues/9

In short, both works started in parallel, unbeknownst to each other, based on my blog post. @xhcoding was quick on the melpa trigger :) but @andrew-christianson 's version ended up getting substantially more development effort.

We all, including @xhcoding , thought it would be best to fold his work into @andrew-christianson 's version, which is the only one working with the new Emacs LSP support.

What would be the best way to do this?

purcell:
> What would be the best way to get this version to be the canonical one on Melpa?

Firstly I'd merge this (done!). Then I'd be happy to merge a further PR removing @xhcoding's package.

I'd also note that this new package would benefit from a little bit of polishing: in particular, turn the option vars into `defcustom`s where possible, with a containing `defgroup`. :-)

PR #6028, opened 2019-02-17
Add Emacs reveal

lechten:

The software emacs-reveal allows to create HTML presentations (with audio explanations if you wish) with reveal.js from Org mode files in GNU Emacs with org-re-reveal (a fork of org-reveal) and several reveal.js plugins. Generated presentations are usable with standard browsers, also mobile and offline.

This is a continuation of PR #5901. Compared to that PR, reveal.js and its submodules have now been moved to a separate repository. Also, org-reveal has turned into org-re-reveal. Its extension org-re-reveal-ref is a required package for emacs-reveal.

https://gitlab.com/oer/emacs-reveal

Maintainer.

**None needed**

### Checklist

Please confirm with `x`:

 It claims that I need Emacs 25, which I don't need as explained in comments underneath my Package-Requires header.
 There are warnings about obsolete functions, which are not used.

### DIFF
--- a/recipes/emacs-reveal
+++ b/recipes/emacs-reveal
@@ -1,0 +1,5 @@
+(emacs-reveal
+ :repo "oer/emacs-reveal"
+ :branch "melpa"
+ :fetcher gitlab
+ :files (:defaults "org-reveal" "org" "css" "title-slide" "reveal.js" "reveal.js-plugins" "Reveal.js-TOC-Progress" "reveal.js-jump-plugin" "reveal.js-quiz"))
--- a/recipes/emacs-reveal
+++ b/recipes/emacs-reveal
@@ -2,4 +2,4 @@
  :repo "oer/emacs-reveal"
  :branch "melpa"
  :fetcher gitlab
- :files (:defaults "org-reveal" "org" "css" "title-slide" "reveal.js" "reveal.js-plugins" "Reveal.js-TOC-Progress" "reveal.js-jump-plugin" "reveal.js-quiz"))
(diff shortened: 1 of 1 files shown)


lechten:
A quick note: I just committed a new version, which contains the new file `emacs-reveal-publish.el` to simplify publication of presentations by others.

lechten:
@purcell Many thanks for this detailed and constructive feedback, which exceeds my hopes by far! I really appreciate the work that you put into this.

I'm currently travelling, so a real answer will take some time.

lechten:
I just opened the new PR #6063 for `oer-reveal`.  In that PR, I addressed your above feedback as follows:

Naming: You are right, emacs-reveal is not a good name for an Emacs package.  Still, I want to keep that name for people who might not be using Emacs yet.  My plan is to have most code in the new package oer-reveal, while emacs-reveal will activate/change custom settings across various packages.

The file `install.el` is removed, as is obsolete code and version information.  Actually, why are self-reporting package versions not needed?  What is the simplest way for a user to find the version, say when reporting a bug?

I’m now using `shell-quote-argument`, the new regexp, `expand-file-name`, `\n`.

File `emacs-reveal-publish.el` was meant for execution in batch mode.  First, themes are not available then (or I’m doing something wrong), so I forced color changes to enable syntax highlighting.  Anyways, in the new PR, this and other changes to the user’s session are moved into functions that need to be called explicitly.  Second, when using `emacs-reveal-publish.el` in batch mode, it starts in a controlled directory, not a random one.  With `oer-reveal`, the location of `oer-reveal-publish.el` does not matter, but files for publication need to be available in the current directory.  This works in batch mode and when exporting an Org file from the current buffer where the directory of the Org file may contain other resources to be published.

Thanks again!

PR #6029, opened 2019-02-17
Add crontab-mode

marsam:

Major mode for `crontab(5)`

https://github.com/emacs-pe/crontab-mode

maintainer

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/crontab-mode
+++ b/recipes/crontab-mode
@@ -1,0 +1,1 @@
+(crontab-mode :repo "emacs-pe/crontab-mode" :fetcher github)


marsam:
Thanks for the heads up; but It does not share code from that mode.
> Needs a (defgroup crontab ...)

fixed in https://github.com/emacs-pe/crontab-mode/commit/9a611bc
> Since you're requiring Emacs 24.3[...]

fixed in https://github.com/emacs-pe/crontab-mode/commit/090ed61

purcell:
Great!

PR #6031, opened 2019-02-19
Add  jk-org2md-hexo package.

loveminimal:
### Brief summary of `jk-org2md-hexo`

A package render `.org` to `.md` which can be rendered by HEXO. 

### Github Repo

https://github.com/loveminimal/jk-org2md-hexo

### About Me

Hello, I'm Jack Liu. 
I build it just for sloving the problem of org's rendering. Now, it make it easies.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/jk-org2md-hexo
+++ b/recipes/jk-org2md-hexo
@@ -1,0 +1,3 @@
+(jk-org2md-hexo 
+  :fetcher github
+  :repo "loveminimal/jk-org2md-hexo")


purcell:
Hi! Thanks for submitting this. It's not yet in shape to be a useful package for wider use, so I'll leave you with some notes to consider:

- The personal `jk-` shouldn't be necessary if this is a package for general use
- All the top-level function/var names must start with the package name, and no "/" separators. (`package-lint` should have warned you about this: did you really run it correctly?) Currently I'd expect to see your main function called `jk-org2md-hexo-md-export`, for example.
- [This](https://github.com/loveminimal/jk-org2md-hexo/blob/bb9a77ae5a76426da397cfcc8e4d747ed999cfbf/jk-org2md-hexo.el#L17-L19) isn't the right way to do package dependencies. Delete that code and look up the `Package-Requires` header format and declare a proper dependency. Note also that it should be possible to simply put both this library and `ox-gfm.el` in a directory on `load-path` and then load them, but your code snippet interferes with this.
- You are using `setq` to implicitly define lots of global variables. That's not the lisp way to do things: use `let` instead.
- Saving and killing the buffer shouldn't be part of the command.

Overall I think this needs to be rewritten as a proper org export backend, using org-gfm as needed. Otherwise this just seems like a function that belongs in your personal config. As such, I'll close this current pull request now, but if you have any questions, just leave a comment here and I'll do my best to help.

loveminimal:
> Hi! Thanks for submitting this. It's not yet in shape to be a useful package for wider use, so I'll leave you with some notes to consider:
> 
> * The personal `jk-` shouldn't be necessary if this is a package for general use
> * All the top-level function/var names must start with the package name, and no "/" separators. (`package-lint` should have warned you about this: did you really run it correctly?) Currently I'd expect to see your main function called `jk-org2md-hexo-md-export`, for example.
> * [This](https://github.com/loveminimal/jk-org2md-hexo/blob/bb9a77ae5a76426da397cfcc8e4d747ed999cfbf/jk-org2md-hexo.el#L17-L19) isn't the right way to do package dependencies. Delete that code and look up the `Package-Requires` header format and declare a proper dependency. Note also that it should be possible to simply put both this library and `ox-gfm.el` in a directory on `load-path` and then load them, but your code snippet interferes with this.
> * You are using `setq` to implicitly define lots of global variables. That's not the lisp way to do things: use `let` instead.
> * Saving and killing the buffer shouldn't be part of the command.
> 
> Overall I think this needs to be rewritten as a proper org export backend, using org-gfm as needed. Otherwise this just seems like a function that belongs in your personal config. As such, I'll close this current pull request now, but if you have any questions, just leave a comment here and I'll do my best to help.

Thanks, sir. It's my first pull-request, so many aspects is week. Hah, your notes is so detailed that I can fixed it.

In a way, thank you, sir. : )

PR #6033, opened 2019-02-19
add greek-polytonic recipe

jhanschoo:

Registers an input method for polytonic Greek to quail (emacs' i18n builtin package)

https://github.com/jhanschoo/greek-polytonic

Package author & maintainer

**None needed** (I hope)

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/greek-polytonic
+++ b/recipes/greek-polytonic
@@ -1,0 +1,2 @@
+(greek-polytonic :fetcher github
+                 :repo "jhanschoo/greek-polytonic")


jhanschoo:
Thanks. I have made the requested change.

Autoloading was not my original intention, but I was persuaded by both of the following:

* I had thought that `quail.el` was always loaded on startup, and associated input methods registered at startup as well.
* The blog post https://swsnr.de/blog/2014/07/02/autoloads-in-emacs-lisp/ advises that
> If your package should automatically register itself in specific Emacs extension points you should add autoloads for these as well to make sure that they are evaluated during package initialization. A typical example is adding a mode to auto-mode-alist [...]

But some poking showed that `quail.el` indeed is not always loaded at startup, so autoloading is indeed unacceptable.

purcell:
Great, thanks for digging into that. Merged now - welcome to MELPA! :tada:

PR #6037, opened 2019-02-19
Added poly-rst recipe.

Xaldew:

This is a [polymode](https://polymode.github.io/) for working with files using the ReST markup syntax. The functionality it provides is similar to [poly-markdown](https://github.com/polymode/poly-markdown), in that regions marked with `code` syntax in ReST will use the appropriate Emacs major mode in those regions.

https://github.com/polymode/poly-rst

I'm co-maintaining the package along with @vspinu.

**None needed**

### Checklist

Please confirm with `x`:

### Misc. Notes

Package-lint will complain about at least one variable, namely `pm-poly/rst`. This appears to be a convention within `polymode`, so I'm afraid that cannot be avoided right now at least.

### DIFF
--- a/recipes/poly-rst
+++ b/recipes/poly-rst
@@ -1,0 +1,3 @@
+(poly-rst
+ :fetcher github
+ :repo "polymode/poly-rst")


Xaldew:
Thanks for the review, totally missed a few of those issues, all fixed now!

As for the `defgroup`s, they already exist in the parent package [polymode](https://github.com/polymode/polymode/blob/master/polymode-core.el), do I still need to include them?

purcell:
Looks good now, so merging. Sorry it has taken a while.

PR #6040, opened 2019-02-22
add maple-package recipe

honmaple:

Speed up emacs startup and add some package functions.

https://github.com/honmaple/emacs-maple-package

maintainer

None needed

### Checklist

Please confirm with `x`:


purcell:
Thanks for submitting this, it looks interesting. However, the only clue about what the package does is the phrase "package configuration". Why would anyone use this? What can they do with it? It's not helpful to publish this for the community if nobody can understand why it might be useful to them.

PR #6041, opened 2019-02-22
Add eldoc-cmake

ikirill:
Support for eldoc when editing CMake files.

Add eldoc-documentation-function that understands CMake functions.

https://github.com/ikirill/eldoc-cmake

I wrote it, it is a simple short one-file package, with hard-coded pre-extracted docstrings for CMake functions.

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/eldoc-cmake
+++ b/recipes/eldoc-cmake
@@ -1,4 +1,3 @@
 (eldoc-cmake
  :fetcher github
- :repo "ikirill/eldoc-cmake"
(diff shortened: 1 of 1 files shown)


purcell:
Thanks! The `:files` spec here is redundant, but I can fix that on this side. Additionally you should add `(require cl-lib)` to be explicit, but you can fix that in your own time and I'll go ahead and merge for now.

PR #6042, opened 2019-02-23
Add bibclean-format recipe

peterwvj:

Commands for prettyprinting and syntax checking BibTeX and Scribe bibliography data base files using bibclean.

https://github.com/peterwvj/bibclean-format

I am the maintainer.

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/bibclean-format
+++ b/recipes/bibclean-format
@@ -1,0 +1,1 @@
+(bibclean-format :fetcher github :repo "peterwvj/bibclean-format")


purcell:
Yay, a use of `reformatter.el` out in the wild! Merging now, but you should fix that the `defgroup` currently claims to be its own parent. :-)

peterwvj:
`reformatter.el` is awesome! Thanks a lot for your feedback! I've updated my package based on your suggestions.

PR #6047, opened 2019-03-01
Added torus recipe

chimay:
Torus let you organize your buffers by creating as many buffer groups
as you need, and quickly navigate between :

* Buffers of the same group
* Buffer groups (circles)

https://github.com/chimay/torus

I’m the only contributor and maintainer of this package.

**None needed**

### Checklist

Please confirm with `x`:

* [x]  The package is released under a [GPL-Compatible Free Software License](https://www.gnu.org/licenses/license-list.en.html#GPLCompatibleLicenses).
* [x]  I've read [CONTRIBUTING.org](https://github.com/melpa/melpa/blob/master/CONTRIBUTING.org)
* [x]  I've used the latest version of [package-lint](https://github.com/purcell/package-lint) to check for packaging issues, and addressed its feedback
* [x]  My elisp byte-compiles cleanly
* [x]  `M-x checkdoc` is happy with my docstrings
* [x]  I've built and installed the package using the instructions in [CONTRIBUTING.org](https://github.com/melpa/melpa/blob/master/CONTRIBUTING.org)

### DIFF
--- a/recipes/torus
+++ b/recipes/torus
@@ -1,4 +1,2 @@
-;;; -*- mode: emacs-lisp; -*-
-
 (torus :repo "chimay/torus"
        :fetcher github)
--- a/recipes/torus
+++ b/recipes/torus
@@ -1,0 +1,4 @@
+;;; -*- mode: emacs-lisp; -*-
+
+(torus :repo "chimay/torus"
+       :fetcher github)


chimay:
Thanks for your notes, it’s my first real lisp code, so it helps !
For instance, I searched for "elisp case" and found  the "(cond ...)" structure.
Anyway, the torus file is updated.

purcell:
So much better! A bunch more notes, mainly related to style:

- [Prefer `when` here](https://github.com/chimay/torus/blob/26eefd161d6c51d309a41dafbbc746ba34503f0c/torus.el#L260) and [here](https://github.com/chimay/torus/blob/26eefd161d6c51d309a41dafbbc746ba34503f0c/torus.el#L234-L235). Just in general, if you have only one branch, use `when` or `unless`.
- You can use `pcase-let` (or just `let`) to bind repeated expressions and simplify code like [this](https://github.com/chimay/torus/blob/26eefd161d6c51d309a41dafbbc746ba34503f0c/torus.el#L260-L267)
- You'll potentially have a bad time [extracting format strings into constants](https://github.com/chimay/torus/blob/26eefd161d6c51d309a41dafbbc746ba34503f0c/torus.el#L482-L483), because you'd have to look back and forth between where the string is defined and where it is used to check that the same number of arguments and placeholders are supplied. Better to just inline the strings.
- Don't `let`-bind a var and then [immediately `setq` it](https://github.com/chimay/torus/blob/26eefd161d6c51d309a41dafbbc746ba34503f0c/torus.el#L489-L490): just initialise it with its initial value in the `let` expression. Even [here](https://github.com/chimay/torus/blob/26eefd161d6c51d309a41dafbbc746ba34503f0c/torus.el#L570-L580) you can use `pcase-let` to set `name` and `args` together from a `cons` returned by a pcase expression like this one, so you can avoid the nil initialisation and subsequent `setq`s.
- You can delete the [`torus-version` var](https://github.com/chimay/torus/blob/26eefd161d6c51d309a41dafbbc746ba34503f0c/torus.el#L59-L60) -- in these days of package.el, it's more reliable to look at the installed package version anyway, so tracking versions internally is a little outdated.

None of these are blockers to merging, but in the interest of making the code as idiomatic as possible - and learning more about elisp - I encourage you to investigate them. :-)

Merging now, then - welcome to MELPA! :tada:

chimay:
Nice, I’ll investigate pcase-let usage, it sounds really powerful.
It’s great to be here, thanks for your patience !

PR #6050, opened 2019-03-03
Add separate

ROCKTAKEY:

This package help you to separate emacs setting by system environment.

https://github.com/ROCKTAKEY/pc-separate

I'm a maintainer of this package.

**None needed**

### Checklist

Please confirm with `x`:


alphapapa:
There are a few issues with this package:

1.  The name is not very descriptive.
2.  It reimplements a lot of existing macros and functions in a non-idiomatic way, e.g. the `separate--mapc-and` function.  I recommend looking at the `dash` package and the `pcase` macro.
3.  The coding style is not idiomatic elisp.  Please check the resources in `CONTRIBUTING.org` and apply them accordingly.

ROCKTAKEY:
First, I'm really sorry to trouble you.

I understood 1. and 2. and fixed the problems, but I couldn't find problems about 3.. Would you mind telling me specific problems?

And, I mistake a order of commits, so check of Travis CI have failed. (I should have commit of my repo before melpa's commit.) So I'd like you to carry out the check again.

Last, I change package name to `system-separate`, but this name is too long to use a prefix of macros, such as `setq` and `cond`, because long macro name reduces those convenience. So I would like to use "ss-" as a prefix instead of "system-separate-", though package-lint regard those as error. If not permitted, I will change prefix to "system-separate-".

I'm terribly sorry to trouble you again and to write long message. Thanks.

alphapapa:
> I understood 1. and 2. and fixed the problems, but I couldn't find problems about 3.. Would you mind telling me specific problems?

Please search `CONTRIBUTING.org` for `coding style` and read the linked guides.

> And, I mistake a order of commits, so check of Travis CI have failed. (I should have commit of my repo before melpa's commit.) So I'd like you to carry out the check again.

What check do you mean?

> Last, I change package name to system-separate, but this name is too long to use a prefix of macros, such as setq and cond, because long macro name reduces those convenience. So I would like to use "ss-" as a prefix instead of "system-separate-", though package-lint regard those as error. If not permitted, I will change prefix to "system-separate-".

1.  Using abbreviated prefixes instead of the package name is against MELPA policy.
2.  I haven't studied your code carefully, but you probably should not need those macros, especially the `cond` one.  Remember that in lisp, everything is an expression, so you can do, e.g.:

```el
(setq var (pcase (system-name)
            ("foo" 'foo)
            ("bar" 'bar)))
```
You could also do something like:
```el
(pcase (system-name)
  ("foo" (setq var1 'foo
               var2 'foo2))
  ("bar" (setq var1 'bar
               var2 'bar2)))
```
So, for example, most of what your `setq` macro does can be done with `pcase`.  You could write some simple extensions to `pcase` (see `pcase`'s implementation's source code) to check system type, etc, so instead of:
```el
(ss-setq foo
  (;; (separator . value)
   (pc-name1      . 1)
   (pc-name2      . 2)
   ("MY-PC-NAME3" . 3)
   (windows-nt    . 4)
   ((:eval (something) (bar))       . 5)
   ((:package-available 'baz)       . 6)
   ((:and windows-nt (:eval (def))) . 7)
   ((:separators "MY-PC-NAME4" gnu/linux) . 8)
   (default . 0)))
```
You could do something like:
```el
(setq foo (pcase (system-info)
            ((name pc-name1) 1)
            ((name pc-name2) 2)
            ((name "MY-PC-NAME3") 3)
            ((os windows-nt) 4)
            ((guard (progn
                      (something)
                      (bar)))
             5)
            ((package-available 'baz) 6)
            ((and (os windows-nt) (pred def)) 7)
            (_ 0)))
```

ROCKTAKEY:
> Please search CONTRIBUTING.org for coding style and read the linked guides.

I've read it twice, but I've overlooked some. I'm sorry. I modified them.

> What check do you mean?

The check of Travis CI which started automatically after this pull request. It might be shown below this comment.

> Using abbreviated prefixes instead of the package name is against MELPA policy.

OK, I got it.  I'm sorry for asking so much of you.

> I haven't studied your code carefully, but you probably should not need those macros, especially the cond one.

This package provides feature to help you to write setting of each environment such as OS,  feature availability, Emacs's version and so on, **without writing environmental conditions repeatedly**. For example, suppose that you write on top of `init.el` below:

``` emacs-lisp
(setq ss-separator-alist '((my-pc1 . (:and windows-nt "hoge"))
                           (my-pc2 . (:and gnu/linux  "fuga" (:package-available 'init-loader)))))
```

Now, on the feature `system-separate`,
symbol of  `my-pc1` is regarded as an environment like "The PC is on Windows, and the PC name is "hoge"".
And symbol of `my-pc2` is regarded as an environment like "the PC is on Linux, the PC name is "fuga", and the feature `init-loader` is available."

So, once you define `my-pc1` and `my-pc2` like above, you can use this symbol instead of writing same condition repeatedly. This is the main advantage of the feature `system-separate`. For example, you can write:

```emacs-lisp
(ss-setq foo ((my-pc1 . 1) (my-pc2 . 2) (default . 3)))

;; ...

;; On other part or file
(ss-setq bar ((my-pc1 . 4) (my-pc2 . 5) (default . 6)))

;; ...
```

instead of:

```emacs-lisp
(setq foo
      (cond ((and (eq system-type 'windows-nt) (string= (system-name) "hoge")) 1)
            ((and (eq system-type 'gnu/linux)  (string= (system-name) "fuga")) 2)
            (t 3)))

;; ...

;; On other part or file
(setq bar
      (cond ((and (eq system-type 'windows-nt) (string= (system-name) "hoge")) 4)
            ((and (eq system-type 'gnu/linux)  (string= (system-name) "fuga")) 5)
            (t 6)))

;; ...
```

Of cource, each S expression above can be written with other function or macro, such as `pcase`. However, this feature make you avoid writing environmental condition repeatedly. This is the main purpose. Other forms, such as your examples, are incidencally prepared in order to use normal conditions with the conditions of symbols like `my-pc1` in same S expression.

This is also true for `ss-cond`. This macro is for the case like:

```emacs-lisp
(setq some-directory
      (expand-filename
        (ss-cond (my-pc1 "hoge1") ("hoge" "hoge2") (my-pc2 "fuga")) use-emacs-directory))
```

In addition, I think branching value by `(system-name)` is frequent for those who share `init.el` with some PCs. So string is regarded as a condition of value of `(system-name)` on this feature. I think this is one of advantages of `system-separate`, too.

alphapapa:
> > Please search CONTRIBUTING.org for coding style and read the linked guides.
> 
> I've read it twice, but I've overlooked some. I'm sorry. I modified them.

Looks much better, thanks.

> > What check do you mean?
> 
> The check of Travis CI which started automatically after this pull request. It might be shown below this comment.

I don't have control over that.  I'm not a MELPA staff member, just a contributor who tries to help out by doing some initial reviews of new packages.

> > Using abbreviated prefixes instead of the package name is against MELPA policy.
> 
> OK, I got it. I'm sorry for asking so much of you.

Not at all.  The policy isn't up to me, anyway, I'm just informing you.

> > I haven't studied your code carefully, but you probably should not need those macros, especially the cond one.
> 
> This package provides feature to help you to write setting of each environment such as OS, feature availability, Emacs's version and so on, **without writing environmental conditions repeatedly**. For example, suppose that you write on top of `init.el` below:
> 
> ```emacs-lisp
> (setq ss-separator-alist '((my-pc1 . (:and windows-nt "hoge"))
>                            (my-pc2 . (:and gnu/linux  "fuga" (:package-available 'init-loader)))))
> ```
> 
> Now, on the feature `system-separate`,
> symbol of `my-pc1` is regarded as an environment like "The PC is on Windows, and the PC name is "hoge"".
> And symbol of `my-pc2` is regarded as an environment like "the PC is on Linux, the PC name is "fuga", and the feature `init-loader` is available."
> 
> So, once you define `my-pc1` and `my-pc2` like above, you can use this symbol instead of writing same condition repeatedly. This is the main advantage of the feature `system-separate`. For example, you can write:
> 
> ```emacs-lisp
> (ss-setq foo ((my-pc1 . 1) (my-pc2 . 2) (default . 3)))
> 
> ;; ...
> 
> ;; On other part or file
> (ss-setq bar ((my-pc1 . 4) (my-pc2 . 5) (default . 6)))
> 
> ;; ...
> ```
> 
> instead of:
> 
> ```emacs-lisp
> (setq foo
>       (cond ((and (eq system-type 'windows-nt) (string= (system-name) "hoge")) 1)
>             ((and (eq system-type 'gnu/linux)  (string= (system-name) "fuga")) 2)
>             (t 3)))
> 
> ;; ...
> 
> ;; On other part or file
> (setq bar
>       (cond ((and (eq system-type 'windows-nt) (string= (system-name) "hoge")) 4)
>             ((and (eq system-type 'gnu/linux)  (string= (system-name) "fuga")) 5)
>             (t 6)))
> 
> ;; ...
> ```
> 
> Of cource, each S expression above can be written with other function or macro, such as `pcase`. However, this feature make you avoid writing environmental condition repeatedly. This is the main purpose. Other forms, such as your examples, are incidencally prepared in order to use normal or one-time conditions with the conditions of symbols like `my-pc1` in same S expression.

Thanks, that helps me understand it better.

So, two issues that are worth considering now, I think:

1.  The naming of the conditions as "separators."  I'm guessing that English isn't your first language, so I don't blame you for the confusion, and naming things is known to be hard.  ;)  As a native English speaker, the word "separator" doesn't seem appropriate or descriptive to me.  I would recommend something else, perhaps "system definition", as in "the code that defines whether a certain system is the local one."  Other people may have better suggestions.  But IMHO, this is an important issue to address before publishing, as the term "separator" seems very confusing to me.

2.  As mentioned before, you might want to consider reimplementing some of the code using `pcase` extensions.  The colon-prefixed lisp DSL you've created isn't bad at all, but it is non-standard, and it will require users to learn a new language.  I have the same problem with `org-super-agenda`--it's like a lispy DSL in which each function is a colon-prefixed symbol, and it can be confusing, even to myself.  Having had a bit more experience since I made `org-super-agenda`, I would probably avoid the colon-prefixed symbols and go for something more like plain lisp.  I've tried to follow that principle better in the `org-ql` package, but I still have a lot to learn.

However, your implementation looks simple and understandable, so you might prefer to leave it as-is.

In any case, the decision whether to merge this package into MELPA isn't mine.  Other MELPA folks, like Jonas (if he has time to look at this) and Steve, who have much more experience than I, can offer better feedback on this issue.

Overall, I have to say, good job.  With a bit more polishing, I think it could be a valuable contribution for users who use multiple systems.

ROCKTAKEY:
> I don't have control over that. I'm not a MELPA staff member, just a contributor who tries to help out by doing some initial reviews of new packages.

OK, I misunderstood your position. Sorry.

> The naming of the conditions as "separators." I'm guessing that English isn't your first language, so I don't blame you for the confusion, and naming things is known to be hard. ;) As a native English speaker, the word "separator" doesn't seem appropriate or descriptive to me. I would recommend something else, perhaps "system definition", as in "the code that defines whether a certain system is the local one." Other people may have better suggestions. But IMHO, this is an important issue to address before publishing, as the term "separator" seems very confusing to me.

Yes, I've thought so, too, but I could not come up with the name of condition. I have firstly used the word "system", which is confusing, too. I will use "system-definition", provisionally. For there are still little or no users. :) Thanks.

> As mentioned before, you might want to consider reimplementing some of the code using pcase extensions. The colon-prefixed lisp DSL you've created isn't bad at all, but it is non-standard, and it will require users to learn a new language. I have the same problem with org-super-agenda--it's like a lispy DSL in which each function is a colon-prefixed symbol, and it can be confusing, even to myself. Having had a bit more experience since I made org-super-agenda, I would probably avoid the colon-prefixed symbols and go for something more like plain lisp. I've tried to follow that principle better in the org-ql package, but I still have a lot to learn.

I agree that. I wanted to use keyword list, such as `use-package`. However, it cannot be used recursively. So unavoidably I adopted curernt style. I'll leave it as-is for now. Thank you for telling.

>In any case, the decision whether to merge this package into MELPA isn't mine. Other MELPA folks, like Jonas (if he has time to look at this) and Steve, who have much more experience than I, can offer better feedback on this issue.

> Overall, I have to say, good job. With a bit more polishing, I think it could be a valuable contribution for users who use multiple systems.

Thank you so much! 

I learned a lot from you. I cannot thank you enough.

alphapapa:
> P.S. I realize the word "system-definition" is not replaceable to "separator", because the word "separator" in system-separate includes one-time conditions. I will use "system-condition" or "system-separator", instead.

The term "separator" is, AFAICT, semantically incorrect, and seems very confusing, so I strongly recommend against it.  I think users will be very confused by it, and I think potential users will not even try the package because of it.  Since they don't already understand what the package does, they will be unlikely to invest enough effort to try to understand what a "separator" is in the context of the package.  It's important to use terms that users are already familiar with.

Probably the best term to use, in lieu of a better one, would be the idiomatic, lispy one: predicate.  They are, after all, predicates, which return nil or non-nil depending on whether certain conditions are met.  So I would suggest something like `system-predicate`, which should be easily understood by Emacs and lisp users.

> I learned a lot from you. I cannot thank you enough.

I'm glad.  You're welcome.

Here are some other resources you might find useful:

* https://github.com/alphapapa/emacs-package-dev-handbook
* ...I was going to add another one or two, but they've slipped my mind...

ROCKTAKEY:
> The term "separator" is, AFAICT, semantically incorrect, and seems very confusing, so I strongly recommend against it. I think users will be very confused by it, and I think potential users will not even try the package because of it. Since they don't already understand what the package does, they will be unlikely to invest enough effort to try to understand what a "separator" is in the context of the package. It's important to use terms that users are already familiar with.

OK, I'll use `system-predicate`. Thanks!

> Here are some other resources you might find useful:
> 
>     https://github.com/alphapapa/emacs-package-dev-handbook
>     ...I was going to add another one or two, but they've slipped my mind...

Thanks! I'd grateful if you tell that when you remember. :)

purcell:
Thanks! Sorry for the delay in getting to this.

I feel like the package introduces a lot of specific arcane syntax for simple things. For example, your sample code in the README is trivially reproduced with Emacs built-ins like this:

```el
(defconst this-system
  (pcase (cons system-type system-name)
    (`(,_ . "my-windows") 'my-win)
    (`(gnu/linux . ,_) 'my-linux)
    (_ 'default)
    ))

(setq foo (pcase this-system
            ('my-win 1)
            ('default 3)
            ('my-linux 1)))
```

And this code doesn't need extensions if you instead want to call `setq-default` instead. In short, it's more idiomatic and composable. As a result, I really wouldn't want to encourage anyone to use the package: if it's helpful to you, then great, but I prefer not to include it in MELPA, sorry. This may seem harsh, but I feel users would be better served by writing Elisp or - perhaps - using `use-package`, which I believe also provides some mechanisms for system-specific configuration.

PR #6052, opened 2019-03-04
Add quilt package into melpa

jstranik:
I was surprised that quilt integration is missing in melpa. Then I
found only abandoned version of quilt code
https://www.emacswiki.org/emacs/Quilt. I did small enhancements and
published the version under a new location.

The package is an integration of quilt into emacs.

The quilt package enables convenient management of patches from 
emacs.

https://github.com/jstranik/emacs-quilt.

I cloned an orphaned package and updated it to modern version of emacs. 
The package is also packaged as package.el package. 

**None Needed**
### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/quilt
+++ b/recipes/quilt
@@ -1,0 +1,1 @@
+(quilt :fetcher github :repo "jstranik/emacs-quilt")


purcell:
Thanks, and sorry for the delay in getting to this!

Mostly looks great, and I'll merge it as-is, but the `quilt-hook` function is unidiomatic and bogus as far as I can see: what you want instead is a global minor mode, probably simply using `define-globalized-minor-mode` on top of `quilt-mode` and the `quilt-hook` function. (`quilt-hook` would then become an internal and non-interactive detail.)

PR #6055, opened 2019-03-05
Add pepita, package to run Splunk search from Emacs.

sebasmonia:

Run a Splunk search from Emacs.

Splunk is a tool that ingests plain data (for example, logs) and makes it searchable. It exposes a REST API, which made this package possible. See https://www.splunk.com for more information.

"Pepita": "nugget" in Spanish.

https://github.com/sebasmonia/pepita

Maintainer.

None.

### Checklist

Please confirm with `x`:


tarsius:
The recipe file has to be named `pepita`, not `pepita.el`.

sebasmonia:
Thanks @tarsius , it seems I'm destined to always get _something_ wrong when I create a new recipe :)

purcell:
> Thanks @tarsius , it seems I'm destined to always get _something_ wrong when I create a new recipe :)

The actions associated with the last checkbox in the checklist above would have caught this... :-)

In any case, this looks really useful. Some further notes from me, then:

- As `package-lint` would have told you when you ran it, mentions of `in Emacs` / `from Emacs` etc. are invariably redundant in elisp packages, and should be removed.
- There's a lot of code which introduces initially-`nil` variables with `let` and then sets them to a meaningful value. In almost all cases these should just be initialised directly to the non-`nil` value. `setq` should be a rarity in the code.
- Relatedly, instead of keeping several local variables like `pepita--buffer-query`, consider storing the whole query in a single variable. And such variables should be defined at the top level with a `defvar-local` instead of implicitly with `make-local-variable`.
- You might consider using `tabulated-list-mode` to display tabular data inside Emacs.
- `pepita--parse-querystring` appears to be doing the exact opposite of parsing. ;-)

Hope that helps!

purcell:
Also, it would be better to use `nil` as an "empty" value instead of [empty strings](https://github.com/sebasmonia/pepita/blob/effb3bf1326f32b968bf95e62fa5ecff5420d1b0/pepita.el#L275-L280).

sebasmonia:
Amazing feedback, thank you!

> As package-lint would have told you when you ran it, mentions of in Emacs / from Emacs etc. are invariably redundant in elisp packages, and should be removed.

I though a lot about how to remove Emacs from the title (copied from the GitHub repo description). "Run Splunk, get results to buffer" but nothing sounded quite right :) will give it another try!

> You might consider using tabulated-list-mode to display tabular data inside Emacs.

Both my previous packages rely heavily in tabulated-list-mode, however this one is different for a few reasons:

- I use the CSV feature of Splunk out of convenience: doesn't require pagination, it's immediately parseable (JSON returns a few objects per iteration so I need to concat them, XML is really, really, really verbose, "raw" format is really hard to parse, etc)

- The number and quantity of columns is variable, so is the amount of data. Sometimes a field has an entire exception stack trace with a few hundred characters

- The CSV format can be easily saved to a file and opened with a spreadsheet app. I thought making the HTML export the default was somewhat forced, ditto with JSON. Dumping the Splunk output sometimes is enough to know "this error message hasn't appeared in the last day"

The reason I explain all these, is just in case you still think tabulated-list if the way to go, or if there are other modes that I haven't considered/know about, I can try to adapt the package to use it.
I did look into exporting to org-tables or a few table-modes I found on MELPA, but nothing seemed to quite work.
Also, didn't want to risk running into the "lines are too long and make Emacs slow" issue. By having the results derived from fundamental-mode and having no font-lock, it keeps things lean.

Will work in the code in the coming days, thanks for the feedback and corrections!

sebasmonia:
`defvar-local` even takes care of adding the "Automatically becomes buffer-local when set." to the variable's help 😮

sebasmonia:
The funny past of "parse-querystring" is that it checked that the string passed was well formed. And since it worked not so well I moved to "let's build it instead.

Also not using `let` as if it were just a variable declaration changed running a new search for the better. `concat` is not bothered by nil, neither is `read-string`.

Learned a lot from these changes.

purcell:
Merged now, thanks for your patience!

PR #6056, opened 2019-03-06
Add cyberpunk 2019 theme

the-frey:

A very bright emacs theme.

https://github.com/the-frey/cyberpunk-2019

I'm the maintainer.

N/A

### Checklist

Please confirm with `x`:


purcell:
Looks good, and I can merge this now, but I'd note that all of your backticks should be simple apostrophes, because no backquote interpolation is used here. If you're not aware of the difference, it's worth reading up on. Welcome to MELPA! :tada:

the-frey:
Thanks! I will have a read and update when I get the chance 👍

PR #6059, opened 2019-03-06
add maple-preview recipe

honmaple:

Preview markdown, org-mode, html file automatically.

https://github.com/honmaple/emacs-maple-preview

maintainer

None needed

### Checklist

Please confirm with `x`:


purcell:
Are you aware that the recipe you have supplied does not include any elisp files? There are also issues in the code which `package-lint` would have reported if you had actually run it. I don't think you can have actually followed the instructions in the checklist above, which makes me sad and frustrated.

PR #6060, opened 2019-03-06
Add kakoune.el recipe file

jmorag:

This package combines ryo-modal-mode, multiple-cursors, and expand-region in order to emulate the [kakoune](https://kakoune.org/) editor.

https://github.com/jmorag/kakoune.el

Maintainer

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/kakoune
+++ b/recipes/kakoune
@@ -1,0 +1,1 @@
+(kakoune :repo "jmorag/kakoune.el" :fetcher github)


purcell:
Thanks. Quick feedback:

- There are many identifiers which don't start with the `kakoune-` prefix, e.g. [everything here](https://github.com/jmorag/kakoune.el/blob/c1dd917ebb429408ae7ec8b7bd2da293fe9f5bc3/kakoune-utils.el)
- Those outboard files should still `require` any necessary libraries so that they compile cleanly.

jmorag:
Thanks for the feedback. Just pushed some changes. Please let me know if there's anything else missing.

purcell:
> Thanks for the feedback. Just pushed some changes. Please let me know if there's anything else missing.

Great! Yep, the symbol prefix should still be `kakoune-`, to match the package, not `kak/`. And my other comment above (about `require` in the outboard files) still applies.

jmorag:
Ok. Changed the prefix to `kakoune-`. I'm not totally sure which files you wanted me to `require`, so I just added calls to `require` all the dependencies in the auxiliary files.

purcell:
> I'm not totally sure which files you wanted me to `require`, so I just added calls to `require` all the dependencies in the auxiliary files.

Yep, that was what was needed: the idea is that each file should directly require all the libraries it refers to.

It's looking pretty nice now, but I still spotted a couple of little issues:

- You need to declare a package dependency on either `(cl-lib "0.6")` or `(emacs "24.3")` because you're using `cl-lib`. `package-lint` should have told you about this. I'd probably suggest the latter because:
- You probably want [this var](https://github.com/jmorag/kakoune.el/blob/09b0347b767d547859b2978ccd370dedc1c84073/kakoune-utils.el#L36) to be maintained on a per-buffer basis, but `defvar` introduces a global variable shared between all buffers. Instead use `defvar-local` (added in Emacs 24.3, when `cl-lib` was also added to core Emacs).
- `progn` like [this](https://github.com/jmorag/kakoune.el/blob/09b0347b767d547859b2978ccd370dedc1c84073/kakoune-utils.el#L48) are redundant in a `let`.
- As a matter of style, you should leave empty lines between top-level declarations like [this](https://github.com/jmorag/kakoune.el/blob/09b0347b767d547859b2978ccd370dedc1c84073/kakoune-utils.el#L16) and avoid [leaving closing parens on their own line](https://github.com/jmorag/kakoune.el/blob/09b0347b767d547859b2978ccd370dedc1c84073/kakoune-utils.el#L113).

purcell:
Ping! :-)

jmorag:
Sorry I'm late getting to this. Just pushed the changes.

riscy:
Hey, having a pass through this.  In `kakoune.el`, package-lint is giving me:
```
2 issues found:
1:1: warning: Including "Emacs" in the package description is usually redundant.
1:1: error: Package should have a Homepage or URL header.
```
Checkdoc on `kakoune-exchange.el` and `kakoune-utils.el` complains about some missing comments (these appear minor but easy fixes).
- I may be mistaken, but it seems your setup example in the README (`my/kakoune-setup`) references a function `ryo-enter` that does not exist in `ryo-modal`
- In `kakoune.el` use `forward-line` instead of `goto-line` (the latter is for interactive use); also use `(goto-char (point-min))` instead of `beginning-of-buffer` for the same reason -- as a small bonus these non-interactive versions are usually more performant and written in C

tarsius:
Ping!

jmorag:
Apologies again for the lateness. @riscy, you're right, `ryo-enter` only exists in my config, but I've now included it into the README. I also rewrote `kakoune-gg` to use the more performant functions and fixed up some of the checkdoc and package-lint issues. Anything else?

purcell:
Looks good to me now, thanks - merged!

jmorag:
Great! Thank you so much for your help with this :)

PR #6064, opened 2019-03-10
Add pdf-view-restore recipe

007kevin:

Support for opening last known pdf position in pdfview mode.

https://github.com/007kevin/pdf-view-restore

maintainer

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/pdf-view-restore
+++ b/recipes/pdf-view-restore
@@ -1,0 +1,1 @@
+(pdf-view-restore :repo "007kevin/pdf-view-restore" :fetcher github)


dzop:
`pdf-tools` is already able to save the state of a pdf using the built-in `bookmark` package with much more information other than just the current page such as your current position in a page. See `pdf-view-bookmark-make-record`.

The only thing missing is support for automatically setting the bookmark when exiting the pdf and restoring it when opening the pdf. Perhaps this package can instead use the bookmark feature of `pdf-tools` and save/restore bookmarks in `kill-buffer-hook`/`find-file-hook`.

One question with this approach is how to prevent clashing with pdf bookmarks created manually by the user? I guess you can use `pdf-view-restore-filename` as a separate bookmark file in which you can save and load the bookmarks created by this package, leaving the default bookmark file alone.

Also the `pdf-view-restore-filename` is not an absolute filename. Is this intended? It would be better to have this file in `user-emacs-directory` instead of having multiple restore files in every directory that contains a pdf.

007kevin:
Thank you for taking the time to review the code, it is much appreciated!

@purcell I've made the changes listed in your notes other than creating my own group since I opted to use the existing pdf-view group instead. As for the alist-get changes, the code is now simpler, but the emacs dependency is changed 26 since I require string equality (since identifiers are based on the file name which could contain non-symbol keys).

@dzop Thank you for pointing out the bookmark feature. Although I could use that instead, I decided to keep the current implementation since my workflow involves mounting a remote directory of pdf documents to be viewed on different computers (i.e work, school, home). Since I may not always close my emacs session, I can't rely on the kill-buffer-hooks to keep my current position updated. Also, simply keeping the save information relative to the pdf made more sense in this workflow.

I've tested the possibility of clashing pdf's opened via bookmarks and the bookmark initiated page changes occur after all the mode hooks have run so it shouldn't be a problem.

purcell:
Thanks. Couple of other things I noticed:

- If `pdf-view-after-change-page-hook` is not automatically buffer-local, then you must pass the "local" flag to `add-hook` and `remove-hook` [here](https://github.com/007kevin/pdf-view-restore/blob/2e907828546d813c69a8a9211263cca89b4efe0b/pdf-view-restore.el#L52), so that your local minor mode doesn't have a global effect.
- Prefer `derived-mode-p` to `(eq major-mode ...)` [here](https://github.com/007kevin/pdf-view-restore/blob/2e907828546d813c69a8a9211263cca89b4efe0b/pdf-view-restore.el#L63)

In any case, I'm merging this now (thanks for your patience!) but please take a look at the above points in your own time. Welcome to MELPA! :tada:

PR #6065, opened 2019-03-10
Add recipe toggle-quotes-plus

jcs090218:

Simple quote toggler that cycle through " ' and `

https://github.com/jcs090218/toggle-quotes-plus

The maintainer.

There is another package name [toggle-quotes](https://github.com/toctan/toggle-quotes.el) by @toctan. I recreate the package by my own method and are slightly different from his package, and I'm not quite sure if he is still maintaining the package since he hasn't update since `20140710.926`. I know is bad to have similar packages on `melpa` at the same time, so feel free to close this down if you don't think this is good enough to be released. :D

#### Differences
* Added `toggle-quotes-plus-chars` for customization.
* Fixed not working in `web-mode`.
* Different solution handling escape character.
* Works outside of string.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/toggle-quotes-plus
+++ b/recipes/toggle-quotes-plus
@@ -1,0 +1,1 @@
+(toggle-quotes-plus :repo "jcs090218/toggle-quotes-plus" :fetcher github)


mattiasb:
How does this compare to [cycle-quotes](http://elpa.gnu.org/packages/cycle-quotes.html) from ELPA?

mattiasb:
Answering this myself it looks like `cycle-quotes` finds the relevant quotations from the major mode in question while `toggle-quotes-plus` punts this to a defcustom?

With that said: it's hard searching for all alternative packages out there. :)

jcs090218:
@mattiasb Hello! :D Sorry if I miss any alternative packages here (Yeah, is hard!). I do have `toggle-quotes-plus-chars` variable that let user customize it themselves to each corresponding mode. I don't know if I should manage this package towards the same as `cycle-quotes`, my habits are just like to have the tool working in any modes instead of hard coding it in each mode. :(

Welcome to anyone who think this package is too similar to other alternative package, please leave a comment. :D

mattiasb:
@jcs090218 Yeah sure, but I think (from reading the `cycle-quotes` code) that it's already possible to fetch the list of relevant quotations from the major mode in question. 

This is how `cycle-quotes` seems to do this:

```elisp
(defun cycle-quotes--set-quote-chars ()
  "Set the quote chars for the current syntax table."
  (let ((syntax-table (syntax-table)))
    (while syntax-table
      (map-char-table
       (lambda (char-code-or-range syntax)
         (when (equal (syntax-class syntax) 7)
           (if (consp char-code-or-range)
               (let ((from (car char-code-or-range))
                     (to (cdr char-code-or-range)))
                 (dolist (char-code (number-sequence from to))
                   (add-to-list 'cycle-quotes--quote-chars char-code)))
             (add-to-list
              'cycle-quotes--quote-chars char-code-or-range))))
       syntax-table)
      (setq syntax-table (char-table-parent syntax-table)))
    (setq-local cycle-quotes--quote-chars-mode major-mode)))
```

jcs090218:
@mattfidler Hmmm, I wonder does `cycle-quotes` works with the thing other than string? I tried `cycle-quotes` and it gives me an error, `Not inside a string`. Like as I mentioned in my repo's `README.md` file, I kind of want to use this in `.md`, `.html` or any related file type.

mattiasb:
@jcs090218 Hm, how do you mean? How would you cycle between quotes without being inside a quotation?

jcs090218:
@mattfidler No worry, I'm not an expert in `elisp` or `Emacs`. I learned while I'm using it. Frankly speaking, my code isn't good and I code it this way because of my limited knowledge on `elisp`. @purcell always gives good advices on the code and keep up the quality of `melpa` packages, very thanks to his effort. :D 

This is how this package works, the package does not check weather the `point` is currently on a `string` or not. It simply search `target` character `before` and `after` from the current `point`. Hence, it will work in any file type because it does not care about the `font-face` in the buffer. I presume `cycle-quotes` uses the same method as `toggle-quotes` to achieve this behavior; that is why I got this error `Not inside a string` when using `cycle-quotes`. If you use `cycle-quotes` or `toggle-quotes` like file in JavaScript; it is totally legal because JavaScript recognized `'` and `"` as a string. When I first try out `cycle-quotes`, I was in a C++ file. C++ recognized `"` but not `'`, therefore it gives me that error. 

Here is why I create this package, 
* Use it everywhere, especially in `.html` and `.md` file. The line of code wouldn't take any effect using `cycle-quotes` or `toggle-quotes` because it doesn't recognize as a `string`.
```html
<meta charset="UTF-8"/>
```
* Maintaining this package, because the author of `toggle-quotes` is no longer maintaining the package and I really want to use this package in `web-mode`.
* Easy to use, I think the current structure of the code can handle most of the request especially because this is such a small package.

mattiasb:
Hm. I don't quite follow what you write but I can see a problem when cycling between quotes with `cycle-quotes`. That is since markdown doesn't have any alternate quotation styles I get this error: `The current mode has no alternative quote syntax`.
markdown-cycle-quotes

I guess having a list of quotes available over all major modes would work around this. :)

jcs090218:
@mattfidler Hmm, cool! I think is a bug to `c++-mode` because it suppose to pop the same error `The current mode has no alternative quote syntax` since c++ doesn't use `'`??? Am I right?

> I guess having a list of quotes available over all major modes would work around this. :)
Not quite sure what you mean by this??? :0

alphapapa:
1.  You should probably consider using major mode syntax tables to determine quote characters.  e.g. `syntax-ppss`.
2.  Rather than write an entire package to do this, it could probably be implemented as a minor extension to an existing package that manipulates surrounding characters, of which there are several.
3.  I see redundant functions like `toggle-quotes-plus-is-beginning-of-buffer-p` (e.g. use `bobp`), `toggle-quotes-plus-is-end-of-buffer-p`, `toggle-quotes-plus-current-char-equal-p `, `toggle-quotes-plus-is-in-list-string`, ` toggle-quotes-plus-current-char`, etc, which aren't necessary, since there are existing built-in functions that do these things, as well as standard libraries like `dash`, etc.  And you don't need to convert characters to strings just to compare them; see the Elisp manual on strings and characters.

So in general:

* This package probably isn't necessary, as other, more general packages should easily fulfill the same needs.
* As is, it needs work to be more idiomatic and standard Emacs lisp.  I recommend consulting the manuals, the style guides in MELPA's `CONTRIBUTING.org` file, and the source code of similar packages that are time-tested and widely used.

purcell:
This is interesting, but I agree it needs some work, including to work idiomatically.

- Storing `point` in a variable and then manually restoring it after doing a bunch of jumping around is a solved problem: see `save-excursion` etc.
- What unit does this work on if there are no quotes around point? It's nice that you want to handle the case where `(point)` is not inside a string, but in that case you should default to the current symbol, or to the region if it's active. I guess it feels like the behaviour is not currently well-defined.
- Getting the quote list from the major mode would be a great idea, and I'd suggest only using the quote chars list as an override. It *might* also make sense to support any kind of pair, not just identical ones: so `()`, `[]` etc. `electric-pair-mode` has code for handling that sort of thing.
- It occurs to me that `smartparens` might already have commands that do similar things to these.
- Escaping rules are different in various languages, so I'm not sure that the hard-coded `"\\"` [handling here](https://github.com/jcs090218/toggle-quotes-plus/blob/master/toggle-quotes-plus.el#L85) is the way to go.

jcs090218:
Thanks for all people who anticipate to this package, I think I will just close this for some reasons!

PR #6066, opened 2019-03-12
Return helm-words back

sergeyklay:

Customizable Helm extension for easy looking up words in various online/offline dictionaries and thesauri.

Refs:
- https://github.com/pronobis/helm-words/pull/2
- https://github.com/melpa/melpa/commit/cc53e7c

https://github.com/pronobis/helm-words

I'm an enthusiastic user

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/helm-words
+++ b/recipes/helm-words
@@ -1,0 +1,2 @@
+(helm-words :repo "pronobis/helm-words"
+            :fetcher github)


alphapapa:
From what I can tell, the package author ignored the issue for over 3 years, and then finally merged a patch without comment.

Is the author actually *maintaining* the package?  If the author is not really actively maintaining it, then it probably should not be in MELPA, especially since it depends on a package like Helm and various online services which are likely to break at any random moment.  Not every Emacs package needs to be in MELPA.

sergeyklay:
Ok. Not a problem at all. Thank you for the feedback

alphapapa:
@sergeyklay Please note that I'm not a MELPA staff member, so my opinion is not a decision, just my two cents.  :)

sergeyklay:
@alphapapa In general, your arguments seemed convincing to me. So MELPA membership was not decisive for me.

PR #6067, opened 2019-03-12
Add recipe reload-emacs.

jcs090218:

Reload Emacs without restart Emacs.

https://github.com/jcs090218/reload-emacs

The maintainer.

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/reload-emacs
+++ b/recipes/reload-emacs
@@ -1,0 +1,1 @@
+(reload-emacs :repo "jcs090218/reload-emacs" :fetcher github)


jcs090218:
This package serve similar purpose to (restart-emacs)[https://github.com/iqbalansari/restart-emacs]. There are several key differences,
* Preserve current workspace after reload Emacs. Meaning the opened buffers will still remain opened.
* Only overwrites the current configuration, not start a new Emacs instance.
* Variables using `defvar`, etc will not change after reload. You will still have to restart Emacs. The rest of configuration changes should be fine.

alphapapa:
Reading this description, I'm afraid I don't understand it at all:

> This package is only useful when you want to test your configurations without restarting Emacs. This package simply reloads all your config files once again, and finally reloads your init.el or .emacs file. If your configuration files originally use load-file and not require, then this package wouldn't really take advantage from your Emacs configuration files.

If a user like me doesn't understand it, most Emacs users probably won't, so you should probably rewrite it to be more descriptive.

purcell:
>     * People can use `restart-emacs` and `recentf` to achieve similar behavior.  (Point 1.) But is laborious if you opened many files.

That's what `desktop-save-mode` is for. I personally wouldn't suggest that anyone use a package like this, because - as you note - it's impossible for them to work completely and reliably, so it's almost always better to either: a) reload small chunks of code selectively, or b) just restart Emacs. It's also worth noting that the *order* in which Emacs features are loaded is extremely important, because some will define functions, variables and (particularly) macros which will influence the *compilation* process of other features. So just reloading all the `.el` files in a random order does not reproduce the initial load behaviour.

A better approach would be unload all possible features and then re-evaluate `init.el`, but note that *even then* there will be differences relative to how Emacs itself runs `init.el` as part of the start-up sequence.

Overall I would say this is an inadvisable approach, then, and I'd prefer not to include the package in MELPA because I think it will be most confusing to the users who are most likely to think they might need it: the relative novices. Sorry, but thanks for taking the time to submit this.

PR #6071, opened 2019-03-14
New package build

milkypostman:
want to pull a new package build but want @tarsius to give a quick look to see if this is going to be fun or not.

### DIFF
--- a/package-build/package-build.el
+++ b/package-build/package-build.el
@@ -246,7 +246,7 @@ is used instead."
      ((and (file-exists-p (expand-file-name ".git" dir))
            (string-equal (package-build--used-url rcp) url))
       (package-build--message "Updating %s" dir)
-      (package-build--run-process dir nil "git" "fetch" "--all" "--tags"))
+      (package-build--run-process dir nil "git" "fetch" "-f" "--all" "--tags"))
      (t
       (when (file-exists-p dir)
         (delete-directory dir t))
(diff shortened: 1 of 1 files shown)


milkypostman:
@tarsius this seems to have the homepage business in it. Didn't this break melpa last time?

tarsius:
I think last time I tried to *remove* that code in favor of `lm-homepage`. This time around I just moved it into a separate function so that it is out of sight. ` package-build--lm-homepage` is a copy of `lm-homepage`. It differs from the removed inline code in that is uses the `lm-with-file` macro. We don't actually pass a file argument, so the only difference is that the buffer is temporarily widened and the search begins at the beginning of the buffer. Looks safe to me.

tarsius:
But it looks like you will have to update mercurial:

```
Cloning https://bitbucket.org/zck/2048.el to /home/travis/build/melpa/melpa/working/2048-game/
Command '(env LC_ALL=C /usr/bin/timeout -k 60 600 hg clone https://bitbucket.org/zck/2048.el /home/travis/build/melpa/melpa/working/2048-game/)' exited with non-zero status 255: warning: connecting to bitbucket.org using legacy security technology (TLS 1.0); see https://mercurial-scm.org/wiki/SecureConnections for more info
abort: error: _ssl.c:510: error:1408F10B:SSL routines:SSL3_GET_RECORD:wrong version number
```

milkypostman:
That’s a Travis CI thing. Not sure how I fix what it’s instaling?

PR #6072, opened 2019-03-14
Add lsp-origami

vibhavp:

origami.el support for lsp-mode

https://github.com/emacs-lsp/lsp-origami

Maintainer/Author

None Needed

### Checklist

### DIFF
--- a/recipes/lsp-origami
+++ b/recipes/lsp-origami
@@ -1,0 +1,1 @@
+(lsp-origami :repo "emacs-lsp/lsp-origami" :fetcher github)


purcell:
This looks neat! I didn't know this was something that lsp supported. Quick notes, then:

- Instead of an `enable` function, provide a proper toggling minor mode. (You can `kill-local-variable` to unset the local modification to `origami-parser-alist`).
- It also looks like you can set `origami-fold-style` locally instead of overriding all of `origami-parser-alist`: in this case you'd just add a global entry to `origami-parser-alist` rather than modify that var locally.
- Your code uses four different and very similar libraries -- dash, dash-functional, cl-lib and seq. I strongly encourage you to cut that down a bit by learning to use some of them more fully. For example, `cl-destructuring-bind` (and even the `pcase-let` built-in) do the same as `-let`, and there are built-ins like `apply-partially` for more functional styles.

Hope that helps!

vibhavp:
Thanks for the suggestions! Almost all of the code in `lsp-origami` has been migrated to `lsp-mode`, though, and the current package is a thin wrapper over lsp functionality. I have rewritten the code to use `origami-fold-style` instead of locally setting `origami-parser-alist`, though.

purcell:
Thanks, looks good - merging now.

P.S. You might prefer to use `kill-local-variable` [here](https://github.com/emacs-lsp/lsp-origami/blob/c7653602a2f2396b1a42d6053fd2be55fce8e0a2/lsp-origami.el#L69) rather than setting the local version of the var to `nil` -- what you probably wanted to do was restore the global default, which this will accomplish.

PR #6074, opened 2019-03-16
Add BNF Mode

sergeyklay:

A GNU Emacs major mode for editing BNF grammars. Currently provides basic syntax and font-locking for BNF files.

https://github.com/sergeyklay/bnf-mode

Maintainer/Author

None Needed

### Checklist

### DIFF
--- a/recipes/bnf-mode
+++ b/recipes/bnf-mode
@@ -1,0 +1,2 @@
+(bnf-mode :repo "sergeyklay/bnf-mode"
+          :fetcher github)


purcell:
Thanks - merging now!

BTW, I strongly recommend removing the use of `pkg-info` and particularly the `bnf-mode-version` function, which is almost the biggest piece of code in this little package. Users can easily call `describe-package` (built-in) or `pkg-info-package-version` interactively if they want to get this information.

sergeyklay:
@purcell Thanks! Changed

PR #6076, opened 2019-03-16
Add catkin package for ROS compilation

gollth:
This package integrates the [catkin](http://wiki.ros.org/catkin) build tool for [ROS](http://ros.org) packages into Emacs. With it you can:

- Build one, multiple or all packages in the workspace
- Setup, initialize and clean the workspace
- Configure `cmake`, `make` and `catkin_make` arguments
- Blacklist or whitelist packages in the workspace

https://github.com/gollth/catkin

Maintainer/Author/Main Requester =P

**None needed**

### Checklist

Please confirm with `x`:

### Note
flycheck, package-lint and also the make file from Melpa give me warnings that my requires of "helm" and "xterm-color" are missing version numbers.

``` Package-Requires: ((emacs "24.3") helm xterm-color (cl-lib "0.5"))```

I understand this warning and tried to provide a "snapshot commit" to both (e.g. "201903015.35" or so) but this was again warned about I shall not do that. When I run `M-x  list-packages` I can only see these version however.

If this warning is a concern for your, please tell me who to find out which version of the packages I'm currently using. Thanks!

### DIFF
--- a/recipes/catkin
+++ b/recipes/catkin
@@ -1,0 +1,4 @@
+(catkin
+ :fetcher github
+ :repo "gollth/catkin"
+ :branch "master")
--- a/recipes/catkin
+++ b/recipes/catkin
@@ -1,4 +1,3 @@
 (catkin
  :fetcher github
- :repo "gollth/catkin"
- :branch "master")
+ :repo "gollth/helm-catkin")

### CODE REVIEW

**Comment by gollth** [comment]
Hi purcell, 

thanks for the constructive feedback so far. 

* The reason why I chose an environment variable is simple. This is my first time doing anything in elisp so I didn't know better =) That being said, I'm happy to work in all you comments as suggested. 

* Regarding the `helm` dependency: Since this is a private side project, I will not be able to maintain two separate packages for now. Therefore I will just rename the repo to `helm-catkin`

I don't know if you have ever worked with `catkin`, but their philosophy is to source a `setup.bash` file in the workspace which sets a lot of environment variables for you (such as `LD_LIBRARY_PATH`, `CMAKE_PREFIX_PATH`, `PYTHONPATH` etc.). This gets a little bit unusable within emacs, since these values are not carried over from sourcing the setup file in a `call-process-shell-command` or so. 

I want to make this package well integrated with the already existing [helm-ros](https://github.com/davidlandry93/helm-ros) on Melpa (which relies on theses envs to be set correctly per workspace). My current workaround is to use the [exec-path-from-shell](https://github.com/purcell/exec-path-from-shell) package (from you, which I just notice =P) and then bind the following to a keybinding:

```elisp
(exec-path-from-shell-copy-envs '("ROS_IP" "PYTHONPATH" "CMAKE_PREFIX_PATH" "ROS_MASTER_URI" "LD_LIBRARY_PATH"))
```

Do you have a better suggestion on the top of your head, how to deal with per workspace changing environment variables?

Thanks again for the feedback. I'll let you know when I worked in all the comments.

gollth:
- [ ] Rename Package
- [ ] Join isolated `)`
- [ ] `if` -> `when`/`unless`
- [ ] environment variable refactoring
- [ ] `defalias` and wrong capital names fixed
- [ ] loop -> cl-loop
- [ ] Helm Keybindings in documentation adjusted

gollth:
Hi,

I refactored the code now according to your requests, please feel free to have a look again. For the F1/F2 helm keybindings: When I do `M-x describe-keymap RET helm-map RET` I get the following:

```
<down>          helm-next-line
<escape>        keyboard-escape-quit
<f1>            ??
<f10>           ??
<f11>           ??
<f12>           ??
<f13>           ??
<f2>            ??
<f3>            ??
```

I refactored all known helm-keybindings in the help strings, but left the F1/F2... in there, because I think they are source specific or so? I hope that is fine for now

The tip with the global emacs variable was very nice and made the code a lot more lightweight. Basically if the var is not set, I will choose the `default-directory` as the workspace, which is fine in my experience.

Let me known, if there is still something you'd like me to change =)

purcell:
Looking good! Couple of points I noticed:

- Don't introduce variables with a `setq` like [this](https://github.com/gollth/helm-catkin/blob/ef9bf51f57c6068e8d0141f03d897d213cbe731e/helm-catkin.el#L377) - use a `defvar` instead.
- [This code](https://github.com/gollth/helm-catkin/blob/ef9bf51f57c6068e8d0141f03d897d213cbe731e/helm-catkin.el#L56-L58) creates two global variables called `ws` and `root`. You want to use `let*` there instead. (Also, wrap the `unless` form onto two lines so it's easier to read.)
- You don't want to be using `ignore-errors` in [basic code you control](https://github.com/gollth/helm-catkin/blob/ef9bf51f57c6068e8d0141f03d897d213cbe731e/helm-catkin.el#L68). What you probably want is to use the `noerror` flag of `re-search-forward` instead. :-)
- You're only using `cl-lib` for `cl-loop`, so perhaps just use the built-in `dolist`?

gollth:
OK, done and pushed to master now =)

purcell:
Nice, merged now - thanks! :tada:

purcell:
The recipe had not been renamed to match the renamed package, but I've fixed that myself now.

PR #6077, opened 2019-03-17
Add recipe for emacs-pubmed

fvdbeek:

Emacs-pubmed is a GNU Emacs interface to the [PubMed database](https://www.ncbi.nlm.nih.gov/pubmed/). It is in an early state of development and far from complete, but already usable.

https://gitlab.com/fvdbeek/emacs-pubmed

I am the maintainer

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/pubmed
+++ b/recipes/pubmed
@@ -1,0 +1,3 @@
+(pubmed
+ :fetcher gitlab
+ :repo "fvdbeek/emacs-pubmed")


purcell:
Thanks for this! Quick notes:

- I seem to recall there's a package which defines the HTTP status codes: perhaps worth using that instead of inlining a list?
- You can omit the `-pkg.el` file, since the required metadata is already in the header of the main `.el` file, so you needn't maintain both.

In any case, this is ready to merge, so welcome to MELPA! :tada:

PR #6078, opened 2019-03-17
Add recipe for gnus-summary-repo

TxGVNN:

A tool will help to import and export files between IMAP and local by using GNUS. It will like a storage service.

https://github.com/TxGVNN/gnus-summary-repo

Maintainer

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/gnus-summary-repo
+++ b/recipes/gnus-summary-repo
@@ -1,0 +1,3 @@
+(gnus-summary-repo
+ :repo "TxGVNN/gnus-summary-repo"
+ :fetcher github)


alphapapa:
This is an interesting package, but I feel like it needs more documentation, considering how unusual it is.  I also recommend implementing some tests before encouraging users to trust it with their data.  That should probably include tests that users can run on their IMAP servers before starting to use the package to ensure their servers work properly with this functionality.

TxGVNN:
Hi @alphapapa, Sorry for reply late and thanks for your suggestion. The package needs to be improved, and I just began with elisp. So it will take a lot of time :).

The first I should have a good document and describe how it works, what needs to be improved. I will do it now.

TxGVNN:
Thank you for great review, 

>     * Since you're already `require`-ing `gnus-sum`, you don't need [this "declare-function" form](https://github.com/TxGVNN/gnus-summary-repo/blob/ef0e676dd7a402039fd517c8cf12f2117dfd8bb2/gnus-summary-repo.el#L24)
> 
I updated
>     * Consider re-using a built-in ignore list like `grep-find-ignored-files` or `grep-find-ignored-directories` instead of [defining yet another one](https://github.com/TxGVNN/gnus-summary-repo/blob/ef0e676dd7a402039fd517c8cf12f2117dfd8bb2/gnus-summary-repo.el#L40).
The purpose of `gnus-summary-repo-import-ignore` is to help people ignore the files they don't want to import  like the password, configuration file,... sensitive data. So I think should not use `grep-find-ignored-files` instead
> 
>     * Use `interactive` to handle [this code](https://github.com/TxGVNN/gnus-summary-repo/blob/ef0e676dd7a402039fd517c8cf12f2117dfd8bb2/gnus-summary-repo.el#L55-L60), e.g. you could just write `(interactive (list (or gnus-summary-repo-dir-local (read-directory-name "Select a directory to import: "))))`
I updated
> 
>     * Shouldn't [this test](https://github.com/TxGVNN/gnus-summary-repo/blob/ef0e676dd7a402039fd517c8cf12f2117dfd8bb2/gnus-summary-repo.el#L61) be `file-directory-p`?
I removed
> 
>     * Similar feedback for [this function](https://github.com/TxGVNN/gnus-summary-repo/blob/ef0e676dd7a402039fd517c8cf12f2117dfd8bb2/gnus-summary-repo.el#L71) and others -- use `interactive` where possible.
I don't know how to handle this [L129 ](https://github.com/TxGVNN/gnus-summary-repo/blob/cbd7b0d182fd37c32cccdbd684d1c8fce68eb538/gnus-summary-repo.el#L129) (read-string subject depends on filename ) and [L84](https://github.com/TxGVNN/gnus-summary-repo/blob/cbd7b0d182fd37c32cccdbd684d1c8fce68eb538/gnus-summary-repo.el#L84) (support number prefix to export many files)
> 
>     * Typos in [these var names](https://github.com/TxGVNN/gnus-summary-repo/blob/ef0e676dd7a402039fd517c8cf12f2117dfd8bb2/gnus-summary-repo.el#L225): modifation -> modification
I fixed

TxGVNN:
>     * Shouldn't [this test](https://github.com/TxGVNN/gnus-summary-repo/blob/ef0e676dd7a402039fd517c8cf12f2117dfd8bb2/gnus-summary-repo.el#L61) be `file-directory-p`?
I mistook and updated https://github.com/TxGVNN/gnus-summary-repo/commit/849acf1050154afc6e501410bbe52e6c5d019ec8. Because of
- with importing action, I want to check it is exist directory
- with export action, I want to check it is NOT a regular file.
What do you think?

purcell:
Thanks. I sent you a couple of fixes in https://github.com/TxGVNN/gnus-summary-repo/pull/2, but I'll go ahead and merge this now. :-)

TxGVNN:
The good news today, thanks

PR #6079, opened 2019-03-17
Include .nosearch in realgud recipe

terlar:
When using realgud together with subdirs.el it will include some
directories that should not be in the load-path, however they should be
part of the distribution.

Adding the .nosearch file will make realgud work together with site-lisp
which is what NixOS use for example.

See:
- https://github.com/realgud/realgud/issues/234
- https://github.com/NixOS/nixpkgs/issues/57622

https://github.com/realgud/realgud

An enthusiastic user

I suppose none needed.

### DIFF
--- a/recipes/realgud
+++ b/recipes/realgud
@@ -1,7 +1,7 @@
 (realgud
  :fetcher github
  :repo "realgud/realgud"
- :files ("realgud.el"
+ :files ("realgud.el" "realgud/.nosearch"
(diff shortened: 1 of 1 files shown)


purcell:
Thanks for this. Can you fix up the conflicts relative to our `master` version of this recipe, please?

terlar:
Sorry, I was in a rush and didn't realize my melpa fork was ancient. I have updated and rebased it now.

PR #6082, opened 2019-03-19
Add recipe isearch-project.

jcs090218:

Incremental search through the whole project.

https://github.com/jcs090218/isearch-project

The maintainer.

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/isearch-project
+++ b/recipes/isearch-project
@@ -1,0 +1,1 @@
+(isearch-project :repo "jcs090218/isearch-project" :fetcher github)


alphapapa:
This looks cool.  Good screencast in the readme.

Looking at the code, I see functions like:

* ` isearch-project-is-contain-list-string`
* ` isearch-project-remove-nth-element`
* `isearch-project-filter-directory-files-recursively`

Which seem to reimplement existing functionality.  I recommend looking at existing libraries like `cl-lib`, `dash`, and `f`.  There is also, e.g. `directory-files-recursively` which has a `REGEXP` argument.

purcell:
>     * ``
> 
> 
> Which seem to reimplement existing functionality. I recommend looking at existing libraries like `cl-lib`, `dash`, and `f`. There is also, e.g. `directory-files-recursively` which has a `REGEXP` argument.

Agree.

Other notes:
- Whenever I see a list like `isearch-project-ignore-paths`, I know that an opportunity to re-use built-in functionality has been missed. You should never need to repeat exclusions for `.git`, `.hg` etc in multiple packages! `grep-find-ignored-directories` is a standard variable for this purpose, for example. Or maybe use `project.el` or `projectile` to help you know what files the user considers "searchable" in the current context. Don't forget there are other considerations like ignoring binary files etc.
- Regarding things like `isearch-project-forward-symbol-at-point`, there's already `compilation-next-error`, and behaviour like this is built into grep-mode etc. Why not use built-in or packaged search functionality internally and simply hide the `*grep*` buffer? This would let you dramatically simplify the code, and would potentially allow it to work with `rg`, `ag` etc. too.
- There's a typo [here](https://github.com/jcs090218/isearch-project/blob/8ba55c00fb96e6f262261f1560254decbf8cbb7b/isearch-project.el#L56), but in any case you shouldn't use `\` to extend initial docstring lines across two lines.

purcell:
(I should add that I really like the idea of letting the user seamlessly switch from searching within a single file to jumping across an entire project. I'd like to see this as seamless and reliable as possible, hence the critical feedback above.)

jcs090218:
> * Whenever I see a list like `isearch-project-ignore-paths`, I know that an opportunity to re-use built-in functionality has been missed. You should never need to repeat exclusions for `.git`, `.hg` etc in multiple packages! `grep-find-ignored-directories` is a standard variable for this purpose, for example. Or maybe use `project.el` or `projectile` to help you know what files the user considers "searchable" in the current context. Don't forget there are other considerations like ignoring binary files etc.

Fixed! I used grep-find-ignored-directories instead of self defined variable `isearch-project-ignore-paths`.

> * Regarding things like `isearch-project-forward-symbol-at-point`, there's already `compilation-next-error`, and behaviour like this is built into grep-mode etc. Why not use built-in or packaged search functionality internally and simply hide the `*grep*` buffer? This would let you dramatically simplify the code, and would potentially allow it to work with `rg`, `ag` etc. too.

I just have it as an option because `isearch` have one and I used it a lot. The reason I don't use `grep` because I never used it, I would definitely like to give it a try. :D

> * There's a typo [here](https://github.com/jcs090218/isearch-project/blob/8ba55c00fb96e6f262261f1560254decbf8cbb7b/isearch-project.el#L56), but in any case you shouldn't use `\` to extend initial docstring lines across two lines.

Fixed!

purcell:
Merged now, thanks

PR #6083, opened 2019-03-19
Add recipe for slime-repl-ansi-color

augfab:

Turn on ANSI colors in [SLIME](https://melpa.org/#/slime) REPL output.

https://gitlab.com/augfab/slime-repl-ansi-color

I am a user of this code snippet, which I had some trouble finding online and integrating into emacs when I started using SLIME.
Making it available on MELPA will make it easier for everyone to find and use it.

I haven't found a maintained version of this package, only copies of the original file to make it available.
Hence I will assume the role of maintainer for this package.

### Checklist

### DIFF
--- a/recipes/slime-repl-ansi-color
+++ b/recipes/slime-repl-ansi-color
@@ -1,0 +1,1 @@
+(slime-repl-ansi-color :fetcher gitlab :repo "augfab/slime-repl-ansi-color")


purcell:
Thanks for this, and sorry for the delay in looking at it. Quick notes:

- Whenenver you have a pair of `-on` and `-off` functions, what you really want instead is a minor mode! So use `define-minor-mode` instead, and then in your advice function you can just check the minor mode's var to see if you should be applying ANSI colors: you can get rid of the current flag var.
- I wonder if the `slime-contrib` definition is useful: the user could just add the minor mode to their `slime-repl-mode-hook` instead.
- The `lambda` [here](https://gitlab.com/augfab/slime-repl-ansi-color/blob/1a52df1baddcaa3494528900c4d618af3902c1da/slime-repl-ansi-color.el#L41) is redundant, even once switched to a minor mode.

Hope that helps!

purcell:
Sent you fixes for the above in https://gitlab.com/augfab/slime-repl-ansi-color/merge_requests/1.

augfab:
Thanks a lot, very interesting feedback, and thanks for the patch!I've just merged it on my side.

purcell:
Great, merged now - welcome to MELPA! :tada:

PR #6084, opened 2019-03-20
Recipe for android-env

fernando-jascovich:

This package is a kind of helper for developing android applications/libraries using emacs.

https://github.com/fernando-jascovich/android-env.el

I'm the only maintainer of the package.

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/android-env
+++ b/recipes/android-env
@@ -1,0 +1,3 @@
+(android-env
+ :fetcher github
+ :repo "fernando-jascovich/android-env.el")


fernando-jascovich:
Thanks Chris, I will update the code to include your suggestions and fix the warnings of `package-lint`.

fernando-jascovich:
Chris, I have updated the code including all your comments and fixing warnings reported by `package-lint`. 
Regarding `cl`, I replaced that method using elisp only, so now I don't have that dependency anymore.
Please let know how do you see the code now.
Thanks.

riscy:
Nice work - the fixes look right and all my checks pass cleanly.  You can checkmark the box about package-lint in the description of this PR. :)

Another thing I notice is that you're using `ido-completing-read` but you could just use `completing-read` which would let your users use `ivy` or `helm` or whatever package they're using for completions rather than being dropped into `ido` -- the let/setq on lines 165/166 might be written `(let ((selected (completing-read "Select avd: " (android-env-avd-list))`.  If I had a final nit it'd be that the horizontal rule on line 245 appears to be one character too short, but that's a reach.

riscy:
Oh yeah, is it worth wrapping your `async-shell-command` calls in a `shell-quote-argument`?  Looks like all of the variables involved are trusted, but it might be worth doing to avoid potential injection attacks.

fernando-jascovich:
Chris,

I have updated my code again including your comments.
Thanks for the review, I really appreciate this level of advice. As you
may have seen, I am not an elisp-guru and this is very helpful for my
understanding of the standards of the language.

-- 
Fernando Jascovich
developer
m: +54 9 3548 63 9833
github: https://github.com/fernando-jascovich/
linkedin: https://www.linkedin.com/in/fernandojascovich/

fernando-jascovich:
I don't follow your comment regarding `shell-quote-argument`. What is the cost of wrapping `async-shell-command` that way?

riscy:
A good rule of thumb is to not let anything go to the shell without quoting it.  In reality it probably doesn't matter.  :)  But here's my contrived example:
- A: hey, I'm using this cool package to develop in Emacs but I'm having trouble getting getting X to work.
- B: I've seen that happen - a recent version of the emulator needs some bytes added to the header, but you can fix it by setqing `android-env-emulator-command` to  `<nasty hex-obfuscated command>` and trying again.

fernando-jascovich:
Ok, there are two calls using `async-shell-command`:

- `android-env-avd`: I used `shell-quote-argument` without problems
  here, please check updated code.

- `android-env-auto-dhu`: I tried to wrap call's arguments inside
  `shell-quote-argument` here but the function (in linux) prepends a
  backslash before the colon, so the command gets executed as
  `$ANDROID_SDK_ROOT/platform-tools/adb forward tcp\:5277
  tcp\:5277`. This causes a problem with `adb forward` which does not
  recognize that as a valid argument. Do you see any way of wrapping
  this argument without having this issue?  

> A good rule of thumb is to not let anything go to the shell without
> quoting it. In reality it probably doesn't matter. :) But here's my
> contrived example:
>
>     A: hey, I'm using this cool package to develop in Emacs but I'm
>     having trouble getting getting X to work.
>     B: I've seen that happen - a recent version of the emulator needs
>     some bytes added to the header, but you can fix it by setqing
>     android-env-emulator-command to <nasty hex-obfuscated command>
>     and trying again.
>
> -
> Reply to this email directly, view it on GitHub, or mute the thread.*
>
>
>

-- 
Fernando Jascovich
developer
m: +54 9 3548 63 9833
github: https://github.com/fernando-jascovich/
linkedin: https://www.linkedin.com/in/fernandojascovich/

riscy:
Thanks for indulging me.  :)  I think I was wrong about that one -- there's nothing coming from the user that you'd need to be careful to escape on that one.  I'm not a MELPA maintainer either, so you can take me with a grain of salt.

fernando-jascovich:
Do you think there is something missing or needing a change in order to merge this PR?

purcell:
Thanks, and sorry for the slow turnaround - I'm very snowed under right now. Couple of quick notes:
- You can replace most of [this function](https://github.com/fernando-jascovich/android-env.el/blob/d494971c4cd3b245f85934d53b8895a62d6e6245/android-env.el#L70-L83) with a call to `locate-dominating-file` (passing a lambda).
- Don't [modify the global version](https://github.com/fernando-jascovich/android-env.el/blob/d494971c4cd3b245f85934d53b8895a62d6e6245/android-env.el#L96-L99) of `compilation-error-regexp-alist`, and in particular, don't do it every time you compile. Instead, add an `android` entry to `compilation-error-regexp-alist-alist`, then add `android` to the *local* version of `compilation-error-regexp-alist` when you compile. Incidentally, you can also use `define-compilation-mode` to wrap this up more neatly.
- Throughout [this code](https://github.com/fernando-jascovich/android-env.el/blob/d494971c4cd3b245f85934d53b8895a62d6e6245/android-env.el#L113-L137) you should be using `shell-quote-argument` for every path.
- Don't separate the `fmt` var [here](https://github.com/fernando-jascovich/android-env.el/blob/d494971c4cd3b245f85934d53b8895a62d6e6245/android-env.el#L124-L127) from the call to `format` -- it just makes things harder to read, since the number of placeholders in one must match the number of args in the other.
- I think you meant to use `when` instead of `while` [here](https://github.com/fernando-jascovich/android-env.el/blob/d494971c4cd3b245f85934d53b8895a62d6e6245/android-env.el#L196-L198).
- You need `shell-quote-argument` [here too](https://github.com/fernando-jascovich/android-env.el/blob/d494971c4cd3b245f85934d53b8895a62d6e6245/android-env.el#L235) -- wrapping in quotes won't help you if the arg contains quotes or backslashes...
- I'm not going to grumble about it too much, but the `hydra`-related code isn't really the way you should do things. Being able to `require` hydra on a particular system doesn't mean that the user wants to use it. Generally option add-on behaviour which requires additional installable packages should go in a separate package with a propert dependency. Here it's marginal, but just wanted to mention the general principle.

Hope that helps!

fernando-jascovich:
Steve,

Thank you so much for the review. I really appreciate this and it helps
me a lot with my understanding on how things should be done at elisp.

I have updated my code trying to comply with the comments, please let me
know your findings.

Regarding hydra add-on behavior, I added a boolean
custom var
(https://github.com/fernando-jascovich/android-env.el/blob/master/android-env.el#L64)
allowing user to make the decision about its usage.

Thanks,

> Thanks, and sorry for the slow turnaround - I'm very snowed under
> right now. Couple of quick notes:
>
>     You can replace most of this function with a call to
>     locate-dominating-file (passing a lambda).
>     Don't modify the global version of
>     compilation-error-regexp-alist, and in particular, don't do it
>     every time you compile. Instead, add an android entry to
>     compilation-error-regexp-alist-alist, then add android to the
>     local version of compilation-error-regexp-alist when you compile.
>     Incidentally, you can also use define-compilation-mode to wrap
>     this up more neatly.
>     Throughout this code you should be using shell-quote-argument for
>     every path.
>     Don't separate the fmt var here from the call to format -- it
>     just makes things harder to read, since the number of
>     placeholders in one must match the number of args in the other.
>     I think you meant to use when instead of while here.
>     You need shell-quote-argument here too -- wrapping in quotes
>     won't help you if the arg contains quotes or backslashes...
>     I'm not going to grumble about it too much, but the hydra-related
>     code isn't really the way you should do things. Being able to
>     require hydra on a particular system doesn't mean that the user
>     wants to use it. Generally option add-on behaviour which requires
>     additional installable packages should go in a separate package
>     with a propert dependency. Here it's marginal, but just wanted to
>     mention the general principle.
>
> Hope that helps!
>
> -
> Reply to this email directly, view it on GitHub, or mute the thread.*
>
>
>

--
Fernando Jascovich
developer
m: +54 9 3548 63 9833
github: https://github.com/fernando-jascovich/
linkedin: https://www.linkedin.com/in/fernandojascovich/

purcell:
Cool, sent you fixes in https://github.com/fernando-jascovich/android-env.el/pull/1, but I think that you're still building commands with unsafe interpolation [here](https://github.com/fernando-jascovich/android-env.el/blob/f86eb174fd99aff15989293e7871eb7481075fb4/android-env.el#L192) and [here](https://github.com/fernando-jascovich/android-env.el/blob/f86eb174fd99aff15989293e7871eb7481075fb4/android-env.el#L199).

fernando-jascovich:
Steve,

Thanks for the fixes, I already merged braches.
Regarding unsafe interpolation, I'm having some issues with
`shell-quote-argument` and/or `shell-quote-wildcard-pattern`.
For example, here:
https://github.com/fernando-jascovich/android-env.el/blob/f86eb174fd99aff15989293e7871eb7481075fb4/android-env.el#L102
The actual string contained in `gradle-cmd` may be something like
"app:clean app:tasks" which after applying shell-quote functions results
in:
- shell-quote-argument: "app\:clean\ app\:tasks"
- sheel-quote-wildcard-pattern: "app:clean\ app:tasks"

Both of the results are incompatible with the format expected by gradle
compiler. Any of the results represents an error for it. The resulting
string should be, in this case, exactly as passed initially ("app:clean
app:tasks").

There is another quoting method that I should use for these kind of
arguments?

Thanks,

> Cool, sent you fixes in fernando-jascovich/android-env.el#1, but I
> think that you're still building commands with unsafe interpolation 
> here and here.
>
> -
> Reply to this email directly, view it on GitHub, or mute the thread.*
>
>
>

-- 
Fernando Jascovich
developer
m: +54 9 3548 63 9833
github: https://github.com/fernando-jascovich/
linkedin: https://www.linkedin.com/in/fernandojascovich/

purcell:
Merged now - thanks for your patience!

> Regarding unsafe interpolation, I'm having some issues with `shell-quote-argument` and/or `shell-quote-wildcard-pattern`. For example, here:

Not really sure, sorry. I don't have time to get my head fully into what's going on here and offer specific suggestions. :-)

PR #6085, opened 2019-03-21
Add recipe for theme-magic

jcaw:

Theme-magic automatically applies the current Emacs theme to all your Linux terminals using [Pywal](https://github.com/dylanaraps/pywal), to create a cohesive color scheme across your desktop. (There's a demo gif in the readme, which may be clearer).

https://github.com/jcaw/theme-magic

Author

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/theme-magic
+++ b/recipes/theme-magic
@@ -1,0 +1,5 @@
+(theme-magic
+ :fetcher github
+ :repo "jcaw/theme-magic"
+ ;; Must explicitly download the python scripts folder
+ :files (:defaults "python"))


jcaw:
Whoops - hadn't pushed a couple of the MELPA compliance changes to GitHub. It's up-to-date now.

jcaw:
Thank! Glad you like it.

>You could (require 'seq) since you use seq-into

Good idea. I'll add that in.

>Admittedly it may not be as easy, but couldn't all of this be done in Emacs using the built-in json.el functions?

I thought about this. The main reason I used Python was that I'd already written a Python suite for manipulating Pywal - I just had to rip out the relevant functions. Aside from that, it gives me more options for adapting to changes in Pywal, since I can import it directly.

(That's also why it's spread across the three files - I ripped them directly then trimmed them down. I'll merge them. It's cleaner.)

>You could potentially remove the Python dependency entirely.

Pywal already requires Python (it's a pip package), so there shouldn't be any additional dependencies. 

>some of the python code looks a little unsafe - running wal --theme "{}" could be an opening for an injection attack; why not use subprocess.call?

I hadn't considered that. I'll switch it.

By the way, if you have any more criticisms please go wild, the more feedback the better.

jcaw:
Suggestions merged here:

https://github.com/jcaw/theme-magic/pull/1

purcell:
Thanks for this, and sorry for the slow response. Mostly looks great. Quick notes:

- `seq` is not available in the Emacs version you're targeting. Add a `Package-Requires` dependency on it.
- You might like to use `zerop` [here](https://github.com/jcaw/theme-magic/blob/master/theme-magic.el#L188)
- You've got `*enable-auto-update` and `*disable-auto-update` functions -- what you want is a global minor mode: just use `define-minor-mode` for this. :-)

Hope that helps!

jcaw:
No worries, I'm sure everyone is busy. 

Suggestions merged again, explicit link here: https://github.com/jcaw/theme-magic/pull/2

I'm not sure why I avoided a minor mode initially. I think it was because Emacs would be dictating the state of Linux, i.e. the scope of the minor mode would be beyond Emacs. Using a minor mode is much neater though, and it makes the state more explicit.

On question - I haven't used the word "global" in the minor-mode name, since it would be verbose. Is it standard practice to do that for all global minor modes?

purcell:
> e state explicit.
> 
> One question - I haven't used the word "global" in the minor-mode name, since it would be verbose. Is it standard practice to do that for all global minor modes?

Nope, that's fine. If there's a pair of local and global modes, which is frequently the case, then it's conventional to name the latter accordingly.

Merging now - welcome to MELPA! :tada:

PR #6086, opened 2019-03-21
In HTML5 the meta and link tags don't need a closing slash.

jbampton:
Minor fix to use HTML5 for the `meta` and `link` tags on the web pages.


purcell:
Thanks.

jbampton:
Thank you Steve 👍

PR #6088, opened 2019-03-23
Add battle-haxe recipe

AlonTzarafi:

This package offers a development system for the Haxe programming language.
Code completion is supported using the [company ](http://company-mode.github.io/) package.
Options like "go to definition" and "find all references" are available, as well as eldoc support.

https://github.com/AlonTzarafi/battle-haxe

I am the author/maintainer.

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/battle-haxe
+++ b/recipes/battle-haxe
@@ -1,0 +1,3 @@
+(battle-haxe :fetcher github
+			 :repo "AlonTzarafi/battle-haxe"
+			 :files ("*.el"))


AlonTzarafi:
@riscy Thank you for all the elisp tips!
I've pushed a commit that fixes those issues.

purcell:
Thanks, and sorry for the slow turnaround. Some notes:

- Odd formatting makes some of this code hard to read. Don't put the "if" condition on a new line, and when the "else" branch is `nil`, just use `when` or `unless` instead of `if`. Same with `let` and `let*`: formatting idiomatically encourages contributors.
- I like that you've made `yasnippet` optional and that you're checking `yas-minor-mode`. I actually think you should drop the `yasnippet` package dependency (you might need to use `declare-function` for `yas-expand-snippet` if so).
- (Minor) You're using `dash` for just a couple of things that built-ins (`mapcar`, `cl-filter` etc.) could handle, so you might consider dropping it.
- The `battle-haxe-helm-find-references` function should be integrated as an `xref` backend.
- You can use `pcase-let` to bind multiple vars in cases like [this](https://github.com/AlonTzarafi/battle-haxe/blob/master/battle-haxe.el#L990-L993), just in case you're interested.
- Re. multiple project support in `battle-haxe-start-server` etc., what you'd usually do is have a separate server buffer per project, then make all vars buffer-local inside that buffer. This way, to find if there's a server running for the current project, you just have to find the corresponding server buffer and check inside it. (You had the right idea in [this commented-out line](https://github.com/AlonTzarafi/battle-haxe/blob/ca49250000fd9834baffe88e5fb6257043ab7b7d/battle-haxe.el#L966))
- Consider the `1-` function instead of `cl-decf`.
- `battle-haxe-services-mode` should probably just be called `battle-haxe-mode`, and you should just inline `battle-haxe-init` there for clarity.

Anyway, I think this can all be addressed after merging, so I'll merge now: welcome to MELPA! :tada:

AlonTzarafi:
Forgot to say but thank you for the tips. It's really helped me a lot!!

PR #6092, opened 2019-03-26
Add recipe for better-jumper

gilbertw1:

A configurable jump list implementation for Emacs that can be used to easily jump back to previous locations. This is a more configurable jump list than the one built into `evil` that plays nicely with packages that manage window configurations such as `persp-mode` and is available to non-evil users as well.

https://github.com/gilbertw1/better-jumper

maintainer

None Needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/better-jumper
+++ b/recipes/better-jumper
@@ -1,0 +1,2 @@
+(better-jumper :repo "gilbertw1/better-jumper"
+               :fetcher github)


dzop:
You may be interested in my branch of `evil` (https://github.com/dzop/evil/blob/jumplist/evil-jumps.el) that also fixes the jumplist implementation to work with packages like `persp-mode` using window properties. It also brings the jumplist implementation more in-line with how jumps work in Vim, e.g. one jump point per line, no duplicated jumps, when jumping from a previous location on the jumplist the jump is "bubbled up" to the top of the jumplist. 

I too was planning on submitting a pull request to `evil`, but given the state of maintenance in `evil` I doubt that it would have ever gotten a proper review as my changes are a fairly large re-implementation of the jumplist. Maybe someday...

Regarding your implementation:

- `let*` is unnecessary [here](https://github.com/gilbertw1/better-jumper/blob/3a0d02071a5aabfb9a9cc9f3e8d2941326cfb610/better-jumper.el#L107), [here](https://github.com/gilbertw1/better-jumper/blob/3a0d02071a5aabfb9a9cc9f3e8d2941326cfb610/better-jumper.el#L203), and [here](https://github.com/gilbertw1/better-jumper/blob/3a0d02071a5aabfb9a9cc9f3e8d2941326cfb610/better-jumper.el#L291)
- [This](https://github.com/gilbertw1/better-jumper/blob/3a0d02071a5aabfb9a9cc9f3e8d2941326cfb610/better-jumper.el#L319-L320) can just be one `let*`
- [This](https://github.com/gilbertw1/better-jumper/blob/3a0d02071a5aabfb9a9cc9f3e8d2941326cfb610/better-jumper.el#L323) can just be `unless`
- You are only using `point` to save jumps on the jumplist (see [here](https://github.com/gilbertw1/better-jumper/blob/3a0d02071a5aabfb9a9cc9f3e8d2941326cfb610/better-jumper.el#L206)). It would be better to use `point-marker` so that the jump position is maintained if text is inserted or deleted.  
- From the README I get the sense that this package can replace `evil`s jumplist, but it is missing the `savehist` integration `evil` provides. When using this package there may be missing entries in the jumplist saved by `evil` if users override `C-o` and `C-i` as stated in the README. This is because `better-jumper--push` may be called in `better-jumper-jump-{backward|forward}` which only adds to the `better-jumper` jumplist and not the `evil` jumplist. Perhaps you can clarify that your package is not a drop in replacement or try to also implement `evil`s `savehist` support.
- Also I think it would be simpler to advise `split-window-internal` (using around advise) instead of hooking into `window-configuration-change-hook` since you will then have a "source" window (passed as the first argument to `split-window-internal`) and a "target" window (provided by calling the original definition of `split-window-internal`). Just my opinion, I don't know how encouraged it is to advise such a function.

gilbertw1:
@dzop Wow, thanks for the detailed feedback! I really appreciate it.

> more in-line with how jumps work in Vim, e.g. one jump point per line, no duplicated jumps, when jumping from a previous location on the jumplist the jump is "bubbled up" to the top of the jumplist.

This definitely seems like desireable behavior. I'll definitely be looking at your implementation to improve the one in `better-jumper`. This also gives me the idea to potentially include these behaviors as optional settings that could be configured in the spirit of customizeablility.

>* let* is unnecessary here, here, and here
>* This can just be one let*
>* This can just be unless

These make sense, I'll make the corresponding updates.

> You are only using point to save jumps on the jumplist (see here). It would be better to use point-marker so that the jump position is maintained if text is inserted or deleted.

This is a side effect of an initial implemention I was attempting to use to specifically save history in a perspective. I'll evaluate moving back to using a `point-marker`, thanks for pointing that out. I'm curious if it will serialize properly when window configurations are saved.

> From the README I get the sense that this package can replace evils jumplist, but it is missing the savehist integration evil provides...

I'm definitely planning on implementing savehist support when `better-jumper`is tracking jumps in the buffer context, probably in the next day or so. I'm not so sure a savehist implementation will make sense in the context of windows, since I'm using persistent window parameters to store the jump lists. This means that the jump lists are stored and loaded alongside window configurations.

I'll have to think about how savehist would work with windows and maybe this is also something that should be configurable. I'll consult your implementation as well here.

This does bring up a bigger issue that I need to clarify the ways that `better-jumper` differs from evil's jump implementation, and steer clear of implying it's an exact drop in replacement. My ambition here is that this is something that can be used in the place of evil's jump list, but is not exactly the same, and can easily be integrated into non-evil users's workflows.

> Also I think it would be simpler to advise split-window-internal (using around advise) instead of hooking into window-configuration-change-hook since you will then have a "source" window...

I'll definitely look into this, thanks for the pointer!


I appreciate all of this feeback a ton. You seem very knowledgeable in this area, if you're interested you're welcome to contribute to `better-jumper`. I feel like it could be a great addition to the Emacs ecosystem if done properly. 

Thanks again!

dzop:
About the savehist integration, yes I think there needs to be some more thought put into this when considering window configurations. I just merge all the jumps from all the windows in the current window configuration to a single jumplist so that I save the most recent jumps in all the currently visible windows. The downside to this is that it doesn't consider jumps in other window configurations, e.g. when using `persp-mode`.

One strategy would be to detect when something like `persp-mode` is being used and save the window jumplists for each perspective (either on a per window basis or merged into a single jumplist). You can make an alist mapping perspective names to the saved jumplists and store that in a variable that savehist saves. Then when re-loading the jumplists on an Emacs restart, whenever a perspective with a name matching one in the saved alist is created, you can associate the corresponding jumplist with the window configuration of the perspective.

> if you're interested you're welcome to be a contributor to `better-jumper`

I'm happy with what I have now, once your package matures I'll maybe re-evaluate. Thanks for the offer.

purcell:
Thanks, and sorry about the slow turnaround. Quick notes:

- You can't have this auto-enable everything when loaded. For starters, it stops people from building their own functionality by using your package as a library. Instead, define a minor mode which people can enable. Bonus: when advising functions, you can then check the value of the minor mode in order to decide whether or not to do the enhanced action. (For example, [here](https://github.com/gilbertw1/better-jumper/blob/385f53b7103e2e7c53515f31283182ff0481532b/better-jumper.el#L482).)
- You might consider using `pcase` to switch on `better-jumper-context` in places like [this](https://github.com/gilbertw1/better-jumper/blob/385f53b7103e2e7c53515f31283182ff0481532b/better-jumper.el#L215-L218) - although in that specific case, you should be using `memq` because the code is identical in both branches of the `cond`.
- Does the `defgroup`'s parent group `jump` actually exist? I suspect not.
- `cl-lib` is built into the version of Emacs you're depending on, so you can drop the package dependency declaration for that library.

Otherwise this all looks pretty solid. Hope that helps!

gilbertw1:
@purcell Thanks for the detailed feedback, it definitely helps a ton!

I'll get started on these updates asap.

gilbertw1:
@purcell 

I've applied each of the suggestions you made to `better-jumper` and added a minor-mode to restrict auto-enable / function advice.

Thanks again for the feedback!

purcell:
Looks good - merged now, thanks!

PR #6095, opened 2019-03-27
Add a recipe for parseedn

bbatsov:

An EDN parser/reader for Emacs. That's a follow up of #5258. Seem we forgot to add a recipe for the second package.

https://github.com/clojure-emacs/parseedn

Co-maintainer. 

### Checklist

Please confirm with `x`:

- [x] I have confirmed some of these without doing them (🤣 )

### DIFF
--- a/recipes/parseedn
+++ b/recipes/parseedn
@@ -1,0 +1,1 @@
+(parseedn :repo "clojure-emacs/parseedn" :fetcher github)


bbatsov:
I've gotten to a point where I could really benefit from using this in CIDER, so I hope someone will merge this quickly.

tarsius:
I think you should extend the library commentary to be more like the first section of the readme.

bbatsov:
Will do! Thanks!

PR #6096, opened 2019-03-27
Move Related package repository from Bitbucket to GitHub

julien-montmartin:
Hello,

I'm moving my package from Bitbucket to GitHub. This is not a new package, but I need to update its recipe.

Please see on the old Bitbucket repo (https://bitbucket.org/julien-montmartin/related/src/master) my last commit "Moving to https://github.com/julien-montmartin/related" to be sure that I'm the real author of the package.

Best regards,

Julien


purcell:
Great, thanks.

PR #6097, opened 2019-03-27
add recipe for which-key-posframe

yanghaoxie:

Let emacs-which-key use posframe to show its popup. 

https://github.com/yanghaoxie/which-key-posframe

Maintainer

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/which-key-posframe
+++ b/recipes/which-key-posframe
@@ -1,0 +1,1 @@
+(which-key-posframe :fetcher github :repo "yanghaoxie/which-key-posframe")


purcell:
Thanks, and sorry about the slow response. Quick notes:

- You can delete [these lines](https://github.com/yanghaoxie/which-key-posframe/blob/d3b5adcbdabf726ce5a26fbae8efdce3a05b3d82/which-key-posframe.el#L34-L35) because you're loading posframe anyway!
- Add `(require 'cl-lib)` because you're using `cl-case` etc.
- Remove the `message` call [here](https://github.com/yanghaoxie/which-key-posframe/blob/d3b5adcbdabf726ce5a26fbae8efdce3a05b3d82/which-key-posframe.el#L136-L146), and change this to a global minor mode if you can, so that the behaviour could (potentially) be toggled off.

In any case, I can merge this now, and you can apply the above suggestions at your convenience. Welcome to MELPA! :tada:

yanghaoxie:
Thank for your excellent review and hard work, I will merge these suggestions into my code later.
I am very excited, since this is my very first little emacs package. haha.
Happy weekend!

PR #6098, opened 2019-03-27
Add recipe for dash-docs

gilbertw1:

The `dash-docs` package is the result of extracting the dash specific logic from `helm-dash`. This will allow dash documentation to be leverage from other packages such as `counsel-dash` without needing to require `helm` and helm specific logic.

There is a detailed issue containing information on this restructuring that you can find here: areina/helm-dash#173

https://github.com/gilbertw1/dash-docs

maintainer

None Needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/dash-docs
+++ b/recipes/dash-docs
@@ -1,0 +1,2 @@
+(dash-docs :repo "gilbertw1/dash-docs"
+                 :fetcher github)


kidd:
Glad that this is finally happening :)

Props to @gilbertw1 for an amazing refactor!

gilbertw1:
@riscy Thanks for the feedback!

I've got a fair bit of cleanup that I am planning on performing, however my plan is to migrate both `helm-dash` and `counsel-dash` to this new package with only minimal changes to the code that was extracted from [helm-dash](https://github.com/areina/helm-dash). Once we've made this transition successfully, I'm planning on doing more.

@kidd Thanks!

purcell:
Nice work, and sorry for the slow response. I'm just now working through around 65 pull requests, and this is around half-way. :-)

This all looks good to me, so merging now.

PR #6100, opened 2019-03-29
Add recipe for github-explorer

TxGVNN:

Explore a Github repository on the fly in Emacs 

https://github.com/TxGVNN/github-explorer

Maintainer

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/github-explorer
+++ b/recipes/github-explorer
@@ -1,0 +1,3 @@
+(github-explorer
+ :repo "TxGVNN/github-explorer"
+ :fetcher github)
--- a/recipes/github-mode
+++ b/recipes/github-mode
@@ -1,0 +1,3 @@
+(github-mode
+ :repo "TxGVNN/github-mode"
+ :fetcher github)
--- a/recipes/github-mode
+++ b/recipes/github-mode
@@ -1,3 +1,0 @@
-(github-mode
- :repo "TxGVNN/github-mode"
- :fetcher github)


riscy:
Byte-compiling:
```
In github-go-at-point:
github-mode.el:60:32:Warning: assignment to free variable ‘item’
github-mode.el:73:36:Warning: reference to free variable ‘item’
github-mode.el:77:1:Warning: Unused lexical variable ‘&allow-other-keys&rest’

In github--tree:
github-mode.el:92:26:Warning: ‘beginning-of-buffer’ is for interactive use
    only; use ‘(goto-char (point-min))’ instead.
github-mode.el:97:26:Warning: ‘beginning-of-buffer’ is for interactive use
    only; use ‘(goto-char (point-min))’ instead.
github-mode.el:106:1:Warning: Unused lexical variable ‘&allow-other-keys&rest’

In github--raw:
github-mode.el:120:26:Warning: ‘beginning-of-buffer’ is for interactive use
    only; use ‘(goto-char (point-min))’ instead.
github-mode.el:128:1:Warning: Unused lexical argument ‘norecord’
github-mode.el:128:1:Warning: Unused lexical argument ‘action’
github-mode.el:128:1:Warning: Unused lexical argument ‘buffer-or-name’
```

TxGVNN:
Hi @riscy, sorry for my ignorance.
I updated https://github.com/TxGVNN/github-mode/commit/cacc636021926eab953c3dba8c298d5ab588a044

akirak:
I'd better suggest using https://github.com/magit/ghub, especially for supporting other services such as gitlab.

TxGVNN:
Hi @riscy, Thanks your continued review

>     * I believe your source file should just be called "github.el" (although my gut feeling is that the name "GitHub" is too generic for what your package actually does -- maybe `github-browse`? -- we should wait for an actual MELPA maintainer to weigh in on this one)
Because this package works directly with GitHub website, so I think it's github-mode. If we really need to rename to new name, I totally followed

> 
>     * It's worth it to include your license in your source file, in case someone passes that around instead of a link to your repository -- just check out how the license is summarized in most other elisp files
I updated 
> 
>     * Everybody makes this mistake, but Github is written GitHub :)
I just check github homepage and updated :D
> 
>     * Replace occurrences of deprecated `function*` with `cl-function`
Do you know? I replaced `cl-function` to `function*` after your first comment =))). I reverted

> 
>     * You could potentially declare and define your keymap all in the same place, e.g.
>       ```emacs-lisp
>       (defvar github-mode-map
>         (let ((keymap (make-sparse-keymap)))
>           (define-key keymap (kbd "o") 'github-go-at-point)
>            ...
>           keymap)
>         "Keymap for GitHub major mode.")
>       ```
Thank you, I learned something. Updated!
> 
>     * I don't think it's necessary to advise `pop-to-buffer` to activate the mode, unless I'm missing something.  Why not just add a call to `(github-mode)` before running `(pop-to-buffer`?
The reason is I need to load major-mode for file type. If I open `README.md`, it will load `markdown-mode`. Do you have a better idea?
> 
> 
> Minor:
> 
>     * The message `Finished!`is perhaps unnecessary (I'm thinking of when we open `dired` -- we don't get a message that says it finished), but I'm opinionated
Removed
> 
>     * There are some scratch comments still in the code (e.g. a call to `github-repo`, a function that no longer exists, is commented out), and some notes about future work that maybe belong in the README.  Worth a quick cleanup?
Cleaned
> 
>     * The double vertical space between functions is probably unnecessary unless you've got a good reason for it.  Note there's a convention of using `^L` on a line by itself to delineate sections in elisp code if you need it.
>
Removed
> 
> Usability-wise (possibly for future):
> 
>     * Is it worth adding a `..` link when you're in a nested directory of a repo to make it easy to get back "up"?
> 
>     * Only being able to browse `master` strikes me as a big limitation -- there are several projects where I'm more interested in e.g. the `develop` branch than the `master` branch.
Yes, It should more future feature. I hope so :(

TxGVNN:
> I'd better suggest using https://github.com/magit/ghub, especially for supporting other services such as gitlab.

Yes, perhaps, but I haven't had time to read and try it.

purcell:
Thanks, and sorry for the slow response. Some notes:

- [Here](https://github.com/TxGVNN/github-mode/blob/d4b45839b26d4222d7881d780a2da7c6baf3021f/github-mode.el#L180-L184) you should use `special-mode` as the parent mode (and then remove the `q` binding [here](https://github.com/TxGVNN/github-mode/blob/d4b45839b26d4222d7881d780a2da7c6baf3021f/github-mode.el#L56)). Also `use-local-map` is called for you by `define-derived-mode`, so you can delete that line too.
- The [`github-mode-name` var](https://github.com/TxGVNN/github-mode/blob/d4b45839b26d4222d7881d780a2da7c6baf3021f/github-mode.el#L45) is pointless: just inline it [here](https://github.com/TxGVNN/github-mode/blob/d4b45839b26d4222d7881d780a2da7c6baf3021f/github-mode.el#L180). That's not a thing that is useful to customise. :-)
- I kinda agree that the package should be called `github-browse` or `github-explorer`. `github-mode` is okay, *except* that all of the symbols then use the `github-` prefix, which is quite likely to clash with other packages. Also, I don't think that people will think "oh, now I want to open this buffer in github mode" - instead, they will think, "I'd like to browse github". The major mode here is kind of an implementation detail.

Hope that helps!

TxGVNN:
Thanks for your help.
>     * [Here](https://github.com/TxGVNN/github-mode/blob/d4b45839b26d4222d7881d780a2da7c6baf3021f/github-mode.el#L180-L184) you should use `special-mode` as the parent mode (and then remove the `q` binding [here](https://github.com/TxGVNN/github-mode/blob/d4b45839b26d4222d7881d780a2da7c6baf3021f/github-mode.el#L56)). Also `use-local-map` is called for you by `define-derived-mode`, so you can delete that line too.
I updated https://github.com/TxGVNN/github-explorer/commit/e36208b8e4f7554f09e5cdaae674c7c08556d1d7
> 
>     * The [`github-mode-name` var](https://github.com/TxGVNN/github-mode/blob/d4b45839b26d4222d7881d780a2da7c6baf3021f/github-mode.el#L45) is pointless: just inline it [here](https://github.com/TxGVNN/github-mode/blob/d4b45839b26d4222d7881d780a2da7c6baf3021f/github-mode.el#L180). That's not a thing that is useful to customise. :-)
In my case, I see this is good feature to anyone want to change the name of mode. Because I sometimes change some modes :3 
> 
>     * I kinda agree that the package should be called `github-browse` or `github-explorer`. `github-mode` is okay, _except_ that all of the symbols then use the `github-` prefix, which is quite likely to clash with other packages. Also, I don't think that people will think "oh, now I want to open this buffer in github mode" - instead, they will think, "I'd like to browse github". The major mode here is kind of an implementation detail.
> 
I renamed to `github-explorer` https://github.com/TxGVNN/github-explorer/commit/e36208b8e4f7554f09e5cdaae674c7c08556d1d7

purcell:
Nice, merging now, thanks!

PR #6101, opened 2019-03-29
Add live-preview recipe

lassik:

Renders a live preview of whatever you are editing in a side window.

You can give any shell command or Emacs Lisp function to render the preview. The preview is rendered whenever you are idle for a few seconds. Different buffers can have different preview commands. There is only one global preview buffer; whenever you go idle in a buffer that has a preview command, the preview buffer is updated with a preview of that buffer.

This is useful for previewing e.g. manual pages or other documentation while writing them. Instead of a preview, could also run a validator or crunch some statistics.

Maybe in the future: graphics support (e.g. render HTML or TeX as an image and show it in Emacs).

https://github.com/lassik/emacs-live-preview

Just wrote it

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/live-preview
+++ b/recipes/live-preview
@@ -1,0 +1,1 @@
+(live-preview :fetcher github :repo "lassik/emacs-live-preview")


lassik:
Thank you for the excellent and thorough review!

> * I think an example of a string to set `live-preview-command` to would be helpful.

I fully agree. It would be a good idea to emphasize that the command needs to accept text from standard input. Your examples are very good.

> * I tried writing a function to work with live-preview.  Am I following best practices here?  Maybe these could be documented too:

Yes, you did everything right. I should have made it more obvious that `M-x live-preview` only accepts a shell command, not an Emacs Lisp function name. (To use a function, you need to do `(setq live-preview-command 'my-live-preview-command)` in a hook or using `M-x eval-expression`. The reason elisp functions are de-emphasized is that I thought only advanced users would be interested in them. I started with shell commands only and added functions as an afterthought, in the spirit of "they're easy to add so why not".

> * `live-preview-show` is a big function, maybe too big:

Yes, it's definitely not pretty code (but then, it also deals with grubby subprocess stuff so there's probably a limit to how clean it can be made). I struggled a lot with some timeout issue that I don't really even remember very well anymore. The idea is that more than likely every user (including me) will set some command by mistake that will run for ages, or will produce tons of output. Since the user's command is invoked repeatedly by an idle timer, `live-preview-show` has to do something reasonable so that Emacs doesn't hang in those cases. It has to be pretty strict, probably more strict than with subprocesses in general.

>   * The nested conditionals in (especially `unless` sometimes in the form `(unless (or x (and y z)) ...)`) get a little hard for me to parse. What are your thoughts on rewriting the first unless to:
>     ```emacs-lisp
>     (unless (or (eq src-buf pre-buf)
>                    (null command)
>                    (and (stringp command) (string-blank-p command)))

I fully agree that the nesting is pretty bad :p Those `unless` expressions could probably be combined, there is not a clear rationale for keeping them separate. Also this code between them could be extracted into its own `live-preview-interrupt` function:

        (when (get-buffer-process (current-buffer))
          (interrupt-process)
          (while (process-live-p (get-buffer-process (current-buffer)))
            (sleep-for 0.1)))

>   * I think there is value in factoring the two `command` processors out into their own functions (e.g. `live-preview--show-string` and `live-preview--show-function`)

Also a good idea. I will do that.

>   * Could `(* 100 1024)` be factored out into a constant?  Or customizable?
>   * Could `2` (as in the idle-timer) be factored out into a customizable?

The maximum preview buffer size `(* 100 1024)` should indeed be named, as you suggest.

Both of these could be variables, but I don't know if there is a point in setting them. At least to me, both of these seem like settings for which some kind of generally useful near-optimum value could be discovered:

* The maximum buffer size is just a sanity check against rogue shell commands. 100 KiB may even be too big.
* The idle timer is something that usability researchers have probably experimentally discovered an optimum value for. However, it may also be affected by command run-time -- if the preview command is not instantaneous, then one might want to run it less frequently. Maybe different preferences due to human temperament also factor into it. I don't know. Perhaps it should be customizable.

A lot of the cruft in this module is explained by the fact that I tried to write it as quickly as I could and then move on to other things (I have a large backlog of other open source work so things like this unfortunately tend to get de-prioritized). I'll try to do the cleanups you suggested by the end of the week. Thanks again for taking the time :)

lassik:
Also if anyone is interested in helping out with this package or even becoming the maintainer, that's great. I don't really have time to do other than the bare minimum myself.

lassik:
Oh well, I got enough momentum from writing the above that I just fixed those things right away :) Results now pushed to github.

riscy:
I like the changes!

I realize you might've written the above in a hurry, and I don't want to sound rude with this.  But if you aren't down for maintaining a package, a better approach might be to find a maintainer to take over, and then have him/her submit to MELPA.

(Also I should stress I'm not a MELPA maintainer -- just a fellow contributor passing through.)

lassik:
No problem. I can maintain it, but with minimum effort (fixing obvious things that people point out as above, merging pull requests and generally replying to issues). We can also leave it out of MELPA until a maintainer is found. Both approaches are fine with me.

Thanks for helping out even though you are not expected to :) It's valuable to the community.

purcell:
Sorry for the slow turnaround - merged now, thanks!

lassik:
No problem. It's important to take as much rest as you need - we are all volunteers here. Thanks for merging :)

PR #6102, opened 2019-03-29
Add nix-haskell-mode

matthewbauer:
URL: https://github.com/matthewbauer/nix-haskell-mode

Configures Haskell projects with packages downloaded from Nix.

https://github.com/matthewbauer/nix-haskell-mode

Creator

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/nix-haskell-mode
+++ b/recipes/nix-haskell-mode
@@ -1,0 +1,1 @@
+(nix-haskell-mode :fetcher github :repo "matthewbauer/nix-haskell-mode")


purcell:
Thanks, and sorry for the slow response. I have some feedback, having taken a few moments to figure out what this package aimed to do:

- Regarding the `flycheck` integration, which some people will not want, my suggestion is this: drop the package dependency and `require` clauses, but retain the code in which sets `flycheck-haskell-ghc-executable` etc. Also, then, don't explicitly enable `flycheck-mode`: a user who wants flycheck enabled will generally add it to hooks themselves, or will enable `global-flycheck-mode`.
- Consider using `or` to tidy [this](https://github.com/matthewbauer/nix-haskell-mode/blob/733198b1799e8e14152d2c2fc5f4e569fc83bbbd/nix-haskell-mode.el#L263-L274) up.
- [These comments are odd](https://github.com/matthewbauer/nix-haskell-mode/blob/733198b1799e8e14152d2c2fc5f4e569fc83bbbd/nix-haskell-mode.el#L547-L548) - are they malformatted file-local variables?

Hope that helps.

matthewbauer:
@purcell Thanks for reviewing! I think I have addressed the review comments.

On the ```flycheck-mode``` - the reason I'm doing this is to force flycheck to reload after those variables are set. Since nix-haskell-mode operates asynchronously, you will see an annoying warning between opening the buffer and nix-haskell-mode finishing its setup. This is kind of annoying so it is best to disable it until after that. At the same time, flycheck doesn't automatically recognize when you've set these variables, so you get a delay based on how often flycheck is set to refresh. This makes the normal hooks hard to work with. But, to address your concerns, I've made ```nix-haskell-flycheck``` determine whether flycheck will be enabled.

purcell:
Cool, thanks for explaining the reasoning. I understand why the async behaviour makes this tricky, and I'll go ahead and merge this now.

But I thought a little bit about alternative approaches to the repeated toggling of flycheck-mode, and I think that there's a more elegant approach: you'd define a `haskell-nix` flycheck backend which would invoke the `haskell-ghc` backend when necessary, but only after its setup is complete. Then tell users to push this backend onto `flycheck-checkers` ahead of the existing backends. Its `:enabled` form would detect when haskell nix files and executables are present. This checker could be a "custom" checker, ie. using elisp instead of an executable: this means flycheck will handle its async nature.

Such a checker would then be a neat orthogonal add-on to the rest of `nix-haskell-mode`, so that you don't need to tangle things up like this, and it could then be a little sibling package with full correct dependencies.

PR #6105, opened 2019-03-31
Update travis config to use Xenial

milkypostman:
Doing as a pull request to test if this works.


purcell:
Ah, I thought about trying this but somehow thought that it was the default already. :-)

PR #6108, opened 2019-04-03
Add recipe for swedish calendar names

matsl:

The package provides a list of names associated with each day in the Swedish calendar. In addition to that it provides a function suitable for inclusion in the `.diary` file so that the names are displayed when viewing the diary from calendar, org-agenda or similar.

https://github.com/matsl/sv-kalender-namnsdagar

**None needed**

### Checklist

Please confirm with `x`:

The function has a free variable `date` as that is bound when it is called from `diary-list-sexp-entries`. This gives a byte compilation warning.

### DIFF
--- a/recipes/sv-kalender-namnsdagar
+++ b/recipes/sv-kalender-namnsdagar
@@ -1,0 +1,1 @@
+(sv-kalender-namnsdagar :repo "matsl/sv-kalender-namnsdagar" :fetcher github)


purcell:
Thanks, and sorry for the slow response. Looks good, so merging now. Note that "celibrated" should be "celebrated". :-)

Welcome to MELPA! :tada:

PR #6111, opened 2019-04-04
Add gn-mode recipe

lashtear:

A major-mode for editing GN (ninja generator) config files in Emacs. Files of this type (e.g. BUILD.gn or *.gni) are common in Chromium-derived projects like Chrome and (Anaheim) Edge.

https://github.com/lashtear/gn-mode

Author / maintainer / recipe writer

None needed yet!

### Checklist

Please confirm with `x`:

    The package is licensed with BSD 3-clause

    ಠ_ಠ

### DIFF
--- a/recipes/gn-mode
+++ b/recipes/gn-mode
@@ -1,0 +1,3 @@
+(gn-mode
+ :fetcher github
+ :repo "lashtear/gn-mode")


riscy:
Also one `package-lint` warning: ```1:73: warning: You should depend on (emacs "24") if you need lexical-binding.```

lashtear:
Great review, thanks!

I hadn't considered `evil-mode`-- I'll try to do so in the future as it really does change motion commands.  I've implemented your fix.

The end-of-buffer related text loss is particularly bad, but apparently never came up because everything is "usually" under-indented.  `(point-max)` shrinking beneath the prior end violates several invariants there, so I've replaced `e` with a (hopefully more readable) `cl-symbol-macrolet` called `work-end` that always evaluates that `min` with `(point-max)`.  Alternatively, we could use a marker, but I think this is lower overhead.

I don't need lexical-binding, but it was strongly advised by the MELPA submission process.  I should have run package-lint again afterward-- added the `(emacs "24")` dependency.

Please take a look at the 0.4 release now; these should all be fixed.  Again, thanks for looking!

riscy:
I'm not a MELPA maintainer -- just a fellow contributor -- but it's working great for me now.

riscy:
Checking back, I may have been wrong about the `forward-char` thing -- it seems like evil doesn't move the cursor backward except after some interactivity.  Sorry for the confusion. :)

lashtear:
Even so, `forward-line` is a bit more clear about what is intended, I think, so leaving that as-is unless there is other trouble with it.  Again, thanks for looking!

purcell:
Looks great now, thanks. Merging now!

PR #6112, opened 2019-04-05
Add shim

twlz0ne:

One package for multiple *env, currently integrated rbenv/pyenv/nodenv/jenv/goenv.

At the same time, provide user an easy way to extend by self:

```elisp
(cl-defun shim-init-foo (&key (major-modes '(foo-mode)) (executable "fooenv"))
  (shim-init
   (make-shim--shim
    :language 'foo
    :major-modes major-modes
    :executable executable)))
```

https://github.com/twlz0ne/shim.el

Author&Maintainer

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/shim
+++ b/recipes/shim
@@ -1,0 +1,3 @@
+(shim
+ :fetcher github
+ :repo "twlz0ne/shim.el")


twlz0ne:
@riscy Thanks for your review. I have made changes based on your comments, please check again.
.

riscy:
I'm not a MELPA maintainer - just a fellow contributor - but everything I noticed looks fixed now!

purcell:
Thanks for this! It's an interesting idea to generalise this pattern.

However, I think the code is more complicated than it needs to be, and that choosing a different design it would make the package much simpler and better. Imagine I have a project which uses `nodeenv` and `pyvenv`: now, as far as I can tell, I can only use this package for one at a time, and `shim-mode` has some internal configuration to know which shim executable it is controlling.

What I would really want would be a `shim-pyvenv-mode` minor mode which I can add to `python-mode-hook`, and a `shim-nodeenv-mode` which I can add to `js2-mode-hook`. This would also eliminate the need to guess the current language, removing all the unnecessary code for that.

Note also that you have a local (ie. non-global) minor mode here which is apparently setting the global values of environment variables from different buffers. I think this is going to lead to clashes between projects which use different shim versions, because calling `shim-auto-set` in any buffer will affect every buffer in Emacs. (Here it's possible to make `process-environment` buffer local and set the `_VERSION` var there instead, to avoid contamination between buffers.)

My suggestion would be to do one of the following:

- Make a `defcustom` that is a list of enabled shims (which can then be set by users in a `.dir-locals.el` if desired), and retain a single `shim-mode`, or
- Provide a macro `shim-define` which expands to define a minor mode with its own name (see my `reformatter.el` for an example of this approach).

If you're only targeting "standard" shims which work in the exact way you assume, then the first approach may be easier/better. It's also similar to the `flycheck` or `company` approach. You can then define new shims either by adding their configuration to an alist of `(shim-name . config)`, or by using the function approach that `company` uses for backends.

Hope that helps!

twlz0ne:
@purcell Thanks for your review and suggestion.

The buffer local `process-environment` problem, it's a serious problem, but I've never noticed it before because I use shim with LSP.  Each time the buffer loads, then the shim changes the shell environment and the LSP server starts, after that, even the shell environment changed by another shim, the previously started LSP process does not seem to be affected.  Now I fixed it at [84096df](https://github.com/twlz0ne/shim.el/commit/84096df4b5dc2378d205cf734bc3e86f98300204) and added some use cases to cover it.

About the particular minor mode (like `shim-pyenv-mode`), it's really necessary?  I see the [LSP](https://github.com/emacs-lsp/lsp-mode/tree/1e062077d01cca6d57e656c8334d03d909c00b30#configure-lsp-mode) also just provides  the `lsp-mode` for all languages:

> ```elisp
> (require 'lsp-mode)
> (add-hook 'XXX-mode-hook #'lsp)
> ```

Every time the buffer loads, the [`lsp--find-clients`](https://github.com/emacs-lsp/lsp-mode/blob/1e062077d01cca6d57e656c8334d03d909c00b30/lsp-mode.el#L4580) is called to find available clients.

About providing a macro `shim-define`, I am thinking of it, it's cleaner than `(cl-defun ...)`.  I'm also thinking about removing the `language` slot from `shim--shim` struct, because it's not inaccurate, for example, **node** is not a language.

tarsius:
@purcell Ping!

purcell:
> @purcell Ping!

I feel like I raised a few design issues here which haven't really been addressed, in particular the unidiomatic configuration. Not sure how to proceed, because I can merge this knowing that it's going to be unnecessarily hard for people (including myself) to use effectively, but I'd rather help it get better and then merge it knowing I'd use it myself.

> About the particular minor mode (like `shim-pyenv-mode`), it's really necessary? I see the [LSP](https://github.com/emacs-lsp/lsp-mode/tree/1e062077d01cca6d57e656c8334d03d909c00b30#configure-lsp-mode) also just provides the `lsp-mode` for all languages:

No, it's not necessary: in my earlier comment I offered an alternative approach.

DamienCassou:

PR #6115, opened 2019-04-07
Add parchment-theme recipe

ajgrf:

Parchment is an editor theme for Vim and Emacs, inspired by Acme and Leuven. You might like it if you prefer light colorschemes but think black-on-white is too hard on your eyes, or if you find most colorschemes too busy.

https://github.com/ajgrf/parchment

I am the the maintainer.

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/parchment-theme
+++ b/recipes/parchment-theme
@@ -1,0 +1,3 @@
+(parchment-theme
+ :fetcher github
+ :repo "ajgrf/parchment")


purcell:
Thanks! This looks good, apart from defining a macro (`style-theme`) that doesn't begin with the package's prefix. Once that is renamed, this should be good to go.

ajgrf:
I just pushed an update to rename the macro to `parchment-style-theme`.

purcell:
Nice, merged now, thanks! :tada:

PR #6117, opened 2019-04-07
Add git-identity

akirak:

Manage local Git identities from inside Emacs

https://github.com/akirak/git-identity.el

I am the maintainer.

**None needed**

### Checklist

Please confirm with `x`:

- [ ] I have confirmed some of these without doing them because **I do them on [CI](https://travis-ci.org/akirak/git-identity.el)**

### DIFF
--- a/recipes/git-identity
+++ b/recipes/git-identity
@@ -1,0 +1,1 @@
+(git-identity :fetcher github :repo "akirak/git-identity.el")


purcell:
Thanks! My only note is that `magit-commit` is already autoloaded, and you shouldn't be defining autoloads for other packages here, so you should remove that line.

Otherwise looks good, so I'll go ahead and merge now. :-)

PR #6118, opened 2019-04-07
Add recipe for `ovpn-mode'

anticomputer:

An openvpn configuration and connection management mode.

https://github.com/anticomputer/ovpn-mode

Maintainer.

N/A

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/ovpn-mode
+++ b/recipes/ovpn-mode
@@ -1,0 +1,1 @@
+(ovpn-mode :repo "anticomputer/ovpn-mode" :fetcher github)


purcell:
Thanks! Some notes:
- When you ran it, `package-lint` should have told you that you need a package dependency for `cl-lib`.
- [This line](https://github.com/anticomputer/ovpn-mode/blob/d39886d619e479acff3b3343ff3a678b582cd0e8/ovpn-mode.el#L169) is redundant, since `define-derived-mode` expands to code which calls `use-local-map` for you.
- Prefer to `let`-bind `inhibit-read-only` in cases like [this](https://github.com/anticomputer/ovpn-mode/blob/d39886d619e479acff3b3343ff3a678b582cd0e8/ovpn-mode.el#L539-L552).
- Nice to see you're carefully using `shell-quote-argument` where appropriate. [This](https://github.com/anticomputer/ovpn-mode/blob/d39886d619e479acff3b3343ff3a678b582cd0e8/ovpn-mode.el#L939) and [this](https://github.com/anticomputer/ovpn-mode/blob/d39886d619e479acff3b3343ff3a678b582cd0e8/ovpn-mode.el#L992) might well not be safe, though.
- You might consider failing fast with code like `(unless cond (error "Could not start openvpn ..."))` to make the [long hanging `if` + `progn` code](https://github.com/anticomputer/ovpn-mode/blob/d39886d619e479acff3b3343ff3a678b582cd0e8/ovpn-mode.el#L724-L796) more comprehensible. The `message` calls there (and elsewhere) should probably be errors in any case.
- [Here](https://github.com/anticomputer/ovpn-mode/blob/d39886d619e479acff3b3343ff3a678b582cd0e8/ovpn-mode.el#L1070-L1073) you're modifying the global `kill-buffer-hook`, then checking every time a buffer is killed to see if it is the open vpn buffer. Instead, in openvpn mode, just modify the local version of `kill-buffer-hook` : see `add-hook`'s "local" argument.

Hope that helps!

anticomputer:
Thanks for the feedback Steve, I'll implement the suggested improvements as well as address the missing package require. Regarding the privileged shell call out safety, I did consider the potential for "shoot in foot" user behavior on that privileged shell interface there as noted in the documentation. I think I will explicitly remove interactive use of that functionality and limit it to the hardcoded xterm and chrome spawners to limit the potential for pilot error. Cheers!

anticomputer:
https://github.com/anticomputer/ovpn-mode/commit/ffd46f97a6b6e6b72f3f795f09807ae3b08a2d18 addresses your feedback. I agree that removing the copious if/progn constructs makes for much more readable code.

purcell:
Awesome, merged now. Welcome to MELPA! :tada:

PR #6119, opened 2019-04-10
Add `ewal` package.

jjzmajic:

`ewal` is a pywal-based theme generator for Emacs. It extracts
special and regular `pywal` colors, and adds extra shades that are
intelligently suppressed in a TTY setting. It also provides functions that
expose those colors for granular theming configurations, and functions that
automatically customize different themes with your preferred colorscheme. It
already provides such a function for `spacemacs-theme`, and more will be added
over time.

https://gitlab.com/jjzmajic/ewal

Maintainer.

The occasional introspective episode.

### Checklist

Please confirm with `x`:

- [x] `M-x checkdoc` is happy with my docstrings (It's not due to one long URL in a docstring.)

### DIFF
--- a/recipes/ewal
+++ b/recipes/ewal
@@ -1,3 +1,4 @@
 (ewal
  :fetcher gitlab
- :repo "jjzmajic/ewal")
+ :repo "jjzmajic/ewal"
+ :files ("ewal.el"))
--- a/recipes/ewal
+++ b/recipes/ewal
@@ -1,0 +1,3 @@
+(ewal
+ :fetcher gitlab
+ :repo "jjzmajic/ewal")
--- a/recipes/ewal-evil-cursors
+++ b/recipes/ewal-evil-cursors
@@ -1,0 +1,4 @@
+(ewal-evil-cursors
+ :fetcher gitlab
+ :repo "jjzmajic/ewal"
+ :files ("evil-cursors/*.el"))
--- a/recipes/ewal-spacemacs-theme
+++ b/recipes/ewal-spacemacs-theme
@@ -1,0 +1,4 @@
+(ewal-spacemacs-theme
+ :fetcher gitlab
+ :repo "jjzmajic/ewal"
+ :files ("spacemacs-theme/*.el"))


purcell:
Thanks! Is there any overlap here with [theme-magic](https://github.com/melpa/melpa/pull/6085), by @jcaw? If there's a way to share a common package for interfacing with `pywal` then I would suggest you join forces.

Re. the code itself:
- You shouldn't [assume that](https://gitlab.com/jjzmajic/ewal/blob/52423994f78e8a379a3193639012ca69b142ff40/ewal.el#L85) just because `display-graphic-p` returns `nil` when this file is loaded, that there is no graphical frame. Emacs may have been started in a terminal or as a daemon, and then later open a graphical frame. Better to only look at `display-graphic-p` close to the time you're doing something.
- [Here](https://gitlab.com/jjzmajic/ewal/blob/52423994f78e8a379a3193639012ca69b142ff40/ewal.el#L139-152) you could just use `ignore-errors`, but the best idea is to write the code so it doesn't fail, which looks like it should be possible here.
- Wouldn't it be better to use the `color` library to [blend colours](https://gitlab.com/jjzmajic/ewal/blob/52423994f78e8a379a3193639012ca69b142ff40/ewal.el#L161-181)? It looks like you're importing that library and not using it.
- I wouldn't suggest that this package should grow in the future to include theme generators for solarized etc. Feels like that would get clunky, and that you should plan to have such things as separate packages (whether or not they're written by you). I could imagine you exposing an `ewal-theme-define` macro for such purposes, which would define a named theme using a user-provided `body` form, with the various `pywal` colours available as bound variables. The current code shouldn't really have anything specific to spacemacs or evil in it, IMO.

jcaw:
Hey, I'm the author of [theme-magic](https://github.com/jcaw/theme-magic). 

I saw your package on Reddit the other day, it looks good. It seems to serve the opposite function to `theme-magic`. Theme magic is Emacs -> Linux (via pywal), this goes Linux (via Pywal) -> Emacs. 

I was actually intending to add the ability to generate a theme from Pywal, using [doom-theme-generator](https://gitlab.com/valtrok/doom-theme-generator). Doom themes look nice, and I can leverage the existing work. If `ewal` does a good job, that's perfect. I'll have to check it out.

I don't see any theoretical limitations to combining the packages, but I'll have to take a look at how ewal functions first. 

>ewal doesn’t only extract special and regular colors from pywal, it adds extra shades to them which it can intelligently suppress in a TTY context. 

👍

jjzmajic:
@purcell
- point taken, will fix it asap
- the idea is to return nil so people can avoid customizing their theme when `ewal` doesn't load if the JSON file doesn't exist for example. I expanded it to all errors because they should skip customisation on any error, but I would welcome any suggestions on refinements. Perhaps returning an appropriate error message depending on type?
- the color library depends on the current frame for almost all of it's functions so it doesn't extend the colours correctly. That's why I'm only using a few functions from it that are frame independent.
- point taken, I'll see what I can do

jjzmajic:
@jcaw, that sounds awesome! Pm me on Gitter to coordinate?

jcaw:
@jjzmajic will do

(Also: just a heads up but I edited the post quite a bit after posting.)

jjzmajic:
@purcell thanks for the head up about theme-magic, @jcaw and I are on it (although still in the discussion stage). I did however implement the changes you suggested, at least to get feedback on them.

The only things that I didn't change exactly as suggested are the error handling (so customization can be avoided on any error), and the color processing functions that need to be there so that color computations are frame agnostic. Do let me know if there is anything else I should do though.

purcell:
Cool, have merged this for now then - thanks!

purcell:
(And welcome to MELPA! :tada:)

jjzmajic:
@purcell, quick question (sorry, still new to melpa):
`ewal-spacemacs-theme` does not extract the proper metadata into a `*-pkg.el` file. I'm sure this is my mistake and has something to do with it being a multi-theme package, but I tried to the best of my ability to follow the docs and emulate the structure of a similar package (`doom-themes`). The metadata is contained in one of the files (`ewal-spacemac.el`). Could it be that the fact that it does not share the name of the package is causing the issue?

Should I rename the package to `ewal-spacemacs-themes` and call the file `ewal-spacemacs-themes.el` (so it does not look like a theme?

aaronjeline:
Just installed through MELPA, 
    getting Required feature ‘term/tty-colors’ was not `provided`

purcell:
> Should I rename the package to `ewal-spacemacs-themes` and call the file `ewal-spacemacs-themes.el` (so it does not look like a theme? Or would renaming the file to `ewal-spacemacs-theme.el` be preferable even if there is some confusion? I'll do the latter for the time being to make the package functional, but if you advise me to change it along with the recipe i'll do so.

Yes, you'd probably want to rename the package to `ewal-spacemacs-themes`, and make sure that the `ewal-spacemacs.el` is called `ewal-spacemacs-themes.el`, to match the package name. Then the metadata it contains will be correct. Just send a PR to rename and adjust the recipes when you're ready and it should be a no-brainer to merge. :-)

jjzmajic:
@aaronjeline that's peculiar, it's part of core Emacs. Are you compiling yours from source by chance? Which version of Emacs are you using?

PS: `ewal-spacemacs-themes` wont work for a bit until the metadata fix discussed above is merged. As soon as that's done, I'll help debug this as much as I can. In the meantime, please let me know about the questions I had above.

purcell:
> Are you compiling yours from source by chance? Which version of Emacs are you using?

Screenshot 2019-05-14 at 14 44 07

It's quite possible the Emacs isn't new enough based on the `git blame` above.

jjzmajic:
@purcell, should I bump the Emacs version to 26.1 then, or remove the require statement? Because the functions I need are available in way older versions of Emacs.

jjzmajic:
@purcell, it's been a while since I removed `ewal-spacemacs-theme` (no **s**), but it still seems to be downloadable. Is there anything I can do to minimize confusion, or should I just wait for a few more builds? PS: @aaronjeline the package should work for you after the next build.

PR #6120, opened 2019-04-10
Add recipe for semaphore

bendlas:

A semaphore based on two monotonic counters and a condition variable.

The condition variable enables it to interact with threads, while the two counters enable it to support a lifetime-count of acquisitions.

https://github.com/webnf/semaphore.el

I'm the creator and maintainer

I expect the API to be totally stable, grow-only. It has been used in a real-world project.

Thanks to the checklist and package-lint, I last-minute renamed `make-semaphore` to `semaphore-create`. Will be happy about further feedback before I tag the first version, since this package assumes a pretty important provision in elisp land.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/semaphore
+++ b/recipes/semaphore
@@ -1,1 +1,2 @@
-(semaphore :repo "webnf/semaphore.el" :fetcher github)
+(semaphore :repo "webnf/semaphore.el" :fetcher github
+           :files ("semaphore.el"))
--- a/recipes/semaphore
+++ b/recipes/semaphore
@@ -1,0 +1,1 @@
+(semaphore :repo "webnf/semaphore.el" :fetcher github)
--- a/recipes/semaphore-promise
+++ b/recipes/semaphore-promise
@@ -1,0 +1,2 @@
+(semaphore-promise :repo "webnf/semaphore.el" :fetcher github
+                   :files ("semaphore-promise.el"))


bendlas:
cc https://github.com/chuntaro/emacs-promise/issues/7

purcell:
Thanks! I suspect this should be a pair of recipes, right? One recipe would specify `:files ("semaphore.el")` and the other recipe would include only `semaphore-promise.el`: `semaphore-promise.el` would then declare package dependencies on `sempahore` and `promise`. Alternatively (and perhaps less ideally) you'd need to declare a package dependency on `promise` in `semaphore.el`, and then leave this as one recipe.

Code itself looks good, but please remove the redundant "semaphore: " string at the beginning of the package description [here](https://github.com/webnf/semaphore.el/blob/d3816842c5ec7ae4634425ccec1f87cd687f53c4/semaphore.el#L1) to keep things tidy.

bendlas:
> I suspect this should be a pair of recipes, right?

The intention was for `promise` to be a kind of optional dependency, as proposed [here](https://github.com/cask/cask/issues/334), but I can also make two recipes. Those can be hosted in the same repo, right?

purcell:
> The intention was for `promise` to be a kind of optional dependency, as proposed [here](https://github.com/cask/cask/issues/334), but I can also make two recipes.

Yes, I think that's ideal. Much better to avoid optional dependencies!

> Those can be hosted in the same repo, right?

Yes, and you can use `:files` in the recipe to specify which file goes in each recipe.

purcell:
Ping! :-)

bendlas:
Is it alright like [this](https://github.com/webnf/semaphore.el/blob/3912621594d3efb4ea4447e14927576db2a534ca/semaphore-promise.el#L10)?

riscy:
That looks right to me.  Let me bump this up the queue --

[/elisp/semaphore-promise.el]
-------------------------------------------------------------------------------

<kbd>M-x checkdoc</kbd>:
```
Warning (emacs): 
semaphore-promise.el:40: Second line should not have indentation
Warning (emacs): 
semaphore-promise.el:40: First line is not a complete sentence
Warning (emacs): 
semaphore-promise.el:40: Argument `semaphore' should appear (as SEMAPHORE) in the doc string
```

```
1 issue found:

```

```
semaphore-promise.el:55:7:Warning: `semaphore' is an obsolete variable (as of
    25.1); use 'semaphore instead
semaphore-promise.el:59:30:Warning: `semaphore' is an obsolete variable (as of
    25.1); use 'semaphore instead
semaphore-promise.el:62:29:Warning: `semaphore' is an obsolete variable (as of
    25.1); use 'semaphore instead
```
(These are interesting.  I am able to "fix" these by changing the variable's name, but I haven't dug any deeper than that.)

<kbd>M-x check-declare-file</kbd> (not on the MELPA checklist):
No issues! 🎉

Minor:
- 9: Prefer `https` over `http` (if possible)

-------------------------------------------------------------------------------
[/elisp/semaphore.el]

<kbd>M-x checkdoc</kbd>:
```
Warning (emacs): 
semaphore.el:53: First line is not a complete sentence
```

No issues! 🎊

No byte-compilation issues! 😎

<kbd>M-x check-declare-file</kbd> (not on the MELPA checklist):
No issues! 🎉

Minor:
- 9: Prefer `https` over `http` (if possible)

purcell:
Thanks, merged now - sorry about the slow turnaround.

PR #6123, opened 2019-04-12
Add 'mood-one-theme' package

jessiehildebrandt:

mood-one-theme is a dark color scheme inspired by the appearance of the Doom One theme from the doom-themes package, but with zero dependencies and a minimal performance footprint.

https://gitlab.com/jessieh/mood-one-theme

Maintainer

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/mood-one-theme
+++ b/recipes/mood-one-theme
@@ -1,0 +1,1 @@
+(mood-one-theme :fetcher gitlab :repo "jessieh/mood-one-theme")


purcell:
Thanks! This looks good, but the fringe bitmap stuff is a bit hokey: if you're providing a function to end users which simply loads up a library, that's a bad sign! All the `eval-after-load`s there don't belong in library code. 

[Here's](https://gist.github.com/purcell/ec8daf96e3634b2b3974e428f68e630e) an alternate (untested) version which I think is more what you should do.

jessiehildebrandt:
> [Here's](https://gist.github.com/purcell/ec8daf96e3634b2b3974e428f68e630e) an alternate (untested) version which I think is more what you should do.

Thank you for the feedback! I was a little uncertain about my approach to that; your version of the fringe bitmap code more appropriately captures what I was trying to do there.

I've integrated your version of that code into the package-- it compiles cleanly and functions as expected when tested with my local Emacs configuration. It should be up in the GitLab repository now.

purcell:
Great! Merged now.

PR #6124, opened 2019-04-12
Add a recipe for gcmh

AndreaCorallo:

Enforce a sneaky Garbage Collection strategy to minimize GC interference with user
 activity.           
                                                     
https://gitlab.com/koral/gcmh

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/gcmh
+++ b/recipes/gcmh
@@ -1,0 +1,1 @@
+(gcmh :repo "koral/gcmh" :fetcher gitlab)


purcell:
Thanks! Looks useful. Notes:

- Typo: "Colelction"
- You should `;;;###autoload` the minor mode.
- The `cond` [here](https://gitlab.com/koral/gcmh/blob/c1bc27aead05fda40bfe5d2959d29ccbcb7d92fd/gcmh.el#L85-107) would be a bit more obvious as an `if`, or even a `pcase`.

In any case, I can go ahead and merge this. :-)

AndreaCorallo:
Thanks!

PR #6125, opened 2019-04-14
add seml-mode recipe

conao3:
`seml-mode` provide major mode for editing SEML (S-Expression Markup Language) files.

`libxml-parse-html-region` returns sexp when feed HTML string but decode function (sexp -> HTML) is missing.
So I implemented it and also implemented the corresponding API.

https://github.com/conao3/seml-mode.el

I'm maintainer.

None

### Checklist

Please confirm with `x`:

---
## package lint warnings
```
1 issue found:

133:1: error: "with-seml-elisp" doesn't start with package's prefix "seml".
```
But macro provide evaluation environment start `with` prefix.
## compiler warnings
```
$ emacs-25.3 -Q --batch -L . -f batch-byte-compile seml-mode.el

In seml-encode-html-from-region:
seml-mode.el:316:34:Warning: function ‘mapcan’ from cl package called at
    runtime
seml-mode.el:326:66:Warning: function ‘mapcan’ from cl package called at
    runtime
seml-mode.el:335:22:Warning: function ‘mapcan’ from cl package called at
    runtime
seml-mode.el:336:48:Warning: function ‘mapcan’ from cl package called at
    runtime

In seml-decode-seml-from-sexp:
seml-mode.el:431:16:Warning: function ‘mapcan’ from cl package called at
    runtime

$ emacs-26.1 -Q --batch -L . -f batch-byte-compile seml-mode.el
(no output)
```
Emacs 25's byte-compiler warn corresponding `mapcan`, but Emacs-26's byte-compiler not warn.
How should I handle this warning?

### DIFF
--- a/recipes/seml-mode
+++ b/recipes/seml-mode
@@ -1,0 +1,1 @@
+(seml-mode :fetcher github :repo "conao3/seml-mode.el")


purcell:
- I think you can just use `rx-to-string` instead of `eval` [here](https://github.com/conao3/seml-mode.el/blob/f62f7a9f5e18ef8f370444a1fce710770227b0f5/seml-mode.el#L114)
- There is a built-in function `consp` which does [this](https://github.com/conao3/seml-mode.el/blob/f62f7a9f5e18ef8f370444a1fce710770227b0f5/seml-mode.el#L124)
- [This global var](https://github.com/conao3/seml-mode.el/blob/f62f7a9f5e18ef8f370444a1fce710770227b0f5/seml-mode.el#L143) is missing the package prefix.
- Try to avoid `let`-binding nil variables and then setting them with `setq`, like [here](https://github.com/conao3/seml-mode.el/blob/f62f7a9f5e18ef8f370444a1fce710770227b0f5/seml-mode.el#L310-L311) -- just assign them in the `let` in a single step.
- This [empty `let`](https://github.com/conao3/seml-mode.el/blob/f62f7a9f5e18ef8f370444a1fce710770227b0f5/seml-mode.el#L59) can be collapsed.
- It's generally not advisable to keep a [version variable](https://github.com/conao3/seml-mode.el/blob/f62f7a9f5e18ef8f370444a1fce710770227b0f5/seml-mode.el#L50) in your code, because it will usually be out-of-sync with the package version from which the code was installed. Instead, omit that variable and tag your stable releases so that the MELPA Stable packages will have the desired package numbers.

> Emacs 25's byte-compiler warn corresponding `mapcan`, but Emacs-26's byte-compiler not warn.
> How should I handle this warning?

Use `cl-mapcan`, from the `cl-lib` library.

> But macro provide evaluation environment start `with` prefix.

The macro is a no-op, so you could just delete it, but in this case the name is okay. Or you could just use the name `seml-elisp` or `seml-elisp-eval`.

Hope that helps!

conao3:
Thank you for your review! Thanks a lot!

I will quote your review with the number.
> 1. I think you can just use `rx-to-string` instead of `eval` [here](https://github.com/conao3/seml-mode.el/blob/f62f7a9f5e18ef8f370444a1fce710770227b0f5/seml-mode.el#L114)
> 2. There is a built-in function `consp` which does [this](https://github.com/conao3/seml-mode.el/blob/f62f7a9f5e18ef8f370444a1fce710770227b0f5/seml-mode.el#L124)
> 3. [This global var](https://github.com/conao3/seml-mode.el/blob/f62f7a9f5e18ef8f370444a1fce710770227b0f5/seml-mode.el#L143) is missing the package prefix.
> 4. Try to avoid `let`-binding nil variables and then setting them with `setq`, like [here](https://github.com/conao3/seml-mode.el/blob/f62f7a9f5e18ef8f370444a1fce710770227b0f5/seml-mode.el#L310-L311) -- just assign them in the `let` in a single step.
> 5. This [empty `let`](https://github.com/conao3/seml-mode.el/blob/f62f7a9f5e18ef8f370444a1fce710770227b0f5/seml-mode.el#L59) can be collapsed.
> 6. It's generally not advisable to keep a [version variable](https://github.com/conao3/seml-mode.el/blob/f62f7a9f5e18ef8f370444a1fce710770227b0f5/seml-mode.el#L50) in your code, because it will usually be out-of-sync with the package version from which the code was installed. Instead, omit that variable and tag your stable releases so that the MELPA Stable packages will have the desired package numbers.
> 
> > Emacs 25's byte-compiler warn corresponding `mapcan`, but Emacs-26's byte-compiler not warn.
> > How should I handle this warning?
> 
> 7. Use `cl-mapcan`, from the `cl-lib` library.
> 
> > But macro provide evaluation environment start `with` prefix.
> 
> 8. The macro is a no-op, so you could just delete it, but in this case the name is okay. Or you could just use the name `seml-elisp` or `seml-elisp-eval`.

1, 5, 6, 7 has fixed (see https://github.com/conao3/seml-mode.el/pull/108 )
Other points will be added in the next comments.

conao3:
## 2. is a point of the review about `seml-pairp`.

`seml-pairp` strictly return t for dot pair.
This property is not found in `consp`.

Because this package is made dependent on this difference, I can not replace `pairp` with `consp`.

```
(let ((list '(a b))
      (pair '(a . b)))
  `(,(seml-pairp list) ,(seml-pairp pair) ,(consp list) ,(consp pair)))
;; => (nil t t t)
```

## 3. is a point of the review about `calculate-lisp-indent-last-sexp`.
My code is for only suppressing compiler errors, and the actual declaration is done in  https://github.com/emacs-mirror/emacs/blob/emacs-26/lisp/emacs-lisp/lisp-mode.el#L898
But this also seems to have not been made...

It is clear that replacing `calculate-lisp-indent-last-sexp` with `seml-calculate-lisp-indent-last-sexp` `(void-variable seml-calculate-lisp-indent-last-sexp)` results in an error.

If `(defvar seml-calculate-lisp-indent-last-sexp nil)` is used as an actual variable declaration instead of pure substitution, `(wrong-type-argument integer-or-marker-p nil)` error occurs and declared as 0 Then `(args-out-of-range # <killed buffer> 0 2)` an error occurs.

I may have to think of another way to deal with this problem. For my purpose, I just want to make the HTML keyword ( https://github.com/conao3/seml-mode.el/blob/522b61579653cc1ada75b7b16f5de2ab4aa612c6/seml-mode.el#L87-L106 ) indent the same as `(declare (indent 2))`.

## 4. is a point of the review about seemingly strange `let`.
This is a dirty technique that tries to recurse with anonymous functions under the influence of `lexical-binding`.
In fact, this is the result follows. I also don't want to use such dirty hacks, so please let me know if there is a better idea about anonymous function recursion.

```
(setq lexical-binding nil)
;; => nil

(let ((fn (lambda (lst)
            (if (null lst)
                0
              (+ (car lst) (funcall fn (cdr lst)))))))
  (funcall fn '(1 2 3)))
;; => 6

;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;

(setq lexical-binding t)
;; => t

(let ((fn (lambda (lst)
            (if (null lst)
                0
              (+ (car lst) (funcall fn (cdr lst)))))))
  (funcall fn '(1 2 3)))
;; => Debugger entered--Lisp error: (void-variable fn)
;;      (funcall fn (cdr lst))
;;      (+ (car lst) (funcall fn (cdr lst)))
;;      (if (null lst) 0 (+ (car lst) (funcall fn (cdr lst))))
;;      (closure (t) (lst) (if (null lst) 0 (+ (car lst) (funcall fn (cdr lst)))))((1 2 3))
;;      funcall((closure (t) (lst) (if (null lst) 0 (+ (car lst) (funcall fn (cdr lst))))) (1 2 3))
;;      (let ((fn (function (lambda (lst) (if (null lst) 0 (+ (car lst) (funcall fn (cdr lst)))))))) (funcall fn '(1 2 3)))
;;      (progn (let ((fn (function (lambda (lst) (if (null lst) 0 (+ (car lst) (funcall fn (cdr lst)))))))) (funcall fn '(1 2 3))))
;;      eval((progn (let ((fn (function (lambda (lst) (if (null lst) 0 (+ (car lst) (funcall fn (cdr lst)))))))) (funcall fn '(1 2 3)))) t)
;;      elisp--eval-last-sexp(t)
;;      #f(compiled-function (eval-last-sexp-arg-internal) "Evaluate sexp before point; print value in the echo area.\nInteractively, ...
;;      apply(#f(compiled-function (eval-last-sexp-arg-internal) "Evaluate sexp before point; print value in the echo area.\nInteractively, ...
;;      eval-last-sexp(t)
;;      eval-print-last-sexp(nil)
;;      funcall-interactively(eval-print-last-sexp nil)
;;      call-interactively(eval-print-last-sexp nil nil)
;;      command-execute(eval-print-last-sexp)

(let (fn)
  (setq fn (lambda (lst)
            (if (null lst)
                0
              (+ (car lst) (funcall fn (cdr lst))))))
  (funcall fn '(1 2 3)))
;; => 6
```

## 7. is a point of the review about macro `with-seml-elisp`
`with-seml-elisp` appears to do nothing at first glance, but it is a macro that executes the S-expression given in the implicit `progn` and returns "always" `nil`.
This is the syntax sugar necessary to execute Elisp freely in SEML, and has a role that only it can do.
Anyway, I found that there was no problem with naming, so I did not change the name.

##
That is all. Please tell me if I need to work more.

purcell:
Okay, thanks for the explanations, and sorry for the long delay - I'm merging this now. Welcome to MELPA! :tada:

conao3:
OK, Thanks!

PR #6126, opened 2019-04-15
Add recipe for leetcode.el

kaiwk:

A LeetCode client for Emacs. Now it can check problems list, try the customized test cases and submit the solution.

https://github.com/kaiwk/leetcode.el

maintainer&contributor

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/leetcode
+++ b/recipes/leetcode
@@ -1,0 +1,1 @@
+(leetcode :fetcher github :repo "kaiwk/leetcode.el")


kaiwk:
Actually, after use `package-lint`, I got this:

>743:1: error: "turn-on-leetcode-loading-mode" doesn't start with package's prefix "leetcode".

purcell:
Thanks! Some notes:

- Don't define faces with hard-coded colors when possible, and in particular never define faces that have only one of `:background` and `:foreground`. The best option is generally to simpy `:inherit` a built-in face so that the default colours are acceptable in every theme, then they can still be overridden by users.
- These [repeated `setq` of the same var](https://github.com/kaiwk/leetcode.el/blob/fc93cefc5a096146b47d01be2fe2de677ce85fd5/leetcode.el#L196-L200) are bogus: this could all be done in one expression. `(setq leetcode--user (list :username ... :solved ...))`
- Check out the `let-alist` library, which will make all the data unpacking code much more tidy.
- You can do [all this](https://github.com/kaiwk/leetcode.el/blob/fc93cefc5a096146b47d01be2fe2de677ce85fd5/leetcode.el#L234-L236) with a single replacement of the regexp `[(),]`.
- Presumably [this code](https://github.com/kaiwk/leetcode.el/blob/fc93cefc5a096146b47d01be2fe2de677ce85fd5/leetcode.el#L528-L559) can break if the user opens any additional windows, even temporarily, or reorganises the windows. What you should do is just create and display buffers as necessary, and then have a function which reorganises them into the desired layout. Then the function can be called whenever it is convenient for the user.
- It's not at all clear why you've got two `loading` minor modes. It should be sufficient to define a single global minor mode, by adding `:global t` to the [local minor mode definition](https://github.com/kaiwk/leetcode.el/blob/fc93cefc5a096146b47d01be2fe2de677ce85fd5/leetcode.el#L783-L789)

> error: "turn-on-leetcode-loading-mode" doesn't start with package's prefix "leetcode".

Yes, `turn-on-` functions are redundant. You can just call the mode function itself with no args.

Hope that helps! :-)

purcell:
I had a look and it seems like most of this is fixed now, so I'll go ahead and merge. A couple of my points still stand, e.g. the hard-coded colors: it would be much better to `:inherit` the faces `success`, `warning` and `error` here, in my view. Welcome to MELPA! :tada:

PR #6127, opened 2019-04-16
Add recipe shift-select

jcs090218:

Shift select text like how other text editor does.

(Basically the shift select mode that does not suck)

https://github.com/jcs090218/shift-select

The maintainer.

None needed.

### Checklist

Please confirm with `x`:


jcs090218:
I will close this down for now cuz I think this could be my configuration issue. :P

PR #6129, opened 2019-04-18
Recipes: add share

o-nly:
Quality: wip; Description: Share is an Emacs interface for sharing media

The package has functions for abstracting over functions you would use to share media. For example, maybe a pastebin command, maybe a command to deploy a file on your nginx server.

https://github.com/o-nly/emacs-share

Maintainer

Very much a work in progress, but already has a few useful functions.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/share
+++ b/recipes/share
@@ -1,0 +1,4 @@
+(share :fetcher github
+             :repo "o-nly/emacs-share"
+             :branch "master"
+             )
--- a/recipes/share
+++ b/recipes/share
@@ -1,4 +1,1 @@
-(share :fetcher github
-             :repo "o-nly/emacs-share"
-             :branch "master"
-             )
+(share :fetcher github :repo "o-nly/emacs-share")


purcell:
Thanks! Some notes:
- Please remove `:branch` from the recipe. It is redundant.
- Don't put trailing `)` on their own lines; use idiomatic lisp formatting to make the code accessible to contributors. Similarly, leave an empty line between top-level definitions.
- Keep all `require` forms at the top of the code.
- Add the missing package dependency declarations for `s` and `f`
- [This](https://github.com/o-nly/emacs-share/blob/master/share.el#L326) is very mysterious

I guess the issue I'm having overall is that despite the long Commentary, it's really not clear how one would use this, and why it would be beneficial. I feel like this might be a general "framework to rule them all" without a specific killer application, and the rather all-encompassing name underlines that. For the specific cases you've implemented, a few functions that use `kill-ring` would be sufficient and more reusable.

o-nly:
> Thanks! Some notes:
> 
> * Please remove `:branch` from the recipe. It is redundant.
> * Don't put trailing `)` on their own lines; use idiomatic lisp >formatting to make the code accessible to contributors. Similarly, >leave an empty line between top-level definitions.

Removed :branch. Package recipe fits in one line.

> * Keep all `require` forms at the top of the code.
> * Add the missing package dependency declarations for `s` and `f`

Removed (require 's) as it's not needed. Moved (require 'f) to top.

> * [This](https://github.com/o-nly/emacs-share/blob/master/share.el#L326) is very mysterious
> 
> I guess the issue I'm having overall is that despite the long >Commentary, it's really not clear how one would use this, and why it >would be beneficial. I feel like this might be a general "framework to >rule them all" without a specific killer application, and the rather >all-encompassing name underlines that. For the specific cases you've >implemented, a few functions that use `kill-ring` would be sufficient >and more reusable.

Right, Thanks.

o-nly:
I should close this; since you are waiting for upstream?

PR #6131, opened 2019-04-19
Add magit-libgit

tarsius:

```
;; This package teaches Magit to use functions provided by the
;; `libegit2' module to perform certain tasks.  That module used the
;; Libgit2 implementation of the Git core methods and is implemented
;; in the `libgit' package.

;; The hope is that using a C module instead of calling out to `git'
;; all the time increases performance; especially on Windows where
;; starting a process is unreasonably slow.

;; This package is still experimental and not many functions have been
;; reimplemented to use `libgit' yet.
```

https://github.com/magit/magit

Author

### Checklist

Please confirm with `x`:

  It's not happy about version strings, but I know what I am doing.

### DIFF
--- a/recipes/magit
+++ b/recipes/magit
@@ -5,4 +5,5 @@
                "lisp/git-rebase.el"
                "Documentation/magit.texi"
                "Documentation/AUTHORS.md"
-               "LICENSE"))
+               "LICENSE"
+	       (:exclude "lisp/magit-libgit.el")))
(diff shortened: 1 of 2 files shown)


tarsius:
I was not able to prevent this on the first install:

```
In end of data:
magit-libgit.el:76:1:Warning: the following functions are not known to be defined:
  libgit-repository-open, libgit-repository-bare-p
```

Also the user will be asked:

```
libgit must be built, do so now? (y or n)
```

tarsius:
I am merging this *begore* pushing the respective changes to `magit/magit` because otherwise `magit` would fail to build due to `libgit` not being installed.

PR #6134, opened 2019-04-20
Add recipe for aio

skeeto:

Provides async/await much like Python's `asyncio`. I wrote an in-depth article about it: [An Async / Await Library for Emacs Lisp](https://nullprogram.com/blog/2019/03/10/). I don't expect to make significant changes to the API at this point, so I'm submitting it as a package. I gave it about a month: skeeto/emacs-aio#1.

I used `:files` in my recipe since `aio-contrib.el` shouldn't be included in the package. Its purpose is to provide examples of how `aio` might be used with other packages and optional Emacs features. It shouldn't be used directly and won't compile cleanly unless all of these extra, mostly unrelated packages are actually installed. `aio` will always be a single-file package, so I explicitly listed all of the relevant files.

https://github.com/skeeto/emacs-aio

**None needed**

### Checklist

Please confirm with `x`:

`M-x checkdoc` has some minor complaints. I fixed one of them, and I'm using personal judgement for the rest.

### DIFF
--- a/recipes/aio
+++ b/recipes/aio
@@ -1,0 +1,3 @@
+(aio :fetcher github
+     :repo "skeeto/emacs-aio"
+     :files ("aio.el" "README.md" "UNLICENSE"))


purcell:
Great, thanks!

PR #6140, opened 2019-04-24
Add json-reformatter-jq package

wbolster:

This Emacs package provides a [reformatter](https://github.com/purcell/reformatter.el) for [json](https://www.json.org/) and [jsonlines](http://jsonlines.org/) files using the `jq` utility.

https://github.com/wbolster/emacs-json-reformatter-jq

Author/maintainer.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/json-reformatter-jq
+++ b/recipes/json-reformatter-jq
@@ -1,0 +1,3 @@
+(json-reformatter-jq
+ :fetcher github
+ :repo "wbolster/emacs-json-reformatter-jq")


purcell:
Great - thanks!

wbolster:
cool, that was quick.

i was slightly worried someone would complain about using also `jsonlines-reformatter-...` command names for a package that is mostly `json-reformatter-...` commands/variables. :wink:  (i argue this is totally fine and alternatives are worse.)

purcell:
Ah, I missed that. You're right that this is a bit questionable. Better would have been to call the whole thing `jq-format`, then you could have had `jq-format-json-*` and `js-format-jsonlines-*`.

purcell:
That's a shorter and nicer package name too.

purcell:
(Easy to rename this if you're willing to make that change.)

wbolster:
i like your suggestions but I don't like not having "json" in the package name for discoverability. ideas?

purcell:
Yep, just put it in the package description: `;;; jq-format.el --- Reformat JSON and JSONLines using jq`.

purcell:
The MELPA search includes package descriptions, btw: https://melpa.org/#/?q=json

wbolster:
ok, makes sense. all done upstream: https://github.com/wbolster/emacs-jq-format/issues/2

corresponding pr for melpa: https://github.com/melpa/melpa/pull/6142

PR #6142, opened 2019-04-26
rename json-reformatter-jq to jq-format

wbolster:
the upstream package changed its name:
https://github.com/wbolster/emacs-jq-format/issues/2

see https://github.com/melpa/melpa/pull/6140 for original discussion.

(backwards compatibility is not a concern; this package is only a week old.)


purcell:
Super, thanks!

PR #6143, opened 2019-04-28
add gvariant.el recipe

wbolster:

This Emacs package provides helpers for GVariant strings.

https://github.com/wbolster/emacs-gvariant

Author.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/gvariant
+++ b/recipes/gvariant
@@ -1,0 +1,3 @@
+(gvariant
+ :fetcher github
+ :repo "wbolster/emacs-gvariant")


purcell:
Thanks! Mostly looks great, so merging now. My one note is that you should use regexp-opt [here](https://github.com/wbolster/emacs-gvariant/blob/bb7c38b55e80344ee7dce70d0f93b813810fa6ac/gvariant.el#L39). :-)

wbolster:
thanks, will definitely look into your advice

PR #6144, opened 2019-04-28
add gsettings.el

wbolster:

This package provides helpers for Gnome GSettings.

https://github.com/wbolster/emacs-gsettings

author

### Checklist

Please confirm with `x`:


wbolster:
note that this package depends on my recent `gvariant.el`, which has a separate pull request at #6143 .

purcell:
Thanks!

PR #6145, opened 2019-04-28
Add recipe for sly-asdf

mmgeorge:

`sly-asdf` is a contrib for [`sly`](https://github.com/joaotavora/sly) (an editing mode for Common Lisp) that provides support for working with ASDF systems. It is a port of the original [`slime-asdf`](https://github.com/slime/slime/blob/master/contrib/slime-asdf.el) contrib with changes to support `package-inferred-system` (released with ASDF 3.1). 

https://github.com/mmgeorge/sly-asdf

I ported the library from SLIME and maintain it. 

None needed, though I do rely on a couple of internal `slynk` functions from the main `sly` project that I would like to get exposed. 

### Checklist

### DIFF
--- a/recipes/sly-asdf
+++ b/recipes/sly-asdf
@@ -1,0 +1,5 @@
+(sly-asdf :fetcher github
+          :repo "mmgeorge/sly-asdf"
+          :files (:defaults
+                  "*.lisp"
+                  "*.asd"))


purcell:
Nice, thanks for this! It looks in good shape and I'm going to merge it, but the one note I have is that [this line](https://github.com/mmgeorge/sly-asdf/blob/c0153d6017359be411878793b91257d8819c48a4/sly-asdf.el#L287-L288) might well be inappropriate: emacs lisp packages should not generally auto-enable themselves as a result of simply being installed (imagine the chaos if all the packages in the standard library did this!), so you should likely remove the autoload cookie comment and/or tell the user to put this line in their config instead. Hope that helps!

Welcome to MELPA! :tada:

PR #6147, opened 2019-04-29
Add whois recipe

lassik:

This package complements (does not replace) the standard whois
functionality of GNU Emacs. It provides:

* A `whois-mode` with font-lock highlighting to make whois responses
  easier to read.

* A `whois-shell` function to make a whois query using the system
  whois program instead of Emacs` own (often not up to date) whois
  client.

https://github.com/lassik/emacs-whois

Just wrote it

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/whois
+++ b/recipes/whois
@@ -1,0 +1,1 @@
+(whois :fetcher github :repo "lassik/emacs-whois")


purcell:
Looks good, merging now.

If the `*whois*` buffer is indeed non-interactive, as I suspect, can I suggest that you use `special-mode` instead of `fundamental-mode` as the parent? This way, standard keybindings like `q` will work automatically. (You'd want to `let`-bind `inhibit-read-only` while writing into the buffer, in this case, since special-mode also makes the buffer readonly iirc.)

lassik:
Thanks :)

> If the `*whois*` buffer is indeed non-interactive, as I suspect

It is indeed non-interactive.

> can I suggest that you use `special-mode` instead of `fundamental-mode` as the parent? This way, standard keybindings like `q` will work automatically.

This would be a good idea, but:

> (You'd want to `let`-bind `inhibit-read-only` while writing into the buffer, in this case, since special-mode also makes the buffer readonly iirc.)

Subprocesses from `start-process` write directly into that buffer without supervision from my Lisp code. This includes my `whois-shell` function as well as the `whois` function that comes with Emacs in `net-utils.el` which I have no control over. I could add a process filter to `whois-shell` that `let`-binds `inhibit-read-only` while inserting text. But I can't set a process filter for the Emacs `whois` command which uses the same buffer. (I could use a different buffer for `whois-shell` but it would be confusing since `whois-shell` is supposed to be a drop-in replacement for `whois` with no increase in cognitive load.)

PR #6148, opened 2019-04-29
Add recipe show-eol.

jcs090218:

Show end of line symbol in buffer.

This feature is from `NotePad++`.
https://unix.stackexchange.com/questions/82536/showing-type-of-newline-character-in-emacs

https://github.com/jcs090218/show-eol

The maintainer.

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/show-eol
+++ b/recipes/show-eol
@@ -1,0 +1,1 @@
+(show-eol :repo "jcs090218/show-eol" :fetcher github)


riscy:
This is obviously my fault, but I realized that if `whitespace-newline` has already been defined to use `inverse-video`, we'll disable it in the buffer when we toggle the mode off again.

In the spirit of the MELPA documentation -- 

> Please avoid defining a face that both :inherit another face and also override their attributes (e.g. by making them bold, underlined or inverse-video). The result could be really bad depending on user customizations. The best approach here is to simply :inherit the faces, and leave the user to customise the rest.

-- the best approach altogether might be to leave the `'whitespace-newline` face alone.  If users want inverse video in their newlines, they can customize the face accordingly.  Thoughts on that?

jcs090218:
@riscy Yeah, I agree with you. Maybe just remove the line here [L122](https://github.com/jcs090218/show-eol/blob/8f841e4b5272406f2f294072442bc52dcf8276c4/show-eol.el#L122)? I was hoping that user can customize the face their own, but I think the ultimate goal is to act like how `inverse-video` does. I'm satisfy with the result here. :D

riscy:
Yeah, I think that's the right move.

jcs090218:
@riscy Okay, I wasn't sure about your comment until I changed that line of code. I would like to have `inverse-video` to have it match to how `Notepad++` displays. I know this might offend melpa's spirit, would this be okay if I leave it like this?

riscy:
I think the right thing to do is to define a customizable face that inherits `whitespace-newline` or perhaps `whitespace-trailing`, and I think use `face-remap-set-base` and `face-remap-reset-base` to apply that customizable face when show-eol-mode is active.  Otherwise users won't have quite the same level of control over their faces (and this is definitely partly my fault.  I didn't think very far ahead when making my PR).

jcs090218:
Hmmm, after few thoughts I reckon rather having another customizable face I would just remove those two lines, [L115](https://github.com/jcs090218/show-eol/blob/8f841e4b5272406f2f294072442bc52dcf8276c4/show-eol.el#L115) and [L122](https://github.com/jcs090218/show-eol/blob/8f841e4b5272406f2f294072442bc52dcf8276c4/show-eol.el#L122). :D

riscy:
Yeah, I think that's probably the best move - let the user decide how the newlines should look (whether it's via show-eol or whitespace-mode)!

PR #6149, opened 2019-04-30
Add hl-block-mode

ideasman42:
This adds a minor mode for recursively highlighting blocks (nested scopes).

This highlights nested blocks of code, to help see the surrounding nested scopes (a feature in QtCreator)

alt text

https://github.com/ideasman42/emacs-hl-block-mode

Maintainer.
### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/hl-block-mode
+++ b/recipes/hl-block-mode
@@ -1,0 +1,3 @@
+(hl-block-mode
+ :repo "ideasman42/emacs-hl-block-mode"
+:fetcher github)


tarsius:
As part of the review I have send you a cleanup pull-request. https://github.com/ideasman42/emacs-hl-block-mode/pull/2

ideasman42:
> Great, looks like I'd use this. Just having another pass over the code:

Thanks for the feedback, I'm still quite new to writing elisp, so likely to make some beginner mistakes.
 
> Checkdoc has some small nags about the comment on `hl-block--overlay-refresh-from-timer`.
> 
> ```
> Buffer comments and tags:  Ok
> Documentation style:       1 Errors
> Message/Query text style:  Ok
> Unwanted Spaces:           Ok
> ```

Fixed (the error was introduced since the package was first submitted).

> (The following are only suggestions and not necessarily blockers for getting this merged)
> 
>     * If I were using this, my workflow would probably be to enable it globally.  Worth adding a call to `define-globalized-minor-mode hl-block-global-mode ...`?

Done.
 
>     * Should `defvar hl-block-overlay nil` be `defvar-local hl-block-overlay nil`?  That way every buffer gets to keep its overlays even if you've got `hl-block-mode` enabled in several of them.

Yes, done.

>     * `(defcustom hl-block-bracket 123` -- you can use `?{` instead of `123`

Done.
 
>     * Are you aware of the builtins `color-darken-name` and `color-lighten-name`?  Maybe you have a reason for not using these -- but if not, these and related functions in `colors.el` might help simplify your code.

Looked at `color.el` and I didn't see a simple function to replace whats there, although I'm not against moving to it as long as the function's not too heavy.

>       Minor:
> 
>     * `versiong` should be `version` (typo), and you could write `'syntax-ppss'` as `` `syntax-ppss' `` to tell Emacs to make a link out of the function in the online help.

Corrected.

riscy:
Just one package-lint issue (renaming `turn-on-hl-block-mode` to `hl-block-mode-turn-on` should do the trick).  I just noticed a flycheck nag you could fix by moving `defun hl-block--overlay-refresh-from-timer` beneath the definition of `hl-block-mode` and, finally, while you're at it, you could also enable lexical binding (advised) by adding `-*- lexical-binding: t; -*-` to the end of your top line.  Sorry for the laundry list, but I thought I'd get it all written down at once.

ideasman42:
Resolved `turn-on-hl-block-mode` warning.

I had `-*- lexical-binding: t; -*-` enabled however the warnings it gives don't have clear solutions AFAICS, and they seem harmless since the code is working. OTOH if there are ways to resolve issues like ```Unused lexical variable ‘i-tint’``` - I'd be interested to know.

riscy:
Yeah, it looks harmless -- in my opinion this is fine.  I've made a PR https://github.com/ideasman42/emacs-hl-block-mode/pull/3 that quiets a compiler warning and made it so that users can have a lighter for the mode, if they so choose.  A couple other things I was curious about:

- Why 300ms to wait for the update?  It feels a little slow.  I can't cite anything scientific, but I remember reading somewhere 200ms is a good threshold from a user interface perspective.
- What do you think about making nil the default value for `hl-block-bracket`?  Any reason not to?  The big benefit is that the mode will then work without modification in (for example) lisp and python mode for certain nested structures.

ideasman42:
> Yeah, it looks harmless -- in my opinion this is fine. I've made a PR [ideasman42/emacs-hl-block-mode#3](https://github.com/ideasman42/emacs-hl-block-mode/pull/3) that quiets a compiler warning and made it so that users can have a lighter for the mode, if they so choose. A couple other things I was curious about:
> 
>     * Why 300ms to wait for the update?  It feels a little slow.  I can't cite anything scientific, but I remember reading somewhere 200ms is a good threshold from a user interface perspective.

I wanted to avoid lag for large files, so successive actions wouldn't lag (scrolling the cursor for eg).

Lowering the delay to 200ms works fine too.

Note that it might be good enough to detect the first surrounding bracket location and compare them to the  last. When the values match don't re-create the overlays. *(while not fool-proof, I think it's unlikely this would fail often in practice - find/replace including could trick it)*.

>     * What do you think about making nil the default value for `hl-block-bracket`?  Any reason not to?  The big benefit is that the mode will then work without modification in (for example) lisp and python mode for certain nested structures.

This is intended to help with C family languages and think it's most useful for this family of languages, maybe this could be configured per-mode?

So `hl-block-bracket-emacs-lisp-mode` could be set to `nil` (or `(`), then other languages can default to `{`.

riscy:
Defaults configured on a per-mode basis makes a lot of sense to me -- `highlight-indentation.el` (https://github.com/antonj/Highlight-Indentation-for-Emacs) does a good job with a lot of different languages and has a similar use case, so may be worth a look.

In any case this looks distinctive enough that I think it warrants a merge.  Thanks for your patience with the back and forth.

PR #6151, opened 2019-05-02
adding recipe for emacs-todoist

abrochard:

interface for todoist.com task tracking

https://github.com/abrochard/emacs-todoist

maintainer

**None needed**

### Checklist

Please confirm with `x`:

- [ ] I have confirmed some of these without doing them - ahah

### DIFF
--- a/recipes/todoist
+++ b/recipes/todoist
@@ -1,0 +1,1 @@
+(todoist :repo "abrochard/emacs-todoist" :fetcher github)


abrochard:
Hi @riscy thank you for the feedback! I'll tackle this asap.

abrochard:
I just pushed a few commits:

- `url-http-end-of-headers` is defined in the `url-http` package so I required it

purcell:
Thanks! Looks good, and I'll merge it now. Couple of little notes along the way:

- Prefer `string-equal` in cases like [this](https://github.com/abrochard/emacs-todoist/blob/c56a0b0e35f29b91453d9d3b5f58a919f5f6a08d/todoist.el#L80). Also when one branch explicitly returns `nil`, I personally find it clearer to use `unless` or `when`.
- `dolist` would be even clearer than `mapc` [here](https://github.com/abrochard/emacs-todoist/blob/c56a0b0e35f29b91453d9d3b5f58a919f5f6a08d/todoist.el#L336)

abrochard:
Thank you! I'll tackle those asap :)

PR #6152, opened 2019-05-04
Python ipdb support as a separate realgud module

rocky:

The replaces to the emacs debugger front-end realgud python ipdb support. 

Note: after this is merged into MELPA, it will removed from realgud
(the core code).

https://github.com/realgud/realgud-ipdb

I am one of the maintainers 

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/realgud-ipdb
+++ b/recipes/realgud-ipdb
@@ -1,5 +1,5 @@
 (realgud-ipdb
  :fetcher github
  :repo "realgud/realgud-ipdb"
- :files ("realgud-ipdb.el"
+ :files (:defaults 
          ("ipdb" "ipdb/*.el")))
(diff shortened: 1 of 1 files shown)


purcell:
Also needs to declare a package dependency on `load-relative`, then it should be good to merge.

rocky:
Another Ooops - it should be there now in https://github.com/realgud/realgud-ipdb/commit/5fd179d0045ee95784beafe5799f69c9a92bd6da

As before I also added a dependency on cl-lib. If that shouldn't be there, let me know and I will remove it. Do I also need a dependency on test-simple for testing? (I haven't converted to ert yet).

As before double check that I did this right.

And most importantly, heartfelt thanks.

purcell:
Thanks for this!

PR #6153, opened 2019-05-04
lldb debugger support in realgud package

rocky:

Adds support for the lldb debugger inside the realgud debugger front-end package

https://github.com/realgud/realgud-lldb

**None needed**

There are at least two more realgud packages in the javascript space, but it is best to learn the ropes and get feedback with this and realgud-ipdb before starting the others.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/realgud-lldb
+++ b/recipes/realgud-lldb
@@ -1,5 +1,5 @@
 (realgud-lldb
  :fetcher github
  :repo "realgud/realgud-lldb"
- :files ("realgud-lldb.el"
+ :files (:defaults
          ("lldb" "lldb/*.el")))
--- a/recipes/realgud-lldb
+++ b/recipes/realgud-lldb
@@ -1,0 +1,5 @@
+(realgud-lldb
+ :fetcher github
+ :repo "realgud/realgud-lldb"
+ :files ("realgud-lldb.el"
+         ("lldb" "lldb/*.el")))


recipes/realgud-lldb
```diff
@@ -0,0 +1,5 @@
+(realgud-lldb
+ :fetcher github
+ :repo "realgud/realgud-lldb"
+ :files ("realgud-lldb.el"
```
purcell: I'll just edit this to use `:defaults` instead of this specific filename, and then merge.

purcell:
I merged this a bit prematurely because the `.el` file is lacking a package dependency on `load-relative`. Can you add that please?

rocky:
Ooops - it should be there now in https://github.com/realgud/realgud-lldb/commit/2296cc465f98d8fdd3b7af2b0dc167feb899b1be  

I also added a dependency on cl-lib. If that shouldn't be there, let me know and I will remove it. Do  I also need a dependency on test-simple for testing? (I haven't converted to ert yet). 

Please double check that I did this right. I make a lot of inadvertent mistakes - that is one reason I work on debuggers. 
 
Sorry for all of the trouble and thanks for all of your help.

purcell:
> Do I also need a dependency on test-simple for testing? (I haven't converted to ert yet)

No, because we don't bundle `*-test.el` or `*-tests.el` files.

> Please double check that I did this right. I make a lot of inadvertent mistakes - that is one reason I work on debuggers.

It all looks good.

> Sorry for all of the trouble and thanks for all of your help.

No trouble at all.

PR #6154, opened 2019-05-05
add magit-diff-flycheck recipe

ragone:

Run `M-x magit-diff-flycheck` in a magit-diff buffer to display a
filtered list of errors for the added/modified lines only.

This is primarily meant to be used on legacy projects where you do
not want to see errors for the whole file but only for the added/modified
lines.

This works across multiple files in the same error-list.

output-2019-05-05-21:49:56

https://github.com/ragone/magit-diff-flycheck

Maintainer and author

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/magit-diff-flycheck
+++ b/recipes/magit-diff-flycheck
@@ -1,0 +1,1 @@
+(magit-diff-flycheck :repo "ragone/magit-diff-flycheck" :fetcher github)


ragone:
Thank you for your feedback @riscy, really appreciate it.

I have pushed fixes for the 

> * The functions `first` and `second` come from the cl which is deprecated.  You can `(require 'cl-lib)` though and use `cl-first` and `cl-second`

Fixed by using `nth` instead.

> * Your email address has two `.com` at the end

Removed.

> I'm trying to run the mode and having some troubles. First I'm running M-x magit-diff and select u (unstaged commits). Then in the diff buffer I run M-x magit-flycheck-diff, but I run into an error -- here's my traceback if it helps:

I am unable to replicate the error and I don't see how `magit-diff-visit-file` can return `nil` either. However, I have now properly handled this. Thank you.

riscy:
Great, error looks like it's gone for me.  After some experimenting I'm noticing:
- I turn on flycheck on my buffer, run magit-diff on unstaged changes, and run `magit-diff-flycheck` -- works great.
- I turn flycheck _off_ in my buffer, run magit-diff on unstaged changes, and run `magit-diff-flycheck` -- no report.
Can you recreate this?  If so, what do you think about checking the variable `flycheck-mode` in the original buffer and offering to turn flycheck on for the user if it's not enabled?

ragone:
That's a great idea! I have added that with my latest commit.

riscy:
This is a great mode -- I'll be one of your users.  Thanks for this, and welcome to MELPA! :)

PR #6155, opened 2019-05-05
magit-libgit: drop spurious trailing quote

ackalker:
While Emacs' Lisp reader is perhaps fine with the dangling double quote at the end of the recipe expression, a (non-Lisp) tool which I wrote for local use caught it and barfed on it.


tarsius:
Thanks!

ackalker:
You're welcome :-)

PR #6157, opened 2019-05-06
phpstan: drop spurious trailing closing paren

ackalker:
Found another one with a trailing character. I believe there are no more recipes with those, so I don't think you need to worry about more spam from me for now :-)


purcell:
Thanks!

ackalker:
You're welcome, again :-)

zonuexe:
I'm sorry. Thanks!

PR #6159, opened 2019-05-07
Add goto-last-point

manuel-uberti:

A minor mode to record and jump to the last point in the buffer.

Original author is @chrisdone, I hope he'll be happy with this. :)

https://github.com/manuel-uberti/goto-last-point

I am the maintainer.

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/goto-last-point
+++ b/recipes/goto-last-point
@@ -1,0 +1,1 @@
+(goto-last-point :fetcher github :repo "manuel-uberti/goto-last-point")


manuel-uberti:
@riscy thanks for the feedback:

- I marked some functions as private as you suggested.
- I think the name is fine and clear as it is, but if there is a need to change it into somewhere else I'll do it.

manuel-uberti:
I've also rephrased the commentary section to better clarify the package purpose.

riscy:
That looks great.  One thing I'm noticing which I missed on my first pass (sorry about that) is this:
```elisp
(declare-function ring-empty-p "ext:ring-empty-p")
(declare-function ring-remove "ext:ring-remove")
(declare-function ring-insert "ext:ring-insert")
```
Why not just `(require 'ring)` ?

manuel-uberti:
Because I'm silly. :)

Fixed it.

riscy:
Great!  This last one is subjective, but it might be nice if turning off the mode killed the variable goto-last-point-stack (`kill-local-variable`).  This way if someone wanted to change the value of `goto-last-point-max-length`, they could just toggle the mode to have that change take effect.  Right now it seems you have to kill the buffer to guarantee you get a new ring.

I'll leave that one up to you!  Thanks both @chrisdone and @manuel-uberti for packaging this up.

manuel-uberti:
Thanks, I'll add `kill-local-variable` to the `else` part of `define-minor-mode`.

purcell:
I feel like I've seen a package like this arrive in MELPA, but I can't find it now... :-)

manuel-uberti:
Sorry, didn't notice a similar package already there. If this is just duplicating functionalities, feel free to remove it of course.

manuel-uberti:
[jumplist](https://github.com/ganmacs/jumplist) offers a similar functionality, from what I understand from the README, although it looks like it has been updated in a while. Which of course could just mean the package works fine as it is.

purcell:
Yeah, jumplist was probably the one I was thinking of. Anyway, not a big deal in this case, I think.

PR #6160, opened 2019-05-07
Add flycheck-clolyze recipe

DLaps:

This package adds clolyze to flycheck.

https://github.com/DLaps/flycheck-clolyze

Author and maintainer

**None needed** 
### Checklist

Please confirm with `x`:

- [] I have confirmed some of these without doing them

### DIFF
--- a/recipes/flycheck-clolyze
+++ b/recipes/flycheck-clolyze
@@ -1,0 +1,1 @@
+(flycheck-clolyze :fetcher github :repo "DLaps/flycheck-clolyze")


riscy:
Code looks good.  Have you thought about contributing this to the existing `flycheck-clojure` project (https://github.com/clojure-emacs/squiggly-clojure/blob/master/elisp/flycheck-clojure/flycheck-clojure.el)?  I would recommend that because they appear to have a solid package over there.

If that turns out to be infeasible, it may be worth renaming/redescribing your Emacs package to include "Clojure" somewhere in it.  People surfing the package index may not know that clolyze has anything to do with Clojure.

purcell:
Merging now - if you end up deciding to contribute the code into `flycheck-clojure` as @riscy suggested, let us know, and we can remove this recipe again. In the meantime, welcome to MELPA!

> it may be worth renaming/redescribing your Emacs package to include "Clojure" somewhere in it. People surfing the package index may not know that clolyze has anything to do with Clojure.

For this reason it would be worth including "Clojure" in the package description on the first line of the `.el` file.

PR #6161, opened 2019-05-08
add rfc-mode

galdor:

rfc-mode is a browser and viewer for RFC documents. The viewer highlights headers and section titles, and create links (buttons) for references to other RFC documents. The browser uses Helm and makes it easy to search and select RFC documents.

https://github.com/galdor/rfc-mode

I wrote the package and maintain it.

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/rfc-mode
+++ b/recipes/rfc-mode
@@ -1,0 +1,1 @@
+(rfc-mode :fetcher github :repo "galdor/rfc-mode")


galdor:
Thank you for the feedback ! This is my first time writing anything significant with elisp, I'm happy to fix any issue.

I pushed a v1.1.0 which hopefully fix these points. I can byte compile the file without any warning or error, and `STABLE=t make recipes/rfc-mode` still works as expected.

riscy:
I've taken a deeper look and have tried using this some more -- I made a PR (https://github.com/galdor/rfc-mode/pull/1) for functionality I thought might be useful so that the mode feels a little bit more batteries-included.

I think there's an issue with the naming convention you're using.  It's okay to call the file itself `rfc-mode`, but your entire namespace is `rfc-mode`-prefixed.  My understanding is that the `-mode` suffix is usually reserved for when you're actually defining a mode (e.g. with `define-derived-mode`).  Here's what I think you should do, but I'm open to your thoughts:

- Decide on a namespace prefix (one that isn't `rfc-mode`).  You shouldn't use `rfc` either since that is taken (https://www.emacswiki.org/emacs/rfc.el) but what about something like `rfc-browse`?
- Rename all of your functions to `rfc-browse-X`, and rename your `define-derived-mode` to `rfc-browse-mode`.
- You could rename `rfc-mode-browse` to the more succinct `rfc-browse` instead of `rfc-browse-browse` :) -- that seems like the most natural entrypoint to using this package.

galdor:
I merged your patches, thank you for the contribution :)

Regarding naming in general, it is called `rfc-mode` because it's originally just a RFC document reader which happens to have a browse functions. I would have called it just `rfc` but decided otherwise since `rfc.el` exists (even though it is not in MELPA).

I'd prefer not having to rename everything if it's all the same to you. If it is a deal breaker for inclusion in MELPA, just tell me and I'll close my PR and will just make the mode available on my website.

riscy:
Ah, that sounds totally reasonable to me then.  Welcome to MELPA! :)

galdor:
Thank you !

PR #6162, opened 2019-05-09
Add recipe for mu2tex

smaret:

In scientific literature, it is common to format the symbols for chemical elements, molecules and unit names by using roman (as opposed to italic) letters for the names. This makes such things cumbersome to type, because one would have to do either `$\mathrm{H_2O}$` or `H$_2$O`. This package make this typing easier by providing the function `mu2tex`. This function converts plain ASCII molecule names and unit expressions into either of the above forms. It also makes it easier to write numbers times powers of 10.

https://github.com/cdominik/mu2tex

I'm a contributor and user of the package. The maintainer is @cdominik

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/mu2tex
+++ b/recipes/mu2tex
@@ -1,0 +1,2 @@
+(mu2tex :fetcher github :repo "cdominik/mu2tex")
+


riscy:
Awesome.  I was a little confused at the beginning that my point had to be at the _end_ of a molecule before the conversion would happen -- that may be worth documenting on your end, but otherwise this looks great.

cdominik:
> Awesome. I was a little confused at the beginning that my point had to be at the _end_ of a molecule before the conversion would happen -- that may be worth documenting on your end, but otherwise this looks great.

Yes, it is meant to be used while writing, so I did not mention this in the documentation.  Will do put that in.

Carsten

PR #6163, opened 2019-05-12
Add phpt-mode recipe

zonuexe:

This package provides major mode for editing [PHPT File format](http://qa.php.net/phpt_details.php).

Files of this format are used as test code for [the PHP interpreter](https://github.com/php/php-src).
Also, a subset of this format is included in [PHPUnit](https://phpunit.de/).

https://github.com/emacs-php/phpt-mode

I'm author.

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/phpt-mode
+++ b/recipes/phpt-mode
@@ -1,0 +1,1 @@
+(phpt-mode :fetcher github :repo "emacs-php/phpt-mode")


riscy:
Great - this works perfectly for me.

PR #6165, opened 2019-05-14
Add docker-cli package

bosko:

Enables running commands in Docker container within Emacs buffer.

https://github.com/bosko/docker-cli

Author

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/docker-cli
+++ b/recipes/docker-cli
@@ -1,0 +1,1 @@
+(docker-cli :repo "bosko/docker-cli" :fetcher github)


Silex:
@bosko: are you ok if I get inspiration from this for https://github.com/Silex/docker.el/issues/7 ? I never took the time to find out what needed to be done for a proper `comint` shell, so I just used `async-shell-command` but how you do it looks easy.

bosko:
@Silex of course not. As a matter of act I will be glad if you do so ;)

bosko:
@riscy Thank you very much for this excellent review. Frankly speaking this is one of the best code reviews I've seen recently. I will try to fix all issues as soon as possible.

mattiasb:
@bosko Just a small drive-by comment: Your README asks the user to install `inf-ruby` instead of `docker-cli` from melpa. :)

bosko:
@riscy I think I've fixed all issues from your review.

I was not able to reproduce package-lint warning so I guessed problem was in lowercase 'running' so I've made it uppercase in the first line of the source file.

I removed mentioning of Rails console and keywords setting because both of them are not ready yet and I want to leave them for the next version.

riscy:
Aw thanks for saying so.  One last thing I'm noticing is that you probably want to set the `require-match` argument on `completing-read` to t, for both the commands and the containers (which must exist).

bosko:
@riscy `completing-read` is set to `t` and in addition I've added check if command or container are empty strings.

riscy:
Looks great now.  Welcome to MELPA, and thank you!

PR #6168, opened 2019-05-17
Add dianyou

redguardtoo:

Search and analyze mails in Gnus

https://github.com/redguardtoo/dianyou

I'm the maintainer.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/dianyou
+++ b/recipes/dianyou
@@ -1,0 +1,2 @@
+(dianyou :fetcher github
+         :repo "redguardtoo/dianyou")


tarsius:
You have to fix the name of the package inside the recipe.

redguardtoo:
@tarsius , I fixed the package name

redguardtoo:
I updated the pull request.

riscy:
Thanks for checking that out.  I'm not reviewing this in depth yet since I'm still getting:
```
In end of data:
dianyou.el:333:1:Warning: the function ‘gnus-group-topic-name’ is not known to
    be defined.
```
Can you confirm you're getting the same?  I believe you need to `(require 'gnus-topic)` to pull in the elisp where this function is defined.

At a glance, you should also require `ivy` (for `ivy-read`).  But in the case of ivy-read, why not just use a `completing-read` instead?  Then people can use whatever completing-string-read their hooks are set up to use.

redguardtoo:
@riscy , I updated the code. I ran `byte-compile-file` on `dianyou.el` and got no warning/error this time.

 I think `ivy` is necessary to give user better experience out of box.

redguardtoo:
On second thought, removing `ivy` is good idea. It could grow user base to `helm` users. I updated code.

riscy:
Looks good over here.  Thanks!

PR #6169, opened 2019-05-17
Realgud support for node inspect

rocky:

Adds newer node --inspect debugger support to the realgud debugger frontend

https://github.com/realgud/realgud-node-inspect 

Author/Maintainer

 **None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/realgud-node-inspect
+++ b/recipes/realgud-node-inspect
@@ -1,0 +1,5 @@
+(realgud-node-inspect
+ :fetcher github
+ :repo "realgud/realgud-node-inspect"
+ :files (:defaults
+         ("realgud-node-inspect" "realgud-node-inspect/*.el")))


rocky:
> Just from a mile high view, this is an idiosyncratic package layout

I don't know what this means in an actionable way. The package layout is not that different from say a python or nodejs package layout, or a C project layout.  I gave a talk about this whole process to a local NYC Emacs/Lisp group. The slides for that are [here](http://rocky.github.io/NYC-Lisp-Elisp-talk/#). (This was back in 2012 and the slides need revamping to a newer package like reveal.js rather than impress.js, but you get the idea - which is that there was thought to how this project is structured. 

Whenever someone sees a problem and tries to address it, there is always going to be _someone_ who says, "but this is different!"  It may be that in the intervening 7 years the Emacs community has adopted a better way to structure projects. If that is so and it addresses the concerns I had 7 years ago, then great -
 point me to where this newer better approach is spelled out. It might take me a while get around to changing things over though. (See below.)

> and not all of them conform in spirit to package-lint and checkdoc.

That's interesting because I did use package-lint and checkdoc. So be specific what problems did you find? If we are talking about colons in names, this kind of discussion is silly. But feel free to put in a PR to address stuff like this. 

> I feel like a bit of cleanup could go a long way.

Great! Since this of course is in a git repository, pblease fork the code, and make pull requests. 

There is much to do, and alas I don't have time to do everything I would like. So I often have to put in FIXME's and move on. If you or others would like to fix the FIXME's, I'd be grateful. Otherwise, they can stay there as a reminder that there is more to do. 

Lastly, let me say that this project is _already_ in [ELPA](http://git.savannah.gnu.org/cgit/emacs/elpa.git/log/?h=externals/realgud-node-inspect). So the question is now whether we would like it to be in MELPA as well. 

From my standpoint, I probably will rarely use this package whether from MELPA/ELPA, github, or my own local git repository of it. For me, [realgud-trepan-ni](https://elpa.gnu.org/packages/realgud-trepan-ni.html) is much nicer experience and interacts better with realgud and emacs better. 

As I too often do, I added this for _others_ who don't want to use [trepan-ni](https://www.npmjs.com/package/trepan-ni). 

I understand that different people have different views about software and so on, so I've come to live with different ideas and views on software. There is a very small chance that someone will get interested in this package and then as a result of that, the improvements made there can be adopted to other similar packages.

riscy:
Hi, I totally sympathize with where you're coming from.  But the checklist helps maintain a certain level of quality across _all_ packages -- I feel obligated to apply the same standard to all contributions, regardless of the package's background or author.  I'll check out your slides.

To try to be more helpful, I have gone through and run the checkers on your code (and a couple additional lints from the built-in Emacs linter and my own eyes).  Again, it would be fair to call some of them false positives due to the package layout, so I haven't listed every single lint.

`checkdoc`:
- `node-inspect-query-cmdline` could have a documentation string
- `realgud:node-inspect-track-mode-hook` could have a documentation string
- `realgud:node-inspect-suggest-invocation`'s docstring should mention DEBUGGER-NAME (as a bonus, Emacs' online help will detect these and fontify them which can be useful)
  - But if we look at the function we can see it isn't used.  Perhaps it isn't necessary, or otherwise you can call the variable `_debugger-name` to indicate to Emacs and your users that it isn't used
- `realgud:node-inspect-reset`'s docstring could make its first line into a complete sentence; otherwise in Help-apropos we just see an incomplete line: ```Node-Inspect cleanup - remove debugger’s internal buffers (frame,```
  - Same with `realgud:node-inspect-remove-ansi-schmutz`, `realgud:node-inspect-reset`, `realgud:node-inspect-command-hash`

One benefit of using a standard package layout is that you would also have seen these from `package-lint`:
- `node-inspect-query-cmdline` might collide with the "namespace" of another hypothetical package called `node-inspect`
  - Same with `node-inspect-track-mode`, `node-inspect-parse-cmd-args`

Misc:
- In `node-inspect.el`: strictly speaking, `declare-function`'s second argument should be a string (`check-declare-file` calls these "malformed declarations"), but I can see the intent.  I think you want to write `realgud:node-inspect-track-mode` instead of `realgud-node-inspect-track-mode` given your current choice of "namespace"
  - I'd be worried about having these declarations collide with another package's namespace -- are they necessary?
- In `node-inspect-parse-cmd-args`: `node-opt-two-args`, `node-two-args` are bound but never used
- In `realgud:node-inspect`, `cmd-buf` is bound but unused except in a commented-out block of code

rocky:
> Hi, I totally sympathize with where you're coming from. But the checklist helps maintain a certain level of quality across all packages -- I feel obligated to apply the same standard to all contributions, regardless of the package's background or author. 

And I totally sympathize where you are coming from too. This reminds me of the counterpoint theorists who criticize J. S. Bach's counterpoint. And I can imagine some diligent music-publisher reviewer making the same kinds of standards criticisms. (And in his manuscripts sometimes I see that his sometimes note stems are to right of the notehead instead of to the left. Also he sometimes uses this weird Soprano C clef - it is a wonder that he ever got his music played!)

I am not a priori opposed to changes you suggest. it just probably won't be done by me. That's why I've opened an issue _in the project_ to record your comments.  (Pssst standards guy: this too is the standard way to request changes to a github project. Normally a PR is about the code in the PR, not code somewhere else. So many standards to follow!)

Should someone (even possibly you) care enough about in the changes you suggest or care about MELPA packaging of this particular package, then she will more likely see the changes there than here. 

The limited amount of time I have for realgud and nodejs that I have will go into [MELPA realgud-trepan-ni](https://melpa.org/#/realgud-trepan-ni)/[ELPA realgud-trepan-ni](https://elpa.gnu.org/packages/realgud-trepan-ni.html).

As for the other packages realgud-node-debug and realgud-jdb they are in a slightly different situation because right now they appear in realgud, but in order to be able to better maintain things, I have created new repositories. Soon I will be pulling them out of realgud. So although those might be nice to have in MELPA, as with this package, those are already in ELPA.

riscy:
> Pssst standards guy: this too is the standard way to request changes to a github project

Sorry about that

Let's leave this open in case a collaborator/maintainer with more experience decides this is fine.

purcell:
Thanks for all the discussion here, and sorry it's taken me a while to get to this. I'm basically okay with this, since we have a bunch of hte other realgud packages here already, so I'll go ahead and merge.

> Lastly, let me say that this project is _already_ in [ELPA](http://git.savannah.gnu.org/cgit/emacs/elpa.git/log/?h=externals/realgud-node-inspect).

I didn't realise this, and generally we don't find it super helpful to host packages that are also in GNU ELPA, because it can lead to confusion about which of two different versions is the canonical one -- here the code would exist both in your github repo and in the Savannah git, where others could also modify it. Is it wise/helpful for us to continue to host it here too?

rocky:
>  and generally we don't find it super helpful to host packages that are also in GNU ELPA, because it can lead to confusion about which of two different versions is the canonical one -- here the code would exist both in your github repo and in the Savannah git, where others could also modify it. Is it wise/helpful for us to continue to host it here too?

Yes, this is weird and not optimal. I also can't imagine how things would have worked out otherwise. (Some of that below). However I would prefer like this to exist in MELPA at least in unstable as that is updated automatically. And I can get statistics of downloads of the packages which I can't in ELPA. 

The github code in the realgud organization serves _both_ git repositories. What is in the ELPA repository is pulled in from the github organization. So right now in fact in order to change what is in ELPA has to be done via the same github repo that MELPA tracks. 

A benefit of having this be in both ELPA and MELPA is that I have gotten help from people on the ELPA side in improving the code, which wouldn't have happened if this were only on github or MELPA. And to the extent that MELPA standards don't conflict with ELPA standards, _in theory_ the packages would (eventually) follow the union of the two standards which presumably leads to goodness.

But having laid out the situation, I leave it up to you how best to proceed. Whatever you want to do MELPA wise is fine with me. 

As always, thanks for all the help. 

@riscy Just today I got around to revising the ELISP software development slides for the [2012 talk I mentioned before](http://rocky.github.io/NYC-Lisp-Elisp-talk/#/) so it might make sense now. There is still some funny centering markup that should be addressed eventually. And I still need to fold in the text into the slides. There is always a bit more to do than I am able to accomplish.

PR #6170, opened 2019-05-18
Add eshell-toggle recipe

a13:

[Eshell-toggle is another drop-down shell (but can be customized to be a scratch buffer of any kind).]

https://github.com/4DA/eshell-toggle

A contributor

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/eshell-toggle
+++ b/recipes/eshell-toggle
@@ -1,0 +1,1 @@
+(eshell-toggle :repo "4DA/eshell-toggle" :fetcher github)


riscy:
Adding label on compiler error fix -- ping when ready.  Also it looks like you need a require on projectile.

alphapapa:
@a13 

1.  Why are you submitting the package instead of its author?
2.  If the package is intended to be used for more than just eshell, it should probably be refactored and renamed, with eshell being one of the prominent example use cases.

purcell:
Thanks @alphapapa , I second those questions.

I also spotted that `condition-case` is [being used](https://github.com/4DA/eshell-toggle/blob/master/eshell-toggle.el#L117) to call `projectile-project-root` but silently fail if `projectile` is not installed, which is very hacky! You should usually consider using `fboundp` to check for its availability instead. (Also note that `ignore-errors` is a neater way of writing that `condition-case`.) But the best solution here is to remove the `eshell-toggle-use-projectile-root` [custom var](https://github.com/4DA/eshell-toggle/blob/2eb91974047f5caf8df3bf3af5014be2cc95ddac/eshell-toggle.el#L69) and instead have one called `eshell-toggle-project-root-function`, which defaults to `'ignore` (i.e. always return nil.) Then users can set it to `projectile-project-root` if they want, and you don't have to have any references to `projectile` in your code.

a13:
@alphapapa thanks!
1. I don't think that the original author cares about the package as much as I do these days :), at the same time I see no reason to make a fork.

2. any suggestions about naming?

@purcell thanks for your comment, I'm not the original author of the projectile-related code there, but I'll rewrite it when I'll have more spare time.

alphapapa:
> I don't think the original author cares about the package as much as I do these days (see the history) :), at the same time I see no reason to make a fork.

If the author doesn't care about the package, then it shouldn't be on MELPA, because a) if it's not maintained, it may not work well; and b) if the author does push changes, he may do so without regard for MELPA users.  So if you do care about it, you should probably fork it, and then you could be the maintainer, and then you could submit it to MELPA as the maintainer.

> any suggestions about naming?

If you plan to refactor and rename it, go ahead and refactor it, and worry about the name later.  One will probably come to you as you refactor it.  :)

purcell:
The author is still committing their own and contributed changes, so I don't think there's any need for forking or renaming unless @a13 plans to strike out alone on a new path. Let's go ahead and merge this, and @a13 can submit a fix for the issues I mentioned as and when ready. :-)

a13:
@purcell Thank you very much!

PR #6174, opened 2019-05-20
node-debug support in realgud

rocky:

Adds realgud support for the older nodejs `node debug` debugger

https://github.com/realgud/realgud-node-debug

Author/maintainer

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/realgud-node-debug
+++ b/recipes/realgud-node-debug
@@ -1,0 +1,5 @@
+(realgud-node-debug
+ :fetcher github
+ :repo "realgud/realgud-node-debug"
+ :files (:defaults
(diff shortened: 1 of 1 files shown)


riscy:
This looks valuable - I do have a similar request as in https://github.com/melpa/melpa/pull/6169 for a bit of cleanup (e.g., make sure all of your new functions are documented).  Also I notice both packages are shipping with very similar versions of `core.el` -- I think your FIXME comment re: `;;; FIXME: DRY this with other *-parse-cmd-args routines` is apt.

Please ping after a bit of cleanup.  Thanks!

rocky:
As I wrote elsewhere, this is probably not going to happen soon. Until then, folks who want to use can pull this from ELPA.

PR #6177, opened 2019-05-21
add ros.el

DerBeutlin:

A package exploiting `completing-read` functions to interact with ROS (Robot operating systems).
In contrast to other available packages it is not limited to be used with helm or ivy and provides a wider range of functionality from building packages, running tests, auto-complete include/imports, interacts with ROS topics, services, nodes...

https://github.com/DerBeutlin/ros.el

Maintainer

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/ros
+++ b/recipes/ros
@@ -1,0 +1,3 @@
+(ros
+  :repo "DerBeutlin/ros.el"
+  :fetcher github)


riscy:
Cool.  I have no experience with ROS but I've left a PR up here https://github.com/DerBeutlin/ros.el/pull/1 addressing a few minor issues -- feel free to edit and use your best judgment as to whether anything I changed doesn't work.

In that PR, I used `shell-quote-argument` only on the `profile-flag` variables "to prevent globbing and word splitting".  But you may want to look into using `shell-quote-argument` on `workspace` as well.  The issue is that the default workspace is "~/", which will get quoted incorrectly.  You could solve this by always wrapping `workspace` in `expand-file-name`, e.g.
```elisp
(setq workspace "~/")
(shell-quote-argument workspace)  ; bad
(shell-quote-argument (expand-file-name workspace))  ; good
```
but I'll leave that up for your consideration.

DerBeutlin:
Thank for your help!
Merged the PR and added `shell-quote-argument` and `expand-file-name` in the necessary locations

riscy:
Sorry I missed these on my first pass!
- You may want to look into enabling lexical bindings (put `-*- lexical-binding: t -*-` at the end of the first line of ros.el)
- If you do that, you will get 3 small warnings when byte-compiling:
```
ros.el:183:1:Warning: Unused lexical argument ‘major-mode-name’
ros.el:489:1:Warning: Unused lexical variable ‘test’
ros.el:508:1:Warning: Unused lexical variable ‘type’
```
In some cases the solution is to silence these by adding an underscore before the variable (telling Emacs and your users the variable is bound but not used); in other cases (for example in the case of the `test` variable) perhaps the binding should be removed completely?

DerBeutlin:
I removed  the `test` and `type` variable and replaces `major-mode-name` with an underscore.
No warnings anymore

riscy:
Thanks.  It looks great!

PR #6178, opened 2019-05-22
Add recipe for ctune

maurooaranda:

Add an easy way of adding and removing CC Noise Macros, as well as saving the changes
project-wide.

[ctune](https://github.com/maurooaranda/ctune)

Author

None needed.

### Checklist

Please confirm with `x`:

- [x] I've read [CONTRIBUTING.org](https://github.com/melpa/melpa/blob/maste/CONTRIBUTING.org)

### DIFF
--- a/recipes/ctune
+++ b/recipes/ctune
@@ -1,0 +1,3 @@
+(ctune :repo "maurooaranda/ctune" :fetcher github
+       :files (:defaults
+	      (:exclude "doclicense.texi" "docstyle.texi")))


maurooaranda:
> Nice - this looks valuable and is clean/well documented. It's appreciated. And thanks for being thorough with the checklist. :)

Thank you.  I was afraid the package could be regarded as too specific.  And no need to thank me about following the guidelines: you guys do a very hard work, so I should make an effort to make it easier for you :)

maurooaranda:
I've now updated `ctune` to define a minor mode, `ctune-mode`.  I checked again the relevant items of the checklist, to make sure no errors were introduced.

I did some testing and I am pretty happy with the results,   I hope I didn't miss anything, but if you find something, please let me know.

riscy:
Thanks again for the attention to detail.  It's specific but I think C developers would find it valuable.

I've had a chance to look in a bit more depth.  Reading the docs for `c-noise-macro-names` it looks like C++ and even Objc mode also support this functionality, but your mode strictly checks for c-mode (`if (derived-mode-p 'c-mode)` and `unless (derived-mode-p 'c-mode)`) -- maybe there's some value in supporting c++-mode, even though it isn't derived from c-mode?  (I don't know anything about objective-C macros).

I also noticed the value of `c-noise-macro-names` was being set globally in some cases, which I think is not intentional.  In `ctune--reset-values`, should some of those should be using `setq-local`?

Another couple minor fixes I noticed on this pass:
- Is there value in just adding `(require 'cc-vars)` rather than using `defvar` and `declare-function` to pacify the compiler?
- `"Save the C Noise Macros to the directory locals file? "` -- doesn't need to be wrapped in `format`.  But maybe explicitly listing the filename `.dir-locals.el` would be slightly less ambiguous

maurooaranda:
> maybe there's some value in supporting c++-mode, even though it isn't derived from c-mode? (I don't know anything about objective-C macros).

Makes sense.  I will add support for c++-mode and objc-mode.

> I also noticed the value of c-noise-macro-names was being set globally in some cases, which I think is not intentional

No, it's not intentional.  I wrongly assumed that CC Mode made the `c-noise-macro-*` local.  I have fixed that, locally.  I'll push the changes later.

>  In ctune--reset-values, should some of those should be using setq-local?

That's not necessary.  Relevant variables are made local in `ctune-mode`.  I think the problem you observed is taken care with the above mentioned fix.

> Is there value in just adding (require 'cc-vars) rather than using defvar and declare-function to pacify the compiler?

I'd rather not do that.  I think three such declarations don't deserve switching to a require.

> "Save the C Noise Macros to the directory locals file? " -- doesn't need to be wrapped in format. But maybe explicitly listing the filename .dir-locals.el would be slightly less ambiguous

Yes.  I was going to mention the .dir-locals.el file, hence the `format` call.  But along the way, I decided not to but forgot to remove the `format` call.  The problem is that I wouldn't like to hardcode .dir-locals.el.
There's the possibility of using the value of `dir-locals-file`, but that doesn't take into account the possibility of having a .dir-locals-2.el file.  So that arises the need to check for that.  Furthermore, simply saying putting .dir-locals.el (or .dir-locals-2.el) in the prompt, will make the prompt ambiguous either way, because it doesn't show the full file name.  Showing the expanded file name will make the prompt very long, and I don't feel comfortable with it.
Given the above, I could abbreviate the file name in the prompt, or keep the prompt as it is (modulo the redundant `format` call).  I'd like to keep it as it is, but have no strong feelings about it.  I could change it to include the abbreviate file name, since I wouldn't like to include the expanded file name.

maurooaranda:
I've now updated the package with the fix mentioned, and others bug fixes, plus updates in the manual, doc strings and the comments.

Thank you for your review so far, it is much appreciated.

riscy:
This is working great for me -- I'm going to go ahead and merge. Welcome to MELPA!

maurooaranda:
Awesome, thanks!

PR #6179, opened 2019-05-22
Add spacebar recipe

matthias-margush:
A vim-like tab bar user interface for workspaces, built in eyebrowse.

https://github.com/matthias-margush/spacebar

I am the creator of `spacebar'.

n/a

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/spacebar
+++ b/recipes/spacebar
@@ -1,0 +1,3 @@
+(spacebar
+ :repo "matthias-margush/spacebar"
+ :fetcher github)


alphapapa:
Cool!  The Deft example might also integrate well with Yequake.  :)

purcell:
Thanks! Couple of notes on the code, which is pretty nice overall:

- Prefer `string-empty-p` [here](https://github.com/matthias-margush/spacebar/blob/de404a0fc2921dfe7adb35c7631770bbf8576851/spacebar.el#L121)
- Prefer `unless` to [`(when (not ..) ...)`](https://github.com/matthias-margush/spacebar/blob/de404a0fc2921dfe7adb35c7631770bbf8576851/spacebar.el#L405)

Also +1 for @alphapapa's comments above.

Overall since this is `eyebrowse`-specific, I suspect that `eyebrowse-tabs` might be a much clearer name for the package.

Hope that helps!

matthias-margush:
Thanks for the feedback! Taking a look at it now.

matthias-margush:
I've [update the code](https://github.com/matthias-margush/spacebar/compare/master@%7B1day%7D...master) based on the feedback here:

```
0e8f024 Don't clobber `display-buffer-alist'
3caa55b Save/restore `cursor-in-non-selected-windows` value
c6dd246 Check that spacebar-mode is active in `spacebar-setup-evil-keys'
df0e5e7 Fix check for `deft' (`fboundp' vs `boundp')
eda1386 Use `string-empty-p' to test for empty string
ec4fe28 Move `(require 'eyebrowse)' to top of file
9126b0d Ensure keybindings take effect
e640509 Update keybinding documentation in README
0cfd994 Reword docstring to avoid "matches" for checkdoc
1f473b4 Remove unused variable `space-window'
fc8251b Prefer `unless' to `(when (not ..) ..)'
c74810e Remove unnecessary `:group spacebar's from `defcustom's
de404a0 Initial version
```

I also ran the plugin in the sandbox.

One note: I haven't been able to reproduce @riscy's report of this behavior:
> Correct initialization of spacebar-mode is dependent on the user's current window configuration. For example, if my current window configuration is vertically split (one window on the left, one window on the right), my spacebar takes over the right window.

Thanks again for the detailed feedback!

riscy:
I've tried the mode out in a couple different environments and don't face the same issue I described in my bare-bones environment -- I must have a package installed that's somehow interfering with this mode, but I couldn't figure out what it was.  Just an fyi.

Beyond the PR I mentioned, the message "evil needs to be installed to use evil keybindings" should probably also cover the case when spacebar-mode is not active.

matthias-margush:
> the message "evil needs to be installed to use evil keybindings" should probably also cover the case when spacebar-mode is not active.

👍 [updated](https://github.com/matthias-margush/spacebar/commit/7d67e19450ddf5738c779d3fc488524837be1f38)

riscy:
Sorry about the turnaround on this.  I've had a chance to do a bit more user testing:

- If you run `spacebar-open` when the variable `spacebar-mode` is nil, there's an error.  I think we need to ensure that spacebar-mode is on before letting the user run this. Here's the backtrace:
  ```
Debugger entered--Lisp error: (wrong-type-argument number-or-marker-p nil)
  eyebrowse-free-slot(nil)
  eyebrowse-create-window-config()
  (if (string= "" (spacebar--name (spacebar--current-config))) nil (eyebrowse-create-window-co$
  (if existsp nil (if (string= "" (spacebar--name (spacebar--current-config))) nil (eyebrowse-$
  (let* ((name (or name (completing-read "Workspace: " (seq-map (function spacebar--name) (eye$
```
- I'm running `spacebar-mode` in bare-bones Emacs (albeit on the terminal) and I've been able to reproduce some glitchy behavior:
  1. <kbd>M-x load-lib spacebar</kbd> and <kbd>M-x spacebar-mode</kbd>
  1. <kbd>M-x spacebar-open</kbd>: I create a workspace called "1"
  1. I similarly create a workspace called "2".  Both workspaces are open to the "*GNU Emacs* splash buffer.  Works great, and I can switch between them without issues.
  1. I move to workspace 2 and run <kbd>C-c C-w ?</kbd> which changes the window configuration by opening a help window, but it also doubles the height of my spacebar:
image

For reference, the only packages I have installed are:
```elisp
(package-install 'package-lint)
(package-install 'pkg-info)
(package-install 'request)
(package-install 'markdown-mode)
(package-install 'eyebrowse)
```
If I'm additionally using `ivy` for completing-reads, I run into the issue almost immediately since ivy changes the size of the minibuffer.  I also run into similar spacebar-resizing issues when I use C-g to interrupt a command at the minibuffer, etc.

Sorry for the verbosity, but are you able to reproduce this?

riscy:
Woops, I forgot: you will need to `(require 'subr-x)` now that you're using `string-empty-p`.

matthias-margush:
Thanks @riscy. I couldn't reproduce the nil for some reason, but it makes sense to guard the interactive functions when the mode is inactive. Updated that here: https://github.com/matthias-margush/spacebar/commit/d56074cffb9f55cc72ac0cc441f1439aeac51746

As far as the resizing issues, I've noticed some similar interactions with other packages as well. I'll need to fiddle around with it more to see what's happening. In the meantime, I created an issue to track this improvement: https://github.com/matthias-margush/spacebar/issues/2

matthias-margush:
After some fiddling around, I was able to reproduce the ivy issue by running `emacs -Q`, then evaluating:
```emacs-lisp
(require 'package)
(add-to-list 'package-archives
	     '("melpa" . "http://melpa.org/packages/") t)
(package-refresh-contents)
(package-install 'package-lint)
(package-install 'pkg-info)
(package-install 'request)
(package-install 'markdown-mode)
(package-install 'eyebrowse)
(package-install 'ivy)

(require 'ivy)
(ivy-mode)
(add-to-list 'load-path "~/code/spacebar")
(require 'spacebar)
(spacebar-mode)
```

Setting [window-size-fixed to t](https://github.com/matthias-margush/spacebar/commit/fcb8015cbf804de07c37c437df7fad27315db4da) seemed to fix it, for me.

riscy:
This works great for me now.  Awesome.  And welcome to MELPA! :)

PR #6181, opened 2019-05-24
Create evil-hop

3e33:

Evil-Hop brings Vim's EasyMotion--like functionality to Emacs. Although written from scratch, it takes its inspiration of how jump keys and Evil integration should work from the way EasyMotion works in Vim.

https://github.com/3e33/evil-hop

Author / maintainer.

None needed.

### Checklist

Please confirm with `x`:

There are are a few warnings but I don't think it's anything major.

I get "invalid syntax" when trying to build with Make under Windows, see comment below.

### DIFF
--- a/recipes/evil-hop
+++ b/recipes/evil-hop
@@ -1,0 +1,2 @@
+(evil-hop :repo "3e33/evil-hop"
+          :fetcher github)


3e33:
I tried to build with Make following your [guide](https://github.com/melpa/melpa/blob/master/CONTRIBUTING.org) under Windows 7 but all I get is this:

```
λ  make recipes/evil-hop
 • Building package evil-hop ...
ERROR: Invalid syntax. Default option is not allowed more than '1' time(s).
Type "TIMEOUT /?" for usage.
make: [Makefile:110: recipes/evil-hop] Error 1 (ignored)
```

noctuid:
Avy is a generic library and can work fine with evil. Have you seen [evil-easymotion](https://github.com/PythonNut/evil-easymotion), for example? A comparison with it might make more sense than a comparison with avy directly.

3e33:
evil-easymotion depends on Avy, it's just an extension of Avy.

noctuid:
I think a specific comparison for those who see no issues with avy would make sense then.

3e33:
Okay thank you for your feedback, I have added an Avy / Evil-Hop comparison, please see the readme.

https://github.com/3e33/evil-hop/blob/master/README.md

noctuid:
So looking at your comparison there, it looks like you don't actually have any fundamental issues with avy as a library just with the default motions avy provides. Avy as a library can be used for any situation where you want to put overlays over various buffer positions, select one (or more), and then take some action. It's not just for motions.

Evil-easymotion also builds the overlay position list by running the actual motion command, supports any motion, and supports puting overlays only after the point instead of throughout the entire window (or before the point when used with a backwards motion).

I'm not trying to say that evil-hop doesn't have a use case (e.g. `evil-hop-entry` looks convenient, and evil-hop looks like it would be a good choice for someone that wants something that is simple and just works without any tinkering), but I think you may still want to consider using avy unless you have some other fundamental problem with it (in which case maybe it would make sense to make an issue about that deficiency). Leveraging avy would give you a lot of nice functionality for free. For example, avy implements customizable keys, has several ways of generating the key sequences that appear and styling/changing the positioning of the overlays, allows customizing which windows overlays appear in (probably not as useful in this case here), allows for resuming an avy command with `avy-resume`, etc.

3e33:
Don't get me wrong, Avy is a nice package with a lot of functionality, I've been using it for a while now since I couldn't find any alternative. But now I've built an alternative, and I don't see myself going back to Avy now that I have something that works like EasyMotion.

I'm not sure why you're trying to convince me to use Avy after I put so much work into building something as an alternative, precisely because I don't want to use Avy. If I saw Avy as an option, I'd still be using it. I've tried tinkering with it and built extensions to make it work more like EasyMotion. I tried evil-easymotion as well, but ended up with the same problems because it also runs everything through Avy.

So I ended up building my own package. If you don't see any use for it, that's fine. I think people coming from Vim will like it more than Avy, because it works like EasyMotion, which I think is far superior to Avy.

Avy is great for some people, I get that, I'm not going back to it, I built this so that I don't have to go back to it. Hopefully other people will also find it useful in case they also have trouble with Avy.

noctuid:
From the issues you've listed, it seems to me that you don't understand what the avy library actually provides at its core (which again *is not* motions but functions for displaying/selecting overlays). Maybe I'm wrong, but I suspect you are unfamiliar `avy-process` and didn't try using it for your package. And if you didn't, I'm not sure why you would assume avy is the cause of any issues you had with evil-easymotion.

If you did try avy and found the library lacking, even if you don't want to contribute the feedback to avy, I think potential users of your package will also be curious as to why and what benefits your package has. That said your package is not really an alternative to avy. It's an alternative to the motion commands provided by avy. Using avy for your package would not involve using any motions from avy. I don't personally care whether or not you use avy to implement your package. I just think it would be easier on you, make your package a lot more featureful (see the custom variables avy provides), and be more consistent with other packages.

If you have some other reason for not liking the helpers avy provides, that's fine, but it's not clear to me that you do, which is why I still suggest you take another look at avy. Avy doesn't have very comprehensive external documentation, so looking at a package that uses it would be more useful. The most basic usage is to pass a list of buffer positions to `avy-process`. Avy places the overlays and then calls `avy-action` with the candidate the user selects (jumps to the position by default). This is all pretty generic functionality, so I'm not sure how it would conflict with the goals of your package.

3e33:
I'm not sure if you've used EasyMotion, but there are differences in how the overlays get built. Anyway I'll close the pull request for now and keep working on Evil-Hop, and see what happens in the future. Thanks for your input!

noctuid:
I'm not sure if you're talking about how the key sequences for the overlays are built, but the keys generated by evil-hop look very similar to or the same as those generated by default with `avy-style` as `at-full` (the screenshot in the readme is `avy-style` as `de-bruijn`). I do think it would be nice if the overlay styling was completely customizable and have made an issue about it.

Also, I have no affiliation with MELPA (in case it's not clear). If avy's customization for overlay styling is too restrictive, or you don't want to use avy for whatever reason, feel free to ignore my comments and reopen the PR.

3e33:
> I'm not sure if you're talking about how the key sequences for the overlays are built, but the keys generated by evil-hop look very similar to or the same as those generated by default with `avy-style` as `at-full` (the screenshot in the readme is `avy-style` as `de-bruijn`).

In the screenshot Avy uses from 2 key presses to 3 key presses, Evil-Hop and EasyMotion use 1 key press to maximum of 2 key presses for the length of positions.

Evil-Hop / EasyMotion have lower number of key-presses the closer to the cursor, Avy generates from the top of the window so the closer to the cursor often ends up with 3 key presses, and further out is 2 key presses.

Evil-Hop and EasyMotion generate from the cursor, closer to the cursor is 1 key press, further away is 2 key presses. Evil-Hop does not use De Bruijn, and I don't think EasyMotion does either.

As far as user interaction goes, less key presses = better. The amount of key presses Avy requires the user to do is not as efficient as EasyMotion and Evil-Hop.

noctuid:
Avy starts with 1 keypress too with at-full (the default style). De Bruijn is just another option. 

The default avy motions start at the window, but you can of course create avy motions that start from the point like evil-easymotion allows.

3e33:
> Avy starts with 1 keypress too with at-full (the default style). De Bruijn is just another option.
> 

I'm not sure I understand, "at-full" from the Avy github is just a display style, part of avy-style.

https://github.com/abo-abo/avy/wiki/defcustom

> at-full: full path on target, obscuring the target and the text behind it.

My screenshots were made with the default Avy style.

I'm not talking about the style choice for displaying the options, I'm talking about how many key presses it actually takes to get to a jump position.

When Evil-Hop displays 1 jump key, it takes 1 keyboard press to jump there. If it displays 2 keys, it takes 2 presses always until the number of jump positions is over 100, which is fairly rare.

noctuid:
`avy-style` controls both the display style and key generation currently (this is another thing I mentioned in the issue I opened for avy). I didn't look closely enough before. It looks like you did use `at-full` but with different keys from evil-hop (which doesn't really make sense for a comparison).

> Anyway I'd suggest trying Avy and EasyMotion / Evil-Hop side-by-side and you'll see the differences, they're pretty obvious in my opinion.

Like I said, they look very similar to me; here what avy looks like for me:

easymotion

3e33:
You're right it was using different keys, not very fair for a comparison I'll have to update that. I'm not sure why avy-style controls key generation that doesn't seem very clear from the documentation.

PR #6182, opened 2019-05-25
Add sly-docker

fisxoj:

This is a contrib (plugin) for sly, a package for interactive editing of common lisp code.  This contrib handles pathname translation into and out of docker containers that the code might be running in for jumping to function definitions.

https://github.com/fisxoj/sly-docker

I am the maintainer!

*None needed**

### Checklist

Please confirm with `x`:

  - There is one override of a function from the sly package, but that will be removed once I get a pull request merged there.  The maintainer is fine with the change.

### DIFF
--- a/recipes/sly-docker
+++ b/recipes/sly-docker
@@ -1,0 +1,5 @@
+(sly-docker :fetcher github
+            :repo "fisxoj/sly-docker"
+            :files (:defaults
+                    "*.lisp"
+                    "*.asd"))


purcell:
Thanks! Notes:

- `package-lint` should have told you that you're missing a package dependency on `cl-lib`, which isn't included in the minimum `emacs` version you're targeting.
- It doesn't make sense to `autoload` [this line](https://github.com/fisxoj/sly-docker/blob/7e372fe8939bde5439febe9b359f715de713040f/sly-docker.el#L126) with the `'append` option, because you have no guarantees at all about which order the autoloads will be executed in. Your contrib could end up being the first one in the var, or it might even overwrite the default value for `sly-contribs`. Better to instead ask the user to load this file explicitly in their config.

>     * There is one override of a function from the sly package, but that will be removed once I get a pull request merged there.  The maintainer is fine with the change.

Use a renamed version of this locally until then rather than potentially clashing with the name of the upstream version. Either way you'll need/want code changes after your upstream change lands.


Hope that helps!

purcell:
Heya @fisxoj, do you plan to fix up the above couple of minor points so we can merge this?

DamienCassou:

PR #6185, opened 2019-05-26
Add janet-mode package

ALSchwalm:

Adds a new major mode for the Janet programming language: https://janet-lang.org/

https://github.com/ALSchwalm/janet-mode/

Package author

None Needed

### Checklist

Please confirm with `x`:

Note that `package-lint` reports 3 errors erroneously.

### DIFF
--- a/recipes/janet-mode
+++ b/recipes/janet-mode
@@ -1,0 +1,1 @@
+(janet-mode :repo "ALSchwalm/janet-mode" :fetcher github)


purcell:
Thanks!

Couple of notes:

- I'm pretty sure there'll be an alternative to [lines like these](https://github.com/ALSchwalm/janet-mode/blob/ed849caede0ef020c21d0bbd4a00930834292bcc/janet-mode.el#L66) which doesn't require `eval`, which should be seen only extremely rarely in normal code.
- I'm pretty sure the syntax table will be set for the mode automatically by `define-derived-mode`, so [this line is redundant](https://github.com/ALSchwalm/janet-mode/blob/ed849caede0ef020c21d0bbd4a00930834292bcc/janet-mode.el#L395)

But in any case, these are not blockers to merging, which I will do now. :-)

PR #6186, opened 2019-05-26
Add recipe `helm-esa`

masutaka:

`helm-esa.el` is a helm interface for [esa](https://esa.io/) which is a document share service.

https://github.com/masutaka/emacs-helm-esa

I am the author.

None needed

### Checklist

Please confirm with `x`:

    ```
    $ emacs -Q -batch --eval '(package-initialize)' -f batch-byte-compile helm-esa.el
    $ echo $?
    0
    ```

### DIFF
--- a/recipes/helm-esa
+++ b/recipes/helm-esa
@@ -1,0 +1,1 @@
+(helm-esa :repo "masutaka/emacs-helm-esa" :fetcher github)


purcell:
Looks great, but why does this require `curl`? The default URL facilities in emacs should be able to handle this case, right? Or one of the wrapper libraries like `request`. Relying on `curl` makes this library less portable between operating systems.

masutaka:
Thanks for your review.

I ported it from the following packages:

* https://github.com/masutaka/emacs-helm-hatena-bookmark
* https://github.com/masutaka/emacs-helm-qiita

Then I did not think deeply about using curl.

I will avoid using curl. Should I fix this time?

masutaka:
I've removed `curl` from the requirements using [request.el](https://github.com/tkf/emacs-request).

Could you review, please?
https://github.com/masutaka/emacs-helm-esa

riscy:
Looks great!

PR #6190, opened 2019-05-27
Add Poetry.el package

galaunay:

`Poetry.el` provides an interface to use [poetry](https://poetry.eustace.io/) in Emacs
(poetry is a python dependency management and packaging tool).
It uses [transient](https://github.com/magit/transient) to provide a magit-like interface.

Poetry.el is almost fully tested (coverage available on the github page).

https://github.com/galaunay/poetry.el

Maintainer (of poetry.el, not poetry)

None Needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/poetry
+++ b/recipes/poetry
@@ -1,0 +1,3 @@
+(poetry
+ :fetcher github
+ :repo "galaunay/poetry.el")


alphapapa:
Cool!  A few notes:

* It isn't necessary to define `:group` in each `defcustom`, as any defined after a `defgroup` inherit the group.

purcell:
Thanks! Notes:

- Delete the `-pkg.el` file: it's redundant, because the metadata in the main `.el` file is complete and sufficient.
- There's little point maintaining a `-version` [variable](https://github.com/galaunay/poetry.el/blob/ed60dd8f4a60410b008ed60d8dc71d8df1517118/poetry.el#L47) these days; the installed package version is what users should use in bug reports, and the likely outcome of a version variable is that you get less accurate bug reports. :-)
- `package-lint` should have told you that `in Emacs` is redundant in text describing an elisp package which can only be used in Emacs.
- Don't use `string-match` + `symbol-name` [here](https://github.com/galaunay/poetry.el/blob/ed60dd8f4a60410b008ed60d8dc71d8df1517118/poetry.el#L57-L60) -- use `eq` or `memq` on the symbols instead.
- Prefer `unless` to `when` + `not`, e.g. [here](https://github.com/galaunay/poetry.el/blob/master/poetry.el#L86)
- [These comments](https://github.com/galaunay/poetry.el/blob/ed60dd8f4a60410b008ed60d8dc71d8df1517118/poetry.el#L249) are pointless because they are redundant, so they should be deleted. As an example, [this one](https://github.com/galaunay/poetry.el/blob/ed60dd8f4a60410b008ed60d8dc71d8df1517118/poetry.el#L277) is already inaccurate.
- Generally you should use `read-string` in `interactive` forms instead of explicitly calling `read-from-minibuffer` e.g. [here](https://github.com/galaunay/poetry.el/blob/ed60dd8f4a60410b008ed60d8dc71d8df1517118/poetry.el#L314)
- You define a [global minor mode](https://github.com/galaunay/poetry.el/blob/ed60dd8f4a60410b008ed60d8dc71d8df1517118/poetry.el#L423) but might it be possible for this to be a local minor mode so that different projects can use different venvs? Or is this a limitation of `pyvenv-activate` etc.
- Use `unless` instead of `when` + `not` in cases like [this](https://github.com/galaunay/poetry.el/blob/ed60dd8f4a60410b008ed60d8dc71d8df1517118/poetry.el#L520).
- You can use `(or x y)` instead of `(if x x y)` to avoid repetition, e.g. [here](https://github.com/galaunay/poetry.el/blob/ed60dd8f4a60410b008ed60d8dc71d8df1517118/poetry.el#L554).
- [This](https://github.com/galaunay/poetry.el/blob/ed60dd8f4a60410b008ed60d8dc71d8df1517118/poetry.el#L598-L603) could be much more neatly written as `(message "[%s] %s" (or (poetry-get-project-name) "Poetry") mess)`.

Hope that helps!

galaunay:
Hi, 
Thanks a lot for the feedback.
I addressed the points you both mentioned.

Regarding this one:
> - You define a global minor mode but might it be possible for this to be a local minor mode so that different projects can use different venvs? Or is this a limitation of pyvenv-activate etc.

I am afraid the documentation of this mode was not very clear (It is updated now).

Poetry automatically creates a virtualenv for every project (with all the dependencies installed). The idea of this minor mode is to automatically activate the relevant project virtualenv when visiting a project file. It then provides tailored, per-project virtualenvs.

In case I am still unclear (and because an example is better than hundred explanations). Let say I have a file `script1.py` that is part of `project1` (with an associated `virtualenv1`) and a file `script2.py`, part of `project2` (with `virtualenv2`):

- When I open `script1.py`, the virtualenv is set to `virtualenv1`.
- As soon as I switch to `script2.py`, the virtualenv is automatically set to `virtualenv2`.
- If I switch back to `script1.py`, the virtualenv is set to `virtualenv1` again.

riscy:
I'm using this with some success.  One thing I might do is make a constant out of the string `"*poetry*"` because it's repeated in a number of places throughout the codebase.  Or you could write a helper function to do this for you (I'll use this helper function in my next comment):
```elisp
(defun poetry--get-buffer-name (_mode)
  "Return the name of the default poetry buffer."
  "*poetry*")
```

One thing that strikes me about some of the poetry-call methods is how they hang Emacs, sometimes for several seconds at a time (at least on my machine).  A lot of work has gone into correctly handling background compilation in Emacs -- Have you considered *not* using the synchronous `call-process` and instead making a call to `compile`?  Let me show you an example:

```elisp
(defun poetry-call (command &optional args)
  "Call poetry COMMAND with the given ARGS."
  (unless (member command '(new update))
    (poetry-ensure-in-project))
  (let* ((prog (or (executable-find "poetry")
                   (poetry-error "Could not find 'poetry' executable")))
         (args (if (or (string= command "run")
                       (string= command "init"))
                   (cl-concatenate 'list (list (symbol-name command))
                                args)
                 (cl-concatenate 'list (list "-n" "--ansi"
                                          (symbol-name command))
                              args))))
    (let ((compilation-buffer-name-function #'poetry--get-buffer-name))
        (compile (concat prog " " (string-join args " "))))))  ; NOTE this requires 'subr-x
```
This is a little bit different from your method because it pops the compilation buffer up -- if you really wanted to, you could suppress that with a `save-window-excursion` -- but the user is free to continue working in Emacs.  The other nice thing is that it removes your xterm-colors dependency, some of the processing on special characters you have to do, gives keybindings for users to visit the various errors (if any), and it'll also ensure all buffers are saved before compilation.

I don't personally consider this a blocker, but I thought I'd see what you thought.

purcell:
Looks ready to merge, so going ahead, thanks folks!

galaunay:
Sorry, I should have gave you some feedback sooner (I was working on it).

Thanks @riscy, the compilation mode is indeed quite nice.
I didn't wanted to make the calls to poetry asynchroneous because of the added complexity (poetry is not working very well when calling it in parallel for the same project).
But I managed to find a way to do it, and it should be robust now.

I also added a few features regarding virtualenvs, updated the transient interface and fix a few issues.
In consequence the code changed quite a lot since my first PR (if it changes anything on your side).
I still made sure that the package follows Melpa specifications.

Thanks for the ideas and enhancements !

PR #6191, opened 2019-05-29
Add recipe for key-quiz

federicotdn:

Key Quiz is a game for GNU Emacs where the player must type in key sequences corresponding to different Emacs commands, chosen at random.

https://github.com/federicotdn/key-quiz

I am the package's maintainer.

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/key-quiz
+++ b/recipes/key-quiz
@@ -1,0 +1,1 @@
+(key-quiz :repo "federicotdn/key-quiz" :fetcher github)


alphapapa:
Cool!  A few notes:

* It isn't necessary to do `:group 'key-quiz` in each `defcustom`, as any defined after a `defgroup` inherit the group.
* It would be better to inherit from existing faces than to use specific colors.  That avoids conflicts with themes that could make text hard to read.
* You may find this shuffle function from `elfeed` interesting:
```el
(defun elfeed--shuffle (seq)
  "Destructively shuffle SEQ."
  (let ((n (length seq)))
    (prog1 seq  ; don't use dotimes result (bug#16206)
      (dotimes (i n)
        (cl-rotatef (elt seq i) (elt seq (+ i (cl-random (- n i)))))))))
```

purcell:
>     * You may find this shuffle function from `elfeed` interesting:

I was naively going to suggest something like `(seq-sort-by (lambda (_) (random)) #'<= seq)`. So i went ahead and benchmarked this, and `elfeed-shuffle` is already the fastest option. (It would be even faster if `reverse` were replaced with `nreverse`.) See benchmarks here: https://gist.github.com/purcell/34824f1b676e6188540cdf71c7cc9fc4

purcell:
`elfeed--shuffle` is faster when `random` is used instead of `cl-random`, which I think is equivalent. :-)

All of this benchmarking depends heavily on the sequence type: `length` is really expensive on `lists`, so my simple version above has the most consistent runtime, while `key-quiz--shuffle-list` and `elfeed--shuffle` are both much faster on vectors and much slower on lists.

purcell:
In any case, the package code looks good, so merging now... :-)

federicotdn:
@alphapapa @purcell Thanks for the feedback! I'll fix the stuff you've mentioned I soon as I can.

alphapapa:
@purcell Thanks Steve, that's very interesting.  It may be even worse than that, though, when GC is accounted for specifically.  I ran your benchmarks with `bench-multi-lexical` and it looks like `elfeed--shuffle` is the slowest on both types: https://gist.github.com/alphapapa/43b7ba264ca13e51c8e8583869fe3ffd

alphapapa:
BTW, maybe `faster-seq-sort-by--shuffle` should be added to `seq.el`.  What do you think, Steve?  :)

alphapapa:
BTW @skeeto you might find this interesting too.  :)

skeeto:
Thanks, @alphapapa.

A couple of things: Using sort with a random comparison function [will not result in a uniform shuffle](https://rufflewind.com/2016-09-16/random-comparator-sort). Some permutations are more likely than others. So comparing performance is comparing apples and oranges. `elfeed--shuffle` is Fisher–Yates, so its results are as uniform as the PRNG.

This is a histogram of the first element after of a 1,000 element sort, done 100,000 times. First the "sort" shuffle:

Random comparison function

Note the spikes showing bias. Here's a Fisher–Yates shuffle:



If I had enough patience to run this longer it would flatten out into a straight line. The first plot would not.

```el
(let ((hist (make-vector 1000 0)))
  (dotimes (_ 100000)
    (let* ((list (cl-loop for i from 0 to 999 collect i))
           (shuffle (shuffle-function list)))
      (cl-incf (aref hist (car shuffle)))))
  (with-temp-file "results.txt"
    (dotimes (i (length hist))
      (insert (format "%d %d\n" i (aref hist i))))))
```

Shuffling a linked list is always going to be inefficient because of the O(n) indexing. If performance is an issue, always use a vector, even if that means converting a list to a vector then back to a list afterwards. I intentionally wrote `elfeed--shuffle` to accept generic sequences so the caller gets to decide. In practice I always *do* use it to shuffle a list in Elfeed, but that list is always short enough that it doesn't actually matter (`elfeed-feeds`). (The point is just to break up patterns in people's feed lists just before fetching them.)

In the benchmarks, the GC results for `elfeed--shuffle` doesn't make sense since it doesn't allocate memory. The GC time should be zero. The rest of it makes sense, though: The biased shuffle is faster since it's not thorough. The three Fisher–Yates shuffles are about even.

purcell:
That's all super interesting, @skeeto. Thanks!

> If performance is an issue, always use a vector, even if that means converting a list to a vector then back to a list afterwards.

Yep, this is the real answer.

> (@alphapapa) BTW, maybe `faster-seq-sort-by--shuffle` should be added to `seq.el`. What do you think, Steve? :)

Well, a proper Fisher-Yates shuffle might be worth adding... :-)

alphapapa:
@skeeto Thanks, Chris.  I always learn a lot from you.  That seems like good material for your blog.  ;)

PR #6192, opened 2019-05-29
Add leaf recipe

conao3:
### Brief summary of what the package does
Hi!
I have developed yet another implementation of `use-package`. I know the functionality is duplicated, but my package has more advanced keywords and does more reasonable keyword processing. Adding keywords is also easy.

The following is a description of the repository.

> leaf.el is yet another use-package.
> 
> leaf solves the stress of using use-package for 2 years. By developing from scratch, we have a cleaner and more predictable implementation than use-package.
> 
> This makes it easy to maintain and add new keywords. (see leaf-keywords.el)
> 
> leaf has keywords almost identical to use-package, but some of usage of the keywords is different.
> 
> The quickest way to solve problem is to use macroexpand-1 to see the unfolded result if it is not what you intended. And also there are also a number of samples in this README and more in the test file.
> 
> In addition, since it works with Emacs-22, you can use your usual init.el as usual, even if you are temporarily using fossil-like Emacs on loan.
> 
> (Of course, there are not many packages that work perfectly with Emacs-22. These packages will be installed by the package manager and will probably report an error.
leaf processes this and cancels the configuration for that package. It then attempts to process the next package.)

### Direct link to the package repository

https://github.com/conao3/leaf.el

### Your association with the package

I'm an author and maintainer.

### Relevant communications with the upstream package maintainer

None.

### Checklist

Please confirm with `x`:

- [x] The package is released under a [GPL-Compatible Free Software License](https://www.gnu.org/licenses/license-list.en.html#GPLCompatibleLicenses).
- [x] I've read [CONTRIBUTING.org](https://github.com/melpa/melpa/blob/master/CONTRIBUTING.org)
- [ ] I've used the latest version of [package-lint](https://github.com/purcell/package-lint) to check for packaging issues, and addressed its feedback
- [x] My elisp byte-compiles cleanly
- [x] `M-x checkdoc` is happy with my docstrings
- [x] I've built and installed the package using the instructions in [CONTRIBUTING.org](https://github.com/melpa/melpa/blob/master/CONTRIBUTING.org)
- [ ] I have confirmed some of these without doing them

## Flycheck warnings
```
leaf.el   274  93 warning         `eval-after-load' is for use in configurations, and should rarely be used in packages. (emacs-lisp-package)
```
However, this line is written for user deployment and is not required for package operation.

Best,

PR #6193, opened 2019-05-29
add leaf-keywords recipe

conao3:

Corresponding #6192, this package provide additional `leaf` keywords for Emacs external packages.

There is room for thought about this package's dependencies, but it is a sort of transpiler and does not guarantee the viability of the conversion result.
Also, because the lines dependent on the external package are enclosed by `eval-after-load`, an error does not occur and does not prevent the user from executing init.el.

As a practical matter, I think that adding all external packages that this package supports to dependencies depends on the fact that many packages not needed by the user will be installed.

https://github.com/conao3/leaf-keywords.el

I'm an author and maintainer.

None

### Checklist

Please confirm with `x`:

## flycheck warnings
```
 leaf-k…    10  39 error           Package leaf is not installable. (emacs-lisp-package)
 leaf-k…    35   1 error           Cannot open load file: No such file or directory, leaf (emacs-lisp)
 leaf-k…    81  31 warning         `eval-after-load' is for use in configurations, and should rarely be used in packages. (emacs-lisp-package)
 leaf-k…   309   1 error           "leaf-key-chord" doesn't start with package's prefix "leaf-keywords". (emacs-lisp-package)
 leaf-k…   326   1 error           "leaf-key-chord*" doesn't start with package's prefix "leaf-keywords". (emacs-lisp-package)
 leaf-k…   331   1 error           "leaf-key-chords" doesn't start with package's prefix "leaf-keywords". (emacs-lisp-package)
 leaf-k…   405   1 error           "leaf-key-chords*" doesn't start with package's prefix "leaf-keywords". (emacs-lisp-package)
```
- Because `leaf` have not part MELPA, could not retreave leaf, the error will solve when leaf recipe merged.
- `eval-after-load` issue is same reason at `leaf`, the line is not for this package, for users.
- `leaf-key-*` function and macros extend `leaf-key` at `leaf`, I think the name is appropriate.

### DIFF
--- a/recipes/leaf-keywords
+++ b/recipes/leaf-keywords
@@ -1,0 +1,1 @@
+(leaf-keywords :fetcher github :repo "conao3/leaf-keywords.el")


riscy:
I've got my eye on your other PR, but in the meantime here's a quick look at `package-lint`:
```
52:1: error: "leaf-insert-list-after" doesn't start with package's prefix "leaf-keywords".
65:1: error: "leaf-insert-list-before" doesn't start with package's prefix "leaf-keywords".
112:33: warning: `eval-after-load' is for use in configurations, and should rarely be used in packages.
350:1: error: "leaf-key-chord" doesn't start with package's prefix "leaf-keywords".
367:1: error: "leaf-key-chord*" doesn't start with package's prefix "leaf-keywords".
372:1: error: "leaf-key-chords" doesn't start with package's prefix "leaf-keywords".
446:1: error: "leaf-key-chords*" doesn't start with package's prefix "leaf-keywords".
```
I don't have all the context, but should some of these have their prefix adjusted?

conao3:
This name is appropriate because the leaf-chord extends the [leaf-key](https://github.com/conao3/leaf.el/blob/d138ca9241d5de278b941a7e933f94dde6e72b10/leaf.el#L439-L508), but I cannot include it in the leaf because it relies on packages other than the Emacs standard attachments.

And I've already noticed that and written to ask the MELPA maintainer to take care of it.
> leaf-key-* function and macros extend leaf-key at leaf, I think the name is appropriate.

Please, thanks.

conao3:
Two weeks have passed since I submitted this and some packages that were submitted after me have been merged. Do @purcell have any comments about this package?

purcell:
> As a practical matter, I think that adding all external packages that this package supports to dependencies depends on the fact that many packages not needed by the user will be installed.

Yep, I agree with this. In its current form, this is a sort of "extra leaf stuff" package. Since it has no extra dependencies, there's no technical reason why it couldn't be included in the `leaf` recipe, but I appreciate you'd like to keep things separate. The best alternative to how you've set this up would be to publish a number of smaller packages, like `leaf-diminish.el`, but that could get unwieldy. For now, then, I'm happy to merge this as it is.

conao3:
Thanks!

PR #6194, opened 2019-05-30
Add helm-evil-markers recipe

xueeinstein:

Show evil markers with helm

https://github.com/xueeinstein/helm-evil-markers

I am an author and maintainer

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/helm-evil-markers
+++ b/recipes/helm-evil-markers
@@ -1,0 +1,3 @@
+(helm-evil-markers
+ :repo "xueeinstein/helm-evil-markers"
+ :fetcher github)


xueeinstein:
@riscy Thank you for your review! I updated the file header and fixed issues from `package-lint`. However, I don't know how to avoid the warnings for unused lexical args. Indeed, I need them as placeholders. How could I deal with this?

riscy:
@xueeinstein No worries - the way you can fix it is to put underscores before them, which'll tell Emacs and your users that you intend for the arguments to be unused:
```elisp

(defun helm-evil-markers-sort (candidates _source)
  "Custom sorting for matching CANDIDATES from SOURCE."
  (let ((pattern helm-pattern))
    (if (string= pattern "")
        candidates
      (sort candidates
            (lambda (_s1 s2)
              (if (string-prefix-p (format "%s> " pattern) (car s2))
                  nil
                t))))))
```

xueeinstein:
@riscy Thank you! I fixed it. Looking forward to merging this PR soon.

purcell:
Thanks. This mostly looks good, but you'll need to declare a dependency on the `evil` package. :-)

xueeinstein:
@purcell Sure, I just declared the dependency of `evil`.

dzop:
This looks pretty useful. I usually only use one marker at a time. I feel like I would use more of them if they are searchable like this.

Is there a reason why you copy the whole buffer string into a temporary buffer in `helm-evil-markers-update-alist`? Since all you are doing is going to the marker positions and extracting the lines, I believe you can just use `save-excursion` as a replacement for `with-temp-buffer` while also removing the `insert-buffer-substring` call.

riscy:
Sorry about the turnaround @xueeinstein.

I think @dzop has a good point, although, since the mode is behaving correctly I feel you could leave that as an issue to tackle later.

One thing I'm noticing is that setting keybindings like this
```elisp
(define-key evil-normal-state-map (kbd "'") 'helm-evil-markers)
(define-key evil-normal-state-map (kbd "m") 'helm-evil-markers-set)
```
could clobber a user's customizations, and they will have no way easy way to go back.  I've seen this handled a couple ways.  The best way is to create a minor mode with a corresponding keymap that can be toggled on and off.  I have also seen people create a function that loads a specific set of keybindings, which is also okay.

Hope that is helpful - let me know if you have any questions.

xueeinstein:
@dzop Thank you for your suggestion! I update the implementation following your advice. 

@riscy Evil uses `emulation-mode-maps` which has higher priority than minor mode maps. So I create a function to load required keybindings. Please check it.

riscy:
> Evil uses emulation-mode-maps which has higher priority than minor mode maps. So I create a function to load required keybindings. Please check it.

Thanks for letting me know about this!  I looked and tested it out and this is great.  The function `helm-evil-markers-toggle` really makes it a friendlier package (e.g. there might be people out there who have bound `'` to something else and use ``` ` ``` to navigate their marks).

PR #6196, opened 2019-05-30
Added Recipe for ruby-extra-mode

samuel-first:

The package provides a minor mode meant to be used along side the ruby major mode.  The minor mode gives the user the ability to easily run linters, pull up documentation, generate documentation, and prototype programs via an interactive ruby interpreter.

https://gitlab.com/samuelfirst/ruby-extra-mode

Maintainer

None needed

### Checklist

Please confirm with `x`:

### Notes

- Checkdoc doesn't like that some of my docstrings contain the word flag.  In those docstrings, flag does not refer to an internal boolean value, but rather to a flag meant to be passed to a command-line program.  It also does not like places where I reference ruby-lint, which it believes is a lisp symbol.  It actually refers to a ruby linting tool that the mode uses as the default linter.
- `-*- lexical-binding: t; -*-` is on the second line, rather than at the end of the first line.  From what I can tell from the documentation I can find, it shouldn't make a difference functionality-wise, however, putting it on the second line allows me to avoid breaking the 80-character limit.
- There are some complaints from `flycheck` about references to free variables defined with `setq`.  I believe that the magic `-*- lexical-binding: t; -*-` comment should fix this, although, I am new to elisp, so if I'm missing something here, or if there is a cleaner way to handle this, I am willing to change these parts of the script.
- `Flycheck` complains about some functions that are not known to be defined.  One of those functions, `markdown-mode` is not a hard requirement, but rather something that is used only if it is available.  The other `eshell-send-input` is defined when `eshell` starts, so it shouldn't be a problem.

### DIFF
--- a/recipes/ruby-extra-mode
+++ b/recipes/ruby-extra-mode
@@ -1,0 +1,1 @@
+(ruby-extra-mode :fetcher gitlab :repo "samuelfirst/ruby-extra-mode")


riscy:
A quick pass over the MELPA checklist (semi-automated, so please forgive mistakes):

`M-x byte-compile-file`:
```
In ruby-extra-lint:
ruby-extra-mode.el:108:48:Warning: assignment to free variable ‘file-name’
ruby-extra-mode.el:108:26:Warning: reference to free variable ‘file-name’

In ruby-extra-lookup-doc:
ruby-extra-mode.el:134:45:Warning: assignment to free variable ‘word’
ruby-extra-mode.el:135:45:Warning: reference to free variable ‘word’

In ruby-extra-generate-doc:
ruby-extra-mode.el:160:55:Warning: assignment to free variable ‘dir’
ruby-extra-mode.el:160:26:Warning: reference to free variable ‘dir’

In ruby-extra-interactive-run:
ruby-extra-mode.el:187:11:Warning: assignment to free variable ‘run-string’
ruby-extra-mode.el:186:4:Warning: reference to free variable ‘run-string’
```

`M-x checkdoc`:
```
Warning (emacs): 
ruby-extra-mode.el:62: Flag variable doc strings should usually start: Non-nil means
Warning (emacs): 
ruby-extra-mode.el:68: Flag variable doc strings should usually start: Non-nil means
```

`M-x package-lint`:
No issues found.

----------
Not on the MELPA checklist, but hopefully helpful --

`M-x check-declare-file`:
No issues found.

Miscellaneous comments/suggestions:
- You actually _do_ need to have `-*- lexical-binding: t; -*-` on the first line, not on the second.  Unfortunately it tends to make everyone's first line long -- oh well. :)
- It's better to use `let` instead of setq in many cases that the compiler's complaining about.  So, for example, in `ruby-extra-lint`, you can write:
  ```elisp
  (let ((file-name (buffer-file-name)))
    (shell-command (concat ruby-extra-linter " " file-name)
          	 "*ruby-extra-lint*"
          	 "*ruby-extra-error*")
  ```
  but, I think, in this particular instance, you don't even need to define a second variable -- I believe you could just replace `file-name` with `(buffer-file-name)`
- The "flag" complaints are because the variable ends in the word `flag` (so there is some existing Emacs-internal nomenclature around variables that end in this word).  Perhaps you could rename the variables from `ruby-extra-interactive-path-flag` to e.g. `ruby-extra-interactive-path-arg`?
- I figure can use `declare-function` to pacify the compiler around `markdown-mode`:
  ```elisp
  (declare-function markdown-mode "ext:markdown-mode.el" nil t)
  ```
  but the better way to check if markdown-mode is available is to check `(featurep 'markdown-mode)` instead of wrapping in `condition-case`.  Similar with `eshell-send-input`.

samuel-first:
Thanks for your help!

I've made the following changes:

* Changed `setq` to `let`.
* Changed from `C-x` keymaps to avoid overwriting other commands.
* Changed names of variables containing the word 'flag'.
* Fixed some of the quotes in my docstrings

I can't seem to get `featurep` to work properly when I test it in the scratch buffer with the
lines below, so I've gone with the `declare-function` solution.  If I'm missing something here,
let me know.

```
(if (featurep 'markdown-mode) (message "t") (message "nil"))
(if (featurep 'eshell-send-input) (message "t") (message "nil"))
```

I also have a question about how the package manager works:  If I run `M-x package-install-file` over the file, will it install it the same way it would from the melpa?

purcell:
Thanks! Some notes:

- The `defgroup` can't be [its own child](https://gitlab.com/samuelfirst/ruby-extra-mode/blob/a4854a83f324b00d0550d818efc9479a03726bff/ruby-extra-mode.el#L33).
- Not every buffer has a `buffer-file-name`, so [this](https://gitlab.com/samuelfirst/ruby-extra-mode/blob/a4854a83f324b00d0550d818efc9479a03726bff/ruby-extra-mode.el#L109) will explode then. Additionally - and very importantly - this code will not work anyway if the file name contains spaces or certain other characters. Use `shell-quote-argument` to avoid this. This same feedback applies to other parts of the code.
- The recurring code like [this](https://gitlab.com/samuelfirst/ruby-extra-mode/blob/a4854a83f324b00d0550d818efc9479a03726bff/ruby-extra-mode.el#L132-142) is indented incorrectly: the `let` form should clearly enclose its body, ie. the body should be indended. Try enabling `aggressive-indent-mode` when editing elisp to avoid this.
- Prompt strings must include a space after the colon, e.g. [here](https://gitlab.com/samuelfirst/ruby-extra-mode/blob/a4854a83f324b00d0550d818efc9479a03726bff/ruby-extra-mode.el#L157).
- The `\` in the [key sequences here](https://gitlab.com/samuelfirst/ruby-extra-mode/blob/a4854a83f324b00d0550d818efc9479a03726bff/ruby-extra-mode.el#L208) are unnecessary. `kbd` understands `C-` to mean "control".
- Don't repeat the keybindings manually [in the docstring](https://gitlab.com/samuelfirst/ruby-extra-mode/blob/a4854a83f324b00d0550d818efc9479a03726bff/ruby-extra-mode.el#L202-205) in case the user rebinds them. Just use the `\\[]` form, or the form that documents the entire keymap automatically (I forget what it looks like, sorry). Related: [this comment](https://gitlab.com/samuelfirst/ruby-extra-mode/blob/a4854a83f324b00d0550d818efc9479a03726bff/ruby-extra-mode.el#L195-198) is redundant: most of your top-level comments should be removed because the docstrings are sufficient.
- The parent mode of [these major modes](https://gitlab.com/samuelfirst/ruby-extra-mode/blob/a4854a83f324b00d0550d818efc9479a03726bff/ruby-extra-mode.el#L218-231) should probably be `special-mode`, not `fundamental-mode`.
- [This code](https://gitlab.com/samuelfirst/ruby-extra-mode/blob/a4854a83f324b00d0550d818efc9479a03726bff/ruby-extra-mode.el#L242-243) auto-enables the package as soon as it is installed. This is not allowed. Just like built-in Emacs packages, it should be possible to have the package installed but not enabled. Instead, tell the user to add this code to their config.

Hope that helps!

samuel-first:
This is awesome, thanks for all the help, guys!

Changes in most [recent commit](https://gitlab.com/samuelfirst/ruby-extra-mode/tree/8550bc7bc1cf1d104979c3a4263fe1feabbe3d74):

- Changed `ruby-extra-custom`'s group to `'programming`
- Fixed `let` indentation
- Wrapped `buffer-file-name` with the function `safe-buffer-file-name`, which will throw an error if `buffer-file-name` returns `nil`.
- Used `shell-quote-argument` before passing strings to `shell-command`
- Added space after ':' in prompt string
- Removed superfluous comments
- Removed unnecessary back-slashes from keybindings
- Made keymap in the docstring dynamic.  This one was a bit tricky to figure out, but after looking through some of the source of `org-mode`, I've determined that if you've created the mode with `define-minor-mode` you can use `\\{<mode-name>-map}` to create the keymap.
- Made the `ruby-mode` hook toggle-able via a customization variable (off by default).  I think this provides a nice middle-ground and will make it easier for less experienced users to enable the hook, however, I am willing to remove it entirely if you think that would be better.
- Changed derived modes from `fundamental-mode` to `special-mode`.

Thanks again, @purcell and @riscy!

purcell:
>     * Made the `ruby-mode` hook toggle-able via a customization variable (off by default).  I think this provides a nice middle-ground and will make it easier for less experienced users to enable the hook, however, I am willing to remove it entirely if you think that would be better.

Yes, that must be removed, and I'll explain why: there's no guarantee that the autoloads won't be completely evaluated before the custom file is loaded. So making this a custom var doesn't help at all, because load order might mess things up and prove very confusing for novices and experts alike. But in fact a custom var for this behaviour already exists: it's called `ruby-mode-hook`. ;-) 

A few more notes, then:

- Use `with-demoted-errors` [here](https://gitlab.com/samuelfirst/ruby-extra-mode/blob/8550bc7bc1cf1d104979c3a4263fe1feabbe3d74/ruby-extra-mode.el#L158)
- Use `or` instead of a repetitive `if` [here](https://gitlab.com/samuelfirst/ruby-extra-mode/blob/8550bc7bc1cf1d104979c3a4263fe1feabbe3d74/ruby-extra-mode.el#L175-177)
- Why isn't [this](https://gitlab.com/samuelfirst/ruby-extra-mode/blob/8550bc7bc1cf1d104979c3a4263fe1feabbe3d74/ruby-extra-mode.el#L99-112) just `(or (buffer-file-name) (error "This buffer has no file."))`? It is unnecessarily complicated afaict, and the specific error symbol combined with empty `catch` forms is unidiomatic. *Also*, the unprefixed `safe-buffer-file-name` symbol obviously lives in the global namespace, which is a no-no.
- Rather than using `shell-command` for [linting](https://gitlab.com/samuelfirst/ruby-extra-mode/blob/8550bc7bc1cf1d104979c3a4263fe1feabbe3d74/ruby-extra-mode.el#L116-136), you should instead use `compile`, and make `ruby-extra-lint-mode` a compilation mode (see `define-compilation-mode`) with its own regexps for errors. This way, you'll get the standard functionality for clicking on errors, navigating from one to the next etc. Alternatively, create a separate `rubylint`, `flycheck-rubylint` or `flymake-rubylint` package, which would arguably benefit more users: more modular is better.
- The use of eshell [here](https://gitlab.com/samuelfirst/ruby-extra-mode/blob/8550bc7bc1cf1d104979c3a4263fe1feabbe3d74/ruby-extra-mode.el#L198-212) seems unusual and unnecessary. This would normally be handled with `comint` instead, but in this case I think the `inf-ruby` package has you covered anyway, so this functionality is probably superfluous.
- Rather than using `switch-to-buffer-other-window`, e.g. [here](https://gitlab.com/samuelfirst/ruby-extra-mode/blob/8550bc7bc1cf1d104979c3a4263fe1feabbe3d74/ruby-extra-mode.el#L157), you should prefer `display-buffer`. This can be customized as needed. And rather than `read-only-mode`, you can enable `special-mode`, so that `q` works as expected in that buffer.

I'd note that give some of the observations above, this increasingly looks like a couple of personal helper functions rather than a proper minor mode. Given `inf-ruby`, you could make a little package for the `ri`-with-markdown-rendering utils, plus a little `rubylint` or `flycheck-rubylint`, and then this whole package would vanish in a positive way.

samuel-first:
@purcell 

> But in fact a custom var for this behaviour already exists: it's called ruby-mode-hook. ;-)

Haha, fair enough, I've removed the customization variable and updated the mode's docstring to reflect that.

I've implemented a few other changes you recommended in [this commit](https://gitlab.com/samuelfirst/ruby-extra-mode/tree/d80847890f69cc0930f79e26f812f5517cf096aa):
- Used `with-demoted-errors` instead of `condition-case`
- Used `or` [here](https://gitlab.com/samuelfirst/ruby-extra-mode/blob/d80847890f69cc0930f79e26f812f5517cf096aa/ruby-extra-mode.el#L97) and [here](https://gitlab.com/samuelfirst/ruby-extra-mode/blob/d80847890f69cc0930f79e26f812f5517cf096aa/ruby-extra-mode.el#L144).
- Removed the `interactive-mode` feature.  The original intent wasn't to have it operate in the same way as a repl, however, I now realize that it offers no real advantage (and several disadvantages) over the `inf-ruby` package.
- Used `display-buffer` and `special-mode` instead of `switch-to-buffer-other-window` and `read-only-mode`.  Is there a way to combine major modes?  Because I'd really like to use a combination of `special-mode` and `markdown-mode` [here](https://gitlab.com/samuelfirst/ruby-extra-mode/blob/d80847890f69cc0930f79e26f812f5517cf096aa/ruby-extra-mode.el#L126), rather than load `special-mode` when `markdown-mode` isn't available.

I'm not quite sure how to properly use `define-compilation-mode`.  Is what I have below correct?

```
(define-compilation-mode ruby-lint "ruby lint"
  "docstring"
  (setq ruby-lint-error-regexp-alist
	'("\\(.*\\):.*\\([[:digit:]]+\\).*\\([[:digit:]]+\\).*$") 1 2 3))
```

> Given inf-ruby, you could make a little package for the ri-with-markdown-rendering utils, plus a little rubylint or flycheck-rubylint, and then this whole package would vanish in a positive way.

This sounds interesting, could you elaborate a bit on how I would go about this?

purcell:
> > Given inf-ruby, you could make a little package for the ri-with-markdown-rendering utils, plus a little rubylint or flycheck-rubylint, and then this whole package would vanish in a positive way.
> 
> This sounds interesting, could you elaborate a bit on how I would go about this?

You'd make a "ri-markdown.el" package, with its own recipe. Then you'd make a "rubylint.el" package or "flycheck-rubylint.el" (or both!). The combination of those 2 (or 3) packages with the pre-existing "inf-ruby" would cover everything that's in this package, but be more widely useful.

riscy:
I'll wait to hear some feedback from the author on this one, since @purcell makes a good point.

DamienCassou:

PR #6197, opened 2019-05-31
Added region-occurrences-highlighter recipe

alvarogonzalezsotillo:

This minor mode marks all the occurrences of the current region (selection), like VSCode does.

https://github.com/alvarogonzalezsotillo/region-occurrences-highlighter

I am the author and the only contributor.

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/region-occurrences-highlighter
+++ b/recipes/region-occurrences-highlighter
@@ -1,0 +1,2 @@
+(region-occurrences-highlighter :repo "alvarogonzalezsotillo/region-occurrences-highlighter" :fetcher github)
+


riscy:
Hi!  I tested this out and it works great for me.  Some miscellaneous comments/suggestions:
- 4: If possible, prefer `https://` to `http://`
- 22: If possible, prefer `https://` to `http://`
- The double-hyphen in region-occurrences-highlighter--group is probably not necessary -- usually this would signify a "private" function or variable that users/developers shouldn't rely on, but a _group_ should definitely be part of your mode's public interface (e.g. users will want to write `M-x customize-group region...group`)
- Similarly, customizable variables (like region-...max-size) probably shouldn't be "private"
- I believe specifying the `:group` on these customizables isn't strictly necessary -- once you've defined the group earlier in the file, they're automatically a part of it.
- Your `;;;###autoload` line has a leading space before it
- You should add a LICENSE or COPYING file to the repository (this will let GitHub add a nice icon that makes it easier for users to know what they're using).  I think the summarized license in your source file is a little bit too informal.

alvarogonzalezsotillo:
I think I have addressed all your comments in the commit https://github.com/alvarogonzalezsotillo/region-occurrences-highlighter/commit/a781ac6dafa72350c2226c00ae6b768e07008afa

purcell:
Thanks, this mostly looks good, but one thing that stands out is that [this line](https://github.com/alvarogonzalezsotillo/region-occurrences-highlighter/blob/master/region-occurrences-highlighter.el#L75) should be redundant, and the reason you found it isn't is because you are failing to pass the `local` flag to `add-hook`/`remove-hook` [here](https://github.com/alvarogonzalezsotillo/region-occurrences-highlighter/blob/master/region-occurrences-highlighter.el#L63-L65). The current code is effectively a non-global minor mode which modifies the global version of `post-command-hook`. :-)

Hope that helps!

alvarogonzalezsotillo:
Solved in https://github.com/alvarogonzalezsotillo/region-occurrences-highlighter/commit/07be4b8026f6a19ef3811bfe7fdc49e7e4909fb4

I hoped I could manage to publish a recipe before anyone noticed I have no idea about elisp  :)

purcell:
Yeah, so now you can (and should) remove that redundant line as noted above. Merging in any case: welcome to MELPA! :tada:

PR #6199, opened 2019-06-01
Add ddskk-posframe recipe

conao3:

Show Henkan tooltip for [ddskk](https://github.com/skk-dev/ddskk) via [posframe](https://github.com/tumashu/posframe).

https://github.com/conao3/ddskk-posframe.el

I'm an author and maintainer.

None.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/ddskk-posframe
+++ b/recipes/ddskk-posframe
@@ -1,0 +1,1 @@
+(ddskk-posframe :fetcher github :repo "conao3/ddskk-posframe.el")


riscy:
Just a very quick look over this -- most checks pass but I'm seeing in `M-x byte-compile-file`:
```
In end of data:
ddskk-posframe.el:246:1:Warning: the function
    ‘ddskk-posframe-display-at-point’ is not known to be defined.
```

conao3:
Fixed!

conao3:
@riscy Thanks for your work! I fixed your point, do you have another comment?

riscy:
Your elisp is super nice, btw.  I've had a chance to look at this a little more deeply.

A couple very minor comments on the frontmatter:
- `ddskk-posframe.el provide Henkan ...` should be `ddskk-posframe.el provides`
- `You should only to enable ...` should be `You only need to enable ...`

And some other comments as I look this over:
- You don't need to specify a `:group` for every customizable variable that follows your `defgroup` -- they'll automatically become a part of `ddskk-posframe`.
- More importantly, ddsk-posframe's group (`defgroup ddskk-posframe`) should probably belong to `completion`, not `lisp`
- `ddskk-posframe-style`'s default value is `'point` which is a symbol, but the `:type` is `'string` (you can set it to `:type 'symbol`)
- Similarly for `ddskk-posframe-font`, whose default value is `nil` but the `:type` is `'string` (perhaps you can set it to `:type '(choice (const :tag "Off" nil) string)`).
  - Similarly for `ddskk-posframe-height`, `ddskk-posframe-min-width`, `ddskk-posframe-min-height`, and `ddskk-posframe-parameters`
- I notice you brought in some customization variables from `ivy-posframe` and haven't updated the comments (for example, you refer to `ivy-min-posframe`); these should probably be updated to avoid confusion.
- I'm looking at `defface ddskk-posframe-border` and CONTRIBUTING.md -- is there any way around this?:
  >>> Please avoid defining a face that both :inherit another face and also override their attributes (e.g. by making them bold, underlined or inverse-video). The result could be really bad depending on user customizations. The best approach here is to simply :inherit the faces, and leave the user to customise the rest.

conao3:
First, thank you for your detailed review.
However, I feel pointing out really minor errors on English is more than the job of MELPA maintainer.
This cares about you. Are you sure you want to review all future packages with the quality of this review?
Anyway, thank you for your review. Now I have corrected most of the points.

---

The non-English fixes (As you pointed out that the word ivy remains.) are due to me inheriting many implementations of this package from `ivy-posframe`.
I think `ivy-posframe` has the same fixes, but what do you think about the fact that many of the packages in MELPA don't exceed your standards, even though there are gatekeepers that go into these details and review them when they are included in MELPA?

https://github.com/tumashu/ivy-posframe/blob/master/ivy-posframe.el#L150-L205

---

As for the last point, I'm not familiar with Emacs Face, so I don't have an effective solution. How should I correct it? At the same time, it points to `ivy-posframe`.

https://github.com/tumashu/ivy-posframe/blob/master/ivy-posframe.el#L222-L230

Thanks,

riscy:
> what do you think about the fact that many of the packages in MELPA don't exceed your standards, even though there are gatekeepers that go into these details and review them when they are included in MELPA?

Honestly, some of my own code on MELPA has slipped below standard -- I learn as I go.  But beyond just making sure code meets some minimum standard, I view the MELPA pull requests as a kind of open forum about how to write good (or at least better) Emacs packages.  I've learned a lot by looking at what other maintainers have had to say.

I have also only relatively recently been able to merge to MELPA, so I'd rather err on the side of pedantic, at least for now.  :)

> How should I correct it? At the same time, it points to ivy-posframe.

I may have to yield to another reviewer with more expertise, but in this case it may be best to inherit from `fringe` or possibly `highlight`?

conao3:
Thank you for answering my captious question. I am convinced that you will be a major player in MELPA.

I don't have the experience to make that judgment about the face. I would like to wait until there is another maintainer's comment.

Thanks,

riscy:
Really, I think since the face in question is just a border face, and since this is just inheriting ivy-posframe's behavior, this is good to go.

conao3:
Good. thanks!

PR #6203, opened 2019-06-05
Add nnreddit recipe

dickmao:

A Gnus backend for Reddit.

https://github.com/dickmao/nnreddit

Maintainer

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/nnreddit
+++ b/recipes/nnreddit
@@ -1,0 +1,4 @@
+(nnreddit
+  :fetcher github
+  :repo "dickmao/nnreddit"
+  :files ("lisp/*.el" "setup.py" "requirements.txt" "nnreddit"))
--- a/recipes/nnreddit
+++ b/recipes/nnreddit
@@ -1,4 +1,4 @@
 (nnreddit
   :fetcher github
   :repo "dickmao/nnreddit"
-  :files ("lisp/*.el" "setup.py" "requirements.txt" "nnreddit"))
+  :files ("lisp/*.el" "setup.py" "nnreddit"))


dickmao:
https://github.com/dickmao/nnreddit/commit/4bb118c6afd8fbe15537f991e5adadddef82478e 
addresses all except:

1. M-x checkdoc
2. Variables starting with `*`, e.g., `*my-global-var*`

Feedback regarding negotiability of the above is requested.

riscy:
Thanks!  I think I'm still seeing:
```
In nnreddit-rpc-get:
nnreddit.el:668:52:Warning: reference to free variable
    ‘nnreddit-python-command’
```
during byte compiling which you can fix with a `(defvar nnredit-python-command)` at the top of the file.

To be honest I have never seen the `*<var>*` style of creating variables before -- package-lint is obviously unhappy, but maybe you can describe how they're needed/help?

dickmao:
I shall yield to [the elisp conventions:](https://www.gnu.org/software/emacs/manual/html_node/elisp/Coding-Conventions.html)

> In some other systems there is a convention of choosing variable names that begin and end with ‘*’. We don't use that convention in Emacs Lisp, so please don't use it in your programs.

Those "other systems" included clisp.

Please respond with the incantation that yields
```
In nnreddit-rpc-get:
nnreddit.el:668:52:Warning: reference to free variable
    ‘nnreddit-python-command’
```
provided you can easily extract it from whatever larger script you're running.  I've fixed this and the naming issue in https://github.com/dickmao/nnreddit/commit/40d8978aacf61e073393501deb92783e8e3a2d67, but want to add your incantation to my `make test`.  I can't seem to get the warning with my current methods.

riscy:
As I look at it now, that warning (re: nnreddit-python-command) was not your fault - I had commented out a couple of lines so that the checkers would run in my sandbox and got sloppy.  Sorry about that mistake -- I hope it didn't waste too much time.

As for giving the checkdoc stuff a pass, we could wait for someone with more experience to see this and make that call.  Or, if you find the time to fix them, feel free to ping me for another look.

dickmao:
> As for giving the checkdoc stuff a pass

I shall yield to `M-x checkdoc`, which apparently is based on the [official elisp documentation](https://www.gnu.org/software/emacs/manual/html_node/elisp/Documentation-Tips.html#Documentation-Tips) (and not some arbitrarily draconian ruleset).  I fixed all the checkdoc warnings in https://github.com/dickmao/nnreddit/commit/36d93d626932939b11a072888f9b225c84216da8 except "Messages should start with capital letter", which afaict does not appear in the conventions.

I prefer to emit messages prefixed with the function name, e.g., "nnreddit-rpc-post: invalid url %s", which does not start with a capital letter.

riscy:
Thanks for your patience on this!  I've had a chance to look into this a bit better.

- It may be worth considering what the upgrade path will be if, for example, you needed your users to upgrade `praw` (which is already on version 6.3).  Perhaps in the future one option would be to make `AuthenticatedReddit` a pypi package that you could run a `pip install --upgrade` on - thoughts?  Right now this hybrid code feels difficult to maintain.
  - Some minor Python stuff - PEP8 would say the `AuthenticatedReddit` module should have a lowercase name, e.g. (`authenticated_reddit.py`)
  - Some more minor Python stuff - the backslashes [here](https://github.com/dickmao/nnreddit/blob/master/nnreddit/AuthenticatedReddit.py#L74) aren't necessary.
- In your eval-and-compile you try to detect an OS-appropriate version of python, but then [here](https://github.com/dickmao/nnreddit/blob/master/lisp/nnreddit.el#L63) you end up calling `python` anyway.  If our system type _did_ end up being `windows-nt`, does this syntax still work with the command interpreter? `"cd %s && python setup.py install"` (I don't actually know).
- Could you combine the defvar [here](https://github.com/dickmao/nnreddit/blob/master/lisp/nnreddit.el#L90) with the variable's eventual definition [here](https://github.com/dickmao/nnreddit/blob/master/lisp/nnreddit.el#L124)?  e.g.,
  ```elisp
  (defvar nnreddit-summary-mode-map
    (let ((map (make-sparse-keymap)))
      (define-key map "r" 'gnus-summary-followup)
    ...
  ```
- My understanding is that you shouldn't auto-enable your mode or parts of it when someone simply loads your library, e.g. [here](https://github.com/dickmao/nnreddit/blob/master/lisp/nnreddit.el#L818).  Maybe this should be a global minor mode that can be toggled off and on?
  ```elisp
  ;;;###autoload
  (define-minor-mode nnreddit-mode
    "Toggle nnreddit-mode on/off."
    :global t
    (if nnreddit-mode
       (progn (remove-hook ...
     ;; activate!
     (add-hook 'gnus-article-mode-hook 'nnreddit-article-mode-activate))
  ```

dickmao:
> Thanks for your patience on this! I've had a chance to look into this a bit better.

I am generally impressed with the thoughtfulness of the vetting.  Granted a lot of melpa packages can "go rogue" in code quality as soon as they pass the initial inspection.

Requested changes in https://github.com/dickmao/nnreddit/commit/08d68551f93fe6a80f9813be08a83da76eed0215 except the following:

> * It may be worth considering what the upgrade path will be if, for example, you needed your users to upgrade `praw` (which is already on version 6.3).  Perhaps in the future one option would be to make `AuthenticatedReddit` a pypi package that you could run a `pip install --upgrade` on - thoughts?

As you might be aware, I lock praw to 6.2.0 in a venv.  Should 6.2.0 become incompatible with reddit, I will change `requirements.txt`, ensure `make test` succeeds, and re-release.  `authenticated_reddit` should not live in pypi.  It has no value outside `nnreddit`.

> * In your eval-and-compile you try to detect an OS-appropriate version of python, but then [here](https://github.com/dickmao/nnreddit/blob/master/lisp/nnreddit.el#L63) you end up calling `python` anyway. 

Correct, and in the venv, the wacky executable name, e.g., `pythonw`, should get normalized to `python`.

> * My understanding is that you shouldn't auto-enable your mode or parts of it when someone simply loads your library, e.g. [here](https://github.com/dickmao/nnreddit/blob/master/lisp/nnreddit.el#L818).  Maybe this should be a global minor mode that can be toggled off and on?

I am not aware of any existing Gnus backends defining global minor modes in this way.  The "deal" is that if the user is so bold as to include `nnreddit` as a `gnus-secondary-select-method`, then I will load the nnreddit code and I will augment various keymaps if the selected newsgroup comes from reddit (see `nnreddit--gate`).  I think this best accommodates "do what I mean" without undue and onerous feature-flagging.  At least one precedent exists for this in `nnmairix.el`.

riscy:
Thanks for clarifying the details around pythonw :)

> As you might be aware, I lock praw to 6.2.0 in a venv. Should 6.2.0 become incompatible with reddit, I will change requirements.txt, ensure make test succeeds, and re-release. authenticated_reddit should not live in pypi. It has no value outside nnreddit.

I think the issue is that the _only_ time the user installs the requirements is when the [nnreddit virtual env didn't initially exist](https://github.com/dickmao/nnreddit/blob/master/lisp/nnreddit.el#L57).  This will only happen once.  Therefore, even if you re-release, your user may still be stuck with the old requirements.  Maybe another option is to change the name of the venv when you need a new set of requirements to be installed (e.g. nnreddit_v1, nnreddit_v2, etc.) -- but whatever upgrade path you choose should take this into consideration.

Re: the Python changes,  per PEP8 the file should be `authenticated_reddit`, but the class names should still be `CapitalizedCamelCase` e.g. [here](https://github.com/dickmao/nnreddit/commit/08d68551f93fe6a80f9813be08a83da76eed0215#diff-630bd4ce9733684dc38ea739d5bf987aR55).  This is fairly minor (you could run pylint on the file) but is easy to point out.

I'll defer to nnmairix's precedent - thanks for the pointer.

riscy:
Ping me and let me know what you think, then I think we're good to go.

dickmao:
> This will only happen once. Therefore, even if you re-release, your user may still be stuck with the old requirements. 

Gee, you're right and I hadn't thought this through.

I now create a venv per 2-tuple of melpa directory and python executable, e.g.,. nnreddit-20190621.1052-python3.  I also got rid of the eval-and-compile.  I do all the python installing at runtime now.  Thanks to your input, things seem much safer and straightforward, if not bug-free.

I also address the PEP8 class naming in https://github.com/dickmao/nnreddit/commit/08d68551f93fe6a80f9813be08a83da76eed0215

riscy:
Cool!  Just a couple more things:

Did you really do the paperwork to sign your copyright to Free Software Foundation, Inc.?  Otherwise, the Copyright is yours and Paul Issartel's.

More importantly, I'm now able to see that this is a fork of a project that appears to at least have some users.  Have you tried to get in touch with Paul using the email on his website, https://paul-issartel.fr/research/ ?  It has been twenty days since you wrote an issue on his repository, but it would be _ideal_ if he could archive his repository (assuming his version of nnreddit no longer works) in order to avoid user confusion.

dickmao:
How do I sign the paperwork?

riscy:
> How do I sign the paperwork?

https://www.gnu.org/prep/maintain/html_node/Copyright-Papers.html

dickmao:
> Did you really do the paperwork to sign your copyright to Free Software Foundation, Inc.?

Is this recipe specifically asked this question because it adds an author who is not party to its inclusion into melpa?

I am vaguely familiar with the logistic hassle of "assignments" to ensure the sanctity of GPL.  It's not something I want to pursue.

The package does not share any meaningful code with github/paul-issartel (I did run `comm` as a minimally robust comparison), although as mentioned in http://github.com/paul-issartel/nnreddit/issues/9, I do reuse an arguably non-trivial message-flow idea.

I don't mind waiting a week for Paul's input, but I also wouldn't mind taking out his authorship credit which I extended only as a courtesy in order to avoid the assignments business.

riscy:
> Is this recipe specifically asked this question because it adds an author who is not party to its inclusion into melpa?

No, it's this copyright line [here](https://github.com/dickmao/nnreddit/blob/bbd1baa28b6bb2625072b730ac0e1b18cadeb9ea/lisp/nnreddit.el#L3).  If you haven't done the paperwork, your copyright should look more like ";; Copyright (C) 2019  <author names>"

dickmao:
Now I see!  Copyright line has been corrected in
https://github.com/dickmao/nnreddit/commit/5bf314ec968fb80a8f14028060a96e71f0590e43

paul-issartel:
Hi, sorry for the delay. I replied to @dickmao via email, but to summarize my answer here:  I don't see any actual code reuse in this package, so I don't think this is a fork of my version, more like a rewrite from scratch.

Thus, I don't believe I should be credited nor have to sign any paperwork for this package.

Since I stopped development of my own nnreddit package a long time ago, and left it in an unfinished state, I agree with @dickmao that his version should be the one put forward. So I made my own repository private to avoid confusion. Though there are still a number of forks of my repo on GitHub that are still called "nnreddit".

Still, since my package is easier to install and set up than @dickmao's version (it has no dependencies), it might make sense to make it available again, only with a different name.

purcell:
This is a nice piece of work: thanks, including for your patience and for going through the code and ensuring it follows the Elisp conventions upon which checkdoc and package-lint are based. It really makes a difference to ease of contributions across the ecosystem when everyone's code follows the community conventions.

Anyhow, merging now. My only remaining note about the package now is that you might consider removing the `-pkg.el` file, and listing the package dependencies directly in the `Package-Requires` header of `nreddit.el`. This saves a bit of duplication and indirection.

Welcome to MELPA! :tada:

dickmao:
Thanks.  The conventional wisdom does seem to deprecate `-pkg.el`.  But I use `cask package` to allow users to install directly from the git repo (and it complains when `-pkg.el` is missing).  I am rather fond of `-pkg.el` as I pull some other tricks with the `define-package` call therein.  I do agree about the duplication and indirection.

PR #6206, opened 2019-06-07
Add org-shoplist

lordnik22:

It handles org-headings as recipes (food-recipes) and collects there ingredients to shopping-lists. It's possible to compose recipes by having a base recipe with flavours (for example: cake and add different flavors:  Strawberry-Cake, Chocolate-Cake, Lemon-Cake, etc.)
You can set a factor as property and scale up the amounts of portions/peoples. 
You can have different formattings for your shoplist (provided are as-todo-list and as-table, but the user also can pass there own formatting-function).

https://github.com/lordnik22/org-shoplist

maintainer

**None needed**

### Checklist

Please confirm with `x`:

- [X] I have confirmed some of these without doing them (i didn't checked if i have the LATEST package-lint and thought that my current version has done there job good enough)

### DIFF
--- a/recipes/org-shoplist
+++ b/recipes/org-shoplist
@@ -1,0 +1,1 @@
+(org-shoplist :fetcher github :repo "lordnik22/org-shoplist")


riscy:
I'll add that the idea for this package really made me smile.  Please ping if you have any questions about these, I'd be happy to help.

lordnik22:
Thanks for the fast reply and review, i'm glad you like the idea of the package. I went through the code and made changes as asked to do so. 
One thing i didn't change is with the eval-after-load and injection of the custom units. I could use a constant storing a higher-order function and eval it in my magic function: ´org-shoplist-ing--transform-amount`, but i'm not quite convinced about that solution.

Because it's my first package, I would be interested in your opinion about the code as such.
Is there a more "elisp-ly" way in doing the job? Are there some bad habits? Principles i missed?

riscy:
Glad to!  Here are a couple more thoughts now that I'm looking more closely --

- You probably haven't signed your copyright to the FSF yet [here](https://github.com/lordnik22/org-shoplist/blob/master/org-shoplist.el#L3).
- You don't need to add quotes around parameters, simply capitalizing them is enough for Emacs' help system to recognize them as variables, e.g. [here](https://github.com/lordnik22/org-shoplist/blob/master/org-shoplist.el#L127)
- I believe instead of inlining the function [here](https://github.com/lordnik22/org-shoplist/blob/master/org-shoplist.el#L128) you should extract it and use `defsubst`.  That will quiet  the compiler warnings about functions being undefined.
- Instead of `(eq nil str)` you should use `(null str)` or `(not str)`, for example [here](https://github.com/lordnik22/org-shoplist/blob/master/org-shoplist.el#L130) but also in a couple other places.  On that particular line, though, you could write it simply as `(setq str (or str "0"))`.  See the list at the end of these comments.
- You can use `cadr` [here](https://github.com/lordnik22/org-shoplist/blob/master/org-shoplist.el#L156), `caddr` [here](https://github.com/lordnik22/org-shoplist/blob/master/org-shoplist.el#L167), etc.
- You can use `(zerop factor)` [here](https://github.com/lordnik22/org-shoplist/blob/master/org-shoplist.el#L208), but on that particular line you should probably drop the `if` and use an `unless`.
- Some will say you should use `string-empty-p` [here](https://github.com/lordnik22/org-shoplist/blob/master/org-shoplist.el#L263) but I personally find `string= "" ...` fine and a quick benchmark I did suggested it was very marginally faster.
- You could simplify the line [here](https://github.com/lordnik22/org-shoplist/blob/master/org-shoplist.el#L328) to `(/= (point) (point-max))` but make sure you use the right slash. :)
- You should use `unless` [here](https://github.com/lordnik22/org-shoplist/blob/master/org-shoplist.el#L278)
- Your call to `org-shoplist--concat-when-broken` is wrapped in a `save-excursion` -- but it looks to me like the `save-excursion` is the responsibility of `org-shoplist--concat-when-broken`, not its callers.  Ideally, when you can, side-effect-free functions are the way to go, especially if they're ultimately just computing a value for you.
- Some of your indentation should be tab-adjusted, e.g. [here](https://github.com/lordnik22/org-shoplist/blob/master/org-shoplist.el#L91)

-------
- 118: You can use `not` or `null`
- 129: You can use `not` or `null`
- 131: You can use `not` or `null`
- 192: You can use `not` or `null`
- 199: You can use `not` or `null`
- 220: You can use `not` or `null`
- 264: You can use `not` or `null`
- 265: You can use `not` or `null`
- 267: You can use `not` or `null`
- 280: You can use `not` or `null`
- 294: You can use `not` or `null`
- 343: You can use `not` or `null`
- 344: You can use `not` or `null`
- 359: You can use `not` or `null`
- 366: You can use `not` or `null`
- 383: You can use `not` or `null`
- 448: You can use `not` or `null`
- 456: You can use `not` or `null`

riscy:
(It looks like my line number links are out of date since your recent update to master -- just an fyi)

lordnik22:
Many thanks! So far everything is done, except that i won't start writting code with auto-fill-mode.
The links with the of-place-line-numbers weren't a problem, you gave me the right direction. 
I'm looking forward to finally write in the "Getting Started"-Chapter: `M-x package-install org-shoplist ENT`  :).
I'm out, have a good one

purcell:
Looking good. Remaining notes:

- The `defgroup` is declared as its own parent group: instead, place it in a standard group, e.g. `applications`.
- I'm a big fan of literal regexp syntax, but you still might find it useful to start using the `rx` macro for complex regexps like [this](https://github.com/lordnik22/org-shoplist/blob/932be7c8711389c0fb27db992d43895b1e26c9f1/org-shoplist.el#L88)
- [Here](https://github.com/lordnik22/org-shoplist/blob/932be7c8711389c0fb27db992d43895b1e26c9f1/org-shoplist.el#L131) (and in several other places) you're declaring a function in exactly the same file. That's nonsensical: instead, you should change the order of declarations as necessary to avoid this.
- Prefer `unless` to [`when` + `null`](https://github.com/lordnik22/org-shoplist/blob/932be7c8711389c0fb27db992d43895b1e26c9f1/org-shoplist.el#L464), and generally several of the conditional blocks of code could be simplified to be more idiomatic. For example, the `if` [here](https://github.com/lordnik22/org-shoplist/blob/932be7c8711389c0fb27db992d43895b1e26c9f1/org-shoplist.el#L345-L352) could be changed to `(when (and recipes (car recipes)) (list ...))`

riscy:
Looks like everything has been addressed.  Welcome to MELPA. :)

PR #6210, opened 2019-06-09
Add hg-histedit

jojojames:

Interface to mercurial/hg's histedit (similar to git rebase -i).

https://github.com/jojojames/hg-histedit

Author/Maintainer

None Needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/hg-histedit
+++ b/recipes/hg-histedit
@@ -1,0 +1,1 @@
+(hg-histedit :fetcher github :repo "jojojames/hg-histedit")


riscy:
These checks are semi-automated - forgive my mistakes.

`M-x checkdoc`:
No issues! 💯

`M-x byte-compile-file`:
No byte-compilation issues! 😎

`M-x package-lint`:
```
2 issues found:

13:39: warning: Use a non-snapshot version number for dependency on "with-editor" if possible.
180:21: warning: `with-eval-after-load' is for use in configurations, and should rarely be used in packages.
```
For the first one, you could require `(with-editor "2.8.3")`

`M-x check-declare-file` (not on the MELPA checklist):
No issues! 🎉

-------
- 7: Prefer `https://` over `http://`, if possible
- 31: Prefer `https://` over `http://`, if possible
- 250: Use `unless ...` instead of `when (not ...))`

jojojames:
Thanks, 

180:21: warning: `with-eval-after-load' is for use in configurations, and should rarely be used in packages.

is left, though that function would be called by the user.

riscy:
Looking at this a bit more closely --

- How come there is both a `Version` and a `Package-Version`?  I may be missing something.
- You list a `HomePage` and a `URL` in your frontmatter, but you could probably just list one.

Finally, since git-rebase.el was used as a template, there are copy fixes to make in a few places (e.g. change "git" to "hg", etc.):
- https://github.com/jojojames/hg-histedit/blob/9e86c1c109c41efb401935f39b2780e968fb1eb3/hg-histedit.el#L6
- https://github.com/jojojames/hg-histedit/blob/9e86c1c109c41efb401935f39b2780e968fb1eb3/hg-histedit.el#L44
- https://github.com/jojojames/hg-histedit/blob/9e86c1c109c41efb401935f39b2780e968fb1eb3/hg-histedit.el#L291
- https://github.com/jojojames/hg-histedit/blob/9e86c1c109c41efb401935f39b2780e968fb1eb3/hg-histedit.el#L106 (this variable should probably be `hg-histedit-description`)

jojojames:
Thanks, updated.

purcell:
Thanks for this. Some notes:

- The `eval-when-compile` [here](https://github.com/jojojames/hg-histedit/blob/550cdc5c968ec0734fccfabbf6395837e0fafda8/hg-histedit.el) seems bogus.
- Prefer to `:inherit` a built-in face [here](https://github.com/jojojames/hg-histedit/blob/550cdc5c968ec0734fccfabbf6395837e0fafda8/hg-histedit.el#L94) so that this new face plays well with themes by default.
- There's no `defgroup` to declare the faces subgroup referenced [here](https://github.com/jojojames/hg-histedit/blob/550cdc5c968ec0734fccfabbf6395837e0fafda8/hg-histedit.el#L98), or the `hg-histedit` group.
- The interactive wrapper for hg-histedit-run [here](https://github.com/jojojames/hg-histedit/blob/550cdc5c968ec0734fccfabbf6395837e0fafda8/hg-histedit.el#L114-L117) is bogus: just rename `hg-histedit-run` to `hg-histedit` and add the appropriate `interactive` form to it.

Looks good otherwise!

jojojames:
Thanks, sounds good, changed.

For:

> Prefer to :inherit a built-in face here so that this new face plays well with themes by default.

I just copied how magit did it.

```
(defface magit-hash
  '((((class color) (background light)) :foreground "grey60")
    (((class color) (background  dark)) :foreground "grey40"))
  "Face for the sha1 part of the log output."
  :group 'magit-faces)
```

Any recommendation here?

riscy:
> Any recommendation here?

If someone was using a color theme whose background was say `"grey50"` then either one of those would be hard to see, so I don't think magit is right in this case.  It seems like it might be more appropriate in you case to inherit `font-lock-constant-face`?

jojojames:
@riscy Thanks, changed.

riscy:
Great!  Thank you, this looks good.

PR #6211, opened 2019-06-09
Make live-py-mode's Python files pattern flexible

donkirkby:
An upcoming version will use a package folder.

Adds a minor mode for live coding in Python. The mode is already packaged, I'm just changing the build.

https://github.com/donkirkby/live-py-plugin

Maintainer

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/live-py-mode
+++ b/recipes/live-py-mode
@@ -1,4 +1,4 @@
 (live-py-mode
  :fetcher github
  :repo "donkirkby/live-py-plugin"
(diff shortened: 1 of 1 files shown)


purcell:
Super, thanks, and sorry for the slow turnaround.

donkirkby:
I'm patient, @purcell . I appreciate you looking after the MELPA system, I think it's the most common way for users to get my software.

PR #6212, opened 2019-06-10
Add face-shift

phikal:

Face-shift distorts the default colours of faces in a buffer, making a white background slightly more blue, green, pink, etc. An example of how this looks like can be found [here](https://icosahedron.website/@zge/102229312631398520).

https://git.sr.ht/~zge/face-shift

None needed.

### Checklist

Please confirm with `x`:

    ~~(I've used the 0.7 release, from melpa-stable.)~~

### DIFF
--- a/recipes/face-shift
+++ b/recipes/face-shift
@@ -1,0 +1,1 @@
+(face-shift :url "https://git.sr.ht/~zge/face-shift" :fetcher git)


alphapapa:
Are you planning to add a readme file to the repo with screenshots?  I had to read other posts in your link to understand that the idea is to change the color of buffers depending on their major mode (which is a very interesting idea, but without that context, I didn't understand the point of this package).

phikal:
alphapapa <notifications@github.com> writes:

> Are you planning to add a readme file to the repo with screenshots? I
> had to read other posts in your link to understand that the idea is to
> change the color of buffers depending on their major mode (which is a
> very interesting idea, but without that context, I didn't understand the
> point of this package).

Yes, I was adding some missing features to all the packages I submitted,
and the next thing on the list is to write README's with examples.
Should be finished within a day or two.

-- 
	With kind regards,
	Philip K.

phikal:
It has been done: https://git.sr.ht/~zge/face-shift

riscy:
This works great for me.  A couple quick notes before this looks ready to go:
- `face-shift.el:91: Argument `color' should appear (as COLOR) in the doc string`
- Looks like the customizable variable `face-shift-colors` should have type `:type '(alist :key-type symbol)`
- Minor nag: some tab characters on lines 118--121 push the lines way way out to the right for me (my tabs are eight spaces wide)

phikal:
Should be fixed with the last few commits.

riscy:
I think there's a typo in the variable name now -- I'm getting:
```
face-shift.el:92: Argument `coulor' should appear (as COULOR) in the doc string
```

phikal:
you're right, I realized that there were inconsistencies in how I was spelling colour/color so I tried to change everything to color but seem to have missed one instance somehow...

phikal:
@riscy since you added that tag, what's specifically missing?

purcell:
Thanks, this is a nice use of face remapping.

First thing I notice here is that you're using `cl-lib` but haven't declared a package dependency on it: `package-lint` should have told you about that.

Other notes:
- `derived-mode-p` already handles multiple possible parent modes, so `cl-some` is unnecessary [here](https://git.sr.ht/~zge/face-shift/tree/master/face-shift.el#L158)
- It's weird and unnecessary to make `face-shift` a function which returns a function. You're doing this in order to make the assignment to mode hooks trivial, but that's not a good reason to do this because that's just one way of using `face-shift`. Instead just make `face-shift` perform its effects when called, and tell users to wrap it in a lambda if they want to use it with a mode hook. Also, this will let you make it an interactive command, which will prompt for the colour shift mode when called via `M-x`.

Hope that helps!

phikal:
Thanks for the reply. I totally missed adding the `cl-lib` dependency, but that is done now. The `derived-mode-p` thing makes sense when one actually looks at the function signature ^^.

Regarding the _function that returns a function_, I think I'll have to reconsider the design a bit, especially since I thought about switching from RGB color space transformations to HSV/HSL. So I'll report back as soon as I think it's properly ready to be published.

DamienCassou:

phikal:
My [last commit](https://git.sr.ht/~zge/face-shift/commit/82490393c003ac2a22a6d2a5c9a87bf04ae934a3) should address most if not all of the issues. The main function is now a global minor-mode, and the code seems cleaner -- at least I hope ^^

I've also checked byte compilation, package-lint and checkdoc, and it all passes. Sorry for all the inconvenience.

phikal:
Also, how do I re-open a request? It seems I don't have the rights to do that...

riscy:
Hi, I reopened it for you.

You should see package-lint complaining that you'll need Emacs 26 for `when-let*`.  As a very minor point, the `progn` in the else part of the minor mode declaration is unnecessary, so you'd write:
```elisp
(if face-shift-mode
      (progn
        (mapc #'face-shift-setup (buffer-list))
        (add-hook 'after-change-major-mode-hook #'face-shift-setup))
    (mapc #'face-shift-clear (buffer-list))
    (remove-hook 'after-change-major-mode-hook #'face-shift-setup))
```
Otherwise this looks great - I'm bound to use this package and looking forward to customizing it. :)

phikal:
Fixed both -- I'm always annoyed by the asymmetry of lisp's `if`, so I guess that's the reason I wrote the minor mode like that... on the other hand I probably wasn't paying too much attention this way or another.

riscy:
Great thanks!

PR #6215, opened 2019-06-10
Add go-capf

phikal:

A gocode based completion backend from `completion-at-point`.

https://git.sr.ht/~zge/go-capf

I'm the maintainer and creator.

**None needed**

### Checklist

Please confirm with `x`:

    ~~I've used version 0.7 from melpa-stable.~~

### DIFF
--- a/recipes/go-capf
+++ b/recipes/go-capf
@@ -1,0 +1,1 @@
+(go-capf :url "https://git.sr.ht/~zge/go-capf" :fetcher git)


riscy:
These checks are semi-automated - please forgive any oversights.

<kbd>M-x checkdoc</kbd>:
```
Warning (emacs): 
go-capf.el:54: Name emacs should appear capitalized as Emacs
Warning (emacs): 
go-capf.el:100: First sentence should end with punctuation
Warning (emacs): 
go-capf.el:100: Argument `str' should appear (as STR) in the doc string
```

```
1 issue found:

105:1: error: "go-completion-at-point-function" doesn't start with package's prefix "go-capf".
```

phikal:
All except the last issue were added by a few extra commits since I opened this pull request, and I've fixed them now.

The last one

    105:1: error: "go-completion-at-point-function" doesn't start with package's prefix "go-capf".

is still a bit "hard" to fix, since go-capf is just short for go-completion-at-point...

purcell:
> ```
> error: "go-completion-at-point-function" doesn't start with package's prefix "go-capf".
> ```
> 
> 
> is still a bit "hard" to fix, since go-capf is just short for go-completion-at-point.

True. I'd probably just name the function `go-capf`, though. (You could `defalias` it to the longer name, and `package-lint` probably wouldn't notice, at least until I start detecting that in `package-lint`...)

phikal:
I renamed the function, but didn't do the defalias. After all, before it's published there's no need to ensure backwards compatibility, and another name is just another name.

purcell:
Avoid the `executable-find` [here](https://git.sr.ht/~zge/go-capf/tree/master/go-capf.el#L37) -- there's no need to lock the executable to the specific version on the PATH at the time this file is loaded, because the PATH might change later. Setting this to simply "go" would work as expected, so just do that. You can detect the absence of the executable later if necessary.

But apart from that, this looks good, so I'll go ahead and merge. Thanks for your patience.

PR #6217, opened 2019-06-11
Add recipe for `json-rpc-server`

jcaw:

This is a full implementation of the JSON-RPC 2.0 protocol for Emacs. You pass in a JSON string, Emacs executes the specified function(s), and a JSON-RPC 2.0 response is returned.

This package is designed sit underneath a transport layer. No transport logic is included. The transport layer is responsible for communicating with external clients. 

This package was originally part of [porthole](https://github.com/jcaw/porthole), a full-fledged RPC server system, but I abstracted it into its own package so it could be used as a framework for other RPC servers. It's just the JSON-RPC 2.0 protocol, plus function logic.

https://github.com/jcaw/json-rpc-server.el

Author & Maintainer

**None needed**

### Checklist

Please confirm with `x`:

---

`package-lint` did warn me about using a contraction as my package prefix. I'm not sure how to stop it from complaining about that. No other issues were raised.

### DIFF
--- a/recipes/json-rpc-server
+++ b/recipes/json-rpc-server
@@ -1,0 +1,3 @@
+(json-rpc-server
+ :fetcher github
+ :repo "jcaw/json-rpc-server.el")


riscy:
Just a quick comment - it looks like you're providing a package called `json-rpc-server` but your defgroup and all of your functions are specifying a prefix of `jrpc`.  I'm not sure whether this is okay, but package-lint is not happy:

```
27 issues found:

92:1: error: "jrpc--error-codes" doesn't start with package's prefix "json-rpc-server".
101:1: error: "jrpc--unknown-error-code" doesn't start with package's prefix "json-rpc-server".
105:1: error: "jrpc--get-error-code" doesn't start with package's prefix "json-rpc-server".
115:1: error: "jrpc--encode-error-response" doesn't start with package's prefix "json-rpc-server".
161:1: error: "jrpc--throw-error-response" doesn't start with package's prefix "json-rpc-server".
173:1: error: "jrpc--throw-invalid-json" doesn't start with package's prefix "json-rpc-server".
187:1: error: "jrpc--throw-invalid-request" doesn't start with package's prefix "json-rpc-server".
196:1: error: "jrpc--throw-invalid-function" doesn't start with package's prefix "json-rpc-server".
205:1: error: "jrpc--throw-error-calling-method" doesn't start with package's prefix "json-rpc-server".
219:1: error: "jrpc--throw-result" doesn't start with package's prefix "json-rpc-server".
226:1: error: "jrpc--encode-result-response" doesn't start with package's prefix "json-rpc-server".
247:1: error: "jrpc-null-p" doesn't start with package's prefix "json-rpc-server".
256:1: error: "jrpc-alist-get" doesn't start with package's prefix "json-rpc-server".
267:1: error: "jrpc--call-function-internal" doesn't start with package's prefix "json-rpc-server".
276:1: error: "jrpc--call-function" doesn't start with package's prefix "json-rpc-server".
287:1: error: "jrpc--execute-request" doesn't start with package's prefix "json-rpc-server".
332:1: error: "jrpc--validate-request" doesn't start with package's prefix "json-rpc-server".
390:1: error: "jrpc--decode-request-json" doesn't start with package's prefix "json-rpc-server".
428:1: error: "jrpc--replace-symbol-strings" doesn't start with package's prefix "json-rpc-server".
476:1: error: "jrpc--replace-unencodable-object" doesn't start with package's prefix "json-rpc-server".
518:1: error: "jrpc--extract-id" doesn't start with package's prefix "json-rpc-server".
535:1: error: "jrpc--decode-id" doesn't start with package's prefix "json-rpc-server".
548:1: error: "jrpc-unknown-error-response" doesn't start with package's prefix "json-rpc-server".
562:1: error: "jrpc-internal-error-response" doesn't start with package's prefix "json-rpc-server".
601:1: error: "jrpc--ammend-id" doesn't start with package's prefix "json-rpc-server".
612:1: error: "jrpc--handle-single" doesn't start with package's prefix "json-rpc-server".
640:1: error: "jrpc-handle" doesn't start with package's prefix "json-rpc-server".
```

jcaw:
>Sorry, I see you've already acknowledged that. 

No worries! I'm sure there must be a formal way to address it, no idea what it is though.

>Another thing I thought I'd add in the meantimeis that there is already a package on MELPA called json-rpc whose namespace you should be aware of.

Yep, another reason I used a contraction. I figure `jrpc` is less likely to clash.

riscy:
The best way to address it is to either change those function prefixes, _or_ you could rename the package to `jrpc` and the file to `jrpc.el` -- if you do that, package-lint will be happy.

purcell:
Mostly looking great, thanks for this.

> The best way to address it is to either change those function prefixes, _or_ you could rename the package to `jrpc` and the file to `jrpc.el` -- if you do that, package-lint will be happy.

^ Yes, this. The symbol prefix should match the package name throughout. `jrpc` would be fine, or `jsonrpc`, or just use `json-rpc-server--` for the private symbols for consistency. Honestly, that last option is the best. Saving a few chars in symbol names isn't a win.

> I've seen the behaviour in other packages, e.g. `smartparens`.

This is the probem with popular long-standing packages which flaunt the emacs conventions: they set a bad example, which is why we wouldn't accept them into MELPA if submitted now in their current form.

I'd also note that the recipe can and should be simplified, please: both the `:branch` and `:files` specs are unnecessary because the defaults are equivalent (and therefore preferred) here.

Hope that helps!

jcaw:
Recipe fixed, package prefix changed to `json-rpc-server`. Sorry for the delay!

riscy:
Great thanks!  Merging.

PR #6218, opened 2019-06-11
Add recipe for `porthole`

jcaw:

Porthole lets you start RPC servers in Emacs. 

These servers allow Elisp to be invoked remotely via HTTP requests. These servers are designed to be automatic (when running locally), meaning they require no configuration from the client to work. All the client needs is a name, and it can execute remote Elisp procedure calls automatically.

Many Porthole servers can run independently without worrying about one another. It's designed to act as a framework that other packages can build on. 

A full Python client is available on `pip`, repo [here](https://github.com/jcaw/porthole-python-client). It allows one-liner RPC calls to Emacs, from Python.

This package depends on [`json-rpc-server`](https://github.com/jcaw/json-rpc-server.el), which handles the underlying JSON-RPC protocol. I've also opened a pull request to add `json-rpc-server` to MELPA [here](https://github.com/melpa/melpa/pull/6217). 

https://github.com/jcaw/porthole

Author & Maintainer

**None needed**

### Checklist

Please confirm with `x`:

---

- `package-lint` didn't want me to include Emacs in the package name - kept it to make it obvious the RPC functionality exposes *Emacs* - Emacs isn't the client, it's the server.
- Byte compilation just gives me two warnings:
  - `porthole.el:839:1:Warning: Unused lexical variable ‘i’` - Not sure how to `dotimes` without a variable to hold the count.
  - `porthole.el:71:1:Warning: Package assoc is obsolete!` - Only gives me this warning sometimes (?), not sure what I should use instead in this instance - should I explicitly use `alist-get` with a `testfn`?

### DIFF
--- a/recipes/porthole
+++ b/recipes/porthole
@@ -1,0 +1,3 @@
+(porthole
+ :fetcher github
+ :repo "jcaw/porthole")


riscy:
Hello -- a quick pass through the checklist:

```
```

```
porthole.el:863:1:Warning: Unused lexical variable `i'
```
(You can fix this by prepending a `_` to the variable)

From line 1042: I believe you should avoid auto-enabling the mode or parts of it when this package is merely loaded.  (For example, not everyone who requires this package will necessarily want elnode-error-log-to-messages set to nil).  I think this should be written as a global minor mode that can be toggled on and off?

jcaw:
>1:1: warning: Including "Emacs" in the package summary is usually redundant.

Sure - in this case I kept it to make it clearer where the servers are being started (i.e. Emacs is the server, they expose Emacs' functionality). You think I should reword?

>(You can fix this by prepending a _ to the variable)

Ah! Of course. Thanks. 

>From line 1042: I believe you should avoid auto-enabling the mode or parts of it when this package is merely loaded. (For example, not everyone who requires this package will necessarily want elnode-error-log-to-messages set to nil). I think this should be written as a global minor mode that can be toggled on and off?

So, my thinking here is that  the behaviour of Porthole in and of itself doesn't really fit cleanly into  the mode-based paradigm.  Packages that use Porthole need to know the functionality will be available and behave as expected.  Is it okay for the global minor mode to be downstream from the basic functionality? I.E, is it ok for Porthole to automatically enable its global minor-mode when a server is started (and perhaps automatically disable it when all servers are stopped?)

>(For example, not everyone who requires this package will necessarily want elnode-error-log-to-messages set to nil).

Just to be overt, it's ok to do this with a global minor-mode? I don't like having to disable it, but Elnode is so chatty Emacs would slow to a crawl if RPC calls were being made frequently (for example, another user on the machine scanning network services would flood your message buffer). I had a look at modifying the behaviour in Elnode (I'm maintaining the MELPA fork at the moment), but it's baked in so it looks like a big job. I don't want to make major changes until I have confirmation Nic Ferrier dropped the project.

purcell:
Thanks, this is interesting. Some quick minor notes from me:

- A simple direct `pcase` on `system-type` [here](https://github.com/jcaw/porthole/blob/57ff55a96119c17b1c07630bc177233f56907878/porthole.el#L141-L153) would eliminate the need for the unnecessary indirection of the `porthole--on-*` vars. They're only used in one place.
- You could write [this](https://github.com/jcaw/porthole/blob/57ff55a96119c17b1c07630bc177233f56907878/porthole.el#L324-L335) more simply with `cl-loop` or `seq-find`.

Otherwise looks good, so once `json-rpc-server` is ready, this can also be merged.

jcaw:
I've rewritten to use `pcase` and `seq-find` from @purcell. Currently, I'm not creating a minor mode but I'm happy to add that too if you still think it's preferable.

I also removed `branch` from the recipe and implemented the suggestions for `json-rpc-server`. Let me know if there's anything else that needs doing.

riscy:
With `json-rpc-server` merged I'm taking a look at this.  It looks like your recipe could be simplified similarly to how `json-rpc-server` was (will the default recipe suffice?).  Also:

```
2 issues found:

```

```
porthole.el:854:1:Warning: Unused lexical variable `i'
```
(you can prefix the variable as `_i`)

Some other miscellaneous items for your consideration:
- It seems like there should be some error handling in `porthole--stop-all-servers` just in case -- it would be unfortunate to block someone from being able to exit Emacs?
- [#L1033](https://github.com/jcaw/porthole/blob/master/porthole.el#L1033): It is beneficial to sharp-quote hooks; use `#'`
- [#L234](https://github.com/jcaw/porthole/blob/master/porthole.el#L234): It is beneficial to sharp-quote function names; use `#'`
- [#L510](https://github.com/jcaw/porthole/blob/master/porthole.el#L510): It is beneficial to sharp-quote function names; use `#'`
- [#L313](https://github.com/jcaw/porthole/blob/master/porthole.el#L313): Function names should be sharp-quoted; use `#'`
- [#L623](https://github.com/jcaw/porthole/blob/master/porthole.el#L623): Function names should be sharp-quoted; use `#'`

jcaw:
Suggestions all merged onto `master`, except removing Emacs from the package name. Wanted to keep it to communicate directionality:

>in this case I kept it to make it clearer where the servers are being started (i.e. Emacs is the server, they expose Emacs' functionality). You think I should reword?

>will the default recipe suffice

I figured I had to manually exclude the `external-tests` folder since it was non-standard - didn't realise all subfolders are ignored by default. I've force-pushed the default recipe instead.

riscy:
Thanks!

PR #6219, opened 2019-06-11
Update lua-mode recipe

immerrr:
There's an [upcoming change](https://github.com/immerrr/lua-mode/pull/148) that would require to accompany the source code with a Lua initializer script for the inferior shell. This is to ensure the script is redistributed accordingly. 

Please note, that `scripts` directory is not there yet, and I'm not sure if the recipe works without it. If that's the case we can wait with merging until the scripts directory is created.

lua-mode is a major mode for editing Lua files

https://github.com/immerrr/lua-mode

I'm the maintainer.

No package.el compatibility changes were introduced.

### Checklist

Please confirm with `x`:

Looks like, lua-mode is quite far behind on the package quality side of things, but if it's OK I would prefer to fix it in a separate PR.

### DIFF
--- a/recipes/lua-mode
+++ b/recipes/lua-mode
@@ -1,4 +1,4 @@
 (lua-mode
  :repo "immerrr/lua-mode"
  :fetcher github
- :files ("lua-mode.el" ("scripts" "scripts/*")))
+ :files (:defaults "scripts"))
--- a/recipes/lua-mode
+++ b/recipes/lua-mode
@@ -1,3 +1,4 @@
 (lua-mode
  :repo "immerrr/lua-mode"
- :fetcher github)
+ :fetcher github
+ :files ("lua-mode.el" ("scripts" "scripts/*")))


tarsius:
> Looks like, lua-mode is quite far behind on the package quality side of things, but if it's OK I would prefer to fix it in a separate PR.

That's reasonable but I would suggest you do the "I've built and installed the package using the instructions in CONTRIBUTING.org" step in order to answer your question "[the] scripts directory is not there yet, and I'm not sure if the recipe works without it."

tarsius:
Err actually... we know that the build part succeeded and the recipe does not affect installation, so all good.

riscy:
@immerrr can you ping when the `scripts` directory is set up?  Then we can verify the files you're specifying are being laid out exactly how you want.

You can refer to the recipe for `auto-complete`:
```elisp
(auto-complete :repo "auto-complete/auto-complete" :fetcher github :files
               ("*.el" "dict"))
```
which creates a file tree that looks like this:
```
├── auto-complete-20170125.245   
   ├── auto-complete-autoloads.el
   ├── auto-complete-config.el   
   ├── auto-complete-config.elc  
   ├── auto-complete-pkg.el      
   ├── auto-complete.el          
   ├── auto-complete.elc         
   └── dict                      
```
...where the `dict` directory is full of files.

purcell:
A better (and equivalent) `:files` pattern would be `(:defaults "scripts")`. I'll make that change and merge this.

immerrr:
Thank you @purcell

PR #6220, opened 2019-06-11
Add eyezoom recipe.

sandinmyjoints:

[Eyebrowse](https://github.com/wasamasa/eyebrowse) and [Zoom](https://github.com/cyrus-and/zoom) are both great packages. Sometimes you may want Zoom to apply only to some Eyebrowse workspaces (window layouts) but not others. Out of the box, it does not support this; it can only be turned on or off. So if you have workspaces you'd prefer to manage yourself, it will mess with them whenever you switch to them. This package lets you specify the tags of which workspaces Zoom should apply to. Workspaces that don't have such a tag are left alone.

https://github.com/sandinmyjoints/eyezoom

Author and maintainer

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/eyezoom
+++ b/recipes/eyezoom
@@ -1,0 +1,1 @@
+(eyezoom :fetcher github :repo "sandinmyjoints/eyezoom")


riscy:
Looks good so far, and thanks for adhering so closely to the checklist!

It looks like this package will automatically enable itself if it's installed and loaded.  Instead of that, it's probably better to define a global minor mode (https://www.gnu.org/software/emacs/manual/html_node/elisp/Defining-Minor-Modes.html) so that the package's behavior can be toggled on and off.

Also, it seems unnecessary to `advice-remove` every time
`eyebrowse-switch-to-window-config gets called -- worth looking into?

A few minor comments:
- Per CONTRIBUTING.org, you should consider using lexical binding
- 73: Specifying this :group is unnecessary (it is set implicitly)
- 56: Prefer `https` over `http` (if possible)

sandinmyjoints:
Thanks for reviewing, @riscy! Great feedback, and I intend to address it, but it might take a few weeks as life is quite busy for me atm. Is it OK if I ping you when I've updated it?

sandinmyjoints:
I took a quick look at a few of these, and found that:
- package-lint complains "Customization groups should specify a parent via :group" when I remove `:group` from `defgroup` -- is that OK?
- the `advice-remove`s come directly from zoom: https://github.com/cyrus-and/zoom/blob/578295532fb1c4ad2a2e95894e65cce02f812b54/zoom.el#L147-L149 `eyezoom-turn-off-zoom-mode` in this package is the same as `zoom--off` from zoom.el except that it doesn't do the last thing `zoom-off` does, which is balance windows, because if it does that, then windows that "don't zoom" will be balanced, whereas this package is meant to ensure they are left alone.

riscy:
> package-lint complains "Customization groups should specify a parent via :group" when I remove :group from defgroup -- is that OK?

Apologies for that - you don't want to remove the `:group 'tools` from `defgroup`, and it's only the `defcustom` around line 73 that is actually already implicitly a member of the group.  It's not a big deal if you keep all the `:group`s either way.

> the advice-removes come directly from zoom [snip]

I'll look into this one some more and see if I can be more helpful re: zoom'e behavior. :)

riscy:
I've taken a look at zoom, too.  This makes sense.

> Would it be preferable to accomplish that by temporarily rebinding frame-list or balance-windows to be noops so that the real zoom--off can run?

I wonder if this might be the best approach.  There's a possibility zoom might change how it handles turning itself off in the future, and in this case your approach would be able to keep pace.  I think you'd need to do something like:

```elisp
(require 'cl-lib)

...

(defun eyezoom-turn-off-zoom-mode (&rest _args)
  "This is `zoom--off' but prevents the rebalancing of windows."
   (cl-letf (((symbol-function #'balance-windows) #'ignore))
     (zoom--off)))
```

purcell:
Yeah, agree - definitely don't want to be disabling advice that was added by another package.  Also setting `zoom-mode` to `nil` is inappropriate if you could call `(zoom-mode -1)` instead.

purcell:
It feels likely that the right thing here might be for `zoom` and/or `eyebrowse` to change slightly so this type of integration requires less magic.

DamienCassou:

sandinmyjoints:
Sounds good. I haven't had the bandwidth to address these great suggestions, but when I do I'll reopen.

PR #6221, opened 2019-06-11
add frog-jump-buffer recipe

waymondo:

`frog-jump-buffer` is a convenience package that allows one to hop to any Emacs buffer in 2-3 key
strokes.

https://github.com/waymondo/frog-jump-buffer

Maintainer & enthusiastic user.

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/frog-jump-buffer
+++ b/recipes/frog-jump-buffer
@@ -1,0 +1,1 @@
+(frog-jump-buffer :repo "waymondo/frog-jump-buffer" :fetcher github)


riscy:
This looks nice!  I think this byte-compile warning could be fixed:

```
frog-jump-buffer.el:82:42:Warning: reference to free variable
    `frog-jump-buffer-current-filter-function'
```
by moving the `defvar` above when the variable is first referenced.

waymondo:
Strange, I'm not seeing that. The file byte compiles for me with no warnings and also `frog-jump-buffer-current-filter-function` is indeed defined before it is referenced.

riscy:
Thanks for this, it looks great.

PR #6226, opened 2019-06-14
Telega recipe

zevlg:

telega is full-featured Telegram client for GNU Emacs.

https://github.com/zevlg/telega.el

Creator, maintainer

**None needed**

### Checklist

Please confirm with `x`:


purcell:
Thanks for this, and sorry for the slow response. Mostly looks great! Some notes:

- Including `test.el` in  the package is inappropriate, as is the `README`. Only include files  that are needed at runtime. In general, assume that the user shouldn't have to find the package installation directory. If you're going to bundle the `Makefile` and C source code, you might like to provide an interactive command which will compile it.
- Identifiers like `telega--on-updateConnectionState`, ie. with names that don't follow the elisp conventions, should be avoided if possible. If you're dynamically finding function names from event names, then it may still be permissible.
- Instead of the bogus `interactive-p` prefix arg in code like [this](https://github.com/zevlg/telega.el/blob/4b14b8a81719c9bc89530e4e7c12f03e4a3f4c84/telega.el#L313-L328), just use `(called-interactively-p)`.

Hope that helps!

-Steve

zevlg:
> Thanks for this, and sorry for the slow response. Mostly looks great! Some notes:
> 
>     * Including `test.el` in  the package is inappropriate, as is the `README`. Only include files  that are needed at runtime. In general, assume that the user shouldn't have to find the package installation directory. If you're going to bundle the `Makefile` and C source code, you might like to provide an interactive command which will compile it.

We have `telega-server-build` interactive command to build `telega-server` binary needed to run telega.  `test.el` is used in `telega-server` installation phase to make sure Emacs and environment is ok to run telega.  Using environment that is not suitable to run telega makes strange runtime errors, that are hard to track down.  So running tests from `test.el` is essential at least in installation phase

> 
>     * Identifiers like `telega--on-updateConnectionState`, ie. with names that don't follow the elisp conventions, should be avoided if possible. If you're dynamically finding function names from event names, then it may still be permissible.

Yeah, this function names builds dynamically, using
```elisp
(intern (format "telega--on-%s" (plist-get event :@type)))
```

> 
>     * Instead of the bogus `interactive-p` prefix arg in code like [this](https://github.com/zevlg/telega.el/blob/4b14b8a81719c9bc89530e4e7c12f03e4a3f4c84/telega.el#L313-L328), just use `(called-interactively-p)`.

just bad name, I'll rename prefix arg to `print-p`

riscy:
I think this package is missing a dependency on `visual-fill-column` (not a part of Emacs).

- A number of the files are missing requires on parts of the package they need.  As one example, telega-webpage.el should require telega-server, telega-ins, telega-media, and visual-fill-column.  The benefit of making these `require`s explicit is that if you ever change your interfaces, you'll know these warnings are actually errors that need to be fixed before they reach users.
- Enabling lexical binding on each of the .el files will make flycheck more useful for detecting unused arguments to functions while you're editing those files.
- I realize this is a big project and so not everything can feasibly be fixed before this should be added to MELPA, but there are a couple easy fixes here and there:
  ```
  telega-tme.el:67: `y-or-n-p' argument should end with "? "
  telega-customize.el:91: Open parenthesis in column 0 should be escaped
  telega-core.el:484: Error messages should *not* end with a period
  telega-chat.el:2815: `y-or-n-p' argument should end with "? "
  ```

riscy:
> We have telega-server-build interactive command to build telega-server binary needed to run telega. test.el is used in telega-server installation phase to make sure Emacs and environment is ok to run telega. Using environment that is not suitable to run telega makes strange runtime errors, that are hard to track down. So running tests from test.el is essential at least in installation phase

I'll chime in with my thoughts on this -- the README.md still is definitely not necessary, but if `test.el` is necessary to distribute to your users in order for them to use the package, then it should look and act like part of the package -- maybe you can make `test.el` into a file called `telega-test.el` containing an interactive command that can be run to initiate the tests for the user?  Then when the user loads `telega`, you could `message` the user to run `M-x telega-test` (or whatever) to validate a fully compatible environment.

zevlg:
> > Then when the user loads `telega`, you could `message` the user to run `M-x telega-test` (or whatever) to validate a fully compatible environment.

I'll split testing into two parts:

1) Internal tests, used only for development (won't be part of package)
2) Emacs environment test, which runs at first start (part of the package)

riscy:
A couple quick things -- apologies for the multiple passes, there is a lot of code. :)

- This line [here](https://github.com/zevlg/telega.el/blob/27173886c0c133ad049b8bd256b6c9bca1340223/telega.el#L8) should be something like `;; Package-Requires: ((emacs "26.1") (visual-fill-column "1.9"))` -- check `package-lint`.
- The key sequence [here](https://github.com/zevlg/telega.el/blob/5333c05d2c96529521a055d647c274d26951c818/telega-webpage.el#L129) is reserved per Emacs convention

zevlg:
> A couple quick things -- apologies for the multiple passes, there is a lot of code. :)
> 
>     * This line [here](https://github.com/zevlg/telega.el/blob/27173886c0c133ad049b8bd256b6c9bca1340223/telega.el#L8) should be something like `;; Package-Requires: ((emacs "26.1") (visual-fill-column "1.9"))` -- check `package-lint`.
> 
>     * The key sequence [here](https://github.com/zevlg/telega.el/blob/5333c05d2c96529521a055d647c274d26951c818/telega-webpage.el#L129) is reserved per Emacs convention

Thanks!  I've fixed this

riscy:
It looks like the major issues are addressed.  Thanks for this -- this package looks like it will be useful to a lot of people.

zevlg:
What is the best way to use tags (as releases) instead of master branch in recipe?  I’m planning to use master branch for telega development

riscy:
That's totally compatible with MELPA.  You can develop on master and folks who use MELPA unstable will be able to see your latest.  Whenever you tag a release, it will show up on MELPA stable.  You should probably keep a `Version` or `Package-Version` in your header to keep package-lint happy, but MELPA handles everything.

tarsius:
Explicitly specifying `:branch master` seems redundant here or am I missing something?

riscy:
No, it's definitely redundant -- I missed that!

zevlg:
> No, it's definitely redundant -- I missed that!

I’ll fix that

PR #6227, opened 2019-06-14
add filetags.el

DerBeutlin:

The package helps to manage filetags which are stored in the filename of files.
It can be used from dired.

https://github.com/DerBeutlin/filetags.el

Maintainer

None needed

### Checklist

Please confirm with `x`:

For some reason it complains about arguments not being in the docstring although they are.

### DIFF
--- a/recipes/filetags
+++ b/recipes/filetags
@@ -1,0 +1,3 @@
+(filetags
+  :repo "DerBeutlin/filetags.el"
+  :fetcher github)


riscy:
These checks are semi-automated - please forgive any oversights.

<kbd>M-x checkdoc</kbd>:
```
Warning (emacs): 
filetags.el:119: Argument `tags-with-prefix' should appear (as TAGS-WITH-PREFIX) in the doc string
Warning (emacs): 
filetags.el:131: Argument `tags-with-prefix' should appear (as TAGS-WITH-PREFIX) in the doc string
Warning (emacs): 
filetags.el:144: Argument `tags-with-prefix' should appear (as TAGS-WITH-PREFIX) in the doc string
Warning (emacs): 
filetags.el:152: Argument `tags-with-prefix' should appear (as TAGS-WITH-PREFIX) in the doc string
Warning (emacs): 
filetags.el:260: Argument `tags-with-prefix' should appear (as TAGS-WITH-PREFIX) in the doc string
Warning (emacs): 
filetags.el:287: Argument `add-candidates' should appear (as ADD-CANDIDATES) in the doc string
Warning (emacs): 
filetags.el:363: Argument `tags-with-prefix' should appear (as TAGS-WITH-PREFIX) in the doc string
```

No byte-compilation issues! 😎

```
1 issue found:

```

<kbd>M-x check-declare-file</kbd> (not on the MELPA checklist):
No issues! 🎉

-------
- 3: Copyright notice may be out of date.
- 40: Specifying :group is unnecessary after defgroup
- 49: Specifying :group is unnecessary after defgroup
- 55: Specifying :group is unnecessary after defgroup
- 62: Specifying :group is unnecessary after defgroup
- 137: Use `unless ...` instead of `when (not ...)`
- 139: Use `unless ...` instead of `when (not ...)`
- 222: Use `unless ...` instead of `when (not ...)`
- 266: Use `unless ...` instead of `when (not ...)`
- 365: Use `unless ...` instead of `when (not ...)`
- 39: Customizable variables shouldn't be private

DerBeutlin:
I fixed:
-  the unless ⇔ when (not ) stuff,
- removed the `:group`
- updated the copyright

I did not address the following:
- I receive the same checkdoc warnings but I don't know why I have the arguments uppercase in the docstrings
- my package summary starts with an upercase character

riscy:
Looks like part of the issue is that you're using the wrong kind of hyphen (character 0x2014) in your docstrings.  Should be character 0X2D.

DerBeutlin:
Ah!! Thanks for the help!
I fixed everything you mentioned.

riscy:
Looking at this a little more deeply --
- The default separator between tags is a space (and the user cannot change this), and `filetags-delimiter` contains spaces.  Some people (me!) prefer to avoid spaces in filenames due to how shells typically tokenize on spaces.  I think it would be *very* worthwhile to make the between-tags-separator customizable, and maybe worth considering having this package not insert spaces by default.
- I believe the `mapconcat` in `filetags-completing-read-tag` requires an additional argument -- this causes me not to be able to use `filetags-dired-update-tags`.
- `filetags-find-dot-filetags-in-upper-tree` seems to not find `.filetags` for me in some circumstances:
  ```elisp
  ;; $ touch .filetags
  (filetags-find-dot-filetags-in-upper-tree ".") ; => nil (perhaps you should use file-name-as-directory on dir)
  (filetags-find-dot-filetags-in-upper-tree "./") ; => finds it!
  ;; $ mv .filetags ../
  (filetags-find-dot-filetags-in-upper-tree "./") ; => can't find it?
  ```

purcell:
Thanks! Couple of notes:

- `seq` package is not in Emacs 24.4, so you need a package dependency for that. `package-lint` should have mentioned this -- please give it a go.
- You're also using the `s`  library without a package dependency. Though I think you're only using `s-starts-with`, so consider `string-prefix-p` instead, which is built in.
- Typo: candidat => candidate
- There's a less repetitive way to write [this](https://github.com/DerBeutlin/filetags.el/blob/7b160fd79af29b613aa418f231f77c6633e6183f/filetags.el#L91-L98) by placing the conditional  inside the list of args passed to `concat`: when `concat` sees `nil`, then that's an empty string, so  you  can say `(concat a (when some-cond b) c)`, for example.

Hope that helps!

DerBeutlin:
Thanks a lot!
@riscy 
- I made the separator between the tags configurable but kept the default as it is compatible with https://github.com/novoid/filetags. 
- mapconcat lacked an argument thanks for that
- I could reproduce the first two cases but not the third. I added `file-name-as-directory` to make sure `DIR` is used as directory

@purcell 
- `seq` was already required. Therefore `package-lint` did not complain
- I replaced `s-start-with?` with `string-prefix-p` and removed the dependency on `s`
- fixed the typo
- moved the conditional inside the concat

riscy:
This looks good to go.  One last thought for your consideration is that the prompt in `filetags-completing-read-tag` is a little bit confusing to me without reading the documentation.  I like the workflow you created, but I wonder if rewriting the prompt to be something like the following might make it a little more user-friendly:
```elisp
(defun filetags-completing-read-tag (collection selected-tags)
  "Completing read function for filetags with collection COLLECTION.
SELECTED-TAGS is a list of already selected tags.
If \"Commit\" is chosen return nil otherwise return the chosen tag."
  (let ((new-tag
         (completing-read
          (concat "Add/remove tags with '+<tag>' or -<tag>"
                  (if selected-tags
                      (format " (%s): " (mapconcat #'identity selected-tags " "))
                    ": "))
          (push "Commit" collection)
          nil
          filetags-enforce-controlled-vocabulary
          nil
          nil
          "Commit")))
    (unless (string= new-tag "Commit") new-tag)))
```
Regardless, good work.

PR #6231, opened 2019-06-16
Add jsonl recipe

ebpa:

This library contains some utility functions to simplify working with line-delimited JSON (JSONL).  Sometimes called "newline-delimited JSON" or "JSON lines", line-delimited JSON is a format for encoding multiple JSON values within a single file:

    {foo: "bar"}
    {baz: "bat"}

https://github.com/ebpa/jsonl.el

I am the sole author.

**None needed**

### Checklist

### DIFF
--- a/recipes/jsonl
+++ b/recipes/jsonl
@@ -1,0 +1,3 @@
+(jsonl
+ :repo "ebpa/jsonl.el"
+ :fetcher github)


riscy:
Awesome!  Works great for me.  A couple of very minor non-issues I'll leave as a comment before I hit merge:

- The example jsonl you give in your frontmatter is not valid jsonl (e.g. `{foo: "bar"}` should be `{"foo": "bar"}`)
- After reading up on it, `jsonl-line-separator` only needs to be `\n` and doesn't ever need to be `\r\n`, because trailing whitespace should be ignored when parsing these files (per http://jsonlines.org)

ebpa:
> The example jsonl you give in your frontmatter is not valid jsonl (e.g. {foo: "bar"} should be {"foo": "bar"})

@riscy good catch!  Thanks!

> After reading up on it, jsonl-line-separator only needs to be \n and doesn't ever need to be \r\n, because trailing whitespace should be ignored when parsing these files (per http://jsonlines.org)

`jsonl-line-separator` is only used when writing jsonl files (pretty much just to accommodate Windows users expecting to see JSON values on separate lines with their viewers/editors of choice).  The reader in this package (`json-read`) happens to ignore the whitespace, so the separator isn't a concern when reading at the moment.  I'll include a remark in the docstring.

Thank you for all your service to the community!

PR #6232, opened 2019-06-18
Added Coleslaw-mode for Common Lisp static content editing/generating

equwal:

A mode for editing files used by [kingcon's "coleslaw" site content generator](https://github.com/kingcons/coleslaw) to generate static html content. The interesting bits are:
- Every file requires a header
- The type of header required changes based on the filetype: `.page` and `.post` files need different headers.
- The user specifies the style of markup in the header, so the major mode needs to be changed accordingly. Emacs can't do this idiomatically, since it is designed to use file type (ie. extension) information for that

I found myself hand writing these headers and making plenty of preventable errors, while also having to `M-x lisp-mode` or `M-x markdown-mode` and  `M-x markdown-preview-eww` every time I wanted to edit or make a page.

There are no hard dependencies, though if a user runs `(coleslaw-setup)` it assumes that the user has the packages installed that are run according to `coleslaw-modes`, which defines its default behaviour in the docstring.

https://github.com/equwal/coleslaw-mode
Docs available at https://spensertruex.com/coleslaw-mode

I am the creator. Coleslaw itself is here: https://github.com/kingcons/coleslaw. My mode is for editing the files needed for kingcons' project.

I am the upstream package maintainer. Coleslaw itself doesn't require any contact, though they have added a link to my repo to the README.

### Checklist

Please confirm with `x`:

Old pull request is here: https://github.com/melpa/melpa/pull/5858
Responding to @purcell and his suggestions
* Coleslaw is a minor mode now
* coleslaw-setup hopefully is idiomatic, no requiring libraries (except to check if they exist without erroring)
* No automatic mode suggestions to users
* The whole mode system I jerry-rigged is dumped in favour of a user-accessible alist.
* No more bogus autoload
* Old lighter was bogus, it is "Coleslaw" this time. Would "Cslaw" is better for not spamming the mode line?

### DIFF
--- a/recipes/hl-block-mode
+++ b/recipes/hl-block-mode
@@ -1,3 +1,3 @@
 (hl-block-mode
  :repo "ideasman42/emacs-hl-block-mode"
-:fetcher github)
+ :fetcher gitlab)
--- a/recipes/inkpot-theme
+++ b/recipes/inkpot-theme
@@ -1,3 +1,3 @@
 (inkpot-theme
  :repo "ideasman42/emacs-inkpot-theme"
- :fetcher github)
+ :fetcher gitlab)
--- a/recipes/run-stuff
+++ b/recipes/run-stuff
@@ -1,3 +1,3 @@
 (run-stuff
  :repo "ideasman42/emacs-run-stuff"
- :fetcher github)
+ :fetcher gitlab)


equwal:
Old pull request is here: https://github.com/melpa/melpa/pull/5858

equwal:
- [x] Copyright notice to 2019
- [x] `(require 'cl-lib)`
- [?] `when (not...)` is not anywhere in my files\
- [?] `auto-insert-alist` is surrounded by the `(when (require 'autoinsert nil t)...)` form, since I want auto-insert as a soft dependency
- [?] I should require `autoinsert` I think, not `auto-insert`. I was having it as a soft dependency, since some users might not like using my autoinsertion stuff (eg. if they have custom yasnippets or custom snippets). I wrapped the require in a `(when coleslaw-auto-insert...)` which defaults to NIL but is set to T by `(coleslaw-setup)`.

I was warned by @purcell in my last request not to put requires below toplevel, but I can't think of a better way to provide a soft dependency (set by a variable) than this way.

equwal:
I resolved the require issue by changing to an `eval-after-load`.

equwal:
I'm considering removing the skeletons from this package, and trying to add them to a templating package like yasnippet instead. That would make my package only do one thing: dispatch the mode based on the format header.

What do you think?

equwal:
I found the `if (not...` and replaced it with an `if...` after swapping the predicates.

equwal:
Byte compiles cleanly now

riscy:
> I'm considering removing the skeletons from this package, and trying to add them to a templating package like yasnippet instead. That would make my package only do one thing: dispatch the mode based on the format header.

Sorry for the slower turnaround -- that sounds like a good idea to me.  You could emulate a package like [this one](https://github.com/toumorokoshi/go-snippets/blob/d437df148879566ffe7f2e503a3cf2602aa9fb28/go-snippets.el) for your skeletons.  The benefit of relying on yasnippet for this is that it probably gracefully handles corner cases we don't know exist.

One quick note - you should depend on (emacs "24.4") if you need `with-eval-after-load'.

equwal:
> > I'm considering removing the skeletons from this package, and trying to add them to a templating package like yasnippet instead. That would make my package only do one thing: dispatch the mode based on the format header.
> 
> Sorry for the slower turnaround -- that sounds like a good idea to me. You could emulate a package like [this one](https://github.com/toumorokoshi/go-snippets/blob/d437df148879566ffe7f2e503a3cf2602aa9fb28/go-snippets.el) for your skeletons. The benefit of relying on yasnippet for this is that it probably gracefully handles corner cases we don't know exist.

I ended up not using yasnippet, since it inserts static content only. I'm still using the built in Emacs `skeleton` stuff though. This way I can insert dynamic information like the date, and do error checking on the user's input.

> One quick note - you should depend on (emacs "24.4") if you need `with-eval-after-load'.
I managed to remove this entirely, I'll just make the users do stuff in their init. See the readme on [coleslaw-snippets](https://github.com/equwal/coleslaw-snippets).

riscy:
I think if they can be two separate packages, without even relying on each other, that's definitely your best bet.  And in that case, I would make a PR for that package separately.

As for this one,

```
coleslaw.el:90:41:Warning: reference to free variable `coleslaw-separator'

In end of data:
coleslaw.el:111:1:Warning: the function `coleslaw--dispatch' is not known to
    be defined.
```

I'm also noticing you have `(add-hook 'coleslaw-mode-hook 'coleslaw-dispatch)` but you already call `coleslaw--dispatch` in `coleslaw-mode`, so I think the hook should be removed.

After that I think we're good to go.

riscy:
Actually, sorry.  I'm looking at this a little more closely:

`coleslaw-dispatch` should only be called when the mode is being activated, but calling `coleslaw-mode` might turn the mode on or off.  You can check the value of the variable `coleslaw-mode` inside your `define-minor-mode` form to see whether the user is turning it off or on again, and behave accordingly.

riscy:
Let me see if I've got the workflow clear so I can be more helpful -- a Coleslaw user opens a `.page` or a `.post` file (these are the file suffices when using coleslaw).  By default, Emacs doesn't have a major mode for handling either of these types of files, so they just open in fundamental mode.  

Depending on the file's header, however, there is likely to be a mode that is more appropriate for editing the file.  So you supply a function that automatically enables the correct mode (either markdown or html or lisp or rst).  Is that about right?

purcell:
So [here](https://github.com/equwal/coleslaw/blob/master/coleslaw.el#L103-L107) you've got a local (ie. non-global) minor mode which - whether being toggled on or off - changes the global `auto-mode-alist`, and then tries to set the major mode for the current file. That's a little backwards.

Looking at this overall, I think that the whole package could be replaced by judicious entries in `magic-mode-alist`, a built-in var which Emacs uses to choose major modes based on file content. One format for that var allows entries of the type `(matcher-function . function)`, so you could provide a pair of `coleslaw-match` and `coleslaw-dispatch` functions and tell users to add those to `magic-mode-alist`. They could do that via `.dir-locals.el`, for example. And the two functions would use `coleslaw-modes` for the mapping. (Incidentally, it only makes sense to have one mode function for each format type.)

equwal:
> Looking at this overall, I think that the whole package could be replaced by judicious entries in `magic-mode-alist`, a built-in var which Emacs uses to choose major modes based on file content. One format for that var allows entries of the type `(matcher-function . function)`, so you could provide a pair of `coleslaw-match` and `coleslaw-dispatch` functions and tell users to add those to `magic-mode-alist`. They could do that via `.dir-locals.el`, for example. And the two functions would use `coleslaw-modes` for the mapping. (Incidentally, it only makes sense to have one mode function for each format type.)

Thanks for the feedback; I'll return if I have something more worthwhile to share. I didn't realise how easy it was to do this without having to make a mode for it.

equwal:
Coleslaw-mode has been rewritten. I hope it takes your feedback well into account:
- defcustom variables for defaults
- No need for `(coleslaw-setup)`. Instead, a user can include auto-insert, or define a key for insertion.
- This no longer is a "mode." No mode is defined.
- Magic alist is used to change to a decent major mode, or it is left as the normal-mode.
- A number of bugfixes; for example, newlines are placed after each form, instead of before, which keeps the cursor at the next point where the skeleton will insert.

equwal:
> So [here](https://github.com/equwal/coleslaw/blob/master/coleslaw.el#L103-L107) you've got a local (ie. non-global) minor mode which - whether being toggled on or off - changes the global `auto-mode-alist`, and then tries to set the major mode for the current file. That's a little backwards.
> 
Indeed. I've made this thing not even a mode anymore, so I may need to rename it (coleslaw-mode is not so good a name if it isn't even a mode).

> (Incidentally, it only makes sense to have one mode function for each format type.)
I decided to keep that syntax, since it allows for specifying a list of modes to try. If one mode is not installed, then use the next one, or the `normal-mode` at the end.

riscy:
Just rolling through:

No issues! 💯

```
1 issue found:

```

```
coleslaw.el:35:27:Warning: assignment to free variable `coleslaw-modes'
coleslaw.el:52:27:Warning: assignment to free variable `coleslaw-types'
coleslaw.el:61:27:Warning: assignment to free variable `coleslaw-separator'

In coleslaw--valid-format:
coleslaw.el:80:31:Warning: reference to free variable `coleslaw-modes'

In coleslaw--bufftypep:
coleslaw.el:96:12:Warning: reference to free variable `coleslaw-types'

In coleslaw--mode-regex:
coleslaw.el:190:11:Warning: reference to free variable `coleslaw-separator'
coleslaw.el:206:22:Warning: reference to free variable `coleslaw-modes'
```

No issues! 🎉

Other:
- coleslaw.el#L94: No need to sharp-quote lambdas (but also no harm)
- coleslaw.el#L150: No need to sharp-quote lambdas (but also no harm)
- coleslaw.el#L152: No need to sharp-quote lambdas (but also no harm)
- coleslaw.el#L211: It is beneficial to sharp-quote hooks; use `#'`
- Including coleslaw-pkg.el is discouraged -- MELPA will create one

equwal:
> ```
> 1 issue found:
> 
> 1:0: warning: The package summary should not end with a period.
> ```

- [X] I must not have this feature on my outdated version.

> 
> M-x byte-compile-file (using Emacs 26.3):
> 
> ```
> coleslaw.el:35:27:Warning: assignment to free variable `coleslaw-modes'
> coleslaw.el:52:27:Warning: assignment to free variable `coleslaw-types'
> coleslaw.el:61:27:Warning: assignment to free variable `coleslaw-separator'
> 
> In coleslaw--valid-format:
> coleslaw.el:80:31:Warning: reference to free variable `coleslaw-modes'
> 
> In coleslaw--bufftypep:
> coleslaw.el:96:12:Warning: reference to free variable `coleslaw-types'
> 
> In coleslaw--mode-regex:
> coleslaw.el:190:11:Warning: reference to free variable `coleslaw-separator'
> coleslaw.el:206:22:Warning: reference to free variable `coleslaw-modes'
> ```

- [X] Oops, should have paid better attention to that. Did not need to `setq` that variable. 

> 
> Other:
> 
>     * coleslaw.el#L94: No need to sharp-quote lambdas (but also no harm)

There is nothing here that looks like a sharp-quoted lambda.

> 
>     * coleslaw.el#L150: No need to sharp-quote lambdas (but also no harm)
> 
>     * coleslaw.el#L152: No need to sharp-quote lambdas (but also no harm)

- [X] Fixed those two.

> 
>     * coleslaw.el#L211: It is beneficial to sharp-quote hooks; use `#'`
> 
>    * Including coleslaw-pkg.el is discouraged -- MELPA will create one

- [X] Removed.

equwal:
Also [updated the docs](https://spensertruex.com/coleslaw).

PR #6233, opened 2019-06-18
Add recipe for helm-file-preview.

jcs090218:

Preview the current helm file selection.

https://github.com/jcs090218/helm-file-preview

The maintainer.

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/helm-file-preview
+++ b/recipes/helm-file-preview
@@ -1,0 +1,1 @@
+(helm-file-preview :repo "jcs090218/helm-file-preview" :fetcher github)


dzop:
I haven't looked at your code so I don't know what features you have in it, but are you aware of `helm-follow-mode` (activated with `C-c C-f` in `helm`)? Based off of your README and the animations in it, `helm-follow-mode` does everything that it seems you are trying to achieve with this package.

jcs090218:
I think there are few differences between this package and using `helm-follow-mode`. Correct me if I am wrong.

#### Configuration
* In `helm` - you use `helm-attrset` to add the source to enable `helm-follow-mode`. So every time if there are new source/package then you will need to add another source in your configuration.
* In this package - I wouldn't consider this is the best solution to you, but this is much better that works for me. Here is works across all packages that uses `helm` by certain rule, and not by source/command. I will no longer to need worry about what source I am currently using anymore, which is great cuz I don't to worry this again.

dzop:
> So every time if there are new source/package then you will need to add another source in your configuration.

I think this can be handled by setting `helm-follow-mode-persistent`. So you would have to enable `helm-follow-mode` for the source once and then that setting would be persisted across Emacs sessions without having to add anything to your configuration.

> works across all packages that uses helm by certain rule, and not by source/command.

I see. `helm-follow-mode` only works for sources that have a `persistent-action` which jumps to the file location. So I guess your rule would mean that the `persistent-action` of a source doesn't need to be looked at?

jcs090218:
> I see. helm-follow-mode only works for sources that have a persistent-action which jumps to the file location. So I guess your rule would mean that the persistent-action of a source doesn't need to be looked at?

Yeah, I think this isn't really matter to the result/action, but I reckon this is the other way to configure it and is much untie between sources and the feature itself. Since this package doesn't really care about the source.

purcell:
I can't really comment on whether this is redundant given `helm-follow-mode`, but I've gone ahead and merged it anyway. It might be worth noting the differences between the two in your README so that users can make the choice they prefer.

PR #6235, opened 2019-06-18
add date2name

DerBeutlin:

A package to manage ISO date and timestamps in files in dired.

https://github.com/DerBeutlin/date2name.el.git

Maintainer

None

### Checklist

Please confirm with `x`:


purcell:
Looks good, thanks: merged. You should add `;;;###autoload` cookie comments for the two interactive functions.

PR #6239, opened 2019-06-21
Add swift-playground-mode

msanders:

Emacs support for Swift playgrounds.

https://gitlab.com/michael.sanders/swift-playground-mode

Maintainer

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/swift-playground-mode
+++ b/recipes/swift-playground-mode
@@ -1,0 +1,5 @@
+(swift-playground-mode :repo "michael.sanders/swift-playground-mode"
+                       :fetcher gitlab
+                       :files (:defaults
+                               "SwiftPlayground.vim/runner.sh"
+                               "SwiftPlayground.vim/*.swift"))


msanders:
Thanks! All fixed.

purcell:
Thanks!
- For the `:files` spec in the recipe, I'd suggest `(:defaults "SwiftPlayground.vim/*.swift")`
- `seq` is not in Emacs 24.4. You'll need to depend on the ELPA package for that.
- Don't [hard-code `/tmp`](https://gitlab.com/michael.sanders/swift-playground-mode/blob/1626e5a0187e03e9e9079b74d7b4b6dfdd6270a1/swift-playground-mode.el#L190) - there's a built-in var for that.
- [Here](https://gitlab.com/michael.sanders/swift-playground-mode/blob/1626e5a0187e03e9e9079b74d7b4b6dfdd6270a1/swift-playground-mode.el#L217-244) you should be changing `after-save-hook` locally, not globally (see the LOCAL flag for `add-hook`), and then that way you don't need to check for the mode being enabled in the callback.
- [This](https://gitlab.com/michael.sanders/swift-playground-mode/blob/1626e5a0187e03e9e9079b74d7b4b6dfdd6270a1/swift-playground-mode.el#L246-253) could actually be a globalized minor mode.
- The `*-toggle-if-needed` function should really be called `*-enable-if-needed`, right?
- Prefer `when` to `if` in cases with no `else` form, e.g. [here](https://gitlab.com/michael.sanders/swift-playground-mode/blob/1626e5a0187e03e9e9079b74d7b4b6dfdd6270a1/swift-playground-mode.el#L127)
- Avoid explicit layout instructions when displaying buffers, e.g. [here](https://gitlab.com/michael.sanders/swift-playground-mode/blob/1626e5a0187e03e9e9079b74d7b4b6dfdd6270a1/swift-playground-mode.el#L67). There's `display-buffer-alist` for users to customise this.
- You don't need both `inhibit-read-only` and `read-only-mode` [here](https://gitlab.com/michael.sanders/swift-playground-mode/blob/1626e5a0187e03e9e9079b74d7b4b6dfdd6270a1/swift-playground-mode.el#L71-72) - just use the former.
- Why `setq-default` [here](https://gitlab.com/michael.sanders/swift-playground-mode/blob/1626e5a0187e03e9e9079b74d7b4b6dfdd6270a1/swift-playground-mode.el#L76)?
- Setting the [parent keymap](https://gitlab.com/michael.sanders/swift-playground-mode/blob/1626e5a0187e03e9e9079b74d7b4b6dfdd6270a1/swift-playground-mode.el#L53) is likely redundant: `swift-mode` will already be a prog mode.

msanders:
Thanks for the help, I've addressed the points outlined and added context below.

> Don't hard-code /tmp - there's a built-in var for that.

This was necessary to interact with the runner script, but I've updated it to read the value directly from the script output instead.

> This could actually be a globalized minor mode.

I looked into this but wasn't sure of the benefit, could you elaborate on what it would add?

> The *-toggle-if-needed function should really be called *-enable-if-needed, right?

This function was incomplete; I've fixed it to correctly close the output buffer when leaving playgrounds.

> Why setq-default here?

Looks like a mistake, changed to `setq` instead.

Thanks for the thorough review. Pretty new to elisp so the pointers were helpful.

riscy:
> I looked into this but wasn't sure of the benefit, could you elaborate on what it would add?

One benefit is that there would be an ergonomic way for your users to turn _off_ the hook if they so chose to.

Tangentially, I'm thinking the buffer-name-check you do feels a lot like something `auto-mode-alist` should be responsible for.  But I don't actually know whether adding minor modes to `auto-mode-alist` is correct.

msanders:
I've updated this to use a globalized minor mode. `auto-mode-alist` looked promising for the buffer name check, but I was unable to get that working without interfering with the major mode for swift files.

riscy:
> seq is not in Emacs 24.4. You'll need to depend on the ELPA package for that.

I think this one needs to be fixed still, e.g. with `;; Package-Requires: ((emacs "24.4") (seq "2.2.0")))`?

> You don't need both inhibit-read-only and read-only-mode here - just use the former.

This is fixed in terms of behavior, but I still have a couple nitpicks about the following:
```elisp
(save-excursion
          (let ((inhibit-read-only t))
            (erase-buffer)
            (funcall transform)
            (read-only-mode t)))
```
You probably don't need to `save-excursion` if you're just erasing the buffer; and at first glance it's confusing to me to set `read-only-mode` inside the `let` that would otherwise inhibit it.  What about:
```elisp
(let ((inhibit-read-only t))
  (erase-buffer)
  (funcall transform))
(read-only-mode t)
```
(Bearing in mind I might be missing some edge case).

Lastly, on line 119, you can sharp-quote the function (e.g. `(apply #'swift-playground--call-process executable args)`).  It's useful to do this because byte-compilation will tell you if that function name ever disappears.

msanders:
Thanks for the detailed feedback. I've applied the above improvements and confirmed it works correctly on Emacs 24.4.

riscy:
This looks great!  I'm going to go ahead and merge.  And, welcome to MELPA! :)

PR #6240, opened 2019-06-23
Add with-emacs

twlz0ne:

Evaluate Emacs Lisp expressions in a separate Emacs process

```elisp
(with-emacs
  :path "/path/to/version/emacs" ;; optional
  :lexical t                     ;;
  (do-something)
  ...)
```

https://github.com/twlz0ne/with-emacs.el

Author & Maintainer

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/with-emacs
+++ b/recipes/with-emacs
@@ -1,0 +1,3 @@
+(with-emacs
+ :fetcher github
+ :repo "twlz0ne/with-emacs.el")


purcell:
This is cool, and I'll merge it as-is, but I expect that it would be useful to put most of the code from the macro into a function so that it can be used directly when helpful, and then just use the macro to provide convenient syntax sugar for inline evaluation.

PR #6241, opened 2019-06-23
New package: execline

KAction:
This package provides major mode for editing execline scripts.

This package provides major mode for editing [execline](http://skarnet.org/software/execline/) scripts.

https://gitlab.com/KAction/emacs-execline

I am maintainer of package.

**None needed** 

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/execline
+++ b/recipes/execline
@@ -1,0 +1,1 @@
+(execline :fetcher gitlab :repo "KAction/emacs-execline")


alphapapa:
"Page not found"

KAction:
Fixed. gitlab, not github.

riscy:
This looks solid.  Browsing around tells me the `.ex` extension is (also) associated with Elixir as well as another language I just discovered called Euphoria.  Maybe there's nothing to be done about that, but I notice that the execline page you link contains source code with a `.txt` extension!  Even the C repo (https://github.com/skarnet/execline) doesn't seem to have any files with extensions.

So, just to clarify, is `.ex` the standard execline extension?

purcell:
Thanks! Couple of notes:
- You should modify the buffer-local version of `completion-at-point-functions` [here](https://gitlab.com/KAction/emacs-execline/blob/deafc87649fa135e030c1b3bfc11b6488ff3fdba/execline.el#L210) rathter than the global version
- Use `\\'` and not `$` to match end-of-string in the `auto-mode-alist` entry [here](https://gitlab.com/KAction/emacs-execline/blob/deafc87649fa135e030c1b3bfc11b6488ff3fdba/execline.el#L220)
- There's a built-in `bolp` function you could use [here](https://gitlab.com/KAction/emacs-execline/blob/deafc87649fa135e030c1b3bfc11b6488ff3fdba/execline.el#L176)

Hope that helps!

KAction:
[2019-06-29 20:50] Chris Rayner <notifications@github.com>
> So, just to clarify, is `.ex` the standard execline extension?

No. As far as I know, there is no standard extension. If `.ex` conflicts
with other languages, I can use another extension, like `.exec'.

Another approach would be to detect major mode based on shebang. It will
work for editing scripts, but not for freshly-created.
-- 
Note, that I send and fetch email in batch, once in a few days.
Please, mention in body of your reply when you add or remove recepients.

KAction:
[2019-06-30 14:49] Steve Purcell <notifications@github.com>
> Thanks! Couple of notes:
> [...]

Thank you. I applied your suggestions and pushed v0.5 tag.
-- 
Note, that I send and fetch email in batch, once in a few days.
Please, mention in body of your reply when you add or remove recepients.

riscy:
> No. As far as I know, there is no standard extension. If `.ex` conflicts with other languages, I can use another extension, like `.exec'.

I think this'd be a good idea since .ex seems to have been claimed by Elixir. :\  Maybe this is your big chance to claim a unique extension for execline, at least in among Emacs users. :)

riscy:
There's a comment [here](https://gitlab.com/KAction/emacs-execline/blob/master/execline.el#L33) that needs updating re: `.exec` instead of `.ex` files, but that's not a blocker.  Altogether this looks great - welcome to MELPA!

KAction:
[2019-07-04 15:01] Chris Rayner <notifications@github.com>

> > No. As far as I know, there is no standard extension. If `.ex`
> > conflicts with other languages, I can use another extension, like `.exec'.

> I think this'd be a good idea since .ex seems to have been claimed by
> Elixir. :\  Maybe this is your big chance to claim a unique extension
> for execline, at least in among Emacs users. :)

Just did it. .ex -> .exec; pushed as tag v0.6
-- 
Note, that I send and fetch email in batch, once in a few days.
Please, mention in body of your reply when you add or remove recepients.

PR #6242, opened 2019-06-23
Add recipe for alt-codes.

jcs090218:

Insert alt codes using meta key.

https://github.com/jcs090218/alt-codes

The maintainer.

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/alt-codes
+++ b/recipes/alt-codes
@@ -1,0 +1,1 @@
+(alt-codes :repo "jcs090218/alt-codes" :fetcher github)


purcell:
Thanks! Agree with the note about making the codes a variable.

Also I wonder if this is the sort of thing that could be written as a quail input method, which is a more standard way to do customised input. Worth looking into, I'd say.

jcs090218:
@riscy Thanks! As always, you comments is very helpful! :D 

@purcell I will definitely look into this! Thank you!

PR #6243, opened 2019-06-23
adding new recipe for new library icsql

plandes:

This library provides an Emacs SQL mode integration to the ciSQL program.  In addition it provides support for multiple buffers and fast/easy buffer (for each SQL connection) using the buffer manage library.

https://github.com/plandes/icsql

Author and maintainer

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/icsql
+++ b/recipes/icsql
@@ -1,0 +1,1 @@
+(icsql :fetcher github :repo "plandes/icsql")


plandes:
I've made the recommended changes per `package-lint`, it byte compiles clean for me and `check-declare-file` provides nothing.

purcell:
Thanks! Couple of notes:
- I suspect the filename handling is unsafe where you're building up curl commands etc. Imagine working directories or filenames containing spaces and/or backslashes. Judicious use of `shell-quote-argument` is necessary here. You've got some ad-hoc wrapping with single quotes, but that's not safe.
- Prefer `unless` [here](https://github.com/plandes/icsql/blob/a8f5965a88eb823e4e4977b9c64b4fb44a1c5b2a/icsql.el#L168)
- Use `line-beginning-position` and `line-end-position` [here](https://github.com/plandes/icsql/blob/a8f5965a88eb823e4e4977b9c64b4fb44a1c5b2a/icsql.el#L283-L288)
- Instead of [these forms](https://github.com/plandes/icsql/blob/a8f5965a88eb823e4e4977b9c64b4fb44a1c5b2a/icsql.el#L377-L379), define a little minor mode which users can enable in `sql-mode` as desired: it can perform this setup. Otherwise you've got a file which, when loaded, has side effects that might be hard to customise or reason about.
- And in contrast, it feels like the code in `icsql-repopulate-sql-product-alist` should just be called once, globally. Feels odd to me that it's getting called again and again in different buffers.
- Don't use [`wget` and `curl`](https://github.com/plandes/icsql/blob/a8f5965a88eb823e4e4977b9c64b4fb44a1c5b2a/icsql.el#L142-L145) when you can just use the `url-copy-file` built-in instead.

Hope that helps!

plandes:
Hi Steve,

I've made changes that reflect your suggestions except for using `url-copy-file`, which currently creates a 0 byte file because (my guess) GitHub now hosts release files on AWS.  This is implemented as a redirect, and it appears I'm [not the only one](https://emacs.stackexchange.com/questions/48004/why-doesnt-url-copy-file-download-these-files) that's experienced this issue.  I got to the point of debugging `url-retrieve-synchronously` and gave up after Emacs forks a process.

riscy:
I can confirm this is an issue -- I think we should go ahead and merge and hope the redirect issue gets fixed down the line.  There is `w3m-download`, but of course that expects you to have `w3m` installed, which is a bigger expectation than having either `curl` or `wget` installed.

PR #6246, opened 2019-06-26
Add symex-mode recipe

countvajhula:

Symex mode is a Vim-like way to edit Lisp code, as an alternative to packages like paredit and lispy.

https://github.com/countvajhula/symex-mode

I am the author.

**None needed**

### Checklist

Please confirm with `x`:

- [x] My elisp byte-compiles cleanly [couple of warnings but nothing important]


purcell:
I'll have a more detailed look soon, but you'd definitely want to declare a package dependency on `evil`. Best not to rely on transitive dependencies.

countvajhula:
Thanks for the feedback @riscy ! I'm traveling on vacation at the moment, but at first glance, those error messages do look legitimate, not sure why I didn't see them when I ran through the steps with byte-compiling and flycheck (there were other errors, now fixed). I must have missed something. Will take a look soon as I'm back next week.

countvajhula:
Hi @riscy,
I went through the 3 steps of `package-lint`, `checkdoc`, and byte-compiling again and made some progress, but I'm still not seeing all of the errors you found above:

First, the package-lint errors you pointed out weren't showing up for me in flycheck mode, but running the checker at the command line did the trick, i.e. via e.g.:
```
INIT_PACKAGE_EL="(progn
  (require 'package)
  (push '(\"melpa\" . \"http://melpa.org/packages/\") package-archives)
  (package-initialize))"

/usr/local/bin/emacs -Q -batch --eval "$INIT_PACKAGE_EL" -l package-lint.el -f package-lint-batch-and-exit ./sandbox/elpa/symex-mode-20190626.1143/symex-utils.el || [ -n "${EMACS_LINT_IGNORE+x}" ]
```
[from the example in https://github.com/purcell/package-lint/blob/master/run-tests.sh]

This still produces some errors like
```
  at 168:1: error: "symex-make-detour" doesn't start with package's prefix "symex-data".
```
... but these don't appear to be legitimate errors. I think these may be because it is a multi-file package, which package-lint may not be handling correctly based on what [this README](https://github.com/purcell/package-lint/blob/master/README.md) says.

For checkdoc, I think all the errors should be addressed now.

For the byte compile output, what's the recommended way to do that? At the command line I tried:

```
/usr/local/bin/emacs --batch --eval "$INIT_PACKAGE_EL" --eval '(byte-recompile-directory "/Users/siddhartha/work/emacs/melpa/sandbox/elpa/symex-mode-20190711.1305" 0 t)'
```

... and that gives some of the errors but generally doesn't seem to work right, giving messages like:
```
In toplevel form:
symex-transformations.el:29:1:Error: Cannot open load file: No such file or directory, symex-utils
Compiling /Users/siddhartha/work/emacs/melpa/sandbox/elpa/symex-mode-20190712.15/symex-traversals.el...
```
But byte-compiling from within Emacs via

```
(byte-recompile-directory "/Users/siddhartha/work/emacs/melpa/sandbox/elpa/symex-mode-20190712.15" 0 t)
```

(which is how I did it the first time around) seems to work and now produces clean output:
```
Compiling file /Users/siddhartha/work/emacs/melpa/sandbox/elpa/symex-mode-20190712.15/symex-computations.el at Fri Jul 12 00:20:26 2019
Entering directory ‘/Users/siddhartha/work/emacs/melpa/sandbox/elpa/symex-mode-20190712.15/’

Compiling file /Users/siddhartha/work/emacs/melpa/sandbox/elpa/symex-mode-20190712.15/symex-data.el at Fri Jul 12 00:20:26 2019

Compiling file /Users/siddhartha/work/emacs/melpa/sandbox/elpa/symex-mode-20190712.15/symex-evaluator.el at Fri Jul 12 00:20:26 2019

Compiling file /Users/siddhartha/work/emacs/melpa/sandbox/elpa/symex-mode-20190712.15/symex-interface-elisp.el at Fri Jul 12 00:20:26 2019

Compiling file /Users/siddhartha/work/emacs/melpa/sandbox/elpa/symex-mode-20190712.15/symex-interface-racket.el at Fri Jul 12 00:20:26 2019

Compiling file /Users/siddhartha/work/emacs/melpa/sandbox/elpa/symex-mode-20190712.15/symex-interface-scheme.el at Fri Jul 12 00:20:26 2019

Compiling file /Users/siddhartha/work/emacs/melpa/sandbox/elpa/symex-mode-20190712.15/symex-misc.el at Fri Jul 12 00:20:26 2019

Compiling file /Users/siddhartha/work/emacs/melpa/sandbox/elpa/symex-mode-20190712.15/symex-mode.el at Fri Jul 12 00:20:26 2019

Compiling file /Users/siddhartha/work/emacs/melpa/sandbox/elpa/symex-mode-20190712.15/symex-primitives.el at Fri Jul 12 00:20:26 2019

Compiling file /Users/siddhartha/work/emacs/melpa/sandbox/elpa/symex-mode-20190712.15/symex-transformations.el at Fri Jul 12 00:20:26 2019

Compiling file /Users/siddhartha/work/emacs/melpa/sandbox/elpa/symex-mode-20190712.15/symex-traversals.el at Fri Jul 12 00:20:26 2019

Compiling file /Users/siddhartha/work/emacs/melpa/sandbox/elpa/symex-mode-20190712.15/symex-utils.el at Fri Jul 12 00:20:26 2019
```
I think it's still missing some errors though, since I didn't see the ones re: "free variables" at any point, even though they show up in your output and may be legitimate complaints. From some quick googling it sounds like adding `(defvar var-name)` is the standard way to handle that error, but I'd ideally like to see the error actually show up before fixing it, so I can be sure I've handled all the instances.

countvajhula:
Ok, I found the issue. I just needed to add the symex package dir to the load path, and then I got the same errors you did. The byte compile warnings should all be fixed now 👍 , please lmk if there's anything else. FTR this is how I ran the byte compiler from the command line:

```
INIT_PACKAGE_EL="(progn
  (require 'package)
  (push '(\"melpa\" . \"http://melpa.org/packages/\") package-archives)
  (setq my-package-dir \"/Users/siddhartha/work/emacs/melpa/sandbox/elpa/\")
  (add-to-list 'load-path my-package-dir)
  (let ((default-directory my-package-dir))
    (normal-top-level-add-subdirs-to-load-path))
  (package-initialize))"

/usr/local/bin/emacs -Q --batch --eval "$INIT_PACKAGE_EL" --eval '(byte-recompile-directory "/Users/siddhartha/work/emacs/melpa/sandbox/elpa/symex-mode-20190712.1455" 0 t)'
```

countvajhula:
Appreciate the detailed review @riscy , I've made all the changes you suggested. Re: keybindings, yeah I think having a minor mode is the right solution there, so I've gone ahead and added it. It does make things a little unusual, with `symex-mode` actually enabling the minor mode rather than pulling up the hydra menu / entering the evil state or "mode" (in Vim terms) which is the main thing provided by the symex mode package, necessitating a separate entry point for this purpose (which I've named `symex-mode-interface`). But I think that's probably to be expected since symex mode is modeled after the Vim idea of a mode, i.e. an evil state, and yet as an emacs package it needs to provide additional safeguards in other evil states like insert state and Emacs state (or just plain Emacs for non-evil users). This latter requirement does seem to imply a minor mode in addition to the evil state, and it also simplifies managing the keybindings and avoids the contention with other modes that you pointed out.

Re: going overboard with package lint, does that also apply to `Package-Requires`? And does it matter which file contains the `Version` header? I've kept it in `symex-mode.el` and removed it from the others.

riscy:
> Re: going overboard with package lint, does that also apply to Package-Requires? And does it matter which file contains the Version header? I've kept it in symex-mode.el and removed it from the others.

It's typically in the "core" file, or the file that all the other files require.  MELPA doesn't pay much attention to the version number (you need to tag stable releases instead).  Similarly with the `Package-Requires`.  You can look at [this package](https://github.com/purcell/color-theme-sanityinc-solarized) for example.

riscy:
Sorry for the multiple passes. I've gotten a chance to do some user testing.  One thing I'm noticing that the hydra is _huge_:
image
It's not easy to see the full list of commands even with a relatively large Emacs terminal.  Maybe this could be fixed by reducing the number of columns, or perhaps by shortening some of the descriptive text a bit (e.g. "go to outermost" could be "outermost" or "escape to higher level" could be "escape up" or "wrap with ()" could just be "wrap" since the keybinding kind of implies what we want to wrap it with, etc.).

- Now that I'm actually using the package, it feels like [this](https://github.com/countvajhula/symex-mode/blob/8cc74f5a760a3164052a5a20bcdb603a53a6303e/symex-mode.el#L97-L128) keybinding should already be handled by smartparens, unless I'm missing something.  Perhaps leave it up to the user to enable smartparens (or not)?
- I'm also noticing that `racket--send-to-repl` should probably start with your own package's prefix, rather than racket's.
- As a small suggestion, what if you called the core file "symex.el" instead of "symex-mode.el"?  That way your users would have to run `(require 'symex)` instead of `(require 'symex-mode)` or <kbd>M-x load-lib symex</kbd> instead of <kdb>M-x load-lib symex-mode</kbd>

countvajhula:
> Sorry for the multiple passes.
No worries, I appreciate your commitment!

OK, `Package-Requires` also removed, `racket--` function renamed, and package renamed to `symex.el` (I like it!).

Re: smartparens, I did look into it and it's not exactly what I'm looking for since it doesn't include the backspace behavior and also allows closing delimiters to be added even if there isn't a matching opening one. Paredit handles this well. I guess we could rely on users enabling paredit (or not) in Lisp-related buffers, but I wonder if that's too heavyweight since all I'm looking for here is the delimiter behavior, which conceivably could be done within symex-mode someday (I plan to reduce reliance on paredit/lispy over time, that is, once I add an explicit AST representation to symex.el which would improve precision overall). Breaking the tree structure of the symex with manual edits also causes some problems in symex mode, so mandating the structure-preserving delimiter behavior is somewhat necessary at the present time rather than a nice-to-have.

Re: the hydra, yes, I'm aware its size would perhaps inspire terror in Hercules himself! The thing though is that the hydra is sort of a necessary evil at this point, so to speak. I'd ideally like to do away with the menu altogether and bring it up on demand, but that's not something that hydra supports at the moment. And even more ideally I'd like to implement the modal interface entirely in evil since that's what evil is for, and maybe launch the hydra (or some other menu) on demand. Unfortunately my attempts at doing it in evil have failed thus far, because evil's "operator state" seems to be hardcoded for standard evil/vim-style motions, and doesn't operate on motions specified in a custom evil state. It'll take some time to understand this more and see if there is a workable solution, but it's possible that a clean solution here would require changes in evil itself, so at any rate it's a little ways out.

Personally I don't even refer to the hydra menu in my usage and it's mostly just there because it's part of how hydra works, and also as trainer wheels which are always nice.

Btw, I'm giving a [talk about symex.el](https://www.meetup.com/Emacs-SF/events/262446850/) (getting used to the new name :)) in a couple of days and it'd be nice if attendees could install it and try it out, do you think merging it as is would be acceptable? I could continue to address things as time goes on and your feedback is definitely super helpful. Totally fine either way though. What do you think?

riscy:
Yeah, that all makes sense to me.  Thanks for your patience with all the linting -- this now looks good to merge right now, in time for your talk. :)  Welcome to MELPA!

countvajhula:
Woot 🎉Thanks for your help @riscy !

tarsius:
@countvajhula Above you have stated that

> [checkmark] The package is released under a GPL-Compatible Free Software License.

I have some doubts about that.

> ;; This program is "part of the world," in the sense described at
> ;; http://drym.org.  From your perspective, this is no different than
> ;; MIT or BSD or other such "liberal" licenses that you may be
> ;; familiar with, that is to say, you are free to do whatever you like
> ;; with this program.  It is much more than BSD or MIT, however, in
> ;; that it isn't a license at all but an idea about the world and how
> ;; economic systems could be set up so that everyone wins.  Learn more
> ;; at drym.org.

I didn't see anything resembling a software license at http://drym.org, then again I did not feel like reading all of that just to make sure.

@purcell What to do about this?

countvajhula:
@tarsius Do you have specific concerns? Happy to address them.

countvajhula:
The work being "part of the world" means that it is treated as part of your "perceptions" in a generalized sense. Since no one should have control over what you do with your perceptions, in the same manner, no one can control what you do with this work. In particular, no one owns this work either, nor does anyone (e.g. me) own copyright on it. The system proposed at drym.org is intended to transcend traditional legal systems (including differentiating ownership from property -- revealing the former to be an illusory idea for which the latter is sufficient to facilitate equivalencies in all cases), hence there is no need for a license since a license presupposes ownership. For the purposes of traditional legal systems, it can be treated as public domain. HTH.

alphapapa:
> The work being "part of the world" means that it is treated as part of your "perceptions" in a generalized sense. Since no one should have control over what you do with your perceptions, in the same manner, no one can control what you do with this work. In particular, no one owns this work either, nor does anyone (e.g. me) own copyright on it. The system proposed at drym.org is intended to transcend traditional legal systems (including differentiating ownership from property -- revealing the former to be an illusory idea for which the latter is sufficient to facilitate equivalencies in all cases), hence there is no need for a license since a license presupposes ownership. 

MELPA and its users operate in the real world and its existing legal systems.

> For the purposes of traditional legal systems, it can be treated as public domain. HTH.

Then please state so in the code.  For example, Elfeed does this: https://github.com/skeeto/elfeed

riscy:
I'm noting this is something I need to pay closer attention to, too.

countvajhula:
I've added an explicit [public domain dedication](https://github.com/countvajhula/symex.el/blob/master/symex.el#L17-L24) in the header. I kept it simple and modified some of the verbiage from your link @alphapapa to indicate that the author relinquishes any claim to copyright rather than that the copyright is dedicated to the public domain (the latter implying that the author accepts the copyright where it applies -- which my work does not since it is intended to supplant copyright and other such legal implements and it would be incorrect to say this). As always, it's good to see the careful attention to detail around MELPA maintenance. Please let me know if there's anything else.

tarsius:
This works for me. Thanks!

PR #6250, opened 2019-06-30
add flymake-go-staticcheck recipe

s-kostyaev:

Flymake backend for Go using [staticcheck](https://github.com/dominikh/go-tools/tree/master/cmd/staticcheck).

https://github.com/s-kostyaev/flymake-go-staticcheck

I am the author.

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/flymake-go-staticcheck
+++ b/recipes/flymake-go-staticcheck
@@ -1,0 +1,1 @@
+(flymake-go-staticcheck :repo "s-kostyaev/flymake-go-staticcheck" :fetcher github)


riscy:
Just noticing that `checkdoc` gives me:
```
Warning (emacs): 
flymake-go-staticcheck.el:63: Some lines are over 80 columns wide
Warning (emacs): 
flymake-go-staticcheck.el:87: Some lines are over 80 columns wide
```
In the case of the first, some careful rewriting could make the first line a complete sentence.  In the case of the second, you can just run `M-q` on the docstring.

Here:
```elisp
(add-hook 'flymake-diagnostic-functions 'flymake-go-staticcheck--checker nil t))
```
You can actually sharp-quote the `#'flymake-go-staticcheck` which will make the compiler notice if you ever change the name of that function.

Very minor, but this line [here](https://github.com/s-kostyaev/flymake-go-staticcheck/blob/master/flymake-go-staticcheck.el#L67) could just be
```elisp
    (let (results) ...
```
since all you're after is an empty list (i.e. nil).

Also very minor, maybe it is time we began using `https` to link to gnu.org's licenses (in your frontmatter).  It's not a big deal but I thought I'd start pointing it out to people, since I think everyone copy-pastes this from each other.

s-kostyaev:
@riscy thanks for review. Fixed.

riscy:
- `flymake-go-staticcheck` feels like it's four words, not three -- why not `flymake-go-static-check`?  Or, the `check` is sort of implied.  What about `flymake-go-static`?
- `error` takes a string as an argument [here](https://github.com/s-kostyaev/flymake-go-staticcheck/blob/805e81282e6fd57e7cd564a823a9d797c2eabe8f/flymake-go-staticcheck.el#L59)- no need to wrap that string with `message`.
- You'll want to capitalize that string for checkdoc's sake -- and maybe you should be telling the user to run `customize-variable` instead of `set-variable`?
- `#'flymake-go-staticcheck--checker` is private by convention (contains `--`) but the expectation is to add it to a hook -- it should be a public function.
- Many people using flymake would probably prefer to simply add the hook to their configuration file -- should doing so be added to your documentation? Also, using `flymake-go-staticcheck-enable` in a configuration could cause strange things like flymake being enabled on the *scratch* buffer.  Perhaps leave it up to the user to enable `flycheck-mode` after calling that function?
- I don't know a lot about writing flymakes, but have you considered using [flymake-easy](https://github.com/purcell/flymake-easy)?

s-kostyaev:
staticcheck is the name of the linter.

s-kostyaev:
@riscy thanks for review. Fixed.

riscy:
Awesome, thank you!

PR #6251, opened 2019-06-30
Add recipe for org-json

jlumpe:

Encodes the abstract syntax tree of an org mode file in JSON format.

https://github.com/jlumpe/org-json

Author/maintainer

None

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/org-json
+++ b/recipes/org-json
@@ -1,0 +1,1 @@
+(org-json :fetcher github :repo "jlumpe/org-json")


alphapapa:
Please do change the name to e.g. `org-export-json`, since its purpose is exporting.

You might also consider refactoring it to define a derived export backend, which is the standard way to do it in Org.  Then your package could be called, e.g. `ox-json`, like other Org export backends.

jlumpe:
Hmm, I had originally wanted to do that, but I wanted to eventually make it extensible with the ability to add post-processing functions which can act on the pre-encoded output before it is converted to a string (unlike the filter system I think, which acts only on encoded output?). After looking a bit more through the developer docs it does look way more extensible in general, though. It probably shouldn't be too much effort.

jlumpe:
Replaced with  #6280

PR #6255, opened 2019-07-01
Add subed

rndusr:

subed is a major mode for editing SubRip/srt subtitles. 

https://github.com/rndusr/subed/

I'm the maintainer.

None needed

### Checklist

- [ ] I have confirmed some of these without doing them (Sneaky! I like it.)

### DIFF
--- a/recipes/subed
+++ b/recipes/subed
@@ -1,0 +1,3 @@
+(subed :fetcher github
+       :repo "rndusr/subed"
+       :files ("subed/*.el"))


rndusr:
Thank you for taking a look at my code.  I found a bug as a result.

> [...]:Warning: reference to free variable [...]

How do I fix this?  If I use my regular Emacs configuration, `byte-compile-file`
runs without warnings.  If I use `emacs -Q`, I can reproduce the warnings, but
requiring subed-config.el results in an error because it can't find that file.

I've built and installed the package as documented in CONTRIBUTING and
everything seemed to work fine and I didn't get any warnings.

> In `subed-mpv.el`:
> <kbd>M-x byte-compile-file</kbd>:
> ```
> subed-mpv.el:217:1:Warning: Unused lexical variable `err'
> ```
> Regarding `err` on line 217, you can just prefix the variable as `_err` to give the compiler and your users a hint.

Just `nil` instead of `err` seems to work, too, and it seems to be intended.
`C-h f condition-case`: "(If VAR is nil, the handler can’t access that
information.)"  Am I misunderstanding how that works?

I think I've fixed or will fix the other issues you found.
Thanks again.

riscy:
>> [...]:Warning: reference to free variable [...]
>How do I fix this?  If I use my regular Emacs configuration, `byte-compile-file` runs without warnings.  If I use `emacs -Q`, I can reproduce the warnings, but requiring subed-config.el results in an error because it can't find that file.

I think you're already on top of this one, but let me know if not.

You're right about `_err` -- sorry about that. :)

rndusr:
On Sun, Jul 07, 2019 at 09:18:58PM -0700, Chris Rayner wrote:
> >> [...]:Warning: reference to free variable [...]
> >How do I fix this?  If I use my regular Emacs configuration, `byte-compile-file` runs without warnings.  If I use `emacs -Q`, I can reproduce the warnings, but requiring subed-config.el results in an error because it can't find that file.
> 
> I think you're already on top of this one, but let me know if not.

Sorry, I'm afraid I still don't have a clue. (This is my first Elisp project.)

Does it have something to do with autoloading?

riscy:
If the changes I made [here](https://github.com/rndusr/subed/pull/8) make it, you should be all set with the checklist.

However, the "abstraction layer" you define is better handled through customization options.  That said, do you plan on making heavy use of that abstraction layer in the near future?  Otherwise simpler is often better, especially as you gain experience.

rndusr:
> However, the "abstraction layer" you define is better handled through customization options.  That said, do you plan on making heavy use of that abstraction layer in the near future?  Otherwise simpler is often better, especially as you gain experience.

I definitely want to make it as easy as possible to support other subtitle
formats.  I'm not sure yet if I'll implement any myself, but the easier it is
the more likely someone (me or a contributor) will do it.

My idea was to load the relevant functions based on the file format.  If I open
a srt file, `subed-forward-subtitle-text` is an alias for
`subed-srt--forward-subtitle-text`.  If I open a ass file,
`subed-forward-subtitle-text` is an alias for
`subed-ass--forward-subtitle-text`.  I want this to be fully automatic and the
user shouldn't need to know `subed-*--forward-subtitle-text` exist.

But I'm open to different approaches.

riscy:
The problem with using `defalias` is if the user has an "ass" file and a "srt" open at the same time (maybe one is a buffer in the background that the user plans to return to at some time), then these re-definitions will collide and create unexpected behavior for the user.

I'll experiment and I'll make a PR for your consideration.

riscy:
I know we're having a discussion in the PR mentioned above, but feel free to ping here when you've arrived at something you can live with. :)

rndusr:
I will. But it's probably going to be a few weeks until I have time again.

riscy:
Ping! :)

rndusr:
Yeah, sorry about the delay.

I think I've found a neat-ish solution for multiple formats: Each format
implementation gets an init function that sets the buffer-local variable
subed-subtitle-format. The generic functions then use that variable to assemble
the function name, and to look up the format-specific
function. E.g. subed-subtitle-id does (concat "subed-" subed-subtitle-format
"--subtitle-id").

I've been doing a little work on that and it seems to work. I just have to
finish it.

I'll have to look again into the other stuff I have to fix for MELPA. I'll
probably contact you again when I'm at that point.

Thank you for your patience.

rndusr:
I think I found an acceptable solution.  Would you please take a look and let me
know if you agree?

Is there anything else I should do?

riscy:
Taking a look now!  Thanks, I definitely thing this is going to be a strong addition.

I see some strong improvements in the code -- but one thing that pops into focus is it seems like the functions `subed-mode`/`subed-mode-enable`/`subed-mode-disable` comprise turning a major mode off and on.  I can see that `subed-mode-enable` is doing things like killing the local variables, setting the keymap, defining the major-mode variable, etc.  I would recommend using `define-derived-mode`:

```elisp
(define-derived-mode subed-mode text-mode "Subed"
  "Docstring...

\\{subed-mode-map}"
  ;; add special hooks, configure local variables, etc.
  ;; no need to setq major mode, or have a variable called
  ;; `subed-mode--enabled-p` -- it's typical to just
  ;; check (eq major-mode 'subed-mode) -- and
  ;; no need to explicitly run hooks
  ;; no need to kill local variables, I believe
)
```
I probably haven't been thorough there, but there are some reasonable tutorials out there about standard ways of declaring major modes.  I'm suggesting text-mode as the base mode from which you derive, but perhaps this input is more programmatic and you'd prefer to derive from prog-mode.

In fact, I'd recommend looking at another established mode for examples.  `json-mode` is one of the smaller major modes that's well established, and `yaml-mode` is quite small as well (although these are programming modes).

As for byte-compiling, I don't mean to slam you with these, but there are quite a few warnings, some of which might point to bugs (e.g. subed-sort not being defined), some of which might indicate missing `require`'s.  I leave these to the end.

One thing that might help you at this early juncture is to inline all of the subed files into a single file -- then you only have to check byte-compilation on a single file and package-lint will give far fewer false-positives on things like package prefix.

Hopefully that's helpful!

```
subed.el:200:1:Warning: the function `subed-sort' is not known to be defined.
subed-mpv.el:358:1:Warning: the following functions are not known to be
    defined: subed-disable-sync-point-to-player-temporarily,
    subed-subtitle-msecs-start
subed-srt.el:356:54:Warning: reference to free variable
    `subed-default-subtitle-length'
subed-srt.el:550:15:Warning: assignment to free variable
    `subed--subtitle-format'
subed-srt.el:557:1:Warning: the following functions are not known to be
    defined: subed--right-pad, subed-save-excursion
subed-common.el:35:11:Warning: reference to free variable
    `subed-debugging-enabled-p'
subed-common.el:36:11:Warning: assignment to free variable
    `subed-debugging-enabled-p'
subed-common.el:38:44:Warning: reference to free variable `subed-debug-buffer'

In subed-disable-debugging:
subed-common.el:50:9:Warning: reference to free variable
    `subed-debugging-enabled-p'
subed-common.el:51:11:Warning: assignment to free variable
    `subed-debugging-enabled-p'
subed-common.el:53:44:Warning: reference to free variable `subed-debug-buffer'

In subed-toggle-debugging:
subed-common.el:61:7:Warning: reference to free variable
    `subed-debugging-enabled-p'

In subed-debug:
subed-common.el:67:21:Warning: reference to free variable `subed-debug-buffer'

In subed-adjust-subtitle-time-start:
subed-common.el:176:28:Warning: reference to free variable
    `subed-enforce-time-boundaries'
subed-common.el:185:52:Warning: reference to free variable
    `subed-subtitle-spacing'

In subed-adjust-subtitle-time-stop:
subed-common.el:221:28:Warning: reference to free variable
    `subed-enforce-time-boundaries'
subed-common.el:225:53:Warning: reference to free variable
    `subed-subtitle-spacing'

In subed-copy-player-pos-to-start-time:
subed-common.el:292:14:Warning: reference to free variable
    `subed-mpv-playback-position'

In subed-copy-player-pos-to-stop-time:
subed-common.el:301:14:Warning: reference to free variable
    `subed-mpv-playback-position'

In subed--insert-subtitle-info:
subed-common.el:499:34:Warning: reference to free variable
    `subed-default-subtitle-length'
subed-common.el:504:72:Warning: reference to free variable
    `subed-subtitle-spacing'
subed-common.el:520:1:Warning: Unused lexical variable `buffer-is-empty'

In subed-insert-subtitle:
subed-common.el:555:57:Warning: reference to free variable
    `subed-default-subtitle-length'
subed-common.el:578:1:Warning: Unused lexical variable `msecs-avail'

In subed-replay-adjusted-subtitle-p:
subed-common.el:634:45:Warning: reference to free variable
    `subed-subtitle-time-adjusted-hook'

In subed-enable-replay-adjusted-subtitle:
subed-common.el:643:60:Warning: reference to free variable
    `subed-subtitle-time-adjusted-hook'

In subed-disable-replay-adjusted-subtitle:
subed-common.el:654:61:Warning: reference to free variable
    `subed-subtitle-time-adjusted-hook'

In subed-sync-point-to-player-p:
subed-common.el:675:41:Warning: reference to free variable
    `subed-mpv-playback-position-hook'

In subed-enable-sync-point-to-player:
subed-common.el:684:67:Warning: reference to free variable
    `subed-mpv-playback-position-hook'

In subed-disable-sync-point-to-player:
subed-common.el:695:68:Warning: reference to free variable
    `subed-mpv-playback-position-hook'

In subed-disable-sync-point-to-player-temporarily:
subed-common.el:729:9:Warning: assignment to free variable
    `subed--point-was-synced'
subed-common.el:731:9:Warning: reference to free variable
    `subed--point-was-synced'
subed-common.el:733:24:Warning: reference to free variable
    `subed-point-sync-delay-after-motion'

In subed-sync-player-to-point-p:
subed-common.el:743:41:Warning: reference to free variable
    `subed-subtitle-motion-hook'

In subed-enable-sync-player-to-point:
subed-common.el:753:67:Warning: reference to free variable
    `subed-subtitle-motion-hook'

In subed-disable-sync-player-to-point:
subed-common.el:764:68:Warning: reference to free variable
    `subed-subtitle-motion-hook'

In subed--sync-player-to-point:
subed-common.el:780:16:Warning: reference to free variable
    `subed-mpv-playback-position'

In subed-loop-over-current-subtitle-p:
subed-common.el:792:7:Warning: reference to free variable
    `subed--subtitle-loop-start'
subed-common.el:792:34:Warning: reference to free variable
    `subed--subtitle-loop-stop'

In subed-enable-loop-over-current-subtitle:
subed-common.el:803:43:Warning: reference to free variable
    `subed--subtitle-loop-start'
subed-common.el:803:70:Warning: reference to free variable
    `subed--subtitle-loop-stop'

In subed-disable-loop-over-current-subtitle:
subed-common.el:817:44:Warning: assignment to free variable
    `subed--subtitle-loop-start'
subed-common.el:817:71:Warning: assignment to free variable
    `subed--subtitle-loop-stop'
subed-common.el:817:6:Warning: reference to free variable
    `subed--subtitle-loop-start'
subed-common.el:817:6:Warning: reference to free variable
    `subed--subtitle-loop-stop'

In subed--set-subtitle-loop:
subed-common.el:833:42:Warning: reference to free variable
    `subed-loop-seconds-before'
subed-common.el:837:47:Warning: assignment to free variable
    `subed--subtitle-loop-start'
subed-common.el:835:41:Warning: reference to free variable
    `subed-loop-seconds-after'
subed-common.el:838:47:Warning: assignment to free variable
    `subed--subtitle-loop-stop'
subed-common.el:840:43:Warning: reference to free variable
    `subed--subtitle-loop-start'
subed-common.el:841:43:Warning: reference to free variable
    `subed--subtitle-loop-stop'

In subed--ensure-subtitle-loop:
subed-common.el:845:14:Warning: reference to free variable
    `subed--subtitle-loop-start'
subed-common.el:845:41:Warning: reference to free variable
    `subed--subtitle-loop-stop'
subed-common.el:846:14:Warning: reference to free variable
    `subed-mpv-is-playing'

In subed-enable-pause-while-typing:
subed-common.el:875:25:Warning: reference to free variable
    `subed-playback-speed-while-typing'

In subed--pause-while-typing:
subed-common.el:905:13:Warning: reference to free variable
    `subed-mpv-is-playing'
subed-common.el:905:34:Warning: reference to free variable
    `subed--player-is-auto-paused'
subed-common.el:906:13:Warning: reference to free variable
    `subed-playback-speed-while-typing'
subed-common.el:923:36:Warning: assignment to free variable
    `subed--player-is-auto-paused'
subed-common.el:921:28:Warning: reference to free variable
    `subed-unpause-after-typing-delay'
subed-common.el:924:56:Warning: reference to free variable
    `subed-playback-speed-while-not-typing'

In subed-guess-video-file:
subed-common.el:944:32:Warning: reference to free variable
    `subed-video-extensions'

In end of data:
subed-common.el:954:1:Warning: the following functions are not known to be
    defined: subed-subtitle-msecs-start, subed-subtitle-msecs-stop,
    subed-backward-subtitle-id, subed-set-subtitle-time-start,
    subed--run-subtitle-time-adjusted-hook, subed-forward-subtitle-id,
    subed-set-subtitle-time-stop, subed-get-milliseconds-adjust,
    subed-jump-to-subtitle-id, subed-jump-to-subtitle-end, subed-mpv-jump,
    subed-subtitle-id, subed-prepend-subtitle, subed-append-subtitle,
    subed-backward-subtitle-text, subed-regenerate-ids-soon,
    subed-srt--msecs-to-timestamp, subed-jump-to-subtitle-text-at-msecs,
    subed--post-command-handler, subed-mpv-pause, subed-mpv-unpause,
    subed-mpv-playback-speed

subed-config.el:283:1:Warning: the following functions are not known to be
    defined: subed-forward-subtitle-text, subed-backward-subtitle-text,
    subed-jump-to-subtitle-text, subed-jump-to-subtitle-end,
    subed-decrease-start-time, subed-increase-start-time,
    subed-decrease-stop-time, subed-increase-stop-time,
    subed-move-subtitle-forward, subed-move-subtitle-backward,
    subed-shift-subtitle-forward, subed-shift-subtitle-backward,
    subed-insert-subtitle, subed-insert-subtitle-adjacent,
    subed-kill-subtitle, subed-sort, subed-mpv-toggle-pause,
    subed-toggle-debugging, subed-mpv-find-video,
    subed-toggle-pause-while-typing, subed-toggle-loop-over-current-subtitle,
    subed-toggle-replay-adjusted-subtitle,
    subed-copy-player-pos-to-start-time, subed-copy-player-pos-to-stop-time,
    subed-toggle-sync-point-to-player, subed-toggle-sync-player-to-point,
    subed-milliseconds-adjust, subed-subtitle-msecs-start,
    subed-disable-sync-point-to-player-temporarily, subed-subtitle-id
```

purcell:
>  I would recommend using `define-derived-mode`

Indeed. To see what code `define-derived-mode` generates for you, you can write a sample form which uses it, and expand that using `pp-macroexpand-last-sexp`. It automates a lot of the old boilerplate.

rndusr:
Thanks for taking another look, and also for your support.

##### Warnings

I hate them and I don't know how to fix them.

Putting thousands of lines in a single file is not an option for me.  All my
first projects were like that and I spent more time jumping from here to there
than being productive.

I'm also having trouble reproducing them consistently.  I get different warnings
when I run byte-compile-file in an interactive Emacs session compared to, e.g.,
emacs -Q -batch --eval "(byte-compile-file \"subed/subed-srt.el\")".

    subed-common.el:954:1:Warning: the following functions are not known to be defined:
        subed-subtitle-msecs-start, subed-subtitle-msecs-stop,
        subed-backward-subtitle-id, subed-set-subtitle-time-start,
        subed-forward-subtitle-id, subed-set-subtitle-time-stop,
        subed-jump-to-subtitle-id, subed-jump-to-subtitle-end, subed-subtitle-id,
        subed-prepend-subtitle, subed-append-subtitle,
        subed-backward-subtitle-text, subed-regenerate-ids-soon,
        subed-jump-to-subtitle-text-at-msecs

These functions are defined when subed-mode is called.  This is by design to
support multiple subtitle formats with the (seemingly) same function.

    subed-common.el:520:1:Warning: Unused lexical variable ‘buffer-is-empty’
    subed-common.el:578:1:Warning: Unused lexical variable ‘msecs-avail’

These variables are return values from another function,
subed--insert-subtitle-info, which returns a lot of values, so I bound them to
variables with cl-multiple-value-bind.  subed--insert-subtitle-info is used in
two places and in one of those places there is no need for these variables
from the warnings.

I've tried to add more (require ...)s and (declare-function ...)s and whatever
else I could think of in the last few days.  Nothing seems to work.  And to be
honest, I'm not very motivated because these warnings are completely harmless,
expected, and they don't seem to appear during normal usage.

##### Derived mode

I've turned subed-mode into a derived mode and that seems to work fine, except:

1. It can't be disabled anymore.  It seems minor modes are supposed to be
   enabled/disabled/toggled while major modes are not, which I think is odd.

2. Calling subed-mode multiple times opens another mpv instance each time.  I've
   told the function that executes mpv to kill the old instance if there is one,
   but that doesn't work when it's called by subed-mode because all the
   buffer-local variables, including the one that stores the process, are
   already killed, turning the old mpv instance into some kind of zombie.

So not a huge fan of that either, although I see the benefits of using
high-level functions for more consistency.

Sorry if I'm being complicated. :/

rndusr:
When I byte-compile with subed-mode enabled (e.g. by opening foo.srt), all files
compile without issue except for these two:

    subed-mpv.el:362:1:Warning: the function ‘subed-mpv--sub-add-after-save’ is
        not known to be defined.

This is because, if the subtitle file doesn't exist yet, I can't tell mpv to
load it, so I add a tell-mpv-to-load-subtitle-file function to
after-save-hook. And to keep things clean, tell-mpv-to-load-subtitle-file
removes itself from after-save-hook and also undefines itself. Is there a better
way to do this?

    subed-common.el:520:1:Warning: Unused lexical variable ‘buffer-is-empty’
    subed-common.el:578:1:Warning: Unused lexical variable ‘msecs-avail’

I've explained this in my previous reply: These variables are return values from
another function, subed--insert-subtitle-info, which returns a lot of values, so
I bound them to variables with cl-multiple-value-bind.
subed--insert-subtitle-info is used in two places and in one of those places
there is no need for these variables from the warnings.

riscy:
> When I byte-compile with subed-mode enabled

It took me a couple reads to understand this.  I was not understanding the workflow in subed--init.  That's my oversight, and I've had some time to put my eyes on the source and understand what's going on.

I think it's fair to say that in virtually all cases someone will be byte-compiling your package as a means to install it, _not_ after it's been enabled, so I'm afraid this does not meet the criteria for byte-compiling cleanly.

Really, it should be possible to compile each file on its own -- warning free -- with a fresh Emacs instance.  Using this workflow, you miss out on things like the fact that subed-common.el should be including subed-config.el because the latter defines subed-debugging-enabled-p.  Or that subed-mpv.el should be including subed-common.el because the latter defines subed-debug.  I'd recommend enabling flycheck, and it should shed some light on some of these.

I think a lot of this has been done with the goal of increasing subed's flexibility and extensibility, which is definitely a worthy goal.  But I think you should strive for simplicity, laying the code out explicitly in a way that the compiler can understand in a single pass, and not worrying too much about a little bit of repeated code.  Some of the least inhabitable code I've been in has ended up that way because of excessive abstraction.

rndusr:
I've somehow fixed most of the warnings without making the code much worse. Yay!

I've added a "test-compile" target in my Makefile that should put everyone on
the same page regarding compiler messages.  Am I doing that correctly?

----

The only thing I can't figure out is how to declare functions programmatically.

This code, inserted in subed.el above "(defun subed--init ...", seems to work fine:

    (cl-loop for func-suffix in subed--generic-function-suffixes do
             (message "%s is a function: %S" func-suffix (fboundp (subed--get-generic-func func-suffix)))
             (defalias (subed--get-generic-func func-suffix)
               (lambda () "Stub." (message "This is just a stub function.")))
             (message "%s is a function: %S" func-suffix (fboundp (subed--get-generic-func func-suffix))))
    (message "%s is a function: %S" 'subed-validate (fboundp 'subed-validate))

It makes functions and Emacs knows about them.  But the compiler still
complains.  I'm almost starting to get cocky and assume that the "undefined
function" detector simply isn't smart enough for my horrible hacks.

riscy:
I've recently updated package-lint, and you require a slightly more recent version of Emacs:
```
;; Package-Requires: ((emacs "25.1"))
```

Per your question, it actually doesn't matter what's in your Makefile -- your Makefile won't be used when this package is being compiled and installed on somebody else's system.

I recommend installing and configuring flycheck for Elisp so that you can see some of these.

Alternatively, open up a fresh Emacs instance that hasn't been configured with your own configurations with `emacs -Q`.  Go to the file subed.el.  Ensure its directory is on your load-path with add-to-list.  Then run <kbd>M-x byte-compile-file</kbd>.  You will see:

```
In end of data:
subed.el:199:1:Warning: the following functions are not known to be defined:
    subed-forward-subtitle-text, subed-backward-subtitle-text,
    subed-jump-to-subtitle-text, subed-jump-to-subtitle-end,
    subed-kill-subtitle, subed-sort
```
This is what's meant by byte compiling cleanly.

rndusr:
> ;; Package-Requires: ((emacs "25.1"))

OK, thanks.

When I open subed.el and run M-x flycheck-mode and then M-x flycheck-compile, it
complains about not finding subed-config, even though it's in my load-path.

I know the Makefile is not used during the installation or build process, but
I'm not going to manually open each file in a separate, clean Emacs session and
M-x byte-compile it.  Why would that be better than doing it in batch mode with
"make test-compile"?

Anyway, as I explained in my previous reply, I think the only issue left is that
byte-compile thinks the auto-generated generic functions are not defined when
they acutally are, and I don't know how to convince it otherwise.

riscy:
Sorry about that, I didn't look closely at your Makefile.  I just looked at it, and yes, it looks correct.

I think we're all on the same page with respect to this output:
```
$ make test-compile
emacs --quick --batch --eval "(progn (add-to-list 'load-path (expand-file-name \"subed\" default-directory)) \
	                                     (byte-compile-file \"subed/subed.el\"))"

In end of data:
subed.el:199:1:Warning: the following functions are not known to be defined:
    subed-forward-subtitle-text, subed-backward-subtitle-text,
    subed-jump-to-subtitle-text, subed-jump-to-subtitle-end,
    subed-kill-subtitle, subed-sort
emacs --quick --batch --eval "(progn (add-to-list 'load-path (expand-file-name \"subed\" default-directory)) \
	                                     (byte-compile-file \"subed/subed-config.el\"))"

In end of data:
subed-config.el:228:1:Warning: the function ‘subed-subtitle-msecs-start’ is
    not known to be defined.
emacs --quick --batch --eval "(progn (add-to-list 'load-path (expand-file-name \"subed\" default-directory)) \
	                                     (byte-compile-file \"subed/subed-common.el\"))"

In end of data:
subed-common.el:950:1:Warning: the following functions are not known to be
    defined: subed-subtitle-id, subed-subtitle-msecs-start,
    subed-subtitle-msecs-stop, subed-backward-subtitle-id,
    subed-set-subtitle-time-start, subed-forward-subtitle-id,
    subed-set-subtitle-time-stop, subed-jump-to-subtitle-id,
    subed-jump-to-subtitle-end, subed-prepend-subtitle, subed-append-subtitle,
    subed-backward-subtitle-text, subed-regenerate-ids-soon,
    subed-jump-to-subtitle-text-at-msecs
emacs --quick --batch --eval "(progn (add-to-list 'load-path (expand-file-name \"subed\" default-directory)) \
	                                     (byte-compile-file \"subed/subed-mpv.el\"))"

In end of data:
subed-mpv.el:378:1:Warning: the following functions are not known to be
    defined: subed-subtitle-msecs-start, subed-subtitle-id
emacs --quick --batch --eval "(progn (add-to-list 'load-path (expand-file-name \"subed\" default-directory)) \
	                                     (byte-compile-file \"subed/subed-srt.el\"))"

In end of data:
subed-srt.el:578:1:Warning: the following functions are not known to be
    defined: subed-subtitle-id, subed-subtitle-relative-point,
    subed-jump-to-subtitle-id
emacs --quick --batch --eval "(progn (add-to-list 'load-path (expand-file-name \"subed\" default-directory)) \
	                                     (byte-compile-file \"subed/subed-debug.el\"))"
```

riscy:
>  I don't know how to convince it otherwise.

You might try reaching out to the community.  I don't have as much bandwidth these days, sorry. :(

rndusr:
> You might try reaching out to the community.  I don't have as much bandwidth these days, sorry. :(

Fair enough. You've been more than helpful already.

purcell:
I took a look at this just now, FWIW. There's a lot to unpick, and the handling of the generic functions etc. makes it extremely hard to follow, so I'm afraid I can't be of much help. But since you're thinking of this as a generic interface with a number of different implementations, this seems like a prime opportunity to use `cl-defgeneric` and co.

rndusr:
It's been a while, but it looks like the functions that do stuff (e.g. increase
start time in .srt subtitles) are just regular functions and they live in one
file per format.

When subed-mode is activated, subed--init is called, which translates the file
name extension (e.g. "srt") into a format-specific init function
(e.g. subed-srt--init) and calls it. The format-specific init function must set
the global variable subed--subtitle-format.

After that, subed--init walks through a list of hardcoded function names and
creates a generic function for each one, e.g. the string "subtitle-id" creates
the generic function subed-subtitle-id. The generic function uses
subed--subtitle-format to translate "subtitle-id" to the format-specific
function name "subed-srt--subtitle-id".


cl-defgeneric looks interesting. I must have come across this while I was
banging my head against that wall, but I can't remember why I didn't use it.

My initial thought was that this doesn't work because defgenerics decide which
defmethod to call based on the arguments, but in subed arguments do not differ
between formats.

Maybe I can just pass the subtitle format as the first argument and somehow use
that?

purcell:
> My initial thought was that this doesn't work because defgenerics decide which defmethod to call based on the arguments, but in subed arguments do not differ between formats. Maybe I can just pass the subtitle format as the first argument and somehow use that?

Yep, you'd probably just use a symbol to represent the format, e.g. `'srt`.

tarsius:
Friendly ping!

rndusr:
I probably won't put any work into this in the forseeable future.  One of my
other projects needs a lot of attention and I've been putting it off for too
long.

I've also come to realize that I may not be a good maintainer for subed because
I only occasionally edit subtitles.  Users were asking about the workflow for
creating subtitles from scratch and I couldn't really give them an answer
because of my lack of experience.

If anyone wants to collaborate or even fork, let me know.

Sorry for backing out. :(

tarsius:
Okay. I am closing this for the time being. If you or someone else finds the time to finish this, then please let us know and we will reopen this pr.

PR #6256, opened 2019-07-01
Include obsolete ESS lisp files

jabranham:
This fixes ESS's recipe to include obsolete files, see https://github.com/emacs-ess/ESS/issues/894


purcell:
Thanjks

PR #6258, opened 2019-07-02
Add helm-chrome-history recipe

xuchunyang:

Browse Chrome history using helm interface.

https://github.com/xuchunyang/helm-chrome-history

I'm the maintainer

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/helm-chrome-history
+++ b/recipes/helm-chrome-history
@@ -1,0 +1,1 @@
+(helm-chrome-history :fetcher github :repo "xuchunyang/helm-chrome-history")


alphapapa:
This looks cool.  Have you considered using a `WHERE` clause in the SQLite query, rather than reading the entire Chrome history into memory in Emacs?  That would likely be faster, since the database is probably indexed on something like URL and page title, and it would avoid the cache becoming stale.

xuchunyang:
Good idea. I will store the sql to a variable, so user can adjust it

```elisp
(defvar helm-chrome-history-sql "select url, title, last_visit_time from urls")
```

It's not an issue for me at least for now, since I don't have too many chrome history, 41525 history records takes 7.4 M and I'm fine with it. One can also use `M-x helm-chrome-history-clear-cache` to reset `helm-chrome-history-candidates` to nil (I assume then Emacs will later free the 7.4M memory) 

```elisp
(length helm-chrome-history-candidates)
;; => 41525

(with-temp-buffer
  (prin1 helm-chrome-history-candidates (current-buffer))
  (file-size-human-readable (buffer-size)))
;; => "7.4M"
```

xuchunyang:
And I have also limit the history to 100000 (that's 20 Mb most, I guess) by default, just in case it consumes too much memory for first time user.

alphapapa:
@xuchunyang What I mean is, you could use a `WHERE` clause to let SQLite do the matching of keywords the user inputs, like:

```sql
SELECT url, title, last_visit_time 
FROM urls
WHERE title LIKE 'emacs'
```
Or something like that.

xuchunyang:
Oh, I have not considered using the sql for filtering. The current implement is obvious and straightforward and easy to me, and It does work.

xuchunyang:
> * I scroll down to one and hit enter -- nothing happens.  _Should something happen_?

Oh, the default SQL is incorrect, I am using the outdated elc so I didn't notice the problem. It has been fixed, the actions work now.

> * Brwose => Browse
> * Url => URL (typically)

Fixed.

> * `helm-chrome-history-file` could be a `defcustom` in case someone wants to use the customization interface -- thoughts?

Yes, `helm-chrome-history-sql` and `helm-chrome-history-action` should also be defined with `defcustom`. I use `defvar` because it is easier and simpler. Maybe I will change my mind.


> * I don't believe [this](https://github.com/xuchunyang/helm-chrome-history/blob/ee613783fac781b90ed97f17358506ecf3f410bc/helm-chrome-history.el#L65) nil needs to be quoted

It must be quoted for matching `nil` otherwise `pcase` will signal an error. Other literal values such as numbers and strings doesn't need to be quoted, but `nil` and `t` does.

> * Rather than defining `lambdas` [here](https://github.com/xuchunyang/helm-chrome-history/blob/ee613783fac781b90ed97f17358506ecf3f410bc/helm-chrome-history.el#L111-L120), why not use named functions?  That would make it easier to hack on the library to debug it.  Currently I had to redefine `helm-chrome-history-action` and then redefine `helm-chrome-history-source` before any changes in those lambdas got picked up.

I agree. `lambda` is just easier. I will use just

```elisp
(("Browse URL" . browse-url) ("Copy URL" . kill-new))
```

riscy:
Great thanks.  I notice that you took some code from [helm-chrome](https://github.com/kawabata/helm-chrome/blob/master/helm-chrome.el) (you attribute `helm-chrome-file` but I think it would be clearer to write `helm-chrome.el` -- maybe even a link to their repo?)

riscy:
I'll leave that one for your consideration -- this one looks good to go.

xuchunyang:
> I notice that you took some code from [helm-chrome](https://github.com/kawabata/helm-chrome/blob/master/helm-chrome.el) (you attribute `helm-chrome-file` but I think it would be clearer to write `helm-chrome.el` -- maybe even a link to their repo?)

Done.

---

Thank @riscy and @alphapapa for the reviewing.

PR #6259, opened 2019-07-02
js/melpa.js: allow https://git.* fallback for other forges

stsquad:
The calculateSourceURL function has special handling for a selection
of forges and other hosting sites. Rather than adding new special
handling for forges like sr.ht lets just assume that a modern https://
website starting with git with do the right thing when a browser
visits.

### DIFF
--- a/html/js/melpa.js
+++ b/html/js/melpa.js
@@ -136,7 +136,8 @@
                 urlMatch(/(gitlab\.com\/[^\/]+\/[^.]+)/, "https://") ||
                 urlMatch(/^lp:(.*)/, "https://launchpad.net/") ||
                 urlMatch(/^(https?:\/\/code\.google\.com\/p\/[^\/]+\/)/) ||
-                urlMatch(/^(https?:\/\/[^.]+\.googlecode\.com\/)/));
+                urlMatch(/^(https?:\/\/[^.]+\.googlecode\.com\/)/) ||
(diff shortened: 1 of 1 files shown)


html/js/melpa.js
```diff
@@ -136,7 +136,8 @@
                 urlMatch(/(gitlab\.com\/[^\/]+\/[^.]+)/, "https://") ||
                 urlMatch(/^lp:(.*)/, "https://launchpad.net/") ||
                 urlMatch(/^(https?:\/\/code\.google\.com\/p\/[^\/]+\/)/) ||
-                urlMatch(/^(https?:\/\/[^.]+\.googlecode\.com\/)/));
+                urlMatch(/^(https?:\/\/[^.]+\.googlecode\.com\/)/) ||
+                urlMatch(/^(https?:\/\/git.*)/));
```
mattiasb: Just a drive-by comment:
`/^(https?:\/\/git.*)/` should probably be `/^(https?:\/\/git\.)/`

purcell:
Looks good, but I agree that the fix noted by @mattiasb appears necessary here, otherwise the regex itself will incorrectly match `gitlab`, `github` etc. (though of course there are separate earlier matchers for those).

purcell:
P.S. did you test-drive this locally?

purcell:
Modified version committed, thanks.

PR #6260, opened 2019-07-02
Show compile-log after building at sandbox

conao3:
When submitting a recipe, you are required to test the package as
`make sandbox INSTALL = <foo>`.

By showing the compile log, developers can easily check for warnings.


purcell:
Makes sense, thanks!

conao3:
Thanks!

Issue #6262, opened 2019-07-02
What is general strategy for multi files with multi dependency?


conao3:
CC: @akirak

Hi!
First, thank you for longtime maintenance for MELPA.

I found broken elisp package which is consist of multi files with each dependency.
This is https://github.com/akirak/org-starter especially `org-starter-swiper.el`.

I get below error when execute `make sandbox INSTALL=org-starter` at the root of the fork melpa/melpa.
```
^L
Compiling no file at Wed Jul  3 08:16:13 2019
Leaving directory ‘/Users/conao/dev/repos/melpa/sandbox/elpa/org-starter-20190701.427’
^L
Compiling file /Users/conao/dev/repos/melpa/sandbox/elpa/org-starter-20190701.427/org-starter-extras.el at Wed Jul  3 08:16:13 2019
Entering directory ‘/Users/conao/dev/repos/melpa/sandbox/elpa/org-starter-20190701.427/’
^L
Compiling file /Users/conao/dev/repos/melpa/sandbox/elpa/org-starter-20190701.427/org-starter-swiper.el at Wed Jul  3 08:16:13 2019
org-starter-swiper.el:35:1:Error: Cannot open load file: No such file or directory, swiper
^L
Compiling file /Users/conao/dev/repos/melpa/sandbox/elpa/org-starter-20190701.427/org-starter.el at Wed Jul  3 08:16:13 2019
```

`org-starter-swiper.el`'s header is below
```
;;; org-starter-swiper.el --- Swiper for org-starter -*- lexical-binding: t -*-

;; Copyright (C) 2019 Akira Komamura

;; Author: Akira Komamura <akira.komamura@gmail.com>
;; Version: 0.1
;; Package-Requires: ((emacs "25.1") (swiper "0.11"))
;; URL: https://github.com/akirak/org-starter
```
but MELPA parses dependency as below cited from https://melpa.org/#/org-starter
```
Downloads  269 (all versions), percentile: 15.06
Source  github (commit 60ccfd)
Dependencies  dash 2.12 / dash-functional 1.2.0 / emacs 25.1
Needed by
```

So, he and we should describe all dependency in main elisp file only?
Or MELPA plans to crawl all elisp files and gather dependencies in the future?
It seems that it related to #1307

purcell:
[...]

PR #6263, opened 2019-07-03
Add describe dependency strategy for multifile package

conao3:
The information provided in #6262 is useful and should be shared.

I considered where this information should be posted, but it
seems that there is not a good place.

Suggestions are welcome, as I believe that they need to be posted
more clearly in more visible places!


purcell:
Excellent, thank you very much.

conao3:
Good, thanks!
Feel free to rewrite you want better.

PR #6264, opened 2019-07-03
Fix elpa package url scheme; use https instead of http

conao3:
Currently, fetch packages which located in elpa from
http://elpa.gnu.org/packages/ but original `package-archives`
definition is below.

```
(defcustom package-archives `(("gnu" .
                               ,(format "http%s://elpa.gnu.org/packages/"
                                        (if (gnutls-available-p) "s" ""))))
  "An alist of archives from which to fetch.
The default value points to the GNU Emacs package repository.

Each element has the form (ID . LOCATION).
 ID is an archive name, as a string.
 LOCATION specifies the base location for the archive.
  If it starts with \"http(s):\", it is treated as an HTTP(S) URL;
  otherwise it should be an absolute directory name.
  (Other types of URL are currently not supported.)

Only add locations that you trust, since fetching and installing
a package can run arbitrary code."
  :type '(alist :key-type (string :tag "Archive name")
                :value-type (string :tag "URL or directory name"))
  :risky t
  :version "26.1")
```

When available gnutls, fetch elpa packages via https is recomended.
So I fix url scheme. But is there any other reasons using `http`
instead of `https`?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -122,7 +122,7 @@ sandbox: packages/archive-contents
 	@$(EMACS_COMMAND) -Q \
 		--eval '(setq user-emacs-directory (file-truename "$(SANDBOX)"))' \
 		-l package \
-		--eval "(add-to-list 'package-archives '(\"gnu\" . \"http://elpa.gnu.org/packages/\") t)" \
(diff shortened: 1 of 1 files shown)


purcell:
Since we're already using `https` here for MELPA, then `https` should also be fine for ELPA. Thanks!

conao3:
Good, thanks!

PR #6269, opened 2019-07-04
Add sysctl recipe

dantecatalfamo:

View and edit sysctl in an org-mode structured buffer

https://github.com/dantecatalfamo/sysctl.el

I'm the package author

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/sysctl
+++ b/recipes/sysctl
@@ -1,0 +1,2 @@
+(sysctl :repo "dantecatalfamo/sysctl.el"
+        :fetcher github)


riscy:
[/elisp/sysctl.el]
-------------------------------------------------------------------------------

<kbd>M-x checkdoc</kbd>:
No issues! 💯

<kbd>M-x package-lint-current-buffer</kbd> (using version 20190621.14):
No issues! 🎊

```
sysctl.el:167:1:Warning: the function `sysctl--run-command' is not known to be
    defined.
```

No issues! 🎉

-------

Ping here when you're ready!

dantecatalfamo:
Fixed the missing function and improved the description 👍

riscy:
> You should use featurep to check if a feature exists here (i.e. (featurep 'flyspell))

After some thought I may have been wrong about this one.  But maybe it's a good idea to specifically check if the function you want to call is bound, rather than just `flyspell-mode`?

dantecatalfamo:
Thank you very much for the feedback! I've implemented all your recommended fixes and cleaned up several functions. Should be good to go!

riscy:
Yikes, forgive me for this one -- the `explicit-shell-file-name` was a bad call.  I checked out the definition of `shell-command-to-string` and it prefers `shell-file-name`:
```elisp
(defun shell-command-to-string (command)
  "Execute shell command COMMAND and return its output as a string."
  (with-output-to-string
    (with-current-buffer
      standard-output
      (process-file shell-file-name nil t nil shell-command-switch command))))
```
I guess the `explicit-shell-file-name` is a convention from `term.el` and `shell.el` that doesn't carry over here.

riscy:
Thanks, this looks cool!  And, welcome to MELPA. :)

dantecatalfamo:
Thanks, happy to be here! 😀

PR #6271, opened 2019-07-05
Add shadowenv recipe

dantecatalfamo:

[Shadowenv](https://github.com/Shopify/shadowenv) integration.

https://github.com/Shopify/shadowenv.el/

I'm the package author

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/shadowenv
+++ b/recipes/shadowenv
@@ -1,0 +1,2 @@
+(shadowenv :repo "Shopify/shadowenv.el"
+           :fetcher github)


dantecatalfamo:
Fixed, thanks!

dantecatalfamo:
Problems fixed, thanks for the check 👌

riscy:
Taking a deeper look at this now - 

- [Here](https://github.com/Shopify/shadowenv.el/blob/master/shadowenv.el#L64-L66) I believe you can use `defvar-local` -- for this and the others.
- [Here](https://github.com/Shopify/shadowenv.el/blob/master/shadowenv.el#L78), `unless if` could be confusing -- this should be refactored into a single `if`
- [This](https://github.com/Shopify/shadowenv.el/blob/master/shadowenv.el#L194) should use `not` or `null`.
- For [these](https://github.com/Shopify/shadowenv.el/blob/master/shadowenv.el#L157-L161) forms, since you rely on Emacs 24, I believe you can use `setq-local`
- [This](https://github.com/Shopify/shadowenv.el/blob/master/shadowenv.el#L112-L120) appears to be a good case for `pcase`
- Should [this hook](https://github.com/Shopify/shadowenv.el/blob/master/shadowenv.el#L163) be removed in `shadowenv-down`?

riscy:
> Here, unless if could be confusing -- this should be refactored into a single if

Well, maybe not a *single* if :)  What about something like:
```elisp
(defun shadowenv-run (data)
  "Run shadowenv porcelain with DATA."
  (if shadowenv-binary-location
     (file-executable-p shadowenv-binary-location)
    (or (executable-find "shadowenv") (error "Cannot find..."))))
```

dantecatalfamo:
I'm not sure how to use `pcase` in this scenario and get it to evaluate and compare the strings instead of setting `shadowenv--set-exported` to the value of `opcode`

riscy:
I think this works:
```elisp
(pcase opcode
  (shadowenv--set-exported   (setenv variable value))
  (shadowenv--unset          (setenv variable))
  (shadowenv--set-unexported
   (if (string= variable "__shadowenv_data")
       (setq shadowenv-data value)
     (warn "Unrecognized operand for SET_UNEXPORTED operand: %s" variable))))

**Comment by dantecatalfamo** [comment]
I tried that earlier when you suggested it but when I expand the macro it only creates this so I'm not sure if it's working as intended:
```elisp
(let
    ((shadowenv--set-exported opcode))
  (setenv variable value))
```

riscy:
You're totally correct -- I had no experience using pcase with non-literals before and figured it would be similar, but it's not.  Thanks for hearing me out.  I'll take another look later today. :)

riscy:
Checkdoc's telling me you'll have to rely on `((emacs "24.3"))` for `setq-local`, and then I think this is all set.  I should have been more specific.  Thanks. :)

dantecatalfamo:
Dependancy updated, thanks a lot for all your help with this 😄

riscy:
Great!

PR #6272, opened 2019-07-06
Add dash-alfred recipe

xuchunyang:

Search Dash docs on-the-fly

https://github.com/xuchunyang/dash-alfred.el

Maintainer

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/dash-alfred
+++ b/recipes/dash-alfred
@@ -1,0 +1,1 @@
+(dash-alfred :fetcher github :repo "xuchunyang/dash-alfred.el")


kidd:
is there any major difference with helm-dash?

xuchunyang:
helm-dash is an independent application. dash-alfred is a search interface to Dash.app, dash-alfred works with Dash.app. The official (suggested by Dash.app) Emacs plugin is [dash-at-point](https://github.com/stanaka/dash-at-point), which can't search on-the-fly.

riscy:
Just because `helm-dash` is an independent application doesn't mean we're not inadvertently duplicating functionality.  Similarly, there is [counsel-dash](https://github.com/dash-docs-el/counsel-dash), which looks to cover some of the feature you've developed here.  Can you comment a bit more on the novelty of your package so we know what to look for?

In the meantime, here's a pass over the code:

- [This](https://github.com/xuchunyang/dash-alfred.el/blob/e7f0f8dc98b1ecc7ca36652daa7b0b34ede1b8d2/dash-alfred.el#L35-L38) looks like something that would be worth making customizable or documenting in your commentary or README.
- [This](https://github.com/xuchunyang/dash-alfred.el/blob/e7f0f8dc98b1ecc7ca36652daa7b0b34ede1b8d2/dash-alfred.el#L56-L63) and [this](https://github.com/xuchunyang/dash-alfred.el/blob/e7f0f8dc98b1ecc7ca36652daa7b0b34ede1b8d2/dash-alfred.el#L107-L114) are so similar that I would strongly recommend factoring them out into a single function -- it'll make bugfixes easier in the future and shorten your code.
- The error message for ivy is `dashAlfredWorkflow fails` but the error message for helm is `dashAlfredWorkflow failed` -- probably both should be "failed"

riscy:
Looking again, I'm seeing that this is an interface to Alfred rather than just dash, which makes sense.  Sorry for my misunderstanding.

xuchunyang:
> Just because `helm-dash` is an independent application doesn't mean we're not inadvertently duplicating functionality. Similarly, there is [counsel-dash](https://github.com/dash-docs-el/counsel-dash), which looks to cover some of the feature you've developed here

That's right. `helm-dash` and `counsel-dash` are third-part client for browsing Dash docsets and probably have the same (and more) functionality. But when I try `helm-dash` last year it didn't work for me (I was unable to download the docsets). Dash.app is the official client for Dash docsets. My package, dash-alfred, isn't another third-part client of Dash docsets, it's just an interface for the Dash.app, thus its implement is mush simpler than `helm-dash` and `counsel-dash`.



> In the meantime, here's a pass over the code:

I have addressed these three points.

xuchunyang:
I also updated the gif on README.md. I hope it makes the purpose of the package clearer.

riscy:
Thanks for that.  This looks good to me!

PR #6273, opened 2019-07-06
Add recipe for egalgo.

ROCKTAKEY:

Genetic Algorithm for Emacs.

https://github.com/ROCKTAKEY/egalgo

I'm maintainer.

 **None needed**

### Checklist

Please confirm with `x`:

`package-lint` warns:
```
```
However, this package name "egalgo" is abbreviation of "Genetic ALGOrithm for Emacs", so
the word "Emacs" is needed.

### DIFF
--- a/recipes/egalgo
+++ b/recipes/egalgo
@@ -1,0 +1,1 @@
+(egalgo :repo "ROCKTAKEY/egalgo" :fetcher github)


ROCKTAKEY:
>Could egalgo--t-p simply be replaced with identity?

No. `egalgo--p` should return `nil` even when the argument is `non-nil` except `t`.

>For brevity you can write: `(or (car rates) 0)` instead of `(if (car rates) (car rates) 0)` here; similarly here.

Oh, you're right. Thanks. I'll fix it.

>pakcage => package
funciton => function
selectror => selector
Genarate => generate
chromsomes => chromosomes
abusolutely => absolutely

Oh, many typo... I'll fix them, too...

riscy:
Right on, this looks good.  I hope you get over your cold. :)

PR #6275, opened 2019-07-07
Update cmake-mode to use its upstream git repository, not a mirror.

xuhdev:

An enthusiastic user


riscy:
Thanks for noticing this!

PR #6276, opened 2019-07-07
Add recipe for line-indicators.

jcs090218:

Line annotation similar to Visual Studio.

This package is basically the same to [line-reminder](https://github.com/elpa-host/line-reminder). But now this doesn't required `linum` and display it using [indicators](https://github.com/Fuco1/indicators.el).

https://github.com/jcs090218/line-indicators

The maintainer

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/line-indicators
+++ b/recipes/line-indicators
@@ -1,0 +1,1 @@
+(line-indicators :repo "jcs090218/line-indicators" :fetcher github)


riscy:
Interesting -- I wonder if it would be feasible to incorporate this into line-reminder?  This wouldn't be unprecedented -- packages like e.g. yascroll let the user toggle the display method.  It would be a little more work to do, but the two packages share quite a bit of configuration, and line-reminder already has a userbase.

A couple quick suggestions otherwise -- the checklist looks great, but in a few places you could just be writing `nil` instead of `'()`.  Also I feel like "line indicators" doesn't tell me a lot about what the package does (after all, there's a package called "indicators" that indicates lines as well).

jcs090218:
> Interesting -- I wonder if it would be feasible to incorporate this into line-reminder? This wouldn't be unprecedented -- packages like e.g. yascroll let the user toggle the display method. It would be a little more work to do, but the two packages share quite a bit of configuration, and line-reminder already has a userbase.

Yeah, I am thinking about that too. Only thing that I need to do is to avoid `(require 'linum)`. Thanks for the suggestion!

> A couple quick suggestions otherwise -- the checklist looks great, but in a few places you could just be writing nil instead of '(). Also I feel like "line indicators" doesn't tell me a lot about what the package does (after all, there's a package called "indicators" that indicates lines as well).

Yeah, I do have problem with naming! Any suggestion will be appreciated!

jcs090218:
I just incorporated to [line-reminder](https://github.com/elpa-host/line-reminder), thank you!

PR #6283, opened 2019-07-10
adding new mark-thing-at recipient

plandes:

This package tracks and creates bindings for region marking functions found in `thingatpt.el`

https://github.com/plandes/mark-thing-at

Author

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/mark-thing-at
+++ b/recipes/mark-thing-at
@@ -1,0 +1,1 @@
+(mark-thing-at :fetcher github :repo "plandes/mark-thing-at")


riscy:

<kbd>M-x checkdoc</kbd>:
No issues! 💯

No issues! 🎊

No byte-compilation issues! 😎

No issues! 🎉

-------
- [#L157](https://github.com/plandes/mark-thing-at/blob/master/mark-thing-at.el#L157): It's beneficial to sharp-quote this command
- [#L137](https://github.com/plandes/mark-thing-at/blob/master/mark-thing-at.el#L137): Use `unless ...` instead of `when (not ...)`
- [#L74](https://github.com/plandes/mark-thing-at/blob/master/mark-thing-at.el#L74): No need to sharp-quote lambdas (but also no harm)

(Thanks for catching everything on the checklist!  Just letting you know I see this.)

riscy:
Having a deeper look now --

In addition to the items listed at the end of my last comment, I think it is fine to support a function like [this](https://github.com/plandes/mark-thing-at/blob/de2aa6f3133351b3175cd84be7ffe8ee4d70529d/mark-thing-at.el#L143), but this looks like the kind of package that should contain a globalized minor mode that can be turned on and off (on a per-buffer basis or globally).  This would let users disable the mode if they've inadvertently caused keybindings to clash, or diagnose keybinding problems with `describe-mode`, or only enable your library in specific modes, etc.

purcell:
More notes from me:
- I wonder if `mark-at-point` might be an even nicer package name.
- Instead of a [function which uses `eval`](https://github.com/plandes/mark-thing-at/blob/de2aa6f3133351b3175cd84be7ffe8ee4d70529d/mark-thing-at.el#L81), you almost certainly want a macro. The `alist` should be a private var, since a user can't practically customise it. The machinery of the library looks a bit complicated, and if that's for extensibility, there's no indication of how a user should do that.
- I agree with @riscy re. providing a minor mode and globalised minor mode pair. [This function](https://github.com/plandes/mark-thing-at/blob/de2aa6f3133351b3175cd84be7ffe8ee4d70529d/mark-thing-at.el#L101) feels a bit random, when the conventional way to do this would be to add a `\\` docstring macro to the mode docstring so that `describe-mode` will include the keybindings for this minor mode.
- Prefer `unless` and `when` to `if` forms with a single branch. It helps readability.

Hope that helps!

plandes:
All great suggestions with most implemented (thanks).  @purcell: regarding changing the name: if you feel strongly about that I'm more than happy to do it.  If so, what's the best/easiest way to do that?  Should I simply change the code, repo name and melpa entry as a modification or should we just torch this pull request and start new?

Regarding the macro/eval.  I don't really see how to do that.  These are functions that are dynamically created, and have to be since each is key bound for each *thing* (i.e. one for `symbol`, another for `file`) etc.  Thus the eval is needed to create the function, doc, and message created for user interaction.  Again, I'm glad to make it a macro if you tell me how.

riscy:
Just to chime in, I think it's better to keep this PR open so things are easier to track.  Changing the code, repo name, and recipe file can happen alongside.

Regarding creating a macro, you can check out [these docs](https://www.gnu.org/software/emacs/manual/html_node/elisp/Simple-Macro.html#Simple-Macro), but here's an example closer to your use-case:
```elisp
(defmacro create-a-function (function-name msg)
  (list 'defun function-name () `(message ,msg)))

(create-a-function hello-world "hi!")

(hello-world)  ; => "hi!"
```

riscy:
Actually, I'm seeing now how the macro would be hard to define.  Scratch my suggestion.

plandes:
@riscy I agree, that in general, a macro is the way to go with this kind of thing.  However, the difference is the use case.  Typically with macros it's a particular call with particular arguments.  In this case, several functions are created programmatically so the only difference would be semantics as the underlying mechanism wouldn't change.

purcell:
Sorry for the radio silence -- I've been on vacation and starting a new job.

>    "Minor mode for on-the-fly syntax checking.

I think this docstring needs fixing. ;-)  Generally you'd just let `define-minor-mode` generate the docstring for you.

Regarding renaming, it's completely up to you, but `mark-at-point` probably better shows the correspondence with the functions in `thingatpt` that people will be familiar with.

Otherwise everything looks good, so let me know what you prefer re. the name and one of us can click "merge" here.

plandes:
@purcell What? wait--you don't get paid for this ;)

Busted!  Yes, it's my first minor mode despite my 20+ years of Emacs Lisp hacking.  That said, things in the last 5 - 10 years have firmed up the language so at times I feel I'm trying to catch up with (somewhat) new conventions.

In favor of laziness and you're excuse not to change the name, I will thusly not change the name :)

After changing the doc string, I tested and get this complaint from Cask:

```shell
mark-thing-at.el:10:37: error: Package choice-program is not installable.
```

However, I have defined the following in `Cask`:
```lisp
(depends-on "choice-program" "0.9")
```

This dependency shows up in Melpa and `melpa` is defined as a source.  Any ideas?

riscy:
I'm seeing a similar error.  Maybe `choice-program` is the culprit - I'm getting a packaging error on `choice-program`:
```
Version dependency for cl-lib appears too high: try 0.6.1
Depend on the latest 0.x version of cl-lib rather than on version "(1 0)".
Alternatively, depend on (emacs "24.3") or greater, in which cl-lib is bundled.
```

plandes:
@riscy Good idea.  However, this didn't work.  I removed the cl-lib dependency in `choice-program` and pointed Cask to the repo after a GitHub push with:
```lisp
(depends-on "choice-program"
	    :git "https://github.com/plandes/choice-program.git"
            :branch "master")
```
and still got the error in `mark-thing-at`.

riscy:
My mistake was I hadn't been doing this in my sandbox - 
```elisp
(add-to-list 'package-archives '("melpa" . "http://melpa.org/packages/"))
(package-refresh)
```
...in which case everything lints fine.  I think this is an issue with Cask, not your package, so this looks good to go.

PR #6286, opened 2019-07-11
Add recipe for lisp-butt-mode

marcowahl:

This is a minor mode to make fat lisp butts appear slim.

https://gitlab.com/marcowahl/lisp-butt-mode

 maintainer, contributor, An enthusiastic user

**None needed** 

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/lisp-butt-mode
+++ b/recipes/lisp-butt-mode
@@ -1,0 +1,1 @@
+(lisp-butt-mode :repo "marcowahl/lisp-butt-mode" :fetcher gitlab)


riscy:

<kbd>M-x checkdoc</kbd>:
```
Warning (emacs): 
lisp-butt-mode.el:88: Name lisp should appear capitalized as Lisp
Warning (emacs): 
lisp-butt-mode.el:99: Name lisp should appear capitalized as Lisp
Warning (emacs): 
lisp-butt-mode.el:111: Name lisp should appear capitalized as Lisp
```

No issues! 🎊

No byte-compilation issues! 😎

No issues! 🎉

-------
- lisp-butt-mode.el#L30: Prefer `https` over `http` (if possible)

marcowahl:
Thanks!  I took your findings into account and here you go.

You must believe me, my local checkdoc passed without issues.

riscy:
Thanks -- I know many people have `sentence-end-double-space` set to `nil` and if your `checkdoc-proper-noun-list` was empty you wouldn't have seen any of those.  Those semi-automated checks are a work in progress.

Looking at this a bit more now, I can see when I activate `lisp-butt-mode` that everything works as expected, even up to rainbow-delimiters, which is great.  But when I try to deactivate the mode with <kbd>M-x lisp-butt-mode</kbd>, the modifications don't go away (Emacs 26.2, tested on macOS and Ubuntu).  This may be related: even though the function `lisp-butt-unfontify` is defined in your code, you never call it anywhere -- and even if you did, I don't know if `font-lock-unfontify-region` is exactly (or only) what you'd want this function to do, because there might be other fontifications on a given `)` .

marcowahl:
Thanks.  The butts go away in my Emacs habitat which is Arch with a relatively recent git Emacs.

The refresh function call was possibly not okay.  This is changed now in the repo.  Would be awesome, if you could give it another try!

lisp-butt-unfontify is a little extra function for the user to just unfontify the butt at point.  This is thought for cases when lisp-butt-mode is wanted in general but(!) just this one butt shall reveal its true nature.  In this case I see it a minor issue what other fontifications are set on this butt.

riscy:
> lisp-butt-unfontify is a little extra function for the user to just unfontify the butt at point. This is thought for cases when lisp-butt-mode is wanted in general but(!) just this one butt shall reveal its true nature. In this case I see it a minor issue what other fontifications are set on this butt.

Ah now I see. I noticed this function was removing colors but not the butt, but(however) if we use `decompose-region` instead, I think we'll always get the intended effect:
```elisp
(defun lisp-butt-unfontify ()
  "Unfontify Lisp butt at point."
  (interactive)
  (while (string= ")" (buffer-substring-no-properties (1- (point)) (point)))
    (goto-char (1- (point))))
  (re-search-forward ")*")
  (unless (= (match-beginning 0) (match-end 0))
    (decompose-region (1+ (match-beginning 0)) (1- (match-end 0)))))
```

> The refresh function call was possibly not okay. This is changed now in the repo. Would be awesome, if you could give it another try!

It didn't work yet -- I think we still want to decompose the region after the mode is deactivated.  What if you did something like this:
```elisp
(defun lisp-butt-unset-slim-display ()
  "Function to undo the nicer Lisp butts."
  (cl-assert (derived-mode-p 'lisp-mode 'emacs-lisp-mode))
  (font-lock-remove-keywords
   nil
   '((")\\())+\\))"
      (1 (compose-region
          (match-beginning 1) (match-end 1)
          (char-to-string lisp-butt-hole))
         nil))))
  (save-excursion
    (goto-char (point-min))
    (while (re-search-forward ")\\())+\\))" nil t)
      (lisp-butt-unfontify))))
```
At this point we have three occurrences of the string `")\\())+\\))"` in the code, so it's probably worth factoring out into a `defconst`, or maybe even a customizable.

purcell:
Couple more notes from me:
- Prefer `when` or `unless` for `if` forms with only one branch.
- Checking `derived-mode-p` against hard-coded values is inappropriate. Instead provide a list of modes.
- The `cl-assert` on `derived-mode-p` is overly defensive, and ends up repeating that unwise code. There's no reason this mode couldn't work in `clojure-mode`, for example, but these assertions will make it explode needlessly there. (`clojure-mode` is a derived mode of `prog-mode`, not `lisp-mode`.)

marcowahl:
Thanks for the feedback.

@purcell I changed the code according your notes.

@riscy: Your issue seems to be more difficult to deal with.  Unfortunately the variation you proposed, i.e. to use `decompose-region` instead of `font-lock-unfontify-region`, does not change the appearance of the butt with my settings.  You wrote you "think" that `decompose-region` does the trick on your side.  Can you confirm this?

riscy:
Here's my buffer (note I tested with and without rainbow-parens enabled, just in case):
image

Using the code in your repository, I activate the mode:
image

I deactivate the mode as I described above.  I see the message "Lisp-Butt mode disabled in current buffer", but my buffer looks exactly the same as above.

I move my cursor to the second-last character and run `lisp-butt-unfontify`.  My buffer looks like this:

image

If I replace your functions with the ones I wrote above, everything unfolds correctly.  Am I using the mode wrong?

marcowahl:
Thanks for your efforts.

At least turning on lisp-butt-mode works as expected.  :-)

I think you use lisp-butt-mode the right way.  The mode is either on or off and then there is this one extra function.

This issue you found looks severe to me.  Unfortunately I can not reproduce the behavior you observe.  And your suggestions break the functionality on my side.

Honestly I don't know what to do.  Shall I withdraw the pull-request?

riscy:
Well maybe we can fix it (or someone out there knows how).  Which version of Emacs are you using?

riscy:
Here, let's try this:
```elisp
;; reset this buffer
(progn
  (decompose-region (point-min) (point-max))
  (emacs-lisp-mode))

(setq regexp ")\\())+\\))")

(setq pattern `((,regexp
                (1 (compose-region
                    (match-beginning 1) (match-end 1)
                    lisp-butt-hole)
                   nil))))

;; apply compose
(progn
  (font-lock-add-keywords nil pattern)
  (font-lock-flush))

;; unapply -- method 1
(progn
  (font-lock-remove-keywords nil pattern)
  (font-lock-flush))

;; unapply -- method 2
(progn
  (save-match-data
    (save-excursion
      (goto-char (point-min))
      (while (re-search-forward regexp nil t)
        (decompose-region (match-beginning 0) (match-end 0))))))

;; unapply -- method 3: both combined!
(progn
  (font-lock-remove-keywords nil pattern)
  (save-match-data
    (save-excursion
      (goto-char (point-min))
      (while (re-search-forward regexp nil t)
        (decompose-region (match-beginning 0) (match-end 0))))
    (font-lock-flush)))
```
Method 1 is yours (as far as I can tell).  Method 2 is sort of what I was suggesting.  Method 3 is both combined!  Methods 2 and 3 work for me for removing the effect.  Maybe method 3 works for both of us?

marcowahl:
Your suggestions looks good.  I "merged" your code and nothing changes for me, which is great, of course.

Thanks so much!

Meanwhile I tried lisp-butt-mode for emacs -Q.  Et voila: switching off the mode fails.  And with your suggestion the mode turns off.

I try to investigate this issue.

marcowahl:
Just an organizeatorial note.  I'm off the grid for about two weeks.

If you find the state of the package as it is acceptable feel free to pull the thing.

Best regards

riscy:
I think `lisp-butt-unfontify` still won't work as intended using `emacs -Q`, and it strikes me that the function also doesn't need to be autoloaded.

Could this checkdoc warning:
```
Warning (emacs): 
lisp-butt-mode.el:81: Disambiguate global-lisp-butt-mode by preceding w/ function,command,variable,option or symbol.
```
be fixed perhaps by rewording the comment as `"Modes considered when running the `global-lisp-butt-mode' command."`?

Ping when you're back!

purcell:
I'll go ahead and merge this, but if you still have questions about how to make things work, feel free to post them here.

marcowahl:
PING!

Thanks for the conversation and thanks for merging the package, @riscy and @purcell.

I added a few tests which give some confidence, that the basic functionality is there for "emacs27 -Q" and also with my config.  BTW I added another explilcit decompose to function lisp-butt-unfontify.

Further the global mode 'global-lisp-butt-mode' has now the name 'lisp-butt-global-mode' which most likely puts the checkdoc warning you (@riscy) mentioned into nirvana.

Happily I integrate any improvements which might come with time from various directions.


Have a nice day!

PR #6287, opened 2019-07-12
Add recipe for evil-ruby-text-objects

porras:

Adds some text objects and keybindings to work with Ruby code with Evil.

https://github.com/porras/evil-ruby-text-objects

Author.

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/evil-ruby-text-objects
+++ b/recipes/evil-ruby-text-objects
@@ -1,0 +1,1 @@
+(evil-ruby-text-objects :fetcher github :repo "porras/evil-ruby-text-objects")


riscy:

<kbd>M-x checkdoc</kbd>:
```
Warning (emacs): 
evil-ruby-text-objects.el:6: Name emacs should appear capitalized as Emacs
Warning (emacs): 
evil-ruby-text-objects.el:15: All variables and subroutines might as well have a documentation string
```

No issues! 🎊

No byte-compilation issues! 😎

No issues! 🎉

-------
- [#L81](https://github.com/porras/evil-ruby-text-objects/blob/master/evil-ruby-text-objects.el#L81): It is beneficial to sharp-quote hooks; use `#'`
- [#L23](https://github.com/porras/evil-ruby-text-objects/blob/master/evil-ruby-text-objects.el#L23): Use `unless ...` instead of `when (not ...)`

porras:
Thanks for the feedback! I'm a newbie and while I addressed some of the issues the linter showed, I had misinterpreted the two remaining ones :roll_eyes: (e.g. as a not native English speaker, I didn't interpret _might as well_ as _must_ but quite the opposite as _nice but not really necessary for internal/private stuff_).

I'd have a question about the capitalization issue, maybe you know the answer. It's triggered by the [`Package-Requires`](https://github.com/porras/evil-ruby-text-objects/blob/master/evil-ruby-text-objects.el#L6), but I've seen the Emacs version dependency expressed like that in a lot of packages. Is that not correct?

I'll ping you when I fix this anyway, thanks!

porras:
@riscy I have fixed all but the capitalization one, I really don't know what should be done there :confused: (see previous comment).

riscy:
I think it's happening because your ```;;; Commentary``` begins before your `Package-Requires` -- you could try putting them in this order:
```elisp
;; Authors:
;; Created:
;; Keywords:
;; URL:
;; Package-Requires:
;; Version:

;; <license summary>

;;; Commentary:

;; <commentary>

;;; Code:

;; <code>
```

porras:
That was exactly it! Thank you, I just fixed it and I think it should be good to go now :smile_cat:

riscy:
Thanks!  I'm able to take a closer look now, and this really looks useful.

In most cases, simply loading a package shouldn't change the behavior of Emacs (e.g. maybe someone wanted to rely on your package to reuse `evil-ruby-text-objects--evil-range`) so a line like [this](https://github.com/porras/evil-ruby-text-objects/blob/master/evil-ruby-text-objects.el#L111) should be left up to the user to add to their config.

purcell:
Yep, good feedback from @riscy there, thanks. I also wonder why use `enh-ruby-mode`. You should be able to get the same functionality here from the built-in `ruby-mode`, which is much better maintained and will be pre-installed in everyone's recent Emacs. No need to make non-users of `enh-ruby-mode` install that package if it can be avoided.

porras:
Wow, thank you for the double code review, I'm an elisp newbie so it's really helpful. I hope it's okay that I do some follow-up questions.

@riscy thank you, I doubted that that was okay but I figured it'd be, since the key binding wouldn't trigger until the user switches to `enh-ruby-mode`. The minor mode solution makes much more sense, I will implement that and ping you again.

@purcell the reason I based it in `enh-ruby-mode` is 1) that's the one I use :roll_eyes: and 2) there is no equivalent in `ruby-mode` to `enh-ruby-up-sexp`, which makes implementing this feature almost trivial. `ruby-mode` has _some_ related movement functions that fiddle around with regexps (as opposed to `enh-ruby-mode`, which keeps around the code tree, using ruby's own parser, which makes it more reliable for this kind of work). But on second thought you are right, it'd be nice to drop that dependency for people who are not already using it, and it _should_ be possible to create an `up-sexp` equivalent using `ruby-mode`'s building blocks. Maybe detecting which mode you are using and providing this replacement when `enh-ruby-up-sexp` is not available. I don't know if that exceeds my elisp skills but I will give it a try!

Thank you again, it was really helpful.

porras:
> there is no equivalent in `ruby-mode` to `enh-ruby-up-sexp`

It turns out that this is not true :roll_eyes: Through what I just learned is the standard way to do this in Emacs, you can get the same with `backward-up-list`. So, while the `ruby-mode` equivalent functions for _up_, _beginning_ and _end_ don't do exactly the same as the `enh-ruby-mode` in all cases, with some fiddling I'm close to get everything working the same in both modes :tada: The main roadblock are oneline methods, which `enh-ruby-mode` gets just right and `ruby-mode` really fails because regexps, but I might just drop support for that, it's not a really important usecase (after all you can very easily operate on the line). I'm learning a lot with this project, which started just as extracting something I had half working on my config!

porras:
Hi @riscy and @purcell, is it ok if I ask you for feedback once again?

I have implemented some changes so that the ruby mode you're using will be detected and used to navigate the code, in this branch: https://github.com/porras/evil-ruby-text-objects/blob/support-ruby-mode/evil-ruby-text-objects.el As a refactoring, I have used the `eieio` object system to abstract away some duplicated code and conditionals. I wondered if it's overkill but since I couldn't find it mentioned as a practice to avoid and it's not an extra dependency I went for it. I think it simplifies the code but I'm biased because I have little experience with elisp and much more with object oriented languages and patterns :smile_cat: I'd be grateful if you let me know what you think, if you have a minute.

riscy:
I think supporting both modes is a great move -- the abstraction looks fine to me, I realize this is a work in progress, but it's easy to run a few quick checks on this diff:

```
evil-ruby-text-objects.el:39:1:Warning: Unused lexical argument `navigator'
evil-ruby-text-objects.el:43:1:Warning: Unused lexical argument `navigator'
evil-ruby-text-objects.el:47:1:Warning: Unused lexical argument `navigator'
evil-ruby-text-objects.el:50:1:Warning: Unused lexical argument `navigator'
evil-ruby-text-objects.el:57:1:Warning: Unused lexical argument `navigator'
evil-ruby-text-objects.el:60:1:Warning: Unused lexical argument `navigator'
evil-ruby-text-objects.el:63:1:Warning: Unused lexical argument `navigator'
evil-ruby-text-objects.el:67:3:Warning: Unused lexical argument `keyword'
evil-ruby-text-objects.el:67:3:Warning: Unused lexical argument `navigator'

In end of data:
evil-ruby-text-objects.el:155:1:Warning: the following functions are not known
    to be defined: ruby-beginning-of-block, ruby-end-of-block,
    enh-ruby-beginning-of-block, enh-ruby-end-of-block, enh-ruby-up-sexp
```
For those you should `(require 'ruby-mode)`, since it is a part of Emacs, and you can use the following to pacify the compiler re: the enh functions:
```elisp
(declare-function enh-ruby-beginning-of-block "ext:enh-ruby-mode.el")
...
```

Misc:
- Unless I'm missing some nuance, it would be clearer to use `unless` [here](https://github.com/porras/evil-ruby-text-objects/blob/support-ruby-mode/evil-ruby-text-objects.el#L87-L99), instead of the `or` (no `progn` required).
- You can use the function `1-` [here](https://github.com/porras/evil-ruby-text-objects/blob/support-ruby-mode/evil-ruby-text-objects.el#L96).
- Some of the extremely long lines like [here](https://github.com/porras/evil-ruby-text-objects/blob/support-ruby-mode/evil-ruby-text-objects.el#L35) and [here](https://github.com/porras/evil-ruby-text-objects/blob/support-ruby-mode/evil-ruby-text-objects.el#L109) would benefit from being linewrapped (e.g. M-q) -- also makes the online help look better.

porras:
Hi, thank you once again, that was really useful. I have addressed all those issues. The only one where I have done different than what you proposed is using `declare-function` also for the ruby-mode functions, instead of `require` (even if it's installed and available, there should be no reason to load it if it's not going to be used; probably premature optimization but I guess it's fine).

The changes no longer trigger warnings and are already in master. Thank you for your help getting there, I think the package really improved thanks to your feedback and reviews. I wasn't sure at the beginning it was a good idea to submit it to melpa but just because of that it was totally worth it! :heart:

riscy:
None of these items are blockers, and this looks good to merge now, but I'll add them for your consideration:

- The double-backticks [here](https://github.com/porras/evil-ruby-text-objects/blob/781f134d9484481b0b4ad32f9cfe90dc00219902/evil-ruby-text-objects.el#L92) should probably be e.g. ``` `ruby-mode' ``` (i.e. leading backtick, trailing apostrophe) -- this just makes the online help a little better by creating hyperlinks.
- Perhaps some of [this](https://github.com/porras/evil-ruby-text-objects/blob/781f134d9484481b0b4ad32f9cfe90dc00219902/evil-ruby-text-objects.el#L43-L55) documentation could be folded into the class definition behind a `:documentation` tag.
- [This](https://github.com/porras/evil-ruby-text-objects/blob/781f134d9484481b0b4ad32f9cfe90dc00219902/evil-ruby-text-objects.el#L22) URL could use https.

...and, welcome to MELPA! :)

porras:
:tada: thanks a lot for all the tips and the welcome :smiley: I want to add that I've been making small contributions to different open source projects for more than 10 years and this was by far the best and most supportive review process. Really appreciated :heart_decoration:

riscy:
Thanks @porras, it's so good to hear that. :)

PR #6291, opened 2019-07-14
add recipes/vterm-toggle

jixiuf:
toggles between the vterm buffer and whatever buffer you are editing.

https://github.com/jixiuf/vterm-toggle

the author.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/vterm-toggle
+++ b/recipes/vterm-toggle
@@ -1,0 +1,2 @@
+(vterm-toggle :fetcher github
+	:repo "jixiuf/vterm-toggle")


recipes/vterm-toggle
```diff
@@ -0,0 +1,6 @@
+(vterm-toggle :fetcher github
+	:repo "jixiuf/vterm-toggle"
+    :files ("*"
+        (:exclude "LICENSE" ".dir-locals.el" ".gitignore" ".clang-format" ".travis.yml"))
```
riscy: This repository feels like it could use the [single file recipe](https://github.com/melpa/melpa#example-single-file-repository) unless I'm missing something.

jixiuf:
fixed, thanks for reviewing

purcell:
Thanks! More notes from me:

- Instead of including any evil-related code here, and then having to be careful that it will still work even if `evil` is not in use, just provide and run hooks for when the vterm is entered and left. Then an Evil user can easily add functions to the hook that set the evil state as desired. Much more idiomatic!
- If I recall correctly, there's a way to have `cl-loop` give you an index variable automatically for cases like [this](https://github.com/jixiuf/vterm-toggle/blob/22f97f2542f8f536d45af5b0a0d1368932e8d9ec/vterm-toggle.el#L278).

Hope that helps!

jixiuf:
@purcell  done.

riscy:
Looks like everything mentioned has been fixed in the code, and this looks good to merge.  Thanks @jixiuf!

PR #6292, opened 2019-07-15
Add jeison recipe

SavchenkoValeriy:

Jeison is a library for a declarative JSON parsing. It maps EIEIO classes' slots onto specific JSON values and constructs these objects from the given JSON objects.

https://github.com/SavchenkoValeriy/jeison

**None needed**

### Checklist

- [ ] My elisp byte-compiles cleanly (there are warnings that don't seem relevant to me)
- [ ] `M-x checkdoc` is happy with my docstrings (there are warnings that don't seem relevant to me)

**NOTE 1** `checkdoc` doesn't like mentions of `jeison--slot` in docstrings.
**NOTE 2** Byte compilation produces 4 warnings: "‘name’ is an obsolete variable (as of 25.1); use 'name instead", it doesn't seem relevant to me (or I simply don't understand what it means).

### DIFF
--- a/recipes/jeison
+++ b/recipes/jeison
@@ -1,0 +1,3 @@
+(jeison
+ :repo "SavchenkoValeriy/jeison"
+ :fetcher github)


riscy:
Move this declaration:
```elisp
(defclass jeison--slot nil
...
```
to the top of your file -- before it's referenced -- and that should solve most of your issues.  Ping to let us know if that works out for you.

SavchenkoValeriy:
@riscy done ([71aa36](https://github.com/SavchenkoValeriy/jeison/commit/71aa36af86a1566a5846cae39a47bfb1c160dd27)), but the problems with `checkdoc` and `byte-compile` maintain.

purcell:
Looks good to me in Emacs 26 - merging.

Consider `pcase-let` instead of `-let` from `dash`, though, since `pcase` is built-in. There's not much need for `dash` in Emacs 26 tbh, given built-in `seq`, `cl-lib`, `pcase` etc.

SavchenkoValeriy:
@purcell You are right and I was trying to avoid `dash` for the most part of `jeison`s history, but I needed to use `-separate` and I didn't see a good CL replacement for it.

PR #6296, opened 2019-07-15
add haskell-tng

tseenshe:

A modern alternative to `haskell-mode`

https://gitlab.com/tseenshe/haskell-tng.el

Author.

### Checklist

Please confirm with `x`:

  - DISCLAIMER: I disagree with some of the lint messages (doesn't like `:` for package-internal symbols, maybe I'll use double dash in a future version)

### DIFF
--- a/recipes/haskell-tng
+++ b/recipes/haskell-tng
@@ -2,4 +2,4 @@
  :repo "tseenshe/haskell-tng.el"
  :branch "tng"
  :fetcher gitlab
- :files ("*.el" "snippets" ".nosearch"))
+ :files (:defaults "snippets"))
--- a/recipes/haskell-tng
+++ b/recipes/haskell-tng
@@ -1,0 +1,4 @@
+(haskell-tng
+ :repo "tseenshe/haskell-tng.el"
+ :branch "tng"
+ :fetcher gitlab)
--- a/recipes/haskell-tng
+++ b/recipes/haskell-tng
@@ -1,4 +1,5 @@
 (haskell-tng
  :repo "tseenshe/haskell-tng.el"
  :branch "tng"
- :fetcher gitlab)
+ :fetcher gitlab
+ :files ("*.el" "snippets" ".nosearch"))


purcell:
>     * I disagree with some of the lint messages (doesn't like `:` for package-internal symbols, maybe I'll use double dash in a future version)

I haven't had a full look yet, but that's a blocker for merging, sorry -- these days we only add packages that follow the community conventions, reasoning that the consistency the conventions promote is a key factor in community contributions to (and collective ownership of) code. Double dash would be the way to go here.

Also noticed you're using keybindings that are reserved for end-user customisation, which `package-lint` didn't detect because it doesn't know about `bind-key`. (You should just use `define-key` rather than depend on `bind-key` for programmatic cases like this. `bind-key` is for use in personal configs.)

tseenshe:
I have renamed all symbols to use a `haskell-tng--` naming convention, and removed the `bind-key` dependency.

purcell:
Sweet, thanks. Will follow up with a more complete review soon. I’m a bit behind!

purcell:
Great, had time to look through this. Epic piece of work, and very exciting. Nice that it's ultimately very compact.

Some notes:

- The provided `:files` spec in the recipe should be written `(:defaults "snippets" ".nosearch")`, though I question the need for `.nosearch` in the directory of an installed elisp package.
- The stuff in `-contrib.el` should be 3+ separate small packages so that there's no need for soft dependencies. It's trivial for a `yasnippet` user to install `haskell-tng-snippets`, so break these things out.
- [This `boundp` check](https://gitlab.com/tseenshe/haskell-tng.el/blob/6acba95224953bae57253f3c2017cafd55f29a8b/haskell-tng-contrib.el#L45) is bad practice, because presence of a var/function doesn't imply that a user wants to use it. It also shouldn't be autoloaded, though in a separate `haskell-tng-snippets` package you could use `(with-eval-after-load 'yasnippet ...)` _and_ `autoload` to do this safely.
- You might consider using my `reformatter` lib for [things like `stylish`](https://gitlab.com/tseenshe/haskell-tng.el/blob/6acba95224953bae57253f3c2017cafd55f29a8b/haskell-tng-contrib.el#L26-32) -- I wrote it because there are so many ways to accidentally mess up this sort of code, and it also gives people an orthogonal way to enable/disable stylish. You (or I) could (should?) even just publish a separate `haskell-stylish` package for that.
- Why [redefine the global mode map](https://gitlab.com/tseenshe/haskell-tng.el/blob/6acba95224953bae57253f3c2017cafd55f29a8b/haskell-tng-mode.el#L85-95) every time a new Haskell buffer is opened? This defeats user customisation: just fill the mode map at the top level with `define-key` forms after the `define-derived-mode` form has auto-declared the map for you.
- `C-c LETTER` bindings are reserved for end-user customisation, so may not be used by a package. `package-lint` would normally catch this, but it doesn't understand your custom `bind` function enough to warn you. This applies to multiple files, including `haskell-tng-compile.el`.
- Nice use of the custom compilation mode.
- Using `prettify-symbols-mode` in the compilation mode ([commented-out currently](https://gitlab.com/tseenshe/haskell-tng.el/blob/6acba95224953bae57253f3c2017cafd55f29a8b/haskell-tng-compile.el#L120-127)) is something I'd avise against because it's hacky.
- As a general point, you really don't need `dash` these days, given `seq` and `cl-lib`. I personally prefer to use these built-in libraries for basics like list manipulation, since they're more standard and contributors, and even if you personally use `dash`, some contributors will give you code that uses built-in equivalent functions.
- It feels pretty wild to [override](https://gitlab.com/tseenshe/haskell-tng.el/blob/6acba95224953bae57253f3c2017cafd55f29a8b/haskell-tng-mode.el#L50) the user's preference for `tab-width` and `indent-tabs-mode`. There may be technical reasons (e.g. your parser support) for that, but even as a supporter of those exact settings, I feel like they're not something to impose on everyone in the long term. :-)

Hope that helps!

tseenshe:
@purcell that all sounds fair.

Good catch on redefining the keymap.

I expect most users will want `C-c c` to just work so I will add those to the installation instructions.

I'm not sure what you had in mind for formatting or what could go wrong, but I'll look at your lib and use it if it makes sense (I didn't know about it!). Admitedly, the popup could do with some work (e.g. `q` should kill it and it should grab the user focus) so I'd be happy not to have that techdebt.

When it comes to wrappers for external haskell tools, my plan is to put them all into the `-contrib` file if there is nowhere else for them to live, along with other untested utility functions (e.g. I have one to copy the current module name).

I agree that the third party elisp deps should be in separate files, that's much less hacky.

I've given up on prettify-symbol in the compilation mode but I'd still like that functionality somehow. Perhaps just as a user.

I considered using seq and cl-lib but I don't know them as well and didn't feel like the digression. Perhaps it's time to return to that, it'd be nice to remove a dep.

I'm cargo culting the `.nosearch`, happy to remove it.

With regards to the tab settings, this is actually specified in the Haskell2010 report so for Haskell I think this is fully justified. I wouldn't do this unless the language spec called for it.

I'll let you know when I've made these changes so you can merge.

tseenshe:
@purcell I have responded to your review. Thank you for that. If you are happy with the changes, I think this is good to merge.

recipes/haskell-tng
```diff
@@ -0,0 +1,5 @@
+(haskell-tng
+ :repo "tseenshe/haskell-tng.el"
+ :branch "tng"
+ :fetcher gitlab
+ :files (:defaults "snippets"))
```
riscy: Is `haskell-tng-testutils.el` only used for testing?  If so, you may need to explicitly :exclude it in this recipe.

riscy: Disregard -- I missed that it was in a subdirectory. :)

tseenshe:
@riscy I am happy to make the small stylistic changes but the `-contrib-*` modules are optional. I don't know of a way to have an optional dependency with `melpa`/`use-package` and it was my understanding that the convention was to let the bytecompile fail but require the user to opt-in to usage.

I originally put these things behind `fboundp` and `boundp` but @purcell said not to do that.

tseenshe:
@riscy @purcell 

Thank you both for your kind reviews of this package. I am going to make the suggested changes but I have also realised that I don't think this package is ready for publication on MELPA just yet, for several reasons:

1. I have approximately ~100 open TODOs that I plan on returning to
2. It is not yet feature complete. I want to add support for the REPL, ghcid, hlint, hsimport, and flycheck.
3. the semantic backend `hsinspect` has hit some big problems that will confuse users. I want to work with the cabal-install / cabal-helper team to work through these to make a seemless experience
4. I want to move lexing into the syntax-table stage (adding categorios) and rewrite all regexps to make use of this markup, which would fix a lot of problems.

At this point I think `haskell-tng` is only really suitable for Haskell developers who are capable of jumping into the emacs-lisp and ghc code to help out. It would cause a lot of confusion if, say, a beginner were to install this package instead of `haskell-mode` and might even put them off Haskell entirely.

For those reasons, I would rather wait until I am happy that `haskell-tng` can be easily installed and used by the mythical "typical user". I will be back! :smile:

riscy:
Sounds great!  Also, thanks for being so quick to tackle issues.

PR #6297, opened 2019-07-15
add proced-narrow

travisjeffery:

This package provdes live filtering of processes in proced buffers.  In general, after calling
proced-narrow you type a filter string into the minibuffer to filter the list of processes.  After
each change proced-narrow automatically reflects the change in the buffer. 

https://github.com/travisjeffery/proced-narrow

Author and maintainer.

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/proced-narrow
+++ b/recipes/proced-narrow
@@ -1,0 +1,3 @@
+(proced-narrow
+ :repo "travisjeffery/proced-narrow"
+ :fetcher github)


riscy:

<kbd>M-x checkdoc</kbd>:
No issues! 💯

No issues! 🎊

```
proced-narrow.el:92:1:Warning: Unused lexical variable
    `proced-narrow-filter-function'
proced-narrow.el:118:1:Warning: Variable `proced-narrow-buffer' declared after
    its first use
proced-narrow.el:124:1:Warning: Variable `proced-narrow-filter-function'
    declared after its first use
```

No issues! 🎉

-------
- [#L130](https://github.com/travisjeffery/proced-narrow/blob/master/proced-narrow.el#L130): It is beneficial to sharp-quote hooks; use `#'`
- [#L133](https://github.com/travisjeffery/proced-narrow/blob/master/proced-narrow.el#L133): It is beneficial to sharp-quote hooks; use `#'`

travisjeffery:
Thanks, I've addressed all the issues.

purcell:
Thanks!

- The `:files` spec in the recipe is unnecessary, so you can remove it.
- Don't put a `Package-Version` [header](https://github.com/travisjeffery/proced-narrow/blob/e93fd5a46afc0c1dc6924290dad6c64570b60bb8/proced-narrow.el#L8) in the file. You don't need it because MELPA will add one. `Version` is enough to keep `package-lint` happy.
- With [this autoload](https://github.com/travisjeffery/proced-narrow/blob/e93fd5a46afc0c1dc6924290dad6c64570b60bb8/proced-narrow.el#L133), anything that uses the minibuffer before this file is loaded will explode. Removing the `;;;###autoload` cookie should fix this.
- You might consider preferring `seq` (or `cl-lib`) to `dash`, since it's built into newer Emacsen, and still available as a separate package for older Emacsen.

Hope that helps!

travisjeffery:
@purcell @riscy thanks I've push fixed for all your comments. Removed the :files spec, removed the Package-version header, removed that autoload, and switched to seq.

riscy:
It looks like the latest `seq` is 2.20 [from Elpa](https://elpa.gnu.org/packages/seq.html) which package-lint complains about too.

Other than that, everything appears to have been addressed -- feel free to ping when ready.

travisjeffery:
Ah, I use emacs 27 and looked at its version of seq.  just pushed an update to use 2.20.

If you mean me to ping instead of @purcell I'm ready anytime.

travisjeffery:
Thanks, @riscy.

travisjeffery:
@riscy @purcell :ping_pong:

riscy:
Doing some user testing on this -- sorry for the long turnaround.

I think you need to `(require 'delsel)` for `minibuffer-keyboard-quit` (`delsel` isn't always necessarily loaded).  It would probably be worthwhile to sharp-quote the functions on [these](https://github.com/travisjeffery/proced-narrow/blob/master/proced-narrow.el#L63-L65) lines -- flycheck would've told you about the issue with `minibuffer-keyboard-quit`.

This helpful hint:

> To bring it back to the original view, you can call `revert buffer' (usually bound to g).

...feels like it belongs in something users will typically access, like the docstring for the mode or the `proced-narrow` function.

Some minor stuff I'm noticing as I look at the source that might be helpful:
- `provdes` -> `provides`
- `minibufferto` -> `minibuffer to`
- `origininal` -> `original`
- `tthem` -> `them`

travisjeffery:
@riscy sorry about those. I've pushed fixes for all of them. Thanks.

riscy:
Great thanks.  I would still recommend sharp-quoting the functions [here](https://github.com/travisjeffery/proced-narrow/blob/df5cce50b3d1219b23d28e23cbf68e0c7807a15c/proced-narrow.el#L64-L66) (it's a good habit to get into).

For this documentation line [here](https://github.com/travisjeffery/proced-narrow/blob/df5cce50b3d1219b23d28e23cbf68e0c7807a15c/proced-narrow.el#L77), I'm realizing you could write the last sentence as: 
```elisp
"To bring it back to the original view, use \\[revert buffer]."
```
This way Emacs will automatically determine the keybinding (i.e. whether it's <kbd>g</kbd> or who-knows-what) when the help is displayed.

Those are more tips for your consideration than blockers, though, and this looks good.  Welcome to MELPA. :)

PR #6300, opened 2019-07-16
Recipe: Add org-ql

alphapapa:

`org-ql` is a lispy query language for Org files. It allows you to find Org entries matching certain criteria and return a list of them or perform actions on them. Commands are also provided which display a buffer with matching results, similar to an Org Agenda buffer.

https://github.com/alphapapa/org-ql

Author and maintainer.

### Checklist

Please confirm with `x`:

- [x] I've used the latest version of [package-lint](https://github.com/purcell/package-lint) to check for packaging issues, and addressed its feedback (*see notes*)
- [x] My elisp byte-compiles cleanly (*see notes*)

# Notes

* Using the latest version of `package-lint`, it's showing these spurious errors for top-level forms in `org-ql.el`, and a similar list in `org-ql-agenda.el`:
```
20:1: error: "cl-lib" doesn't start with package's prefix "org-ql".
21:1: error: "org" doesn't start with package's prefix "org-ql".
22:1: error: "org-habit" doesn't start with package's prefix "org-ql".
23:1: error: "seq" doesn't start with package's prefix "org-ql".
24:1: error: "subr-x" doesn't start with package's prefix "org-ql".
26:1: error: "dash" doesn't start with package's prefix "org-ql".
30:1: error: "when" doesn't start with package's prefix "org-ql".
53:1: error: "cl-defmacro" doesn't start with package's prefix "org-ql".
63:1: error: "cl-defmacro" doesn't start with package's prefix "org-ql".
114:1: error: "cl-defun" doesn't start with package's prefix "org-ql".
234:1: error: "define-hash-table-test" doesn't start with package's prefix "org-ql".
308:1: error: "cl-defun" doesn't start with package's prefix "org-ql".
```

* Also, the `*Package-Lint*` buffer doesn't properly jump to lines when I press `RET` on the items.  Something seems to be wrong with it.  Is it just me?

* These errors from `package-lint` are not spurious, but the aliases are necessary for compatibility with Org versions 9.0-9.1:
```
31:2: error: Aliases should start with the package's prefix "org-ql".
32:2: error: Aliases should start with the package's prefix "org-ql".

54:2: error: Aliases should start with the package's prefix "org-ql-agenda".
```

* As well, when byte-compiling, the `org-get-tags` compatibility alias may cause a spurious warning about number of arguments, depending on the version of Org being used.  It may be safely ignored, and won't matter in the future when support for Org <9.2 is dropped.

### DIFF
--- a/recipes/org-ql
+++ b/recipes/org-ql
@@ -1,0 +1,1 @@
+(org-ql :fetcher github :repo "alphapapa/org-ql")


riscy:
I'm not totally versed in the `org-get-tags` vs `org-get-tags-at` issue, but what about doing defining a function like:
```elisp
(defun org-ql-get-tags ()
  (if (version< org-version "9.2")
      (org-get-tags-at)
    (org-get-tags)))
```
and then just using `org-ql-get-tags` everywhere in your code?  (Maybe I'm not aware of the deeper interactions; if so, sorry).  The current solution feels like it will clobber my `org-get-tags` function until I reload org-mode.

Since not everyone will have `org-super`, the `declare-function` can mark this as an external file e.g. `(declare-function org-super-agenda--group-items "ext:org-super-agenda")`

And some minor checks I'm automating (so grain of salt):

- org-ql.el#L47: Alias targets should be sharp-quoted; use `#'`
- org-ql.el#L48: Alias targets should be sharp-quoted; use `#'`
- org-ql.el#L4: Prefer `https` over `http` (if possible)
- org-ql.el#L30: Prefer `https` over `http` (if possible)
- org-ql-agenda.el#L157: Prompt strings should usually end with a space
- org-ql-agenda.el#L49: Alias targets should be sharp-quoted; use `#'`
- org-ql-agenda.el#L4: Prefer `https` over `http` (if possible)
- org-ql-agenda.el#L25: Prefer `https` over `http` (if possible)

alphapapa:
> I'm not totally versed in the `org-get-tags` vs `org-get-tags-at` issue, but what about defining a function like:
> 
> ```emacs-lisp
> (defun org-ql-get-tags ()
>   (if (version< org-version "9.2")
>       (org-get-tags-at)
>     (org-get-tags)))
> ```
> 
> and then just using `org-ql-get-tags` everywhere in your code?

That would cause the version check to run every time the function is called, which would be unnecessary and cause poor performance.  This function is called repeatedly in queries that get tags.

>  (Maybe I'm not aware of the deeper interactions; if so, sorry). The current solution feels like it will clobber my `org-get-tags` function until I reload org-mode.

The `org-get-tags` function only exists in Org >= 9.2.  In Org 9.0-9.1, only the `org-get-tags-at` function exists.  The alias is only defined in Org < 9.2, where the alias doesn't exist, so it doesn't clobber anything.

> And some minor checks I'm automating (so grain of salt):

Thanks, fixed.

riscy:
For me:
```elisp
(org-version) ; => 9.1.9
#'org-get-tags ; => exists
#'org-get-tags-at ; => exists
```

Also, a line from org-archive.el on version 9.1.9 that makes me leery of clobbering:
```elisp
	  (let* ((all-tags (org-get-tags-at))
		 (local-tags (org-get-tags))
```

Re: poor performance, what about this?

```elisp
(defvar org-ql-prefer-get-tags-at (version< org-version "9.2"))
(defun org-ql-get-tags ()
  (if org-ql-prefer-get-tags-at (org-get-tags-at) (org-get-tags)))
```
Or, it may be prematurely optimizing, but you could even nest two different defuns inside a conditional.

alphapapa:
@riscy Thanks, I took a patch I received for granted instead of verifying that `org-get-tags` was not a function in Org <9.2.  I just pushed a fix that defines a new function to call the appropriate Org function, so the existing aliases are not changed.  The timestamp one shouldn't be a problem, because the new alias doesn't exist in Org <9.2.

riscy:
It looks like you're setting an alias for `org-timestamp-to-time` but not using it, e.g. [here](https://github.com/alphapapa/org-ql/blob/master/org-ql.el#L446-L447) and [here](https://github.com/alphapapa/org-ql/blob/master/org-ql.el#L608-L609).

That said, I'm leery of using a version check to decide whether to define an alias that _might_ (probably won't, but _might_) clobber an existing function.  If you were to do this:

```elisp
;; Compatibility buffer for older versions of `org-mode':
(if (version< org-version "9.2")
    (progn
      (defun org-ql-timestamp-to-time (timestamp &optional end)
        (org-timestamp--to-internal-time timestamp end))
      (defun org-ql--get-tags (&optional pos local)
        (org-get-tags-at pos local)))
  (defun org-ql-timestamp-to-time (timestamp &optional end)
       (org-timestamp-to-time timestamp end))
  (defun org-ql--get-tags (&optional pos local)
    (org-get-tags pos local)))
```

...`package-lint` would be totally happy, and there'd be no (slight) risk of a minor bump to version 9.1 of org-mode  (e.g. 9.1.10) introducing a function that you would be unnecessarily shadowing.

alphapapa:
> It looks like you're setting an alias for `org-timestamp-to-time` but not using it, e.g. [here](https://github.com/alphapapa/org-ql/blob/master/org-ql.el#L446-L447) and [here](https://github.com/alphapapa/org-ql/blob/master/org-ql.el#L608-L609).

Thanks for noticing that.  Another case of my taking a patch for granted instead of verifying that it did everything it needed to.

> That said, I'm leery of using a version check to decide whether to define an alias that _might_ (probably won't, but _might_) clobber an existing function. If you were to do this:
> 
> ```emacs-lisp
> ;; Compatibility buffer for older versions of `org-mode':
> (if (version< org-version "9.2")
>     (progn
>       (defun org-ql-timestamp-to-time (timestamp &optional end)
>         (org-timestamp--to-internal-time timestamp end))
>       (defun org-ql--get-tags (&optional pos local)
>         (org-get-tags-at pos local)))
>   (defun org-ql-timestamp-to-time (timestamp &optional end)
>        (org-timestamp-to-time timestamp end))
>   (defun org-ql--get-tags (&optional pos local)
>     (org-get-tags pos local)))
> ```
> 
> ...`package-lint` would be totally happy, and there'd be no (slight) risk of a minor bump to version 9.1 of org-mode (e.g. 9.1.10) introducing a function that you would be unnecessarily shadowing.

There's no risk of that, because Org doesn't release patches to old branches--9.1 is a dead branch.  And I'd prefer to avoid adding another level of indirection there, as this function is called in tight loops when processing timestamps in entries.

Thanks for your thorough reviews here.

riscy:
Okay, sounds reasonable.  Looks like a very cool package. :)

alphapapa:
Thanks.  Publishing it on MELPA has been a long time coming.

PR #6303, opened 2019-07-19
Add sacred themes

matthias-margush:

Nature inspired dark and light themes.

https://github.com/matthias-margush/sacred-theme

n/a

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/sacred-theme
+++ b/recipes/sacred-theme
@@ -1,0 +1,3 @@
+(sacred-theme
+ :repo "matthias-margush/sacred-theme"
+ :fetcher github)


purcell:
Thanks! This needs moving around a bit. First of all, `sacred-theme.el` is named like a theme and is in a directory on the theme path, so it will potentially show up in the theme list, even though it doesn't itself define a custom theme. I suggest you rename this (and the package/recipe) to `sacred-themes`. Also, then, for more consistent naming, the other files should be `sacred-forest-theme.el` and `sacred-paper-theme.el` (just like the functions you've provided to enable them!).

Apart from that, everything looks good, but consider providing the `DOC` argument to your `deftheme` forms so that there is a human-readable name for each theme.

Hope that helps!

riscy:
@matthias-margush Ping! :)

riscy:

PR #6304, opened 2019-07-19
Add streamline recipe

matthias-margush:

Minimalist's Mode Line

https://github.com/matthias-margush/streamline

n/a

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/streamline
+++ b/recipes/streamline
@@ -1,0 +1,3 @@
+(streamline
+ :repo "matthias-margush/streamline"
+ :fetcher github)


purcell:
Thanks! This is neat, and the code mostly looks good. Couple of notes:

- The keybinding stuff is unnecessarily idiosyncratic and could be made much simpler: you have a global minor mode, so just give that mode a keymap and put the toggle keybinding in it. That's the standard way for users to customise bindings, and changes they make will always be reflected immediately, unlike in the current code. (`C-s-g` uses "Super", right? If so, that's not a great default binding, when not all keyboards can produce Super.)
- The name `streamline--hide` is a bit confusing. Why not `streamline--show-original` for clarity?
- The `eval`s in [these forms](https://github.com/matthias-margush/streamline/blob/master/streamline.el#L47-L49) seem undesirable to me. What am I missing?
- Rather than direct people from the source code [to an online README](https://github.com/matthias-margush/streamline/blob/master/streamline.el#L32-L33), prefer to place a little documentation in the `Commentary`. Among other things, this text is used to document your package on its MELPA page and when people hit `?` in the `M-x list-packages` page.

Hope that helps!

jcs090218:
@matthias-margush Hi, may I know the differences between this package and [feebleline](https://github.com/tautologyclub/feebleline)? :D Because both these two packages look similar to me.

matthias-margush:
Thanks for the PR feedback. I'll be able to take a look next weekend. 

As far as the difference between feebleline and this:
- streamline goes out of its way to display nothing (modeline nor minibuffer), feebleline goes out of its way to display stuff in the minibuffer.
- the author of feebleline admits to using an ugly theme, and I would never do such a thing (:aghast:)

That said, feebleline looks great and  I might use feebleline myself and leave this for the true luddites.

purcell:
Ping! :-)

terlar:
There is also this one, that hides the modeline:
https://github.com/hlissner/emacs-hide-mode-line

Together with the built-in window devider I think it looks the same?
```
(setq window-divider-default-places t
      window-divider-default-bottom-width 1)
(window-divider-mode 1)
```

riscy:
Please feel free to reopen if/when this is ready.  Thanks!

PR #6305, opened 2019-07-19
Add orgraphy recipe

matthias-margush:

Enhance any theme with visual improvements for org mode, markdown

https://github.com/matthias-margush/orgraphy

n/a

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/orgraphy
+++ b/recipes/orgraphy
@@ -1,0 +1,3 @@
+(orgraphy
+ :repo "matthias-margush/orgraphy"
+ :fetcher github)


riscy:
Checklist looks great on a quick first pass.  I'll look in more detail soon...

In the meantime, I couldn't help but try it out and notice my `ivy-read` fonts are changed, for example when I'm switching buffers.  This is apparently because the counsel faces inherit from org-mode, which caught me by surprise, so I'm passing it along fyi (not that there's much you can do but document it for users):

image

matthias-margush:
Interesting. From the `counsel-outline-face-style` doc, it looks like this is a feature `counsel-outline`. It's default setting uses the `counsel-outline-default` face in the outline selection buffer (so no scaling.) It offers a few other choices that will display the outline with scaled fonts in various ways. So from what I can tell, the font scaling here is intended by ivy, and would additionally need to be wanted & turned on by the user explicitly.

purcell:
Thanks! Some notes:

- The font heights and styles are hard-coded, and cannot be customised. This makes the whole file seem like a bundle of personal preferences rather than a package that will be useful to  other people. The theme system supports multiple active themes at one time, so you could simply write an additional theme for those customisations.
- Is [this](https://github.com/matthias-margush/orgraphy/blob/6799e5265e168e518983afd2eb6fec84898c234d/orgraphy.el#L161) really what the minor mode does?
- Why are you [altering `org-mode-hook`](https://github.com/matthias-margush/orgraphy/blob/6799e5265e168e518983afd2eb6fec84898c234d/orgraphy.el#L139-L155) from a global minor mode when you could instead provide a local minor mode that the user can activate in `org-mode-hook`? This is unidiomatic. Whenever you find yourself writing `init` and `deinit` functions, you're probably writing a minor mode. Same applies when you find yourself saving previous values of faces or variables to hash tables etc.
- Same comment here re. linking to the README. At least tell the user roughly how to use the package.

purcell:
Ping! :-)

riscy:

PR #6308, opened 2019-07-20
Add hercules.el recipe.

jjzmajic:

An auto-magical, `which-key` based `hydra` banisher.

In at most 7 lines of set-up code, `hercules` lets you call any group
of related command sequentially with no prefix keys, while showing a
handy popup to remember the bindings for those commands. He can create
both of these (the grouped commands, and the popup) from any
keymap.

https://gitlab.com/jjzmajic/hercules.el

Maintainer.

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/hercules
+++ b/recipes/hercules
@@ -1,0 +1,1 @@
+(hercules :repo "jjzmajic/hercules.el" :fetcher gitlab)


purcell:
Thanks, looks neat. Some notes:

- The documentation phrase ["In at most 7 lines of set-up code"](https://gitlab.com/jjzmajic/hercules.el/blob/7c6d63886724a60aed5568c9ecd80909bc40e4ce/hercules.el#L30) isn't going to age well!
- Don't [start docstrings with a blank line](https://gitlab.com/jjzmajic/hercules.el/blob/7c6d63886724a60aed5568c9ecd80909bc40e4ce/hercules.el#L111)
- When you provide a pair of `enable` and `disable` functions, that's often just an incomplete minor mode. Maybe you can provide a minor mode instead! You can even give it a keymap, which might possibly help you avoid the odd "PSEUDO-MODE" stuff. Does this make sense here?
- [This aliasing](https://gitlab.com/jjzmajic/hercules.el/blob/7c6d63886724a60aed5568c9ecd80909bc40e4ce/hercules.el#L60) is bogus. At the very least, advise the `which-key--hide-popup` function instead to make it do nothing when necessary, e.g. when your new minor mode is enabled. :-)  It should also not be necessary to change `pre-command-hook` if you're changing the behaviour of `which-key--hide-popup` anyway.
- I believe that since "hercules" isn't the name of a var or function, you shouldn't quote it [like this](https://gitlab.com/jjzmajic/hercules.el/blob/7c6d63886724a60aed5568c9ecd80909bc40e4ce/hercules.el#L112) in docstrings.

Hope that helps!

jjzmajic:
Noted and working on it! One thing though, hercules provides an arbitrary number of of enable and disable functions, including 0, and also need to be able to exit it the moment you press a key outside the keymap you pass in. This could be done by creating a mode, and then having each function call that mode to toggle it, but I feel it adds unnecessary complexity and transient keymaps are the way to go. At least that's the path hydra went down. 

Am I making any sense?

jjzmajic:
Better now?

jjzmajic:
Nope, wait, big refactor coming up, sorry about that.

jjzmajic:
Done. A lot has changed though. Sorry to yank your chain like this.

riscy:
Thanks for the update!  I'm seeing a few minor package-lint issues in the latest:
```
3 issues found:

9:1: warning: You should include standard keywords: see the variable `finder-known-keywords'.
10:1: error: Expected (package-name "version-num"), but found (which-key).
123:27: warning: `with-eval-after-load' is for use in configurations, and should rarely be used in packages.
```
For the first, you could perhaps add `convenience` to your keywords list.  For the second, you just need a version after `which-key` (similar to your emacs version requirement).  The third happens inside a defun, so looks fine.

Minor: it looks like in one docstring `hecules-graylist` should be `hercules--graylist`

jjzmajic:
Got it! Can't track down the `hercules-graylist` though. Must have removed it since.

riscy:
[This](https://gitlab.com/jjzmajic/hercules.el/blob/master/hercules.el#L121) line to be more specific. :)

jjzmajic:
Oh god, I didn't see the heCkules part, I though it was just the missing dash. I am totally using that one for my next package. Thanks man! All fixed.

jjzmajic:
Not rushing you guys, we all have our day jobs, but is there anything else I can do? Just so I don't delay things with another stupid mistake.

riscy:
No worries.  Last weekend was a vacation for me and I was in transit yesterday, but I'll take another look this weekend.

riscy:
I just did some user testing.  I came up with this "minimal" example -- let me know if I did something wrong:
```elisp
(defvar hercules-keymap
    (define-key keymap (kbd "!") #'hercules-hi)
    (define-key keymap (kbd "?") #'hercules-bye)
    (define-key keymap (kbd ".") #'hercules-toggle)
    keymap)
  "Keymap for hercules example.")

(defun hercules-toggle ()
  (interactive)
  (message "(Toggle)"))

(defun hercules-hi ()
  (interactive)
  (message "Hi"))

(defun hercules-bye ()
  (interactive)
  (message "Bye!"))

(hercules-def
 :toggle-funs '(hercules-toggle)
 :hide-funs #'hercules-bye
 :show-funs #'hercules-hi
 :keymap 'hercules-keymap)
```
After evaluating all of those forms, things seem to work mostly as I'd expect, with a small issue <kbd>M-x hercules-toggle</kbd>: this brings up which-key with the options `!` `.` and `?`, but if I try any of these options, nothing happens.  If I try one of the options a second time, then it works.  It's like I have to press a key before the options take effect, given the above setup.

This is happening in GUI and Terminal Emacs 26.2 for me.  Am I doing something wrong, or is there an issue that we can sort out?

jjzmajic:
Nope, you're doing everything right. It's definitely something to fix. What's more, it seems that only every other key press registers. Which is strange because this does not happen with say the following setup for macrostep:

```elisp
(hercules-def
           :toggle-funs #'macrostep-mode
           :keymap 'macrostep-keymap)
```
Now that I pay close attention, it doesn't respond to the first key press either. But unlike this example, every subsequent key press is fine. I tried changing the advice from before to :after, but it seems to cause even more trouble... Thoughts?

I've been doing my best to contribute, but I'm still fairly new to elisp, so I could definitely use some help if you have the time.

riscy:
I'm not too familiar with the guts of `which-key`, but I'll poke around -- I think it will take some careful step-by-step debugging to understand what's going on.

jjzmajic:
Think I fixed it. r/emacs to the rescue again.

riscy:
Works great!  Thanks. :)

PR #6315, opened 2019-07-20
add recipe for dispwatch

mnp:

This package watches the current display geometry (pixel width and height) and gives your
hook a call if it changes.  Intended use case is plugging/unplugging a monitor.

https://github.com/mnp/dispwatch

Maintainer

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/dispwatch
+++ b/recipes/dispwatch
@@ -1,0 +1,1 @@
+(dispwatch :repo "mnp/dispwatch" :fetcher github)


purcell:
Thanks! Quick notes:

- Instead of a single-branch `if` and a `progn`, just use `unless` [here](https://github.com/mnp/dispwatch/blob/86905b6f820b1984cf7b9d736348c9bc5bcd8965/dispwatch.el#L90-L93) and elsewhere. (Or `when`, as applicable.)
- Your pair of `enable`/`disable` functions indicate that this is an informal minor mode. Instead, declare a proper global minor mode with `define-minor-mode`.
- It would be preferable to pass the new geometry to hook functions as 2 numeric arguments or a cons cell rather than constructing a geometry string. It's easy enough for the user to call `format` if required, and a cons cell can be compared using `eq`, and passing the dimensions individually makes it easy for hook functions to detect, say, narrow or small displays.
- Even polling every 2 seconds would be minimal overhead, and might be a slightly nicer experience out of the box.

Hope that helps.

purcell:
Ping! :-)

mnp:
Yep, thanks for the reminder, and for the detailed review. Will address all comments.

riscy:
> Ping ACK.

Does this mean you're ready for another pass? :)

mnp:
Yep please have at it.  Thanks again for the kind review.

riscy:
```
1 issue found:

1:64: warning: You should depend on (emacs "24") if you need lexical-binding.
```
Also on dispwatch.el#L3: This copyright notice may need updating?

Once those are checked, this looks good to merge.

mnp:
Comments addressed.  Thanks!

riscy:
Great thanks!  And, welcome to MELPA. :)

PR #6318, opened 2019-07-21
Add recipe for marquee-header.

jcs090218:

Code interface for displaying marquee in header.

https://github.com/jcs090218/marquee-header

The maintainer.

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/marquee-header
+++ b/recipes/marquee-header
@@ -1,0 +1,1 @@
+(marquee-header :repo "jcs090218/marquee-header" :fetcher github)


riscy:

No issues! 💯

No issues! 🎊

```
marquee-header.el:177:1:Warning: the function `spaces-string' is not known to
    be defined.
```
(looks like we need a require on `rect`. I didn't know about this little function, neat)

No issues! 🎉

jcs090218:
Thanks for the review! Fixed! :D

riscy:
Looking more closely at this now.  This is really great.

One small thing I notice is that if I run `(marquee-header-notify "Test")` in reasonably rapid succession, then the marquee gets "stuck" after completion (I think the value of `marquee-header--previous-header-line-format` gets overwritten with a marquee string).  It feels like the current marquee should be cleared before another one is allowed to start.

It's a matter of style, but just in case you're not aware you can write blocks like [this](https://github.com/jcs090218/marquee-header/blob/4648b2689d6bb58317b1e515ee2c1da394d3c76a/marquee-header.el#L96-L101) as:
```elisp
    (setq marquee-header--message msg
          marquee-header--direction direction
          marquee-header--frame-counter frame-counter
          marquee-header--message-decoration deco
          ...)
```

Since [this](https://github.com/jcs090218/marquee-header/blob/4648b2689d6bb58317b1e515ee2c1da394d3c76a/marquee-header.el#L150-L151) function is not interactive, any code using it will presumably `(require 'marquee-header)` which makes the autoload cookie [unnecessary](https://www.gnu.org/software/emacs/manual/html_node/elisp/When-to-Autoload.html#When-to-Autoload).

I think you only need to check `stringp msg` [here](https://github.com/jcs090218/marquee-header/blob/4648b2689d6bb58317b1e515ee2c1da394d3c76a/marquee-header.el#L155-L156).   If it's nil then it's not a string -- similarly with `time` below that, unless I'm missing something.

jcs090218:
Thanks! I think all problems you pointed out have been fixed right now!

riscy:
Looks great.  Nice work!

PR #6321, opened 2019-07-23
Add flymake-quickdef

karlotness:

Helps write diagnostic/backend functions for the new Flymake (functions like the [example in the docs](https://www.gnu.org/software/emacs/manual/html_node/flymake/An-annotated-example-backend.html)).

It provides a skeleton that handles running an external check program, processing its output, and cleaning up scratch buffers and files.

https://github.com/karlotness/flymake-quickdef

I am the author

n/a

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/flymake-quickdef
+++ b/recipes/flymake-quickdef
@@ -1,0 +1,1 @@
+(flymake-quickdef :fetcher github :repo "karlotness/flymake-quickdef")


riscy:
This looks really useful, and thank you for your close adherence to the checklist.  Here are a couple thoughts for your consideration:
- It's almost always beneficial to sharp-quote function names like [here](https://github.com/karlotness/flymake-quickdef/blob/master/flymake-quickdef.el#L223).
- As a small matter of style, [this](https://github.com/karlotness/flymake-quickdef/blob/master/flymake-quickdef.el#L201-L234) feels like it could be split out into a private function (e.g. `flymake-quickdef--sentinel`?) -- its inlining contributes to making the containing function quite large.
- I'm not super familiar with the flymake space, but would there be any benefit to making a PR to an established package like [flymake-easy](https://github.com/purcell/flymake-easy)?

karlotness:
Thanks for taking a look, I appreciate it. Thanks for pointing out the missed sharp quote. I've fixed that.

For the sentinel, I could spin that out into a separate macro or something. The only tricky piece is that part of the lambda body is also generated during the macro expansion. So one way or another I think something is going to have to provide a generated lambda. The macro could also potentially generate two defuns, with a separate one for the sentinel, but since it should only be used internally by the new backend, keeping a lambda might work better to keep that internal.

I think flymake-easy helps set up tools to work with the system Flymake had before it was recently updated. Those setups work differently and I think still work after the Flymake updates (with the [proc backend](https://www.gnu.org/software/emacs/manual/html_node/flymake/The-legacy-Proc-backend.html#The-legacy-Proc-backend)). It seems like it might make sense to keep this helper for the new backends separate from flymake-easy, especially since flymake-easy hasn't been updated in a while and would support older Emacs than 26+ where the new Flymake ships built-in.

riscy:
Okay, this looks great.  And, welcome to MELPA! :)

PR #6322, opened 2019-07-24
Add new recipe for helm-org.el

thierryvolpiatto:

Helm for org headlines and keywords completion

https://github.com/emacs-helm/helm-org

Maintainer

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/helm-org
+++ b/recipes/helm-org
@@ -1,3 +1,3 @@
-(helm-recoll :fetcher github
-             :repo "emacs-helm/helm-org"
-             :files ("helm-org.el"))
+(helm-org :fetcher github
+          :repo "emacs-helm/helm-org"
+          :files ("helm-org.el"))
--- a/recipes/helm-org
+++ b/recipes/helm-org
@@ -1,0 +1,3 @@
+(helm-recoll :fetcher github
+             :repo "emacs-helm/helm-org"
+             :files ("helm-org.el"))


recipes/helm-org
```diff
@@ -0,0 +1,3 @@
+(helm-org :fetcher github
+          :repo "emacs-helm/helm-org"
+          :files ("helm-org.el"))
```
riscy: This looks like you could use the [default recipe](https://github.com/melpa/melpa#recipe-format) instead of specifying a particular file.

alphapapa:
BTW, please note that `helm-org.el` is already distributed in the package `helm`; it's the same code that's been distributed in it for years.  We're splitting it into a separate package for easier, shared maintenance.

alphapapa:
I've addressed some of those minor issues in two recent commits.  The others can be ignored.  Thanks.

recipes/helm-org
thierryvolpiatto: We use this at Emacs-helm this for all our Helm external packages, don't know which code is running on Melpa, but I guess specifying a file (for simple package like this) is much faster that listing a directory and extracting the "*.el" files.
If you really want we change this, please let me know.
Thanks.

riscy:
Thanks for addressing those, @alphapapa, especially the byte-compile issues.

thierryvolpiatto:

> Merged #6322 into master.

Thanks.


-- 
Thierry

Get my Gnupg key:
gpg --keyserver pgp.mit.edu --recv-keys 59F29997

PR #6324, opened 2019-07-24
Add doom-community-themes recipe

flatwhatson:

Additional themes based on doom-themes.

https://github.com/Emiller88/emacs-doom-community-themes

Collaborator

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/doom-community-themes
+++ b/recipes/doom-community-themes
@@ -1,0 +1,1 @@
+(doom-community-themes :repo "Emiller88/emacs-doom-community-themes" :fetcher github)


purcell:
Thanks. I see there's already a big pack of Doom themes, which are also presumably Doom community themes. Is there not a way to merge these with them to avoid confusion? /cc @hlissner

ema2159:
Hi @purcell! As I understand, the intention with this is to separate the core from the themes in order to standarize Emacs theming through Doom-themes, that was Mr. Lissner's intention with `doom-community-themes`, and also for him to be able to manage only the issues and PRs  related to the theming core, delegating the specific theme issues an PRs to this new package.

purcell:
Ah, so the themes [here](https://github.com/hlissner/emacs-doom-themes/tree/master/themes) would eventually move to this new repo?

purcell:
Note that, in any case, what I'd suggest is to make each theme a separate package, which would be trivial. Bundling unrelated themes into a single package implies users will be installing more than necessary, and it brings up the issue of the package name not matching its contents.

ema2159:
Yes. Some of then will remain there (the themes maintained by Mr. Lissner himself) but the others will move to the new repo as I understand.

ema2159:
> Note that, in any case, what I'd suggest is to make each theme a separate package, which would be trivial. Bundling unrelated themes into a single package implies users will be installing more than necessary, and it brings up the issue of the package name not matching its contents.

I would beg to disagree. The value I see from these kind of packages (multi-theme packages) is that they offer the possibility for the user to have access to multiple theming options within a single package instead of having to download a lot of individual packages simplifying a lot of aspects like switching themes.

 There are several packages like this like Kaolin themes or Base16 and they are really popular for this reason, users simply like the idea of just having 2/3 theme packages with a lot of themes than having 37+ individual theme packages. 

I personally think (and many users also think this way) that this is the best way of implementing themes, so I ask you to reconsider!

purcell:
Yep, and I have my own package of the 5 Tomorrow themes. I think in those cases (including Base 16) there _may_ be more in common between the group of themes than with these ones.

I'm not particularly concerned, but what worries me is that the `doom-community-themes` package might end up containing `x-theme.el` where `x-theme.el` is also included either in the core `doom-themes` package or a standalone package: this is why there's the rule of thumb that `.el` files inside a package should start with the package name (with some exceptions for major modes).

This concern would go away completely if the themes were called `doom-community-gruvbox-theme.el` etc., and it's also why I asked above if all the Doom themes might potentially end up in a single `doom-themes` package.

ema2159:
Ok ok now I understand! Yes, let me communicate this and I'll come back with an answer

flatwhatson:
Thanks @purcell, you're right, it will be better to just have a single `doom-themes` package.  Henrik has agreed to have some collaborators there to assist with getting more themes merged.

Many thanks for your wisdom & vigilance!

purcell:
Great! 👍

PR #6326, opened 2019-07-28
Recipe: Add ts.el

alphapapa:

`ts` is a date and time library for Emacs. It aims to be more convenient than patterns like `(string-to-number (format-time-string "%Y"))` by providing easy accessors, like `(ts-year (ts-now))`.

https://github.com/alphapapa/ts.el

Author and maintainer.

### Checklist

Please confirm with `x`:

- [x] I've used the latest version of [package-lint](https://github.com/purcell/package-lint) to check for packaging issues, and addressed its feedback (*Note that this package actually exposes a bug in the current version of `package-lint`, which I'll file.*)
- [x] `M-x checkdoc` is happy with my docstrings (*Note one erroneous flag about spacing after a period which does not end a sentence.*)

### DIFF
--- a/recipes/ts
+++ b/recipes/ts
@@ -1,0 +1,1 @@
+(ts :fetcher github :repo "alphapapa/ts.el")


riscy:
Cool!

PR #6327, opened 2019-07-28
flutter-l10n-flycheck: new recipe

amake:

An optional Flycheck checker companion package for flutter.el.

https://github.com/amake/flutter.el

Maintainer

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/flutter-l10n-flycheck
+++ b/recipes/flutter-l10n-flycheck
@@ -1,0 +1,4 @@
+(flutter-l10n-flycheck
+ :repo "amake/flutter.el"
+ :fetcher github
+ :files ("flutter-l10n-flycheck.el"))


riscy:

```
Warning (emacs): 
flutter-l10n.el:163: First line is not a complete sentence
```

```
1 issue found:

102:1: error: "forward-dart-string" doesn't start with package's prefix "flutter-l10n".
```

No byte-compilation issues! 😎

No issues! 🎉

Misc. automated checks:
- flutter-l10n.el#L23: Prefer `https` over `http` (if possible)

riscy:
Wooops, sorry about that.  Wrong elisp file. :(

In the _correct_ file I only see a minor packag-elint issue:
```
1 issue found:

```

amake:
That's OK, the other file is a dependency so we might as well talk about it.

> flutter-l10n.el:163: First line is not a complete sentence

What is the best practice here? I couldn't find a good way to phrase it without going over 80 chars (to be honest, I didn't try that hard), so I wrapped it. I can squash the warning by escaping the line break, but then there's a super long line when inspecting docs with e.g. `describe-function`.

> 102:1: error: "forward-dart-string" doesn't start with package's prefix "flutter-l10n".

This is to support `bounds-of-thing-at-point` from `thingatpt`:

```
;; The function bounds-of-thing-at-point finds the beginning and end
;; positions by moving first forward to the end of the "thing", and then
;; backwards to the beginning.  By default, it uses the corresponding
;; forward-"thing" operator (eg. forward-word, forward-line).
```

The other option is to use symbol properties, but those seemed a bit magical so I went with the function.

> 1:1: warning: The package summary should start with an uppercase letter or a digit.

I've now fixed this.

riscy:
> What is the best practice here?

Just looking at your latest code, I think it would be straightforward to write:
```elisp
  "Return a hash table of existing IDs found in `flutter-l10n-file'.
Values are t."
```
One rationale is that `apropos` shows a listing that _only_ contains the first line of every docstring, so it's best if that line is somehow complete.

Thanks re: `thingatpt`, I wasn't aware of that!  I'll take a deeper look at the code soon.

amake:
Thanks for the suggestion. I've pushed fixes for all outstanding issues.

> Thanks re: thingatpt, I wasn't aware of that! I'll take a deeper look at the code soon.

I just bit the bullet and went with the symbol property solution.

riscy:
Great thanks!  Sorry about the long turnaround. :)

PR #6334, opened 2019-07-31
recipes: add kconfig-mode

elmarco:

A kernel config mode (kconfig)

https://github.com/delaanthonio/kconfig-mode

User

https://github.com/delaanthonio/kconfig-mode/issues/2

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/kconfig-mode
+++ b/recipes/kconfig-mode
@@ -1,0 +1,3 @@
+(kconfig-mode
+ :fetcher github
+ :repo "delaanthonio/kconfig-mode")


riscy:
Thanks, this definitely looks valuable.  Here's some feedback I hope is helpful:

```
kconfig-mode.el:121:1:Warning: Unused lexical variable `i'
```
This looks like a false positive though -- more on this below re: the `result` variable.

[This](https://github.com/delaanthonio/kconfig-mode/blob/fce71f2650a18eb974a5a26027d6986d90970ff5/kconfig-mode.el#L123) expression doesn't seem like it will have any effect.

The result variable [here](https://github.com/delaanthonio/kconfig-mode/blob/fce71f2650a18eb974a5a26027d6986d90970ff5/kconfig-mode.el#L126) seems unnecessary except to ensure that a value is returned by the function, but in that case why not write the function as:
```elisp
(defun kconfig-outline-level ()
  "Kconfig function to compute nesting level."
  (let ((prefix (match-string 0))
        (result 0))
    (dotimes (i (length prefix))
      (setq result (+ result
                      (if (equal (elt prefix i) ?\s)
                          1 tab-width))))
    result))
```
This also has the effect of fixing the compiler's false-positive on the `i` variable mentioned above, when lexical binding is enabled.

Sharp-quote suggestions:
- kconfig-mode.el#L150: It is beneficial to sharp-quote hooks; use `#'`

Miscellaneous and minor:
- kconfig-mode.el#L26: Prefer `https` over `http` (if possible)
- kconfig-mode.el#L5: This copyright notice may need updating to the current year.

purcell:
Thanks - the `auto-mode-alist` entry looks wrong here, though. Check using `M-: (string-match-p "pattern" "/example/full/path")` that the pattern matches the files you want to be opened in this mode. Or explain here which files you'd like to match and we can help you out with the regexp.

delaanthonio:
Upstream has been updated with the requested changes.

riscy:
> Thanks - the auto-mode-alist entry looks wrong here, though. Check using M-: (string-match-p "pattern" "/example/full/path") that the pattern matches the files you want to be opened in this mode. Or explain here which files you'd like to match and we can help you out with the regexp.

Just to chime in on this:

```elisp
(add-to-list 'auto-mode-alist '("\\Kconfig\\\'" . kconfig-mode))
```
Looks like there is an extra `\` near the end of the string (probably we just want `\\Kconfig\\'`) but ultimately it evaluates to the same string (the extra `\` just escapes the `'`).  This should match `Kconfig` files, so I believe it works correctly.

delaanthonio:
@riscy I updated the line to be `(add-to-list 'auto-mode-alist '("\\Kconfig\\'" . kconfig-mode))` as you requested

riscy:
Great, thanks for this!  Welcome, both of you, to MELPA.

PR #6337, opened 2019-08-01
Added recipe for clj-mvt-el.

davidneu:

Adds functionality to inf-clojure to support clj-mvt.

https://github.com/davidneu/clj-mvt-el

Maintainer

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/clj-mvt
+++ b/recipes/clj-mvt
@@ -1,0 +1,1 @@
+(clj-mvt :repo "davidneu/clj-mvt-modes" :fetcher github)
--- a/recipes/clj-mvt-el
+++ b/recipes/clj-mvt-el
@@ -1,0 +1,1 @@
+(clj-mvt-el :repo "davidneu/clj-mvt-el" :fetcher github)
--- a/recipes/clj-mvt-el
+++ b/recipes/clj-mvt-el
@@ -1,1 +1,0 @@
-(clj-mvt-el :repo "davidneu/clj-mvt-el" :fetcher github)


riscy:

No issues! 💯

```
No issues found.
```

```
clj-mvt-el.el:190:1:Warning: the following functions are not known to be
    defined: first, second, third
```
These can be replaced with `cl-first`, `cl-second`, etc. from `cl-lib`.

No issues! 🎉

Some sharp-quoting suggestions:
- clj-mvt-el.el#L171: Advice functions should be sharp-quoted; use `#'`
- clj-mvt-el.el#L172: Advice functions should be sharp-quoted; use `#'`
- clj-mvt-el.el#L175: Advice functions should be sharp-quoted; use `#'`
- clj-mvt-el.el#L176: Advice functions should be sharp-quoted; use `#'`
- clj-mvt-el.el#L179: Advice functions should be sharp-quoted; use `#'`
- clj-mvt-el.el#L180: Advice functions should be sharp-quoted; use `#'`
- clj-mvt-el.el#L183: Advice functions should be sharp-quoted; use `#'`
- clj-mvt-el.el#L184: Advice functions should be sharp-quoted; use `#'`
- clj-mvt-el.el#L140: It's beneficial to sharp-quote this command
- clj-mvt-el.el#L141: It's beneficial to sharp-quote this command
- clj-mvt-el.el#L142: It's beneficial to sharp-quote this command
- clj-mvt-el.el#L143: It's beneficial to sharp-quote this command
- clj-mvt-el.el#L144: It's beneficial to sharp-quote this command
- clj-mvt-el.el#L156: It's beneficial to sharp-quote this command
- clj-mvt-el.el#L157: It's beneficial to sharp-quote this command
- clj-mvt-el.el#L158: It's beneficial to sharp-quote this command
- clj-mvt-el.el#L159: It's beneficial to sharp-quote this command

Other miscellaneous checks:
- clj-mvt-el.el#L12: This statement may not currently be accurate (probably the package is not part of Emacs!)
- clj-mvt-el.el#L7: Prefer `https` over `http` (if possible)
- clj-mvt-el.el#L25: Prefer `https` over `http` (if possible)

riscy:
Since I'm here looking at the source, one issue I'm seeing is that you want to try to avoid changing Emacs's behavior simply when a package is loaded or its buffer is being evaluated -- simultaneously, it's a good idea to not advise functions using lambdas, because it makes it harder to remove that advice after the fact.

You _have_ declared a minor mode, though, which puts you in a good position to fix this and only have this advice fire when the minor mode is active.  Instead of this:
```elisp
(advice-add 'inf-clojure-load-file :before (lambda (&optional _switch-to-repl _file-name) (save-some-buffers)))
```
You could rip out that lambda and have it only fire when `clj-mvt-el-src-minor-mode` is active, e.g.
```elisp
(defun clj-mvt-el-save-some-buffers (&optional _switch-to-repl _file-name)
    (when clj-mvt-el-src-minor-mode (save-some-buffers)))
```
Incidentally, wrapping the advice-adds in `progns` is unnecessary.

I'm trying to make one of my own packages even less obtrusive in how it implements advice, which you can see [here](https://github.com/riscy/shx-for-emacs/blob/develop/shx.el#L1021-L1046).

As a last comment while I'm taking a close look, I think `clj-mvt-mode` may be a better name for the minor mode, and correspondingly `clj-mvt` is a better name for the package -- the `el` and especially the `minor-mode` tokens feel redundant.

davidneu:
Thank you for taking the time to review the package.  I think I've implemented all of your suggestions.  In keeping with the renaming, I've changed the name of the github repo to be https://github.com/davidneu/clj-mvt-modes (there are two minor modes).  Could you let me know the best way to handle that?  Thanks again!

davidneu:
... and the package has been renamed to be clj-mvt.

riscy:
You'll have to update the repository URL in your recipe.  Here's another quick pass for your reference:

```
Warning (emacs): 
clj-mvt.el:138: All variables and subroutines might as well have a documentation string
Warning (emacs): 
clj-mvt.el:141: All variables and subroutines might as well have a documentation string
Warning (emacs): 
clj-mvt.el:144: All variables and subroutines might as well have a documentation string
Warning (emacs): 
clj-mvt.el:147: All variables and subroutines might as well have a documentation string
```

```
No issues found.
```

No byte-compilation issues! 😎

No issues! 🎉

Miscellaneous and minor:
- clj-mvt.el#L12: This statement may not currently be accurate

davidneu:
I've added the doc strings and updated the recipe.  Thanks for all of your help!

purcell:
Nice!

Okay, still one key issue outstanding here: you've (reasonably) got a local minor mode, which means it can be toggled on/off in different buffers independently. However, when the minor mode is toggled, it changes global behaviour, ie. the advice. So turning the mode off in one buffer will break it in all the others. This is why @riscy suggested that the advice functions should check the mode variables. (Note that these kinds of patterns aren't terribly well documented anywhere, so don't feel bad for not knowing them.)

riscy:
@davidneu How's it working out?

riscy:

PR #6341, opened 2019-08-04
Add scihub

marsam:
Emacs integration with Sci-Hub

https://github.com/emacs-pe/scihub.el

maintainer

none needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/scihub
+++ b/recipes/scihub
@@ -1,0 +1,1 @@
+(scihub :fetcher github :repo "emacs-pe/scihub.el")


riscy:
Very nice.  This looks good to merge.

Neither of these are blockers, but in the docstring for `scihub-fetch-domains-lovescihub`, I think "working Sci-Hub working domains" should just be "working Sci-Hub domains" -- similarly for `scihub-fetch-domains-scihub_ck`.

Also, it's unlikely to become an issue, but I think you should consider adding `(require 'cl-lib)` for `cl-loop` and `cl-assert`, since it's best not to rely on transitive imports.

PR #6345, opened 2019-08-05
Add recipes/terraform-doc

TxGVNN:

Look up the Terraform documentation on the homepage

https://github.com/TxGVNN/terraform-doc

Maintainer

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/terraform-doc
+++ b/recipes/terraform-doc
@@ -1,0 +1,3 @@
+(terraform-doc
+ :repo "TxGVNN/terraform-doc"
+ :fetcher github)


DamienCassou:
The fact that you have checked "I have confirmed some of these without doing them" is kind of a bad sign.

TxGVNN:
> The fact that you have checked "I have confirmed some of these without doing them" is kind of a bad sign.

Oh, my mistake. I created 3 PR before, so this is like a habit. 
I almost tested it by myself then just tick without reading.

I will learn from experience. Thank you

riscy:
- [This](https://github.com/TxGVNN/terraform-doc/blob/f353ae26a0c453b9204ca9b94ac6a2fafd862616/terraform-doc.el#L105) seems unusual, but maybe I'm missing something.  How come we need the `with-current-buffer` wrapper?
- If possible, you should prefer to write a loop like the following instead of using [`perform-replace`](https://github.com/TxGVNN/terraform-doc/blob/f353ae26a0c453b9204ca9b94ac6a2fafd862616/terraform-doc.el#L114) -- note that `perform-replace` echoes messages too, which isn't what you want:
  ```elisp
  (while (re-search-forward "foo[ \t]+bar" nil t)
    (replace-match "foobar" nil nil))
  ```
- [This](https://github.com/TxGVNN/terraform-doc/blob/f353ae26a0c453b9204ca9b94ac6a2fafd862616/terraform-doc.el#L106-L118) chunk and [this](https://github.com/TxGVNN/terraform-doc/blob/f353ae26a0c453b9204ca9b94ac6a2fafd862616/terraform-doc.el#L135-L146) chunk feel like they have quite a lot of repeated code.  Also, both feel like they should be pulled into their own function(s).  It will make the code easier to read, and since these contain a lot of (fragile) parsing logic, you want them to be easier to edit and update in isolation from other functions.
- None of the checkers complain about it due to transitive dependencies, but I would personally be leery about relying on e.g. `shr-render-region` being available without adding an explicit require on `shr`.

TxGVNN:
Hi @riscy,

Thanks for your great comment,
I have updated the 4 points which you mentioned. Could you have a look?

purcell:
Looks good now, merging - thanks!

PR #6348, opened 2019-08-06
Add navigel

DamienCassou:

Facilitate the creation of tabulated-list based UIs

https://gitlab.petton.fr/DamienCassou/navigel/

I'm the author

none required

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/navigel
+++ b/recipes/navigel
@@ -1,0 +1,1 @@
+(navigel :fetcher github :repo "DamienCassou/navigel")


Silex:
Can you describe the differences with https://github.com/politza/tablist a little bit?

DamienCassou:
> Can you describe the differences with https://github.com/politza/tablist a little bit?

@Silex: Thank you for the suggestion. I just [improved the README](https://gitlab.petton.fr/DamienCassou/navigel/commit/b25d96d1268e72cce7750fab22afbdabe0bb8fc8) a bit to clarify the difference. I copy/paste it here:

> This package depends on [[https://github.com/politza/tablist][tablist]] to get support for marking, deletion and more.  On top of the tablist package, navigel provides an easy way to specify the content of your tabulated lists: through entities specified with method overrides.  This makes it a breath to have tablist-based navigation within domain objects.

Silex:
:+1: 

I already use `tablist` in docker.el, I might consider navigel :wink:

DamienCassou:
Philippe Vaucher <notifications@github.com> writes:
> I already use `tablist` in docker.el, I might consider navigel :wink:

please don't hesitate to give feedback (being missing doc, missing
features, broken stuff…). It's in early stage of development even though
it is inspired by former packages I've created.

-- 
Damien Cassou

tarsius:
Wouldn't it be better if `tablist` and `navigel` were merged into `tabulated-list` itself?

DamienCassou:
> Wouldn't it be better if `tablist` and `navigel` were merged into
> `tabulated-list` itself?

regarding tablist, I think it could be a good idea. I'm not sure for
navigel though because it heavily relies on CL methods for the public
API which is not really how Emacs code is usually structured. I think it
would be a bit alien for Emacs core.

-- 
Damien Cassou

purcell:
Very nice, thanks - merged.

tarsius:
@DamienCassou 

The recipe actually references the github repository not the repository from your gitea instance that you mentioned above. Is the former intended as a mirror of the later? The later is ahead of the former.

Melpa gets 19 of your packages from github while it gets `json-process-client` from https://gitea.petton.fr. What's up with that?

DamienCassou:
> The recipe actually references the github repository not the
> repository from your gitea instance that you mentioned above. Is the
> former intended as a mirror of the later?


The github repository is supposed to mirror the one on gitea. The goal
of referencing github from melpa instead of gitea is to reduce the
likelihood that the repository is down when melpa is building. Moreover,
when Nicolas Petton switched from GitLab to Gitea recently, I lost the
automatic synchronization of my repositories. I have to pay more
attention when I commit on gitea to also propagate the changes to Github.


> The later is ahead of the former.


This is fixed now.


> Melpa gets 19 of your packages from github while it gets
> `json-process-client` from https://gitea.petton.fr. What's up with
> that?


I should have pushed json-process-client to github, just like the
others. Sorry about that.

-- 
Damien Cassou

tarsius:
Well, do you think it is more likely that petton.fr goes down than it is for you to forget to push to the mirror? :stuck_out_tongue_winking_eye:

But you don't actually have to remember to do so yourself anyway. Use this instead:

```
git remote set-url --push origin https://gitea.petton.fr/DamienCassou/navigel.git
git remote set-url --add --push origin git@github.com:DamienCassou/navigel.git
```

The first command is required even though this just duplicates the value of the fetch url.

DamienCassou:
> Well, do you think it is more likely that petton.fr goes down than it
> is for you to forget to push to the mirror?
> :stuck_out_tongue_winking_eye:


You have a point :-).


> But you don't actually have to remember to do so yourself anyway. Use
> this instead: […]


This is actually a very nice tip, thank you so much. I'm not sure it
will solve all my problems (e.g., when I merge a PR in gitea, master
will change there but not in github) but it will certainly help me quite
a bit. It's too bad I can't specify this in my ~/.emacs.d/.gitmodules
file.

Best,


-- 
Damien Cassou

PR #6352, opened 2019-08-07
Add recipe for dart-server

bradyt:

Factors out the features from dart-mode.el that relied on external executables, thus improving availability, reliability, testing and development of dart-mode.el.

~~Once on melpa, I can push the slimmed down dart-mode.el to master.~~

https://github.com/bradyt/dart-server

Commits splitting project into two roughly disjoint projects: 
- https://github.com/bradyt/dart-mode/commit/c5681185359c7db54756d7935de421da8eef37c7
- https://github.com/bradyt/dart-server/commit/a9f2034f29ad106b14947c10950da6671687e73a

I am the maintainer of the original combined package, and am nearly the sole author of what is left now in [current working tree of dart-mode.el](https://github.com/bradyt/dart-mode).

**None needed**

### Checklist

Please confirm with `x`:

- [ ] ~~`M-x checkdoc` is happy with my docstrings~~

### DIFF
--- a/recipes/dart-server
+++ b/recipes/dart-server
@@ -1,0 +1,2 @@
+(dart-server :fetcher github :repo "bradyt/dart-server"
+             :version-regexp "dart-server-\\(.*\\)")


duianto:
The link text: `current working tree of dart-mode.el`
with the url: https://github.com/bradyt/dart-mode/tree/scratch/remove-server
leads to a 404 page.

bradyt:
@duianto Thank you, I've updated the link.

riscy:
Bearing in mind this is (part of) an established package, I'm seeing a couple issues as part of the standard review process.  It feels likely we'll grandfather this in (I'll defer to more experienced reviewers), but you are probably aware of package-lint's issue with the namespace and the following warning:
```
300:55: warning: This key sequence is reserved (see Key Binding Conventions in the Emacs Lisp manual)
```

And while I'm looking, `dart-popup-mode-map` could possibly benefit from being encapsulated in a single `defvar`?:
```elisp
(defvar dart-popup-mode-map
    (set-keymap-parent dart-popup-mode-map help-mode-map)
    ;; Unbind help-specific keys.
    (define-key dart-popup-mode-map (kbd "RET") nil))
    ...
```

These sharp-quote suggestions that may be helpful:
- dart-server.el#L403: It is beneficial to sharp-quote hooks; use `#'`
- dart-server.el#L404: It is beneficial to sharp-quote hooks; use `#'`
- dart-server.el#L405: It is beneficial to sharp-quote hooks; use `#'`
- dart-server.el#L1231: It is beneficial to sharp-quote hooks; use `#'`

Miscellaneous and minor:
- dart-server.el#L1244: It is unusual to derive from `fundamental-mode`; perhaps `special-mode` is more appropriate?
- dart-server.el#L11: Copyright notice may be out of date.
- dart-server.el#L99: Use `unless ...` instead of `when (not ...)`
- dart-server.el#L24: Prefer `https` over `http` (if possible)

bradyt:
Package-lint is not compatible with straight.el. Nor is it compatible with Cask. Can we ignore the package-lint issues until it is more available?

bradyt:
- https://github.com/purcell/package-lint/issues/73
- https://github.com/purcell/package-lint/issues/108

bradyt:
@riscy I don't get all these errors. It might be because package-lint is apparently not compatible with straight.el or Cask. Can you clarify how you are installing dash, flycheck, s. Are you only using package-lint? Can you clarify all the tools you are using, and exactly how you call them, and how your method of installation?

Thank you!

bradyt:
> Bearing in mind this is (part of) an established package, I'm seeing a couple issues as part of the standard review process. It feels likely we'll grandfather this in (I'll defer to more experienced reviewers),

Yes, please grandfather it in, I haven't made time to understand this part of the package (now the present entirety). All the issues you describe are from previous authors and maintainers. I'd like these commands to be easily available to anyone that was still using them.

> but you are probably aware of package-lint's issue with the namespace and the following warning:
>
> ```
> 300:55: warning: This key sequence is reserved (see Key Binding Conventions in the Emacs Lisp manual)
> ```

Both done.

(My package-lint script was silently broken, though indeed I knew of these issues from older drafts of testing setups.)

> And while I'm looking, `dart-popup-mode-map` could possibly benefit from being encapsulated in a single `defvar`?:
> 
> ```emacs-lisp
> (defvar dart-popup-mode-map
>   (let ((keymap (make-sparse-keymap)))
>     (set-keymap-parent dart-popup-mode-map help-mode-map)
>     ;; Unbind help-specific keys.
>     (define-key dart-popup-mode-map (kbd "RET") nil))
>     ...
> ```

I'm not sure I understand. May I please have precise information on the benefit. I'd wonder if the `dart-popup-mode-map` would make sense inside the `defvar`. Can this be filed as an issue at repo and be dealt with later? Or if it's fine with you, I already have a general issue at the repo pointing to this thread.

> These sharp-quote suggestions that may be helpful:
> 
> * dart-server.el#L403: It is beneficial to sharp-quote hooks; use `#'`
> * dart-server.el#L404: It is beneficial to sharp-quote hooks; use `#'`
> * dart-server.el#L405: It is beneficial to sharp-quote hooks; use `#'`
> * dart-server.el#L1231: It is beneficial to sharp-quote hooks; use `#'`

All four done.

> Miscellaneous and minor:
> 
> * dart-server.el#L1244: It is unusual to derive from `fundamental-mode`; perhaps `special-mode` is more appropriate?

I would like to research this at a later date. If it is fine with you, I have a general issue on the repo pointing to this thread.

> * dart-server.el#L11: Copyright notice may be out of date.

I don't understand. The whole repo was transferred to me in 2018, with the copyright dates as they are now. And the half that is now dart-server.el, I've modified very little. Does this require contacting google (previous maintainer?) and go authors for clarification?

> * dart-server.el#L99: Use `unless ...` instead of `when (not ...)`
> * dart-server.el#L24: Prefer `https` over `http` (if possible)

Both done.

riscy:
> May I please have precise information on the benefit. I'd wonder if the dart-popup-mode-map would make sense inside the defvar.

I should have replaced `dart-popup-mode-map` with `keymap` in that case.  More like this:
```elisp
(defvar dart-popup-mode-map
    (set-keymap-parent keymap help-mode-map) ...
```

The main benefit is if I were to define my own `dart-popup-mode-map` in my config, and then load `dart-server`, my keybindings wouldn't be clobbered (if I did it in the reverse order I'd be fine... as long as I didn't inadvertently reload `dart-server`). The same goes for `dart-server-map`.

> Can this be filed as an issue at repo and be dealt with later? 

I'd anticipate it would be a very quick fix and would only improve the package (even though I made a copy paste mistake).  Someone might even make a PR for it if you're pinned.

> [re: special mode] I would like to research this at a later date

That's fine.

> [re: copyright] I don't understand

Copyright notice may also _not_ be out of date. :)  That's an automatic check I wrote since sometimes these are neglected or copy-pasted incorrectly.

bradyt:
Okay, keymap code has been moved to defvar for both.

purcell:
Thanks @bradyt, this all broadly makes sense. I'm okay with merging this as-is, given its history, but I have a few notes along the way for you to take into account as you move forward with this code:

- Consider splitting out `dart-flycheck` as a separate package. This would make the packaging more coherent, and pave the way for `flymake` support etc. later. Doing this will generally help the modularity of the code.
- `pcase` is nice in cases like [this](https://github.com/bradyt/dart-server/blob/c1cc631fb1649ca19942c94741e38cc14c4c7e0f/dart-server.el#L681-L686)
- [Here](https://github.com/bradyt/dart-server/blob/c1cc631fb1649ca19942c94741e38cc14c4c7e0f/dart-server.el#L261-L265) you've got a `when-let` containing a clause that should also be - but isn't - a `when-let`.
- You can canonicalise paths using `file-truename`, e.g. [here](https://github.com/bradyt/dart-server/blob/c1cc631fb1649ca19942c94741e38cc14c4c7e0f/dart-server.el#L392-L395)
- If possible, make `dart-server-expand` work as a `completion-at-point` function. (See `completion-at-point-functions`.) Binding `M-/` to an ad-hoc completion mechanism really isn't ideal!

Hope that helps.

bradyt:
Thank you, great suggestions, notes copied to issue at https://github.com/bradyt/dart-server/issues/7.

PR #6353, opened 2019-08-08
Add compdef recipe.

jjzmajic:

Sets local completion backends for capf and company with some convenience auto-magic thrown in.

https://gitlab.com/jjzmajic/compdef

Maintiner.

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/compdef
+++ b/recipes/compdef
@@ -1,0 +1,1 @@
+(compdef :fetcher gitlab :repo "jjzmajic/compdef")


jjzmajic:
For the record, I know it's simplistic. It just feels like a lot of us do a whole bunch of copy pasting to set this up for every major mode. If it doesn't contribute enough to be on MELPA I get it.

riscy:

No issues! 💯

```
No issues found.
```

```
compdef.el:70:40:Warning: assignment to free variable `company-backends'
```
Could 

No issues! 🎉

Sharp-quote suggestions:
- compdef.el#L63: Alias targets should be sharp-quoted; use `#'`

Miscellaneous and minor:
- compdef.el#L66: With emacs 24.3 you can use `setq-local`
- compdef.el#L69: With emacs 24.3 you can use `setq-local`
- compdef.el#L24: Prefer `https` over `http` (if possible)

jjzmajic:
Better? Not sure how to use `#'` on a fynamically generated symbol though? Is there a function that does this?

Jumping on hercules.el now!

riscy:
> Not sure how to use #' on ...

That was my oversight.  I'll start a deeper look when time allows!

purcell:
Thanks for this. I wonder if this package promotes good practice, though. For example, it's actually standard practice to write `company` backends so that they are no-ops when they are not applicable, e.g. outside certain major modes, so then it's normal to add company backends to the global `company-backends` value. Additionally, this package really doesn't save much code compared to the regular boilerplate, particularly since both `modes` and `hooks` must be passed. For that particular point, see the built-in function `derived-mode-hook-name`, but I still suspect that this package is a case of premature factorisation.

jjzmajic:
That's fair enough, like I said, I'm leaving this one up to you. The only reason I made this PR is because it seemed to be quite popular on r/Emacs.

Just want to clarify a couple of things about functionality before you make the call though:
- Either modes or hooks or both can be passed (which saves a bit of code). The idea is that most times mode hooks follow a convention, and the hooks parameter is only to be used when they do not.
- The code saving bit comes when you're dealing with multiple modes or a mode with a repl (e.g. I want the same backends in emacs-lisp-mode and ielm). You can just pass a list of modes and hooks.
- The bad practice is kind of already there. Spacemacs, Doom and Centaur Emacs all separately implement this functionality because of it. The example I can think of off the top of my head where this was handy is here https://github.com/syl20bnr/spacemacs/pull/11933. Note that I had to remove company-capf completely because it just wouldn't work together with company-nimsuggest. Some completion backends place snippets to this effect in the README precisely because it's a frequent problem.
- Speed. Once you have 30 function calls before you hit company-dabbrev which is what you really might want, you feel that 0.00001s lag every time you press TAB

Again not pushing for this to be merged. Just wanted the pros to be outlined as well as the cons before a decision is made. Happy to close this if you still fill it doesn't cut it

jjzmajic:
PS: added the fix you suggested.

jjzmajic:
Accidentally closed... Just wanted to drop some more examples straight from lsp-mode:
``` elisp
    (when (and lsp-enable-completion-at-point (lsp--capability "completionProvider"))
      (setq-local completion-at-point-functions nil)
      (add-hook 'completion-at-point-functions #'lsp-completion-at-point nil t))
```

```elsip
      (add-to-list 'company-backends 'company-lsp)

      ;; make sure that company-capf is disabled since it is not indented to be
      ;; used in combination with lsp-mode (see #884)
      (setq-local company-backends (remove 'company-capf company-backends))
```

And there are countless READMEs recommending stuff like this: https://github.com/nsf/gocode/tree/master/emacs-company

40-odd backends... one is bound to cause trouble

riscy:
Hi jizzmajic, I think you make a reasonable argument for inclusion in MELPA -- but I think the checklist items have slipped with the time that has passed.  I realize I nag a lot, but sometimes fixing these addresses other problems before a closer look at your code.

```
9 issues found:

85:21: warning: `with-eval-after-load' is for use in configurations, and should rarely be used in packages.
96:2: error: Aliases should start with the package's prefix "compdef".
97:2: error: Aliases should start with the package's prefix "compdef".
98:2: error: Aliases should start with the package's prefix "compdef".
115:1: error: "use-package-handler/:capf" doesn't start with package's prefix "compdef".
115:1: error: `use-package-handler/:capf' contains a non-standard separator `/', use hyphens instead (see Elisp Coding Conventions).
129:2: error: Aliases should start with the package's prefix "compdef".
133:1: error: "use-package-handler/:compdef" doesn't start with package's prefix "compdef".
133:1: error: `use-package-handler/:compdef' contains a non-standard separator `/', use hyphens instead (see Elisp Coding Conventions).
```

```
compdef.el:138:1:Warning: the following functions are not known to be defined:
    use-package-list-insert, use-package-normalize-symlist,
    compdef--plist-multi-delete, use-package-plist-delete,
    compdef--use-package-get, use-package-handler/:capf
```

jjzmajic:
@riscy You aren't nagging, just maintaining healthy standards. Sorry for putting you in a position where you have to do it in the first place. Is the code better now?

NOTE: I removed the `defalias` call as using lambdas allows for much simpler `use-package` keyword handling. Besides, I don't want to completely nuke the namespace with unnecessary functions.

riscy:
Hi @jjzmajic, it looks like there are still byte-compiler warnings:

```
compdef.el:115:1:Warning: the following functions are not known to be defined:
    use-package-list-insert, compdef--use-package-handler
```

What I recommend is you create a block of byte-compiler-pacifier at the top of your code (not inside the eval-after-load) like:

```elisp
;; Compiler pacifier
(defvar use-package-keywords)
(declare-function use-package-concat "ext:use-package-core")
(declare-function use-package-list-insert "ext:use-package-core")
(declare-function use-package-process-keywords "ext:use-package-core")
(declare-function use-package-normalize-symlist "ext:use-package-core")
```

And move the `defun compdef--use-package-handler` out of the eval-after-load as well, so that the byte-compiler can know that the function exists.  When you do that, you will get checkdoc warnings on the documentation for that function -- maybe something like this suffices for that docstring:

```
  "Handle each `compdef' `use-package' KEYWORD for package NAME.
Note this function should not be called with KEYWORD :compdef."
```

Let me know if that sounds about right to you.

jjzmajic:
Oh dammit. I switched to doom recently, and there's all kinds of magic I don't understand. When I byte compile something like this in my config the warnings pop up. I wanted to keep the functions declarations inside the `with-eval-after-load` to avoid namespace pollution, but I guess you guys know best given how much code you see. 

Just a question since I obviously missed something basic here:
I thought the whole point of `with-eval-after-load` (as compared to `eval-after-load`) was for the compiler to see inside and not treat the sexps like data. But you're right, when I compile it in my vanilla config, the warnings show up. What am I missing?

jjzmajic:
PS: I made the changes you suggested.

riscy:
Looks like there's still a byte-compile warning in case `use-package` hasn't been brought in -- you still need this one for a clean byte-compile:

```elisp
(declare-function use-package-list-insert "ext:use-package-core")
```

riscy:
Also, I think you should revise the short description of your package.  "A stupid completion definer." is not something a person scrolling/searching the list of packages is going to be able to make complete sense of.

jjzmajic:
Nothing I can do but apologize. I don't now what happened to my settings. I manually checked the compile log this time since I don't trust it to pop up.

riscy:
No need to apologize!  I byte-compile PR's with a featureless Emacs install that brings in a minimal number of packages, including packages like `use-package`.  I suspect most people aren't byte-compiling in a clean room like that (so I really wouldn't blame your settings).  It's probably a little bit overkill on my part.

riscy:
Here we go.  Thanks!

jjzmajic:
Yay!

purcell:
>  byte-compile PR's with a featureless Emacs install that brings in a minimal number of packages

FWIW, I think when `flycheck` runs its emacs lisp checker, it's byte-compiling in a separate emacs process, so it should pick up some of the same issues.

jjzmajic:
Thanks @purcell!

PR #6354, opened 2019-08-08
Add vcsh

stepnem:
(comment edited by @riscy) 

Brief summary of what the package does

The package provides Emacs/Magit integration with the vcsh CLI tool.

Direct link to the package repository

vcsh: https://github.com/RichiH/vcsh

This package (vcsh.el): https://gitlab.com/stepnem/vcsh-el

Your association with the package

I am the author.

Relevant communications with the upstream package maintainer

None needed

Checklist

Please confirm with x:

- [x] The package is released under a GPL-Compatible Free Software License.
- [x] I've read CONTRIBUTING.org
- [x] I've used the latest version of package-lint to check for packaging issues, and addressed its feedback
- [x] M-x checkdoc is happy with my docstrings
- [x] I've built and installed the package using the instructions in CONTRIBUTING.org

### DIFF
--- a/recipes/vcsh
+++ b/recipes/vcsh
@@ -1,0 +1,1 @@
+(vcsh :fetcher gitlab :repo "stepnem/vcsh-el")


alphapapa:
Why did you delete the checklist template?

riscy:
In the meantime :) 

```
Warning (emacs): 
vcsh.el:118: `y-or-n-p' argument should end with "? "
```

```
2 issues found:

103:3: error: You should depend on (emacs "25") if you need `when-let'.
```

No byte-compilation issues! 😎

No issues! 🎉

stepnem:
> vcsh.el:118: `y-or-n-p' argument should end with "? "

This is a checkdoc bug; it only naïvely checks for a single string
argument, but in this case the string is computed (using `format') and
does end with "? ".

> 1:1: warning: The package summary should start with an uppercase letter or a digit.

I'm aware of this (BTW I'm sorry I didn't check the PR template boxes, I
bypassed the web interface so I didn't see it when submitting) and I'd
like to suggest you remove this check from package-lint. As far as I
know such a requirement isn't documented anywhere, checkdoc doesn't
require it, many libraries out there (including Emacs core) don't follow
it and often it doesn't make much sense, such as here where "vcsh" is
the name of the tool (e.g., all occurences in the manpage start
sentences beginning with it with lowercase letter).

> 103:3: error: You should depend on (emacs "25") if you need `when-let'.

Oops, sorry for that, I added `when-let' only yesterday and forgot to
rerun package-lint... Fixed!

Thanks.

riscy:
Noted on the checkdoc bug -- thanks for checking.

> I bypassed the web interface

I updated the comment -- unless for purposes of accessibility (let me know) or otherwise, would you mind double-checking the last couple boxes?

stepnem:
> I updated the comment -- unless for purposes of accessibility (let me
> know) or otherwise, would you mind double-checking the last couple
> boxes?

Done, thanks and sorry again for the extra work.

purcell:
It looks like this package could be useful without Magit, no? In that case, it should be 2 little packages, with `magit-vcsh` being an optional add-on for `magit` users which would depend on `magit` and `vcsh`. As it stands, some users would get `magit` unexpectedly installed without needing it. Alternative is to just call this whole package `magit-vcsh`. The code itself looks good, btw. :-)

stepnem:
> It looks like this package could be useful without Magit, no?

Yeah, I guess you're right. Without some kind of GUI front end I don't
see the point of using M-x vcsh-link instead of "vcsh enter" or M-x
vcsh-new instead of "vcsh init" + "git add", but it should be possible
to use this e.g. with vc instead of Magit.

I'll split the package in two as you suggest.

Thanks.

stepnem:
> I'll split the package in two as you suggest.

Done as #6380

purcell:
Super, thanks!

PR #6355, opened 2019-08-09
Recipe: Add org-rifle, change helm-org-rifle

alphapapa:
`helm-org-rifle` is being split into separate packages, `org-rifle` and `helm-org-rifle`.  Both are built from the same repo.  

I have a branch about ready to push to `master` at https://github.com/alphapapa/org-rifle/tree/wip/org-rifle.

What's the best way to coordinate the transition?  If the new `helm-org-rifle` recipe is merged to MELPA before I push the new branch to `master`, or if I push to `master` before the recipe is merged, users who try to upgrade it will get an error because the `org-rifle` package won't yet be buildable and won't be available for installation, which the new `helm-org-rifle` package depends on.

Thanks.

### DIFF
--- a/recipes/helm-org-rifle
+++ b/recipes/helm-org-rifle
@@ -1,1 +1,2 @@
-(helm-org-rifle :fetcher github :repo "alphapapa/helm-org-rifle")
+(helm-org-rifle :fetcher github :repo "alphapapa/org-rifle"
+                :files ("helm-org-rifle.el"))
--- a/recipes/org-rifle
+++ b/recipes/org-rifle
@@ -1,0 +1,2 @@
+(org-rifle :fetcher github :repo "alphapapa/org-rifle"
+           :files ("org-rifle.el"))


akirak:
There was a similar situation in the past: https://github.com/melpa/melpa/pull/5571

In the case of `lv` now separated from `hydra`, the recipe for `lv` was first added. At this point, the recipe for `hydra` hadn't been changed. `lv.el` was duplicated in both `hydra` and `lv` packages, but it didn't cause build failures on `hydra` package. Afterwards, the recipe for `hydra` was changed to exclude `lv.el`, which now looks as follows: https://github.com/melpa/melpa/blob/master/recipes/hydra

According to the example, you can prevent the build failure by first adding the recipe for `org-rifle`, while `helm-org-rifle` continues to have no `:files` spec. The user can choose either `org-rifle` for one library or `helm-org-rifle` for both. Later on, you can send a new PR for modifying the recipe for `helm-org-rifle`. 

You probably didn't have to rename the repository from `helm-org-rifle` to `org-rifle`, but it doesn't matter, since the old repository URL is redirected to the new one. `helm-org-rifle` is the only helm-based package I regularly use (I prefer ivy in most situations), but it would be useful in several situations to have a separate `org-rifle` package.

purcell:
Thanks @akirak, yep, that makes sense. @alphapapa, let me know what you'd like to do.

alphapapa:
Thanks to both of you.  So if I understand correctly, I should take these steps:

1.  Push new branch that splits the code into `org-rifle.el` and `helm-org-rifle.el` to `master`.  The `helm-org-rifle` package will then include both files, and `helm-org-rifle.el` will be able to `(require 'org-rifle)`.
2.  Add recipe for package `org-rifle` to MELPA that excludes file `helm-org-rifle.el`.  Package `org-rifle` will then be an installable package that only includes files `org-rifle.el`, and package `helm-org-rifle` will be an installable package that includes files `org-rifle.el` and `helm-org-rifle.el`.
3.  Add to file `helm-org-rifle.el` a `Package-Requires` dependency on package `org-rifle`.  Since package `org-rifle` will by then be an installable package, this will cause package `org-rifle` to be installed when package `helm-org-rifle` is installed.  The file `org-rifle.el` will be installed by both packages in their own directories, but that should not cause a problem, because, assuming both packages are up-to-date, the contents of the file will be the same in both packages.
4.  Update recipe for package `helm-org-rifle` to exclude file `org-rifle.el`, and update recipe for package `org-rifle` to exclude file `helm-org-rifle.el`.  The package `helm-org-rifle` will then be an installable package that includes only the file `helm-org-rifle.el`, and the package `org-rifle` will then be an installable package that includes only the file `org-rifle.el`.  When package `helm-org-rifle` is installed or updated, it will install package `org-rifle` if it's not already installed.

So, following those steps will prevent any gap in which the package `helm-org-rifle` would not be able to install its dependency package `org-rifle`, and users of `helm-org-rifle` who update it during the transition should not experience any errors.

A verbose explanation, to be sure.  :)  Am I understanding correctly?

Thanks for your help.

akirak:
@alphapapa It must be what I intended. Thank you for the detailed explanation.

> 3. Add to file helm-org-rifle.el a Package-Requires dependency on package org-rifle. 

If you follow the plan, please don't forget you first have to remove `org-rifle` from the `Package-Requires` header.

riscy:
Just pinging the participants on this one.  I must have mistakenly removed the awaiting-upstream label.  Thanks!

alphapapa:
@riscy Thanks.  I'm not sure when I'll have time to work on this plan, so I'll close this PR for now.

PR #6358, opened 2019-08-10
Add fast-scroll to MELPA

ahungry:

Disables certain Emacs features when it detects the user is scrolling quickly.

https://github.com/ahungry/fast-scroll

Owner

It's myself

### Checklist

Please confirm with `x`:

- [x ] I've built and installed the package using the instructions in [CONTRIBUTING.org](https://github.com/melpa/melpa/blob/master/CONTRIBUTING.org)
- [ haha, nice last item] I have confirmed some of these without doing them

### DIFF
--- a/recipes/fast-scroll
+++ b/recipes/fast-scroll
@@ -1,0 +1,3 @@
+(fast-scroll
+  :repo "ahungry/fast-scroll"
+  :fetcher github)


riscy:

```
Warning (emacs): 
fast-scroll.el:108: All variables and subroutines might as well have a documentation string
Warning (emacs): 
fast-scroll.el:109: All variables and subroutines might as well have a documentation string
Warning (emacs): 
fast-scroll.el:110: All variables and subroutines might as well have a documentation string
Warning (emacs): 
fast-scroll.el:111: All variables and subroutines might as well have a documentation string
```

```
1 issue found:

9:1: warning: You should include standard keywords: see the variable `finder-known-keywords'.
```

```
fast-scroll.el:131:1:Warning: the following functions are not known to be
    defined: evil-scroll-up, evil-scroll-down
```

No issues! 🎉

Miscellaneous and minor:
- fast-scroll.el#L27: Prefer `https` over `http` (if possible)

ahungry:
Yes, for one line function alias bindings, I do not believe it's worthwhile to include a docstring echoing what is basically a wrapper to advice different functions, so I chose not to as that's just a checkdoc warning.  If this causes rejection in Melpa, I feel that's to the communities detriment.

I don't want to force inclusion of evil mode as a dependency, but I do want evil-mode users to receive the convenient advice setups.

riscy:
Sorry, but especially for a small package like this, please do the checklist.

ahungry:
Sure, updated on master in my repo, I re-ran all your linters and no complaints or warnings :)

purcell:
Thanks for this - interesting idea. A few notes:

- You've got ad-hoc functions to enable and disable the advice, and code which calls `global-set-key`. What this suggests, and what would really be better for end users, is a conventional global minor mode which users can enable.
- Disabling and enabling flycheck is not advisable for a few reasons, not the least of which is that a generalised package like this really shouldn't have any hard-coded references to other specific packages it doesn't depend on. Why not provide two hooks that users could use to do that sort of thing if they really decide they want it? You could even add your own functions for minimising/restoring the modeline and disabling font lock to those hooks by default. All of this would make the package more "emacsy" and generally usable.

ahungry:
For the first question - the reason I opted to go with advice (and an optional top level binding) vs a mode is that this feature is intended to be globally useful to handle a global problem (Emacs has terrible scroll latency in any significantly large buffer when font lock and other modes are enabled) - as such, if a user wants a global "fix" to their scrolling speed, this package is for them.  A user could certainly conditionally bind the functions that I "wrap" for them without the advice if they only wanted to use this mode in a very limited subset of Emacs modes.

For the second one - I thought that a functionp guard to check for the existence or not (of flycheck) would be suitable, but I suppose that could run into a possible issue where the user has flycheck, yet has not enabled it (where this package would tend to leave it enabled as a side-effect).

A hook is a good idea (a list of modes that should be flipped off when scrolling and set back to on when done).  I'll work on adding that.

For a minor-mode, is there mode-specific advice?  So I could wrap common scroll keys in the defadvice way, but limited to just the mode?  (I guess I could always opt into wrapping as I do currently, but abort early as a no-op function call if the "fast-scroll" minor mode was not active in the given buffer).

purcell:
> , if a user wants a global "fix" to their scrolling speed, this package is for them.

A global minor mode would seem ideal then.

> but I suppose that could run into a possible issue where the user has flycheck, yet has not enabled it (where this package would tend to leave it enabled as a side-effect).

Yes, exactly.

> For a minor-mode, is there mode-specific advice? So I could wrap common scroll keys in the defadvice way, but limited to just the mode? (I guess I could always opt into wrapping as I do currently, but abort early as a no-op function call if the "fast-scroll" minor mode was not active in the given buffer).

Yeah, the typical way to do this is to check the mode var in the advice function, or - even better for a _global_ minor mode - enable/disable the advice in the mode body, according to whether the mode is being enabled or disabled.

ahungry:
The package removed the flycheck reference and has hooks support now.

For the advice - are you suggesting repeatedly adding/removing advice around the functions for scrolling each time the mode is enabled/disabled?  Would that run into a problem if a user wanted them advised in one buffer, such as long_file.c, and didn't want them enabled in short_file.el, but then upon switching back to long_file.c, would they not still be disabled?  Or can advise be buffer specific?

ahungry:
Latest master has global minor mode support, great call on that, thank you.

riscy:
I'm just doing some user testing.  One thing I notice is that if I scroll down fast enough for the minor-mode to activate itself, and then quickly do e.g. `other-window`, my mode-line gets "stuck" on the fast-scroll version.  I discovered this in the natural course of editing.  It seems like we also want to identify the buffer in which `fast-scroll-run-fn-minimally` was called, and to have `fast-scroll-end` take a buffer as an argument.

I realize how painstaking it must have been to get a lot of this to line up with all of Emacs' idiosyncrasies in play, but something about `fast-scroll-end` needing to call `fast-scroll-end-p` to check timing information feels like `run-with-idle-timer` may be a better choice for resetting the minor mode state.  Thoughts?

Ultra minor:
- I think you could just call the minor mode `fast-scroll-mode` rather than `fast-scroll-minor-mode` -- the latter is atypical of minor modes (e.g. we don't call it `evil-minor-mode`)
- `fast-scroll-default-mode-line` feels like it should be customizable?
- In the future (definitely not a blocker) maybe it's worth supporting `mwheel-scroll` too?

ahungry:
@riscy sorry for the delay on that.  Please see tag 0.0.5 or later. I have addressed the following:

- Your timing issue should no longer be possible
- The mode has removed the extra -minor- label that was present

Regarding the run-with-idle-timer suggestion - it seems like an interesting approach, although I'm uncertain if it would have any change in the end result (currently with a think w530, I can scroll through a 50,000 line C file with flycheck + complicated mode line turned on without getting a single Emacs pause/hangup).  I think the setup I use is quite modest, and not that "beefy" compared to an average one, but I will probably tinker with the idea/implementation if I find free time (or be open to a PR if it can be shown to make a difference).

ahungry:
The one problem I can think of off the top of my head is that run-with-idle-timer would probably only run if the entire Emacs was idle, which is a regression in current behavior:

Currently:

- I can scroll quickly, triggering the mode
- I can stop scrolling and immediately begin navigating in the buffer with letter/word/line movements
- The buffer will un-minimize and show proper syntax / modeline etc., as I would like

With the proposition:

- Same scenario would leave the buffer minimized until I actively stopped pressing any keys for a period of time (I believe).

riscy:
Yeah, that makes sense to me.  This looks great -- thanks for addressing everything that came up in the review, and apologies for the longer turnaround time.

ahungry:
Awesome! Glad to see it merged :)

PR #6362, opened 2019-08-10
Add kaocha-runner

magnars:

The package runs Clojure tests with the Kaocha library via CIDER (the Clojure IDE for Emacs).

https://github.com/magnars/kaocha-runner.el

Maintainer.

### Checklist

Please confirm with `x`:

Good checklist! The README and this are both very useful to improve quality. Thanks.

### DIFF
--- a/recipes/kaocha-runner
+++ b/recipes/kaocha-runner
@@ -1,0 +1,1 @@
+(kaocha-runner :fetcher github :repo "magnars/kaocha-runner.el")


riscy:

No issues! 💯

```
1 issue found:

38:0: warning: Private functions generally should not be autoloaded.
```
Although this isn't a function, it is still a customizable variable.  In general, defcustoms shouldn't be private (as indicated by the double-hyphen) since by definition they should be visible (and customizable) to the user.  I also think the autoload cookies here and in the defgroup [aren't necessary](https://www.gnu.org/software/emacs/manual/html_node/elisp/When-to-Autoload.html#When-to-Autoload), but maybe you can clarify why they might need to be.

No byte-compilation issues! 😎

No issues! 🎉

Miscellaneous and minor:
- kaocha-runner.el#L21: Prefer `https` over `http` (if possible)

magnars:
All good points. Thank you, I'll fix those at once.

The autoload for group/custom is just a piece of cargo culting on my part, having seen it in other packages. If they are not needed, I'll remove them.

magnars:
Also, sorry for not catching the lint error. I added the autoloading without rerunning the linter. Shame on me.

magnars:
All feedback has now been followed.

purcell:
Looks good, even if the Clojurey-but-not-emacs-lispy "?" suffix in place of "-p" hurts my eyes a bit. ;-)  Merging now, thanks.

magnars:
Thanks, Steve! :)

lør. 17. aug. 2019 kl. 07:23 skrev Steve Purcell <notifications@github.com>:

PR #6363, opened 2019-08-10
add eshell-eldoc

defaultxr:

Eshell-eldoc provides Eldoc functionality for Eshell. It provides the normal Eldoc for Elisp code when Elisp functionality is being called, or a summary of the command from `whatis` when an external command is being called.

https://github.com/defaultxr/eshell-eldoc

Maintainer.

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/eshell-eldoc
+++ b/recipes/eshell-eldoc
@@ -1,0 +1,1 @@
+(eshell-eldoc :fetcher github :repo "defaultxr/eshell-eldoc")


riscy:

```
Warning (emacs): 
eshell-eldoc.el:0: The first line should be of the form: ";;; package --- Summary"
```

```
No issues found.
```

```
eshell-eldoc.el:75:1:Warning: Unused lexical variable `eshell-cmd'
```

No issues! 🎉

Automated sharp-quote suggestions for your consideration:

Miscellaneous and minor:
- eshell-eldoc.el#L25: Prefer `https` over `http` (if possible)

defaultxr:
> eshell-eldoc.el:75:1:Warning: Unused lexical variable `eshell-cmd'

Fixed.

> eshell-eldoc.el:0: The first line should be of the form: ";;; package --- Summary"

I wasn't sure if the `-*- lexical-binding: t -*-` declaration needed to be the first line or would be fine on the second. If it's fine on the second line, I can swap them.

riscy:
> I wasn't sure if the `-*- lexical-binding: t -*-` declaration needed to be the first line or would be fine on the second. If it's fine on the second line, I can swap them.

Surprisingly enough, they both have to be on the first line -- add the lexical binding stuff at the end. :)

defaultxr:
Ahh, gotcha! OK, I changed that now too, hopefully all looks good, let me know if not.

walseb:
I'm not associated with melpa but here are a couple of points. You should clarify that you have to run `sudo mandb`  to get `whatis` working or else you will get a lot of confused people asking why it doesn't work. Since you already listed one similar library I think you should add [esh-help ](https://github.com/tom-tan/esh-help) too. It's kind of the same thing except it lists the possible command line arguments instead of the `whatis` output but it was also really slow with complex commands last time I tried it. Maybe if they cache the outputs like you did it will run a lot faster. Anyways thanks for the great lib!

defaultxr:
Thanks @walseb ! I've implemented your suggestions. I actually wasn't aware of esh-help and I didn't find it by searching (probably would've found it if it was called "eshell-help" instead), otherwise I probably wouldn't have written my own library.

As for speed, both my library and esh-help run too quickly on my machine for me to tell a difference (though I've only tested esh-help on a few commands). I'm glad my library is of use to you, but maybe I will see if my efforts could be merged into esh-help instead. Perhaps its author would accept a PR adding an option to use `whatis` instead of `man`. From a cursory glance it looks like esh-help gets its information by parsing man pages, so if one is really long then it could be the cause of the slowdown you're seeing.

For now I'm going to close this PR since it looks like esh-help is already covering this functionality. If/when I get around to submitting a PR to esh-help I might tag you to help test any speed improvements I implement, if that's alright. Thanks again!

walseb:
No problem! Yeah that's probably what's happening and it doesn't help that I have eldoc delay set to 0. Looking forward to a merge of these packages, just tag me and I will check for any problems!

riscy:
Thank you both for this resolution!

PR #6365, opened 2019-08-11
add org-analyzer

rksm:

Visualizes org-mode time tracking data.

https://github.com/rksm/clj-org-analyzer

maintainer

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/org-analyzer
+++ b/recipes/org-analyzer
@@ -1,5 +1,4 @@
 (org-analyzer :fetcher github
               :repo "rksm/clj-org-analyzer"
               :files ("org-analyzer-el/*.jar"
-                      "org-analyzer-el/*.el"
-                      (:exclude "org-analyzer-el/org-analyzer-pkg.el")))
+                      "org-analyzer-el/*.el"))
--- a/recipes/org-analyzer
+++ b/recipes/org-analyzer
@@ -1,0 +1,5 @@
+(org-analyzer :fetcher github
+              :repo "rksm/clj-org-analyzer"
+              :files ("org-analyzer-el/*.jar"
+                      "org-analyzer-el/*.el"
+                      (:exclude "org-analyzer-el/org-analyzer-pkg.el")))


recipes/org-analyzer
```diff
@@ -0,0 +1,3 @@
+(org-analyzer :fetcher github
+              :repo "rksm/clj-org-analyzer"
+              :files ("org-analyzer-el/*.el"))
```
riscy: This recipe looks like it will include `org-analyzer-pkg.el`, which it shouldn't.

rksm:
Thanks for checking! The issues should be addressed know.

alphapapa:
This looks really cool.  

Is there a MELPA policy about packages that download and run more executable code at runtime?  In other words, the Elisp package is not self-contained.  I feel like users may be surprised by this.  I would expect to be prompted for confirmation.

Also, if the `jar` download becomes unavailable, and the user doesn't already have it, the Elisp package won't work.

rksm:
I've added a prompt regarding the download. And I'm sorry but I won't have time to rewrite the server in pure elisp.

alphapapa:
I wouldn't expect you to do that.  It'd probably slow, anyway.

What if you package the jar in the elisp package?

rksm:
Sure, this is no problem. I've updated [rksm/clj-org-analyzer](https://github.com/rksm/clj-org-analyzer) accordingly.

alphapapa:
Cool.  I hope MELPA approves of that.

rksm:
Thank you for your help!

purcell:
Looks good to me, thanks. I'm going to edit the recipe before merging to remove the exclusion of the `-pkg.el` file, but you should really just delete that file from the repo: MELPA will read package metadata from the `.el` file which matches the package name, so it's redundant to have a `-pkg.el` file for our sake. Other than that, welcome to MELPA!

rksm:
Thank you!

PR #6366, opened 2019-08-12
Update `picolisp-mode` recipe.

flexibeast:


tarsius:
Posting the same request as at https://github.com/flexibeast/picolisp-mode/commit/d608324457f436a9e404ecf107abaf2e2b7772a7#commitcomment-34716387:

> A fork of tj64's `inferior-picolisp' (https://github.com/tj64/picolisp-mode/), stripped down to only provide the minimum necessary for Org Babel session support, and modified to be compatible with flexibeast's `picolisp-mode' (https://github.com/flexibeast/picolisp-mode/). Initial work on the fork, removing unneeded functionality, was done by cryptorick (https://github.com/cryptorick).

Please consider renaming your fork. As part of my work on the Emacsmirror I try to make sure that each feature is only provided by a single package. That benefits users because they won't end up accidentally using one version of a library when they intended to use the other version just because the wrong version came earlier on the load-path. Thanks!

flexibeast:
@tarsius: 

Thanks for the comment about the commentary being on a single line; i'll try to address that issue in this package, and my other packages, as soon as i can.

More generally, see my reply [here](https://github.com/flexibeast/picolisp-mode/commit/d608324457f436a9e404ecf107abaf2e2b7772a7#commitcomment-34716680). :-)

tarsius:
> Thanks for the comment about the commentary being on a single line; i'll try to address that issue in this package, and my other packages, as soon as i can.

That would be much appreciated. I think you should limit the length of both code and comment lines to ~80 lines. Or maybe 90 if you prefer; but you shouldn't go higher than that.

flexibeast:
Okay, i'll go with an 80-character limit. :-) Thanks again!

purcell:
I've removed the :files spec from the recipe entirely, since the defaults will pull in both .el files. While I have the same concerns as @tarsius re. the name clashes and possible confusion leading from the forked repo, this particular addition to the pre-existing MELPA recipe doesn't make things worse. Obviously the best outcome in the long term would be for the community to have a single generally-supported `picolisp-mode`, so joining forces with tj64 where possible would make sense.

PR #6367, opened 2019-08-13
Add mozc-cand-posframe

akirak:

Posframe frontend for mozc.el, the Japanese input method by Google

https://github.com/akirak/mozc-posframe

I am the maintainer

**None needed**

### Checklist

Please confirm with `x`:

`mozc.el` which is a dependency of this package doesn't have a version, so I was unable to include its minimum version in the header as in the below. mozc.el has no stable version, even though it hasn't been updated for a few years. package-lint warns about that, but it is not a problem.

```
;; Package-Requires: ((emacs "26.1") (posframe "0.5.0") (mozc "0"))
```

### DIFF
--- a/recipes/mozc-cand-posframe
+++ b/recipes/mozc-cand-posframe
@@ -1,0 +1,1 @@
+(mozc-cand-posframe :fetcher github :repo "akirak/mozc-posframe")


riscy:
Looks great to me.  If you find you need to, you can also refer to snapshot versions of mozc, e.g. `(mozc "20180101.800")`

PR #6369, opened 2019-08-14
Add awscli-capf, a completion at point function for the AWS CLI

sebasmonia:

A completion at point function, with company support, for commands and parameters form the AWS command line tool.

https://github.com/sebasmonia/awscli-capf

Mantainer

N/A

### Checklist

Please confirm with `x`:

package-lint and I disagree on the prefix for private functions. I am open to suggestions on

1. Changing the package name and thus the prefix
2. Changing the prefix only
3. Doing what package-lint suggests if you agree with its feedback.

Basically `awscli-capf--` vs `awscli--capf-` for internal functions.

### DIFF
--- a/recipes/awscli-capf
+++ b/recipes/awscli-capf
@@ -1,0 +1,1 @@
+(awscli-capf :repo "sebasmonia/awscli-capf" :fetcher github)


riscy:
Before a closer look, please look into these (semi-automated) checks:

No issues! 💯

```
15 issues found:

37:1: error: "awscli--capf-script-dir" doesn't start with package's prefix "awscli-capf".
38:1: error: "awscli--capf-data-file" doesn't start with package's prefix "awscli-capf".
39:1: error: "awscli--capf-services-info" doesn't start with package's prefix "awscli-capf".
40:1: error: "awscli--capf-global-options-info" doesn't start with package's prefix "awscli-capf".
42:1: error: "awscli-add-to-capf" doesn't start with package's prefix "awscli-capf".
91:1: error: "awscli--capf-help-buffer" doesn't start with package's prefix "awscli-capf".
97:1: error: "awscli--capf-annotation" doesn't start with package's prefix "awscli-capf".
106:1: error: "awscli--capf-store-data-in-file" doesn't start with package's prefix "awscli-capf".
113:1: error: "awscli--capf-read-data-from-file" doesn't start with package's prefix "awscli-capf".
122:1: error: "awscli--capf-param-strings-only" doesn't start with package's prefix "awscli-capf".
126:1: error: "awscli--capf-service-completion-data" doesn't start with package's prefix "awscli-capf".
136:1: error: "awscli--capf-command-completion-data" doesn't start with package's prefix "awscli-capf".
151:1: error: "awscli--capf-parameters-completion-data" doesn't start with package's prefix "awscli-capf".
211:1: error: "awscli--capf-service-data-from-cli" doesn't start with package's prefix "awscli-capf".
237:1: error: "awscli--capf-command-data-from-cli" doesn't start with package's prefix "awscli-capf".
```
I think you just want to rename these to e.g. `awscli-capf--command-data-from-cli`.

```
awscli-capf.el:265:1:Warning: the following functions are not known to be
    defined: position, subseq, concatenate
```
To solve these, you should `(require 'cl-lib)` and prefer `cl-position`, `cl-subseq`, and `cl-concatenate` if you can -- use of the `cl` package is deprecated.

sebasmonia:
Thank you for your feedback!

*Package prefix*: left a note regarding this point above, regardless I went with option 3 and your suggestion, and renamed all functions to agree with `package-lint`

*cl-lib*: An earlier version of this library required `cl`, the functions mentioned have been corrected to use `cl-lib` as the rest of the package.

You can also see in the list of issues a few other items to address. I think the biggest one is, some strings could be stored and read back propertized instead of doing it on each completion.

riscy:
Thanks!  I'm still seeing one more byte-compile lint (you should use `cl-concatenate` instead of the deprecated `concatenate`).  [Here](https://github.com/sebasmonia/awscli-capf/blob/10de5d3efeaf2d2352c6a2345f8137bc01d03c68/awscli-capf.el#L37), `load-file-name` will be nil causing an error if someone tries to run `eval-buffer` -- you can instead write something like:
```elisp
(defconst awscli-capf--script-dir 
  (if load-file-name (file-name-directory load-file-name) default-directory)
  "...")
```

Using the ``` `' ```-style quoting on `candidates` does no harm [here](https://github.com/sebasmonia/awscli-capf/blob/10de5d3efeaf2d2352c6a2345f8137bc01d03c68/awscli-capf.el#L94) or [here](https://github.com/sebasmonia/awscli-capf/blob/10de5d3efeaf2d2352c6a2345f8137bc01d03c68/awscli-capf.el#L101), but that quoting is usually used to create links in Emacs' online help to variables and functions with supporting documentation -- just fyi.  Also, the variable name is `candidate` not `candidates` at that point.

I may be missing something, but it looks like `company` is a soft dependency and (only?) used to generate a help buffer.  Could the hard dependency be removed with some judicious use of `require` and `declare-function`?

Minor: Sometimes it's worth turning a literal like "aws" [here](https://github.com/sebasmonia/awscli-capf/blob/10de5d3efeaf2d2352c6a2345f8137bc01d03c68/awscli-capf.el#L183) into a customizable -- sometimes users might have two versions of the cli installed (for example) and this affords a bit more flexibility.

Also very minor, "ocurrences" should be "occurrences" in a few places.

sebasmonia:
Took a while to start revisiting the package. Here a run down item by item:

* Changed `concatenate` to `cl-concatenate`
* Modified definition of `awscli-capf--script-dir` as per your suggestion
* `candidates` in the comments you mentioned referred to the `candidates` list in the function awscli-capf, modified the comments to better express this idea.
* Corrected spelling of "occurrences"

Will add a defcustom for `aws` as both the process to call and after giving some thought, to the keyword that kicks off completion [here](https://github.com/sebasmonia/awscli-capf/blob/10de5d3efeaf2d2352c6a2345f8137bc01d03c68/awscli-capf.el#L56).

Regarding the company dependency, all I do is add some text properties that company will interpret is present. So if company is not installed, everything else works as usual via CAPF.

Will test a bit more without company + after adding customization for those options, and resubmit.
Thank you!

sebasmonia:
Added defcustom options. Documentation strings validated with `checkdoc-eval-current-buffer`.

I think this is the last feedback item to address.

riscy:
Thanks - sorry for the multiple turn arounds but I'm still seeing a compiler warning:
```
awscli-capf.el:282:1:Warning: the function `company-doc-buffer' is not known
    to be defined.
```

sebasmonia:
Don't be sorry at all. I appreciate your effort and all feedback.

I think I need to byte compile uninstalling company, and use fboundp, to see if that removes the warning.

When I tested without company, I tried `C-h`, I know it didn't break anything (without company that binding is not present).

Need to mind the compiler too. Fix coming _soon_!

Thank you,
Sebastián

sebasmonia:
Pushed a change that removed `company` from the requirements list and guards the call to `company-doc-buffer` with `fboundp` to eliminate the compiler warning.

riscy:
Thanks!  And, thanks for your patience.  This is looking good to merge.

Issue #4154, opened 2016-08-16
Automate purcell


purcell:
I'm getting burnt out on MELPA, or rather my time for it is increasingly limited. I hope that both authors and the general level of package quality on MELPA have benefited from there being someone available to provide code reviews and - occasionally - turn away packages that aren't sufficiently novel or widely useful.

So my question is: if I scale back my time commitment, how can we continue to keep MELPA awesome? Here are some of my thoughts:

There are some basic packaging issues (malformed headers etc., redundant or deprecated `:files` specs) which don't get caught by our Travis build. We should fail faster and with clear information, so that we can simply refer contributors to the failing build output. This might involve having a "pendantic" setting in package-build, to escalate some of what are currently debug messages about such issues. `package-build-current-recipe` should also fail noisily in such cases, to head off submission of not-really-working-well recipes.

Some packaging-specific issues (e.g. `cl-lib` deps, long description lines) can already be detected by the code in `flycheck-package`. Further work is needed for packaging-related issues such as using `setq-local` or `advice-add` but not declaring a corresponding dependency such as `(emacs "24.4")`.

Other packaging issues span files, e.g. including `company` or `helm` backends in a package which doesn't declare a package dependency on the corresponding package. Those checks can't be built into `flycheck-package`. I imagine a separate package linter which `flycheck-package` could use.

There are a ton of other stylistic and code quality issues I routinely check for when reviewing upstream code, e.g. inconsistent symbol prefixes, crappy docstrings, missing autoloads, empty commentaries etc. There needs to be a great linter for these problems too, probably combined with the package linter mentioned above.

Additionally, even after recipes get merged, code or packaging issues can arise later, so it'd be great to provide the same detailed feedback about packages on an ongoing basis: it could be generated each time we build them.

Other issues such as whether a package adds to the ecosystem are a bit more subjective: we need to be asking authors, "how is this different to packages X & Y?", and "could you work with Z to add this functionality directly there?" I have the advantage of having reviewed thousands of packages at this point, so I can make mental connections, but it's a useful thing which we should continue to do.

It probably makes sense for me to write an exhaustive checklist of all the stuff I do when reviewing packages, but I'm also keen to get input from others.

lunaryorn:
@purcell I don't think it's likely that we can get a good Emacs Lisp linter quickly, before you burn out entirely :grin:  Linting is hard, the more the "softer" quality criteria gets.  No Emacs Lisp linter will ever be able to check for "good" docstrings.

[...]

PR #6370, opened 2019-08-14
Add recipe: grip-mode.

seagle0128:

Instant Github-flavored Markdown preview using [grip] https://github.com/joeyespo/grip) (Github README Instant Preview).

https://github.com/seagle0128/grip-mode

N/A

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/grip-mode
+++ b/recipes/grip-mode
@@ -1,0 +1,1 @@
+(grip-mode :repo "seagle0128/grip-mode" :fetcher github)


seagle0128:
Updated the checklist. Please merge. Thanks！

markhepburn:
@seagle0128  I just came here wondering when it would be merged too, but probably also worth considering https://github.com/melpa/melpa/issues/4154

(Love your package btw!)

riscy:
Thanks for the close adherence to the checklist.  One comment I have is that looking for the executable once [here](https://github.com/seagle0128/grip-mode/blob/8792144d856aff7dd5c3eed7698987db3f073a51/grip-mode.el#L53) means that, if the user installs grip, they'll have to restart Emacs (or painstakingly re-set the variable) before grip is detected correctly -- and the error message doesn't tell them as much.

Maybe when the check [here](https://github.com/seagle0128/grip-mode/blob/8792144d856aff7dd5c3eed7698987db3f073a51/grip-mode.el#L72-L73) fails, you could try once more to locate the binary.  Or you could try to locate the binary on every call to `grip-start-process`, which would be (undetectably) slower but more robust.

Another ultra minor thing I'm seeing is that "Github" is usually branded "GitHub".

seagle0128:
@riscy Thanks a lot for the suggestion! I've updated the codes according to your comments.

purcell:
> Maybe when the check [here](https://github.com/seagle0128/grip-mode/blob/8792144d856aff7dd5c3eed7698987db3f073a51/grip-mode.el#L72-L73) fails, you could try once more to locate the binary. Or you could try to locate the binary on every call to `grip-start-process`, which would be (undetectably) slower but more robust.

Yeah, it's generally best not to try to locate a specific binary path at all, because that's what `exec-path`/`$PATH` are for. If the user wants to provide a full path when customising an executable path, then they can still do that.

A couple more notes from me:
- You should provide a `:group` for your `defgroup` form, perhaps `applications`.
- The `buffer-file-name` handling [here](https://github.com/seagle0128/grip-mode/blob/15f6d5777f9124cda5940a4f70ee6905edc0d144/grip-mode.el#L111-L126) is a bit inconsistent: in one function, nothing happens if the buffer is not backed by a file, and in the other, and error will occur. The ideal would be that temp files should be used when the buffer does not have a filename: it's generally best not to assume that every buffer has a file.
- Don't use `concat` to build file names, in any case. In fact, [here](https://github.com/seagle0128/grip-mode/blob/15f6d5777f9124cda5940a4f70ee6905edc0d144/grip-mode.el#L113) you should probably just use `default-directory` instead of trying to get the buffer file's directory.

Hope that helps! This is okay to merge in the meantime, but please take a look at those points above in your own time.

seagle0128:
Thank you very much, @purcell !

The comments are helpful. I am working on it.

PR #6371, opened 2019-08-14
Add counsel-test recipe

xmagpie:

An emacs package that integrates ivy/counsel with different testing frameworks and allows interactive test discovery and execution.

https://github.com/xmagpie/counsel-test

**None needed** 

### Checklist

Please confirm with `x`:

- [ ] `M-x checkdoc` is happy with my docstrings. There are several false positive warnings about the *tests* word in several docstrings. I beleive these can be ignored.

### DIFF
--- a/recipes/counsel-test
+++ b/recipes/counsel-test
@@ -1,0 +1,4 @@
+(counsel-test
+ :repo "xmagpie/counsel-test"
+ :fetcher github
+ :files (:defaults "counsel-test.el"))


recipes/counsel-test
```diff
@@ -0,0 +1,3 @@
+(counsel-test
+ :repo "xmagpie/counsel-test"
```
riscy: I think you will have to be more explicit with this recipe, since the [default recipe](https://github.com/melpa/melpa#recipe-format) will ignore `-test.el` files.

purcell: Yep, that should have been obvious when building and testing the package as per the checklist. Please let us know if the instructions for that were not clear.

purcell:
Thanks, I left a comment above, but the code itself looks good to me, so once the recipe is fixed and tested, this should be good to merge.

recipes/counsel-test
xmagpie: I am sorry, that is my bad. All instructions are good, it seems like I have not followed all of them properly, so it is just my lack of attention.

xmagpie:
Thanks for the review! I'll provide the fix tomorrow, when I get home.

xmagpie:
@purcell, @riscy  I have explicitly specified `counsel-test.el` for inclusion together with `:defaults`. I have checked the package archive contents after running `C-c C-c` for the recipe file and also checked package with `make sandbox INSTALL=counsel-test`.

I hope, I did not misunderstand or forget anything. Let me now if something still needs any fixes.

riscy:
Great thanks!  And, welcome to MELPA. :)

xmagpie:
Thanks, this is a great pleasure :)

PR #6372, opened 2019-08-15
New pack/elispfl

cireu:
Extra font-lock for emacs-lisp mode

https://github.com/cireu/elispfl

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/elispfl
+++ b/recipes/elispfl
@@ -1,0 +1,1 @@
+(elispfl :repo "cireu/elispfl" :fetcher github)


alphapapa:
From looking at the screenshot, I can't tell what this adds to the default.  You should explain in words what exactly it does.

cireu:
@alphapapa Add control group's screenshot to README.org now 

https://github.com/cireu/elispfl/blob/master/README.org

alphapapa:
It would be helpful to explain in words what it does, and how it differs from other packages like https://github.com/alphapapa/highlight-function-calls

cireu:
@alphapapa  I didn't notice there was such package before. And I have a simple look on your package.

1. Your package contains so many whispers in code, like https://github.com/alphapapa/highlight-function-calls/blob/master/highlight-function-calls.el#L105. It's really confusing.
elispfl is simple and clear

2. elispfl differentiate atomic functions(e.g `mapcar` `car` cdr`) and lisp/userland functions, they will be colorized in different face.

3. elispfl use `(eq (char-before (match-beginning 0)) ?\()` to check whether a symbol is subroutine call or not, this way can work well with general list literals like

```
'(
  foo
  bar
)
```

alphapapa:
> 1. Your package contains so many whispers in code, like https://github.com/alphapapa/highlight-function-calls/blob/master/highlight-function-calls.el#L105. It's really confusing.
>    elispfl is simple and clear

??

cireu:
> > 1. Your package contains so many whispers in code, like https://github.com/alphapapa/highlight-function-calls/blob/master/highlight-function-calls.el#L105. It's really confusing.
> >    elispfl is simple and clear
> 
> ??

So what's the point of `(0 . nil)`? I don't use this magick cons but elispfl can do the highlight.

cireu:
@alphapapa There was no much different between to package, so I close this PR.

purcell:
Thanks folks, I agree it's best to avoid duplication. :-)

alphapapa:
> So what's the point of (0 . nil)? I don't use this magick cons but elispfl can do the highlight.

You mean this?

```el
     ;; NOTE: The (0 nil) is required, although I don't understand
     ;; exactly why.  This was confusing enough, following the
     ;; docstring for `font-lock-add-keywords'.
     (0 nil)
```
You will have to refer to the docstring for `font-lock-keywords`.  Maybe you can understand it better than I could at the time I wrote that.

> I don't use this magick cons but elispfl can do the highlight.

According to the documentation, there are at least 5 or 6 different ways to write font-lock keywords and matchers.  I don't know which is the best or most appropriate for this case.  It seems to work well and fast, so I think it's fine, but if you can suggest improvements, that would be welcome.

> There was no much different between to package, so I close this PR.

I think that's probably the best thing to do, since they're so similar.  However, if you would like to propose improvements to `highlight-function-calls`, please do.  Thanks.

cireu:
On Sun, 18 Aug 2019 17:46:04 +0800,
alphapapa wrote:
> 
> [1  <text/plain; UTF-8 (7bit)>]
> [2  <text/html; UTF-8 (7bit)>]
>  So what's the point of (0 . nil)? I don't use this magick cons but elispfl can do the highlight.
> 
> You mean this?
> 
>      ;; NOTE: The (0 nil) is required, although I don't understand
>      ;; exactly why.  This was confusing enough, following the
>      ;; docstring for `font-lock-add-keywords'.
>      (0 nil)
> 
> You will have to refer to the docstring for font-lock-keywords. Maybe you can understand it better than I could at the time I wrote that.
> 
>  I don't use this magick cons but elispfl can do the highlight.
> 
> According to the documentation, there are at least 5 or 6 different ways to write font-lock keywords and matchers. I don't know which is the best or most appropriate for
> this case. It seems to work well and fast, so I think it's fine, but if you can suggest improvements, that would be welcome.
> 
>  There was no much different between to package, so I close this PR.
> 
> I think that's probably the best thing to do, since they're so similar. However, if you would like to propose improvements to highlight-function-calls, please do.
> Thanks.
> 
> ―
> You are receiving this because you modified the open/close state.
> Reply to this email directly, view it on GitHub, or mute the thread.

Thanks for your recommendation and detailed explanation, but I may not have a try on
`highlight-function-calls` because `elispfl` was 
extracted from my config(in fact doom-emacs's config), I have used it for a long 
time. 

If you want to add option to differentiate builtin subroutine and userland's for
`highlight-function-calls`. I' ll be happy if my implementation can help you.

a13:
@alphapapa There's also an older package [Highlight-defined](https://github.com/Fanael/highlight-defined)

PR #6374, opened 2019-08-15
Add header-tracker

nathanpc:

Quickly and easily keep track and switch between C header and source files.

https://github.com/nathanpc/header-tracker.el

**None needed**

### Checklist

### DIFF
--- a/recipes/header-tracker
+++ b/recipes/header-tracker
@@ -1,0 +1,1 @@
+(header-tracker :fetcher github :repo "nathanpc/header-tracker.el")


terlar:
Isn't this already available via the built-in `find-file` package?

purcell:
> Isn't this already available via the built-in `find-file` package?

Yes, agree - this appears to be a reinvention of that more general functionality. On the plus side, that means you had a good idea @nathanpc! :-)  I'll go ahead and close this, then, but I look forward to your next submission.

PR #6375, opened 2019-08-15
Add vdf-mode

pmiddend:

The package adds highlighting and indentation for Valve's VDF files.

https://github.com/plapadoo/vdf-mode

I am the sole author of the package.

Not sure?

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/vdf-mode
+++ b/recipes/vdf-mode
@@ -1,0 +1,1 @@
+(vdf-mode :repo "plapadoo/vdf-mode" :fetcher github)


riscy:

No issues! 💯

```
3 issues found:

22:0: error: Package should have a non-empty ;;; Commentary section.
70:3: error: You should depend on (emacs "24.3") if you need `setq-local'.
71:3: error: You should depend on (emacs "24.3") if you need `setq-local'.
```

```
vdf-mode.el:44:1:Warning: Unused lexical argument `posEnd'
```
Incidentally, I notice right away that you're using `camelCase` instead of the elisp standard (which is `hyphenated-lowercase`) for your variable names.  You should switch those over.

Added miscellaneous checks (automated):
- vdf-mode.el#L66: It is unusual to derive from fundamental-mode; use prog-mode in this case
- vdf-mode.el#L41: Parentheses should be gathered together on one line
- vdf-mode.el#L53: Parentheses should be gathered together on one line
- vdf-mode.el#L20: Prefer `https` over `http` (if possible)
- vdf-mode.el#L68: You can use `setq-local` here
- vdf-mode.el#L72: You can use `setq-local` here

pmiddend:
@riscy Thanks for the review! I think I've fixed the most important warnings now and also switched to `hyphenated-lowercase`.

riscy:
- In `vdf-count`, `pos-end` still isn't used [here](https://github.com/plapadoo/vdf-mode/blob/611e470f2479a9a743bad75e23d472ff790d65fd/vdf-mode.el#L46).  If that's ok with you, the variable can be removed and the documentation can be updated to prevent someone from being misled.  I mentioned this in my first comment, but that may have gotten buried.  Enabling lexical binding would have helped you spot this.
- `synTable` -> `syn-table` (or, even better, `syntax-table`).
- You're still deriving from `fundamental-mode` instead of `prog-mode`.

Now that I look at the code, I also see `defvar` followed by `setq` is unnecessary up [here](https://github.com/plapadoo/vdf-mode/blob/611e470f2479a9a743bad75e23d472ff790d65fd/vdf-mode.el#L29-L43).  You can (and should) just make one call to `defvar`.

pmiddend:
@riscy Thank you for the thorough review. I'm sorry for my rookie mistakes, but it's my first real elisp package that's not 100% copy&paste, so I'm still learning lots of stuff.
I've applied your suggestions and am, at last, grateful for lexical binding. It has helped.
I've deliberately not changed `fundamental-mode` to `prog-mode` at first, because I didn't really understand *why* I should derive from `prog-mode`, since `vdf-mode` seems to be working finde with `fundamental-mode`.

pmiddend:
Ah, I see. A very good reason!

I've now changed this to `prog-mode`. Is there anything else left to do?

riscy:
I'm just doing some user testing on this.  It seems like automatic (electric) indentation isn't working, which is probably down to the `vdf-indent-line` function -- do you experience similar?

I'm not well versed in the norms of writing indent functions, unfortunately, but it seems like you might want to do something more along these lines:
```elisp
(defun vdf-indent-line ()
  "Indent current line as VDF code."
  (let* ((open-count (save-excursion
                       (beginning-of-line)
                       (vdf-count-backwards "{" (point-min))))
         (close-count (save-excursion
                        (end-of-line)
                        (vdf-count-backwards "}" (point-min))))
         (depth (- open-count close-count)))
  (if (<= (current-column) (current-indentation))
      (indent-line-to depth)
    (save-excursion (indent-line-to depth)))))
```

riscy:
That could probably be simplified even further if `vdf-count-backwards` took a `pos-begin` as well as a `pos-end` (and then we wouldn't need to mess around with the `save-excursion/beginning-of-line` stuff).

pmiddend:
Thanks for the feedback @riscy, I'm learning a lot! I'm on vacation the next week, so don't worry if the PR isn't updated for this time. ;)

riscy:
Sure!  Feel free to ping when ready.

riscy:
Ping! :)

pmiddend:
@riscy Oh wow, thanks for the reminder, I'll get on it next week!

pmiddend:
@riscy Okay, I finally got around to refactoring this a bit. You can test again now. :)

riscy:
Great, thanks!

PR #6376, opened 2019-08-15
add org-tanglesync recipe

mtekman:

Pulling external file changes to a tangled org-babel src block is surprisingly not an implemented feature. This package does just that if a difference between an org src tangled block and the actual external tangled file are detected.

https://github.com/mtekman/org-tanglesync.el

Maintainer

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/org-tanglesync
+++ b/recipes/org-tanglesync
@@ -1,0 +1,1 @@
+(org-tanglesync :fetcher github :repo "mtekman/org-tanglesync.el")


riscy:

No issues! 💯

```
No issues found.
```

```
org-tanglesync.el:113:1:Warning: Unused lexical variable `buffer-modified-p'
org-tanglesync.el:124:1:Warning: Unused lexical argument `org-buffer'
org-tanglesync.el:124:1:Warning: Unused lexical argument `external'
org-tanglesync.el:124:1:Warning: Unused lexical argument `internal'
org-tanglesync.el:132:1:Warning: Unused lexical argument `org-buffer'
org-tanglesync.el:144:1:Warning: Unused lexical argument `internal'
org-tanglesync.el:169:1:Warning: Unused lexical variable `method-do'

In end of data:
org-tanglesync.el:274:1:Warning: the function `org-src-source-buffer' is not
    known to be defined.
```

- org-tanglesync.el#L270: It is beneficial to sharp-quote hooks; use `#'`

purcell:
Thanks for this. Couple of notes:

- You're using `progn` in a bunch of places, but I don't think any of them is required. Perhaps you've misunderstood what it's for? In any case, the code would be clearer without them.
- Is there a way to toggle this behaviour? It looks like `org-src-mode-hook` is changed globally, but I suspect it should be possible to enable/disable this behaviour on a per-file or per-directory basis, so I'd probably expect to see a minor mode instead.

mtekman:
@purcell Thanks for the comprehensive review! I have removed the `progn`s and reworded my conditionals to not rely on them so much.

As for `org-src-mode-hook` issue. The idea was that as long as this package is installed, the default behaviour should be that the `org-tanglesync-user-edit-buffer` function will be enabled... but I added another custom variable which the user can toggle to change this. To enable this on the file/directory level, I leave this to user to change the `org-tanglesync-enable-src-mode-hook` var.

I hope that works, looking forward to your next review!

riscy:
I'm catching a compilation warning still:

```
In end of data:
org-tanglesync.el:304:1:Warning: the function ‘method-do’ is not known to be
    defined.
```

For `pushnew` you want to `(require cl-lib)` and for `string-trim` you want to `(require 'subr-x)` (rather than rely on transitive imports).

mtekman:
Hi @riscy , thanks a lot for the review - I added the required imports and fixed the missing function.

riscy:
Apologes for missing this the first go around -- `pushnew` should be `cl-pushnew` (the older version is deprecated).  You should keep the `(require 'cl-lib)` however.

elisp distinguishes between variables and functions, so this for example isn't valid:
```elisp
(defun org-tanglesync-perform-action (internal external org-buffer method-do)
  (method-do internal external org-buffer))
```
but this is:
```elisp
(defun org-tanglesync-perform-action (internal external org-buffer method-do)
  (funcall method-do internal external org-buffer))
```
The byte-compiler should warn you about situations like this.

> As for org-src-mode-hook issue. The idea was that as long as this package is installed, the default behaviour should be that the org-tanglesync-user-edit-buffer function will be enabled... but I added another custom variable which the user can toggle to change this. To enable this on the file/directory level, I leave this to user to change the org-tanglesync-enable-src-mode-hook var.

The best solution is to implement a minor mode -- which is the preferred Emacs convention for toggling behavior off and on, rather than directly changing the value of some variable.  Then, your hook function (`org-tanglesync-user-edit-buffer`) can check if the mode is active or not, rather than the value of the variable `org-tanglesync-enable-src-mode-hook`.

mtekman:
Thanks @riscy, I added a minor mode and removed the defcustom, as well as using the right cl function and funcall :-)

riscy:
Great!  You could just call the minor mode `org-tanglesync-mode` (rather than `org-tanglesync-minor-mode`) if you wanted, but I'll leave that up to you to consider.  For [this line](https://github.com/mtekman/org-tanglesync.el/commit/19abbadc227eb064cd36fba432706d251fa3b60d#diff-adde777c328ad02c33580c856faed592R291), you again don't need to check whether the minor-mode variable is bound, because declaring the minor mode is guaranteed to do that for you.  

These aren't blockers, so this looks ready to merge.  Thanks for this!

PR #6378, opened 2019-08-16
Add comint-hyperlink

matthewbauer:

A filter for comint to add terminal hyperlinks as detailed in https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda

https://github.com/matthewbauer/comint-hyperlink

Maintainer

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/comint-hyperlink
+++ b/recipes/comint-hyperlink
@@ -1,0 +1,1 @@
+(comint-hyperlink :fetcher github :repo "matthewbauer/comint-hyperlink")


alphapapa:
Reading the discussion on the gist is scary.  The "author" of the "proposal" doesn't even understand that URLs are more than just links to web sites:

> If you're worried about opening a URL having unexpected, bad consequences, e.g. due to lack of proper CSRF prevention, then it's those websites that should be subject to doubt, not the terminal.

This is after someone else already showed him examples like iOS bugs which allowed apps to open URLs to cause the phone to repeatedly call police departments.

He keeps dismissing security vulnerabilities people explain, while he's filing over 24 bug reports asking developers to implement support for his dangerous idea.  And he insists on keeping the discussion on that Gist, where he can delete critical comments, which he already admitted to doing.

Look at his primary statement about security:

> This feature doesn't introduce anything that's not already present while browsing the web. Therefore we believe this feature doesn't have security aspects to worry about.

Well, that should be just fine, bringing web-like security vulnerabilities into our terminals.  Nothing to worry about.  Just move along and accept the feature request.

His most recent comment:

> I have no plans to convert it to a repo. In fact, I'm tempted to convert it to a static page without the possibility of commenting.  If I get one more comment questioning the security of the feature in general, without pointing out an actual bug in this very specification (not elsewhere!), I'm probably going to do this.

None so deaf as those who will not hear.

matthewbauer:
I think as an opt-in feature it is reasonable to allow this kind of thing. You're right that security should be a concern and this does introduce a vector for phishing. But it is also very useful to have a standard way to annotate the location of things in the terminal, without having to show the user a full URL. systemd outputs these codes by default and coreutils has an option for ls to have it already. I've added a "filter" option as well to my package for users who just want to hide these codes. 

Most terminals (and emacs via url-handler-mode) provide some highlighting for URLs, and this is not controversial, but this control sequence takes things a step farther by making those URLs hidden from the user. I think the best way to address these concerns is just to make the URL very visible when you click on it. Perhaps some confirmation should also be necessary to make this a little bit harder?

alphapapa:
It's Free Software, so implement whatever features you want.

But I think it's a very bad idea for a terminal to do this, and I think it's a mistake to encourage it.

purcell:
Thanks, interesting, and I agree with the concerns, @alphapapa.

Are there actually any reputable programs which output URLs to the console in this manner?

matthewbauer:
> Are there actually any reputable programs which output URLs to the console in this manner?

systemctl outputs it by default and coreutils ```ls``` does with the ```--hyperlink``` flag.

riscy:
Just to chime in on this -- after some experimenting, I find that the `comint-preoutput-filter-function` `xterm-color-filter` (from the `xterm-color` package) is incompatible with your package (something to be aware of), but the built-in `ansi-color-filter-apply` filter appears to be compatible.

riscy:
I'm taking a look at this, even while the discussion above still seems unresolved.  Just a minor lint:

```
comint-hyperlink.el:111:1:Warning: Unused lexical variable `start'
```

My two cents is that it's not so bad, but I think we could address some of the concerns here by (optionally) making the behavior a little bit more mistake-proof.  Rather than immediately browsing a URL that gets clicked on (deliberately or not), perhaps we could prompt the user with a `yes-or-no-p` prompt.  This would *ensure* that the user is actually looking at the raw URL before an action is taken (relying purely on the `'help` field may not be enough).  Another option would be to add a user-customizable regexp for "safe" URL's that the user is comfortable opening without a prompt.

matthewbauer:
> Per CONTRIBUTING.org, consider [using lexical binding](https://github.com/melpa/melpa/blob/master/CONTRIBUTING.org#fixing-typical-problems) (and depend on at least (emacs "24"))
> (The following assumes lexical binding is enabled)
> 
> ```
> comint-hyperlink.el:111:1:Warning: Unused lexical variable `start'
> ```

Missed that, thanks! Updated in https://github.com/matthewbauer/comint-hyperlink/commit/0b7f6f44c1d13d14acb817c52a37525e592e22d0

> My two cents is that it's not so bad, but I think we could address some of the concerns here by (optionally) making the behavior a little bit more mistake-proof. Rather than immediately browsing a URL that gets clicked on (deliberately or not), perhaps we could prompt the user with a `yes-or-no-p` prompt. This would _ensure_ that the user is actually looking at the raw URL before an action is taken (relying purely on the `'help` field may not be enough). Another option would be to add a user-customizable regexp for "safe" URL's that the user is comfortable opening without a prompt.

Agreed that `yes-or-no-p` makes this safer. The idea is that you can always define a custom function for `comint-hyperlink-action`, but the default should probably be safe. Added in https://github.com/matthewbauer/comint-hyperlink/commit/e1e326a2c08e2de66b74074cfcd0bc48bb50d79a.

riscy:
Thanks for the added safety check.  I think this looks good, so I'm going to go ahead and merge.  Including a link to the gist in the frontmatter is a great idea.  You may also want to document the `xterm-color-filter` issue I mentioned above in your documentation somewhere, but that's not a blocker.

PR #6382, opened 2019-08-17
Add fancy-dabbrev

jrosdahl:

fancy-dabbrev essentially wraps the Emacs built-in dabbrev functionality, with two improvements:

1. **Preview**: The first expansion candidate will/can be shown when any text has been entered.
2. **Popup menu**: The first call to fancy-dabbrev-expand will expand the entered word prefix just like dabbrev-expand. But the second call will show a popup menu with other candidates (with the second candidate selected). The third call will advance to the third candidate, etc. It is also possible to go back to a previous candidate with fancy-dabbrev-backward. Selection from the menu can be canceled with C-g. Any cursor movement or typing will hide the menu again.

https://github.com/jrosdahl/fancy-dabbrev

I'm the maintainer.

N/A

### Checklist

Please confirm with `x`:

    - **I chose to ignore some suggestions about writing documentation for internal functions, though.**
    - **Installed using straight.el, not package.el.**

### DIFF
--- a/recipes/fancy-dabbrev
+++ b/recipes/fancy-dabbrev
@@ -1,0 +1,1 @@
+(fancy-dabbrev :repo "jrosdahl/fancy-dabbrev" :fetcher github)


riscy:

```
Warning (emacs): 
fancy-dabbrev.el:270: All variables and subroutines might as well have a documentation string
Warning (emacs): 
fancy-dabbrev.el:275: All variables and subroutines might as well have a documentation string
Warning (emacs): 
fancy-dabbrev.el:280: All variables and subroutines might as well have a documentation string
Warning (emacs): 
fancy-dabbrev.el:289: All variables and subroutines might as well have a documentation string
Warning (emacs): 
fancy-dabbrev.el:292: All variables and subroutines might as well have a documentation string
Warning (emacs): 
fancy-dabbrev.el:295: All variables and subroutines might as well have a documentation string
Warning (emacs): 
fancy-dabbrev.el:310: All variables and subroutines might as well have a documentation string
Warning (emacs): 
fancy-dabbrev.el:314: All variables and subroutines might as well have a documentation string
Warning (emacs): 
fancy-dabbrev.el:328: All variables and subroutines might as well have a documentation string
Warning (emacs): 
fancy-dabbrev.el:331: All variables and subroutines might as well have a documentation string
Warning (emacs): 
fancy-dabbrev.el:335: All variables and subroutines might as well have a documentation string
Warning (emacs): 
fancy-dabbrev.el:342: All variables and subroutines might as well have a documentation string
Warning (emacs): 
fancy-dabbrev.el:356: All variables and subroutines might as well have a documentation string
Warning (emacs): 
fancy-dabbrev.el:364: All variables and subroutines might as well have a documentation string
Warning (emacs): 
fancy-dabbrev.el:374: All variables and subroutines might as well have a documentation string
Warning (emacs): 
fancy-dabbrev.el:400: All variables and subroutines might as well have a documentation string
Warning (emacs): 
fancy-dabbrev.el:423: All variables and subroutines might as well have a documentation string
Warning (emacs): 
fancy-dabbrev.el:430: All variables and subroutines might as well have a documentation string
```

```
No issues found.
```

```
fancy-dabbrev.el:276:9:Warning: looking-back called with 1 argument, but
    requires 2-3

In fancy-dabbrev--post-command-hook:
fancy-dabbrev.el:318:14:Warning: reference to free variable
    `fancy-dabbrev-mode'
```

purcell:
Interesting idea, and I'm a big `dabbrev` fan, but I'm wondering how this compares to:

1. `dabbrev-completion`, a built-in alternative to `dabbrev-expand`
2. `company` with `company-dabbrev` included in `company-backends`.

All things being equal, it would be preferable to avoid adding a package which duplicates more established functionality elsewhere.

jrosdahl:
Thanks for the review!

@riscy wrote:
> Before a closer look, please check the following:
> [...]
> fancy-dabbrev.el:270: All variables and subroutines might as well have a documentation string

As noted in the PR description (but maybe it was too hidden?), I had already seen those warnings and chosen to ignore them for most internal functions and variables since I didn't feel that it would contribute with much clarity. Is that OK with you?

> fancy-dabbrev.el:276:9:Warning: looking-back called with 1 argument, but requires 2-3

Fixed in https://github.com/jrosdahl/fancy-dabbrev/commit/372f8f569dc7f7d0c9216550fcdc3b42b4849570, thanks. I don't know why that one slipped through.

> In fancy-dabbrev--post-command-hook:
> fancy-dabbrev.el:318:14:Warning: reference to free variable `fancy-dabbrev-mode'

Fixed in https://github.com/jrosdahl/fancy-dabbrev/commit/664094a62d1fa8a6451ad56222bb3396865cd7a6 I hope. I don't get that warning with Emacs 26.1 so I can't check if it works though.

@purcell wrote:
> Interesting idea, and I'm a big dabbrev fan, but I'm wondering how this compares to:
>
> 1. dabbrev-completion, a built-in alternative to dabbrev-expand
> 2. company with company-dabbrev included in company-backends.
>
> All things being equal, it would be preferable to avoid adding a package which duplicates more established functionality elsewhere.

Right, good question. Here's an attempt of explaining why I wrote my own package:

For reference, here's how `dabbrev-expand` works:

1. The user enters a string prefix.
2. The user presses the expansion key.
3. If the user is happy with the expansion, the user selects the candidate by _not pressing the expansion key again_.
4. If the user wants the next candidate, the user _presses the expansion key again_.

I very much like this behavior, probably since I've grown accustomed to it after using `dabbrev-expand` for like 25 years. It only uses one key for selection and no extra key press is needed to accept a candidate.

Unless I've missed some configuration option, `dabbrev-completion` works like this when completing:

1. The user enters a string prefix.
2. The user presses the expansion key.
3. A new window is opened, showing the candidates. The user can then switch to the other window, find a candidate and then press RET to select it (or use the mouse, which I'm not interested in doing). Or if using something like ivy/counsel (which I do), the candidate list is shown in the context next to the cursor and the user can select a candidate with C-n/C-p or down/up arrow keys and pressing RET.

In other words: The user selects a candidate by navigating to the candidate with movement commands and then pressing RET. I think that this is much less convenient compared to how `dabbrev-expand` works; I like being able to press the same key again to go to the next candidate. Also, if the first candidate is the one I want, which often is the case, I'm all done when using `dabbrev-expand` while I have to press an extra RET with `dabbrev-completion`.

`company` behaves similar to `dabbrev-completion` with ivy/counsel except that it pops up the menu of candidates "spontaneously" after typing some characters. I think that this is too intrusive as it hijacks the down/up movement keys – if I write some string and stop writing then the menu appears, and then the down/up keys suddenly no longer mean moving the cursor but choosing between candidates.

`fancy-dabbrev` keeps the basic `dabbrev-expand` behavior of using the same key for the initial expansion and for selecting subsequent candidates, and it requires no key to accept the candidate. It also behaves like `company` in the sense that it displays the first candidate "spontaneously" but in a less intrusive way: if you don't like the candidate, just do nothing.

By the way, I don't think that the name "fancy-dabbrev" is particularly great; I welcome suggestions of better names.

manuel-uberti:
FWIW, to stop `company` from automatically popping up the menu you can set `company-idle-delay` to `nil`, and then add a key binding for `company-complete` to show the menu only when you need it.

jrosdahl:
Come to think of it, I wrote [iflipb](https://melpa.org/#/iflipb) many years ago for switching between buffers with a somewhat similar model:

* One key for selecting the first candidate.
* The same key for selecting the next candidate.
* A list of the next and previous candidates is presented.
* Immediate feedback of what the result would if the candidate is accepted.
* Accept a candidate by _not pressing the key again_.

Then somebody [added it to Melpa](https://github.com/melpa/melpa/commit/db14d02c098c9a58e199151216afe70f1e15d5e5) :slightly_smiling_face:, so I thought that fancy-dabbrev might be useful for others as well.

> FWIW, to stop company from automatically popping up the menu you can set company-idle-delay to nil, and then add a key binding for company-complete to show the menu only when you need it.

Right, that's a step towards the model I like, but only a small step.

jrosdahl:
As of https://github.com/jrosdahl/fancy-dabbrev/commit/3959a383b81fae7b0e57d63a6b50ecfe07255a32, fancy-dabbrev.el is checkdoc clean.

riscy:
The only nit I have with this that you're defining faces that may be incompatible with a user's current theme.  Someone might have a background near "#777", for example, in which case the preview face will disappear.  It may be best to simply :inherit some faces and thereby have implicit support for your package using all of the existing color themes.

Hope that's helpful.  Beyond that, this is looking good to me, so ping when you're ready.

jrosdahl:
> It may be best to simply :inherit some faces

Thanks, done in jrosdahl/fancy-dabbrev@b40dc6100a2c.

riscy:
Great.  I think you make a good case for inclusion (UI is super important) so I'll go ahead and merge.  Thanks!

PR #6383, opened 2019-08-17
Add auto-clang-format

jrosdahl:

auto-clang-format provides is a minor mode (auto-clang-format-mode) that runs clang-format-buffer before the buffer is saved to its file.

https://github.com/jrosdahl/auto-clang-format

I'm the maintainer.

N/A

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/auto-clang-format
+++ b/recipes/auto-clang-format
@@ -1,0 +1,1 @@
+(auto-clang-format :repo "jrosdahl/auto-clang-format" :fetcher github)


riscy:

```
Warning (emacs): 
auto-clang-format-mode.el:84: Lisp symbol `clang-format' should appear in quotes
Warning (emacs): 
auto-clang-format-mode.el:89: All variables and subroutines might as well have a documentation string
```
> I chose to ignore some suggestions about writing documentation for internal functions, though.

For now, please do your best to document internal functions.

No byte-compilation issues! 😎

purcell:
Thanks! I feel like this minor mode is a little bit backwards, because the user enables it, and then it conditionally does stuff. Better would be to provide a mechanism where the minor mode only gets enabled when it is applicable, and then operates unconditionally. For example, a user could add a lambda to `rust-mode-hook` which enables the minor mode if a `.clang-format` file is found, and I feel like this would be more idiomatic.

jrosdahl:
Interesting. When I renamed the branch GitHub closed the PR, which wasn't what I intended. I'll keep the old branch name instead then.

jrosdahl:
Thanks for the review!

@riscy wrote:
> Before a closer look, please check the following:
> [...]
> auto-clang-format-mode.el:84: Lisp symbol `clang-format' should appear in quotes

That's a false positive warning since `.clang-format` refers to a filename, not a Lisp symbol.

> auto-clang-format-mode.el:89: All variables and subroutines might as well have a documentation string
> [...]
> For now, please do your best to document internal functions.

As you already have guessed I don't agree that adding redundant documentation strings for small internal functions makes the code better. But sure, I've made a try now in https://github.com/jrosdahl/auto-clang-format/commit/637aa9a71862ddc9a400315d3ffd6cdbe99b9d9a.

@purcell wrote:
> I feel like this minor mode is a little bit backwards, because the user enables it, and then it conditionally does stuff. Better would be to provide a mechanism where the minor mode only gets enabled when it is applicable, and then operates unconditionally. For example, a user could add a lambda to rust-mode-hook which enables the minor mode if a .clang-format file is found, and I feel like this would be more idiomatic.

Right, I can see why you think so. I thought that it was a good idea to perform the detection at each save because the situation can have changed since the mode was activated, but it would probably be just as good in practice to tie the detection to mode activation. It's just that I don't have a good feeling for how to do that in an idiomatic and easy to use way.

I made an attempt in https://github.com/jrosdahl/auto-clang-format/commit/5c42a8e35fcc7af27712f2306c19143d06f17303. I'm not 100% happy with the name `auto-clang-format-enable-if-appropriate`. Adding `-mode` somewhere in the name would make it feel even clunkier. Do you have any suggestion here?

jrosdahl:
I just found [`clang-format+`](https://melpa.org/#/clang-format%2B) which wasn't around when I wrote auto-clang-format-mode. I'll join efforts with that package instead!

riscy:
Awesome!  That's always great when that can happen.

PR #6384, opened 2019-08-18
Add with-proxy

twlz0ne:

Temporarily enable proxy for evaluating expressions:

``` elisp
(with-proxy
  ...)
```

https://github.com/twlz0ne/with-proxy.el

Author & Maintainer

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/with-proxy
+++ b/recipes/with-proxy
@@ -1,0 +1,3 @@
+(with-proxy
+ :fetcher github
+ :repo "twlz0ne/with-proxy.el")


riscy:
I can understand the desire to write human-readable function names, but Emacs doesn't have a way to protect namespaces beyond how we name functions -- and the number of packages continues to grow.  If someone created a package called `with-shell` (for example) you'd be at risk of colliding with their namespace, unfortunately. :(  I'd lean toward keeping package-lint happy.

Even though `checkdoc` doesn't complain about missing docstrings on `cl-defmacro` functions, it's still worthwhile to document these functions -- they are after all the main entrypoints in your package.

A couple very minor things I can point out:
- with-proxy.el#L83: It's safer to sharp-quote function names; use `#'`
- paire -> pair

Hopefully that's helpful.

twlz0ne:
@riscy Thanks for your review

I have fixed all the issues you pointed out in https://github.com/twlz0ne/with-proxy.el/commit/787b9af8c75684415148d620eedb593e89c3f4bd, please check again.

riscy:
Thanks!  This looks great now.

PR #6387, opened 2019-08-18
Add stock-tracker

beacoder:
stock-tracker enable users to track chinese stocks prices in emacs with json-api provided by netease company.

https://github.com/beacoder/stock-tracker

I'm the maintainer.

N/A

### Checklist

Please confirm with `x`:

- [x ] I've read [CONTRIBUTING.org](https://github.com/melpa/melpa/blob/master/CONTRIBUTING.org)

### DIFF
--- a/recipes/stock-tracker
+++ b/recipes/stock-tracker
@@ -1,0 +1,1 @@
+(stock-tracker :repo "beacoder/stock-tracker" :fetcher github)


alphapapa:
Since the package is only for Chinese stocks using a certain proprietary API, the name should be descriptive.  `stock-tracker` implies much more general usefulness.

beacoder:
@alphapapa currently it did only utilize a certain proprietary API which used for chinese stock market.
But in the future, I may add more open APIs to retrieve data from stocks which may come from other markets, so I think the name is appropriate.
thanks.

riscy:
Just a quick check since these packages might be quite different, but any relationship (or possible opportunity for mutual benefit) with https://github.com/hagleitn/stock-ticker?

beacoder:
@riscy , nope, no relationship with this stock-ticker.
the stock-tracker is based on net-ease open api for tracking stocks listed CHN and US.
in stock-tracker, there's a dedicated buffer showing stocks.
seems in stock-ticker, it just show stock information in mode-line.

riscy:
Taking a closer look at this now.

Personally I'm okay with the name of the package (`stock-ticker`, after all, accesses Yahoo's API, and that appears to be US-only at a quick glance).  More descriptive is almost always better, so I'd be thrilled if this package could be renamed `netease-stocks` or similar.  But that's not a blocker.

However, I think it would be better if stock-ticker's package description mentioned it was US stocks -- and by that same token, I think it would be good if `stock-tracker`'s description mentioned it was Chinese stocks.  This will help people searching for packages know what they're installing.

To be equivocal about this, I've raised an issue at the `stock-ticker` author's repository: https://github.com/hagleitn/stock-ticker/issues/6

Some other miscellaneous advice I have:

- stock-tracker.el#L238: It's safer to sharp-quote function names; use `#'`
- stock-tracker.el#L263: It's safer to sharp-quote function names; use `#'`
- stock-tracker.el#L384: It is beneficial to sharp-quote hooks; use `#'`
- stock-tracker.el#L25: Prefer `https` over `http` (if possible)
- stock-tracker.el#L71: Prefer `https` over `http` (this one seems like it might be a very good idea if possible)

Minor:
- `error-reponse` -> `error-response`
- `moniter` -> `monitor`

beacoder:
Thanks for these great advices, I will take time to go over these one by one.
thanks a lot.

beacoder:
@riscy As I have updated in the https://github.com/beacoder/stock-tracker/blob/master/README.org
This package now support US stocks as well.
Please see the note region.
Thanks.

beacoder:
@riscy I have fixed all 7 issues you mentioned, thanks a again.

riscy:
Great, thanks! :)

PR #6392, opened 2019-08-20
Add rbtagger

thiagoa:

A ctags-based Emacs utility to index Ruby projects along with gems. It aims to provide context-aware, accurate tag lookup by parsing the current Ruby file and an easy-to-use tags solution that works out of the box.

https://github.com/thiagoa/rbtagger

I am the maintainer.

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/rbtagger
+++ b/recipes/rbtagger
@@ -1,0 +1,4 @@
+(rbtagger
+ :fetcher github
+ :repo "thiagoa/rbtagger"
+ :files (:defaults "bin"))


thiagoa:
@riscy Thank you for the feedback! Everything has been addressed [here](https://github.com/thiagoa/rbtagger/pull/11).

Regarding the free variable warning, I wasn't able to see it back when I opened the PR, probably because `enh-ruby-mode` is installed here and therefore `(require 'enh-ruby-mode nil 'noerror)` succeeds.

riscy:
Great.  And, welcome to MELPA. :)

PR #6394, opened 2019-08-21
Add howdoyou

thanhvg:

Searches your query all accross stackoverflow and it’s sisters’ sites. They are:
    stackoverflow.com
    stackexchange.com
    superuser.com
    serverfault.com
    askubunu.com

The result is then showed in an `org-mode` buffer.

https://github.com/thanhvg/emacs-howdoyou

I am the maintainer.

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/howdoyou
+++ b/recipes/howdoyou
@@ -1,0 +1,1 @@
+(howdoyou :fetcher github :repo "thanhvg/emacs-howdoyou" :branch "master")


riscy:
```
howdoyou.el:242:1:Warning: Unused lexical variable `i'
howdoyou.el:242:1:Warning: Unused lexical variable `i'
```
These look like false positives.  You can (and should) silence the compiler so that the warnings don't confuse users installing the package -- one way would be to rewrite the function body as:
```elisp
  (let* ((answer-nodes (dom-by-class (cdr result) "answercell"))
         (question-dom (car (dom-by-id (cdr result) "^question$")))
         (title (car (dom-by-class (cdr result) "question-hyperlink")))
         (number-of-answers (min (length answer-nodes) howdoyou-number-of-answers))
         (tags (howdoyou--get-so-tags (cdr result)))
         (score-nodes (dom-by-class (cdr result) "js-vote-count"))
         acc
         scores)
    (dotimes (i number-of-answers)
      (setq acc (append acc (dom-by-class (nth i answer-nodes) "post-text"))))
    (dotimes (i (1+ number-of-answers))
      (setq scores (append scores `(,(dom-text (nth i score-nodes))))))
    (list (car result)
          (dom-text title)
          (dom-by-class question-dom "post-text")
          acc
          scores
          tags)))
```
but the function feels more complicated than it needs to be.

You should `(require 'subr-x)` for `if-let` (rather than rely on transitive imports).  Similarly you should `(require 'url-http)`, 

And some minor stuff:
- howdoyou.el#L198: Consider `unless ...` instead of `when (not ...)`
- howdoyou.el#L21: Prefer `https` over `http` (if possible)
- Try to be precise with your indentation, e.g. [here](https://github.com/thanhvg/emacs-howdoyou/blob/2915c4815abc6b3e708d66a396e60b7277a769b9/howdoyou.el#L375-L378), since it can obfuscate the scope of your expressions.  (Emacs should be able to indent elisp code consistently.)
- accross -> across
- Roate -> ?
- shawdow -> shadow
- insead -> instead
- Othewise -> Otherwise

thanhvg:
Thanks so much for the review.
 
> And some minor stuff:
>     * howdoyou.el#L198: Consider `unless ...` instead of `when (not ...)`

Fixed, thanks

>     * howdoyou.el#L21: Prefer `https` over `http` (if possible)
>     * accross -> across
>     * Roate -> ?
>     * shawdow -> shadow
>     * insead -> instead
>     * Othewise -> Otherwise

I have fixed these typos. Sorry for my bad English

>     * Try to be precise with your indentation, e.g. [here](https://github.com/thanhvg/emacs-howdoyou/blob/2915c4815abc6b3e708d66a396e60b7277a769b9/howdoyou.el#L375-L378), since it can obfuscate the scope of your expressions.  (Emacs should be able to indent elisp code consistently.)

Indeed, the code is indented correctly and always done automatically by Emacs. It is a promise-chain macro and the first parameter of the chain is completely different to the rest that why it is indented different to the rest. Anyway I move it to the same line with promise-chain to avoid confusion.

> These look like false positives. You can (and should) silence the compiler so that the warnings don't confuse users installing the package -- one way would be to rewrite the function body as:
> 
> ```emacs-lisp
>   (let* ((answer-nodes (dom-by-class (cdr result) "answercell"))
>          (question-dom (car (dom-by-id (cdr result) "^question$")))
>          (title (car (dom-by-class (cdr result) "question-hyperlink")))
>          (number-of-answers (min (length answer-nodes) howdoyou-number-of-answers))
>          (tags (howdoyou--get-so-tags (cdr result)))
>          (score-nodes (dom-by-class (cdr result) "js-vote-count"))
>          acc
>          scores)
>     (dotimes (i number-of-answers)
>       (setq acc (append acc (dom-by-class (nth i answer-nodes) "post-text"))))
>     (dotimes (i (1+ number-of-answers))
>       (setq scores (append scores `(,(dom-text (nth i score-nodes))))))
>     (list (car result)
>           (dom-text title)
>           (dom-by-class question-dom "post-text")
>           acc
>           scores
>           tags)))
> ```
> 
> but the function feels more complicated than it needs to be.

I didn't know that we need to make the linter happy at all cost.  But your suggestion didn't work neither. I made it work by using `mapcar` instead of `dolist`

> You should `(require 'subr-x)` for `if-let` (rather than rely on transitive imports). Similarly you should `(require 'url-http)`,

I am not sure I understand what you meant. Can you elaborate? I require `promise` and `request` so I can have good abstraction layer to work on these low level apis. And if promise and request fail  on those apis, explicit require them won't help either and I would rather my code crash as expected

riscy:
> Indeed, the code is indented correctly and always done automatically by Emacs. It is a promise-chain macro and the first parameter of the chain is completely different to the rest that why it is indented different to the rest. Anyway I move it to the same line with promise-chain to avoid confusion.

Sorry about the misfire. :)

> I didn't know that we need to make the linter happy at all cost.

If we can minimize the number of warnings that scroll by when people are byte-compiling their installed packages, it's for the best, I think. Some byte-compile warnings actually point to errors that need attention. Thanks for looking into that!

> Can you elaborate?

`if-let` is a macro defined in subr-x, so when [you use it](https://github.com/thanhvg/emacs-howdoyou/blob/master/howdoyou.el#L139), you should explicitly import subr-x.  subr-x is a part of Emacs and a lot of packages import it, but you should `(require 'subr-x)` at the top of your file anyway, just to be sure.  Similarl for the function url-http-parse-headers, which you use [here](https://github.com/thanhvg/emacs-howdoyou/blob/master/howdoyou.el#L169).  It's not a part of `promise` or `request`, it's a part of `url-http`, and an explicit require is a good idea

thanhvg:
It's all clear to me now with your explanation :). I added 

``` elsip
(require 'subr-x)
(require 'url)
```
My Emacs (26.2) said `url` instead of `url-http`.

Thanks

riscy:
This looks like it's running great for me and I love the execution.  One suggestion I have is to add an alias, `howdoyou`, against `howdoyou-query`, to make it even easier for people to use from <kbd>M-x</kbd>.  But that's not a blocker for merging.

Welcome to MELPA! :)

tarsius:
Why is `:branch` set explicitly in the recipe?

riscy:
Woops - I missed that.  @thanhvg, fyi, you don't have to specify a branch if you're just using master.  I'm now noticing we have a few of these (`omnisharp`, `org-babel-eval-in-repl`, `l-eval-in-repl`, `proof-general`, `telega` (which I believe I merged), and even `scala-mode`.

PR #6395, opened 2019-08-22
Add annalist.el

noctuid:
`annalist` is a package for recording information and later printing it using org-mode headings and tables. Both general.el and evil-collection use org-mode for displaying recorded keybindings. `annalist` generalizes and extracts this functionality into a library that can be used by both (and other packages).

https://github.com/noctuid/annalist.el

I am the author.

**None needed**

### Checklist

(`package-lint` and also `elisp-lint` checks pass and I've tested it in the sandbox)

### DIFF
--- a/recipes/annalist
+++ b/recipes/annalist
@@ -1,0 +1,2 @@
+(annalist :fetcher github
+          :repo "noctuid/annalist.el")


riscy:
Looks convenient.  Thanks. :)

PR #6398, opened 2019-08-24
recipes/plain-org-wiki: Add plain-org-wiki

abo-abo:

Quickly browse or create Org files.

https://github.com/abo-abo/plain-org-wiki

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/plain-org-wiki
+++ b/recipes/plain-org-wiki
@@ -1,0 +1,1 @@
+(plain-org-wiki :fetcher github :repo "abo-abo/plain-org-wiki")


riscy:

No issues! 💯

```
No issues found.
```

```
plain-org-wiki.el:63:1:Warning: Unused lexical variable `ffip-project-root'

In end of data:
plain-org-wiki.el:115:1:Warning: the function `ffip-project-search' is not
    known to be defined.
```

- plain-org-wiki.el#L23: Prefer `https` over `http` (if possible)

abo-abo:
@riscy Thanks, fixed.

riscy:
Thanks!

```
1 issue found:
1:79: warning: You should depend on (emacs "24") if you need lexical-binding.
```
Sorry about this, my message up there should mention the Emacs 24 dependency required.

```
plain-org-wiki.el:117:1:Warning: the function `ffip-project-search' is not
    known to be defined.
```
One way to address this one is to depend on `(find-file-in-project "5.7.7")` for example.  So altogether your `Package-Requires` would read like:
```elisp
;; Package-Requires: ((emacs "24") (find-file-in-project "5.7.7"))
```

Another thing I notice is that `(require 'helm)` is buried in `plain-org-wiki-helm`, but you could also soft-require it using `(require helm nil t) (declare-function helm "ext:helm.el")`

riscy:
Hi!  I saw [your commit](https://github.com/abo-abo/plain-org-wiki/commit/e246b65dfecb94758cb97493f368ea52047c5f8a#diff-109a883063be58ae8183b9d358084664R8) -- feel free to ping when you're ready for another pass.  At this very moment it looks like you still need a `require` on `find-file-in-project`.

abo-abo:
Requiring `find-file-in-project` is not essential to the functionality of the package, in my opinion.

riscy:
That's fair, but your elisp should byte-compile cleanly:
```
In end of data:
plain-org-wiki.el:118:1:Warning: the function `ffip-project-search' is not
    known to be defined.
```
and your function `plain-org-wiki-files-recursive` calls `ffip-project-search`.

abo-abo:
OK, it should now byte compile cleanly.

purcell:
This is missing a package dependency declaration [for `ivy`](https://github.com/abo-abo/plain-org-wiki/blob/5b3ea15507f0e42c4210495480fe566b398460b9/plain-org-wiki.el#L33).

> Another thing I notice is that `(require 'helm)` is buried in `plain-org-wiki-helm`, but you could also soft-require it using `(require helm nil t) (declare-function helm "ext:helm.el")`

@riscy Eeagerly loading `helm` (which might be installed, but not used) wouldn't be ideal.  So the solution here  would be the `declare-function` you suggest, and to remove the `(require 'helm)` and `(require 'helm-multi-match)` completely.

(Really, to package this properly you'd have a separate little `plain-org-wiki-helm` package with a proper dependency on `plain-org-wiki` and `helm`. These byte compilation workarounds are generally a sign that the code is doing things the wrong way. For this little case I wouldn't insist.)

abo-abo:
@purcell Thanks, added the dependency on ivy.

riscy:
Hi @abo-abo, I see the push -- I think this issue still isn't addressed:

> the solution here would be the declare-function ..., and to remove the (require 'helm) and (require 'helm-multi-match) completely.

riscy:
Currently on the master branch I see:

```elisp

;;;###autoload
(defun plain-org-wiki-helm ()
  "Select an org-file to jump to."
  (interactive)
  (require 'helm)
  (require 'helm-multi-match)
  (helm :sources
        '(((name . "Projects")
           (candidates . plain-org-wiki-files)
           (action . plain-org-wiki-find-file))
          ((name . "Create org-wiki")
           (dummy)
           (action . plain-org-wiki-find-file)))))
```
@purcell's suggestion is to use `declare-function`, so at the top of your file you would have:
```elisp
(require 'ivy)
(declare-function helm "ext:helm.el" t)
```
Another small thing I'm noticing with the latest `package-lint` -- you want to depend on Emacs 24.3 rather than Emacs 24.1 for `cl-mapcan`.

After these I think everything's sorted and this is good to merge.

abo-abo:
@riscy Thanks.

riscy:
Great, thanks!

abo-abo:
Thanks.

PR #6403, opened 2019-08-25
add frecentf

FelipeLema:

Pervasive recentf using frecency

https://launchpad.net/frecentf.el

Maintainer

Versions of `persist` previous to 0.4 do not work (cannot persist a `hash-table`). I had real trouble trying to make sure this is the version that I was using (I have less trouble installing stuff from melpa than from elpa).

### Checklist

Please confirm with `x`:

- [x] I've used the latest version of [package-lint](https://github.com/purcell/package-lint) to check for packaging issues, and addressed its feedback... although I use `straight` so I may have [missed something](https://github.com/purcell/package-lint/issues/108).

### DIFF
--- a/recipes/frecentf
+++ b/recipes/frecentf
@@ -1,0 +1,3 @@
+(frecentf :fetcher git
+	  :url "https://git.launchpad.net/frecentf.el"
+	  :commit "origin/master")
--- a/recipes/frecentf
+++ b/recipes/frecentf
@@ -1,3 +1,2 @@
 (frecentf :fetcher git
-	  :url "https://git.launchpad.net/frecentf.el"
-	  :commit "origin/master")
+	  :url "https://git.launchpad.net/frecentf.el")


riscy:

No issues! 💯

```
No issues found.
```

```
frecentf.el:128:23:Warning: reference to free variable `frecency-struct'
frecentf.el:126:58:Warning: reference to free variable `path'
frecentf.el:145:1:Warning: Unused lexical variable `ivy-sort-functions-alist'

In frecentf--pick-by:
frecentf.el:153:48:Warning: reference to free variable `frecency-struct'
frecentf.el:151:69:Warning: reference to free variable `path'

In end of data:
frecentf.el:212:1:Warning: the following functions are not known to be
    defined: assert, multiple-value-bind, path
```
The vast majority of these are fixed by using `cl-assert` and `cl-multiple-value-bind` functions instead of their deprecated counterparts.

FelipeLema:
now fixed upstream

had originally done `byte-compile-file` on fully loaded Emacs config and didn't get those warnings. Did byte compile with `emacs -Q` and loading only `straight.el` and requires packages and got the warnings.

Thanks for pointing them out.

alphapapa:
Hi,

I'm glad to see another package based on `frecency`.  :)  Here's a little feedback:

-   `frecentf-do-not-add-under-these` seems to have the wrong docstring.
-   `frecentf-do-not-add-under-these` is also kind of awkwardly named.  I'd recommend something like `frecentf-ignore-paths`.
-   `frecentf-add-path` has a spelling error in its docstring.
-   The commentary says, "On each buffer added, its directory is also added."  I recommend making this optional, disabled by default, because it's not what most users will expect.
-   `frecentf--add-entry` has a `∈` in it.  It's better to use plain English to describe things like that than mathematical symbols, because not everyone is familiar with them.
    -   The asserts in that function could be simply, e.g. `(member type-of-path '(file dir))`.
-   You may find the `ht` library helpful for working with hash tables.
-   Instead of `frecentf--table-as-list`, you could use `map-into`.
-   `frecentf--pick-by` has `ivy`-related variables mentioned in it.  You should generally avoid referring to variables from packages you don't actually use.
-   You add functions to two hooks. Those should be added and removed in a minor mode.  Simply loading your package's file shouldn't modify any hooks.
-   You use `throw` in `frecentf--pick-by` without a `catch`, and `condition-case` in `frecentf-pick-file` as a sort of implied catcher.  The manual seems to indicate that this is possible, but I think it's generally unidiomatic Elisp.  I don't think it's necessary, anyway, just to display a message in case the user changes his mind and doesn't choose a file.

recipes/frecentf
```diff
@@ -0,0 +1,3 @@
+(frecentf :fetcher git
+	  :url "https://git.launchpad.net/frecentf.el"
+	  :commit "origin/master")
```
riscy: I think we should remove this line.  This looks like a branch name, and even then, MELPA will take `master` by default.

FelipeLema: I guess you meant the "commit" property (github highlights the three lines)

FelipeLema:
all remarked issues are resolved

recipes/frecentf
riscy: Yep!  Thanks.

riscy:
@alphapapa, thanks a ton, that was thorough.  @FelipeLema, this is looking ready to go.  Thanks!

Silex:
@FelipeLema: I made a bug report at https://bugs.launchpad.net/frecentf.el/+bug/1848294, just curious can I know why you host it on launchpad instead of github?

FelipeLema:
the main reason is that, AFAIK, launchpad does not deny services to US-sanctioned countries (Cuba, Iran, probably Venezuela too).

FelipeLema:
yes, they do... see https://gitlab.com/gitlab-com/support-forum/issues/3795

Silex:
That's really a silly move from the US.

PR #6408, opened 2019-08-28
add balanced-windows.el

wbolster:

Automatically keep windows balanced.

https://github.com/wbolster/emacs-balanced-windows

Author.

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/balanced-windows
+++ b/recipes/balanced-windows
@@ -1,0 +1,3 @@
+(balanced-windows
+ :fetcher github
+ :repo "wbolster/emacs-balanced-windows")


riscy:
I'm surprised there's nothing on MELPA that does exactly this.  Have you compared to [zoom](https://melpa.org/#/zoom) for example?  A couple quick notes:

```
balanced-windows.el:43:11:Warning: assignment to free variable
    `evil-auto-balance-windows'
```

- balanced-windows.el#L38: Advice functions should be sharp-quoted; use `#'`

wbolster:
i am not aware of the proper way to get rid of that warning. i tried:

- a hard dependency on `evil`: not desirable since this has nothing to do with `evil`.

- adding a `defvar`, gets rid of the warning, but introduces another one about defining variables with the wrong prefix.

wbolster:
it seems `zoom.el` is based on the idea that the selected window is special and gets automatic resizing (which is configurable), and all other remaining windows get balanced automatically. this can result in quite some automatically moving content on the screen when selecting another window, or when resizing the frame, or when dragging the window sizes manually using the mouse.

my (way simpler) `balanced-windows.el` aims for the opposite: make the windows layout as stable as possible, minimising the amount of movement on the screen. it does not perform automatic resizing all the time, by design. this also means that manual resizes (e.g. to have a better view on a wide buffer) stay intact... until the window configuration changes, after which everything will be perfectly balanced again.

some background: i am very sensitive to "flicker", moving things on the screen, or generally large screen updates. this package makes the screen layout extremely stable and predictable (or "boring"), which helps me a lot since i don't get distracting visual stimuli. i use a 2, 3 or 4 column window layout 100% of the time (depending on monitor size), and this package keeps my screen completely stable. i have been using this for years but recently moved this code out of my `init.el` into its own package after a fellow emacs user expressed an interest in this behaviour.

riscy:
That sounds fair to me.  I was actually surprised to find that there are no other packages (after some Googling) that do this very thing.  Let me welcome this package to MELPA. :)

wbolster:
thanks! tips for getting rid of the warning still welcome btw.

i was also surprised i couldn't find a package providing a simple feature like this. without the boilerplate that i added to turn this into a package, this lived as a couple of lines of lisp (without a toggle) in my `init.el`. but it seems we agree this is non-trivial enough to warrant its own package.

riscy:
Sorry about that, I forgot to try to byte-compile your code and merged prematurely, which has an error in it now.  This should fix it: https://github.com/wbolster/emacs-balanced-windows/pull/2

wbolster:
> tips for getting rid of the warning still welcome btw.

thanks for teaching me that 1-arg `defvar` is a thing. 👍

PR #6411, opened 2019-08-31
add asx recipe

ragone:

This package allows you to search any StackExchange site specified in `asx-sites` and insert the top post in an Org-mode buffer. For better searching, this uses Google and DuckDuckGo as it is superior to StackExchange’s searching capabilities. More search engines can be defined using `asx-search-engine-alist`.

example

https://github.com/ragone/asx

Maintainer and author

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/asx
+++ b/recipes/asx
@@ -1,0 +1,1 @@
+(asx :repo "ragone/asx" :fetcher github)


riscy:
I really like this!  A couple quick notes and then this looks ready to merge to me:

- asx.el#L7: Use `Version` instead of `Package-Version` (MELPA automatically adds `Package-Version`)
- It seems to me the functions `asx-next-post` through `asx-go-back-to-first-post` do not need the autoload cookie, since they aren't initial entrypoints in the same way that `asx` is.  See [when to autoload](https://www.gnu.org/software/emacs/manual/html_node/elisp/When-to-Autoload.html).

ragone:
Thank you for your feedback @riscy.
I have updated master with the changes you suggested.

riscy:
Looks great, thanks!  And, welcome to MELPA. :)

PR #6412, opened 2019-08-31
Add recipe for diminish-buffer.

jcs090218:

Diminished buffers from being shown in buffer menu.

https://github.com/jcs090218/diminish-buffer

The maintainer.

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/diminish-buffer
+++ b/recipes/diminish-buffer
@@ -1,0 +1,1 @@
+(diminish-buffer :repo "jcs090218/diminish-buffer" :fetcher github)


alphapapa:
The function `diminish-buffer--is-contain-list-string-regexp` is only used in one place and calls a single function.  Why not inline it in `diminish-buffer-clean` to avoid polluting the obarray?

jcs090218:
Is just one of my coding habit, I think have it this way is more human readable!

alphapapa:
It's less readable, because I don't know what that function does until I look it up.  In contrast, what the `(cl-some ...` call does is obvious.  But it's your code.

jcs090218:
All good, I just have the habit having the function’s name express on its own. Hence, is-contain-list string-regexp explain on its own.

alphapapa:
A function name like `list-matches-regexp-p` would be easier to understand and more idiomatic.  But since it's essentially a one-line function, it would be preferable to inline it.  If it needs explanation, a comment would suffice.

riscy:
This looks good.  The only thing I would suggest is to change your package short-description (first line of your elisp file) from: "Diminished buffers from being shown in buffer menu." to something like: "Prevent buffers from being shown in buffer-menu." or "Diminish (hide) buffers from buffer-menu."

(Writing it as "buffer-menu" makes it clearer that we're referring to the `buffer-menu` elisp command so people scanning the package list know exactly where it applies).

That's not a blocker, but left for your consideration.

PR #6413, opened 2019-09-02
Add the new package rev-mode

gaballench:

`rev-mode` is an IDE package for carrying out analyses with the bayesian software [revbayes](https://revbayes.github.io/). It allows text redirection for script evaluation with line, selection, and buffer redirection capabilities. 

https://github.com/gaballench/rev-mode

Maintainer: Gustavo A. Ballen

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/rev-mode
+++ b/recipes/rev-mode
@@ -1,0 +1,1 @@
+(rev-mode :fetcher github :repo "gaballench/rev-mode" :branch "release")


gaballench:
Hello @riscy. Thanks for your comments. I noted these warnings via flycheck but did not think these were somewhat serious and thus did not try to figure out how to correct them. Since there  was a lot of input from your checks, I want to get through it by pieces.

First, I am considering tom move the add-hook sentence in l.285 to the install instructions so that the user puts that specific line in their .emacs file. That way, loading the package would not have that unwanted behavior as far as I can see. What do you think?

The reason I was including that line in the code is that my package is in a sense a minor mode derived from sh-mode that allows to open a process of the program revbayes as a shell process and then allows text redirection to the buffer with such process. Also, as far as I can see, the package is only loaded when a .rev file is opened and thus only "alters" the behavior of Emacs on load of such a file, but I can be wrong so any suggestion is very welcomed.

Thanks,

Gustavo

> Double-checking the checklist:
> 
> M-x checkdoc (using version 0.6.1):
> 
> ```
> Warning (emacs): 
> rev-mode.el:279: All variables and subroutines might as well have a documentation string
> ```
> 
> I am just looking quickly but I also notice, underneath this function, the line `(add-hook 'sh-mode-hook 'rev-mode-sh-hook)`. Simply loading a package shouldn't alter Emacs's behavior -- toggling keybindings should be the responsibility of minor modes and major modes.

riscy:
> The reason I was including that line in the code is that my package is in a sense a minor mode derived from sh-mode that allows to open a process of the program revbayes as a shell process and then allows text redirection to the buffer with such process.

It may be a good idea to define a minor mode in this case!  Mentioning the hook in the install instructions is a good idea, and much better than adding a hook that [modifies the user's keybindings](https://github.com/gaballench/rev-mode/blob/master/rev-mode.el#L278-L285) (and potentially overwrites their customizations).

tarsius:
Closing due to inactivity.

PR #6414, opened 2019-09-02
Add libbasecampel

DamienCassou:

The package is an Emacs library to interact with Basecamp
(https://basecamp.com), a project management web application. The
package depends on an external proxy implemented in JS (the same way
Indium is implemented https://melpa.org/#/indium).

https://gitlab.petton.fr/basecampel/libbasecampel

Author

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/libbasecampel
+++ b/recipes/libbasecampel
@@ -1,0 +1,1 @@
+(libbasecampel :fetcher github :repo "DamienCassou/libbasecampel")


riscy:
Great.  One suggestion for your consideration is that in libbasecampel-proxy.el#L51, it strikes me that customizable variables shouldn't be "private".

DamienCassou:
> it strikes me that customizable variables shouldn't be "private"

fixed. Thank you.

riscy:
Thanks!

PR #6415, opened 2019-09-03
Update url for helm-fuzzy.

jcs090218:
I moved this repo to under an organization, update the url. :D

### DIFF
--- a/recipes/helm-fuzzy
+++ b/recipes/helm-fuzzy
@@ -1,1 +1,1 @@
-(helm-fuzzy :repo "jcs090218/helm-fuzzy" :fetcher github)
+(helm-fuzzy :repo "elpa-host/helm-fuzzy" :fetcher github)


riscy:
Just out of curiosity, why don't these packages start out in the organization? :)

jcs090218:
Oh, it's just my workflow. If the package doesn't get merged then I might transfer the package to somewhere else. :0

tarsius:
> If the package doesn't get merged then I might transfer the package to somewhere else.

By "doesn't get merged" do you mean "is not added to Melpa because the maintainers reject it"?

Where would you transfer it to?

> Oh, it's just my workflow.

It sounds like this workflow makes things more complicated for Melpa maintainers, who now have to figure out (and optionally manage to remember for a long time) what you are doing and why.

Could your workflow be adjusted so that it doesn't involve two pull-requests if things go well? I.e. please consider going with `elpa-host/PACKAGE` right away.

tarsius:
On #6445 @jcs090218 said:

>  Starting from next package I will start the package from organization.

PR #6418, opened 2019-09-03
Add helm-fd.el

lerouxrgd:

Same as `helm-find` but using [fd](https://github.com/sharkdp/fd) instead of `find`.
`fd` is written in Rust and is much faster than `find`.
There is also a `helm-fd-project` function to find files in VCS projects.

https://github.com/lerouxrgd/helm-fd

I am the author.

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/helm-fd
+++ b/recipes/helm-fd
@@ -1,0 +1,2 @@
+(helm-fd :fetcher github
+         :repo "lerouxrgd/helm-fd")


riscy:
Looks good, and thanks for sticking to the checklist.  Non-blocking suggestions:

You may want to raise an error or info of some kind if `helm-fd-bin` turns out to be nil (e.g. the executable was not found).  I figure if users are savvy enough to have `fd` installed they'll be able to figure this out, but just in case.

One thing I also suggest is to include a LICENSE file in the repo (MELPA will exclude it) -- as a bonus, GitHub will give your repo a nice icon that says "GPL 3.0" (or whatever).  I'd also recommend linking to http://www.gnu.org/licenses/ using https in your front-matter (in an effort to modernize the copy-paste GPL summary everyone uses).

riscy:
...and, welcome to MELPA. :)

lerouxrgd:
Thank you!

PR #6419, opened 2019-09-03
Add swift-helpful recipe

danielmartin:

swift-helpful is a package that includes curated Swift documentation from Apple in GNU Info format
(original license is Creative Commons: https://swift.org/documentation/).

Invoke `M-x swift-helpful-configure-manuals` first to make sure Emacs can search the installed GNU Info manuals using `info-look`.

After that, if you open a `.swift` file, place the point on a keyword and invoke `M-x swift-helpful`, you will see a documentation snippet explaining the programming concept represented by that keyword.

If lsp-mode is configured, API documentation is also shown.

https://github.com/danielmartin/swift-helpful

I created the package.

**None needed** 

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/swift-helpful
+++ b/recipes/swift-helpful
@@ -1,0 +1,3 @@
+(swift-helpful :repo "danielmartin/swift-helpful"
+               :fetcher github
+               :files ("*.el" "swift-info/*.info"))
--- a/recipes/swift-helpful
+++ b/recipes/swift-helpful
@@ -1,3 +1,3 @@
 (swift-helpful :repo "danielmartin/swift-helpful"
                :fetcher github
-               :files ("*.el" "swift-info/*.info"))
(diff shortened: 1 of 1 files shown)


riscy:
This looks good, thanks!  A few non-blocking suggestions:
- swift-mode.el defines a customization group "swift" that you could declare as a parent of your group "swift-helpful"
- Since ripgrep won't be available by default on every machine, when you define `swift-helpful-ripgrep-executable` you may want to use `executable-find` which will return nil if the executable can't be found, and then raise appropriate warnings. (On the other hand, maybe appropriate warnings are already raised when "rg" fails...?)
- swift-helpful-regex.el#L120: It's safer to sharp-quote function names; use `#'`
- swift-helpful-regex.el#L16: Prefer `https` over `http` (if possible)

PR #6420, opened 2019-09-04
add recipe for lsp-julia

gdkrmr:

Julia support for `lsp-mode`.

https://github.com/non-jedi/lsp-julia

Maintainer of the package

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/lsp-julia
+++ b/recipes/lsp-julia
@@ -1,0 +1,5 @@
+(lsp-julia :fetcher github
+           :repo "gdkrmr/lsp-julia"
+           :files ("*.el"
+                   "README.org"
+                   "languageserver"))
--- a/recipes/lsp-julia
+++ b/recipes/lsp-julia
@@ -1,5 +1,3 @@
 (lsp-julia :fetcher github
            :repo "non-Jedi/lsp-julia"
-           :files ("*.el"
-                   "README.org"
-                   "languageserver"))
+           :files (:defaults "languageserver"))
--- a/recipes/lsp-julia
+++ b/recipes/lsp-julia
@@ -1,5 +1,5 @@
 (lsp-julia :fetcher github
-           :repo "gdkrmr/lsp-julia"
+           :repo "non-Jedi/lsp-julia"
            :files ("*.el"
                    "README.org"
                    "languageserver"))


gdkrmr:
One thing: This package does not call any functions from `julia-mode` but `julia-mode` is required for `lsp-mode` to be able to start the server so I added it as a dependency in order to be able to test it with `make sandbox`.

riscy:
> One thing: This package does not call any functions from julia-mode but julia-mode is required for lsp-mode to be able to start the server so I added it as a dependency in order to be able to test it with make sandbox.

On the one hand, if it's only strictly required for testing, it might be prudent to exclude it.  On the other, it seems like this mode can't be expected to function without julia-mode (is that correct?), so I think what you said above is fine.

non-Jedi:
In the github comment, you link to my fork, but the actual recipe links to yours. I'm don't have any particular issue with either, but you should be consistent there. There may be some advantages to keeping mine as canonical to avoid link rot.

gdkrmr:
> this mode can't be expected to function without julia-mode (is that correct?)

`lsp-mode` starts the server according to the major mode of the file, so I guess this is a hard dependency. If you work around this, you could probably make it run without `julia-mode` but I guess this does not make much sense.



> In the github comment, you link to my fork, but the actual recipe links to yours. I'm don't have any particular issue with either, but you should be consistent there. There may be some advantages to keeping mine as canonical to avoid link rot.

Sorry, this was unintended.

riscy:
Great, thanks!  To me, this looks good to merge.  I have a few comments that I consider non-blocking, but please consider them.

---------

In your "manual installation" instructions, you should recommend `add-to-list` instead of `push` (the latter lets you push duplicate entries into the list), but that's not a blocker.

The function `lsp-julia--get-depot-path` looks like it could be simplified to:
```elisp
(defun lsp-julia--get-depot-path ()
  "Get the (Julia) depot path."
  (or (getenv "JULIA_DEPOT_PATH") lsp-julia-default-depot))
```

This docstring:
```elisp
"The default depot path, used if `JULIA_DEPOT_PATH' is unset."
```
..doesn't need to ``` `' ```-quote `JULIA_DEPOT_PATH` -- typically those kinds of quotes are reserved for hyperlinking Emacs symbols in the online help.  But no harm!

riscy:
I've also updated your recipe to prefer the `:defaults` (which will now exclude `README.org`).

PR #6421, opened 2019-09-06
Add mldoc recipe

zonuexe:

MLDoc is an extension of ElDoc with a function definition helper and a multiple configuration mechanism.

https://github.com/emacs-php/mldoc

I'm the author.

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/mldoc
+++ b/recipes/mldoc
@@ -1,0 +1,1 @@
+(mldoc :fetcher github :repo "emacs-php/mldoc")


riscy:
Your minor mode declaration `(define-minor-mode mldoc-mode ...` should handle toggling the mode both off and on, but at the moment only appears to handle turning the mode on.  (See [here](https://www.gnu.org/software/emacs/manual/html_node/elisp/Defining-Minor-Modes.html)).  The automatic turning on of eldoc presents an interesting issue -- should eldoc be turned (back) off when your minor mode is toggled off?  Should it remain on?  I'll leave that part up to you.

In your docstrings, the ``` `backtick/apostrophe' ```-style quoting is typically just used to wrap symbols (and in that way creates hyperlinks in the online help).  So for example this:
```elisp
  "Return propertized string by `ARGS' list, `CURRENT-ARG', `ARG-SEPARATOR' and `ARG-SPEC'."
```
Should probably be written:
```elisp
  "Return propertized string by ARGS list, CURRENT-ARG, ARG-SEPARATOR and ARG-SPEC."
```
and
```elisp
  "Return list for function `propertize' by `PROPERTIZER' alist."
```
Should be written:
```elisp
  "Return list for function `propertize' by PROPERTIZER alist."
```
(Emacs will automatically detect that these capitalized forms are part of your function signature and highlight them in the online help as well).

Hope that helps!

zonuexe:
@riscy Thank you, I fixed.

riscy:
Thanks - I've had some time to look a little more deeply at this.  Just as a general code thing, I notice I'm having trouble evaluating your buffer due to the definition of the minor mode.  It looks like if you don't pass _any_ keyword arguments (e.g. `:group` or `:lighter`) then the arguments "init-value, lighter, and keymap" must be provided.  So I'd recommend supplying `:group` and potentially a customizable `:lighter` to your minor mode to make sure that block is evaluable.

Hopefully that's helpful...

Further, I'm reading your documentation and I am having trouble understanding the workflow for this.  Maybe you can clarify -- is it up to a developer to write additional code in order to support mldoc-style documentation?  Your README has an example, but I am having trouble seeing things end-to-end.

tarsius:
Closing due to inactivity.

PR #6422, opened 2019-09-08
added recipe for org-working-set

marcIhm:

Maintain a small and changing subset of your org-nodes to visit with ease.

https://github.com/marcIhm/org-working-set

I am the author and maintainer of this package.

**None needed**

### Checklist

Please confirm with `x`:

Remark: I use two prefixes for my package: org-working-set- and org-ws-- ; I do this for brevity although package-lint-current-buffer is complaining; I hope this is okay for you.

### DIFF
--- a/recipes/org-working-set
+++ b/recipes/org-working-set
@@ -1,0 +1,2 @@
+(org-working-set :fetcher github
+	   :repo "marcIhm/org-working-set")


alphapapa:
You may find `org-now` interesting.

> Remark: I use two prefixes for my package: org-working-set- and org-ws-- ; I do this for brevity although package-lint-current-buffer is complaining; I hope this is okay for you.

MELPA requires that the prefix guidelines be followed.

riscy:
Yes, and in addition to @alphapapa 's comment, please consider the following:

```
org-working-set.el:416:31:Warning: reference to free variable `org-mode-map'

In org-ws--bottom-of-node:
org-working-set.el:572:60:Warning: reference to free variable
    `org-outline-regexp-bol'

In end of data:
org-working-set.el:725:1:Warning: the following functions are not known to be
    defined: org-entry-get, org-id-get-create, org-with-limited-levels,
    org-clock-in, org-with-point-at, org-id-find, org-goto-marker-or-bmk,
    org-id-get, org-get-heading, org-id-goto, org-format-outline-path,
    org-get-outline-path, org-back-to-heading, org-end-of-subtree,
    org-at-heading-p, org-reveal, org-entry-put, org-show-context,
    org-fill-paragraph
```
It seems like you may need to just `(require 'org)`?

Did you really do the paperwork to sign your package's copyright to the FSF? (`Copyright (C) 2011-2018 Free Software Foundation, Inc.`).  I like to double-check this one just in case.

And some added miscellaneous hints (automated):
- org-working-set.el#L615: It's safer to sharp-quote function names; use `#'identity`
- org-working-set.el#L616: It's safer to sharp-quote function names; use `#'`
- org-working-set.el#L26: Prefer `https` over `http` (if possible)

marcIhm:
Hi, thanx for your feedback !
Addressed all the technical points:
- (require 'org)
- added # to 'identity
- https
- changed org-ws-- into org-working-set--
Regarding org-now: Did not know about it; but I think the use-case is distinct: org-working-set can be helpful if you work on a handful of tasks one after the other; org-now seems better suited for working on tasks simultanously.
And I should mention org-mru-clock too, which has more overlap with org-working-set.
However, I think all use-cases are distinct enough, so I just put a reference of the other two packages into my description.
Best regards
Marc

marcIhm:
Forgot to answer riscy's previous question about the FSF-Paperwork: Yes I can confirm: I did sign the paperwork ! My other package org-index has previously been part of org-contrib, so I had to sign the paperwork back then.

riscy:
Thanks!  I'm looking at the code a little more closely now...

The keymap in `org-working-set--menu-install-keyboard-shortcuts` has a couple issues in it.  
- First, the keymap that it returns is hidden inside the function definition, so it makes it hard for a user to modify.  Defining the keymap at the top level with a `defvar` is usually a good idea because it exposes it to users. 
- Second, linking keystrokes to anonymous lambda commands should usually be avoided and, instead, each one of those lambdas should be a separately defined command.  This way users can fetch help about keystrokes and be linked to actual functions with definitions they can read and modify.  This is beneficial for you as the maintainer too, since it'll let you change the definition of these functions while the mode is running.

The while/progn loop on lines 668--672 is a little surprising -- it may be worth adding a body to it if that's correct, but something seems wrong.

Regarding the use of `message (concat` on lines 251 and 591, you may or may not know that `message` takes f-strings, so you can write e.g. `(message "%s%s." message again)` for example.  Something small I thought I'd point out.

marcIhm:
Hi riscy, seems I overlooked the awaiting-upstream label for a while ... Thanx for your comments !
Especially: I will modify org-working-set--menu-install-keyboard-shortcuts as you suggested.
However, the while/progn loop on lines 668--672 is intended that way, as I need to go through the body at least once.
Finally: My use of message is probably personal taste only, as I think it reads easier that way (eyes only moving in one direction :-)
So I will pick up most of your suggestions, then do some integration with org-mru-clock and then reply again.
Thanx for taking the time, your comments are interesting and helpful !

marcIhm:
Just addressed your main point and added two keymaps org-working-set-circle-keymap and org-working-set-menu-keymap which each have a set of interactive functions bound to various keys. That way users now may change keybindings as they preferred.
Hope, this makes the pull request fit for acceptance ...

riscy:
Great, thanks for that and for your patience with the process. :)  Unfortunately since our last conversation, package-lint has been updated with some more accurate version checks:

<kbd>M-x package-lint-current-buffer</kbd> (using version 20191027.20):
```
3 issues found:

280:21: error: You should depend on (emacs "25.1") or the seq package if you need `seq-intersection'.
281:44: error: You should depend on (emacs "25.1") or the seq package if you need `seq-difference'.
625:7: error: You should depend on (emacs "25.1") if you need `cursor-intangible-mode'.
```

But once these are sorted out, this looks good to merge!

marcIhm:
That was easy :-)
Just incremented the required version ...

riscy:
Great, thanks again for your patience.

PR #6424, opened 2019-09-08
Recipe: Add org-sidebar

alphapapa:

This package presents a helpful sidebar view for Org buffers. At the top is a chronological list of scheduled and deadlined tasks in the current buffer (similar to the Org agenda ,but without all its features), and below that is a list of all other non-done to-do items. If the buffer is narrowed, the sidebar only shows items in the narrowed portion; this allows seeing an overview of tasks in a subtree.

https://github.com/alphapapa/org-sidebar

Author and maintainer.

### Checklist

Please confirm with `x`:

Thanks.

### DIFF
--- a/recipes/org-sidebar
+++ b/recipes/org-sidebar
@@ -1,0 +1,1 @@
+(org-sidebar :fetcher github :repo "alphapapa/org-sidebar")


riscy:
Nice! :)

PR #6426, opened 2019-09-08
Add ‘tramp-auto-auth’ recipe.

oitofelix:
----

TRAMP automatic authentication library

This library provides `tramp-auto-auth-mode`: a global minor mode
whose purpose is to automatically feed TRAMP sub-processes with
passwords for paths matching regexps.  This is useful in situations
where interactive user input is not desirable or feasible.  For
instance, in sub-nets with large number of hosts or whose hosts have
dynamic IPs assigned to them.  In those cases it’s not practical to
query passwords using the `auth-source` library directly, since this
would require each host to be listed explicitly and immutably in a
Netrc file.  Another scenario where this mode is useful are
non-interactive Emacs sessions (like those used for batch processing
or by evaluating `:async` Org Babel source blocks) in which it’s
impossible for the user to answer a password-asking prompt.

When a TRAMP prompt is encountered, `tramp-auto-auth-mode` queries the
alist `tramp-auto-auth-alist` for the auth-source spec value whose
regexp key matches the correspondent TRAMP path.  This spec is then
used to query the auth-source library for a presumably phony entry
exclusively dedicated to the whole class of TRAMP paths matching that
regexp.

To make use of the automatic authentication feature, on the Lisp side
the variable `tramp-auto-auth-alist` must be customized to hold the
path regexps and their respective auth-source specs, and then
`tramp-auto-auth-mode` must be enabled.  For example:

#### **`~/.emacs.el`**
```elisp
(require 'tramp-auto-auth)

(add-to-list
 'tramp-auto-auth-alist
 '("root@10\\.0\\." .
   (:host "Funny-Machines" :user "root" :port "ssh")))

(tramp-auto-auth-mode)
```

After this, just put the respective sacred secret in an
authentication source supported by auth-source library.  For
instance:

#### **`~/.authinfo.gpg`**
```elisp
machine Funny-Machines login root password "$r00tP#sWD!" port ssh
```

In case you are feeling lazy or the secret is not so secret (nor so
sacred) -- or for any reason you need to do it all from Lisp --
it’s enough to:

```elisp
(auth-source-remember '(:host "Funny-Machines" :user "root" :port "ssh")
		         '((:secret "$r00tP#sWD!")))
```

And happy TRAMPing!

https://github.com/oitofelix/tramp-auto-auth

I’m the original author.

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/tramp-auto-auth
+++ b/recipes/tramp-auto-auth
@@ -1,0 +1,1 @@
+(tramp-auto-auth :fetcher github :repo "oitofelix/tramp-auto-auth")


riscy:

No issues! 💯

```
No issues found.
```

```
tramp-auto-auth.el:127:13:Warning: Unused lexical variable `spec'
tramp-auto-auth.el:127:13:Warning: Unused lexical variable `pre-secret'
```

- tramp-auto-auth.el#L9: Use `Version` instead of `Package-Version` (MELPA automatically adds `Package-Version`)

oitofelix:
Hi Chris,


>
> tramp-auto-auth.el:127:13:Warning: Unused lexical variable `spec'
> tramp-auto-auth.el:127:13:Warning: Unused lexical variable `pre-secret'

Don’t know how to fix these, because both variables are clearly used.
Actually, when I prefix them with an underscore (‘_’), the byte compiler
complains they are both used.


> Added miscellaneous hints (automated):
>
>   • tramp-auto-auth.el#L9: Use Version instead of Package-Version (MELPA automatically adds Package-Version)

This has been fixed.


Thanks!

-- 
 88888  FFFFF Bruno Félix Rezende Ribeiro (oitofelix) [0x28D618AF]
 8   8  F     http://oitofelix.freeshell.org/
 88888  FFFF  mailto:oitofelix@gnu.org
 8   8  F     irc://chat.freenode.org/oitofelix
 88888  F     xmpp://oitofelix@riseup.net

riscy:
Hi!  I'm inclined to believe the compiler's found something (perhaps an unnecessary line).  If you're 100% sure, I can try to step through it in a little while.

oitofelix:
I’m rarely 100% sure of anything ;).  In this case I can’t see what I’m missing, though.  Your help is very welcome!

riscy:
Ok, I had to read a bit on pcase.  I ran `macroexpand` on the pcase and I get the following:

```elisp
(let* ((val (or (car (last vec)) ""))
                (x1868 (funcall (function (lambda (expval) (assoc-default expval tramp-auto-auth-alist (function string-match-p)))) val)))
           (if x1868 (let* ((sym (let* ((spec x1868)) (plist-get (car (apply (function auth-source-search) spec)) :secret))))
                       (if sym (let* ((sym (let* ((pre-secret sym)) (if (functionp pre-secret) (funcall pre-secret) pre-secret))))
                                 ;; *** here are the problematic bindings ***
                                 (if sym (let ((secret sym) (pre-secret sym) (spec x1868))
                                           (process-send-string proc (concat secret tramp-local-end-of-line)))
                                   (funcall tramp-action-password proc vec)))
                         (funcall tramp-action-password proc vec)))
             (funcall tramp-action-password proc vec)))
```

I think we can blame this on `pcase`'s expanded form binding variables it doesn't strictly need to.

We can still silence the byte compiler (which I think is a worthy goal, since then when other byte compilation errors crop up we treat them more seriously).  One option is to use the variables, even in a redundant way:

```elisp
(pcase (or (car (last vec)) "")

			((and
                          (app (lambda (expval) (assoc-default expval tramp-auto-auth-alist #'string-match-p)) spec)
                          (guard spec)

                          (let pre-secret (plist-get (car (apply #'auth-source-search spec)) :secret))
                          (guard pre-secret)

                          (let secret (if (functionp pre-secret) (funcall pre-secret) pre-secret))
                          (guard secret))

                         (when (and spec pre-secret spec)  ; not required but pacifies byte compiler
                           (process-send-string proc (concat secret tramp-local-end-of-line))))

                        (_ (funcall tramp-action-password proc vec)))
```

Let me know what you think about that, and thanks for your patience.

oitofelix:
Hello Chris,

> We can still silence the byte compiler (which I think is a worthy
> goal, since then when other byte compilation errors crop up we treat
> them more seriously). One option is to use the variables, even in a
> redundant way:

I did just that (and documented it).  Thanks for the suggestion.

-- 

riscy:
Thanks, this looks interesting.

PR #6427, opened 2019-09-09
Add ‘elpa-deploy’ recipe.

oitofelix:
----

ELPA deployment library

This library provides `elpa-deploy`: a complement to the function
`package-upload-file` from `package-x.el`, which automatizes the
deployment of simple and multi-file packages.  This function
automatizes the upload of a library to an ELPA directory so that no
other action is needed.  Particularly useful for rapid ELPA deployment
aiding testing and releases.  The procedures automatically taken care
of consist of updating the version field of a package source using the
current time-stamp, generating its tar archive if multi-file, and
uploading the results to a specified ELPA directory, while deleting
any previous version of the same package already deployed.

https://github.com/oitofelix/elpa-deploy

I’m the original author.

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/elpa-deploy
+++ b/recipes/elpa-deploy
@@ -1,0 +1,1 @@
+(elpa-deploy :fetcher github :repo "oitofelix/elpa-deploy")


riscy:
Thanks for this -- you've probably seen my other lint checks.  For this one the only thing I'm seeing on an automated first pass is this one:

- elpa-deploy.el#L10: Use `Version` instead of `Package-Version` (MELPA automatically adds `Package-Version`)

I need to check the documentation around this one, so it may not even be your fault. :)

oitofelix:
Hello Chris,

>   • elpa-deploy.el#L10: Use Version instead of Package-Version (MELPA automatically adds Package-Version)

This has been fixed.

Thank you!

-- 
 88888  FFFFF Bruno Félix Rezende Ribeiro (oitofelix) [0x28D618AF]
 8   8  F     http://oitofelix.freeshell.org/
 88888  FFFF  mailto:oitofelix@gnu.org
 8   8  F     irc://chat.freenode.org/oitofelix
 88888  F     xmpp://oitofelix@riseup.net

riscy:
Hi!  You opened this PR before some recent updates to `package-lint`, which will tell you that you need to depend on Emacs 24.4.  It's a small change, and I was considering grandfathering it in, but it's also a very easy change to make.

Another small thing I notice is that you use unicode in some of the strings you display to users (e.g. `Package definition file ‘%s’ doesn’t exist`).  For usability purposes and for those who might be stuck on simpler terminals, it may be worth converting those to ASCII, but your call.

PIng me when you have `;; Package-Requires: ((emacs "24.4") (f "0.0"))` and we should be all set!

oitofelix:
Hi Chris,


> PIng me when you have ;; Package-Requires: ((emacs "24.4") (f "0.0"))
> and we should be all set!

Done.


> Another small thing I notice is that you use unicode in some of the
> strings you display to users (e.g. Package definition file ‘%s’
> doesn’t exist). For usability purposes and for those who might be
> stuck on simpler terminals, it may be worth converting those to ASCII,
> but your call.

I’d prefer to make that change once the first user in those
circumstances complain about it.


Thanks!

-- 

riscy:
Great, thanks!

PR #6429, opened 2019-09-09
Add :files to org2blog.

grettke:
Now Org2Blog uses README.org in its package so add :files, specifying README.org.

Tested against Emacs 26.3 by building the package at the command line, inspecting the tar file, and then in Emacs installing the package using package-install-file and inspecting the directory under the elpa dir, and the package loaded successfully as expected.

### DIFF
--- a/recipes/org2blog
+++ b/recipes/org2blog
@@ -1,1 +1,1 @@
-(org2blog :repo "org2blog/org2blog" :fetcher github)
+(org2blog :repo "org2blog/org2blog" :fetcher github :files (:defaults "README.org"))


tarsius:
By default `README*` files are not included in any packages, not even if that file is named `README.md` or just `README`. Are you aware of that and if so, then why do you think an exception should be made for this package? To me it seems that in this case bundling the readme file would be even less useful than in many other cases because it consists primarily of screenshots and these image files are not distributed with the package.

grettke:
Hi @tarsius I forgot to share that I rewrote the readme [in another branch](https://github.com/org2blog/org2blog/tree/v1.1.0), and that readme would serve to justify this request. In light of that omission I replied to your questions in chronological order because it seemed easier to read.

> By default README* files are not included in any packages, not even if that file is named 
> README.md or just README. Are you aware of that 

When I read the [recipe format](https://github.com/melpa/melpa#recipe-format) I saw in the `:files` documentation for `:defaults` that README.org is not included as a default. That is why I added README.org to the `org2blog` recipe. In that way I was aware of it. Is that what you mean?

> and if so, then why do you think an exception should be made for this package? 

Org2Blog v1.1.0, the pending release, contains completely rewritten documentation and additional functionality. For that I consolidated everything essential for using it into one place: README.org. 

Org2Blog can be configured for very different styles of blogging. There are a lot of features for doing the blogging. First time bloggers (Org2Bloggers) can start by setting up a very simple configuration to get started. There are way too many features to use them all right away, and you shouldn't. Over time as you get more comfortable you skim the documentation to see what is possible or specifically for what you want to do. As you iterate and refine your blogging process you expect more of the software. That is when you dig deeper into the functionality. That iterative process continues of refining your blogging process and demanding more of the software. Finally you reach an advanced level where you want to do things like migrate your blog into Org2Blog or track your posts or configure multiple blogs or develop the code yourself. Throughout this iterative process you end up using the documentation in a non-linear fashion because you wear the beginner, intermediate, and advanced hats at different times. You end up referring to the documentation at all levels because there are so many features that it is hard to remember all of them and details about how they work because you always have to consider how it is working with the WordPress side of things. You end up reading a lot, taking notes, and updating your own config: this package helps by including that readme. Blogging is fun with Org2Blog when you have the manual available always nearby when you need it. 

That is why an exception should be made to include README.org with this package.

> To me it seems that in this case bundling the readme file would be even less useful than in 
> many other cases

Totally understood and agreed. I forgot to share that I'm doing all of the v1.1.0 work in [this branch](https://github.com/org2blog/org2blog/tree/v1.1.0). I haven't moved it into `master` yet though because I wanted to make sure that the new recipe worked correctly first. 

I'm sorry that I forgot to mention that absolute key part to me sharing any of this.
 
> because it consists primarily of screenshots and these image files are not 
> distributed with the package.

I considered including all of the screenshots. Blogging is very visual and it helps to see what happens. What I noticed though was that once see how it works once I never look at the screenshot again. While referring to the readme as a reference the context is already there so I don't think the screenshots are necessary. 

This package is an application. If it were a desktop app I would include all of the images and other user-oriented but non essential documentation and the SVG formats of the logo and icon in the package. I get the sense though that MELPA packages are not supposed to include all of that kind of stuff like how you would package up a desktop app. Do I have it right? 

Org2Blog is such a great software that I inherited as maintainer and I felt that it deserved a logo that captured the fun and spirit of writing and blogging. For this v1.1.0 release I hired a graphic artist to make a logo and icon for Org2Blog that really did Org2Blog justice because it had none. It turned out great. I wanted to include the logo and icon in the package simply because it I think they turned out super. However I did not because I figured it isn't critical and couldn't justify it. Can packages simply include what they wish? It is MELPA hosting the files and paying for the bandwidth and since I don't consider it critical I didn't include it.

tarsius:
Now that I know that the README.org is basically being turned into a manual I agree that it should be included in the package--after all we do that for info manuals as well.

One thing that you might want to consider though is to include a "proper" manual instead. You can even author the manual using Org and export that to info. If you do that, then I recommend you give my [`ox-texinfo+`](https://github.com/tarsius/ox-texinfo-plus) a try. I use that for Magit and a few other of my packages (which you should look at for examples).

PR #6432, opened 2019-09-10
add hnreader

thanhvg:
This package renders hackernews website at https://news.ycombinator.com in an org buffer. Almost everything works.

The org-mode buffer feature interactive links similar to html.

https://github.com/thanhvg/emacs-hnreader

I am the maintainer

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/hnreader
+++ b/recipes/hnreader
@@ -1,0 +1,1 @@
+(hnreader :fetcher github :repo "thanhvg/emacs-hnreader")


riscy:
Nice, I keep getting surprised by what people are doing with org-mode buffers.  As a convenience to users, I might recommend adding a function (or alias) simply called `hnreader` to give users an easy/obvious entrypoint -- maybe it would just alias to `hnreader-newest`.

That's not a blocker, though, and thanks for catching everything on the checklist.

PR #6433, opened 2019-09-11
Add amixer recipe

remvee:

Basic commands to control audio volume via amixer.

https://github.com/remvee/amixer-el

**None needed**

### Checklist

### DIFF
--- a/recipes/amixer
+++ b/recipes/amixer
@@ -1,0 +1,1 @@
+(amixer :fetcher github :repo "remvee/amixer-el")


riscy:
You may want to make the customization group explicit with, e.g.

```elisp
(defgroup amixer nil "Functions to call out to amixer."
  :prefix "amixer-" :group 'convenience)
```

but this isn't a blocker.  Thanks for your close adherence to the checklist -- and, welcome to MELPA!

remvee:
Thanks for the tip!

tarsius:
Another, older, `amixer.el` also exists: http://www.sugarshark.com/elisp/mylisp/amixer.el. I'll just leave this here (and update the Emacsmirror to mirror your implementation), but would encourage you to look at the other implementation, possibly contact its author, and ideally come up with an implementation that is "better" than both current implementations, whatever that means in this case. But I do not want to be involved in the process (done that too often already, for other packages I personally am not really interested in).

tarsius:
@riscy @purcell maybe the review process should involve checking whether another package by the same name already exists on the Emacsmirror (*including* the Emacsattic)?

remvee:
Maybe I should also put that in my hey-lets-whip-something-up-process, it's almost identical.

riscy:
That's my mistake!  It's outlined well already in CONTRIBUTING, but I haven't formalized it well enough as a part of my own process.

remvee:
Please let me know if I can do anything, rename for instance.

riscy:
Thanks hugely for the offer, @remvee.  Now that I actually take the time to google, I see one more amixer, which is part of emacspeak -- https://github.com/tvraman/emacspeak/blob/master/lisp/amixer.el

In light of that, perhaps it would be a good idea to rename -- I'd still encourage you to reach out to the amixer author @tarsius listed in case there's possibility for collaboration.

remvee:
@tarsius let's start by dropping this recipe from melpa.  This package is the result of me trying to tidy up some of my emac.d setup, thinking somebody might also want to use it.  It only yielded 10 downloads so far at this moment (according to https://melpa.org/#/amixer) so not many people will be effected by this, if at all.  I'm not sure what the policy on pulling packages is on melpa but can you revert the merge?

Meanwhile I'll try an come up with another name and submit another PR when that's done.

tarsius:
Considering that this only got added a few days ago, I think we can just remove it again. Please open a pull-request for that.

PR #6434, opened 2019-09-12
Add didyoumean

kisaragi-hiu:

The user types `emacs he<TAB><RET>` in the shell expecting the file name to be expanded to `hello.el`; Emacs opens up an empty `hello.` because there is a `hello.elc` in the same folder. This package makes Emacs ask which of these files the user actually wants to edit.

https://gitlab.com/kisaragi-hiu/didyoumean.el

I am the maintainer.

None needed.

### Checklist

Please confirm with `x`:

- [x] The package is released under a [GPL-Compatible Free Software License](https://www.gnu.org/licenses/license-list.en.html#GPLCompatibleLicenses). (MIT)

### DIFF
--- a/recipes/didyoumean
+++ b/recipes/didyoumean
@@ -1,0 +1,1 @@
+(didyoumean :fetcher gitlab :repo "kisaragi-hiu/didyoumean.el")


riscy:
This appears to work as an alternative to heavier-handed completion packages like helm and ivy -- and nice use of `-unload-function`.

This feels like it could be a global(?) minor mode, since that would give users a familiar way to toggle the feature off and on.  It should be easy to write with your existing functionality, e.g.:

```elisp
(define-minor-mode didyoumean-mode
   "Did you mean...?"
   :global t
   (if didyoumean-mode
     (add-hook 'find-file-hook #'didyoumean)
    (didyoumean-unload-function)))
```

kisaragi-hiu:
Done in https://gitlab.com/kisaragi-hiu/didyoumean.el/commit/6d0c4203eb192d73d89261b3a9bad52951e394af.

riscy:
This isn't a blocker, but your license front-matter says "didyoumean.el is free software", but "free software" sometimes means GPL-licensed software.  I think altogether the text is clear, but it might be better to just write:

```
;; didyoumean.el can be redistributed it and/or modified
;; under the terms of the MIT License.
```

Or even paste your license there verbatim, since the MIT license is deliberately short.

kisaragi-hiu:
Will do, thank you!

PR #6437, opened 2019-09-15
Add comware-router-mode recipe

daviderestivo:

Major mode for editing Comware routers and switches configuration files.

https://github.com/daviderestivo/comware-router-mode

I'm the author and current maintainer of this mode.

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/comware-router-mode
+++ b/recipes/comware-router-mode
@@ -1,0 +1,1 @@
+(comware-router-mode :repo "daviderestivo/comware-router-mode" :fetcher github)


riscy:

No issues! 💯

```
No issues found.
```

```
comware-router-mode.el:168:24:Warning: `beginning-of-buffer' is for
    interactive use only; use `(goto-char (point-min))' instead.
```

- comware-router-mode.el#L259: It looks like you're defining your major-mode in an unusual way -- check out [this](https://www.gnu.org/software/emacs/manual/html_node/elisp/Major-Mode-Conventions.html) section of the elisp manual.
- comware-router-mode.el#L168: Don't quote lambdas; this prevents them from being compiled

daviderestivo:
> Per CONTRIBUTING.org, consider [using lexical binding](https://github.com/melpa/melpa/blob/master/CONTRIBUTING.org#fixing-typical-problems) (and depend on at least (emacs "24"))
> (The following assumes lexical binding is enabled)

The mode was already depending on  (emacs "24"). I have enabled lexical binding.

> 
> ```
> comware-router-mode.el:168:24:Warning: `beginning-of-buffer' is for
>     interactive use only; use `(goto-char (point-min))' instead.
> ```

Valid point. I have fixed it.

> Added miscellaneous hints (automated):
> 
> * comware-router-mode.el#L259: It looks like you're defining your major-mode in an unusual way -- check out [this](https://www.gnu.org/software/emacs/manual/html_node/elisp/Major-Mode-Conventions.html) section of the elisp manual.
I guess the mode is properly defined using an autoload function.

> * comware-router-mode.el#L168: Don't quote lambdas; this prevents them from being compiled

Fixed it.

Thanks for your review :)

riscy:
Thanks for addressing that first pass.  I'm taking a more in-depth look now:

You probably want to use the `define-derived-mode` macro and derive from `prog-mode` -- I should probably have recommended that in my first comment.  The reason for this is that some users will add hooks to prog-mode to do programming language-specific things, such as enabling line numbers, certain personalized syntax highlighting, etc.  Likely they will want the same for your mode.  I recommend `define-derived-mode` because you have to write less code this way (e.g. you don't have to explicitly run hooks or call `use-local-map` or define `comware-router-mode-hook` and so forth), and so there's less room to make a mistake.  I'd recommend taking a look at a package that is similar to yours (maybe another prog-mode extension).

Both `comware-router-mode-map` and `comware-router-mode-syntax-table` can be defined by wrapping a `let` inside a `defvar` -- it's better to do it this way because you don't want a user's customizations to be clobbered if they reload your feature or re-evaluate the source buffer.  An example of what I mean:
```elisp
(defvar comware-router-mode-syntax-table
  (let ((table (make-syntax-table)))
    (modify-syntax-entry ?_ "w" table))
  "Syntax table for comware router mode.")
```

Good work with defining the faces using inherited faces only.

I think once those two (broad) issues above are addressed, this is good to merge!

daviderestivo:
> I think once those two (broad) issues above are addressed, this is good to merge!

@riscy I would like to thank you for having spent your time in reviewing my code. This is very valuable for me so I can improve my elisp coding skills :). 

Btw, I think I have fixed both issues.

riscy:
Wooops, I made a mistake -- I forgot to make sure the `defvar` assigns the value of the eventual table.

```elisp
(defvar comware-router-mode-syntax-table
  (let ((table (make-syntax-table)))
    (modify-syntax-entry ?_ "w" table)
    table) ;; <-- forgot
  "Syntax table for comware router mode.")

**Comment by daviderestivo** [comment]
> Wooops, I made a mistake -- I forgot to make sure the `defvar` assigns the value of the eventual table.

Fixed it.

**Comment by riscy** [comment]
One last nag, I promise. :)  You actually need Emacs 24.3 for `setq-local` now that it's been added:

```
;; Package-Requires: ((dash "2.16.0") (emacs "24.3"))
```

daviderestivo:
> ;; Package-Requires: ((dash "2.16.0") (emacs "24.3"))

Fixed it as well 👍

riscy:
Great!

PR #6438, opened 2019-09-17
[fix_codcut] fix missed recipe name in #6406

riscy:
fyi @dgopsq, I made a mistake when I merged the recipe - the build expects the package to have the same name as the elisp file.  I should have rebuilt your PR after the name change.


riscy:
Great, thanks!  I wanted to be sure.

PR #6442, opened 2019-09-20
Add global-tags

xuhdev:

Tastable Emacs Lisp API that wraps around GNU global tag system

https://launchpad.net/global-tags.el

An enthusiastic user?

### Checklist

Please confirm with `x`:

cc @FelipeLema

When I'm testing the package built via `make recipes/global-tags`, I keep getting the error:

    package-buffer-info: Search failed: ";;; global.el ends here"

I have no idea how to fix this...

### DIFF
--- a/recipes/global-tags
+++ b/recipes/global-tags
@@ -1,0 +1,2 @@
+(global-tags :fetcher git
+             :url "https://git.launchpad.net/global-tags.el")


walseb:
> When I'm testing the package built via make recipes/global-tags, I keep getting the error:
> package-buffer-info: Search failed: ";;; global.el ends here"
> I have no idea how to fix this...

The problem is that in the first line of the file you name the package "global.el" while when you put the end marker you refer to it as "global-tags.el"

riscy:
Feel free to ping when this is ready!

xuhdev:
Bug submitted to upstream: https://bugs.launchpad.net/global-tags.el/+bug/1844960

FelipeLema:
I've fixed the bug and complied with the checklist upstream, please double check and update @xuhdev.

xuhdev:
I can confirm that things work well now

xuhdev:
ping @riscy (in case you didn't receive the notification)

riscy:
Sorry about that delay!  A quick pass over the checklist:

No issues! 💯

```
1 issue found:

```

```
global-tags.el:214:45:Warning: reference to free variable `v'

In end of data:
global-tags.el:293:1:Warning: the following functions are not known to be
    defined: with-parsed-tramp-file-name, tramp-make-tramp-file-name,
    tramp-file-name-method, tramp-file-name-user, tramp-file-name-domain,
    tramp-file-name-host, tramp-file-name-port, tramp-file-name-hop
```

- global-tags.el#L133: It's safer to sharp-quote function names; use `#'`
- global-tags.el#L52: Customizable variables shouldn't be private

FelipeLema:
all items have been addressed. I wasn't getting the `tramp` ones on my main devenv, so I used another machine (with different OS+Emacs) to confirm them.

Thanks.

riscy:
Hi, I'm taking a closer look now --

This feels like it is modifying the behavior of Emacs with the `cl-defmethods` redefining the behavior of built-in functions like `xref-backend-references`.  It's strict, but simply loading a library should not affect Emacs's behavior.  In this case I think this should be a minor mode than can be toggled off and on... does that seem feasible?

The minor stuff that I saw:

Your `declare-function`s should contain strings pointing to where the functions are defined, maybe along these lines:
```elisp
(declare-function with-parsed-tramp-file-name "tramp" t t)
(declare-function tramp-make-tramp-file-name "tramp" t t)
```
(You can check the helpdoc on `declare-function` and run `check-declare-file` on your code for more.)

The `:group 'global` in `defcustom global-tags-global-command` seems like a wrongly named group.  It may be a good idea to lead with a `defgroup` and to name the group `'global-tags` instead of `'global` (which is too vague).

FelipeLema:
I've fixed the `declare-function` items and changed the `:group` for `global-tags-global-command` as you mentioned.

Also

> This feels like it is modifying the behavior of Emacs with the cl-defmethods redefining the behavior of built-in functions like xref-backend-references. It's strict, but simply loading a library should not affect Emacs's behavior. In this case I think this should be a minor mode than can be toggled off and on... does that seem feasible?

These `cl-defmethod` for xref.el and project.el functions do not change Emacs behaviour: they expand the support for xref/project for more/other backends but, at the same time, do not change the default backends being shipped with Emacs. Just doing `(require 'global-tags.el')` will not alter Emacs behaviour at all.

These methods will only be called in case you change the default backends, but this is done in two steps:
- Adding "I should handle the backend for project.el/xref.el for this context (current file)" functions in `project-find-functions` and `xref-backend-functions` (not done explicitly in global-tags.el, though suggested).
- Actually reporting as valid backend when these aformentioned functions are called. (global-tags.el has `global-tags-xref-backend` and `global-tags-try-project-root`.

FelipeLema:
I don't know if my ~mansplaination~ explanation convinced @riscy, but so far I don't see the need to change anything upstream.

riscy:
Thanks for your patience --

> These cl-defmethod for xref.el and project.el functions do not change Emacs behaviour

That was my misunderstanding.  I wasn't familiar with the `cl-defgeneric`/`cl-defmethod` behavior.  Sorry for the confusion.

About the novelty of the package:  are you aware of the existing GNU global packages listed [here](http://melpa.org/#/?q=gnu%20global&sort=downloads&asc=false)?  One of these is even in the 93rd percentile of MELPA downloads -- have you checked it or the other xref gnu global packages out?  Can you take a look at some of these and comment on what makes your package unique?

FelipeLema:
I've compared this package against 3/4 most downloaded packages you pointed out in [the README file](https://git.launchpad.net/global-tags.el/tree/README.adoc). Here's a complete comparison:

* [counsel-gtags](https://github.com/syohex/emacs-counsel-gtags): unmaintained and non-re-usable [counsel](https://github.com/abo-abo/swiper) interface. (my maintained fork [here](https://github.com/FelipeLema/emacs-counsel-gtags/))
* [ggtags](https://github.com/leoliu/ggtags): interface using `compilation-mode`. Not a re-usable library (see https://github.com/leoliu/ggtags/issues/175#issuecomment-428986369)
* [gxref](https://github.com/dedi/gxref/): unmaintaned
* [helm-gtags](https://github.com/syohex/emacs-helm-gtags): unmaintaned
* [agtags](https://github.com/vietor/agtags): no tests up to date, provides custom interface to final user (adds shortcuts to new commands)

I'll update the README for the missing packages listed here later tomorrow.

riscy:
Alright, thanks a lot -- and thanks both for your patience.

PR #6445, opened 2019-09-22
Update url for diminish-buffer.

jcs090218:
Refer to this #6415. Starting from next package I will start the package from organization. Thanks!


tarsius:
Thanks @jcs090218

PR #6444, opened 2019-09-21
Stan mode packages

kaz-yos:
### Brief summary of what the package does
example

These are a family of emacs packages to provide programming support for Stan, a probabilistic programming language for Bayesian statistical inference.

- UPDATE stan-mode for indentation and syntax highlighting
- UPDATE stan-snippets for function argument completion
- NEW company-stan for auto-completion via company
- NEW eldoc-stan live help for function arguments
- NEW flycheck-stan on-the-fly syntax checking with stanc (stan compiler)
- NEW ac-stan for auto-completion via auto-complete

Note that they are kept in a single repo because they must be kept in sync with the common stan language definition file.

### Direct link to the package repository

https://github.com/stan-dev/stan-mode/

### Your association with the package

Current Maintainer

### Relevant communications with the upstream package maintainer

Not applicable

### Checklist

Please confirm with `x`:

- [x] The package is released under a [GPL-Compatible Free Software License](https://www.gnu.org/licenses/license-list.en.html#GPLCompatibleLicenses).
- [x] I've read [CONTRIBUTING.org](https://github.com/melpa/melpa/blob/master/CONTRIBUTING.org)
- [x] I've used the latest version of [package-lint](https://github.com/purcell/package-lint) to check for packaging issues, and addressed its feedback: No issues on [Travis-CI](https://travis-ci.org/stan-dev/stan-mode/jobs/587722072#L1977)
- [x] My elisp byte-compiles cleanly: No byte-compilation issues on [Travis-CI](https://travis-ci.org/stan-dev/stan-mode/jobs/587722072#L1482)
- [x] `M-x checkdoc` is happy with my docstrings: No issue on [Travis-CI](https://travis-ci.org/stan-dev/stan-mode/jobs/587722072#L2109). Run via elisp-lint
- [x] I've built and installed the package using the instructions in [CONTRIBUTING.org](https://github.com/melpa/melpa/blob/master/CONTRIBUTING.org)
- [ ] I have confirmed some of these without doing them

PR #6450, opened 2019-09-22
Add eldoc-stan

kaz-yos:
This is a split version of https://github.com/melpa/melpa/pull/6444

## Brief summary of what the package does

The `eldoc-stan` provides live help for function arguments via `eldoc` for all functions in Stan, a probabilistic programming language for Bayesian statistical inference.

This package was newly added in https://github.com/stan-dev/stan-mode/commit/d3c045c74cce32121e0e9c13a7222c537a14bbf7

https://github.com/stan-dev/stan-mode
https://github.com/stan-dev/stan-mode/eldoc-stan

Maintainer

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/eldoc-stan
+++ b/recipes/eldoc-stan
@@ -3,4 +3,5 @@
  :repo "stan-dev/stan-mode"
  :files ("eldoc-stan/*.el"
 	 "eldoc-stan/eldoc-stan.json"
-         (:exclude "eldoc-stan/test-*.el")))
+         (:exclude "eldoc-stan/test-*.el"
+                   "eldoc-stan/eldoc-stan-create-json.el")))
--- a/recipes/eldoc-stan
+++ b/recipes/eldoc-stan
@@ -1,0 +1,6 @@
+(eldoc-stan
+ :fetcher github
+ :repo "stan-dev/stan-mode"
+ :files ("eldoc-stan/*.el"
+	 "eldoc-stan/eldoc-stan.json"
+         (:exclude "eldoc-stan/test-*.el")))


riscy:
Hi again! :)  Thanks for your close adherence to the checklist.

- Minor, but it looks to me like `eldoc-stan-eldoc-documentation-function` does not need to be autoloaded per [when to autoload](https://www.gnu.org/software/emacs/manual/html_node/elisp/When-to-Autoload.html).
- You require `c-eldoc`, but unless I missed something it looks like you don't actually refer to `c-eldoc-format-arguments-string`.  Moreover c-eldoc does not appear to be part of Emacs (thus would be a `Package-Require` if you actually need it).

kaz-yos:
This commit addresses the issues: https://github.com/stan-dev/stan-mode/commit/e52f234f9a158af44092011d0b2b8d3bab5d38ae

- Do not autoload `eldoc-stan-eldoc-documentation-function`
- Do not require `c-eldoc` unnecessarily
  - `c-eldoc-format-arguments-string` had been replaced with `eldoc-stan--format-arguments-string`

riscy:
I'm just noticing eldoc-stan-create-json.el is included with your package.  But at a quick glance it looks like that file is just required by the Makefile (which isn't included) to create the .json file required by your package.  Should the .el file be excluded in your recipe?

kaz-yos:
Thanks for picking this up. I have excluded the file in question. That file is not required for regular users after the `eldoc-stan.json` is created.

riscy:
Great.  Thank you for your patience!

kaz-yos:
Thank you!

PR #6451, opened 2019-09-22
Add flycheck-stan

kaz-yos:
This is a split version of https://github.com/melpa/melpa/pull/6444

## Brief summary of what the package does

The `flycheck-stan` provides on-the-fly syntax check for Stan via its compiler `stanc`. Stan is a probabilistic programming language for Bayesian statistical inference.

This package was newly added in https://github.com/stan-dev/stan-mode/commit/0d875ad012d85c898fe6682b62df6182295b6871 based on a very early form that was never released.

https://github.com/stan-dev/stan-mode
https://github.com/stan-dev/stan-mode/flycheck-stan

Maintainer

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/flycheck-stan
+++ b/recipes/flycheck-stan
@@ -3,4 +3,5 @@
  :repo "stan-dev/stan-mode"
  :files ("flycheck-stan/*.el"
 	 "flycheck-stan/error_msgs.txt"
-         (:exclude "flycheck-stan/test-*.el")))
+         (:exclude "flycheck-stan/test-*.el"
+                   "flycheck-stan/flycheck-stan-error-msgs.el")))
--- a/recipes/flycheck-stan
+++ b/recipes/flycheck-stan
@@ -1,0 +1,6 @@
+(flycheck-stan
+ :fetcher github
+ :repo "stan-dev/stan-mode"
+ :files ("flycheck-stan/*.el"
+	 "flycheck-stan/error_msgs.txt"
+         (:exclude "flycheck-stan/test-*.el")))


riscy:
Some very small things, but otherwise I'm thinking this looks good to merge:

- You should add a dependency on Emacs 25.1 (instead of just 25) in both files, according to the latest version of package-lint. 
- Minor typo: `neline at the end of the tring`-> `newline at the end of the string`
- For the variable `flycheck-stan-regexp-error-msgs-start`, it feels like a decision hasn't been made about whether the list of strings should exist in `error_msgs.txt` or in the elisp code.  I think you should just pick one.  Perhaps you could use some elisp like this if you prefer to be able to paste the lines from a text file without the `cons` noise:
```elisp
(split-string "
could not find include file
PARSER FAILED TO PARSE INPUT COMPLETELY
Function 
Function declared, but not defined.
No matches for:
" "\n")
```

kaz-yos:
This commit addresses the first two issues: https://github.com/stan-dev/stan-mode/commit/2ba214ef80564ffe95a2d030212f10c3c219bace

I will work on the last one and come back.

kaz-yos:
This commit improves the ad-hoc pattern addition: https://github.com/stan-dev/stan-mode/commit/4fb2489a165ac57437508f72b3ee3244fc5bddff#diff-d15f67018b6ca976ee176545bc90c252

Now the patterns are added during the `error_msgs.txt` file generation. No ad-hoc addition is performed after loading this file.

riscy:
Similar to my previous comment, do we need to package `flycheck-stan-error-msgs.el` with this, or will that not be code required by the users?

kaz-yos:
Thanks for picking this up. Excluded the file in question. That file is not required for regular users after the `error_msgs.txt` is created.

riscy:
Thanks once more!

kaz-yos:
Thank you so much for your assistance!

PR #6452, opened 2019-09-23
Add nix-env-install

akirak:

This package helps you install various packages using Nix. At present, it supports installing npm packages. It is a companion to an upcoming package [lsp-server.el](https://github.com/akirak/lsp-server.el), which discovers and installs a language server for the current buffer.

https://github.com/akirak/nix-env-install

I am the maintainer

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/nix-env-install
+++ b/recipes/nix-env-install
@@ -1,0 +1,1 @@
+(nix-env-install :fetcher github :repo "akirak/nix-env-install")


riscy:
Great, this looks novel -- and thanks for minding the checklist!

- Do you necessarily need to autoload `nix-env-install-cachix-exists-p`, `nix-env-install-cachix-use` and `nix-env-install-cachix` (see [when to autoload](https://www.gnu.org/software/emacs/manual/html_node/elisp/When-to-Autoload.html)).  I may be missing something, but these seem like they aren't entrypoints that _always_ need to be available.
- The type for the customizable `nix-env-install-after-npm-function` should be `'(choice regexp nil)` I believe.
- This is more of a question than a suggestion, but do you think the definitions of "s2" and "s1" may be a good use case for a macro?  It may serve to simplify the code a little.

akirak:
Thank you for your comment. I think I have fixed the issues.

> Do you necessarily need to autoload nix-env-install-cachix-exists-p, nix-env-install-cachix-use and nix-env-install-cachix (see when to autoload). I may be missing something, but these seem like they aren't entrypoints that always need to be available.

IMO it is necessary to autoload `nix-env-install-cachix-use`, because it may be called by a human before he installs other packages manually using this package. The intended use of this package is with `lsp-server.el`, but it is useful in itself. Using cachix is important for installing some big packages - like ghcide which takes a few hours to build without remote cache, which I plan on supporting in the near future.

> The type for the customizable nix-env-install-after-npm-function should be '(choice regexp nil) I believe.

I made it a hook variable. Actually, it is not necessary for now, but it was added as an option.

> This is more of a question than a suggestion, but do you think the definitions of "s2" and "s1" may be a good use case for a macro? It may serve to simplify the code a little.

Instead of a macro, I've defined a function. It is hard for me to get macros right, especially when I want to pass variables as arguments. Functions are more straightforward.

riscy:
A quick scan through the checklist:

<kbd>M-x package-lint-current-buffer</kbd> (using version 20191002.501):
```
1 issue found:

75:48: error: `show-buffer' was removed in Emacs version 26.1.
```
I think this is just a variable name issue.  You could call it `display-buffer-p` or something to get rid of this false positive.

```
nix-env-install.el:73:1:Warning: Unused lexical variable ‘proc’

In nix-env-install--start-process:
nix-env-install.el:99:42:Warning: reference to free variable ‘process’
```
I think these two together point to a small bug?

- nix-env-install.el#L26: Prefer `https` over `http` (if possible)

purcell:
> I think this is just a variable name issue. You could call it `display-buffer-p` or something to get rid of this false positive.

Yeah, that's a false positive, sorry: noted in https://github.com/purcell/package-lint/issues/144.

riscy:
How's this one going?

akirak:
Sorry. I was thinking about how to lint my packages and didn't fix the errors of this package. I will work around the issues soon.

Also, lsp-mode wants automatic server installation as a built-in feature, and we were talking about how to merge/integrate my lsp-server package with lsp-mode: https://github.com/emacs-lsp/lsp-mode/issues/506

akirak:
@riscy Now it passes linting. Please add this package.

@purcell Thank you for your recent work in nix-emacs-ci. Now I can run CI without issues!

riscy:
Apologies for the long delay.  One small nag (not a blocker):

- nix-env-install.el#L26: Prefer `https` over `http` (if possible)

And this looks good to go.  Thanks!

PR #6453, opened 2019-09-23
json-process-client: Migrate from Nicolas' gitlab to gitea

DamienCassou:
2019-09-23-134457_1257x148_scrot


tarsius:
https://gitlab.petton.fr/nico/json-process-client

tarsius:
Thanks!

PR #6454, opened 2019-09-23
Move python packages from @proofit404 to extra pythonic-emacs org

dakra:
@proofit404 created an organization for his Python Emacs packages.
See https://www.reddit.com/r/emacs/comments/d34qk7/im_looking_for_maintainers_for_anacondamode/

CC @proofit404 @CeleritasCelery @jcs090218


tarsius:
Thanks!

PR #6443, opened 2019-09-20
remove ensime and return scala-mode/sbt-mode to original author

fommil:
The ENSIME project is shutting down, all repositories are now archived.

The `ensime` package is no longer maintained (and confirmed it no longer works for recent versions of Scala). This PR deletes the `ensime` package and returns ownership of `scala-mode` / `sbt-mode` to @hvesalai, the original and best author.

My conclusion is that community initiatives are a waste of time: individual contributors are the best way to advance free software. I apologise to Heikki Vesalainen for roping him into this and I encourage anybody who wants to contribute to Free Software to take the following hard-earned advice to heart: https://medium.com/@fommil/hide-your-real-name-in-open-source-3d67e74a8c56

Note to @hvesalai if you want to retain the issue tracker and PR history from the `ensime/` versions of this repository I can assign you the ownership of the repos but github says that you must delete your fork first (I tried it already). https://github.com/ensime/emacs-scala-mode/settings and https://github.com/ensime/emacs-sbt-mode/settings. Specifically it says "hvesalai already has a repository in the ensime/emacs-scala-mode network and You can only transfer a repository from an organization to yourself at this time "

PR #6456, opened 2019-09-25
hvesalai maintains sbt-mode and scala-mode

hvesalai:
See #6443

### DIFF
--- a/recipes/sbt-mode
+++ b/recipes/sbt-mode
@@ -1,3 +1,3 @@
 (sbt-mode
- :repo "ensime/emacs-sbt-mode"
+ :repo "hvesalai/emacs-sbt-mode"
  :fetcher github)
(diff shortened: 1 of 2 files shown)


hvesalai:
This pr doesn't remove ensime. Is that a problem? Will there be a conflict for sbt-mode and scala-mode that were part of ensime before?

hvesalai:
@tarius this needs to be merged. The emacs-scala-mode repository name was wrong in pull request #6465

tarsius:
Ah crap... I opened a pr in order to run the tests to avoid exactly this kind of mistake and then I didn't wait for the tests to complete?!

tarsius:
Fixed with eb0b5735e9d930502ea7346e29d350ba8068440c.

PR #6459, opened 2019-09-27
Updated logpad.el

dertuxmalwieder:
I got rid of Bitbucket. No code changes since the initial commit.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/logpad
+++ b/recipes/logpad
@@ -1,2 +1,2 @@
-(logpad :fetcher bitbucket
-        :repo "tux_/logpad.el")
+(logpad :fetcher github
+        :repo "dertuxmalwieder/logpad.el")


tarsius:
Could you do something to prove that you are the same person as **tux_** on bitbucket?

* https://bitbucket.org/tux_/logpad.el says that the repository now lives at https://code.rosaelefanten.org/logpad.el.

* https://code.rosaelefanten.org/logpad.el tells the user to "Grab it on MELPA!" ...

* ... however it also says "This page was generated in about 0.003s by **Fossil** 2.8 [f8d7f76bfd] **2019-02-20** 15:01:32" ...

* ... so this would have given an attacker a large time frame to take over this package.

Can you modify the first page (or else the second) to say that this request is legit?

dertuxmalwieder:
The first one is generated automatically. The second one is the Fossil build date... it will be changed when my server has Fossil 2.9.

You could compare the code...?

dertuxmalwieder:
https://bitbucket.org/tux_/itsreallyme/

Here you go

tarsius:
Thanks!

PR #6460, opened 2019-09-27
Add package-former.el ^_^

alhassy:

This program is intended to reduce the burden in selecting
the form of records and other kinds of packages in Agda.
For example, the decision of whether a record element should be
declared as a field or as a parameter no longer needs to be performed
prematurely but rather may be selected when necessary.

https://github.com/alhassy/next-700-module-systems

I am the creator. 

None needed

### Checklist

Please confirm with `x`:

    - [ ] But with many warnings.  
 
    - [ ] Mostly, it dislikes that I have extra (desirable) indentation.
    
### DIFF
--- a/recipes/agda-next-700-module-systems
+++ b/recipes/agda-next-700-module-systems
@@ -1,0 +1,3 @@
+(agda-next-700-module-systems :fetcher github
+                              :repo "alhassy/next-700-module-systems"
+                              :files ("prototype/agda-next-700-module-systems.el"))
--- a/recipes/package-former
+++ b/recipes/package-former
@@ -1,0 +1,4 @@
+(package-former :fetcher github
+                :repo "alhassy/next-700-module-systems"
+                :files ("prototype/package-former.el")
+                )
--- a/recipes/package-former
+++ b/recipes/package-former
@@ -1,4 +1,0 @@
-(package-former :fetcher github
-                :repo "alhassy/next-700-module-systems"
-                :files ("prototype/package-former.el")
-                )


tarsius:
Maybe this would be more discoverable if the package name contain "agda"?

Also maybe the package name should not use the `package-` prefix because that suggest that it has something to do with `package.el` (which is Emacs package manager for emacs-lisp packages) or emacs-lisp packages in general. Like for example the `package-build` and `package-lint` packages.

riscy:
Thanks @alhassy. Also chiming in here:

> My elisp byte-compiles cleanly  - But with many warnings.

:) I think most of these should be addressed -- per CONTRIBUTING.org, I'd also strongly recommend you use [using lexical binding](https://github.com/melpa/melpa/blob/master/CONTRIBUTING.org#fixing-typical-problems) (and depend on at least (emacs "24")) as you clean these up.

The latest `package-lint` also has some warnings that you should address, and many of the 30 or so `checkdoc` warnings are legitimate.  The reason for the formatting restrictions is due to Emacs' "self-documenting" nature -- your documentation is not only viewed by people reading your code, but by people using the Emacs online help.

alhassy:
@tarsius I agree that the name could be misleading to other Emacs users, however a minor aim of this work is to bring the whole of Emacs and Emacs Lisp to the attention of Agda users ---a large number of whom mainly use Emacs as an Agda-IDE and nothing more. As such, I've tried to name that package for the task it serves: An implementation of the package-former research proposal.

( In particular, the package exposes some functions *without* the chosen prefix `pf-` since they are intended to be used by Agda users in the process of creating their own Agda packages, written in ELisp but running in Agda. ) 

I'm not committed to the name and would be happy to change it. I'm considering `agda-meta-modules`, `agda-bundling-mechanisms`, `agda-package-former`. Any suggestions would be appreciated ---perhaps members of the Agda core community may advise on a name; @jespercockx, @nad, @UlfNorell.

@riscy I'm happy to make the minor accommodations of depending on emacs 24, and using lexical binding. The former is necessary for my use of, say, `when-let`, which is why I currently depend on 24.4, but the latter however is not necessary. The checkdoc warnings are all about the extra indenation, which is considered to be poor Lisp doc style, but I thought as *style*, it would be no real issue if I deviated ever so slightly by adding in the desirable white space. Unless strongly urged, I'm not inclined to remove the whitepace ---I like the indentation as it is consistent with other indentation rules of Lisp constructs.

@tarsius @riscy Thank-you for your feedback; please advise of any necessary next steps.

jespercockx:
@alhassy You are correct that many Agda users see Emacs mainly as an Agda-IDE, I am one of them :) (though I also use it for Haskell and LaTeX). 

Could you point me to where I could find some more detailed description of what this package does exactly and perhaps some example use cases? Maybe then I would be better able to help picking a name for it.

tarsius:
> Unless strongly urged, I'm not inclined to remove the whitepace
> ---I like the indentation as it is consistent with other indentation
> rules of Lisp constructs.

I am afraid we're gonna have to insist.  This is not how `describe-function` should display this:

```
pf--substring-delimited-here is a Lisp function in
`/tmp/next-700-module-systems/prototype/package-former.el'.

(pf--substring-delimited-here CONTEXT STRING)

Assuming ‘context’ = “⟪prefix⟫ $here ⟪suffix⟫”
   and ‘string’ ≈ ⋯‘prefix’⟪needle⟫‘suffix’⋯, return the /first/ such needle.

  NOTE: ⟪prefix⟫ and ⟪suffix⟫ cannot be empty strings!

  We convert all adjacent whitespace
  characters to a single space in the input ‘string’ and trim any surrounding
  whitespace from the resulting output needle string.
  

[back]
```

It should look like this:

```
pf--substring-delimited-here is a Lisp function in
`/tmp/next-700-module-systems/prototype/package-former.el'.

(pf--substring-delimited-here CONTEXT STRING)

Assuming CONTEXT = “⟪prefix⟫ $here ⟪suffix⟫”
and STRING ≈ ⋯‘prefix’⟪needle⟫‘suffix’⋯, return the /first/ such needle.

NOTE: ⟪prefix⟫ and ⟪suffix⟫ cannot be empty strings!

We convert all adjacent whitespace characters to a single space
in the input STRING and trim any surrounding whitespace from
the resulting output needle string.
  
[back]
```

The remaining quoting stuff is highly unusual for emacs-lisp, but I guess it is an agda thing and its use justified.  Or maybe `⟪prefix⟫` should be written the emacs-lisp way as `PREFIX`.  You probably should not use these characters `“”‘’` literally.

Also the first line should be a complete sentence, which does not continue on the next line, which ends with a period, and which if at all possible isn't longer than 74 characters.  Maybe something like "Return the first needle from CONTEXT as specified by STRING works." I don't really understand what this does.

Its quite difficult for me to know what you should be doing.  You are violating some basic rules where there is very strong consensus that you should not do what you are doing and that makes it hard to know whether some other unusual styling is actually justified or even necessary.  So I would suggest you read the rules carefully and try to stick to them.

Also don't put a newline for the closing `"`.

So you probably should write something like this as:

``` emacs-lisp
(cl-defun pf--substring-delimited-here (context string)
  "Return the first needle from CONTEXT as specified by STRING works.
  
Assuming CONTEXT = \"PREFIX $here SUFFIX\"
and STRING = \"... PREFIX NEEDLE SUFFIX ...\", return the first such needle.

NOTE: PREFIX and SUFFIX cannot be empty strings!

We convert all adjacent whitespace characters to a single space
in the input STRING and trim any surrounding whitespace from
the resulting output needle string."
  (-let [pre-post (s-split "$here" context)]
    (pf--substring-delimited (s-trim (car pre-post)) (s-trim (cadr pre-post)) string)))
```

If it is a big issue for you that the "Return" and "Assuming" do not begin in the column, then you could write (which is unusual but okay I guess):

``` emacs-lisp
(cl-defun pf--substring-delimited-here (context string) "\
Return the first needle from CONTEXT as specified by STRING works.
  
Assuming CONTEXT = \"PREFIX $here SUFFIX\"
...
```

alhassy:
@jespercockx Please see here, which also links to a recent GPCE19 paper: https://alhassy.github.io/next-700-module-systems/prototype/package-former.html

@tarsius The final form with `\"` seems like a good place for a compromise. I will make such changes promptly. 

Thank-you.

jespercockx:
@alhassy That looks very impressive, I would like to play around with it once I find some time. For the name, may I perhaps suggest `agda-700`? Otherwise I also like `agda-meta-modules` from your suggestions.

JacquesCarette:
`agda-meta-modules` seems like a better name than `agda-700`, both for findability and expressiveness of purpose.

riscy:

No issues! 💯

```
65 issues found:

1:1: warning: The package summary is too long. It should be at most 60 characters.
54:1: error: "pf--get-indentation" doesn't start with package's prefix "agda-next-700-module-systems".
60:1: error: "pf--get-children" doesn't start with package's prefix "agda-next-700-module-systems".
123:1: error: "pf--substring-delimited" doesn't start with package's prefix "agda-next-700-module-systems".
140:1: error: "pf--substring-delimited-here" doesn't start with package's prefix "agda-next-700-module-systems".
160:1: error: "pf-folding" doesn't start with package's prefix "agda-next-700-module-systems".
164:1: error: "pf--buffer-substring-delimited" doesn't start with package's prefix "agda-next-700-module-systems".
197:1: error: "pf--buffer-substring-delimited-whole-buffer" doesn't start with package's prefix "agda-next-700-module-systems".
222:1: error: "rename-mixfix" doesn't start with package's prefix "agda-next-700-module-systems".
254:1: error: "pf-generated-suffix" doesn't start with package's prefix "agda-next-700-module-systems".
259:1: error: "pf--generated-file-name" doesn't start with package's prefix "agda-next-700-module-systems".
263:1: error: "pf--extract-imports" doesn't start with package's prefix "agda-next-700-module-systems".
267:3: error: You should depend on (emacs "25") if you need `thread-last'.
273:1: error: "λ" doesn't start with package's prefix "agda-next-700-module-systems".
321:1: error: "pf--package-formers" doesn't start with package's prefix "agda-next-700-module-systems".
324:1: error: "pf-highlighting" doesn't start with package's prefix "agda-next-700-module-systems".
332:1: error: "pf--open-pf" doesn't start with package's prefix "agda-next-700-module-systems".
362:1: error: "element-replace" doesn't start with package's prefix "agda-next-700-module-systems".
376:1: error: "parse-name" doesn't start with package's prefix "agda-next-700-module-systems".
386:1: error: "parse-elements" doesn't start with package's prefix "agda-next-700-module-systems".
435:1: error: "pf--load-package-former" doesn't start with package's prefix "agda-next-700-module-systems".
476:1: error: "pf--special" doesn't start with package's prefix "agda-next-700-module-systems".
488:1: error: "show-element" doesn't start with package's prefix "agda-next-700-module-systems".
504:1: error: "pf--show-package-former" doesn't start with package's prefix "agda-next-700-module-systems".
527:8: error: You should depend on (emacs "25") if you need `thread-last'.
533:1: error: "pf--variationals" doesn't start with package's prefix "agda-next-700-module-systems".
536:1: error: "pf-variational-composition-operator" doesn't start with package's prefix "agda-next-700-module-systems".
540:1: error: "pf--ensure" doesn't start with package's prefix "agda-next-700-module-systems".
561:1: error: "pf--wf" doesn't start with package's prefix "agda-next-700-module-systems".
601:1: error: "𝒱𝒸" doesn't start with package's prefix "agda-next-700-module-systems".
662:1: error: "𝒱-" doesn't start with package's prefix "agda-next-700-module-systems".
665:3: error: You should depend on (emacs "25") if you need `thread-last'.
671:1: error: "𝒱" doesn't start with package's prefix "agda-next-700-module-systems".
746:26: error: You should depend on (emacs "25") if you need `thread-last'.
764:1: error: "pf--load-variational" doesn't start with package's prefix "agda-next-700-module-systems".
778:3: error: You should depend on (emacs "25") if you need `thread-last'.
796:1: error: "pf--load-instance-declaration" doesn't start with package's prefix "agda-next-700-module-systems".
837:9: error: You should depend on (emacs "25") if you need `when-let'.
872:6: error: You should depend on (emacs "25") if you need `thread-last'.
937:1: error: "pf--annotations" doesn't start with package's prefix "agda-next-700-module-systems".
942:1: error: "pf--load-pf--annotations" doesn't start with package's prefix "agda-next-700-module-systems".
1029:1: error: "pf--tooltipify" doesn't start with package's prefix "agda-next-700-module-systems".
1044:1: error: "pf--insert-generated-import" doesn't start with package's prefix "agda-next-700-module-systems".
1060:1: error: "pf-waiting-for-agda-threshhold" doesn't start with package's prefix "agda-next-700-module-systems".
1066:1: error: "pf--reify-package-formers" doesn't start with package's prefix "agda-next-700-module-systems".
1152:1: error: "pf--menu-bar" doesn't start with package's prefix "agda-next-700-module-systems".
1157:1: error: "pf-enable-package-formers" doesn't start with package's prefix "agda-next-700-module-systems".
1168:1: error: "pf-disable-package-formers" doesn't start with package's prefix "agda-next-700-module-systems".
1179:1: error: "pf-package-formers-about" doesn't start with package's prefix "agda-next-700-module-systems".
1199:1: error: "pf-bare-bones" doesn't start with package's prefix "agda-next-700-module-systems".
1226:1: error: "pf-show-variationals" doesn't start with package's prefix "agda-next-700-module-systems".
1234:1: error: "pf-show-pfs" doesn't start with package's prefix "agda-next-700-module-systems".
1243:1: error: "pf-fold-annotations" doesn't start with package's prefix "agda-next-700-module-systems".
1295:1: error: "♯standard-variationals" doesn't start with package's prefix "agda-next-700-module-systems".
1299:1: error: "graph-map" doesn't start with package's prefix "agda-next-700-module-systems".
1341:1: error: "--graph-map" doesn't start with package's prefix "agda-next-700-module-systems".
1379:9: error: You should depend on (emacs "25") if you need `thread-last'.
1438:1: error: "reify-to-list" doesn't start with package's prefix "agda-next-700-module-systems".
1454:6: error: You should depend on (emacs "25") if you need `thread-last'.
1465:1: error: "to-subscript" doesn't start with package's prefix "agda-next-700-module-systems".
1477:1: error: "is-sort" doesn't start with package's prefix "agda-next-700-module-systems".
1486:1: error: "target" doesn't start with package's prefix "agda-next-700-module-systems".
1500:9: error: You should depend on (emacs "25") if you need `thread-last'.
1529:1: error: "targets-a-sort" doesn't start with package's prefix "agda-next-700-module-systems".
1581:1: error: "homify" doesn't start with package's prefix "agda-next-700-module-systems".
```

```
agda-next-700-module-systems.el:540:11:Warning: macro `pf--ensure' defined too
    late

In pf--reify-package-formers:
agda-next-700-module-systems.el:1081:26:Warning: reference to free variable
    `♯standard-variationals'
agda-next-700-module-systems.el:1131:12:Warning: reference to free variable
    `agda2-highlight-in-progress'
agda-next-700-module-systems.el:1293:1:Error: Symbol's function definition is void: λ
```

And a couple more:
- agda-next-700-module-systems.el#L133: Consider `unless ...` instead of `when (not ...)`
- agda-next-700-module-systems.el#L136: Consider `unless ...` instead of `when (not ...)`
- agda-next-700-module-systems.el#L446: Consider `unless ...` instead of `when (not ...)`
- agda-next-700-module-systems.el#L22: Prefer `https` over `http` (if possible)

riscy:
Apologies,it should be "emacs 25.1" above, not "emacs 25" for e.g. `thread-last`

alhassy:
Thank-you for your time and feedback; I have learned more about Lisp and its namespacing issues in this venture.  

The package I've written is targeted to Agda users, of which I know only 2 who are daily Emacs users and the community as a whole encourages use of Emacs as an Agda IDE and little else ---as some Agda users have complained at how "old" Emacs looks and that itself is off-putting to them; perhaps the Agda community would at least recommend Spacemacs, which has an Agda layer.

Consequently, MELPA support would only be useful to a small minority of potential users ---which does not warrant the strain on users via the necessary prefix naming conventions. Indeed, the package is intended to allow metaprogramming of Agda programs, targeting non-Lisp users, and so the names need to be unqualified for the sake of accessibility.

Again, thank-you for your time and feedback ^_^

PR #6461, opened 2019-09-28
Add awscli-capf recipe

sebasmonia:

Provide completion at point for the AWS CLI

https://github.com/sebasmonia/awscli-capf.git

Author

N/A

### Checklist

Please confirm with `x`:

- [ X `M-x checkdoc` is happy with my docstrings

NOTE: I found out some issues running the completion data refresh under Linux. I saw the pre-packaged data as optional but until I can work out the issue I prefer to include in the package for better usability, hence the change in the recipe.

### DIFF
--- a/recipes/awscli-capf
+++ b/recipes/awscli-capf
@@ -1,0 +1,1 @@
+(awscli-capf :repo "sebasmonia/awscli-capf" :fetcher github)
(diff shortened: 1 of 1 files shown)


sebasmonia:
For more details see https://github.com/sebasmonia/awscli-capf/issues/7

tarsius:
The original subject was very confusing. This pull-request does not improve a pre-existing recipe, it adds a new recipe.

tarsius:
It gets more confusing.

#6369 (which has already been merged) adds the same package and both prs were opened by the author of the package. I am closing this as it appears to be a duplicate.

PR #6465, opened 2019-09-29
Remove ensime, and return scala-mode and sbt-mode to original author

tarsius:
Replaces #6443 and #6456.

### DIFF
--- a/recipes/ensime
+++ b/recipes/ensime
@@ -1,2 +1,0 @@
-(ensime :fetcher github
-	:repo "ensime/ensime-emacs")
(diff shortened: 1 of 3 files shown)


tarsius:
I did rebase the commit from #6443, adjusted the paths, and made cosmetic changes. I would then have pushed to @fommil's branch from which #6443 was created, but that is not possible because he did not use a dedicated feature branch but pushed straight to `master`. Maintainers can never push to the contributors `master` branch (or whatever is the default branch) even when the contributor normally allows that.

So I had to create a new pull-request and #6443 has to forever remain "unmerged".

If you want to learn more about this, see https://github.com/magit/magit/wiki/Dedicated-pull-request-branches.

PR #6468, opened 2019-09-30
Add recipe gotn

yauhen-l:

Utility tool for Go programming language to run test at point.

https://github.com/yauhen-l/gotn

Maintainer

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/gotn
+++ b/recipes/gotn
@@ -1,1 +1,1 @@
-(gotn :fetcher github :repo "yauhen-l/gotn" :files ("gotn.el"))
+(gotn :fetcher github :repo "yauhen-l/gotn")
--- a/recipes/gotn
+++ b/recipes/gotn
@@ -1,0 +1,1 @@
+(gotn :fetcher github :repo "yauhen-l/gotn" :files ("gotn.el"))


riscy:
I believe you can safely use the default recipe (see README.md)

yauhen-l:
@riscy I dropped `:files` in recipe to use defaults.

riscy:
Hi, thanks!

The URL listed in your front matter, https://github.com/yauhenl/gotn, doesn't appear to lead anywhere.

Have you considered adding this behavior to the [gotest](https://github.com/nlamirault/gotest.el/tree/36e09a6bf1face4c56d4a7707935c992786e0076) package?

As for the code itself, one quick thing I see:
```
gotn.el:44:1:Warning: Lexical argument shadows the dynamic variable mode-name
```
You can rename the variable to `_mode-name` since it is unused, which fixes that.

In the declaration of your major mode, you have:
```elisp
(define-derived-mode gotn-mode compilation-mode "gotn"
  "Major mode for the gotn compilation buffer."
  (use-local-map gotn-mode-map)
  (setq major-mode 'gotn-mode)
  (setq mode-name "gotn")
  (setq truncate-lines t)
  (font-lock-add-keywords nil nil))
```
But you don't need to `setq` on `major-mode` or `mode-name` -- these should be set automatically when using `define-derived-mode`.

Hope that helps, and let me know your thoughts on `gotest`.

yauhen-l:
Hi @riscy ,
Thank you very much! I've fixed all issues. A funny part with `setq on major-mode or mode-name` that I took it from `gotest`: https://github.com/nlamirault/gotest.el/blob/master/gotest.el#L141
Regarding `gotest`:
`gotest` uses Elisp to parse and identify Go tests names, while `gotn` uses general Go approach of providing simple Go program for this task, so it can be used by any editor and not only Emacs.
Also I am not so experienced Elisp developer to implement same functionality in `gotest`.

riscy:
> gotest uses Elisp to parse and identify Go tests names, while gotn uses general Go approach of providing simple Go program for this task, so it can be used by any editor and not only Emacs.

Ah, interesting.  The ability to support editors besides Emacs is definitely a plus for `gotn`, so I think you've done nice work there, but that seems irrelevant when we're talking about MELPA packages.

> Also I am not so experienced Elisp developer to implement same functionality in gotest

To be honest, you seem experienced enough that I believe it would be within your reach.  Can I recommend you try to open an issue with them?

tarsius:
Closing due to inactivity.

PR #6470, opened 2019-10-02
Add lastfm recipe

mihaiolteanu:

[Last.fm](https://www.last.fm/api/) API for Emacs Lisp. 

https://github.com/mihaiolteanu/lastfm.el

Author and maintainer.

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/lastfm
+++ b/recipes/lastfm
@@ -1,0 +1,1 @@
+(lastfm :fetcher github :repo "mihaiolteanu/lastfm.el" :files ("lastfm.el"))
--- a/recipes/lastfm
+++ b/recipes/lastfm
@@ -1,1 +1,1 @@
-(lastfm :fetcher github :repo "mihaiolteanu/lastfm.el" :files ("lastfm.el"))
+(lastfm :fetcher github :repo "mihaiolteanu/lastfm.el")


alphapapa:
This looks cool.

Note that the way you're generating API functions is rather unorthodox.  You're defining methods/endpoints in a variable, and using a macro to map a function across the variable, defining all the variables with a single call to the macro.

The more idiomatic way would be to define a macro which defines a function, something like:

```el
(defmacro last.fm-defmethod (name args docstring &rest body)
  (declare (debug (symbolp listp stringp def-body))
           (indent defun))
  (let ((fn-name (intern (concat "last.fm-api-" (symbol-name name)))))
    `(cl-defun ,fn-name ,args
       ,docstring
       (last.fm--parse-response
        ,@body))))
```

Then you would use it to define functions, something like:

```el
(last.fm-defmethod get-top-tracks (artist &key (limit 10) (page 1))
  "Return top tracks for ARTIST."
  ("track > name" "playcount" "listeners"))
```

There are a few other unidiomatic things, like:

* Instead of `lastfm--local-file-path`, you probably want something like `expand-file-name` or some of the other built-in functions that work on paths.
* You have several functions, like `lastfm--method-params`, that just call e.g. `cl-fourth` on the argument.  If you use a macro like I recommended, you won't need those.  But anyway, you should probably use `cl-defstruct` or `cl-defclass` rather than your own pseudo-structs/classes.

mihaiolteanu:
"Unorthodox" is a nice way of putting it.  

Indeed, your method seems to be a more sane approach. Would you say that the benefits are consistency across the packages and a more easier time for future maintainers, or are there others, like easier to debug? I often times look inside a package implementation for ways of configuring it, understanding how it works, etc., and yes, I hate to see nasty surprises and clever ways of doing things if it brings no actual benefit. 

Is there a reason you're using `last.fm-api-` and not simply `lastfm-` as a prefix, or is this just an example to express the gist of the solution? I need to also have the `artist` there in the name of the method, not just `lastfm-get-top-tracks`, since there is also a `lastfm-tag-get-top-tracks`, a `lastfm-user-get-top-tracks` and a few others besides `lastfm-artist-get-top-tracks`. But that is easy to tweak.

`(expand-file-name "README.md")` just points me to `~/README.md` and not to `~/.emacs.d/list/lastfm/README.md` as I would like. Anyway, I'm still not sure how to generate the documentation for the package in a nice, seamless way. I do have the package general overview in a `README_overview.md` file. The `README_api.md` is generated when the API functions are generated and then the two files are merged in the final `README.md` file. It works nice while I'm developing the package, but the byte-compiler doesn't agree. I can't seem to bring macros under my full control yet, so I've had to comment some stuff out and only uncomment them in the development phase.

Yes, I've tried using structures for a time. It would have only added `make-method`s everywhere in the lists for no apparent gain. That's why I've decided for `cl-fourth` and friends. 

I will implement your suggestions and see how that works.

Thank you for your time and input,
Mihai

alphapapa:
> Would you say that the benefits are consistency across the packages and a more easier time for future maintainers, or are there others, like easier to debug? 

Both.  I've never seen an Emacs package do it the way yours currently does, but the way I showed is common.  It also makes debugging easier, and the code is much easier to find and understand.

> Is there a reason you're using last.fm-api- and not simply lastfm- as a prefix, or is this just an example to express the gist of the solution? 

It's just an example.  Note that you *can* put `.` in symbol names, so you could if you wanted.

> (expand-file-name "README.md") just points me to ~/README.md and not to ~/.emacs.d/list/lastfm/README.md as I would like.

There are other functions, e.g. `locate-user-emacs-file` is probably the one you're looking for.  Check the manual for more.  Also, I recommend using Helm, as it makes such functions easy to find quickly (I don't have them all memorized, and I only check the manual when I need to; otherwise, Helm makes it easy to find by typing a few related words into `C-h f`).

>  Anyway, I'm still not sure how to generate the documentation for the package in a nice, seamless way. 

That is definitely an issue.  I've tried to do something like that for some of my packages.  See:

-  https://raw.githubusercontent.com/alphapapa/org-ql/master/README.org at the bottom in the `* COMMENT Code` heading (which is hidden in GitHub's renderer)
- https://github.com/alphapapa/unpackaged.el#convert-elisp-to-org-format
- https://github.com/alphapapa/emacs-package-dev-handbook#documentation

I usually end up having to do sync some docstring changes to the readme manually, because sometimes it's quicker than redoing the whole thing and tidying up all the little things.

There's definitely space for a package that helps generate Elisp documentation cleanly, in a configurable way...

>  It works nice while I'm developing the package, but the byte-compiler doesn't agree

Sometimes you can use `eval-and-compile` or `cl-eval-when` to help with issues like that.

> Yes, I've tried using structures for a time. It would have only added make-methods everywhere in the lastfm--methods list for no apparent gain. That's why I've decided for cl-fourth and friends.

I'm not sure exactly what you mean, but using a macro like I suggested will probably obviate that problem anyway.  Note as well that Emacs's EIEIO is like Emacs's version of CLOS.  It's used extensively by some major packages like Magit.

Here are some other things you may find generally useful: https://github.com/alphapapa/elexandria

One other small suggestion: you should probably store the API rc file either in the Emacs user directory, or at least in a filename that indicates it's Emacs-related.  If you just write a Lisp list to a file and read it in, you could name it e.g. `last.fm-api.el`, even though it may not contain executable code.

riscy:
Why not use the default recipe? (See README.md)

mihaiolteanu:
Neat. That was very instructive. I've changed the implementation to use a `lastfm--defmethod` macro instead of the big list. Not only is the result much cleaner than before, it is also much shorter and easier to reason about. A keyboard macro worked nicely to transform one into the other, too. 

> I've never seen an Emacs package do it the way yours currently does.

Well, don't I deserve a medal, then?

> Also, I recommend using Helm.

I've tried it and it was the first completion framework that I've used, at least for a while. I like the speed and minimalism of Counsel better, even though is not so complete like Helm. At least the configuring of where to open the minibuffer when calling Helm is much saner now than the gymnastics required some years ago.

I've managed to add the documentation for each method, at the place where that method is defined, into a global string which will then be saved into the md file. I'm using a custom variable to only generate the documentation when it is set to `t`, so that the user will not see any new files besides the lastfm.el file if he leaves that custom var alone.

> Here are some other things you may find generally useful: https://github.com/alphapapa/elexandria

More than `alexandria` in elisp, I'm beginning to now want `dash` in cl. I'm beginning to like the anaphoras too much, maybe.

Let me know if I've missed anything,
Mihai

mihaiolteanu:
> Why not use the default recipe? (See README.md)

You mean something like `(lastfm :fetcher github :repo "mihaiolteanu/lastfm.el")`. That is, dropping the filename since there only is a single `.el` file?

riscy:
Yeah!  It's the preferred approach.

alphapapa:
> More than alexandria in elisp, I'm beginning to now want dash in cl. I'm beginning to like the anaphoras too much, maybe.

Indeed, anaphoric macros are so helpful.  You might find the `anaphora` package useful.  And if you haven't already seen them, the books *On Lisp* by Paul Graham  and *Let Over Lambda* by Doug Hoyte.

riscy:
Hi again and apologies for the turnaround.  One thing I'm seeing with the current code and byte-compilation is the following:

```
lastfm.el:217:1:Error: Symbol's function definition is void: lastfm--api-fn-name
```
Since you're a first-time contributor, I would recommend steering clear of metaprogramming if you can -- aim for explicit code.

While I'm at it, I like to check whether you have done the paperwork to sign your code to the FSF  (people often copy and paste that line)?

mihaiolteanu:
> lastfm.el:217:1:Error: Symbol's function definition is void: lastfm--api-fn-name

That's not good. I did manage to get rid of that error with an `eval-and-compile` around this `defun`. But I agree, I'm really not that confident in my understanding of the byte-compiler and what is really going on at a deeper level.

> Since you're a first-time contributor, I would recommend steering clear of metaprogramming if you can -- aim for explicit code.

The whole package is built around this macro, actually. It gets called 53 times to generate the package API. There would be a lot of copy/paste code, with insignificant differences between the code, without this macro.

> While I'm at it, I like to check whether you have done the paperwork to sign your code to the FSF (people often copy and paste that line)?

I guess I'm one of those people, then. I can get that process started, though.

Thanks for your time.

riscy:
> I guess I'm one of those people, then. I can get that process started, though.

There's no requirement to do so!  Just change the copyright to point to yourself. :)

mihaiolteanu:
> There's no requirement to do so! Just change the copyright to point to yourself. :)

Ok, I'll take the easy way out then :)

riscy:
Since I forgot to mention -- ping when you're ready for another look. :)

mihaiolteanu:
Oh, so that's what `awaiting-upstream` means..

How are new versions or updates to the package handled once the package is added to melpa? Aren't there any code reviews after that? Or is it assumed and hoped that everybody will behave and not introduce any crap?

Anyway, I'm ready for another round.

riscy:
> How are new versions or updates to the package handled once the package is added to melpa? Aren't there any code reviews after that? Or is it assumed and hoped that everybody will behave and not introduce any crap?

"Packages are versioned based on the most recent commit date to package files based on commits to upstream package repository" - which basically means, if you make a commit, your package is updated on MELPA.

There are no follow-up code reviews, but in the past I've seen some MELPA maintainers do periodic checks and open PR's against packages that could use some MELPA-related update or another.  It is hoped that everyone will behave, but it's easy to unindex packages.

riscy:
```
lastfm.el:485:63:Warning: reference to free variable `it'

In lastfm--parse-response:
lastfm.el:511:14:Warning: reference to free variable `it'

In end of data:
lastfm.el:566:1:Warning: the following functions are not known to be defined:
    alet, awhen
```

Extra hints:
- lastfm.el#L22: Prefer `https` over `http` (if possible)
- lastfm.el#L71: Prefer `https` over `http` (if possible)
- lastfm.el#L128: Prefer `https` over `http` (if possible)

mihaiolteanu:
Sadly, I don't get those warnings. Presumably, because the `anaphora` package is already required by other package(s) at startup?! As an experiment, if I remove the `(require 'elquery)`, then I do get a warning for functions not known to be defined  (`elquery-text` and `elquery-$`). Presumably, because no other package loads 'elquery?! Is this an instance of "emacs has a global namespace and that kinda sucks"?

How does one byte-compile a file with confidence, then? I've tried both to `batch-byte-compile` the file from the command line and to `byte-compile-file` from an `emacs -q` instance (no init file). In both cases, all the `require` calls fail with `no such file or directory`. If I then do an `emacs -q` (no init file) and do the minimum of initialization,

```
(require 'package)
(package-initialize)
```
then I get rid of the errors. But then, I also get no warnings for missing symbols even though `anaphora` is not a required package. Presumably, again, because it gets required somewhere else.

Anyway, I've added `anaphora` to the Package-Requires list. Now I have another weirdness to ask. This anaphora package has a v.1.0.4 [1], but if I require that version with `(anaphora "1.0.4")` I get an error when invoking `package-build-current-recipe` saying that there is no such version available. So I use `(anaphora "1.0.0")` which gets the `20180618.2200` version, which is fine. But this seems to be a newer version than 1.0.0, since 1.0.0. is from 2013 and `20180618.2200` was byte-compiled in 2018, if I get these things right. So how do you choose these versions with confidence when adding to Package-Requires?

I've also replaced `http` with `https`, so I'm ready for another round.

[1] https://github.com/rolandwalker/anaphora/releases

riscy:
> Presumably, because the anaphora package is already required by other package(s) at startup?!

In this case I think it was anaphora using the `###autoload` cookie on its macro definition of `anaphoric-let` -- so if you had anaphora installed (and you did on your machine), then `alet` was defined whether or not you explicitly loaded the anaphora library (see: [when to autoload](https://www.gnu.org/software/emacs/manual/html_node/elisp/When-to-Autoload.html) for discussion).

> How does one byte-compile a file with confidence, then?

The flycheck linter works quite well, so if you haven't enabled that I'd recommend it.  When I'm reviewing these PR's, I run the MELPA checks checks with a script inside a Docker container with a minimal emacs installation.

> Now I have another weirdness to ask [...]

This is beyond my reckoning right now. :)  It _should_ work with 1.0.4 and https://stable.melpa.org has indexed the 1.0.4 version, so I'm at a loss.

riscy:
Okay, I took another look --

First, this line:
```elisp
(lastfm--generate-documentation (file-name-directory load-file-name))
```
...prevents the buffer from being using the `eval-buffer` command, since `load-file-name` is not defined in that context.  Instead you can use something like this to consistently get the filename:
```elisp
(if load-file-name (file-name-directory load-file-name) default-directory)
```

Second, the default value of the customizable variable `lastfm-enable-doc-generation` is `'nil` but you can just write `nil`.

It's not a blocker, but documentation (and documentation-generating-code) generally doesn't need to ship with the package and could be factored out into another `.el` file, for example under a `docs/` directory.

Once the first and second are fixed, I think this is looking good to go.

mihaiolteanu:
Fixed the first and the second. I'll keep in mind the doc generation hint for future packages.

I've clarified the v.1.0.4 not being available. I was under the impression that `package-build-current-recipe` also installs the missing packages. I only get that error from above if I remove the `anaphora` package from `~/.emacs.d/elpa`. I've updated the package to use `1.0.4`, as a result.

Thanks for your patience and good advice.

mihaiolteanu:
Hello! :)

riscy:
I think this is a non-blocker, but this line:
```elisp
    (when (yes-or-no-p "Did you grant the application persmission
to access your Last.fm account? ")
...
```
Should probably be rewritten to remove the newline from the string.  You could write it as:
```elisp
    (when (yes-or-no-p (concat "Did you grant the application permission "
                               "to access your Last.fm account?"))
```
if you want to avoid the long line.  Also, `persmission` -> `permission`

In any case, this looks good -- and, welcome to MELPA!

PR #6476, opened 2019-10-09
Added ivy-backup

walseb:

This allows the package helm-backup to be used with ivy

EDIT:
This allows the package provides an interface for git-backup using ivy

https://github.com/walseb/git-backup-ivy

~https://github.com/walseb/ivy-backup~

Author

**None needed**
(I attempted to get this into helm-backup itself here https://github.com/antham/helm-backup/pull/35
But I think the extra modularity is good)

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/git-backup-ivy
+++ b/recipes/git-backup-ivy
@@ -1,0 +1,1 @@
+(git-backup-ivy :fetcher github :repo "walseb/git-backup-ivy")


walseb:
There is if @antham were to split `helm-backup` up into `backup` and `helm-backup`. `helm-backup` doesn't actually use helm for anything other than the frontend so it shouldn't be hard (but would take some time to do since you would have to do a lot of renaming, submitting a new package to melpa, etc), but that's up for him/her to decide. I don't mind it anyways

riscy:
Sorry for the two-week delay.  I see the situation.

It would be ideal to have an independent `backup` package that both yours and helm-backup were based on, but it's not fair to burden the helm-backup maintainers with that -- and they really seem focused on helm.

This is just my personal take, but I think the next best solution would be for you to copy-paste the relevant backend code into your repo (helm-backup is only about 150--200 lines of code unless I'm not seeing something) and acknowledge them accordingly.  You lose the benefit of their implied maintainership, but in this case I think it's better than having your package glued to helm.

walseb:
Yeah I could do that but the problem is that the package is complex with long and advanced CLI commands mixed with code. And I don't think I could effectively maintain the package if something broke without putting in a lot of time to learning how everything works

walseb:
Yeah I could do that but the problem is that the package is complex with long and advanced CLI commands mixed with code. And I don't think I could effectively maintain the package if something broke without putting in a lot of time to learning how everything works

walseb:
Also translating commits made in `helm-backup` to `ivy-backup` is not something I want to do

riscy:
Okay, fair enough.  I dug around a little more.  At least according to this issue [here](https://github.com/antham/helm-backup/issues/12), @antham appears open to the idea of splitting the backend from the frontend (see the last couple items of the list in the issue).  That's a few years old, of course, but it's encouraging.

Maybe you could see if @antham's willing to work with you on that in the near term?  If it's a clear no, then I'll concede that the transitive dependency on helm is okay, if not the best outcome for users.

antham:
Yes it was an idea from some years ago, I never took the time to work on this, I'm going to check now what I can do about that.

riscy:
Just pinging the participants here. :)

antham:
Yes I've almost finished the creation, I still have some CI issues

antham:
Here we go

riscy:
I think we're all set with https://github.com/melpa/melpa/pull/6581 merged.

walseb:
Great! First I'll have to rewrite this, I will try to do that tomorrow.

walseb:
Okay so I rewrote the package and everything should be good to go now

riscy:
Awesome.  Thanks both -- this is ready to go.

Non-blocking, but you might want to change your summary line from `A ivy interface to git-backup ` to `An ivy interface to git-backup`

walseb:
Thanks! Will do!

PR #6479, opened 2019-10-10
add zero-input package.

sylecn:

zero is a Chinese input method framework written as Emacs minor mode. A pinyin input method is included with zero.

https://gitlab.emacsos.com/sylecn/zero-el/tree/pkg

I am maintainer of the package.

**None needed**

### Checklist

Please confirm with `x`:

  Apache License Version 2.0

package-lint still reports two errors
```
206:38: error: You should depend on (emacs "25.1") if you need `window-absolute-pixel-position'.
211:11: error: You should depend on (emacs "25.1") if you need `frame-edges'.
```

These two functions are tested using functionp, a fallback
function will be used if they are not available.

  zero-framework.el still have 5 Errors. Other files are clean.
  checkdoc doesn't like `zero-mode' in docstring, which I don't know how to fix. Adding function or symbol prefix would only make the docstring worse. For example
```emacs-lisp
(defvar zero-mode-map
  ;; content omitted for brevity
  nil
  "Keymap for `zero-mode'.")

(defun zero-on ()
  "Turn on `zero-mode'."
  ;; content omitted for brevity
  )
```
  nope :)

// edit: updated info on package-lint.
// eidt2: update git repo url to link to pkg branch.

### DIFF
--- a/recipes/zero-input
+++ b/recipes/zero-input
@@ -1,0 +1,4 @@
+(zero-input
+ :fetcher git
+ :url "https://gitlab.emacsos.com/sylecn/zero-el"
+ :branch "pkg")


tarsius:
If the package is named `zero`, then there should also be a library named `zero.el`.

recipes/zero
```diff
@@ -0,0 +1,5 @@
+(zero
+ :fetcher git
+ :url "https://gitlab.emacsos.com/sylecn/zero-el"
+ :branch "master"
```
riscy: No need to specify `master` branch.

sylecn:
Thanks for the review.

I thought package lint was only applicable to zero-pkg.el. Didn't know it
checks much more than metadata for package.el. That's why I didn't get
those errors in the first place.

I will look into the issues.

About the one file vs multi-file package. For development, I would like
different modules to be in separate files. It's easier to read, git
diff/git log would be more meaningful, also makes it easier to add new
input methods. Is it good practice to add an automated build step which
would generate a single file for package.el style distribution? That will
make running checkdoc/package-lint easier for review for example. Similar
to how javascript framework build components to a single js file for
distribution. I can create a standalone git repository just to store the
generated single file zero distribution (with license file etc).

riscy:
> I thought package lint was only applicable to zero-pkg.el. 

No worries.  Actually, per CONTRIBUTING.org, it's encouraged to not have a `-pkg.el` file -- MELPA creates one automatically.

> Is it good practice to add an automated build step which would generate a single file for package.el style distribution?

That seems far too complicated.  Following the checklist **on each file** individually will be more than enough, but working with a single file will be easier for you (and for us to check).  If you're confident you can maintain MELPA standards in the multi-file format, go for it.

recipes/zero
```diff
@@ -0,0 +1,5 @@
+(zero
+ :fetcher git
+ :url "https://gitlab.emacsos.com/sylecn/zero-el"
+ :branch "pkg"
+ )
```
Silex: Don't put closing parenthesis on a line of their own...

sylecn: fixed that in a new force push.

sylecn:
Updates:
I created a dedicated branch "pkg" for one-file release. Now it's just one zero.el.

The remaining checkdoc problems is still about the usage of `zero-mode' in docstring. 

The remaining package-lint problems:
```
206:38: error: You should depend on (emacs "25.1") if you need `window-absolute-pixel-position'.
211:11: error: You should depend on (emacs "25.1") if you need `frame-edges'.
```
These two functions are tested using functionp, a fallback function will be used if they are not available.

sylecn:
I notice awaiting-upstream label is still applied. If there is something I can do please let me know.

tarsius:
All reviewers are busy at the moment. Please hold.

alphapapa:
Looking at the packages in MELPA, there isn't a strong convention, but at least some of the input-mode packages have names ending with `-input`.  I recommend that this package be either prefixed or suffixed with `input` to help identify it as an input-mode package.  One of the biggest obstacles to packages being used is discovery and identification by potential users, and descriptive names help with that.

sylecn:
Hi, I have renamed the package to zero-input.el

- file renamed. All symbols/functions renamed. Build scripts updated.
- package tested locally, works fine after rename.
- updated package receipt in the pull request to use the new name.

riscy:
Taking a closer look at this now, thanks for your patience, and for putting everything in a single file -- it makes things quite a lot easier.  A few minor comments:

- zero-input.el#L17: Use `Version` instead of `Package-Version` (MELPA automatically adds `Package-Version`)
- The variable `zero-input-full-width-mode` might be renamed to `zero-input-full-width-p` so as not to confuse it with a major/minor mode-enabling function.  You could rename the function `zero-input-toggle-full-width-mode` to `zero-input-toggle-full-width` correspondingly without any loss of clarity, I feel.
- A function like `zero-input-toggle` should not be necessary -- simply calling `zero-input-mode` should do the toggling on and off.  Inside the function `zero-input-mode` you can check the variable `zero-input-mode` and run the turn-on or turn-off behavior depending on that variable's value.
- If variables like `zero-input-punctuation-level` are always going to be local, you should use `defvar-local` instead of `defvar` everywhere, and then you won't need to call `make-local-variable 'zero-input-punctuation-level` in the definition of `zero-input-mode`.
- You have a a couple sections called `;; test data` in your code, but it's generally recommended not to ship test code with your package -- especially the stuff around `zero-input-table-set-table` modifies Emacs' state in an unnecessary way when the package is merely loaded.
- You have `(provide 'zero-input-pinyin)`, `(provide 'zero-input-pinyin-service)`, `(provide 'zero-input)` but only the latter is required (no doubt a result of catting your files together -- but maybe easy to fix).
- Minor typo: `"Call METHOD on zero-input-pinin-service asynchronously.` -- `pinin` should be `pinyin` and you could add ``` `emacs-style' ``` quotes around that.

Hope that helps!

sylecn:
riscy, thanks for the review. Some quite good advice there.

Here are changes I have made.

- dropped Package-Version metadata. Version was there from the start.
- zero-input-full-width-mode renamed to zero-input-full-width-p.
- zero-input-toggle is now an obsolete alias for zero-input-mode.
- use defvar-local when variable is always buffer local.
- moved demo/test data to debug/test files. Not shipped in release code.
- Comments on keeping just one `provide' call for zero-input.el

  It's easy to fix in build script. But I have kept multiple provide calls for
  the following reasons:

  - All features provided are in fact independent modules. They can be
    optionally loaded, they could have alternative implementations, and some
    can be reused in other emacs lisp projects.
  - Test file for each feature only require that feature, not all of
    zero-input.el. If only one provide is used, test file have to require the
    whole thing, or I will need different require call for development code
    and released code.
  - The one file zero-input.el is only created for easier distribution. I
    would like to switch to multi-file distribution when package-lint supports
    multi-file packages.

  If possible, I would like to keep the multiple provide calls.
- fixed typo. zero-input-pinin-service > `zero-input-pinyin-service'.
- build & tested v2.0.4 on localhost.

Please let me know your thoughts on the multiple `provide' calls.

Thanks again for your review.

riscy:
In my humble opinion that shouldn't be a problem, since each of them starts with `zero-input`.

Welcome to MELPA!

tarsius:
> // eidt2: update git repo url to link to pkg branch.

This was a feature branch right? You didn't do that on `master` because this had the potential of breaking things and such. But now you are done with this. The package is now named `zero-input` and it consists of a single library `zero-input.el`. I think `pkg` should now be merged into `master`. Then `pkg` should be deleted and the Melpa recipe adjusted to not specify a branch anymore. You might also want to rename the repository to https://gitlab.emacsos.com/sylecn/zero-input-el (or drop the `el` prefix). If you do, then adjust the recipe accordingly as well.

sylecn:
pkg branch is not a feature branch. It is a branch dedicated for package.el style single file distribution.

The zero-input.el file is built from multiple files in master branch. In master, type "make" will create/update zero-input.el in master branch. The same file is then copied and committed in pkg branch when a release is made.

I didn't use master branch because when melpa build the package, extra el files may interfere the build process, such as byte compile file ordering, including unnecessary files etc.

I will think about the repo renaming later.

tarsius:
I strongly urge you to give up on these unnecessary complications.

Above it was suggested that you use a single file. I have only skimmed the discussion above and have not looked at the code. But regardless of whether doing these things would result in my seconding the suggestion to use only a single file or not, I would accept your decision to stick with using many files. It may "unnecessary" here and even "unidiomatic".

So if that's what you want to do, then please stick to using multiple files!

Neither `package.el` nor Melpa enforce some silly restriction that a package must consist of only one library. We do however very strongly prefer to build the Melpa packages from the same "version" of the package that is also used by the developer and users who get the package directory from the git repository.

sylecn:
The only problem of using multiple files is package-lint check would fail on some files because it doesn't support multi-file package yet. e.g. it requires duplicated metadata on every .el file, enforcing package prefix based on file name instead of package name, etc. If "pass package-lint check" restriction could be lifted, I would be happy to make zero-input as a multi-file package. In that case, pkg branch will no longer be needed.

Relevant package-lint issue:
Support multi-file packages
https://github.com/purcell/package-lint/issues/111

tarsius:
That's really a defect of `package-lint`. If `package-lint` is spewing false-positives, then that is no reason to reject a package. I am perfectly fine with such a package but lets quickly summon @purcell just in case. @riscy should have another look as well. (This may take a while.)

riscy:
I'm tracking this -- I did write this:

> That seems far too complicated. Following the checklist on each file individually will be more than enough

But then of course it was me who merged it. :)

I think my comment about using a single file came from a bit of frustration with people under the assumption that they only had to run package-lint (and checkdoc, and byte-compile) on a single -- and sometimes arbitrary -- file in their package.

sylecn:
I have created (and tested) a feature branch multi-file-pkg which hosts multi-file package for zero-input.

https://gitlab.emacsos.com/sylecn/zero-el/tree/multi-file-pkg

This branch is also tested works with melpa (make recipes/zero-input)

If multi-file package is preferred, this is what I would submit to melpa.

PR #6481, opened 2019-10-10
Add git-walktree recipe

10sr:

A Git ls-tree browser with following features:

- Browse tree (and blob) objects of arbitrary revisions
  without checking them out to working directory
  - Go up and down tree objects
  - Visit blob objects
  - Switch to parent and child revisions
- Checkout (dump) tree or blob objects to current working directory

https://github.com/10sr/git-walktree-el

The maintainer.

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/git-walktree
+++ b/recipes/git-walktree
@@ -1,0 +1,1 @@
+(git-walktree :repo "10sr/git-walktree-el" :fetcher github)


riscy:
Hi!  Just a quick first pass:

- There is a file `git.el` in the repo that defines `git-repo`, `git-branches`, etc. but your package is called `git-walktree`, so these variable/function names should be prefixed correctly.  (The line `(require 'git)` feels bound to backfire.)  The functions should also have docstrings per checkdoc.  Perhaps this file's contents could be inlined into the only place you require it.
- In git-walktree.el, "You should depend on (emacs "26.1") if you need `replace-buffer-contents'."
- I would recommend making this package into a single file.  It will make the linters more effective for you.

tarsius:
> There is a file `git.el` in the repo that defines `git-repo`, `git-branches`

There used to be at least two Emacs packages already which were named just `git` (and therefore came with a library named `git.el`). One of these packages was distributed with Git itself (not the Emacs package but *the* Git). You should not use `git.el` for a library of your own.

> (The line `(require 'git)` feels bound to backfire.)

Even ignoring the above conflict, this `require` form is not going to work. I suggest you move the library to the top-level and name it something like `git-walktree-vars.el` and adjust the symbol prefix as suggested.

Additionally you added `(git "0.1.1")` to `Package-Requires`, which definitely is wrong.

10sr:
> Additionally you added (git "0.1.1") to Package-Requires, which definitely is wrong.

No. This package depends on the [git](https://melpa.org/#/git) library. I cannot see why you think this dependency is wrong.

This file is under `tests` directory, which means that this file is used only for testing.
Currently, this file only exists to avoid byte-compile warnings. https://github.com/10sr/git-walktree-el/blob/2accf02175b1d897142e1619ef07b9f95e476ba4/Makefile#L20

Because this recipe uses the default for `:files`, this `git.el` file will not be included in the package (yes, I made sure it *actually* was not included). https://github.com/melpa/package-build/blob/master/package-build.el#L578

When testing this package loads `tests/lib/git.el`, and for actual usage it loads external `git` library.


> In git-walktree.el, "You should depend on (emacs "26.1") if you need `replace-buffer-contents'."

Thanks! I updated it.

> I would recommend making this package into a single file.

I think this library is relatively big: this library has 1226 lines in total. IMO generally it is a good practice to split code into several files if possible when a library has this level of lines. 
It is surprising to me if MELPA officially recommends always merging code into a single file, really should I do this?

riscy:
Thanks - the oversight with the git.el file was mine and some in-development tools I'm playing with to try to make reviewing a little more automatic.   I think I've created some confusion with that, so I apologize.

In this case I shouldn't have necessarily recommended merging the file.  The reason I (sometimes) recommend that is to help contributors solve issues that the linters probably would've caught.  In this case it was my mistake to lump your git.el file in with the rest.

riscy:
Ok, looks good.  The latest package-lint has a few extra lints for you (`warning: Closing parens should not be wrapped onto new lines.`) but that's not a blocker since those were added relatively recently to package-lint.

10sr:
Thanks!

Issue #6484, opened 2019-10-12
Migrate packages away from Mercurial on Bitbucket


tarsius:
About two months ago [Bitbucket announced](https://bitbucket.org/blog/sunsetting-mercurial-support-in-bitbucket) that they would drop support for Mercurial by June 1, 2020.  Lets not wait until that deadlines has flown by before we migrate.

Of course you have the option to stick with Mercurial (in which case you have to migrate to another forge) *or* you could stick with Bitbucket (in which case you have to convert to Git), but since Magit is getting quite a bit of praise and most of the maintainers already have an account on Github, I recommend you switch to Git and Github.

Below you can see all the packages that are currently using Mercurial on Bitbucket.  It turns out that most of the maintainers of these packages already have an account on Github and that the username is the same as on Bitbucket.  (I also made sure that either the legal name as also the same and/or the Github user has forked Melpa.  But please excuse me if I have summoned you by mistake.

| ☑   | Package (36)         | User (23)        | Bitbucket repository                                     | Emacsmirror repository                              |
| --- | -------------------- | ---------------- | -------------------------------------------------------- | --------------------------------------------------- |
| ☑   | kanban               | @ArneBab         | https://bitbucket.org/ArneBab/kanban.el                  | https://github.com/emacsmirror/kanban               |
| ☑   | digit-groups         | @adamsmd         | https://bitbucket.org/adamsmd/digit-groups               | https://github.com/emacsmirror/digit-groups         |
| ☑   | ahg                  | @agriggio        | https://bitbucket.org/agriggio/ahg                       | https://github.com/emacsmirror/ahg                  |
| ☒   | grin                 | @dariusp686      | https://bitbucket.org/dariusp686/emacs-grin              | https://github.com/emacsmirror/grin                 |
| ☐   | nose                 | @durin42         | https://bitbucket.org/durin42/nosemacs                   | https://github.com/emacsmirror/nose                 |
| ☑   | lognav-mode          | @ellisvelo       | https://bitbucket.org/ellisvelo/lognav-mode              | https://github.com/emacsmirror/lognav-mode          |
| ☑   | multi-project        | @ellisvelo       | https://bitbucket.org/ellisvelo/multi-project            | https://github.com/emacsmirror/multi-project        |
| ☑   | achievements         | @gvol            | https://bitbucket.org/gvol/emacs-achievements            | https://github.com/emacsmirror/achievements         |
[...]

PR #6486, opened 2019-10-15
Migrate project-root from Bitbucket to Github

piranha:
As @tarsius has requested in #6484 I have migrated my packages
from Bitbucket to Github.  The old repositories were located at:

* https://bitbucket.org/piranha/project-root


tarsius:
Thanks!

PR #6492, opened 2019-10-16
add evil keybindings to kubel

snowiow:
This package provides evil bindings to the kubel package. This is added as an additional package similar to the kubernetes -> kubernetes-evil, so that people can decide whether they want to include evil bindings or not.

https://github.com/abrochard/kubel

Contributor of the evil keybindings to kubel

**None needed**, because the preparations were already taken here: https://github.com/melpa/melpa/pull/6390

### Checklist

Please confirm with `x`:

- [x] `M-x checkdoc` is happy with my docstrings (At least it's happy soon: https://github.com/abrochard/kubel/pull/6)

### DIFF
--- a/recipes/kubel-evil
+++ b/recipes/kubel-evil
@@ -1,0 +1,3 @@
+(kubel-evil :repo "abrochard/kubel"
+            :fetcher github
+            :files ("kubel-evil.el"))


riscy:
Nice, thanks for this. :)

I see one small package-lint:
```
```

More generally, a set of of additional keybindings should probably happen in the context of a minor mode.  That gives users the option to toggle the keybindings off and on again, rather than being "stuck" after they load the package.

snowiow:
Hi, 
```
```
I made a PR for this, so this should be fixed soon. Thanks for pointing that out.

Regarding the minor mode, these are no normal keybindings. They are added to the evil minor mode as motions bindings. On the other hand I'm not very proficient in emacs, that's why I tried to be as close to similar packages, like the kubernetes package does it with it's kubernetes evil keys: https://github.com/chrisbarrett/kubernetes-el/blob/master/kubernetes-evil.el which also doesn\t define a minor mode for the evil bindings.

I also had a look into the [evil-collection](https://github.com/emacs-evil/evil-collection), which provides keybindings for a lot of other modes. There is also not a single example of a package defining these keys as a minor mode. On the other hand I also can't tell, why the addition of those keys are not handled as additional minor-modes, because I'm pretty new in the emacs world. Maybe someone more knowledgeable can answer this question, thanks in advance :slightly_smiling_face: .

Silex:
@jojojames: is this something https://github.com/emacs-evil/evil-collection should deal with?

jojojames:
We could definitely take some form of it in. Glossing over the file, some notes:

The transients will have to go and the bindings that starts the transient.

Normal state instead of motion.

riscy:
Have you considered making a PR to evil-collection, @snowiow?

I'm not extremely familiar with the nuances of defining evil keybindings, but if you're curious about an example of defining evil keybindings within a minor mode, evil-org might be a place to start.  You might want to look up the documentation on `define-globalized-minor-mode` to get the keybindings to be enabled by default.

snowiow:
> Have you considered making a PR to evil-collection, @snowiow?

snowiow:
> Have you considered making a PR to evil-collection, @snowiow?

Not yet. I wanted to put it as a package here, because the [kubernetes](https://github.com/melpa/melpa/blob/master/recipes/kubernetes) plugin also provides it's [evil keybindings](https://github.com/melpa/melpa/blob/master/recipes/kubernetes-evil) as an additional package here. But it shouldn't be a problem to include in evil-collection otherwise.

riscy:
Did you get to take a look at how org-evil handles the use of a minor mode?  One more benefit is that might apply here is that minor mode keybindings have precedence over other keybindings.

I'll try to sell you a little more.  In my opinion, good Emacs packages are modular.  Simply loading them does not change how Emacs behaves, any changes they effect can be "undone", and <kbd>M-x unload-feature</kbd> is actually effective at removing them from the environment.

riscy:
Ping!  (I think I can't insist on emulating evil-org, but just wanted to see what you thought.)

snowiow:
Hi, I had a look and it definitely makes sense to make a minor mode of it. However I didn't find the time to take a proper look into how to implement it. Apparently there are two evil-org repositories. [org-evil](https://github.com/GuiltyDolphin/org-evil) and [evil-org-mode](https://github.com/edwtjo/evil-org-mode). I followed along the latter, because it looked more intuitive to me. You can find my implementation [here](https://github.com/abrochard/kubel/pull/8). From testing it myself it works exactly like before and the minor mode appears in the minor mode list as well. Is there anything else to check, before it should be merged?

riscy:
Left some comments!  Thanks for accommodating.  Ping me whenever that PR gets merged or is ready to go.

snowiow:
@riscy the changes regarding the minor mode are now merged in the kube lrepo

riscy:
I don't think this one is worth blocking on, but just so that it looks nice in the search:

```
1 issue found:
```

Apologies for the lengthy wait on this one.  And, welcome to MELPA!

PR #6493, opened 2019-10-17
Delete unmaintained packages.

echosa:
I'm removing several packages that I had added that have been unmaintained for years.

### DIFF
--- a/recipes/emacs-setup
+++ b/recipes/emacs-setup
@@ -1,3 +1,0 @@
-(emacs-setup
- :fetcher github
- :repo "echosa/emacs-setup")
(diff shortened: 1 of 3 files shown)


tarsius:
I only meant to suggest that `emacs-setup` should probably be removed. It's the sort of package that has a high probability to fall apart as soon as the former maintainer looks away.

The other two packages that you are removing do not fall into the same category. They might continue to do their work just fine years after you stopped using them. Maybe those should not be removed?

echosa:
I think they're quite out of date and hardly used anyway. I think we can delete them, but if you insist they stay, I'll leave them.

tarsius:
Okay lets remove them.

PR #6494, opened 2019-10-18
replace yafolding with my updated fork

vindarel:
We change the source of an abandoned package to our's with fixes.

yafolding is a code folding tool based on indentation.

It is already in MELPA.

my updated fork: https://github.com/vindarel/yafolding.el

I am a minor contributor and a user, and I want to fix a feature.

This issue was raised 2 years ago (may 2017): https://github.com/zenozeng/yafolding.el/issues/34

My PR is unanswered for one year: https://github.com/zenozeng/yafolding.el/pull/38

My simple change is: replace the un-developed discover mode by a hydra.

### Checklist

Please confirm with `x`:

(yafolding is already in MELPA)

### DIFF
--- a/recipes/yafolding
+++ b/recipes/yafolding
@@ -1,1 +1,1 @@
-(yafolding :fetcher github :repo "zenozeng/yafolding.el")
+(yafolding :fetcher github :repo "vindarel/yafolding.el")


tarsius:
Have you tried contacting the maintainer by email?

vindarel:
Yes, also 6 to 12 months ago. Actually it was someone else, interested by the issue being fixed, who contacted me and the maintainer. We got no reply from him.

last ping to @zenozeng !

tarsius:
@vindarel Are you committed to maintain this package? I.e. would you look into merging the other prs too?

@zenozeng Would you be willing to transfer the repository to someone else? I have created the [Emacsorphanage](https://github.com/emacsorphanage) for that purpose.

vindarel:
I would not develop the package further, but I would consider PRs and check for code quality. (there are no other PR than mine at the moment).

I didn't know about emacs orphanage. It seems a better solution. If it's a way for the community to maintain the package it's good, if I can get push access to yafolding it would be nice too. I'd prefer yafolding to be under a community driven shelter, it doesn't have to be under my name.

tarsius:
@vindarel The orphanage seems suitable in that case. I'll make you an admin of this repo if it is transferred. If you haven't done so yet, then please send another, final ping by email.

@zenozeng If you would like to transfer the repository to the orphanage, then tell me so--I have to make you a member before you can do so. Over the last few months a few maintainers have dropped of *all* their packages because they no longer have any time/interest to maintain them themselves. If you wish then you can do that too.

vindarel:
email sent.

tarsius:
Any response?

vindarel:
no, no email or github response.

tarsius:
I did review your changes a bit and eventually I came to the conclusion that we do not really need to fork just because of this.

What you are essentially doing is adding a hydra, right? That should be done in a separate package anyway, so you might as well maintain that package in a separate repository. Am I missing something?

vindarel:
I am actually fixing a broken `discover` command, then I replace it by a hydra.

yafolding advertises this:

    (global-set-key (kbd "s-d y") 'yafolding-discover)

but it is broken: https://github.com/zenozeng/yafolding.el/issues/34

I first committed a fix to this issue and proposed a PR (still awaiting to be merged, so the bug still fools users), then I changed `discover` by a hydra.

Thanks for the review.

tarsius:
So how do you feel about my proposal about you creating a new package `yafolding-hydra` and maintaining it in a separate repository?

Sure that won't fix the `discover` integration, but that should not be used anyway. We can comment in zenozeng/yafolding.el#34, telling other users to give up on that implementation and to use your package instead.

`discover` and even more so the `makey` package it relies on became obsolete not that long after they were created. `makey` was a friendly fork of the `magit-key-mode.el` library which back in the days was used to implement Magit's popup buffers. At the time the fork happened I was already working on the replacement `magit-popup.el`.

Once that came out `discover` should have been ported to use that instead but that never happen.

Meanwhile `hydra` was released which is much more suitable and much less of a hack than `makey`. And `magit-popup` too got replaced, this time with `transient`. These later iterations of approximately the same idea learned from the major limitations and other mistakes of `magit-key-mode.el`. Since `makey` is basically a copy of `magit-key-mode.el`, including all of its issues, one really shouldn't use it anymore.

vindarel:
I went over your reviews and rebased everything in two commits:

- remove discover, explain why, give the hydra in the readme, not in the code
- one commit of linting

I don't think having a separate yafolding-hydra package is worth the trouble. I'd add hydra as a dependency, but giving a single function as snippet in the readme is good enough.

Now it would be nice if yafolding went into the orphanage repository.

> we do not really need to fork just because of this.
> Sure that won't fix the discover integration, but that should not be used anyway.

I'd like the yafolding package in melpa to not have this broken discover command, somehow.

PR #6495, opened 2019-10-18
Convert metaweblog recipe to be a multiple package in on repo recipe in org2blog's repo

grettke:
When we moved `metaweblog` [to it's own repo](https://github.com/org2blog/org2blog/issues/52) we did it with the best of intentions: software engineering best practices and reuse and all of that. It was a great idea. In practice [8 years later](https://github.com/org2blog/metaweblog/issues/8) though no other packages were using it. Only `org2blog` uses it. 

The plan is to reintegrate the `metaweblog` package into `org2blog` by

1. Move `metaweblog.el` into the [org2blog repository](https://github.com/org2blog/org2blog)
2. Convert this recipe to a ["Multiple Packages in one Repository" recipe](https://github.com/melpa/melpa#example-multiple-packages-in-one-repository) to point at the `org2blog` repo
3. Archive the [metaweblog](https://github.com/org2blog/metaweblog) repository.

Reviewed with @tarsius and this is that recipe change.

### DIFF
--- a/recipes/metaweblog
+++ b/recipes/metaweblog
@@ -1,1 +1,1 @@
-(metaweblog :fetcher github :repo "org2blog/metaweblog")
+(metaweblog :fetcher github :repo "org2blog/org2blog" :files ("metaweblog.el"))
--- a/recipes/org2blog
+++ b/recipes/org2blog
@@ -1,1 +1,1 @@
-(org2blog :repo "org2blog/org2blog" :fetcher github :files (:defaults "README.org"))
+(org2blog :repo "org2blog/org2blog" :fetcher github :files (:defaults "README.org" (:exclude "metaweblog.el")))


tarsius:
You also have to adjust the `org2blog` recipe:

```elisp
:files (:defaults "README.org" (:exclude metaweblog))
```

And I forgot to mention that you also have to add `metaweblog` to `org2blog`'s `Package-Requires`.

grettke:
Gotcha will do that.

tarsius:
I'll merge once you have updated `org2blog`'s `Package-Requires`.

tarsius:
Uh, I just noticed that you have a file `org2blog-pkg.el` and also specify `Package-Requires` inside `org2blog.el`. You should only do one or the other.

Tools that are part of Emacs and that could be used to create something like Melpa require the `*-pkg.el` file for multifile packages. But those tools are not used Melpa and we usually encourage maintainers to use the library header instead I think. Summoning @purcell.

grettke:
I do update both `Package-Requires` and `org2blog-pkg`. I do that so that anyone who wanted to build a package themselves would be able to do that using tar. That had got me wondering if the `org2blog-pkg.el` was important to MELPA and you answered my question before I finished updating the code :).

I just pushed the update.

grettke:
At some point I need to increment the `org2blog` version by +0.0.1. But in my minds eye I can't see in what order it needs to happen among the commit for the changes to the recipes and creating the tag for `org2blog`. 

Although I think that once these recipe changes are made that it will be seamless from the `org2blog` usage side of things, I won't make that increment now because I'm not totally sure. I will wait until this is finished and increment it then.

In the meantime I'm going to put it in a flow chart to see of I am getting the order right.

What do you think?

tarsius:
> Although I think that once these recipe changes are made that it will be seamless from the `org2blog` usage side of things

Yeah that should just work.

grettke:
Thank you for all of your help on this!

tarsius:
You are welcome!

PR #6498, opened 2019-10-21
Add trashed recipe

shingo256:

Viewing/editing system trash can -- open, view, restore or permanently delete trashed files or directories in trash can with Dired-like look and feel.  The trash can has to be compliant with freedesktop.org spec in <http://freedesktop.org/wiki/Specifications/trash-spec/>

https://github.com/shingo256/trashed

Maintainer

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/trashed
+++ b/recipes/trashed
@@ -1,0 +1,2 @@
+(trashed :repo "shingo256/trashed"
+         :fetcher github)


shingo256:
I have reopened the pull request.  Please check.

alphapapa:
This should probably be implemented using `tabulated-list-mode`.

shingo256:
Thank you for the great advise.  Now I know that and will try to revise my code shortly.

Silex:
@shingo256: consider using https://github.com/politza/tablist, it's a wrapper around tabulated-list which make some things easier.

shingo256:
I took a close look at both tabulated-list and tablist and ended up using tabulated-list and revised Trashed with it as I thought it's a better fit.

I realized that tablist is richer and covers some functionalities which I implemented in Trashed.  However, looks like it might be too much for this simple trash can editing, because it's not a build-in package and actually much bigger than Trashed even though it would simplify just a part of Trashed.  And also, since I would need to sort out and add necessary functionalities, the code might get more complicated from another perspective.

I also improved Trashed a little bit with functionalities such as column movement and regexp-based marking inspired by these packages, and now I'm hoping it's good to go.  I also would continue to consider using tablist in future versions at the time when I need to implement more elaborated functionalities.  Any further advice is appreciated and would also appreciate it if you could resume the checking process with this.

manuel-uberti:
@shingo256 I am not part of the MELPA team, but FWIW `trashed` looks better now and it works as expected for me. Great job! :)

shingo256:
Thank you!

alphapapa:
Great job.  I sent a few more suggestions as an issue on the repo's tracker.

shingo256:
Thank you for the bug reports and the suggestion.  I quickly fixed my code for the 2 issues and looking at your code.  Let's discuss it in the issue threads.

shingo256:
I further improved it to allow user to customize file size and time format.

riscy:
This mostly looks great, I just have a couple small comments:

- I believe you shouldn't have to define `trashed-mode-hook` -- it's defined automatically when you `define-derived-mode trashed-mode`, unless you have a specific reason for wanting it to be customizable. 
- It may be going over my head, but I'm having trouble understanding the need for `trashed-load-hook` -- really loading trashed should only happen once, but maybe I'm missing something.

Some minor suggestions:
- trashed.el#L620: It is beneficial to sharp-quote hooks; use `#'`
- trashed.el#L22: Prefer `https` over `http` (if possible)

And a couple minor typos:
- `restration` -> `restoration`, `Restiration` -> `Restoration`
- `acording` -> `according`

shingo256:
Hi,

Thank you for the advice.  

- `trashed-mode-hook`: Since it is declared with `defvar` in `define-derived-mode`, I declared it here with `defcustom` just to make it customizable and show up in the customize list.  I added a comment to explain this, but if you know any better way, please advise.

- `trashed-load-hook`: Yes, it is run just once at the very end of the trashed.el loading.  I just implemented it in the same way as `dired-load-hook` following Dired convention.  User may use it for their own customization such as function override, etc.

If I miss or misunderstand something, please let me know.  I quickly updated the code with your suggestions and corrected typos.  Thank you so much again.

riscy:
>  I just implemented it in the same way as `dired-load-hook` following Dired convention.

I just learned something about dired.  Okay, thanks for clarifying. :)

This looks good.  Welcome to MELPA!

PR #6499, opened 2019-10-22
new package sql-sqlline

matteoredaelli:

Emacs comes with a SQL interpreter which is able to open a connection to databases and present you with a prompt you are probably familiar with (e.g. mysql>, pgsql>, sqlline>, etc.). This mode gives you the ability to do that for Sqlline.

https://gitlab.com/matteo.redaelli/sql-sqlline

I am the maintainer

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/sql-sqlline
+++ b/recipes/sql-sqlline
@@ -1,0 +1,4 @@
+(sql-sqlline
+ :fetcher gitlab
+ :repo "matteo.redaelli/sql-sqlline"
+ :files ("artifacts/*.el"))


recipes/sql-sqlline
```diff
@@ -0,0 +1,4 @@
+(sql-sqlline
+ :fetcher gitlab
+ :repo "matteo.redaelli/sql-sqlline"
+ :files ("artifacts/*.el"))
```
riscy: I think this points to the wrong directory -- maybe you can move your .el files to the project root and use the default recipe?

matteoredaelli: I think I removed all warnings related to  M-x package-lint-current-buffer

riscy: But there are no `.el` files in the directory `artifacts` - your recipe needs fixing, I believe.

matteoredaelli: You had to run "make" to see files *.el in artifacts/
They are a copy from files from "src/*el" and I did not want to save them in git
In any case now I have changed my .gitignore file and you should see files in artifacts/

I hope now it is ok

riscy:
Thanks for the update!  This is looking good to me.  I have the following comments that you'll probably want to address, but none of these are a blocker:

- This bit in your commentary ```(e.g. `mysql>', `pgsql>', `sqlline>', etc.)``` should probably exclude `sqlline`, since that's what you're adding? 
- Your `Contributing` commentary has a small typo: "Please open **GitHub** issues and issue pull requests."
- I didn't know about sqlline, so a link to their repository (https://github.com/julianhyde/sqlline) might be helpful.

...and, welcome to MELPA!

PR #6500, opened 2019-10-22
Add recipe for parse-it.

jcs090218:

Basic Parser in Emacs Lisp.

https://github.com/jcs-elpa/parse-it

The maintainer.

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/parse-it
+++ b/recipes/parse-it
@@ -1,0 +1,1 @@
+(parse-it :repo "jcs-elpa/parse-it" :fetcher github)


alphapapa:
I'm a little confused.  I feel like it must take more than [51 lines of code](https://github.com/jcs-elpa/parse-it/blob/master/parse-it-c%2B%2B.el) and some regexps to parse C++ to an AST.  And this supposedly also parses XML with regexps?

As well, Emacs already comes with facilities for building parsers, but there's no mention of them in the readme.

jcs090218:
This is build for `charp-mode` and `typescript-mode` because those modes are using regexp base highlighting. For c++ and xml, I know Emacs itself has its parser for those languages? :/ Hence, the parser for those modes might missing many things (not needed). TBH, I'm not the expert on writing a parser. Just trying to solve the highlighting for certain modes that don't have the parser/AST built.

riscy:
This is pretty cool, but I can see @alphapapa's point here -- if we're not truly generating an AST, maybe we should be careful to say so.  Would it be fair to call these "regexp-based syntax tree approximations?"  (Most formal grammars need significantly more than regexps to be correctly parsed -- entire books are written on the subject -- even though a regexp might be sufficient for syntax highlighting).

> Just trying to solve the highlighting for certain modes that don't have the parser/AST built.

From what you wrote above, it sounds like the two main modes suffering from this are c-sharp mode and typescript mode.  Have you thought about opening PR's to directly improve the syntax highlighting for those modes?  I'm inferring that may be a more direct way for you to make the contribution you're aiming to make here.

Meanwhile, a quick pass through the checklist:
```
Warning (emacs): 
parse-it-util.el:61: Argument `ind' should appear (as IND) in the doc string
parse-it-ast.el:36: Argument `type' should appear (as TYPE) in the doc string
```

```
parse-it-xml.el:62:1:Warning: the function `parse-it-ast-build' is not known
    to be defined.
parse-it-typescript.el:66:1:Warning: the function `parse-it-ast-build' is not
    known to be defined.
parse-it-swift.el:44:11:Warning: reference to free variable
    `parse-it-objc--token-type'
...
parse-it-lex.el:171:1:Warning: the function `s-replace-regexp' is not known to
    be defined.
parse-it-python.el:80:1:Warning: the function `parse-it-ast-indent-build' is
    not known to be defined.
parse-it-util.el:80:1:Warning: the function `spaces-string' is not known to be
    defined.
parse-it-asm.el:103:11:Warning: reference to free variable
    `parse-it-lex--token-type'
parse-it-asm.el:105:1:Warning: Unused lexical variable
    `parse-it-lex--token-type'

In end of data:
parse-it-asm.el:116:1:Warning: the following functions are not known to be
    defined: parse-it-lex-tokenize-it, parse-it-ast-build
```

jcs090218:
> This is pretty cool, but I can see @alphapapa's point here -- if we're not truly generating an AST, maybe we should be careful to say so. Would it be fair to call these "regexp-based syntax tree approximations?" (Most formal grammars need significantly more than regexps to be correctly parsed -- entire books are written on the subject -- even though a regexp might be sufficient for syntax highlighting).

True, I agree this is a regexp based syntax tree (maybe or may not be). TBH, I don't have too much experiences on writing a parser/AST. The way I implemented this are searching token through regexp in order, hence if you have the correct order then it should gives you what you want to identify. At least at what I am observing it is actually pretty accurate. Maybe not 100 percent correct but it will goes approximately 99 percent in theory.

> From what you wrote above, it sounds like the two main modes suffering from this are c-sharp mode and typescript mode. Have you thought about opening PR's to directly improve the syntax highlighting for those modes? I'm inferring that may be a more direct way for you to make the contribution you're aiming to make here.

I have opened it to `csharp-mode` and someone else has opened it on `typescript-mode`. The maintainers are asking people to do a parser/AST like what `js2-mode` does. Hence, I came up with the universal solution. Now this package supports indentation base (python, etc.) and c-like base programming languages. 

> Meanwhile, a quick pass through the checklist:

All issues should be fixed now. :D

riscy:
>> Have you thought about opening PR's to directly improve the syntax highlighting for those modes?

> I have opened it to csharp-mode and someone else has opened it on typescript-mode

Thanks, I'm seeing an issue you raised [here](https://github.com/josteink/csharp-mode/issues/142).  I guess what I'm getting at is, you've obviously done some good thinking to get to this point, but as you stated this is about achieving better syntax highlighting, which in turn feels like it should belong with the major modes, not in an external package.  I'd feel the same about, for example, someone writing a package that defined standardized "indent-line-function"s for 25 different languages.  Not to mention how difficult it might be to convince a major-mode maintainer to add this as a dependency to their library.

jcs090218:
> Thanks, I'm seeing an issue you raised here. I guess what I'm getting at is, you've obviously done some good thinking to get to this point, but as you stated this is about achieving better syntax highlighting, which in turn feels like it should belong with the major modes, not in an external package. I'd feel the same about, for example, someone writing a package that defined standardized "indent-line-function"s for 25 different languages. Not to mention how difficult it might be to convince a major-mode maintainer to add this as a dependency to their library.

My first thought is like this.
parse-it-spec
I do agree maybe this shouldn't be an external package and should built into each major mode. But I think overall Emacs need a standard parser but not too aggressive to each programming languages/major modes. What I mean not too aggressive is to decouple from each major mode to its programming language. For instance, `js2-mode` has their own parser which is great but aren't interchangeable (No needed, just an example). `parse-it` only dose the parsing and the major-mode maintainer would have to do the rest, like syntax highlighting, syntax check, etc. 

I know your concern and yet I think this is still a better idea. I have implements the parsing as easy as possible. There are only 4 core files that people need to know how to use this.

1. `parse-it.el` - Entry.
2. `parse-it-lex.el` - Lexer functionalities.
3. `parse-it-ast.el` - Build AST after lexing.
4. `parse-it-uitl` - Other hepler.

IMO, this package is fast and not too aggressive and it should help major-mode maintainer who still having the issue to parse their source file. Implementing the new programming language parser wouldn't be that difficult, you would only need to know regular expression to identify the tokens and know the targeting programming language itself to build the AST correctly.

riscy:
> I do agree maybe this shouldn't be an external package and should built into each major mode.

Yeah, this is far more desirable as I see it -- especially since, this way, the maintainer of the major mode is the maintainer of the syntax-highlighting parser.  

But this package might be useful for writing e.g. simple static analysis tools that aren't necessarily part of a language major mode, so I can see the usefulness.

Also, forgive me, I missed something:

```
61:0: error: `parse-it-c--c-type-inc/dec-operators-token-type' contains a non-standard separator `/', use hyphens instead (see Elisp Coding Conventions).
```
but this looks about ready to go.

jcs090218:
> **Also, forgive me, I missed something:
M-x package-lint-current-buffer (using version 20191124.132):**

Fixed! :D

riscy:
```
parse-it-actionscript.el:47:11:Warning: reference to free variable
    `parse-it-c--c-type-inc/dec-operators-token-type'
```

```
parse-it-asm.el:98:11:Warning: reference to free variable
    `parse-it-c--c-type-inc/dec-operators-token-type'
```

jcs090218:
Can't believed I didn't realize these util now. Fixed!

PR #6503, opened 2019-10-25
added scrum

ianxm:

Generates a scrum board and reports from an org buffer.

https://github.com/ianxm/emacs-scrum

I am the maintainer.

Not needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/org-scrum
+++ b/recipes/org-scrum
@@ -1,0 +1,2 @@
+(org-scrum :repo "ianxm/emacs-scrum"
+           :fetcher github)
--- a/recipes/scrum
+++ b/recipes/scrum
@@ -1,0 +1,2 @@
+(scrum :repo "ianxm/emacs-scrum"
+       :fetcher github)
--- a/recipes/scrum
+++ b/recipes/scrum
@@ -1,2 +1,0 @@
-(scrum :repo "ianxm/emacs-scrum"
-       :fetcher github)


alphapapa:
This looks pretty cool.  It should probably be called `org-scrumm` rather than `scrumm` or `emacs-scrumm`,  which fits the naming convention to communicate that it's based on Org,.

ianxm:
I assume you're talking about the package name and not the repository name. I'm fine with changing it to `org-scrum`. Did you mean to use the double 'm'? I will check with the orgmode list if there is any objection to my using that name.

alphapapa:
> I assume you're talking about the package name and not the repository name.

Right, the issue is helping users find it in MELPA.

> Did you mean to use the double 'm'?

No, I must have been reading about SCUMM recently.  :)

> I will check with the orgmode list if there is any objection to my using that name.

I replied to your post there approving the request.  ;)  Seriously, though, there is no process by which to approve package names with `org-` prefixes.  There are many of them on MELPA, and new ones are added frequently.  The only way it could be a problem would be, e.g. if there were already a feature by that name in `org` or `org-contrib`.

ianxm:
Thanks for the "approval." I suppose the `org-` prefix doesn't imply it is part of org or endorsed by org, but I thought some might see it that way so I at least wanted to let the people behind org have a chance to disavow it.

I've changed the name in the emacs-scrum package and updated this PR. Let me know if you'd like me to squash the commits.

ianxm:
So, does thumbs up mean you acknowledge the name change, or does it mean you want me to squash the commits?

riscy:
> So, does thumbs up mean you acknowledge the name change, or does it mean you want me to squash the commits?

No need.

riscy:
- It feels like the `:group` of `defgroup org-scrum` should be `org` as opposed to `org-scrum` itself.

Some notes on the style -- I feel like the code could be better flattened and factored out for better maintainability:
- The lambda spanning around 256--268 almost feels like it could just be its own function; it is also underindented, which makes the region a little difficult to parse.
- Similarly with the really big lambda beginning around 133 -- this one _really_ feels like it should be a function -- it even contains a lambda within the lambda.
- Similarly with the lambda around 365--383.  And around here, why not continue to use the `let` block to define the variables instead of using `setq`?  Something like: ```elisp
            (let* ((hdg (nth 4 (org-heading-components)))
                   (bracket (string-match "\\[" hdg))               ; index of bracket character
                   (id (or (org-entry-get (point) "TASKID") "\\_\\_\\_"))
                   (owner (or (org-entry-get (point) "OWNER") "\\_\\_\\_"))
                   (est (or (org-entry-get (point) "ESTIMATED") "\\_\\_\\_")))
              (when bracket (setq hdg (substring hdg 0 (1- bracket))))
              (setq str (concat str (format "...
```

And some other very minor things:
- org-scrum.el#L111: It's safer to sharp-quote function names; use `#'`
- org-scrum.el#L112: It's safer to sharp-quote function names; use `#'`
- org-scrum.el#L157: It's safer to sharp-quote function names; use `#'`
- org-scrum.el#L158: It's safer to sharp-quote function names; use `#'`
- org-scrum.el#L249: It's safer to sharp-quote function names; use `#'`
- org-scrum.el#L266: It's safer to sharp-quote function names; use `#'`
- org-scrum.el#L130: It's safer to sharp-quote function names; use `#'`
- org-scrum.el#L6: Prefer `https` over `http` (if possible)
- org-scrum.el#L23: Prefer `https` over `http` (if possible)
- In the README, "here is are example" -> "here is an example"

ianxm:
thanks for the detailed review. All are good suggestions. I've pushed the changes.

riscy:
No trouble.  This looks good to go -- the following are non-blocking but I'll leave them for your consideration:

- org-scrum.el#L187: Single-line comments should begin with `;;`
- org-scrum.el#L190: Single-line comments should begin with `;;`

ianxm:
changes made. thanks!

PR #6504, opened 2019-10-25
Add recipe for define-it.

jcs090218:

Define, translate, wiki the word.

https://github.com/jcs-elpa/define-it

The maintainer.

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/define-it
+++ b/recipes/define-it
@@ -1,0 +1,1 @@
+(define-it :repo "jcs-elpa/define-it" :fetcher github)


alphapapa:
There are quite a few similar packages on MELPA.  It would be good to compare and contrast them in your readme.  Even better would be to improve those packages, if possible.

jcs090218:
> There are quite a few similar packages on MELPA. It would be good to compare and contrast them in your readme. Even better would be to improve those packages, if possible.

I should have done but I don't see similar packages because this packages is the combination of few packages. eg. `google-translate`, `wiki-summary`. I mean their packages doing the correct job and this package is just reusing these packages' functionalities and put it into one buffer. That's it.

I don't think I should compare `define-word` or any related dictionary packages because this isn't just dictionary. :/ I could write it on the README page but I don't see the meaning of this.

riscy:
When I use <kbd>M-x define-it</kbd> I get:

```
Debugger entered--Lisp error: (wrong-number-of-arguments ((t) (word) "Define by inputing WORD." (interactive "MWord: \ni\nP") (if word nil (user-error "[WARNINGS] Invalid search string: %s" word)) (let ((show-count (define-it--show-count))) (if (= 0 show-count) (progn (user-error "[CONFIG] Nothing to show: %s" show-count)))) (setq define-it--current-word word) (define-it--register-events word) (define-it--reset-timer)) 3)
```

The timeout interval (300 seconds I believe -- should this be customizable?) feels way too long.  For large definitions it covers my entire frame.  How do we clear this after it appears?

jcs090218:
> The timeout interval (300 seconds I believe -- should this be customizable?) feels way too long. 

Yes, this should be customize. Added!

>  For large definitions it covers my entire frame. How do we clear this after it appears?

If you are using `pos-tip`, after you moved your cursor the tip should be removed? If the tip covers you frame, you can set `(setq define-it-output-choice 'view)` to see the definition in the new buffer.

riscy:
Apologies for the delay -- I wanted to write some feedback here.

I feel like this package does what it needs to do, but @alphapapa has a point that there are similar packages on MELPA.  You've described this as a kind of meta-package, and I think there's value in those (I'm reminded of [helpful](http://melpa.org/#/helpful) for instance), but I think for MELPA inclusion we should ensure that the execution is top notch.

(And I know you have the talent to bring this package in that direction, which is why I think pushing for quality over quantity could really pay off.)

Here's an example of the output of this package:
```
[DEFINE]: elephant
=>=>=>=>=>=>=>=>=>=>=>=>=>=>=>=>=>=>=>=>=>=>=>=>=>=>=>=>=>=>
Word forms:  plural  elephants 

 1.  countable noun An  elephant  is a very large animal with a long, 
 flexible  
 nose  
 called  a 
 trunk , which it uses to 
 pick  up things. Elephants 
 live  in 
 India  and Africa.

 2.     See also   
 white elephant 
=>=================================
Elephants are mammals of the family Elephantidae and the largest existing land animals. Three species are currently recognised: the African bush elephant, the 
...
```
This is almost the same output as `wiki-summary`, but with a dictionary entry inserted above.  I have to say, the dictionary entry itself is much more raw than a package like [define word](http://melpa.org/#/?q=define%20word) in that the wordwrapping is strange and filled with extra spaces, the "countable", "noun" signifiers blur into the definition text, that sort of thing.  (Incidentally, does the `define-word` package already achieve much of the parsing you aim to do?)  In the "view" mode of your package (as opposed to the "pop" mode) I feel you have an untapped opportunity to propertize the text and include links (maybe make that "see also" text a bit more useful) that would enrich the output.

Beyond that, the default delimiters (which I realize are customizable) look haphazard to me due to their different width (`=>=>...` vs `=>====...`).  It would be more useful if the output was delimited by headers indicating which service the output was from.

Minor:

You rely on Emacs 25 and cl-lib, but I believe all the cl-lib you need is included in Emacs 25?

I still have the issue where the popup doesn't go away and blocks my entire window for 5 minutes, but that's on macOS.  Maybe `define-it-output-choice` would be safer with its default set to `'pop` instead of `'view`?

jcs090218:
> This is almost the same output as wiki-summary, but with a dictionary entry inserted above. I have to say, the dictionary entry itself is much more raw than a package like define word in that the wordwrapping is strange and filled with extra spaces, the "countable", "noun" signifiers blur into the definition text, that sort of thing. (Incidentally, does the define-word package already achieve much of the parsing you aim to do?) In the "view" mode of your package (as opposed to the "pop" mode) I feel you have an untapped opportunity to propertize the text and include links (maybe make that "see also" text a bit more useful) that would enrich the output.

TBH, I do have the issue implementing this. All dictionary services required different format/parsing process. Is not perfect yet but I think is readable now (hopefully) and I will improve the parsing process in the near future. Thanks for your advice to propertize the text, I currently do not know how to, so I might hold on to this.

> Beyond that, the default delimiters (which I realize are customizable) look haphazard to me due to their different width (=>=>... vs =>====...). It would be more useful if the output was delimited by headers indicating which service the output was from.

Not quite sure what you mean by `service the output was from`? You mean [Collins Dictionary](), [Google Translate]() and [Wikipedia]()? Currently I think I can replace it with `^L` and the `page-break-lines` yet I don't know if it will work inside popup/pos-tip. Any other suggestions? 😕 

Minor:

> You rely on Emacs 25 and cl-lib, but I believe all the cl-lib you need is included in Emacs 25?

Do you mean that I should remove `cl-lib` and not just have Emacs 25? Fixed.

> I still have the issue where the popup doesn't go away and blocks my entire window for 5 minutes, but that's on macOS. Maybe define-it-output-choice would be safer with its default set to 'pop instead of 'view? 'view instead of 'pop?

Changed! 😄

riscy:
> I will improve the parsing process in the near future.

Awesome.  Have you looked at how define-word does it?  That package's output is pretty slick.

> Not quite sure what you mean by service the output was from? You mean Collins Dictionary, Google Translate and Wikipedia? [...] Any other suggestions?

Yeah, exactly.  I'm thinking for example "Dictionary definition: ", "Wikipedia summary: ", that sort of thing.  Here's an example from helpful, since I used it as an example:
image

i.e., a nicely presented collection of information from different sources.  I'm not saying to emulate that exactly, of course.

All good on the last couple of points, thanks. :)

jcs090218:
> Awesome. Have you looked at how define-word does it? That package's output is pretty slick.

I did, I think I know how `define-word` is implemented and we have similar parsing process. I just haven't put enough work on parsing the document yet!

> Yeah, exactly. I'm thinking for example "Dictionary definition: ", "Wikipedia summary: ", that sort of thing. Here's an example from helpful, since I used it as an example:

This is a great idea! I should definitely do this! I don't know why I haven't thought about this. This is what it look like now. What do you think? 

```
--{{ DEFINE }}--

elephant

--{{ DICTIONARY }}--

Word forms:  plural  elephants 

 1.  countable noun An  elephant  is a very large animal with a long, 
 flexible  
 nose  
 called  a 
 trunk , which it uses to 
 pick  up things. Elephants 
 live  in 
 India  and Africa.

 2.     See also   
 white elephant

--{{TRANSLATION }}--

象

--{{ WIKIPEDIA SUMMARY }}--

Elephants are mammals of the family Elephantidae and the largest existing land animals.
...
```

riscy:
> I think I know how define-word is implemented and we have similar parsing process

At least on first glance it looks like word wrapping is handled correctly in the define-word package.

> This is what it look like now. What do you think?

I think it's better, but have you thought about using the `propertize` function?  Another option I've seen is people setting the buffer to org-mode and writing org-mode markup, which makes fontification and linking much easier.

jcs090218:
I made the improvements on the parsing and should work right now. This is the screenshot of the current version.

define-it-demo

alphapapa:
That looks much better.  A couple more ideas:

1.  Remove the `--{{` brackets.  Instead, just use faces to make the headings stand out.  You might even try a larger font size.
2.  Enable `visual-line-mode`, or wrap the paragraphs with, e.g. `fill-paragraph`.

jcs090218:
Updated!

> 1. Remove the --{{ brackets. Instead, just use faces to make the headings stand out. You might even try a larger font size.

Is custom variables now, anyway I removed it as default!

> 2. Enable visual-line-mode, or wrap the paragraphs with, e.g. fill-paragraph.

Fixed with `visual-line-mode` enabled.

#### Updated screenshot.

demo

riscy:
This looks quite a lot better, thanks!  The dictionary lookup is giving me blank text right now (maybe due to your most recent commit), but probably easy to fix.  In the meantime:

- define-it.el#L172: Use `re-search-*`'s built-in NOERROR argument
- define-it.el#L178: Use `re-search-*`'s built-in NOERROR argument
- define-it.el#L162: Use `search-*`'s built-in NOERROR argument
- define-it.el#L166: Use `search-*`'s built-in NOERROR argument
- define-it.el#L318: Use `search-*`'s built-in NOERROR argument

jcs090218:
Should all fixed! :D

riscy:
Great, this looks good to go and very useful.  A couple more non-blockers:

- When I try to translate from a language to itself (e.g. English to English ) it throws an error (I'm sure nobody would try this but thought I'd mention it)
- It would be nice if the `TRANSLATION` heading had the language next to it `TRANSLATION (French)` but that's obviously minor.

jcs090218:
Hurray!

> When I try to translate from a language to itself (e.g. English to English ) it throws an error (I'm sure nobody would try this but thought I'd mention it)

Hmmm, I don't know if this is the issue from me or the `google-translate` package. I will keep this in mind.

> It would be nice if the TRANSLATION heading had the language next to it TRANSLATION (French) but that's obviously minor.

Yeah, this is a great idea! I will implement this in the next release. :D

PR #6508, opened 2019-10-28
Use github fetcher for github project

thblt:
emacs-jabber is hosted on github, but used the git fetcher. This PR fixes this.


tarsius:
Thanks!

PR #6509, opened 2019-10-29
Add recipe for multi-shell.

jcs090218:

Managing multiple shell buffers.

https://github.com/jcs-elpa/multi-shell

The maintainer.

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/multi-shell
+++ b/recipes/multi-shell
@@ -1,0 +1,1 @@
+(multi-shell :repo "jcs-elpa/multi-shell" :fetcher github)


alphapapa:
The code in this package could be much simpler and easier to understand if it used tools in `cl-lib` and `dash`.  For example, instead of:

```el
(defun multi-shell-opened-p ()
  "Check if at least one shell opened."
  (let ((index 0) (shell-list (multi-shell-select-list)) (break nil) (opened nil))
    (while (and (< index (length multi-shell--live-shells)) (not break))
      (when (get-process (nth index shell-list))
        (setq opened t)
        (setq break t))
      (setq index (1+ index)))
    opened))
```
You could write:
```el
(cl-loop for (id . buffer) in multi-shell--live-shells
         thereis (get-process buffer))
```
And instead of:
```el
(defun multi-shell--get-current-shell-index-by-id (&optional id)
  "Return the current shell index by ID."
  (unless id
    (setq multi-shell--current-shell-id (multi-shell--name-to-id (buffer-name)))
    (setq id multi-shell--current-shell-id))
  (let ((index 0) (break nil) (sp nil) (fn-index -1))
    (while (and (< index (length multi-shell--live-shells))
                (not break))
      (setq sp (nth index multi-shell--live-shells))
      (when (= (car sp) id)
        (setq fn-index index)
        (setq break t))
      (setq index (1+ index)))
    fn-index))
```
You could use `-find-index`.

And instead of:
```el
(defun multi-shell-kill-all ()
  "Kill all shell buffers."
  (interactive)
  (while (not (= 0 (length multi-shell--live-shells)))
    (multi-shell-kill (nth 0 multi-shell--live-shells))))
```
You could use `mapc` and `(setq multi-shell--live-shells nil)`.

Those are a few examples, but there are other, similar "code smells" in the package.

jcs090218:
> And instead of:
```el
(defun multi-shell-kill-all ()
  "Kill all shell buffers."
  (interactive)
  (while (not (= 0 (length multi-shell--live-shells)))
    (multi-shell-kill (nth 0 multi-shell--live-shells))))
```
> You could use mapc and (setq multi-shell--live-shells nil).

I think the logic here is simply enough? I don't see there are benefits to change to `mapc`.

For others, I am not really good at either `dash` or `cl-lib`. I would learn it if I needed to, thanks!

riscy:
I like this functionality.  One minor issue is that it seems possible to get out of sync by killing a buffer without using `multi-shell-kill` or `multi-shell-kill-all`, and when that happens you get this error:
```
Debugger entered--Lisp error: (error "Selecting deleted buffer")
  set-buffer(#<killed buffer>)
  (save-current-buffer (set-buffer (cdr sp)) (erase-buffer))
* (progn (save-current-buffer (set-buffer (cdr sp)) (erase-buffer)) (kill-process (cdr sp)) (kill-buffer (cdr sp)) (setq multi-shell--live-shells (remove sp multi-shell--live-shells)) (multi-shell--correct-buffer-name multi-shell--current-shell-id))
  (if sp (progn (save-current-buffer (set-buffer (cdr sp)) (erase-buffer)) (kill-process (cdr sp)) (kill-buffer (cdr sp)) (setq multi-shell--live-shells (remove sp multi-shell--live-shells)) (multi-shell--correct-buffer-name multi-shell--current-shell-id)))
  multi-shell-kill((1 . #<killed buffer>))
```

One small nag:

```
1 issue found:

```

jcs090218:
> One minor issue is that it seems possible to get out of sync by killing a buffer without using multi-shell-kill or multi-shell-kill-all, and when that happens you get this error:

Should be sync for now.

> One small nag:

Fixed! :D

riscy:
Hi!  Taking another pass to dig a little deeper now -- thanks for your patience.  Have you had a chance to compare to some existing packages on MELPA?

- http://melpa.org/#/bshell
- http://melpa.org/#/shell-switcher
- http://melpa.org/#/helm-switch-shell

Is there an opportunity to contribute to/improve any of these?

jcs090218:
I just took a look at these packages. I think these are the differences compare to my package.

* http://melpa.org/#/bshell
  - Way more features than mine!!
  - My package automatically rename the entry by order. `bshell` doesn't.
* http://melpa.org/#/shell-switcher
* http://melpa.org/#/helm-switch-shell
  - These two packages do not handle multiple shell buffer? Only focus on switching the shell.

I could be wrong with these, correct me if I am wrong! I think `bshell` could definitely replace this package yet I think this has way too many features to me... Here I am creating a package that only build on top Emacs built-in `shell`. No other things. Hence, I would compare to [multi-term]() or [multi-libvterm](#6512).

riscy:
> - http://melpa.org/#/bshell
> Way more features than mine!!
> My package automatically rename the entry by order. bshell doesn't.

I hear where you're coming from.  Is automatically renaming the entries by order something that could be incorporated to bshell to improve it?   bshell isn't hugely featureful (weighing in at 146 lines), but perhaps there are UI simplifications you could propose/PR that would improve that package?

I think by default MELPA should reject packages that duplicate functionality, and push for collaboration on existing packages where possible (especially for small packages like these that -- while both useful -- are a couple hundred lines each).  But let me know if that sounds reasonable.

jcs090218:
Yeah, I think this is reasonable. I will close this now. And yet, thank you for letting me know `bshell` exists! :D

PR #6512, opened 2019-10-30
Add multi-libvterm recipe

suonlight:

This is similiar with multi-term.el but it integrates with emacs-libvterm

https://github.com/suonlight/multi-libvterm

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/multi-libvterm
+++ b/recipes/multi-libvterm
@@ -1,0 +1,1 @@
+(multi-libvterm :fetcher github :repo "suonlight/multi-libvterm")
--- a/recipes/multi-libvterm
+++ b/recipes/multi-libvterm
@@ -1,1 +1,1 @@
-(multi-libvterm :fetcher github :repo "suonlight/multi-libvterm")
+(multi-vterm :fetcher github :repo "suonlight/multi-libvterm")


riscy:
Are you sure you got everything on the checklist? :)

suonlight:
@riscy I have checked again and fixed. Could you please help me review again? Thank you

suonlight:
@riscy Thank you for your careful review. I learn a lot from it. Anyway, I have fixed your concerns. The buffer name has already been surrounded by asterisks. You can see on multi-libvterm.el#L160, L162, L173, L239

riscy:
Taking a deeper look now -- a couple small code suggestions:

```elisp
  (interactive)
  (if (projectile-project-p)
    (if (buffer-live-p (get-buffer (multi-libvterm-projectile-get-buffer-name)))
      (if (string-equal (buffer-name (current-buffer)) (multi-libvterm-projectile-get-buffer-name))
        (delete-window (selected-window))
        (switch-to-buffer-other-window (multi-libvterm-projectile-get-buffer-name)))
      (let (vterm-buffer)
        (setq vterm-buffer (multi-libvterm-get-buffer 'projectile))
        (setq multi-libvterm-buffer-list (nconc multi-libvterm-buffer-list (list vterm-buffer)))
        (set-buffer vterm-buffer)
        (multi-libvterm-internal)
        (switch-to-buffer-other-window vterm-buffer)))
    (message "This file is not in a project")))
```
Be careful of your indentation around ifs, because there should be a (subtle) visual indication as to which is the "if" and which is the "else".  This is actually what Emacs prefers to do if you run `indent-region` on your entire file.

If you do this, it makes sense to put the shorter conditional block first (since the "else" part has less indentation, so less of your code gets indented), and instead of a `let` followed by a `setq`, you can combine those lines:
```elisp
  (if (not (projectile-project-p))
      (message "This file is not in a project")
    (if (buffer-live-p (get-buffer (multi-libvterm-projectile-get-buffer-name)))
        (if (string-equal (buffer-name (current-buffer)) (multi-libvterm-projectile-get-buffer-name))
            (delete-window (selected-window))
          (switch-to-buffer-other-window (multi-libvterm-projectile-get-buffer-name)))
      (let ((vterm-buffer (multi-libvterm-get-buffer 'projectile)))
        (setq multi-libvterm-buffer-list (nconc multi-libvterm-buffer-list (list vterm-buffer)))
        (set-buffer vterm-buffer)
        (multi-libvterm-internal)
        (switch-to-buffer-other-window vterm-buffer)))))
```

I notice you write e.g.:
```elisp
Will prompt you shell name when you type `C-u' before this command."
```
but if you remove the ``` `/' ``` delimiters, you'll get a useful checkdoc warning telling you:
```
Warning (emacs): 
multi-libvterm.el:105: Keycode C-u embedded in doc string.  Use \\<keymap> & \\[function] instead
```
You should follow this, since not everyone will have `C-u` bound to what you expect. In this case you want to write `\\[universal-argument]` or even write in the documentation that the function takes a universal ARG (people will know what this means, see e.g. the help for `klll-line`).  But as I look at this, it seems like the function doesn't even support taking an argument?  I'd be looking for something like `(interactive "P")`, but maybe I'm missing something.

The way you are identifying which shell to run (in `multi-libvterm-shell-name`) is unusual.  I think you should follow the convention set by `term.el` and `shell.el` , e.g.:
`(or explicit-shell-file-name (getenv "ESHELL") shell-file-name)` if possible.

Hopefully those help!

Minor:
- "Will prompt you shell name" -> "Will prompt you for a shell name"
- "is not exist" -> "does not exist"
- "Close current vterm buffer when `exit' from vterm buffer." -> "when exiting from vterm buffer."
- "Go to the next|previous term buffer" -> "next|previous vterm buffer"

purcell:
I sent you some fixes in https://github.com/suonlight/multi-libvterm/pull/5, which gave me a chance to look closer and try the code out.

The docstrings talk about `C-u` to cause prompting, but the code doesn't actually do that at all. I think this needs a little more work in that regard, or just remove those parts of the docstrings! To get it to work, you would use `(interactive "p")` -- read about how that works. The arg for [this function](https://github.com/suonlight/multi-libvterm/blob/faa605936c0cc9854d27c368c0845cb89d25f386/multi-libvterm.el#L73) is unnecessary and won't do anything.

Other notes:

- The whole package should be called `multi-vterm`, not `multi-libvterm` IMO. There's already a `vterm` package, and nobody needs to know about the "lib" bit, I think. :-)
- You should probably use the `window-text-height` function [here](https://github.com/suonlight/multi-libvterm/blob/faa605936c0cc9854d27c368c0845cb89d25f386/multi-libvterm.el#L233) instead of doing your own calculations

Hope that helps!

suonlight:
@riscy Thank you again for your very careful review. I think @purcell fixed almost your concerns in his PR. I've already merged it.

@purcell I think `multi-vterm` can make others confused with `multi-term`, that's why I use the name `multi-libvterm`.  I would like to hear your ideas. The other note `window-text-height`, I'll refactor it later.

riscy:
I actually agree that `multi-vterm` would be better.  I don't think there's any risk of confusion, and probably less so (term is to multi-term as vterm is to multi-vterm).  Just my two cents.

Very tiny checkdoc here:
```
Warning (emacs): 
multi-libvterm.el:84: Documentation strings should not start or end with whitespace
Warning (emacs): 
multi-libvterm.el:84: First sentence should end with punctuation
```
but otherwise looks ready.

suonlight:
@riscy the package is renamed to `multi-vterm`. And I also fixed the docstring as well.

purcell:
Merging now, thanks. You can optionally rename the repo using Github's web interface, and everything will still work, if you like.

purcell:
Welcome to MELPA! :tada:

tarsius:
You forgot to rename the recipe file so the package was not actually build. I have fixed that now.

tarsius:
The test should not have succeeded but did:

```
Building recipes touched in commits be594de074dc3513c773d78c38e823be53061612...07dd16cda5c46abae362bdefae627377d145447e

fatal: bad object 8d513bf81c227cbf0eed65b00e50dd5834a2880b

Build successful

The command "./run-travis-ci.sh" exited with 0.

Done. Your build exited with 0.
```

tarsius:
@suonlight You might want to consider renaming the repository as well. If you do, then please open a pull-request to update the location in the recipe. Thanks.

suonlight:
@tarsius I have renamed the repo on https://github.com/melpa/melpa/pull/6805.

PR #6516, opened 2019-10-31
Add recipe for modus-operandi-theme

protesilaos:

Light theme designed for colour-contrast accessibility. Each combination of foreground and background colour values has a minimum contrast ratio of >= 7:1. This complies with the highest accessibility standard: WCAG AAA.

https://gitlab.com/protesilaos/modus-themes

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/modus-operandi-theme
+++ b/recipes/modus-operandi-theme
@@ -1,0 +1,4 @@
+(modus-operandi-theme
+ :fetcher gitlab
+ :repo "protesilaos/modus-themes"
+ :files ("modus-operandi-theme.el"))


alphapapa:
You might consider combining these themes into a single package, ala `doom-themes`.

protesilaos:
Thanks for the feedback! I just thought it would be preferable to offer them as separate packages because one is not a pre-requisite for the other. Furthermore, and based on feedback I have received, users choose one or the other.

My approach on this was further inspired by the [dired-hacks repo](https://github.com/Fuco1/dired-hacks) that I use, which offers standalone packages like `dired-subtree`, `dired-collapse`…

At any rate, I am perfectly willing to combine them into a single package if the MELPA maintainers feel it is a better choice, for reasons they may provide.

riscy:
That sounds like good reasoning to me.  I have only one comment, but this one is not a blocker:

- modus-operandi-theme.el#L24: Prefer `https` over `http` (if possible)

With that said, welcome to MELPA! :)

protesilaos:
I just addressed that lint. Thank you for this and all the work you are doing.

PR #6524, opened 2019-11-02
Add flask-mode recipe

FlexW:

Minor mode for working with Flask (https://palletsprojects.com/p/flask/) projects.

https://gitlab.com/flexw/flask-mode

maintainer

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/flask-mode
+++ b/recipes/flask-mode
@@ -1,0 +1,1 @@
+(flask-mode :fetcher gitlab :repo "flexw/flask-mode")


riscy:
No issues! 💯

```
No issues found.
```

```
flask-mode.el:144:21:Warning: `end-of-buffer' is for interactive use only; use
    `(goto-char (point-max))' instead.
```

Other:
- flask-mode.el#L175: It's safer to sharp-quote function names; use `#'`
- flask-mode.el#L246: It is unusual to derive from fundamental-mode; try special-mode
- flask-mode.el#L203: Consider `unless ...` instead of `when (not ...)`
- flask-mode.el#L92: You can use `not` or `null`
- flask-mode.el#L101: You can use `not` or `null`
- A progn isn't necessary for the `else` part of an `if`, and a progn isn't necessary underneath a `let` (in `flask-run-server` and `flask-try-setting-project-root` on first glance)
- It looks like the way you define keys (e.g. `(flask-key (kbd "C-c , f r") #'flask-run-server)`) could overwrite a user's modifications to `flask-minor-mode-map`.  Try something along the lines of: `(defvar my-map (let ((km (make-sparse-keymap))) (define-key ...) km))`")
- Adding hooks to the mode you're defining (e.g. `(add-hook 'flask-test-mode-hook...`) should be unnecessary -- you can run the code you need inside the definition of the mode.

tarsius:
Closing due to inactivity.

PR #6527, opened 2019-11-04
Add counsel-jq recipe

munen:

With `counsel-jq`, you can quickly test queries and traverse a complex JSON structure whilst having live feedback.

https://github.com/200ok-ch/counsel-jq

maintainer

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/counsel-jq
+++ b/recipes/counsel-jq
@@ -1,0 +1,3 @@
+(counsel-jq
+  :repo "200ok-ch/counsel-jq"
+  :fetcher github)


alphapapa:
Nifty "fallout" from Organice, I suspect.  :)

munen:
@alphapapa Partially^^ I've been meaning to write this package for a longer time, actually. We've been fans of jq for a longer time now[1] and integrating it with counsel just makes sense to turbo-charge it. Finally, I got around to doing it(;

1. https://200ok.ch/posts/jq-my-new-favorite-tool-to-work-with-json-on-the-command-line.html

alphapapa:
Cool.  That might even be a feature that would draw in new Emacs users!

munen:
I thought so, too.

With a little bit of documentation in the README on how to setup ivy with counsel-jq in the README this could be easily used from an otherwise vanilla Emacs without needing to know too many intrinsics of Emacs itself.

Happy to hear this hunch confirmed by someone else(;

riscy:
Very cool.  This looks ready to merge, but here are a couple quick comments:

This line:
```
      ;; This expression uses the most recent other buffer (see
```
feels like it could [also] be part of the docstring, to tell users (and not only people looking at the source) about the behavior.

```
1 issue found:

```

Should `counsel-jq` be autoloaded, as the main entrypoint to the package?

munen:
Hi @riscy

Thank you for your feedback! I made the proposed changes where applicable. Here's feedback to your comment:

> This line:
> 
>       `;; This expression uses the most recent other buffer (see`
> 
> feels like it could [also] be part of the docstring, to tell users (and not only people looking at the source) about the behavior.

You're right, this behavior should be visible to the users and not just people reading the code. Since it's the main behavior (and  That's why I had this sentence in the top level README:

> Whenever you're happy, hit =RET= and the results will be displayed to you in the buffer =*jq-json*=.

Adding it to the docstring of the interactive function `counsel-jq` is a good idea! Hence, I did that.

> M-x package-lint-current-buffer (using version 20191124.132):
> 
> 1 issue found:
> 
> 1:0: warning: The package summary should not end with a period.

Thanks! I was the most recent version of package-lint when I created the PR which was `package-lint-20191103.1154`. Thanks for continuing to push the capabilities of that package! :pray: I just updated and made sure that `package-lint-current-buffer` has no issues.

> Should counsel-jq be autoloaded, as the main entrypoint to the package?

Very good point! Done(;

Thanks for your time and feedback, it's much appreciated! :pray: 

I hope `counsel-jq` is good to merge, now. If you have further input, I'll happily iterate on it, of course^^

riscy:
> Thanks for continuing to push the capabilities of that package!

package-lint is all purcell's hard work. :)

I just realized you might consider adding a `(require 'swiper)` (just in case), although `ivy-read` is autoloaded so the compiler finds it without issue.  Not a blocker.

Thanks a lot for this.  And, welcome to MELPA!

munen:
@riscy Thank you very much for pulling it in, that makes my day! :pray: 

> I just realized you might consider adding a (require 'swiper) (just in case), although ivy-read is autoloaded so the compiler finds it without issue. Not a blocker.

I did include it, so it'll be in the next build. 
NB: For that to lint correctly, I had to change the `Package-Requires` dependency to `ivy` which makes sense, because swiper has a dependency on ivy.

PR #6529, opened 2019-11-06
Add recipe for reveal-in-folder.

jcs090218:

Reveal current file in folder.

https://github.com/jcs-elpa/reveal-in-folder

The maintainer.

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/reveal-in-folder
+++ b/recipes/reveal-in-folder
@@ -1,0 +1,1 @@
+(reveal-in-folder :repo "jcs-elpa/reveal-in-folder" :fetcher github)


alphapapa:
This package seems to hard-code certain possibilities rather than making use of built-in systems in Emacs and the OS.  For example:

```el
     ((eq system-type 'gnu/linux)
      (cond
       ((reveal-in-folder--safe-execute-p "nautilus ."))
       ((reveal-in-folder--safe-execute-p "nemo ."))
       ((reveal-in-folder--safe-execute-p "caja ."))
       ((reveal-in-folder--safe-execute-p "thunar ."))
       (t (user-error "[USER-ERROR] No desktop environment detected"))))
```
I don't have any of those installed on my Linux system, and I've only heard of two of them.  But I can already do `xdg-open ~` and have my home directory opened in the file manager I've configured with XDG settings.  And Emacs already provides tools to discover this information, e.g.

```el
(mailcap-mime-info "inode/directory") ;;=> "dolphin %i -caption %c %s"
```

jcs090218:
>This package seems to hard-code certain possibilities rather than making use of built-in systems in Emacs and the OS

You are indeed correct.

> I don't have any of those installed on my Linux system, and I've only heard of two of them. But I can already do xdg-open ~ and have my home directory opened in the file manager I've configured with XDG settings. And Emacs already provides tools to discover this information, e.g.

Cool, so does that mean `xdg-open` will work in any linux distro? TBH, I didn't have enough experience with linux yet. For function `mailcap-mime-info`, I'm not sure how you use this to get the information I need... I have tested on my Windows and Ubuntu but it only returns `nil`.

alphapapa:
> Cool, so does that mean xdg-open will work in any linux distro? 

It might, but `mailcap-mime-info` does the same thing, in what should be a cross-platform way.  I recommend studying the code in `org-open-file`, which ultimately calls `mailcap-mime-info` when it's appropriate.

jcs090218:
Cool, updated! I think is good enough as long it uses the default settings. I have tested on my Linux distro and it works.

> I recommend studying the code in org-open-file, which ultimately calls mailcap-mime-info when it's appropriate.

I will definitely do if I needed this to work on Linux. Thanks!

alphapapa:
Consider this command:
```el
(defun ffap-open-directory (filename)
  (interactive (list (ffap-file-at-point)))
  (let* ((directory (file-name-directory filename))
         (command (car (split-string (mailcap-mime-info "inode/directory") nil t))))
    (start-process "ffap-open-directory" nil
                   command directory)))
```
I don't know if it works on Windows (it likely does when using Cygwin); it probably works on Mac OS.

jcs090218:
> I don't know if it works on Windows (it likely does when using Cygwin); it probably works on Mac OS.

Unfortunately, it doesn't work on Windows. I don't have a macOS with me now so I couldn't test it. BTW, I don't prefer any external software if needed and especially for such these simple tasks.

alphapapa:
What external software are you referring to?

jcs090218:
cygwin

alphapapa:
Cygwin's just a convenient way to install Emacs and GNU userland on Windows.

jcs090218:
I know Cygwin but I still don't prefer it. And I use Windows' build version of Emacs.

riscy:
```
reveal-in-folder.el:35:1:Error: Cannot open load file: No such file or directory, f
```


Looks like we need a `Package-Requires` on `f`

jcs090218:
> Looks like we need a Package-Requires on f

Fixed!

riscy:
Thanks both -- @alphapapa some really good guidance here.

riscy:
There is an older but quite popular package that does this specifically for macOS, fyi: https://melpa.org/#/reveal-in-osx-finder although it goes through AppleScript as opposed to using the general-purpose `open` command on macOS (@kaz-yos, maybe you can comment on why that was).

Maybe there is a possibility for collaboration.

kaz-yos:
It opens the enclosing folder and highlight the relevant file in macOS Finder. I am not sure if there is a CLI way to do this. 

https://github.com/kaz-yos/reveal-in-osx-finder/blob/master/reveal-in-osx-finder.el#L29

jcs090218:
> I am not sure if there is a CLI way to do this.

In theory, I think this should be possible?

I just googled but I couldn't test it because I don't have macOS with me. Do you think these links help?

* https://superuser.com/questions/1010008/launching-finder-in-mac-os-x-terminal-and-highlighting-specific-files-selected-f
* https://stackoverflow.com/questions/39214539/opening-finder-from-terminal-with-file-selected

kaz-yos:
Wow, I was not aware of this. Indeed the `-R` (reveal) option to `open` seems to be the shell equivalent that highlights the specified file. It does function that way in a few examples I examined.

jcs090218:
> Maybe there is a possibility for collaboration.

Base on @riscy word, If we are going to collaborate, my biggest concern is to have this feature cross OSs.

> Wow, I was not aware of this. Indeed the -R (reveal) option to open seems to be the shell equivalent that highlights the specified file. It does function that way in a few examples I examined.

Maybe we can merge the two packages together if you don't mind doing it! Of course, you don't have to! :D

PR #6530, opened 2019-11-06
Add recipe magit-patch-changelog

dickmao:

Generate a patch according to emacs-mirror/CONTRIBUTING.

https://github.com/dickmao/magit-patch-changelog

### Checklist

Please confirm with `x`:

- [ ] package-lints modulo redefinition of `string-trim-left` for emacs-25
- [ ] `M-x checkdoc` is happy with my docstrings modulo "Messages should start with capital letter"

### DIFF
--- a/recipes/magit-patch-changelog
+++ b/recipes/magit-patch-changelog
@@ -1,0 +1,2 @@
+(magit-patch-changelog :repo "dickmao/magit-patch-changelog"
+                       :fetcher github)


alphapapa:
This sounds interesting, but the screenshot is missing from the readme, so I can't see what this package actually does.

Also, the usage instructions don't help much, and they seem to expect the user to have certain key bindings, which may not be the case.

It's interesting to see `easy-mmode-defmap` used to define the keymap--I haven't seen that before.  I wasn't aware that function exists.  You may need to `(require 'easy-mmode)`.

riscy:
Hi!   The screenshot/animated demo looks quite informative now.  Some quick feedback:

You depend on Emacs 25.1 (for when-let and if-let I imagine), but you also redefine string-trim-left on versions of Emacs earlier than 26.1.  However, you only need Emacs 24.4 for string-trim-left -- is there a reason it's redefined?

If you wanted, you could fix these:
```
Warning (emacs): 
magit-patch-changelog.el:563: Messages should start with a capital letter
Warning (emacs): 
magit-patch-changelog.el:582: Messages should start with a capital letter
```
by wording to, e.g. `(user-error "Failed magit-commit-diff")`.  That way you don't get used to seeing warnings, so new warnings have a better chance of getting your attention as you develop.

dickmao:
CR> string-trim-left -- is there a reason it's redefined?

I need `string-trim-left` to take more than one argument.  emacs-25's only
accepts one.

CR> That way you don't get used to seeing warnings, so new warnings have a better chance of
CR> getting your attention as you develop.

I am a big proponent of "no warnings".  My `make test` always runs
[package-lint](https://github.com/dickmao/magit-patch-changelog/blob/master/tools/package-lint.sh#L11),
granted one that is doctored to ignore the "Messages should start" warning.

riscy:
> I need `string-trim-left` to take more than one argument.

Ah, I see -- apologies for missing that.  With that said, redefining a builtin feels a little risky -- a mistake, or a decision not to byte-compile your library, could result in buggy/reduced performance for an Emacser in a global way.  (Not that I think there's a mistake, but defensively speaking..) Also if someone is <kbd>M-x h f string-trim-left</kbd> they'll be redirected to your package.  

Why not do something like this?
```elisp
(defun magit-patch-changelog-string-trim-left (string &optional regexp)
      "Backward-compatibly trim STRING of leading string matching REGEXP.
If variable `emacs-version' >= 26.1, use function `string-trim-left'.
Otherwise, REGEXP defaults to \"[ \\t\\n\\r]+\"."
 (if (version<= "26.1" emacs-version)
     (string-trim-left string regexp)
   (if (string-match (concat "\\`\\(?:" (or  regexp "[ \t\n\r]+")"\\)") string)
       (replace-match "" t t string)
     string)))
```

riscy:
> that is doctored to ignore the "Messages should start" warning.

Another option is -- rather than ignoring this check in your CI -- to write the messages as 

```elisp
(user-error "`magit-commit-diff' failed")
```
(I'll probably recommend this to people going forward unless someone convinces me it's a bad idea.)  Emacs nicely syntax-highlights ``` `magit-commit-diff' ```, you don't get the checkdoc warning, and uncapitalized messages you write in the future will not be missed by your CI.

dickmao:
[You got it!](https://github.com/dickmao/magit-patch-changelog/pull/6/files)

conao3:
Hi! I'm interested in this package too!

1. Please use proper indent in the whole Elisp file.
    Proper indent helps us understand Lisp code quickly.
2. Please add `*.elc` entry to `.gitignore` file to ignore Elisp compiled files
3. Please consider using `list` for proper align keyword instead of `backquote`.
4. Please concider using awhen instead of aif which have no else form.
    Please remove top-level sexp.
    All packages should not have side-effects just require package for user-experience. see https://www.gnu.org/software/emacs/manual/html_node/elisp/Coding-Conventions.html#Coding-Conventions

    Simply loading a package should not change Emacs's editing behavior. Include a command or commands to enable and disable the feature, or to invoke it.

I suggest all packages should have *-setup and *-teardown function if some setup is needed.

alphapapa:
> I suggest all packages should have *-setup and *-teardown function if some setup is needed.

That's what global minor modes are for.

conao3:
Yes, it makes sense.

However, the case where `*-teardown` is executed is hard to imagine...
If `*-setup` is prepared, the behavior is not changed just by requiring (this is the minimum condition), and if it also has a `*-teardown` function, the user will thank it enough.
In addition, I use the 200 Emacs package, but if one-quarter of them provide a global minor mode, the display width of the mode line will be insufficient.
(Although there are also many packages that can improve this situation ...)

However, the approach of creating a global minor mode is the certainly correct way.

riscy:
Ok, nice. :)

The last couple of nags I see are that your package summary "Git format-patch for CONTRIBUTE" feels a little ambiguous -- you could use your commentary text "Generate a patch according to emacs-mirror/CONTRIBUTE" and I think it would be clearer at a glance (and we can ignore the package-lint that triggers).

And, lastly, I've been advised to be more strict regarding licensing -- to that end, please add an [automatically detectable](https://github.com/licensee/licensee) LICENSE file to your repository (e.g. no markup).

Looks like we'll be good to go after that.

dickmao:
[You got it](https://github.com/dickmao/magit-patch-changelog/commit/ecc25954c6ea4ce0bad54ec0d75f89fb1ca39c13)

riscy:
Thanks!

PR #6531, opened 2019-11-06
Add recipe for eglot-jl

non-Jedi:

Load support for the Julia language server into eglot.

https://github.com/non-Jedi/eglot-jl

author

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/eglot-jl
+++ b/recipes/eglot-jl
@@ -1,0 +1,3 @@
+(eglot-jl :fetcher github
+          :repo "non-Jedi/eglot-jl"
+          :files (:defaults "*.jl" "*.toml"))


alphapapa:
Is the `LANGUAGEeglot` naming a convention for `eglot` packages?  If not, you might consider `eglot-jl`.  :)

non-Jedi:
Nah. There's no convention. Support for most languages is actually bundled with eglot, but there's a bit more going on with Julia, so it made sense with me to package it up separately. I think eglot-jl is reasonable, so I'll go ahead and change it.

riscy:
Sorry for the long wait on this one -- this looks great!

tarsius:
@non-Jedi Is it very important to you to say this and nothing else?:

```
;;; License:

;; To the extent possible under law, Adam Beckmeyer has waived all
;; copyright and related or neighboring rights to eglot-jl. This
;; work is published from: United States.
```

Or would you be willing to add a header keyword

```
;; License: public-domain
```

Do the above things mean different things to you? If so, is the latter a good enough approximation?

----

Melpa is supposed to only contain packages that are released under terms that are compatible with version 3 of the GPL. "public-domain" is compatible as far as I know, while I have no idea if that is also the case for the words you have used.

----

@riscy Please look into license issues before merging.

riscy:
> Please look into license issues before merging.

Ah, thanks!   I'll start checking more closely

riscy:
Just since there's no update visible here, I see the repo is [CC0](https://www.gnu.org/licenses/license-list.en.html#CC0) licensed now.

tarsius:
Unfortunately https://github.com/licensee/licensee is not able to detect that license.

@non-Jedi Please add this to your library:

```
;; License: CC0
```

Thank you.

----

Maybe you could incorporate `elx-license` from [`elx`](https://github.com/emacscollective/elx) into your workflow?

non-Jedi:
Odd that it's not picked up: https://github.com/licensee/licensee/blob/master/vendor/choosealicense.com/_licenses/cc0-1.0.txt

Will modify header and look at elx-license.

Thanks

non-Jedi:
Odd that it's not picked up: https://github.com/licensee/licensee/blob/master/vendor/choosealicense.com/_licenses/cc0-1.0.txt

Will modify header and look at elx-license.

Thanks

tarsius:
Thanks!

PR #6533, opened 2019-11-10
Add 'diff-at-point' package

ideasman42:

Improved diff navigation, using the current context when diffing the current repository.

Demo video: https://youtu.be/bR4czpEqah8

https://gitlab.com/ideasman42/emacs-diff-at-point

Maintainer.

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/diff-at-point
+++ b/recipes/diff-at-point
@@ -1,0 +1,3 @@
+(diff-at-point
+ :repo "ideasman42/emacs-diff-at-point"
+ :fetcher gitlab)


riscy:
Looks cool -- just coming through with a quick pass:

```
2 issues found:

221:18: warning: Closing parens should not be wrapped onto new lines.
```

```
diff-at-point.el:314:1:Warning: the function `diff-goto-source' is not known
    to be defined.
```

Other:
- diff-at-point.el#L302: It's safer to sharp-quote function names; use `#'`
- diff-at-point.el#L131: You can use `not` or `null`
- diff-at-point.el#L166: You can use `not` or `null`
- diff-at-point.el#L167: You can use `not` or `null`
- diff-at-point.el#L175: You can use `not` or `null`
- diff-at-point.el#L185: You can use `not` or `null`
- diff-at-point.el#L196: You can use `not` or `null`

ideasman42:
Thanks for the hints, applied changes.
(note that I went over these steps but did some changes after :/ so missed the linter error).

riscy:
Thanks!  It feels like this one:

```
diff-at-point.el:305:1:Warning: the function `diff-goto-source' is not known
    to be defined.
```

...could be solved with a `(require 'diff-mode)`

Another small note I see is that `progn` isn't needed for the else part of an if, e.g.:
```elisp
              (if fallback-point-is-begin
                (progn
                  (goto-char fallback-point-begin)
                  (forward-line 1)
                  (forward-char))
                (progn
                  (goto-char (or fallback-point-end current-filename-diff-point-max))
                  (forward-line -1)
                  (forward-char)))
```
could be:
```elisp
(if fallback-point-is-begin
    (progn
      (goto-char fallback-point-begin)
      (forward-line 1)
      (forward-char))
    (goto-char (or fallback-point-end current-filename-diff-point-max))
    (forward-line -1)
    (forward-char))
```

ideasman42:
Thanks, updated.

alphapapa:
@ideasman42 Have you considered submitting this to Emacs itself?  It seems like a basic, useful feature that ought to exist by default.  Why keep it as a separate package?

ideasman42:
@alphapapa yes, I proposed this, I submitted a video with the functionality to see if they were interested to add this but didn't hear back.

https://debbugs.gnu.org/cgi/bugreport.cgi?bug=36526

riscy:
Looking below the surface now, I wonder if this functionality is a part of magit?  e.g. are you familiar with`magit-status-here`?

ideasman42:
> Looking below the surface now, I wonder if this functionality is a part of magit? e.g. are you familiar with`magit-status-here`?

No, I don't use maggit, however I installed it and compared the functionality.

There are some differences:

- Only works with git (I use this with subversion too).
- Doesn't jump to the nearest hunk (this is handy when working in large files).

riscy:
Thanks for checking!  Here's another quick nag:

```
diff-at-point.el:241:15:Warning: `message' called with 1 args to fill 0 format
    field(s)
```

but this is looking close to ready to me.

ideasman42:
> Thanks for checking! Here's another quick nag:
> 
> M-x byte-compile-file (using Emacs 26.3):
> 
> ```
> diff-at-point.el:241:15:Warning: `message' called with 1 args to fill 0 format
>     field(s)
> ```
> 
> but this is looking close to ready to me.

Corrected

riscy:
Ok, great.  These aren't blockers, but I'm just having one last check:
- `When SCROLL-RESET is t `  -> `When SCROLL-RESET is non-nil` (a few places)

Thanks for this!

PR #6534, opened 2019-11-10
Add recipe for grammarly.

jcs090218:

Grammarly API interface.

https://github.com/jcs-elpa/grammarly

The maintainer.

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/grammarly
+++ b/recipes/grammarly
@@ -1,0 +1,1 @@
+(grammarly :repo "jcs-elpa/grammarly" :fetcher github)


riscy:
Thanks for this, on first glance this is looking like a good addition to MELPA.

For ultra-clean byte compilation, you should `(require 'subr-x)`, resolving this small nag:
```
grammarly.el:233:1:Warning: the following functions are not known to be
defined: string-trim, string-empty-p
```
- The defgroup on `grammarly` is a subgroup of `tool`, but I don't see any `tool` group in Emacs 26.3 -- there is a `tools` group, but it is for programming language tools.  Would `text` be a better choice?
- going -> going
- Kil -> kill

jcs090218:
> ultra-clean byte compilation

How do I get ultra-clean because when I compiled it. It doesn't show that I should `(require 'subr-x)`.

> * Kil -> kill

I didn't see where's this. Is this one line L200 => `Kill the timer.`?

---

Others are fixed! :D

riscy:
> How do I get ultra-clean because when I compiled it. It doesn't show that I should (require 'subr-x).

One way you can see these is enabling flycheck while editing your elisp files -- if you just byte-compile in your normal Emacs session (I guess I'd call that a "dirty compile"), you may have already imported things that your code relies on, but flycheck runs a clean compile.

The `Kil` is on line 194 ("Kil the websocket.")

I'm also noticing your URL is out of date -- https://github.com/jcs090218/grammarly should be https://github.com/jcs-elpa/grammarly now (the redirect is working but we should point to jcs-elpa where the package is hosted).

The byte-compile looks good now.  Just ping me when the URL is fixed and it's a fast merge.

Also some sharp-quote stuff while I'm here:
- grammarly.el#L228: You can sharp-quote this one
- grammarly.el#L229: No need to sharp-quote lambdas, but also no harm

jcs090218:
> One way you can see these is enabling flycheck while editing your elisp files -- if you just byte-compile in your normal Emacs session (I guess I'd call that a "dirty compile"), you may have already imported things that your code relies on, but flycheck runs a clean compile.

I did enable flycheck but it gives me error on line `36` saying `Cannot open load file: No suck file or directory s`. I don't know why this is happening. :/ 

---

All should be fixed including the updated URL.

@riscy Ping! :)

riscy:
This one's my fault for using ambiguous language, but on grammarly.el#L229 you're now regular-quoting instead of sharp-quoting the lambda.  In fact you don't need to quote it at all.  Regular-quoting is worse, because it prevents the compiler from compiling the lambda.

That's not a blocker, just something to look at.  Looks good to go!

Thanks. :)

PR #6536, opened 2019-11-10
Add recipe for humanoid-themes package

tasmo:

A color theme made for coding and working with a dark and a light version.
It works well with true color as well as in a 256 color terminal.

https://github.com/humanoid-colors/emacs-humanoid-themes
https://gitlab.com/tasmo/emacs-humanoid-themes

I'm the author and the maintainer of the package.

**None needed**

### Checklist

### DIFF
--- a/recipes/humanoid-theme
+++ b/recipes/humanoid-theme
@@ -1,0 +1,2 @@
+(humanoid-theme :repo "tasmo/humanoid-theme"
+                :fetcher github)
--- a/recipes/humanoid-theme
+++ b/recipes/humanoid-theme
@@ -1,2 +1,0 @@
-(humanoid-theme :repo "humanoid-colors/emacs-humanoid-theme"
-                :fetcher github)
--- a/recipes/humanoid-theme
+++ b/recipes/humanoid-theme
@@ -1,2 +1,2 @@
-(humanoid-theme :repo "tasmo/humanoid-theme"
+(humanoid-theme :repo "humanoid-colors/emacs-humanoid-theme"
                 :fetcher github)
--- a/recipes/humanoid-themes
+++ b/recipes/humanoid-themes
@@ -1,2 +1,2 @@
-(humanoid-theme :repo "humanoid-colors/emacs-humanoid-themes"
-                :fetcher github)
+(humanoid-themes :repo "humanoid-colors/emacs-humanoid-themes"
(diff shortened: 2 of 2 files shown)


riscy:
Hi, these look nice.  Some quick thoughts on this:

- Including humanoid-theme-pkg.el is discouraged -- MELPA will create a `-pkg.el` file
- You `provide humanoid-common` in your main file, but most (all?) of your definitions don't use that as a prefix -- they use `humanoid-theme` instead.   Best to choose one.

Some other relevant nags:
```
94:0: error: "true-color-p" doesn't start with package's prefix "humanoid-common".
100:0: error: "create-humanoid-theme" doesn't start with package's prefix "humanoid-common".
```

tasmo:
Hi @riscy,
thank you for these very useful hints.
I tried to implement all the mentioned issues.

Afterwards I did `M-x package-lint-current-buffer` again and got just the Emacs version error comments on removed and added features which this theme colorizes as well.

riscy:
Thanks for the patience on this!

These are fyi, but these aren't blockers:
(contains `no-byte-compile: t` but checking anyway)
```
humanoid.el:99:1:Warning: Unused lexical variable `head5-bg'
humanoid.el:99:1:Warning: Unused lexical variable `act1'
```

The last thing I notice is that your package is called `humanoid-theme`, but it strikes me that it should be `humanoid-themes` in the plural i.e. [here](
https://github.com/melpa/melpa/pull/6536/files#diff-45233a9eec99b717bdba6213a88d6da5R1).  (I realize I said "humanoid-theme" up above, but the file/package is not itself a single theme.  Apologies for that one.  I'm always a bit rusty when it comes to color themes.)

Similarly, your main file is `humanoid.el`, but now that we're without a `-pkg.el` file we want its name to line up with the name of the package (i.e. `humanoid-themes.el`).

With those names changed we're good to go, so please ping me when ready!

tasmo:
@riscy Just want to ask kindly, if you got the upstream info.

riscy:
Thanks so much for your patience.  I didn't catch the upstream info. :)

Humanoid looks ready to go 🤖 --I'll mention this last thing as a non-blocker so we can get this merged: you're still defining `head5-bg` but not using it.  Maybe you're keeping that for future.

And, welcome to MELPA!

PR #6537, opened 2019-11-10
erc-hipchatify: remove since hipchat is long since dead

seanfarley:
Hipchat was shutdown a long time ago.


purcell:
Great, thanks.

PR #6538, opened 2019-11-11
Move some repositories from bitbucket to gitlab

gvol:
This was necessitated due to bitbucket's discontinuing of mercurial
support.

I admit to not checking everything since they are pre-existing packages, but I can if it's really necessary.  @tarsius, what do you think?  I did build the packages with MELPA's make to ensure that works.


tarsius:
Thanks!

> what do you think? 

Well... you don't have to. But it never hurts. For example I recently spell-checked all the packages I have installed... oh boy. :wink:

PR #6540, opened 2019-11-11
Add recipe for flycheck-grammarly.

jcs090218:

Grammarly support for Flycheck.

https://github.com/jcs-elpa/flycheck-grammarly

The maintainer.

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/flycheck-grammarly
+++ b/recipes/flycheck-grammarly
@@ -1,0 +1,1 @@
+(flycheck-grammarly :repo "jcs-elpa/flycheck-grammarly" :fetcher github)


riscy:
Just taking a quick look at this -- it appears `grammarly` is not installable.  Do you mean [this](https://github.com/mmagnus/emacs-grammarly)?

---------

- flycheck-grammarly.el#L174: It is beneficial to sharp-quote hooks; use `#'`

jcs090218:
> Just taking a quick look at this -- it appears grammarly is not installable. Do you mean this?

No, I mean [this](https://github.com/melpa/melpa/pull/6534). :D

> * flycheck-grammarly.el#L174: It is beneficial to sharp-quote hooks; use #'

Fixed!

riscy:
My bad.  Thanks. :)

riscy:
### flycheck-grammarly.el
```
flycheck-grammarly.el:206:1:Warning: the function `jcs-is-end-of-buffer-p' is
not known to be defined.
```

- No issues! 💯

- No issues! 💯

Also:
- `flycheck-grammarly--check-all`'s docstring seems to refer to parameter BUF but there is no such parameter
- It seems you `(require 's)` to only use one function -- `s-replace` -- but you could reduce your requirements by using `replace-regexp-in-string`.  Minor, but sometimes it's worth it to minimize dependencies
 - `"On closse Grammarly API."` -- looks like a typo around closse?
- `(add-to-list 'grammarly-on-open-function-list 'flycheck-grammarly--on-open)`: I might be missing something, but this seems strange to me.  The function `flycheck-grammarly--on-open` only writes a message about trying to connect.  If we had N packages that were all using the grammarly API the way flycheck-grammarly is, then we'd get N such messages echoed to us, right?  That seems a little odd.
  - Similarly for flycheck-grammarly--on-message -- the `(message "[INFO] Receiving data from grammarly,...` seems superfluous, but correct me if I'm wrong
- Also, `grammarly-on-open-function-list` is user-customizable, so we have code here that is changing a user-customizable variable whenever the package is loaded.  This is more an issue with the `grammarly` package than `flycheck-grammarly`, but perhaps this variable shouldn't be a `defcustom` if the intend is to change its value progammatically
-

jcs090218:
> * (add-to-list 'grammarly-on-open-function-list 'flycheck-grammarly--on-open): I might be missing something, but this seems strange to me. The function flycheck-grammarly--on-open only writes a message about trying to connect. If we had N packages that were all using the grammarly API the way flycheck-grammarly is, then we'd get N such messages echoed to us, right? That seems a little odd.
Similarly for flycheck-grammarly--on-message -- the (message "[INFO] Receiving data from grammarly,... seems superfluous, but correct me if I'm wrong

Oh, yes... you are right. My previous design is to just have a function receiving it then I changed to a list instead of just a function. I thought about this and assume people wouldn't use `grammarly` while this package is on. Anyway, I added the check before messaging. Any suggestion is welcome!

> * Also, grammarly-on-open-function-list is user-customizable, so we have code here that is changing a user-customizable variable whenever the package is loaded. This is more an issue with the grammarly package than flycheck-grammarly, but perhaps this variable shouldn't be a defcustom if the intend is to change its value progammatically

I thought `defcustom` is also for other package too. I mean `flycheck-grammarly` depends on `grammarly` and they are two separate packages. Is there something wrong to have `flycheck-grammarly` uses `defcustom` from `grammarly`? 😕 

---

Any other issue should be fixed! :D

riscy:
> I thought defcustom is also for other package too. I mean flycheck-grammarly depends on grammarly and they are two separate packages. Is there something wrong to have flycheck-grammarly uses defcustom from grammarly?

defcustom is for what are also called "user" options -- using it enables a user to [use Emacs's customize feature to set their values][https://www.gnu.org/software/emacs/manual/html_node/eintr/defcustom.html].  (Of course, they don't have to.  They can use setq or any other interface: "So long as you remember where the values are set, you will have no trouble.").

I think the solution is easy - `grammarly-on-open-function-list` and its siblings are not user options, and so shouldn't be declared with defcustom (sorry I didn't realize this when I was reviewing `grammarly` in isolation).

I'm not too familiar with `flycheck-define-generic-checker` but it looks like `text-mode` is a parent of `latex-mode`, `markdown-mode`, and `org-mode` (and `outline-mode`, `nroff-mode`, etc.) -- is it necessary to list all of `(text-mode latex-mode org-mode markdown-mode))` or can you just list `text-mode`?

If I'm understanding correctly, this package will call `(message "[INFO] Reset!")` every time the user modifies the buffer?  Wouldn't this be disruptive?  Also, this message has no context about where it's coming from -- why should a user keep seeing the message "Reset!" over and over?

That's all for now -- hope that helps.

jcs090218:
> defcustom is for what are also called "user" options -- using it enables a user to [use Emacs's customize feature to set their values][https://www.gnu.org/software/emacs/manual/html_node/eintr/defcustom.html]. (Of course, they don't have to. They can use setq or any other interface: "So long as you remember where the values are set, you will have no trouble.").

I think the solution is easy - grammarly-on-open-function-list and its siblings are not user options, and so shouldn't be declared with defcustom (sorry I didn't realize this when I was reviewing grammarly in isolation).

Fixed!

> I'm not too familiar with flycheck-define-generic-checker but it looks like text-mode is a parent of latex-mode, markdown-mode, and org-mode (and outline-mode, nroff-mode, etc.) -- is it necessary to list all of (text-mode latex-mode org-mode markdown-mode)) or can you just list text-mode?

Oh, okay. This is my first `flycheck` package so I don't know about this! Fixed!

> If I'm understanding correctly, this package will call (message "[INFO] Reset!") every time the user modifies the buffer? Wouldn't this be disruptive? Also, this message has no context about where it's coming from -- why should a user keep seeing the message "Reset!" over and over?

Okay, I added the option `flycheck-grammarly--show-debug-message` and default as `nil`. When first design this, I wouldn't want the user just to hang there and waiting for the requests. But I think you are right; after all, the message is quite annoying. Thanks!

riscy:
Cool, thanks and apologies for all the passes this is taking.  I think this is getting there, but there could be more care taken to make this a top notch package.  I'm looking at `flycheck-grammarly--after-change-functions`:
```elisp
  (unless (string=
           (flycheck-grammarly--limit-changes-buffer flycheck-grammarly--last-buffer-string)
           (flycheck-grammarly--limit-changes-buffer (buffer-string)))
   ;;; <-- kill a timer and restart it, etc. -->
```
And this guard is being called as part of our `after-change-functions`, which means that a user typing at (say) 100WPM is going to call `flycheck-grammarly--limit-changes-buffer` at a very rapid pace (the first call doesn't seem like it should be necessary, but I digress).  I look at that function and I see this:
```
(defun flycheck-grammarly--limit-changes-buffer (str)
  "Minify the STR to check if any text changed."
  (with-temp-buffer
    (insert str)
    (delete-whitespace-rectangle (point-min) (point-max))
    (call-interactively #'mark-whole-buffer)
    (goto-char (point-min))
    (while (search-forward "\n" nil t) (replace-match "" nil t))
    (buffer-string)))
```
...and I haven't benchmarked this, but this looks inefficient (both in the creation of temporary buffers and the creation/comparison of potentially long strings).  Moreover, per the docs, use of mark-whole-buffer is recommended against in elisp code (`(call-interactively #'mark-whole-buffer)` feels like a hack), which hints to me that something could possibly be improved.  Unless I'm missing something, why couldn't something like this be used?
```elisp
(defun flycheck-grammarly--limit-changes-buffer (str)
  (declare (side-effect-free t))
  (md5 (replace-regexp-in-string "[[:space:]\n]+" " " str)))
```
That way you still achieve invariance when reformatting due to spaces, and you don't create a temporary buffer and move a mark around inside it at a very rapid pace. The use of `md5` means that you then aren't comparing two (potentially very long) strings, you're just comparing a short hash, although I don't know if this is ultimately faster.  And I would personally rename this function to something like `flycheck-grammarly--minified-string` or similar.

All of this is being done in `after-change-functions` so that we can freshen a timer.  But this feels like an approximation of what `run-with-idle-timer` does.  Does it seem feasible to start a repeating idle-timer in the background that looks for these meaningful changes in the buffer (say, after 500ms of inactivity) rather than doing it on every change?

jcs090218:
> That way you still achieve invariance when reformatting due to spaces, and you don't create a temporary buffer and move a mark around inside it at a very rapid pace. The use of md5 means that you then aren't comparing two (potentially very long) strings, you're just comparing a short hash, although I don't know if this is ultimately faster. And I would personally rename this function to something like flycheck-grammarly--minified-string or similar.

OMG, You are a genius! I do have issue with this performance and haven't thought improving this. Thank you so much! :flushed:

> All of this is being done in after-change-functions so that we can freshen a timer. But this feels like an approximation of what run-with-idle-timer does. Does it seem feasible to start a repeating idle-timer in the background that looks for these meaningful changes in the buffer (say, after 500ms of inactivity) rather than doing it on every change?

I have implemented it with `run-with-idle-timer` but the result are difference and not quite stable. I might pass the changes here as long the plugin work well right now.  😖 

Anyway, this is the implementation using `run-with-idle-timer` [click here](https://github.com/jcs-elpa/flycheck-grammarly/blob/08a46238f0e703e1f73e8c60196ed38d5b98556f/flycheck-grammarly.el).

---

One thing, I have revert the [03131b3d77a475ff6a2e1fb7b80c750960a2f8a8](https://github.com/jcs-elpa/flycheck-grammarly/commit/03131b3d77a475ff6a2e1fb7b80c750960a2f8a8) because it wasn't working in `markdown-mode` so I assume other modes would not work too. Let me know if I should still get rid of this! And thanks again! 😄

riscy:
> I have revert the 03131b3d77a475ff6a2e1fb7b80c750960a2f8a8 because it wasn't working in markdown-mode

That's my oversight, I wasn't sure whether you could only specify a parent mode or not.  I guess I know now. :)

Looking over it, I think this is ready to go.  Thanks!

PR #6543, opened 2019-11-13
added nv-delete-forward recipe

nivaca:

The package `nv-delete-forward` replicates the forward delete behavior of modern text editors like oXygen XML or Sublime Text. It implements the function `nv-delete-forward-all` which deletes all white space (including newlines) until, but not including, the following text. 

https://gitlab.com/nivaca/nv-delete-forward

I am creator and maintainer.

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/nv-delete-forward
+++ b/recipes/nv-delete-forward
@@ -1,7 +1,1 @@
-<<<<<<< HEAD
-(nv-delete-forward :fetcher github
-  :repo "nivaca/nv-delete-forward"
-  :files ("*.el" ("doc/*.md" "doc/*.gif")))
-=======
 (nv-delete-forward :fetcher gitlab :repo "nivaca/nv-delete-forward")
->>>>>>> upstream/master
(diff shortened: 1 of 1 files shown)


purcell:
Check out the built-in `c-hungry-delete-forward` command.

riscy:
Have you had a chance to look at `c-hungry-delete-forward`?  It's short enough I can include it here:

```elisp

(defun c-hungry-delete-forward ()
  "Delete the following character or all following whitespace
up to the next non-whitespace character.
See also \\[c-hungry-delete-backwards]."
  (interactive)
  (let ((here (point)))
    (c-skip-ws-forward)
    (if (/= (point) here)
	(delete-region (point) here)
      (funcall c-delete-function 1))))
```

A couple small static checks I can also offer while here, although these might be unimportant depending on your answer to the above:
- nv-delete-forward.el#L7: Use `Version` instead of `Package-Version` (MELPA automatically adds `Package-Version`)
- nv-delete-forward.el#L24: Prefer `https` over `http` (if possible)

tarsius:
Closing due to inactivity.

PR #6544, opened 2019-11-14
recipes/helm-fish-completion: Add helm-fish-completion

Ambrevar:

Helm interface for fish-completion.

https://github.com/emacs-helm/helm-fish-completion

Maintainer.

 **None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/helm-fish-completion
+++ b/recipes/helm-fish-completion
@@ -1,0 +1,3 @@
+(helm-fish-completion
+ :repo "emacs-helm/helm-fish-completion"
+ :fetcher github)


riscy:
Should your `package-requires` include `fish-completion`?  It also looks like you need `(require 'esh-mode)`

Ambrevar:
Fixed, thanks!

riscy:
Great, and thanks for your patience.  There's a small package-lint nag:

```
1 issue found:

```
and my usual https nag :) 
- helm-fish-completion.el#L27: Prefer `https` over `http` (if possible)

Beyond that, I see an autoload cookie on `helm-fish-completion-make-eshell-source`, but maybe it's not necessary?  See [when to autoload](https://www.gnu.org/software/emacs/manual/html_node/elisp/When-to-Autoload.html).

Beyond that, the code is looking good -- ping me for another check.

Ambrevar:
Thanks for the review, I've fixed the 2 issues.

Regarding the autoload: The idea is that the user only needs to add

(setq helm-esh-pcomplete-build-source-fn #'helm-fish-completion-make-eshell-source)

to their Emacs configuration to enable Helm Fish completion.
Alternatively, I could remove the autoload and let the user call
`(require 'helm-fish-completion)` themselves.  Thoughts?

riscy:
Looks good!  Thanks for this.

PR #6545, opened 2019-11-15
Add recipe for ebuku.

flexibeast:

EBuku provides a basic interface to the Buku Web bookmark manager.

https://github.com/flexibeast/ebuku

Creator.

n/a

### Checklist

Please confirm with `x`:

- [x] My elisp byte-compiles cleanly (originally with one exception, described below; now addressed)

Byte-compiling produces:

    ebuku.el:242:1:Warning: Unused lexical argument ‘proc’

which is indeed correct - the `proc` argument is not needed by that function, but Emacs calls sentinels with `proc event` as arguments.

### DIFF
--- a/recipes/ebuku
+++ b/recipes/ebuku
@@ -1,0 +1,3 @@
+(ebuku
+ :fetcher github
+ :repo "flexibeast/ebuku")


riscy:
> the proc argument is not needed by that function

You can give the byte compiler a hint by renaming it to `_proc`.

flexibeast:
@riscy: Thanks for the tip! Renamed in https://github.com/flexibeast/ebuku/commit/67b5307ad93627a9171a93854e824f8c923ccddd.

flexibeast:
(Now that i've addressed the "byte-compiles cleanly" item in the checklist, do i update the checklist to reflect that, or is that for a reviewer to do once they've confirmed it?)

riscy:
Oh yeah, you can check it off. :)  But while I'm here...

```
Warning (emacs): 
ebuku.el:0: The first line should be of the form: ";;; package --- Summary"
```

<kbd>M-x package-lint-current-buffer</kbd> (using version 20191103.1154):
```
No issues found.
```

No byte-compilation issues! 😎

Extra hints:
- ebuku.el#L184: This define-key could overwrite a user's keybindings.  Try: `(defvar my-map (let ((km (make-sparse-keymap))) (define-key ...) km))`

flexibeast:
Thanks for the further feedback. :-)

> Oh yeah, you can check it off. :)

Okay, i'll do that shortly.

> ebuku.el:0: The first line should be of the form: ";;; package ---

Yeah, i saw that, but wasn't sure how that should be combined with enabling lexical scoping? Do i just move `-*- lexical-binding: t; -*-` to the end of the `;;; package` line?

> This define-key could overwrite a user's keybindings

Oops, indeed! i'll fix that.

riscy:
The lexical binding line has to go on the first line of the file -- also, I wonder if mentioning Emacs in it is redundant?  Maybe:

```elisp
;;; ebuku.el --- Interface to the Buku Web bookmark manager -*- lexical-binding: t; -*-
```

This block in your `define-derived-mode` is actually redundant (define-derived-mode handles it all):
```elisp
  (kill-all-local-variables)
  (setq-local major-mode 'ebuku-mode)
  (setq-local mode-name "EBuku")
  (use-local-map ebuku-mode-map)
  (run-mode-hooks 'ebuku-mode-hook))
```
...so you could actually just remove those.

Minor, but the `if` on line 351 could be a `when` (which means you don't need to use `progn`).  Even more minor, I see a couple places where `progn` isn't necessary and the code could be made more terse, for example an expression like this --
```elisp
(if bookmark
            (progn
              (let ((title (cdr (assoc 'title bookmark)))
                    (url (cdr (assoc 'url bookmark)))
                    (comment (cdr (assoc 'comment bookmark)))
                    (tags (cdr (assoc 'tags bookmark))))
                (setq title (read-from-minibuffer "Title? " title))
                (setq url (read-from-minibuffer "URL? " url))
                (setq comment (read-from-minibuffer "Comment? " comment))
                (setq tags (read-from-minibuffer "Tags? " tags))
                (with-temp-buffer
                  (if (ebuku--call-buku `("--update" ,index
                                          "--title" ,title
                                          "--url" ,url
                                          "--comment" ,comment
                                          "--tag" ,tags))
                      (progn
                        (ebuku-refresh)
                        (message "Bookmark updated."))
                    (error "Failed to update bookmark")))))
          (error (concat "Failed to get bookmark data for index" index)))
```
can possibly be written like this instead (but check my work):
```elisp
(if bookmark
    (let ((title (read-from-minibuffer   "Title? "   (cdr (assoc 'title bookmark))))
          (url (read-from-minibuffer     "URL? "     (cdr (assoc 'url bookmark))))
          (comment (read-from-minibuffer "Comment? " (cdr (assoc 'comment bookmark))))
          (tags (read-from-minibuffer    "Tags? "    (cdr (assoc 'tags bookmark)))))
      (with-temp-buffer
        (if (not (ebuku--call-buku `("--update" ,index
                                     "--title" ,title
                                     "--url" ,url
                                     "--comment" ,comment
                                     "--tag" ,tags)))
            (error "Failed to update bookmark")
          (ebuku-refresh)
          (message "Bookmark updated."))))
  (error (concat "Failed to get bookmark data for index " index)))
```
which for me makes it easier to spot "wrong code" (there was none in this case, except maybe the missing space at after `data for index`).

Beyond those nags, I think this is looking good.  Ping me for another check when you're ready.

flexibeast:
@riscy:

Thank you again for the feedback!

i've now made the requested changes, with two exceptions:

* "the if on line 351 could be a when (which means you don't need to use progn)." i've not changed this, as there in changes in train which result in the `progn` becoming necessary.

* i've not exactly followed the layout changes you suggest for `ebuku-edit-bookmark`, as doing so pushed line lengths over 80 characters; i've been told elsewhere to try to avoid this. Hopefully the layout i've chosen, which works for me, is okay for you also. :-)

riscy:
This looks good to me.  Thanks for your patience!

Super minor and not blockers, but while I'm here:
- asychrononous -> asynchronous
- ebuku.el#L27: Prefer `https` over `http` (if possible)
- ebuku.el#L106: Prefer `https` over `http` (if possible)

flexibeast:
Thanks for reviewing and merging this!

i've just changed the gnu.org URLs from `http` to `https`, but i'm not sure where 'asychrononous' is mentioned .... ?

PR #6547, opened 2019-11-16
Delete recipes/jsx-mode

abicky:

jsx-mode.el had been created for JSX, an AltJS provided by DeNA, before JSX, an XML-like syntax extension, provided by Facebook was released, and I added its recipe by https://github.com/melpa/melpa/pull/1013.
However many people misunderstand that jsx-mode.el is for the XML-like syntax.
cf. https://github.com/jsx/jsx-mode.el/issues/13

In addition, the AltJS seems to be no longer maintained, so I think it is better to delete its recipe to avoid misunderstanding.

https://github.com/jsx/jsx-mode.el

The maintainer

None needed

### Checklist

Please confirm with `x`:

    - I think I don't need to check these items to delete an existing package.

### DIFF
--- a/recipes/jsx-mode
+++ b/recipes/jsx-mode
@@ -1,3 +1,0 @@
-(jsx-mode :repo "jsx/jsx-mode.el"
-          :fetcher github
-          :files ("src/*.el"))


riscy:
Thanks!  Have you considered renaming the package to, e.g., `jsxlang` or similar?

abicky:
I thought of changing the description in README as described in https://github.com/jsx/jsx-mode.el/issues/13, but I think it is better to delete it than changing the description or changing the name. It is hard to avoid misunderstanding because almost all software engineers writing JavaScript codes associate the word "JSX" with a XML-like syntax. 

In addition, as you can see, the last commit of [JSX](https://github.com/jsx/JSX/commits/master) was made about 4 years ago. The situation surrounding JavaScript has changed drastically for the past 4 years, so it is outdated and few want to install jsx-mode.el from MELPA.

riscy:
Okay, it can definitely be unindexed.  There may be some value in preserving it, somehow, under whatever name, but I will leave that up to you.  Thank you for keeping an eye on this.

PR #6548, opened 2019-11-16
Do not apt-get update: too costly.

equwal:
You should not apt-get update from TravisCI.
https://docs.travis-ci.com/user/installing-dependencies#installing-packages-on-standard-infrastructure

So that line is removed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,6 @@ addons:
     - texinfo
 
 before_install:
-  - sudo apt-get -yq update
   - sudo apt-get -yq install git mercurial texinfo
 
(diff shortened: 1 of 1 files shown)


Silex:
That doesn't make sense, the page you link explicitely tells that you can/should run update.

It says that you shouldn't run upgrades tho:

```
Do not run apt-get upgrade in your build as it downloads up to 500MB of packages and significantly extends your build time. Additionally, some packages may fail to update, which will lead to a failed build.
```

Can you maybe quote the line where it advices against update?

equwal:
Correct.

PR #6559, opened 2019-11-25
add recipe for anki-vocabulary

lujun9972:

anki-vocabulary is a plugin that helps you to create vocabulary cards in Anki.

It only supports translate English to Chinese right now.

https://github.com/lujun9972/anki-vocabulary.el

I am the maintainer

 **None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/anki-vocabulary
+++ b/recipes/anki-vocabulary
@@ -1,0 +1,1 @@
+(anki-vocabulary :fetcher github :repo "lujun9972/anki-vocabulary.el")


riscy:
To `(require 'pdf-view)` you need a `Package-Requires` on `pdf-tools`.

alphapapa:
There does not appear to be any English documentation for how to use the package or what it actually does ("Help you to create vocabulary card in Anki" does not explain what it actually does or how).

As well, if it only supports Chinese-English vocabulary, the name ought to reflect that.  The name `anki-vocabulary` suggests that it's generally useful for studying vocabulary with Anki, which does not appear to be the case.

lujun9972:
@riscy Thanks, since `anki-vocabulary` is workable without pdf-view, so I remove the require and define the `anki-vocabulary--get-pdf-text` only when the `pdf-view` is present (using featurep).

@alphapapa I have added an English README, and there are some steps to telling how to specify the original language and target language. (English is not my native language. If there are grammar mistakes or inappropriateness,  please let me know, thank you.)

riscy:
Great, that works.  The byte-compiler can step all the way through now --

### anki-vocabulary.el

```
Warning (emacs): 
anki-vocabulary.el:208: First sentence should end with punctuation
```

```
No issues found.
```

```
anki-vocabulary.el:264:1:Warning: the following functions are not known to be
    defined: pushnew, anki-vocabulary--get-pdf-text
```

```
anki-vocabulary.el:45:Warning (check-declare): said
    `pdf-view-active-region-text' was defined in pdf-view.el: file not found
anki-vocabulary.el:46:Warning (check-declare): said
    `pdf-view-assert-active-region' was defined in pdf-view.el: file not found
```
You can mark these as external with:
```elisp
(declare-function pdf-view-active-region-text "ext:pdf-view" ())
(declare-function pdf-view-assert-active-region "ext:pdf-view" ())
```

Other static checks:
- anki-vocabulary.el#L192: It's safer to sharp-quote function names; use `#'`
- anki-vocabulary.el#L10: Prefer `https` over `http` (if possible)
- anki-vocabulary.el#L25: Prefer `https` over `http` (if possible)
- anki-vocabulary.el#L30: Prefer `https` over `http` (if possible)

lujun9972:
@riscy It is strange that I marked pdf-view-assert-active-region as  external with:
```
(declare-function pdf-view-assert-active-region "ext:pdf-view" ())
```

But I still get the warning
```
anki-vocabulary.el:46:Warning (check-declare): said
    ‘pdf-view-assert-active-region’ was defined in
    .emacs.d/elpa/pdf-tools-20191128.1731/pdf-view.el: function not found
```

riscy:
Ah, I installed `pdf-tools` and I see that the symbol in question is a macro, not a function.

You can really force `declare-function` to relax if you want by just asking it to check the file existence:
```elisp
(declare-function pdf-view-assert-active-region "ext:pdf-view" () t)
```
Per the docs:

> This is intended for function definitions that check-declare does not recognize, e.g., defstruct.

lujun9972:
@riscy Thanks, and I have one more problem.
I want to redefine `anki-vocabulary--get-pdf-text` and `anki-vocabulary--get-text` after loading `pdf-view`.
```
(with-eval-after-load 'pdf-view
  (defun anki-vocabulary--get-pdf-text ()
    "Get the text in pdf mode."
    (pdf-view-assert-active-region)
    (let* ((txt (pdf-view-active-region-text))
           (txt (string-join txt "\n")))
      (replace-regexp-in-string "[\r\n]+" " " txt)))
  (defun anki-vocabulary--get-text ()
    "Get the region text."
    (if (derived-mode-p 'pdf-view-mode)
        (anki-vocabulary--get-pdf-text)
      (anki-vocabulary--get-normal-text))))
```
But I got an warning when `M-x package-lint-current-buffer`
```
1 issue found:

128:1: warning: `with-eval-after-load' is for use in configurations, and should rarely be used in packages.
```
I have tried `eval-after-load` and still got a similar warning. 

Is there any way to fix the problem?

riscy:
Would it be simpler to just use `featurep`?

```elisp

(defun anki-vocabulary--get-text ()
  "Get the region text."
  (if (and (featurep 'pdf-view) (derived-mode-p 'pdf-view-mode))
        (anki-vocabulary--get-pdf-text)
    (anki-vocabulary--get-normal-text)))

(defun anki-vocabulary--get-pdf-text ()
    "Get the text in pdf mode."
    (pdf-view-assert-active-region)
    (let* ((txt (pdf-view-active-region-text))
           (txt (string-join txt "\n")))
      (replace-regexp-in-string "[\r\n]+" " " txt)))
```

lujun9972:
@riscy Thank you for your advice.
I check whether `pdf-tools` installed in `anki-vocabulary--get-pdf-text`. It will throw an error if `pdf-tools` not installed, otherwise do the `(require 'pdf-view)`. I hope it OK. 
Please check the code again. Thank you very much.

riscy:
That solution is fine, but I wonder if `(derived-mode-p 'pdf-view-mode)` is true, will `(package-installed-p 'pdf-tools)` automatically be true too?  (So it's impossible for the error to ever get thrown.)

The above question is a minor issue, but I'll wait to hear back from you before hitting merge.  In the meantime, here are a few more minor issues (also probably non-blockers, but worth checking):

```
Warning (emacs): 
anki-vocabulary.el:128: Messages should start with a capital letter
```
Maybe you could write the line as:
```
(error "`pdf-tools' is required!"))
```

Minor:
- defauld -> default
- meanning -> meaning
- sentece -> sentence
- In a couple places: `"Translate `SENTENCE' using youdao."`: you only need to capitalize SENTENCE, not ``` `quote' ``` it -- the quoting is typically reserved to indicate actual elisp symbols and will create links in the online help

lujun9972:
@riscy Yes, you are right. There is no need to check whether package-tools has installed, so I commented out the code.

I think I have fixed the issues above. Please check the code again. Thank you very much.

riscy:
Great thanks!  I'm looking at some existing MELPA packages and just wanted to double-check, is this package distinct from https://melpa.org/#/anki-editor ?  I haven't looked too deeply, but I was wondering if they occupied the same space.

lujun9972:
@riscy
They are quite different. anki-editor is more flexible, It is actually convert the org into an Anki card， so you have to input the content manually.
While anki-vocabulary is more rigid but more convenient. It is used to create vocabulary cards only(the model fields are fixed). It will do the translation and make the Anki card automatically.

PR #6560, opened 2019-11-26
Add recipe for ffmpeg-player.

jcs090218:

Play video using ffmpeg.

https://github.com/jcs-elpa/ffmpeg-player

The maintainer.

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/ffmpeg-player
+++ b/recipes/ffmpeg-player
@@ -1,0 +1,1 @@
+(ffmpeg-player :repo "jcs-elpa/ffmpeg-player" :fetcher github)


alphapapa:
Do I understand correctly that this package works by extracting every frame of a video into a separate image file on-disk and repeatedly displaying those images in an Emacs buffer?

jcs090218:
> Do I understand correctly that this package works by extracting every frame of a video into a separate image file on-disk and repeatedly displaying those images in an Emacs buffer?

Not sure what you meant by `repeatedly displaying those images`. This package extract frame from a video using `ffmpeg`, and play those frame base on the calculated time. Calculated time is base on FPS and delta time, hence the outcome should be accurate. 

Here is some circumstances that the frame images will be deleted:

* Frame images are completely deleted when you called `ffmpeg-player-clean`. 
* The previous video's frame images are deleted when you played a new video (Try to not to waste the disk space). 
* When you replay the video then frame image get regenerated. (Just to prevent something goes wrong).

Hope this solve your confusion! :D

alphapapa:
I'm not confused.  What I said is how the package works.  It extracts all frames of a video to disk as images, and it repeatedly displays images (not the same image repeatedly) in a buffer to imitate a video player.

That's an interesting idea, but I can't imagine that it's practical to do so in Emacs.  For example, consider an hour-long, 1080p video.  For the sake of argument, let's say it's encoded with H.264 and takes up 1 GB of space (that's probably significantly less space than it would really consume at average bitrates).  To extract all of its frames as PNGs would probably consume at least 10 GB of space, depending on PNG compression (and more compression requires more CPU).

Now, how fast can Emacs read and display images?  Can it keep up with a 30 fps video at 1080p?  60 fps?

How fast can `ffmpeg` dump frames to disk?  30 fps?  60 fps?  How hard is that going to churn I/O while it's doing so?

It's a neat experiment, but I can't see it being a practical or recommended way to "watch videos in Emacs."  It already depends on `ffmpeg` and `ffplay`--so why would a user who has those tools want to use this package instead, consuming excessive disk space, I/O, and CPU?

Packages on MELPA ought to be ones that can be generally recommended as good to use.  As creative as this package is, I can't see it being something that can be recommended for actual use.

jcs090218:
> That's an interesting idea, but I can't imagine that it's practical to do so in Emacs. For example, consider an hour-long, 1080p video. For the sake of argument, let's say it's encoded with H.264 and takes up 1 GB of space (that's probably significantly less space than it would really consume at average bitrates). To extract all of its frames as PNGs would probably consume at least 10 GB of space, depending on PNG compression (and more compression requires more CPU).

You are correct for all this. But this is the best solution I can think of. And here are some performance improvements

* This encode to JPG, not PNG. (You could, but not recommended)
* By changing `ffmpeg-player-display-width` and `ffmpeg-player-display-height` can limit the file size. The smaller the size is the smaller the disk space it takes.

> Now, how fast can Emacs read and display images? Can it keep up with a 30 fps video at 1080p? 60 fps?

30 fps is fine for my PC. I don't know if there are 60 fps video. At least I never see one. The output size is depends on the user by changing `ffmpeg-player-display-width` and `ffmpeg-player-display-height`.

> How fast can `ffmpeg` dump frames to disk? 30 fps? 60 fps? How hard is that going to churn I/O while it's doing so?

In my use case is faster than the actual playing video (It depends on video's FPS and your PC).

> It's a neat experiment, but I can't see it being a practical or recommended way to "watch videos in Emacs." It already depends on ffmpeg and ffplay--so why would a user who has those tools want to use this package instead, consuming excessive disk space, I/O, and CPU?

When you want to watch video when you are coding? The below video image is 672 x 378.

watch-video

> Packages on MELPA ought to be ones that can be generally recommended as good to use. As creative as this package is, I can't see it being something that can be recommended for actual use.

I can play video on my computer. Try it yourself!

---

Overall, This is the best solution I could think of unless you have any other solution to play video file in Emacs. Thanks!

riscy:
This is an interesting demo, cool. :)  Apologies for the extremely long delay -- there are quite a few PR's open and I've fallen behind.

I think you want to use `shell-quote-argument` in `ffmpeg-player--form-command-video` since (weirdly enough) it is possible for filenames to have `"` in them, which could throw a wrench into interpolating `"ffmpeg -i \"%s\" %s \"%s%s%s.%s\""`. Similarly with `ffmpeg-player--command-play-audio`.  And probably also for `ffmpeg-player--async-delete-directory`, since the directory name may have a `"` in it.

Whenever I see a `setq` under a `let` I wonder if the pattern could be simplified.  What about changing:
```elisp
(let ((start-pt -1))
      (search-forward "Duration: ")
      (setq start-pt (1- (point)))
      (search-forward ",")
      (substring (buffer-string) start-pt (- (point) 2)))
```
to
```elisp
(search-forward "Duration: ")
(let ((start-pt (1- (point))) ; ...
```

And as a last thing I wonder, as a small suggestion, if these functions:
```elisp
(defun ffmpeg-player-volume-dec-5 ()
  "Decrease volume by 5."
  (interactive)
  (ffmpeg-player--move-volume -5))
```
could be rewritten to take an argument (but retain the default value):
```elisp
(defun ffmpeg-player-volume-decrease (amount)
  "Decrease volume by AMOUNT (default -5)."
  (interactive)
  (ffmpeg-player--move-volume (or amount -5)))
```
...just as a means to increase their usability, if it ever came to that.  That's really minor though.

Minor:
- ffmpeg-player.el#L444: It's safer to sharp-quote function names; use `#'`
- ffmpeg-player.el#L219: It's safer to sharp-quote function names; use `#'`
- ffmpeg-player.el#L321: It's safer to sharp-quote function names; use `#'`
- "no longer alived" -> "no longer live"
- inforamtion -> information
- "Timer that try to" -> "Timer that tries to"

jcs090218:
All issues have fixed! 😄 Thanks!

conao3:
Hi!

1. Please fix Cask file. Package file name have changed?

jcs090218:
Cask file fixed! Thanks!

riscy:
Okay, this looks good.  Potentially an interesting way to "exercise" an SSD, too. :)

PR #6562, opened 2019-11-27
add recipe for ksm-window

karrick:
# ksm-window

provides handy window management functions

## Why

I frequently find myself in a situation where I have half a dozen
Emacs windows in a frame, and I want to focus on just one of those
windows for a few moments, and then later return my window state to
its present configuration.  Emacs already has facilities for
accomplishing such tasks, but I wanted a quick toggle function like
tmux' "C-b z" pane toggling feature.

I put this file somewhere Emacs will find it, then added the below to
my ~/.emacs/init.el file.

```Lisp
   (require 'ksm-window)
   (global-set-key (kbd "C-x w") #'ksm-window-zoom-out) ;; restore frame layout from stack
   (global-set-key (kbd "C-x z") #'ksm-window-zoom-in)  ;; save frame layout, then zoom in to buffer
```

I also frequently find myself working with half a dozen buffers in as
many windows when interrupted to help someone else. I want to save my
entire frame layout, associated with a key word, then go help the
person who asked for it, and afterwards, quickly restore my frame
layout.

To do this, I add the following two more lines to my
~/.emacs.d/init.el file:

```Lisp
   (global-set-key (kbd "C-x j") #'ksm-window-config-restore) ;; prompt for NAME and restore config
   (global-set-key (kbd "C-x p") #'ksm-window-config-save)    ;; prompt for NAME and save config
```

provides handy window management functions

https://github.com/karrick/ksm-window

I am the package author.

None Needed

### Checklist

Please confirm with `x`:

I did all the steps, so I'm not sure what to put in the final task, "I have confirmed some of these without doing them".

### DIFF
--- a/recipes/ksm-window
+++ b/recipes/ksm-window
@@ -1,0 +1,3 @@
+(ksm-window :repo "karrick/ksm-window"
+            :fetcher github
+            :files ("ksm-window.el"))
--- a/recipes/ksm-window
+++ b/recipes/ksm-window
@@ -1,3 +1,1 @@
-(ksm-window :repo "karrick/ksm-window"
-            :fetcher github
-            :files ("ksm-window.el"))
+(ksm-window :repo "karrick/ksm-window" :fetcher github)


alphapapa:
It seems like these functions are already provided by Emacs commands.  For example, instead of:

```el
   (require 'ksm-window)
   (global-set-key (kbd "C-x w") #'ksm-window-zoom-out) ;; restore frame layout from stack
   (global-set-key (kbd "C-x z") #'ksm-window-zoom-in)  ;; save frame layout, then zoom in to buffer
```
You can use these built-in tools:
```el
(winner-mode)
(global-set-key (kbd "C-x w") #'winner-undo) ;; restore frame layout from stack
(global-set-key (kbd "C-x z") #'delete-other-windows)  ;; save frame layout, then zoom in to buffer
```
And instead of:
```el
   (global-set-key (kbd "C-x j") #'ksm-window-config-restore) ;; prompt for NAME and restore config
   (global-set-key (kbd "C-x p") #'ksm-window-config-save)    ;; prompt for NAME and save config
```
You could use registers.  From the manual:
```
13.4 Saving Window Configurations in Registers
==============================================

You can save the window configuration of the selected frame in a
register, or even the configuration of all windows in all frames, and
restore the configuration later.  *Note Windows::, for information about
window configurations.

‘C-x r w R’
     Save the state of the selected frame’s windows in register R
     (‘window-configuration-to-register’).
‘C-x r f R’
     Save the state of all frames, including all their windows, in
     register R (‘frameset-to-register’).

   Use ‘C-x r j R’ to restore a window or frame configuration.  This is
the same command used to restore a cursor position.  When you restore a
frame configuration, any existing frames not included in the
configuration become invisible.  If you wish to delete these frames
instead, use ‘C-u C-x r j R’.
```

riscy:
The code looks good, with two small nags:

- ksm-window.el#L41: No need to sharp-quote lambdas (but also no harm)
- ksm-window.el#L12: Use `Version` instead of `Package-Version` (MELPA automatically adds `Package-Version`)

@alphapapa makes a good point, and it's often better if you can implement your package using builtins.  There are also some existing packages on MELPA that I think (but I might be wrong) might form a basis for similar behaviors --

- http://melpa.org/#/edwina
- http://melpa.org/#/es-windows
- http://melpa.org/#/switch-window
- And more generally you might check the list of other packages that manipulate windows [here](http://melpa.org/#/?q=window).

You may be able to add to these packages if they do (almost) what you want, but ping me back if you've had a chance to take a look.

tarsius:
Closing due to inactivity.

PR #6563, opened 2019-11-28
Add helm-switch-eshell

jamesnvc:

Provides a helm interface for easily switching between (e)shells

https://github.com/jamesnvc/helm-switch-shell

I am the author.

**None needed**

### Checklist

### DIFF
--- a/recipes/helm-switch-eshell
+++ b/recipes/helm-switch-eshell
@@ -1,0 +1,1 @@
+(helm-switch-eshell :fetcher github :repo "jamesnvc/helm-switch-eshell")
--- a/recipes/helm-switch-eshell
+++ b/recipes/helm-switch-eshell
@@ -1,1 +1,0 @@
-(helm-switch-eshell :fetcher github :repo "jamesnvc/helm-switch-eshell")
--- a/recipes/helm-switch-shell
+++ b/recipes/helm-switch-shell
@@ -1,0 +1,1 @@
+(helm-switch-shell :fetcher github :repo "jamesnvc/helm-switch-shell")


manuel-uberti:
Two minor notes:
- could you make the colours customizable? A theme developer can then adjust them as preferred
- by replacing `eshell` with `shell` this works for `shell-mode` as well. Would it make sense to make this a more generalised switcher?

jamesnvc:
The only place this actually uses a specific face is for the `[+]` indicator. I made a face for that & the rest should just be whatever the default is.  Will that work, or should I explicitly make faces for everything?

I made the change to make it also find `shell-mode` buffers, but I suppose I should rename this to just something like `helm-switch-shell` now? As you point out, there isn't actually anything eshell-specific...

jamesnvc:
If it's going to allow switching to both `shell`s and `eshell`s, I suppose it should also allow creating either. I will make that an option as well.

manuel-uberti:
> The only place this actually uses a specific face is for the `[+]` indicator. I made a face for that & the rest should just be whatever the default is. Will that work, or should I explicitly make faces for everything?

Great, the face for indicator should be enough, I don't see other hard-coded colours.
 
> I made the change to make it also find `shell-mode` buffers, but I suppose I should rename this to just something like `helm-switch-shell` now? As you point out, there isn't actually anything eshell-specific...

`helm-switch-shell` should be good, but I'll leave that to the MELPA people which I'm sure are going to review your PR more properly than me.

Thanks for this package, I look forward to using it. :)

manuel-uberti:
@jamesnvc thank you for the latest changes. 

I added support for your package to the theme I use ([sanityinc-tomorrow-night](https://github.com/purcell/color-theme-sanityinc-tomorrow)), and I specifically set it to use `shell` instead of `eshell`. Your package works as expected. :+1:

manuel-uberti:
@jamesnvc I noticed you are using `string-prefix-p "*shell*"` to get the candidates to display. 

Would it make sense to use `string-prefix-p "*shell"` instead? This should make switching between project shells (the ones you can spawn via Projectile) easier, because they are named something like `*shell <project-name>`. 

What do you think?

jamesnvc:
@manuel-uberti Sure, I can do that; I typically just use eshell myself, so info on what will make this work better for shell-mode are more than welcome.

manuel-uberti:
It works as expected now, thank you. 

I mainly use project-based shells (`shell-mode` that is), so with regard to my usual workflow you are covering pretty much everything I need. :)

riscy:
Thanks both, this looks good!  There are two small things I'll point out:

- helm-switch-shell.el#L8: Use `Version` instead of `Package-Version` (MELPA automatically adds `Package-Version`)
- helm-switch-shell.el#L24: Prefer `https` over `http` (if possible)

...but after these, this looks ready to go.

jamesnvc:

> Thanks both, this looks good!  There are two small things I'll point out:
>
> - helm-switch-shell.el#L8: Use `Version` instead of `Package-Version` (MELPA automatically adds `Package-Version`)
> - helm-switch-shell.el#L24: Prefer `https` over `http` (if possible)

Done; thank you!

riscy:
Great.  Another tiny thing I'm noticing is you don't need to bring in cl-lib, since you're already package-requiring emacs 25, so you could have:

```
;; Package-Requires: ((emacs "25") (helm "2.8.8") (dash "2.16.0") (s "1.12.0"))
```

This one doesn't appear to be a blocker to me, but I'd fix that line when you get a chance.  And, welcome to MELPA! :)

PR #6575, opened 2019-12-06
Add undo-fu

ideasman42:

Helpers to expose undo/redo functionality on top of Emacs own undo commands
*(without modifying the internal undo data which tends to be buggy, [example](https://www.reddit.com/r/emacs/comments/85t95p/undo_tree_unrecognized_entry_in_undo_list/))*.

- Undo marks the initial step when activated.
- Redo will not redo past this initial undo action.
- Redo will not undo if the last action in the undo stack is not an undo.

These constraints can be disabled temporarily by pressing Ctrl-G.

https://gitlab.com/ideasman42/emacs-undo-fu

Author/Maintainer.

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/undo-fu
+++ b/recipes/undo-fu
@@ -1,0 +1,3 @@
+(undo-fu
+ :repo "ideasman42/emacs-undo-fu"
+ :fetcher gitlab)


manuel-uberti:
Neat idea. At a first (quick) try it works as expected, so thanks for this package.

One minor note: are the newlines [here](https://gitlab.com/ideasman42/emacs-undo-gofer/blob/master/undo-gofer.el#L98) and [here](https://gitlab.com/ideasman42/emacs-undo-gofer/blob/master/undo-gofer.el#L160) really necessary?

manuel-uberti:
Oh, another note: you have two `(provide 'undo-gofer)` at the end.

ideasman42:
Thanks for checking this out. Corrected double provide, the parens on own line is the code-style I prefer, where every block is indented by 2, and multi-line blocks never proceed single line blocks, While it's a little odd, It makes it easy to know how deep blocks are nested and works well with auto-formatting.

riscy:
Just quickly passing through --
```
undo-tastic.el:82:1:Warning: Unused lexical argument `list-to-find'
undo-tastic.el:109:1:Warning: Unused lexical variable `steps'
```
Usually these are easy to fix (prefix with `_`) but especially in the case of `steps` it seems like that variable wasn't meant to be ignored?

ideasman42:
@riscy thanks, both should have been used. corrected.

ideasman42:
Anything I can do to help get this included?

I posted an [RFC to reddit](https://www.reddit.com/r/emacs/comments/eab4ry) and got some positive responses.

Silex:
@ideasman42: @riscy has computer problems, but there's nothing that seem to prevent your package from being included. Just be a bit patient :wink:

riscy:
Very much appreciate seeing the RFC on Emacs, btw (I would encourage everyone to do this).  Thanks for your patience -- this looks good to go!

Silex:
Just wanted everyone to be aware of this interesting undo-tree update that explains the bug: http://www.dr-qubit.org/Lost_undo-tree_history.html

PR #6576, opened 2019-12-08
add dash.info to dash package recipe

emacs18:
Add dash.texi to recipe of dash package so that dash.info would be built and included in the package.  I discussed this with author of dash who gave his blessing to this idea.


Fuco1:
Thanks!

PR #6577, opened 2019-12-08
add ppp recipe

conao3:

Extended `pp` package (builtin).

https://github.com/conao3/p.el

Author and maintainer.

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/ppp
+++ b/recipes/ppp
@@ -1,0 +1,1 @@
+(ppp :fetcher github :repo "conao3/ppp.el")


riscy:
Thanks for your close adherence to the checklist!  A couple quick things I'm noticing:

- The name "p" doesn't make me think of "pp.el extensions", and your source code and docs don't make any mention of `pp` and how this sits in the same niche -- only your PR here does.  "p" is also the name of a new/currently developed programming language.  I realize p is nice and short, but maybe `pp-extensions` or `ppp` or something would be a less risky name/namespace to take on?
- I may be missing something but why are some of the defuns wrapping progns?:
```elisp
(defun p-plist (form)
  "Output the pretty-printed representation of FORM suitable for plist."
  (progn
    (let ((str (with-p--working-buffer form
                 (when (and form (listp form))
                   (forward-char)
                   (ignore-errors
                     (while t (forward-sexp 2) (newline)))
                   (delete-char -1)))))
      (princ (concat str "\n")))
    nil))
```
Couldn't this be written:
```elisp
(defun p-plist (form)
  "Output the pretty-printed representation of FORM suitable for plist."
  (let ((str (with-p--working-buffer form
                                     (when (and form (listp form))
                                       (forward-char)
                                       (ignore-errors
                                         (while t (forward-sexp 2) (newline)))
                                       (delete-char -1)))))
    (princ (concat str "\n")))
  nil)
```
But on that same vein, why return `nil`?

conao3:
`ppp` is good! I rename project `p` to `ppp`.

If this project function use in `*scratch*` returning string is too abundant so I think these functions should return nil.

```emacs-lisp
(ppp-list
 '(ppp-a ppp-b ppp-c))
(ppp-a
 ppp-b
 ppp-c)
"(ppp-a
 ppp-b
 ppp-c)
"

(ppp-list
 '(ppp-a ppp-b ppp-c))
(ppp-a
 ppp-b
 ppp-c)
nil
```

These output strings with `princ` could catch via `with-output-to-string`.

riscy:
That makes sense to me.  Actually I've noticed that functions that yield big return values are not very REPL friendly (yet another advantage of [command/query separation](https://abdullin.com/command-query-separation/)).

This looks good to go!

conao3:
Thanks!

PR #6579, opened 2019-12-09
Add recipe for company-quickhelp-terminal.

jcs090218:

Terminal support for `company-quickhelp'.

https://github.com/jcs-elpa/company-quickhelp-terminal

The maintainer.

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/company-quickhelp-terminal
+++ b/recipes/company-quickhelp-terminal
@@ -1,0 +1,1 @@
+(company-quickhelp-terminal :repo "jcs-elpa/company-quickhelp-terminal" :fetcher github)


jcs090218:
I had tried to collaborate to `company-quickhelp`. Unfortunately the author not willing to maintain the terminal support. Hence this package is created.

* https://github.com/expez/company-quickhelp/issues/99
* https://github.com/expez/company-quickhelp/pull/100

riscy:
Great, thanks for your close adherence to the checklist.

You have a function `company-quickhelp-pos-tip-available-p` which makes these two checks -- `(fboundp 'x-hide-tip) (fboundp 'x-show-tip)` -- but I think this will always be true: these functions are defined in C.

Related: you're relying on`pos-tip-` but it's `company-quickhelp` that is loading `pos-tip` for you -- it's probably better to make the import specific and drop the `pos-tip-available-p` function.

I get the feeling this should be a global minor mode -- someone might load this package, but they will only want to activate it when they're using a terminal-based Emacs session, unless I'm missing something.

jcs090218:
> You have a function company-quickhelp-pos-tip-available-p which makes these two checks -- (fboundp 'x-hide-tip) (fboundp 'x-show-tip) -- but I think this will always be true: these functions are defined in C.

Yeah, I was just overriding the `company-quickhelp-pos-tip-available-p` function. 

> Related: you're relying onpos-tip- but it's company-quickhelp that is loading pos-tip for you -- it's probably better to make the import specific and drop the pos-tip-available-p function.

I am not relying on `pos-tip` yet is `popup` actually. 

> I get the feeling this should be a global minor mode -- someone might load this package, but they will only want to activate it when they're using a terminal-based Emacs session, unless I'm missing something.

I think I handle it by having `(display-graphic-p)` while displaying tool tip hence it wouldn't hamper anything while running this package. And TBH, I don't think this should be a minor mode because this is such a simple package and just one little purpose. 

---

BTW, I have follow the thread [here](https://github.com/company-mode/company-quickhelp/pull/94).

Unfortunately, no one wants to support terminal version. ;(

riscy:
> And TBH, I don't think this should be a minor mode because this is such a simple package and just one little purpose.

Let me try to incentivize it. :)

Imagine someone has a bug that boils down to `(company-quickhelp--cancel-timer)` being called at an unexpected time (which is private, signaling that the author did not expect third parties to rely on it).  They look at the list of minor modes that are active and begin to turn them off in an effort to find out what is causing the problem.  I went through exactly this workflow recently to diagnose a tramp issue.

They will never find the issue if it was because of your package, because your package is overriding the behavior of functions (again, one of them private) without a toggle.

riscy:
That wasn't really an incentive, haha.  More of a justification. :)

jcs090218:
> Imagine someone has a bug that boils down to (company-quickhelp--cancel-timer) being called at an unexpected time (which is private, signaling that the author did not expect third parties to rely on it). They look at the list of minor modes that are active and begin to turn them off in an effort to find out what is causing the problem. I went through exactly this workflow recently to diagnose a tramp issue.

> They will never find the issue if it was because of your package, because your package is overriding the behavior of functions (again, one of them private) without a toggle.

I saw your reason! I am encouraged and added the minor mode! Thanks for your efforts! :D

purcell:
:heart: minor modes

PR #6584, opened 2019-12-12
add tracker

ianxm:

generate graphs of personal metrics from diary entries

https://github.com/ianxm/emacs-tracker

maintainer

none needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/diary-metrics
+++ b/recipes/diary-metrics
@@ -1,2 +1,0 @@
-(diary-metrics :repo "ianxm/emacs-tracker"
-               :fetcher github)
--- a/recipes/diary-metrics
+++ b/recipes/diary-metrics
@@ -1,0 +1,2 @@
+(diary-metrics :repo "ianxm/emacs-tracker"
+               :fetcher github)
--- a/recipes/metrics-tracker
+++ b/recipes/metrics-tracker
@@ -1,0 +1,2 @@
+(metrics-tracker :repo "ianxm/emacs-tracker"
+                 :fetcher github)
--- a/recipes/tracker
+++ b/recipes/tracker
@@ -1,0 +1,2 @@
+(tracker :repo "ianxm/emacs-tracker"
+           :fetcher github)
--- a/recipes/tracker
+++ b/recipes/tracker
@@ -1,2 +1,0 @@
-(tracker :repo "ianxm/emacs-tracker"
-           :fetcher github)


ianxm:
sorry bad link in request. project is here: https://github.com/ianxm/emacs-tracker

alphapapa:
Have you considered a more descriptive name?  i.e. something that would indicate it's for tracking personal metrics in a diary file.

As well, there's already software named Tracker, and there's some Emacs support for it here: https://www.emacswiki.org/emacs/Tracker

riscy:
In addition to @alphapapa's comment, some quick feedback:

### tracker.el

No issues! 💯

```
1 issue found:

368:2: warning: Closing parens should not be wrapped onto new lines.
```

```
tracker.el:88:29:Warning: reference to free variable `diary-file'

In end of data:
tracker.el:522:1:Warning: the following functions are not known to be defined:
    orgtbl-mode, org-ctrl-c-ctrl-c
```
(Probably a `(require 'org)` here)

No issues! 🎉

Other static checks:
- tracker.el#L93: No need to sharp-quote lambdas (but also no harm)
- tracker.el#L130: No need to sharp-quote lambdas (but also no harm)
- tracker.el#L384: No need to sharp-quote lambdas (but also no harm)
- tracker.el#L482: No need to sharp-quote lambdas (but also no harm)
- tracker.el#L93: It's safer to sharp-quote function names (i.e. `encode-time`); use `#'`
- tracker.el#L130: It's safer to sharp-quote function names; use `#'`
- tracker.el#L240: It's safer to sharp-quote function names; use `#'`
- tracker.el#L384: It's safer to sharp-quote function names; use `#'`
- tracker.el#L482: It's safer to sharp-quote function names; use `#'`
- tracker.el#L127: Consider `unless ...` instead of `when (not ...)`

ianxm:
@alphapapa: good point, I wasn't aware of the other tracker. I didn't see anything in `list-packages` so I thought I was clear. I didn't initially go with a more specific name because I wanted it to be short and this time around I was just matching the name I'd chosen for the original tool. anyway, I agree I should rename the emacs package. how about `diary-metrics-reports` for the name and `dmr` as the prefix to namespace the functions and variables? suggestions welcome.

@riscy: sorry, I really had run those tools before submitting, I must have reintroduced those issues later. I corrected them, except
- I left in the sharp-quotes for lambdas. it's a soft warning and when I removed them the emacs 25.2 byte compiler complained.
- I don't see the byte compile warnings about `diary-file` and `orgtbl-mode` so I am not certain I've resolved them.  I added `requires` for the packages containing them. Let me know if there's still an issue.

alphapapa:
MELPA requires that symbol prefixes match the package name.  How about just `diary-metrics`?

Or if you want something more creative, maybe combine it with your username, something like `metrixian`, which is similar to the word *metrician*.  Or you could get even sillier and do something like `diatrician`, sort of combining *diary* with *metrics* and *ian*.  :)

ianxm:
renamed to diary-metrics

riscy:
Thanks for doing that - did you want to rename the GitHub repository as well (or perhaps you've got your own system for naming your repositories, which is fine too).

ianxm:
actually I had a better idea. I will rename it to metrics-tracker. this matches the domainname I have for the original app (metricstracker.com -- it is down at the moment as I've lost my free hosting but I plan to get it up again and allow this package to sync to it). this also disambiguates it from the other tracker and relates to the repo name. the drawback is that it doesn't mention the diary, but anyone that gets as far as the the package summary will see that.

riscy:
Sounds good.  I feel like the package could safely be named metrics-tracker too (or maybe that's your plan already).  Ping me when ready. :)

ianxm:
renamed the package and recipe to metrics-tracker

riscy:
This is cool.  Thanks for your patience!

I wouldn't consider this a blocker, but rather than toggling read-only mode just to clear the buffer such as in `metrics-tracker--setup-output-buffer` with `(read-only-mode -1)`, you can also run something like this:
```elisp
(let ((inhibit-read-only t)) (erase-buffer))
```
and should use that form whenever you need a lisp program to modify a read-only buffer.

PR #6585, opened 2019-12-12
Add posix-manual recipe

lassik:

Provides a `posix-manual-entry` command with tab completion. It lets you easily call up POSIX manual pages in your web browser.

The package contains a database of all POSIX manual page names and their URLs. It's scraped from the Open Group website by a Racket script which is included in the repo.

https://github.com/lassik/emacs-posix-manual

Just wrote it

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/posix-manual
+++ b/recipes/posix-manual
@@ -1,0 +1,1 @@
+(posix-manual :fetcher github :repo "lassik/emacs-posix-manual")


lassik:
This is my first multi-file package, so quick questions:

- Am I using (require 'posix-manual-data) correctly? Should I just (load ...) that file instead?

- Are MELPA and package.el able to extract the long package description from the comment in posix-manual.el?

The package itself is almost trivial. The only reason it's split into two files is because one of them is auto-generated. It's much easier to generate an entire file from scratch than splice generated content into a file that also has manually written parts.

lassik:
Actually, I could replace this package with a more generic one that provides web-based manual page lookup for different Unixes, not only Posix.

riscy:
Hi!

First, including posix-manual-pkg.el in your recipe is discouraged -- MELPA will create a `-pkg.el` file.

> Am I using (require 'posix-manual-data) correctly? Should I just (load ...) that file instead?

Yes but I believe you want to add a `provide` so that the feature gets registered, i.e. `posix-manual-data.el:2179: The footer should be: (provide 'posix-manual-data)\n;;; posix-manual-data.el ends here` (checkdoc should whine about those).

> Are MELPA and package.el able to extract the long package description from the comment in posix-manual.el?

Not sure where to look - what's the long comment (or a line number)?

-------
### posix-manual.el

No issues! 💯

```
3 issues found:

1:0: warning: "Version:" or "Package-Version:" header is missing. MELPA will handle this, but other archives will not.
1:0: error: package.el cannot parse this buffer: Package lacks a "Version" or "Package-Version" header
1:53: warning: You should depend on (emacs "24") if you need lexical-binding.
```

No byte-compilation issues! 😎

No issues! 🎉

### posix-manual-data.el

```
Warning (emacs): 
posix-manual-data.el:0: The first line should be of the form: ";;; package --- Summary"
Warning (emacs): 
posix-manual-data.el:0: You should have a section marked ";;; Commentary:"
Warning (emacs): 
posix-manual-data.el:3: You should have a section marked ";;; Code:"
Warning (emacs): 
posix-manual-data.el:2179: The footer should be: (provide 'posix-manual-data)\n;;; posix-manual-data.el ends here
```

(We can skip this since it's a support file and the function/variable prefixes look correct)

No byte-compilation issues! 😎

No issues! 🎉

Other static checks:
- posix-manual-data.el#L2018: Don't hardcode /tmp.  Use `temporary-file-directory`
- posix-manual-data.el#L2019: Don't hardcode /tmp.  Use `temporary-file-directory`

lassik:
Thank you for the feedback. The new template is really nice :)

> First, including posix-manual-pkg.el in your recipe is discouraged -- MELPA will create a -pkg.el file.

In that case, where do I put the package metadata (keywords, version, summary, etc.)? Does MELPA have some way to figure out what the "main file" of the package is, and look there for the package metadata? Should I have duplicate metadata in both `posix-manual.el` and `posix-manual-data.el`?

> Not sure where to look - what's the long comment (or a line number)?

I was referring to the usual `;;; Commentary:` part.

I'll add the lexical-binding declaration.

The lint warnings about `/tmp` are spurious; in that file, `/tmp` is a substring of public web URLs, not local filenames.

riscy:
> Does MELPA have some way to figure out what the "main file" of the package is, and look there for the package metadata?

I'm not too familiar with the guts of package-build but I think the answer is [here](https://github.com/melpa/melpa/blob/master/package-build/package-build.el#L807-L823).  Basically, MELPA looks for a `-pkg.el` file and if it finds it, it uses that.  If it can't find it, it looks in `<your-package-name>.el`, so that is effectively the "main file".  In fact, if you have different `Package-Requires` among your source files, the Package-Requires that aren't in `<your-package-name.el>` are ignored, and there is at least one package in MELPA that accidentally does this.

Incidentally, it is okay to have a `-pkg.el` file, but doing it incorrectly can break the build.  If you put your package info in your main file then we can use `package-lint` to catch mistakes and enforce consistency.

> I was referring to the usual ;;; Commentary: part.

Oh -- I think these can be quite long, but I don't know which package currently holds the record. :)

riscy:
One other thing I notice with a quick glance -- I can't seem to find a compatible license listed in your repo or in your files.

lassik:
Sorry about the huge delay in getting back to this. I thought it would be better to add the POSIX manual functionality into another, more generic package but it didn't work out so well. Hence it's best to keep POSIX manuals as a separate package after all.

> One other thing I notice with a quick glance -- I can't seem to find a compatible license listed in your repo or in your files.

* Added a `LICENSE` file (ISC).
* Also removed the hand-crafted `-pkg.el`. I'll let MELPA make it own.
* Added `lexical-binding` and docstrings to the auto-generated `posix-manual-data.el`.
* M-x checkdoc is now happy with both `posix-manual.el` and `posix-manual-data.el`.
* M-x package-lint-current-buffer is happy with `posix-manual.el`.
* M-x package-lint-current-buffer is *not* happy with `posix-manual-data.el`, but since MELPA ignores any package header comments in this auxiliary file, it's probably best to keep out such comments and have them only in the main file?

lassik:
Just tried this locally in my clone of the MELPA repo:

```
make clean
make recipes/posix-manual
make sandbox INSTALL=posix-manual
```

Seems to be working fine - `*Compile-Log*` shows no errors in the vanilla Emacs given by `make sandbox` and I'm able to use `M-x posix-manual-entry` from it.

Contents of the tar package generated by MELPA:

```
$ tar -tf packages/posix-manual-20200227.1808.tar
posix-manual-20200227.1808/
posix-manual-20200227.1808/posix-manual-pkg.el
posix-manual-20200227.1808/posix-manual-data.el
posix-manual-20200227.1808/posix-manual.el
```

riscy:
Awesome, thank you. :)  Looks good!

PR #6586, opened 2019-12-13
Add aftersave package

roman:
When developing software, I tend to notice sometimes that I do repetitive tasks
on my session; at some point, I stop and create a quick function on my
`*scratch*` that executes a bunch of steps in one go and add an
`after-save-hook` for it on my current buffer.

This package helps removing manual steps in the development cycle. If you
need to execute the same commands after a save (run tests, restart browser,
etc.), this package allows you to add an Emacs function in the
after-save-hook of the editor (or the local buffer).

https://github.com/roman/aftersave.el

Author and Maintainer

None needed

### Checklist

Please confirm with `x`:

- [ ] I have confirmed some of these without doing them <= cute

### DIFF
--- a/recipes/aftersave
+++ b/recipes/aftersave
@@ -1,0 +1,1 @@
+(aftersave :repo "roman/aftersave.el" :fetcher github)


alphapapa:
How does this compare with other, similar packages on MELPA?  For example, searching for keywords like "hook", "save", etc, I see several similar ones.  If they don't do exactly what this does, maybe they could be improved instead of making a new, slightly different package.

roman:
Hello @alphapapa,

Thank you for taking the time to review this PR. 

> your readme still has the outdated command names with / in them.

Agh, thank you, `checkdoc` and package-linter can do only so much

> How does this compare with other, similar packages on MELPA?

To be honest, it didn't occur to me to look up for the keyword hook, I just did it for the keyword save.

The only that relates to this package (that I can tell) is the [hookify](https://github.com/Silex/hookify) package, it seems you may be able to accomplish a similar behavior, but with some differences:

hookify:
* accepts a hook name as part of the command 
* only works with custom ELISP forms as parameters

The package in this PR on the other hand: 
* Has it's functions only working with `after-save-hook` 
* Offers an add function that works with existing functions in your env (it autocompletes them when calling add interactively)
* Offers a remove function that removes already registered functions to the `after-save-hook` (it also lets you select which function you want to remove)

Keeping in mind that I might be biased :see_no_evil:, I argue, for the purpose of adding `after-save-hook`s on the fly to enhance your development workflow, `aftersave` offers a better UX story. Also, has less dependencies.

About extending `hookify`, I can foresee it was intended to be more general purpose, you can use any hook you want, and you can only use custom forms. This package is very specific and with a non ambiguous purpose.

Looking forward for your thoughts/impressions.

alphapapa:
Hi Roman,

I'm not sure exactly how many similar packages there are on MELPA, but one that comes to mind is https://github.com/wasamasa/firestarter.  It says:

> firestarter lets you execute all kinds of things after saving a buffer. You can run shell commands, execute interactive Emacs commands and evaluate arbitrary pieces of Emacs Lisp. 

That sounds very similar to what your package does.

roman:
Ahh, indeed, `firestarter` seems to be more aligned with `aftersave`, it even offers a way to register shell and custom sexps. I'm going to go ahead and extend that library with the `aftersave/add-hook` functionality and drop this package instead.

Thank you!

PR #6596, opened 2019-12-18
Add recipe for `emake' package.

doublep:

Emake is an Emacs-based build system, targeted solely at Elisp projects. It is an alternative to Cask. Unlike Cask, Emake itself is fully written in Elisp and its configuration files are also Elisp programs. If you are familiar with Java world, Cask can be seen as a parallel to Maven — it uses project description, while Emake is sort of a parallel to Gradle — its configuration is a program on its own.

Emake doesn't have to be installed in user's Emacs, as it is a command-line tool, but its normal installation consists of 1) user copying a (relatively) small `emake` shell script to his machine (using `curl`, for example); 2) Emake bootstrapping itself from Melpa. The second step obviously requires the package being published on Melpa.

https://github.com/doublep/emake

I'm the author.

**None needed**

### Checklist

Please confirm with `x`:

- [ ] `M-x checkdoc` is happy with my docstrings [*]

[*] Since Emake uses docstrings on its own functions to provide help in batch mode, `M-x checkdoc` cannot be really happy. E.g. first line will usually not contain full sentence, etc.

### DIFF
--- a/recipes/emake
+++ b/recipes/emake
@@ -1,0 +1,2 @@
+(emake :repo "doublep/emake"
+       :fetcher github)


alphapapa:
There's already a similar project named emake: https://github.com/vermiculus/emake.el

doublep:
I guess I didn't find it, since it is not on Melpa. It is sort-of similar, but still provides considerably different functionality, so I'm not going to throw my work away.

But obviously I need to rename my package even if there is no confusion on Melpa. I will create a new pull request later.

PR #6598, opened 2019-12-19
Add kubectx-mode recipe

terjesannum:

Change kubectl context/namespace and show in mode line

https://github.com/terjesannum/emacs-kubectx-mode

I am the maintainer

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/kubectx-mode
+++ b/recipes/kubectx-mode
@@ -1,0 +1,3 @@
+(kubectx-mode
+ :fetcher github
+ :repo "terjesannum/emacs-kubectx-mode")


riscy:
Great thanks, and the checklist looks good.  A quick bit of extra feedback:

- kubectx-mode.el#L46: It's safer to sharp-quote function names; use `#'`
- kubectx-mode.el#L100: It's safer to sharp-quote function names; use `#'`
- kubectx-mode.el#L7: Use `Version` instead of `Package-Version` (MELPA automatically adds `Package-Version`)
- kubectx-mode.el#L37: This define-key could overwrite a user's keybindings.  Try: `(defvar my-map (let ((km (make-sparse-keymap))) (define-key ...) km))`
- kubectx-mode.el#L38: This define-key could overwrite a user's keybindings.  Try: `(defvar my-map (let ((km (make-sparse-keymap))) (define-key ...) km))`
- kubectx-mode.el#L91: Consider `unless ...` instead of `when (not ...)`
- kubectx-mode.el#L22: Prefer `https` over `http` (if possible)
- Whether or not a user is turning the minor mode on or off, lines 91 and 92 will run -- line 91 in particular (`(when (not global-mode-string) (setq global-mode-string '("")))` seems like something you *don't* want to do, and rather handle the cases when the variable is nil inside your own functions).

terjesannum:
Thanks for your feedback, I think all of this is fixed now.

riscy:
I tested this out and it seems to work fine on my machine.

However, as I scan for related work, it looks like there is already a tool called [kubectx](https://github.com/ahmetb/kubectx) which is quite popular and about 3 years old.  It looks like this package is not related to that tool.

It looks feasible that someone would actually want to write an Emacs wrapper around the original kubectx, so I feel like this package's name might open us up for confusion.  Would you consider renaming this package to resolve the ambiguity?

terjesannum:
That kubectx is what made me write this emacs package. As the functions is just a wrapper around kubectl, there was no reason to depend on that tool and better to use kubectl directly.
I've added some info about this in the readme.

riscy:
Okay, cool.  Thanks for this, and welcome to MELPA! :)

PR #6599, opened 2019-12-20
Add vimgolf recipe

timvisher:

A major mode that let's you play VimGolf challenges in Emacs.

https://github.com/timvisher/vimgolf.el

I'm the maintainer.

**None needed**

### Checklist

Please confirm with `x`:

- [x] The package is released under a [GPL-Compatible Free Software
  License](https://www.gnu.org/licenses/license-list.en.html#GPLCompatibleLicenses).
- [x] I've built and installed the package using the instructions in
  [CONTRIBUTING.org](https://github.com/melpa/melpa/blob/master/CONTRIBUTING.org)

[VimGolf]: https://www.vimgolf.com/
[vimgolf.el]: https://github.com/timvisher/vimgolf.el

### DIFF
--- a/recipes/vimgolf
+++ b/recipes/vimgolf
@@ -1,0 +1,2 @@
+(vimgolf :repo "timvisher/vimgolf.el" :fetcher github)
+


riscy:
Cool. :)

Just to help out the process, could you restore the original [PR template](https://github.com/melpa/melpa/blob/master/.github/PULL_REQUEST_TEMPLATE.md)?

timvisher:
@riscy Thanks. :)

How's that?

riscy:
Awesome, thank you for that.  I have a hidden motive, which is that it makes the checklist followup semi-automated:

### vimgolf.el

```
Warning (emacs): 
vimgolf.el:170: Argument `separator' should appear (as SEPARATOR) in the doc string
Warning (emacs): 
vimgolf.el:203: All variables and subroutines might as well have a documentation string
Warning (emacs): 
vimgolf.el:307: Argument `challenge-id' should appear (as CHALLENGE-ID) in the doc string
Warning (emacs): 
vimgolf.el:312: Argument `challenge-id' should appear (as CHALLENGE-ID) in the doc string
Warning (emacs): 
vimgolf.el:354: First sentence should end with punctuation
Warning (emacs): 
vimgolf.el:398: Argument `force-pull' should appear (as FORCE-PULL) in the doc string
Warning (emacs): 
vimgolf.el:416: First sentence should end with punctuation
Warning (emacs): 
vimgolf.el:416: Argument `replace' should appear (as REPLACE) in the doc string
Warning (emacs): 
vimgolf.el:489: Argument `arg' should appear (as ARG) in the doc string
Warning (emacs): 
vimgolf.el:546: First sentence should end with punctuation
Warning (emacs): 
vimgolf.el:546: Argument `challenge-id' should appear (as CHALLENGE-ID) in the doc string
```

```
3 issues found:

1:0: error: Package should have a Homepage or URL header.
393:0: error: "*vimgolf-browse-list*" doesn't start with package's prefix "vimgolf".
561:0: warning: Closing parens should not be wrapped onto new lines.
```

```
vimgolf.el:350:16:Warning: reference to free variable
    `url-http-end-of-headers'
vimgolf.el:432:1:Warning: Unused lexical variable `start'

In vimgolf-browse-list:
vimgolf.el:482:40:Warning: `beginning-of-buffer' is for interactive use only;
    use `(goto-char (point-min))' instead.
vimgolf.el:487:1:Warning: Unused lexical argument `arg'

In vimgolf-show-description:
vimgolf.el:539:42:Warning: `beginning-of-buffer' is for interactive use only;
    use `(goto-char (point-min))' instead.
```

No issues! 🎉

Other static checks:
- vimgolf.el#L188: It's safer to sharp-quote function names; use `#'`
- vimgolf.el#L203: It's safer to sharp-quote function names; use `#'`
- vimgolf.el#L172: It's safer to sharp-quote function names; use `#'`
- vimgolf.el#L172: It's safer to sharp-quote function names; use `#'`
- vimgolf.el#L188: It's safer to sharp-quote function names; use `#'`
- vimgolf.el#L188: It's safer to sharp-quote function names; use `#'`
- vimgolf.el#L203: It's safer to sharp-quote function names; use `#'`
- vimgolf.el#L203: It's safer to sharp-quote function names; use `#'`
- vimgolf.el#L24: Prefer `https` over `http` (if possible)
- vimgolf.el#L35: Prefer `https` over `http` (if possible)
- vimgolf.el#L294: Prefer `https` over `http` (if possible)
- vimgolf.el#L296: Prefer `https` over `http` (if possible)
- vimgolf.el#L297: Prefer `https` over `http` (if possible)
- vimgolf.el#L402: You can use `not` or `null`

timvisher:
LOL. Oof. I guess I'll get to that if I can. :)

riscy:
I sympathize. :)  Those won't be so bad though!

timvisher:
@riscy Where do the other static checks warnings come from?

timvisher:
@riscy I'm also struggling with the free variable reference warning. That's declared [here](https://git.savannah.gnu.org/cgit/emacs.git/tree/lisp/url/url-http.el#n45) but `(require 'url-http)` doesn't seem to help the byte compiler find it.

timvisher:
I think I've taken care of everything else.

timvisher:
ping. :)

timvisher:
OK I've incorporated your changes into the branch as well as a few of my own. How're we looking now?

riscy:
If I haven't mentioned that this is cool, let me do that now. ;)

One thing I will note is that this is fairly unconventional:
```elisp
(defun vimgolf-refresh-keystroke-log
    ()
```
but I don't consider it a blocker (usually in elisp the argument form is on the same line as the defun, which you do in some places, but not most).

One last nag that I probably missed in my PR (sorry) is that you should make it possible for users to customize `vimgolf-browse-mode-map` too (right now, changes they make to that variable will be clobbered every time the mode is initialized).  You can use something like this:

```elisp
(defvar vimgolf-browse-mode-map
    (define-key keymap (kbd "TAB") #'vimgolf-show-description)
    (define-key keymap "g" #'vimgolf-browse-refresh)
    (define-key keymap "n" #'vimgolf-browse-next)
    (define-key keymap "p" #'vimgolf-browse-previous)
    keymap)
  "Keymap for browsing VimGolf.")

(define-derived-mode vimgolf-browse-mode special-mode "VimGolf browse"
  "A major mode for completing vimgolf challenges.
\\{vimgolf-browse-mode-map}")
```
Minor:
- "Got title for the challenge at point." -> "Get title..."

Once you fix the keymap, ping me and this looks good to go!

timvisher:
> If I haven't mentioned that this is cool, let me do that now. ;)

Thanks! :)

I'm glad someone ping me on it not being in MELPA anymore. :)

> One thing I will note is that this is fairly unconventional:
> 
> ```emacs-lisp
> (defun vimgolf-refresh-keystroke-log
>     ()
> ```
> 
> but I don't consider it a blocker (usually in elisp the argument form is on the same line as the defun, which you do in some places, but not most).

Ah. I'm glad that's not a blocker. I thought I was consistent throughout the file but I did it by hand so I guess I missed some. I've fixed that now.

> One last nag that I probably missed in my PR (sorry) is that you should make it possible for users to customize `vimgolf-browse-mode-map` too (right now, changes they make to that variable will be clobbered every time the mode is initialized). You can use something like this:
> 
> ```emacs-lisp
> (defvar vimgolf-browse-mode-map
>   (let ((keymap (make-sparse-keymap)))
>     (define-key keymap (kbd "TAB") #'vimgolf-show-description)
>     (define-key keymap "g" #'vimgolf-browse-refresh)
>     (define-key keymap "n" #'vimgolf-browse-next)
>     (define-key keymap "p" #'vimgolf-browse-previous)
>     keymap)
>   "Keymap for browsing VimGolf.")
> 
> (define-derived-mode vimgolf-browse-mode special-mode "VimGolf browse"
>   "A major mode for completing vimgolf challenges.
> \\{vimgolf-browse-mode-map}")
> ```

Good point!

> Minor:
> 
> * "Got title for the challenge at point." -> "Get title..."

Nice.

> Once you fix the keymap, ping me and this looks good to go!

w00t!

timvisher:
@riscy Pushed all that. Let me know if there's anything else.

riscy:
Non-blocking, but fyi:

```
Warning (emacs):
vimgolf.el:569: First sentence should end with punctuation
```

Thanks for your patience. :)

timvisher:
:D https://github.com/timvisher/vimgolf.el/commit/af492bfb869623d0cb4378005e6fb3aa79213101

> Non-blocking, but fyi:
> 
> M-x checkdoc (using version 0.6.1):
> 
> ```
> Warning (emacs):
> vimgolf.el:569: First sentence should end with punctuation
> ```

:D https://github.com/timvisher/vimgolf.el/commit/af492bfb869623d0cb4378005e6fb3aa79213101

tarsius:
Thanks for using the GPL.

Unfortunately the tools used to detect the license are not able to do so in this case because you formatted the license text using markdown. This includes tools such as `licensee`, which is used by Github to display the license in the repository (where it says "View license" for this repository instead). It also includes the tools I use for the Emacsmirror.

You should replace `LICENSE.md` with a new `LICENSE`, which contains the plain text version, which you can copy from e.g. https://raw.githubusercontent.com/magit/magit/master/LICENSE.

Please also add the permission statement to the library header. See e.g. https://github.com/magit/magit/blob/master/lisp/magit.el#L24-L37. Note how this is placed directly before the commentary.

Thanks!

riscy:
@tarsius I'll start being even more stringent about this and making sure the GitHub API is recognizing the licenses.

timvisher:
@tarsius: Thanks for bringing that to my attention.

timvisher:
@tarsius I'm not sure if you've tried this before but I think it would've made more sense to make an Issue on the vimgolf.el repo rather than add a comment to this thread. I'm assuming that a tool you wrote made the comment so it would possibly be easy to change it to open an issue instead or I guess even if you did it by hand I still think it would've made more sense. That way the Issue can be closed directly as activity on the repo rather than in a far-flung comment.

Just a thought. Thanks for everything you do for emacs. :)

tarsius:
In my experience it helps to raise such issues here because:

1. It causes @riscy to improve his tooling. :stuck_out_tongue_winking_eye: 
2. The submitter realizes that while their package was added despite the defect in question, this is still part of the Melpa review. I.e. its more likely that the issue gets addressed in a timely manner than what may appear as some drive-by issue in the upstream repository.

timvisher:
@tarsius Got it. Yeah those reasons make sense. Maybe a both/and approach would be appropriate but obviously that's more work. Thanks for taking the time to explain it to me. :)

purcell:
I really need to tackle https://github.com/purcell/package-lint/issues/83 so that `package-lint` raises this directly too. It's good if Github understands what the license is, but an unambiguous metadata header directly in the elisp is a strong starting point.

PR #6605, opened 2019-12-23
Add recipe for `ob-elm`

BonfaceKilz:

Org-Babel support for evaluating Elm code

https://github.com/BonfaceKilz/ob-elm

I'm the author

None needed

### Checklist

Please confirm with `x`:

#### Note:

- When using package-lint, I found:
```
6 issues found:

6:3: warning: You should include standard keywords: see the variable `finder-known-keywords'.
51:0: error: "org-babel-elm-eoe" doesn't start with package's prefix "ob-elm".
105:0: error: "org-babel-elm-initiate-session" doesn't start with package's prefix "ob-elm".
112:0: error: "org-babel-load-session:elm" doesn't start with package's prefix "ob-elm".
112:0: error: `org-babel-load-session:elm' contains a non-standard separator `:', use hyphens instead (see Elisp Coding Conventions).
134:0: error: "org-babel-elm-var-to-elm" doesn't start with package's prefix "ob-elm".
```

I ignored the above errors because fixing them would mean going against conventions set [here](https://code.orgmode.org/bzg/worg/raw/master/org-contrib/babel/ob-template.el) in the officicial org template for extending languages

### DIFF
--- a/recipes/ob-elm
+++ b/recipes/ob-elm
@@ -1,0 +1,1 @@
+(ob-elm :fetcher github :repo "BonfaceKilz/ob-elm")


riscy:
Thanks, looks useful.

### ob-elm.el

No issues! 💯

```
6 issues found:

6:3: warning: You should include standard keywords: see the variable `finder-known-keywords'.
<snip>
```
The org-babel conventions seem fine, but the above one could be addressed.

```
ob-elm.el:64:29:Warning: `assq' called with 8 args, but requires 2
ob-elm.el:81:36:Warning: assq called with 8 arguments, but accepts only 2
ob-elm.el:64:47:Warning: reference to free variable `pThe'
ob-elm.el:64:52:Warning: reference to free variable `header-args'
ob-elm.el:64:64:Warning: reference to free variable `passed'
ob-elm.el:64:71:Warning: reference to free variable `to'
ob-elm.el:64:74:Warning: reference to free variable `the'
ob-elm.el:64:78:Warning: reference to free variable `src'
ob-elm.el:64:82:Warning: reference to free variable `block\.arams'

In end of data:
ob-elm.el:153:1:Warning: the following functions are not known to be defined:
    org-trim, org-strip-quotes
```
I think a typo snuck in for the first chunk.  A `(require 'org)` is needed for `org-trim`, and I think `org-strip-quotes` isn't available with Emacs 26.3's version of org (I don't know about Emacs 27), so you should add a later-versioned `Package-Require` on org.  Or you could use [org-babel-strip-quotes](https://github.com/bzg/org-mode/blob/master/lisp/org-compat.el#L630-L631) for backward compatibility.

BonfaceKilz:
@riscy Done. I've opted to specify the org version. Here's the merged(upstream) PR: https://github.com/BonfaceKilz/ob-elm/pull/9/files

riscy:
This is almost there, but I'm having trouble with this expression here (maybe you can clarify):
```elisp
  (add-hook 'elm-mode-hook
            (lambda ()
              (setq-local comint-prompt-regexp
                          (concat elm-prompt-regexp "\\|^λ?> "))))
```
Does this have any effect?  For instance I don't think `elm-mode` pays any attention to `comint-prompt-regexp` (but perhaps `elm-repl` does?  Do we care it does for the purposes of this package?)

For reference, I see ob-haskell has something similar,
```elisp
(add-hook 'inferior-haskell-hook
            (lambda ()
              (setq-local comint-prompt-regexp
                          (concat haskell-prompt-regexp "\\|^λ?> "))))
```
but the hook here is being added to inferior-haskell (i.e. the REPL) instead of haskell-mode.

Since I'm not too familiar with ob stuff, I'm not sure what the goal of changing the prompt is.  But usually a change like that (one that a user might want to undo) should be done through a minor mode, not as a hook that gets snuck in when a function is called.  In addition, anything you add as a hook in a package should probably be a named function, not an anonymous one -- named functions can be easily redefined, and named functions are easier to remove from hooks once they're in there.

That was a bit long-winded, but if that expression isn't having any effect, I wonder if you could just remove it?

BonfaceKilz:
Thanks for the review. I've learnt something about things to watch out for when creating a package :smile: :+1: 

I'm not an expert on `ob` myself... I really wanted to evaluate `elm` inside org so I hacked up something that works. With regards, to the minor hook, you are right... It has no effect on the *elm* buffer(and anyways I don't think it matters because the output in the org buffer is what matters). Anyways, I've removed it [here](https://github.com/BonfaceKilz/ob-elm/pull/14/commits/15408f73671d6ad8d8b57812a173fee2b2329745)

riscy:
Great, thanks for this.  And, welcome to MELPA!

PR #6606, opened 2019-12-23
Add ob-graphql recipe

jdormit:

[`ob-graphql`](https://github.com/jdormit/ob-graphql) adds an org-babel execution backend for GraphQL snippets that allows users to execute GraphQL source blocks against their servers and see the results inline in their `.org` files.

https://github.com/jdormit/ob-graphql

I'm the maintainer.

N/A

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/ob-graphql
+++ b/recipes/ob-graphql
@@ -1,0 +1,1 @@
+(ob-graphql :fetcher github :repo "jdormit/ob-graphql")


riscy:
Thanks for this!  For this package-lint warning:
```
7:69: warning: Use a non-snapshot version number for dependency on "request" if possible.
```
I believe it is possible to use a non-snapshot in this case (such as 0.3.2) so you could have a package-requires that looked like this instead:
```elisp
((emacs "24.4") (graphql-mode "20191024.1221") (request "0.3.2"))
```
but that is minor enough to not be a blocker -- but something you should check into.

And, welcome to MELPA!

PR #6609, opened 2019-12-24
Add recipe for ivy-resize.

jcs090218:

Auto resize `ivy' window base on number of candidates

https://github.com/jcs-elpa/ivy-resize

The maintainer.

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/ivy-resize
+++ b/recipes/ivy-resize
@@ -1,0 +1,1 @@
+(ivy-resize :repo "jcs-elpa/ivy-resize" :fetcher github)


alphapapa:
I see in the readme that you linked to the issue where you posted your code only 13 days ago: https://github.com/abo-abo/swiper/issues/649.  This seems like a simple feature that should be built-in to Ivy, and @abo-abo said about the idea:

> Unfortunately, due to the way Emacs minibuffer works, I currently don't know if it's possible to shrink the minibuffer window. Suggestions welcome.

So he seems amenable to including the feature.  Why don't you wait and see if it can be merged into Ivy instead of publishing another package that users will have to discover and install and configure.

jcs090218:
Sorry I am kind of the guy that are being rush and hoping this feature get implemented! I can do both, and not sure if @abo-abo has the time. I think he is being busy or not available recently?

riscy:
The code looks good, but I tend to agree with @alphapapa here -- this behavior feels like it could be part of `ivy`.

In my opinion, the best path forward for users would probably be for you to open a PR (although I'm sure lots of people appreciate seeing that you posted a solution in the issue as a stop-gap).

jcs090218:
I will go for your suggestion, thanks!

jcs090218:
I will reopen this since @abo-abo has mention that he prefer to avoid extra work in `post-command-hook` at [swiper/2389](https://github.com/abo-abo/swiper/pull/2389#event-2922582638). :scream:

alphapapa:
He said:

> Thanks for the PR. But I prefer to avoid doing extra work in post-command-hook. Please review the referenced commit and see if it works for you.

The commit he referenced seems to implement a feature like this.  You haven't commented on the commit nor explained why your package is still necessary.

jcs090218:
Okay, I just saw the reference commit. I wasn't sure what he meant in the first place. Thanks! :D

PR #6613, opened 2019-12-27
add oblique strategies

zzkt:

A version of Oblique Strategies accessible within emacs

https://github.com/zzkt/oblique-strategies

maintainer

N/A

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/oblique
+++ b/recipes/oblique
@@ -1,0 +1,4 @@
+(oblique
+ :repo "zzkt/oblique-strategies"
+ :fetcher github
+ :files ("oblique.el" "README.org" "strategies"))


alphapapa:
IIUC the data files are not released under a GPL-compatible license, so this cannot be published on MELPA.

Although that moots these points, FYI and for future reference:

* https://github.com/zzkt/oblique-strategies/blob/a9cf993075883d2783d51014c297df38ea310f55/oblique.el#L64: For `:group` you should use an actual, existing Emacs customization group.
* You've defined functions and variables whose symbols do not begin with the package name.  `package-lint` warns about these issues, but you checked the box claiming that you addressed `package-lint`'s feedback.
* You need not define `random-elt`, because `seq-random-elt` already exists in Emacs.
* Since you load strategy lines at load time, users who change the desired strategies set won't see any changes until they restart Emacs.  The simplest solution is probably to not read them into a list of lines, but to simply pick one at insertion time.  e.g. something like:

```el
(defun oblique-insert-random-strategy ()
  (insert (with-temp-buffer
            (insert-file-contents oblique-strategies-file)
            (goto-char (random (buffer-size)))
            (buffer-substring (line-beginning-position) (line-end-position)))))
```

zzkt:
Thanks for the feedback. Since the the data files are not released under a GPL-compatible license I'll close the PR

PR #6614, opened 2019-12-27
add litanize recipe

zzkt:

A "Latour Litany" generator for emacs 

https://github.com/zzkt/litanizer

maintainer

N/A

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/litanize
+++ b/recipes/litanize
@@ -1,0 +1,1 @@
+(litanize :repo "zzkt/litanizer" :fetcher github)


riscy:
Thanks for this.  Quick pass:

### litanize.el

```
litanize.el:101:1:Warning: the function `s-chop-suffix' is not known to be
defined.

```
This needs a `(require 's)` which should be added to your `;;; Package-Requires:` line as well.

```
Warning (emacs):
litanize.el:61: First sentence should end with punctuation
Warning (emacs):
litanize.el:71: First sentence should end with punctuation
Warning (emacs):
litanize.el:71: Argument `length' should appear (as LENGTH) in the doc string
Warning (emacs):
litanize.el:84: First sentence should end with punctuation
Warning (emacs):
litanize.el:90: First sentence should end with punctuation

```

```
7 issues found:

1:51: warning: You should depend on (emacs "24") if you need lexical-binding.
8:0: error: Expected (package-name "version-num"), but found (enlive).
9:3: warning: You should include standard keywords: see the variable `finder-known-keywords'.
60:0: error: "random-wp-title" doesn't start with package's prefix "litanize".
70:0: error: "latour-litany" doesn't start with package's prefix "litanize".
89:0: error: "insert-litany" doesn't start with package's prefix "litanize".
```

Other static checks:
- litanize.el#L42: Prefer `https` over `http` (if possible)

zzkt:
Thanks for the feedback.

- `(require 's)` has been added and `Package-Requires:` updated
- `checkdoc` is all OK
- `package-lint-current-buffer` still gives a single error... 

```
90:0: error: "insert-litany" doesn't start with package's prefix "litanize".
```

...however, I think the `insert-` prefix is preferable in this situation (mode independent and more or less consistent with existing `insert-` functions)

riscy:
> ...however, I think the insert- prefix is preferable in this situation (mode independent and more or less consistent with existing insert- functions). If the linting is a problem, does adding something like(fset 'insert-litany 'litanize-insert-litany) work?

I see your point, but it's difficult for me to unilaterally make exceptions.  To be fair to everyone contributing, I try (?) to be consistent.  If you feel strongly about this, it might be a good idea to raise an issue either in `package-lint` or within this repository, where other maintainers with more experience can weigh in.  So this is not a flat "no", but that process might take considerably more time.

You aso make a good point wrt the Emacs documentation, and in fact `package-lint` /does/ have some exceptions along these lines:
```elisp
      (string-match-p (rx-to-string `(seq string-start (or "define" "defun" "defvar" "with") "-" ,prefix)) name)
      (string-match-p (rx-to-string  `(seq string-start "global-" ,prefix (or "-mode" (seq "-" (* any) "-mode")) string-end)) name)
```
But `insert` is not currently one of them -- and if it was, we would have to go with `insert-litanize` (MELPA prefers this admittedly strict policy to prevent collisions, once again in the name of fairness).

I'm still open to a conversation on this, and I'd be willing to help open those issues if it helps. :)

Other than that, I think the code for this package is looking good, and the concept is interesting!

zzkt:
I've renamed the function `insert-litany` to `litanize-at-point` so the package now passes the linting (I've also added a conditional `fset` for the symbol which appears to work but feels like a loophole in `package-lint` and easy to remove if it's a problem) 

If you think it's useful to open the relevant issues for MELPA and/or `package-lint` please go ahead. tag me and I'll follow up if needed...

riscy:
Thank you for accommodating!  If you're okay with `litanize-at-point` that might be the way to go.  I believe using `fset` is a loophole :( -- it's basically the same as using `defalias`, which `package-lint` will complain about in a similar way.  I'll actually raise _that_ one as an issue.

purcell:
> (I've also added a conditional `fset` for the symbol which appears to work but feels like a loophole in `package-lint` and easy to remove if it's a problem)

Yeah, the `fset` workaround is bogus. If you're *really really* worried about transition for existing users (and you really shouldn't be, IMO), then you can consider using `define-obsolete-function-alias` in this case. Otherwise, users can just generally `defalias` whatever names they find most convenient in their own configs.

zzkt:
I'm more curious than worried. so removed the `fset` clause and hope it's ok to merge now.

PR #6615, opened 2019-12-27
Add recipe for autobuild

erjoalgo:

Define and execute custom build rules and compilation pipelines.

https://github.com/erjoalgo/autobuild

Author, maintainer.

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/autobuild
+++ b/recipes/autobuild
@@ -1,0 +1,1 @@
+(autobuild :fetcher github :repo "erjoalgo/autobuild")


alphapapa:
There are several similar such packages on MELPA.  See, e.g. this recent PR for one and the packages mentioned in the comments: https://github.com/melpa/melpa/pull/6586  It would be good to compare and contrast your package with existing solutions.

erjoalgo:
This package is different from either aftersave or firestarter in that it provides a menu of various applicable build options, as opposed to a providing a single sequence of hooks that are always executed. This allows users to switch back-and-forth between different options for building/compiling.
27-Dec-2019-19:15:27

It also makes it very easy to define applicable build rules based on some context. For example. I can easily define that on `ngnix-mode` buffers I want the options to syntax-check the file, or restart nginx:

```lisp
(autobuild-define-rule autobuild-nginx-restart (nginx-mode)
  "sudo service nginx restart")

(autobuild-define-rule autobuild-nginx-test-config (nginx-mode)
  "sudo nginx -t")
```

For example, while making significant changes to an nginx file, initially I would like to build using `autobuild-nginx-test-config` to check that the file syntax is still correct. But when making small changes, like changing a port number or server name, I would like to use `autobuild-nginx-restart` directly. `autobuild` makes it easy to define and switch between such distinct flows. 

It is also possible to add more complex logic for the applicability of a build rule, its priority, and its build definition. This can be useful when defining custom flows to custom contexts, for example,  languages internal to a company associated with a custom set of build tools.

alphapapa:
Cool, it would be great if that part were documented in the readme too.

riscy:
Thanks for that explanation.  Quick pass:

### autobuild.el

```
autobuild.el:275:1:Warning: Unused lexical variable
`ansi-color-for-comint-mode'
```
This of course comes from this:
```elisp
    (let* ((ansi-color-for-comint-mode t))
      (compile cmd))))
```
which is correct, and you could skip the byte-compilation warning with this,
```elisp
    (let* ((ansi-color-for-comint-mode t))
      ansi-color-for-comint-mode  ; compiler pacifier
      (compile cmd))))
```
but people who haven't turned on `ansi-color-comint-mode` may have [avoided it for a reason](https://no-color.org/), which you may want to take under consideration.

No issues! 💯

No issues! 💯

Other static checks:
- autobuild.el#L353: No `format` required; messages are already f-strings
- autobuild.el#L11: Prefer `https` over `http` (if possible)
- autobuild.el#L357: It is beneficial to sharp-quote hooks; use `#'`

erjoalgo:
Thanks for all the feedback, I've addressed it in https://github.com/erjoalgo/autobuild/pull/3.

riscy:
Great, thanks for addressing those.  Looking a little more closely now:
```elisp
;; TODO use pipeline hook, not compilation hook
(add-hook 'compilation-finish-functions #'autobuild-notify)
```
Ideally the behavior of Emacs isn't changed simply from loading a package.  In this case, it looks like you want to define a global minor mode which, when toggled, adds and removes this hook.

- It looks like you could *almost* use the default recipe for this (which would exclude your test file automatically) if you were to move `autobuild-examples.el` into a subdirectory.  There is a preference for using the default recipe when you can, so I'd say that's something to consider.
- You don't have to `progn` the else part of an `if` expression, so on lines 159 and 184 you could remove the `progn`
- autobuild.el#L402: It's safer to sharp-quote function names; use `#'`

erjoalgo:
Regarding the concern about modifying emacs' behavior upon loading the package.

In autobuild add a few functions to compilation hooks/advice:
```
(add-hook 'compilation-finish-functions #'autobuild-pipeline-continue)
(advice-add #'compilation-start :around #'autobuild-compilation-buffer-setup)
(add-hook 'compilation-finish-functions #'autobuild-notify)
```

The first two do nothing unless the user has specifically used autobuild to create the compilation buffer, and this behavior is gated by an autobuild-specific buffer-local flag `(bound-and-true-p autobuild-pipeline-rules-remaining)`

My understanding with the way autoload works is that even if a package is available in the load-path, it will not be loaded until the user specifically requires it, or until the user makes use of a function that triggers the autoload.

I can think of a few drawbacks to adding a minor mode to add/remove these hooks:

- some of the `autobuild-*` functions may or may not work properly depending on whether `autobuild-mode` is enabled. For example, if someone executes an autobuild pipline of two steps, where the first step is a compile command, the user will see that the first command executes successfully, but mysteriously not the second if `autobuild-mode` was accidentally disabled because the hooks required for `autobuild-pipeline` to work had been disabled.
- adds usage complexity. now instead of just a set of functions and macros that the user may invoke freely independently of state, now a state variable `autobuild-mode` is introduced which needs to be handled by the user.
- unlike other modes which visibly influence emacs' behavior, for example, show-paren-mode which toggles highlighting of the parenthesized expression at point, the minimal autobuild setup code only takes effect when the user explicitly invokes an autobuild command or function.*

Regarding this:

1. Is your concern with all three hooks, or specifically with `(add-hook 'compilation-finish-functions #'autobuild-notify)`, which does provide a default that does change emacs' behavior? If the latter, could this be addressed by making the default do nothing? This could be achieved by setting the default value of `autobuild-notification-function` to nil and gating the hook function on this value being non-nil.
2. If the former, would it be appropriate to enable the global minor mode by default? This is to reduce the risk of `autobuild-*` functions not working after the user explicitly invoked them or explicitly required autobuild because they forgot to also enable the global minor mode.
3. One last alternative I can think of is to check for the state of `autobuild-mode` whenever a function's correctness depends on the prior setup of autobuild hooks.

erjoalgo:
Re: default package. I was planning to clean up `autobuild-examples` and make it available as a separate package, as a convenient way to load some useful and widely-applicable autobuild rules. Since I am planning to keep both files in the same repo (shared history, and simply makes sense), I think that at least one of the two recipes will still need the `:files` specification even if I moved the examples file to a subdirectory. Given this, would it still be beneficial to try to omit the `:files` spec and is there any significant drawback to it?

erjoalgo:
I am addressing this feedback in https://github.com/erjoalgo/autobuild/pull/6. Will merge once other questions are resolved.

riscy:
> In autobuild add a few functions to compilation hooks/advice:
>
> (add-hook 'compilation-finish-functions #'autobuild-pipeline-continue)
> (advice-add #'compilation-start :around #'autobuild-compilation-buffer-setup)
> (add-hook 'compilation-finish-functions #'autobuild-notify)
>
> The first two do nothing unless the user has specifically used autobuild to create the compilation buffer, and this behavior is gated by an autobuild-specific buffer-local flag (bound-and-true-p autobuild-pipeline-rules-remaining)

Gating behavior is one of the things minor modes are made for.  Here's a sketch of what I'm thinking:
```elisp
;; NOTE we have to define-minor-mode before using the `autobuild-mode` variable
(define-minor-mode autobuild-mode
  "Define and execute build rules and compilation pipelines."
  :lighter " a"
  :global t
  (if autobuild-mode
      (add-hook 'compilation-finish-functions #'autobuild-pipeline-continue)
    (remove-hook 'compilation-finish-functions #'autobuild-pipeline-continue)))

(defun autobuild-pipeline-continue (compilation-buffer finish-state)
  "Internal.  Used to resume an asynchronous pipeline.

   COMPILATION-BUFFER FINISH-STATE are the arguments passed
   to functions in ‘compilation-finish-functions'."
  (when autobuild-mode  ; <-- you get this variable for free to tell you whether autobuild-mode is on
    (with-current-buffer compilation-buffer
      (when (bound-and-true-p autobuild-pipeline-rules-remaining)
        ;; NOTE I took some liberties with rewriting this if to avoid progn:
        (if (autobuild-compilation-succeeded-p finish-state)
            (progn
              (message "continuing with pipeline: %s" autobuild-pipeline-rules-remaining)
              (autobuild-pipeline-run autobuild-pipeline-rules-remaining))
          (message "aborting pipeline: %s" autobuild-pipeline-rules-remaining)
          (setq-local autobuild-pipeline-rules-remaining nil))))))
```

> My understanding with the way autoload works is that even if a package is available in the load-path, it will not be loaded until the user specifically requires it, or until the user makes use of a function that triggers the autoload.

Simply loading a package should not change Emacs' behavior.  If someone were to load your package to make use of a single function or reuse a handy constant you've defined, they'll all of a sudden have hooks that will be nontrivial for them to remove.  I realize these hooks have a fairly localized behavior by design, but it's better to be defensive and use minor modes (when possible) to gate these kinds of modifications.

I'll stop there so I can make sure we're on the same page.  In the meantime, I'll also add:

- "autobuild-mode-filer-applicable-p" -> "autobuild-mode-filter-applicable-p"

erjoalgo:
Thanks for the feedback, I will address it.

riscy:
Thanks!  Ping when you're ready.

erjoalgo:
Please take another look. There are a couple of checkdoc warnings which I believe to be false-positives.

riscy:
Great, thanks.  One thing I'm seeing is this
```elisp
(advice-add #'compilation-start :around #'autobuild-compilation-buffer-setup)
```
at the top-level, and I'd say the same thing about adding advice as I would hooks -- when it's possible, it seems to fit better as something you add/remove when enabling/disabling the mode?

One easy way around the current checkdoc issue is to quote the mode-name inside the string:
```elisp
(error "`autobuild-mode' is not enabled")
```
(I like this solution because it incidentally adds elisp syntax highlighting.)

Another thing that I've been advised to be more rigorous about when reviewing (and this affects you, sorry :) ) is to ensure that packages that use GitHub are being packaged with LICENSE files that are detectable by the GitHub API, for example you might use the one [MELPA itself uses](https://raw.githubusercontent.com/melpa/melpa/master/LICENSE.md).

This one is minor, but I notice not all of the arguments of `autobuild-candidate-select` are documented (checkdoc in Emacs 26 has a bug where if you list a couple out of the order in which they appear, it won't check for the rest).

Also minor:
- piepine -> pipeline?

erjoalgo:
Thanks for the detailed review. Addressed in https://github.com/erjoalgo/autobuild/pull/8.

riscy:
Re: the license, the GitHub API is still showing "Other".  Maybe it is taking a second to refresh, but alternatively you may just want to copy over the MELPA LICENSE.md, which is correctly detected.  Ping me if that changes!

Also regarding putting other packages in the same repository (and not using the default recipe), I can quote tarsius:

> I generally encourage maintainers to use more than a single repository for related, yet distinct libraries/packages. The reason for this is that when those only partially related libraries are located in the same repository, then it becomes difficult to use just one of the libraries. Users of Melpa are not affected by this, but people who install packages by cloning their Git repository are. These likely are the users that contribute back, so you shouldn't make it harder for them. Users might end up loading outdated bundled copies of third-party libraries because of this.

It's a small thing in this case, I think, but I'll leave that for your consideration.

erjoalgo:
Thanks. Actually, I am using the same `LICENSE.md`, so I am not sure why github has not yet detected it:

```
█[debianx1e][git][0]$ sha256sum ./{autobuild,melpa}/LICENSE.md
ccb349b4132ed7737f25e5adebfe61f3d52dca33708df1e50352320438d1d4c2  ./autobuild/LICENSE.md
ccb349b4132ed7737f25e5adebfe61f3d52dca33708df1e50352320438d1d4c2  ./melpa/LICENSE.md
```

I'd still prefer to keep `autobuild.el` and `autobuild-examples.el` in the same repository since these are not really distinct libraries. Right now, the repository is simple and small enough that I haven't needed to create subdirectories. Is there a strong downside to using the `:files` argument?

riscy:
The benefit is it allows your package to grow without opening additional PR's to MELPA (less downstream maintenance).  I'm not sure what's going on with the license... you could do me a favor and try using a plaintext one and see if that works?  I'm still trying to figure this out too.

riscy:
> you could do me a favor and try using a plaintext one and see if that works? I'm still trying to figure this out too.

I can test this out myself, too.  I'll get back to you.

riscy:
Yeah, if you could make your license a plaintext file [like this](https://github.com/riscy/license-tests) it will work.  I'm not sure why the MELPA repository can get away with the license it's using -- there may be some additional check I'm missing, but in the meantime just leaving it as plaintext will work for our purposes.

erjoalgo:
Thanks, using the plaintext LICENSE file did work.
I am using the default package now.

riscy:
Thanks so much for your patience. :)

PR #6617, opened 2019-12-29
Add recipe for sfz-mode

jpcima:

This is a major mode for editing the SFZ sampled instrument format.

https://github.com/sfztools/emacs-sfz-mode

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/sfz-mode
+++ b/recipes/sfz-mode
@@ -1,0 +1,1 @@
+(sfz-mode :fetcher github :repo "sfztools/emacs-sfz-mode")


riscy:
Looks solid!  Quick run-through:

### sfz-mode.el

```
sfz-mode.el:90:53:Warning: reference to free variable `char'

In end of data:
sfz-mode.el:159:1:Warning: the following functions are not known to be
defined: seq-doseq, char, seq-some, seq-contains, seq-filter, seq-empty-p
```
All fixed with `(require 'seq)` I believe.

No issues! 💯

No issues! 💯

Other static checks:
- sfz-mode.el#L107: It's safer to sharp-quote function names; use `#'`

I notice you're manually defining the major mode, but maybe `define-derived-mode` would be more appropriate, since then you could derive from `prog-mode` (this does look like a code mode), e.g. something like:
```elisp
(define-derived-mode sfz-mode prog-mode "SFZ"
  "Major mode for editing SFZ files.
\\{sfz-mode_map}"
  :group 'sfz
  :syntax-table sfz-mode-syntax-table
  (set (make-local-variable 'font-lock-defaults) '(sfz-font-lock-keywords))
  (set (make-local-variable 'indent-line-function) 'sfz-indent-line))
```

jpcima:
Hi @riscy, thanks for review and help. I have updated the upstream as suggested.

- the function `identity` was sharp-quoted;
- the library `seq` was required;
- I have used `define-derived-mode`, and tested.

I run the same Emacs version, somewhat the warnings have escaped me on the first run of `byte-compile-file`.

(also I have assumed the sharp-quoting to not apply to `indent-line-function`, referring to GNU Emacs's own modes I find regular quoting as being common practice about this)

riscy:
> (also I have assumed the sharp-quoting to not apply to indent-line-function, referring to GNU Emacs's own modes I find regular quoting as being common practice about this)

Yeah, `indent-line-function` is in fact a variable, and sharp-quoting is to indicate the names of functions.  Not every codebase will sharp-quote functions, especially a lot of legacy elisp, but it's a [good idea](https://endlessparentheses.com/get-in-the-habit-of-using-sharp-quote.html).

I notice you're distributing under the MIT license which is fine, but you don't write the words "MIT license" anywhere, which you should.  It may also be a good idea to add a LICENSE file to your repository (which GitHub can detect, thus giving your repo a little icon and enabling the API to return your license).

Another couple small nags:
- sfz-mode.el#L118: Single-line comments should (usually) begin with `;;`
- sfz-mode.el#L120: Single-line comments should (usually) begin with `;;`
- sfz-mode.el#L123: Single-line comments should (usually) begin with `;;`
- sfz-mode.el#L125: Single-line comments should (usually) begin with `;;`

Beyond those the code is looking good, so ping when ready!

jpcima:
Hi @riscy, it's ready.
I've applied some changes as you have suggested in the review.
sfztools/emacs-sfz-mode@614506b

riscy:
Apologies for getting so far behind on this.  This looks good to go!

Welcome to MELPA. :)

PR #6620, opened 2019-12-30
Add recipe: clipetty

spudlyo:

Clipetty is a minor mode for terminal (TTY) users that sends text that you kill in Emacs to your Operating System's clipboard.

https://github.com/spudlyo/clipetty

I am the author, maintainer, and enthusiastic user.

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/clipetty
+++ b/recipes/clipetty
@@ -1,0 +1,1 @@
+(clipetty :repo "spudlyo/clipetty" :fetcher github)


riscy:
This looks great.  These are non-blockers, but some quick feedback:

- clipetty.el#L166: Consider `unless ...` instead of `when (not ...)`
- clipetty.el#L104: Could this be simplified to `(if (and tmux-ssh-tty...`?
- Since you defined a minor mode, the function `clipetty-p` isn't necessary I believe -- you can just access the variable `clipetty-mode` directly.

spudlyo:
Thanks for the feedback, these suggestions are great!

PR #6621, opened 2019-12-31
Add org-gtd recipe

Trevoke:

Org-gtd is a wiring of org-mode and (currently) org-brain that aims to follow faithfully the GTD processing framework.

https://github.com/Trevoke/org-gtd.el

None needed.

### Checklist

Please confirm with `x`:

I'm not checking that last box :p 

I'm not sure what "cleanly" means - I get a warning for  `org-gtd.el:35:1:Warning: cl package required at runtime` Because I'm using `find-if` and I don't know how else to do it.

Also, I'm not sure if this actually _fits_ the requirements for a package that goes into Melpa - to some extent, it's mostly just org-mode code and anyone could do it this way. I'll let you be the judge :D

### DIFF
--- a/recipes/org-gtd
+++ b/recipes/org-gtd
@@ -1,0 +1,1 @@
+(org-gtd :fetcher github :repo "trevoke/org-gtd.el")


alphapapa:
This looks like a package that I'm sure many Org users will be interested in!

> I'm not sure what "cleanly" means - I get a warning for org-gtd.el:35:1:Warning: cl package required at runtime Because I'm using find-if and I don't know how else to do it.

Use `cl-lib` and `cl-find-if` instead of `cl` and `find-if`.

Here's some random feedback:

* I think it really calls for a narrated video demonstrating how it's used.  Even for users familiar with GTD, seeing how these commands you've made are used to implement it would be very helpful.
* > At this early stage of the code, the package is destructive. It will overwrite the following variables. 

  You should definitely avoid overwriting those variables, because loading your package should not overwrite user settings like that.  If you think that's fundamental to the package, then, sadly, it probably shouldn't be published on MELPA.  But maybe you could move them to a setup command and instruct users to either configure the variables themselves or use the setup command, in full knowledge that their settings will be overwritten.

* > Then run (org-gtd-init). Run this function every time you change one of the above values.

  You can use `defcustom` setters to do these kinds of things automatically, e.g. https://github.com/alphapapa/magit-todos/blob/40d268f9b487ad7e3b7aff2da2bc44e596bc8052/magit-todos.el#L171

* In general, you may find `org-ql` helpful for some of your code.
* When defining a `defcustom`, it's not necessary to specify the group, because it automatically uses the `defgroup` last defined in the file.
* I'm not sure why you used `:risky t` for so many of your `defcustoms`--that's probably not necessary.
* You should probably not be doing these, because simply loading your file shouldn't do these:

```el
(setq org-edna-use-inheritance t)
(org-edna-load)
```
* https://github.com/Trevoke/org-gtd.el/blob/2e8693a2437a9bbc6f43e5a44d15d66bfb07e35c/org-gtd.el#L117  See `find-buffer-visiting` and `find-file-noselect`.
* https://github.com/Trevoke/org-gtd.el/blob/2e8693a2437a9bbc6f43e5a44d15d66bfb07e35c/org-gtd.el#L123 `org-map-entries` can be quite slow on large files.  `org-ql-select` can help when appropriate.
* https://github.com/Trevoke/org-gtd.el/blob/2e8693a2437a9bbc6f43e5a44d15d66bfb07e35c/org-gtd.el#L144  Use `cl-case` or `pcase`.
* https://github.com/Trevoke/org-gtd.el/blob/2e8693a2437a9bbc6f43e5a44d15d66bfb07e35c/org-gtd.el#L155 You should probably use `outline-next-heading`-type functions when appropriate, rather than line-oriented ones.
* https://github.com/Trevoke/org-gtd.el/blob/2e8693a2437a9bbc6f43e5a44d15d66bfb07e35c/org-gtd.el#L193 It may be a matter of style, but the Org Agenda negates the need to refile things into strict outline hierarchies as their statuses change.  You might want to reconsider this part of the workflow.
* It would help a lot if the `org-brain` parts of the functionality were demonstrated or explained more thoroughly.
* https://github.com/Trevoke/org-gtd.el/blob/2e8693a2437a9bbc6f43e5a44d15d66bfb07e35c/org-gtd.el#L206 This function has some "code smells."  Erasing a buffer and inserting an entire other buffer shouldn't generally be necessary.
* https://github.com/Trevoke/org-gtd.el/blob/2e8693a2437a9bbc6f43e5a44d15d66bfb07e35c/org-gtd.el#L243 I can't tell what this function is supposed to be doing, but it's almost certainly doing it wrong.  :)  
* In general, you don't need to say `Private function` in the docstrings.  Try to write them imperatively, like `Return VALUE for ARG`, or `Set foo for BAR`, so users will know what the functions' purpose is.

Hope this helps.  Keep working on this, because I think it could be very popular.

Trevoke:
Whoa, such feedback! Alright, let's see.

### Video demo

Sure, that sounds like it'd be helpful. Any suggestions for format? GIF.. Youtube.. Something else?

### Clean compilation through cl-lib

This will be done.

### Using defcustom's `:set`

I tried it that way for a bit, had some issues with putting together the two-way logic: if the root directory changes, the paths to the files change, and the paths to the files are dependent on the names of the files, so if the names change, the path changes.

I think for v1 I'm going to simplify it and not let the filenames be customizable - that'll allow me to drop a whole category of complex code.

### org-ql

Maybe? I looked at it when it came out, and I looked at after your comment, but I don't completely understand the documentation. Like.. It returns non-nil ? Does this mean it returns `t` ? Or does it return a list of org elements?

I don't know if I'd have a use for it beyond what I use org-edna for, speaking of which..

### org-edna

I need org-edna to automatically change the following sibling heading from TODO to NEXT when a NEXT becomes DONE, and I need the variable to be set before I load org-edna because that's how org-edna works - I think for this limited requirement, I could actually just do it programmatically. I'll see if I can drop the dependency entirely.

### defcustom and groups

I know that the documentation says I don't need to do that but I ran into some flychecking issues. I'll try that again.

### `:risky t`

I'll remove that! The less code the better.

### `find-file-noselect`

Okay, that seems like a better thing to do programmatically.

### `org-map-entries`

What's "large files" ?

This function should be used daily on the inbox, which means I suppose it could get up to 200 entries in a single day if things got quite hectic, but the only time it might be a very, very large file is if someone is just starting with GTD and they're spending the suggested two-three days to capture all of their incoming stuff, in which case it might be, I dunno, 3,000 at a guess? And it probably wouldn't be worth it to put *all of it* in the electronic inbox, so I doubt we'd get to that kind of size.

### `cl-case`

Huh, that hadn't gotten flagged by the compiler, I'll switch it to `cl-case` though.

### `outline-next-heading` type functions

What I'm actually trying to do here is add a timestamp as text under the current org heading. I could have sworn there was a function for that but I completely failed to find one, soooooo I'm doing it like this. I'm open to better ways of doing it, though I don't think `outline-next-heading` will help here.

### Refiling and strict hierarchies

You're right, I canjust move this to the One-Offs instead. There's two reasons for having at least one separate hierarchy (the `Projects`):
- finding stuck projects
- triggering a state change in a project (TODO -> NEXT) as indicated above when an action goes from NEXT to DONE

### `org-brain`

Yeah. The tricky part is that to some extent if I do that I'm just explaining how `org-brain` works -- which of course, its documentation does well. I could add something (text/video) for how we interact with it in the readme.

### The `project` buffer, and erasing the whole buffer

As written, yes, it's pretty smelly, however:
- this is a temp buffer, and it'll be used while the inbox gets processed. This means the user might see this buffer more than once, and it'll be for different purposes. It's either "erase the buffer" or "kill the buffer and create a new one"
- I'm inserting "a whole buffer", though it's a buffer narrowed to a single org heading, so I'm inserting a single org heading in the buffer

So while it's kind of smelly, I think it's actually the right behavior.

### "wtf is this function doing"

Yeah, probably. This function will go away when  I remove the ability to customize the filenames, which will probably make it easier to adjust the internal behavior of the package as I go anyway.

### `Private function` in the docstring

I think they're already written imperatively, in no small part because flycheck / already pings me about this, but I'll check again.

I'll remove `Private function` from the docstrings.

I'll ping here again when I've made the adjustments I mentioned above (and I'll list them in the new comment) -- in the meanwhile, I'm curious to hear your replies to my questions and thoughts. Thanks for the feedback!

alphapapa:
> Whoa, such feedback! Alright, let's see.
> 
> ### Video demo
> Sure, that sounds like it'd be helpful. Any suggestions for format? GIF.. Youtube.. Something else?

I'd recommend both, actually: a narrated YouTube video to show the whole workflow, and short, "targeted" GIFs in the readme showing simple features (`emacs-gif-screencast` is great for the GIFs, and my Bashcaster script can be helpful for recording outside of Emacs, if you don't already have a solution for that).

> I think for v1 I'm going to simplify it and not let the filenames be customizable - that'll allow me to drop a whole category of complex code.

Sounds like a good plan.  Incremental development, release-early-release-often, etc.  :)

> ### org-ql
> Maybe? I looked at it when it came out, and I looked at after your comment, but I don't completely understand the documentation. Like.. It returns non-nil ? Does this mean it returns `t` ? Or does it return a list of org elements?

Your question here is more about Lisp and predicate functions than about `org-ql`.  Check the code examples in the readme for more information, and look carefully at the results returned (in the `;;=>` comments).

> I need org-edna to automatically change the following sibling heading from TODO to NEXT when a NEXT becomes DONE, and I need the variable to be set before I load org-edna because that's how org-edna works - I think for this limited requirement, I could actually just do it programmatically. I'll see if I can drop the dependency entirely.

I didn't mean that you shouldn't use `org-edna`, just that you shouldn't have those lines of code there that run when your package is loaded.

> ### `org-map-entries`
> What's "large files" ?

Good question.  :)

> This function should be used daily on the inbox, which means I suppose it could get up to 200 entries in a single day if things got quite hectic, but the only time it might be a very, very large file is if someone is just starting with GTD and they're spending the suggested two-three days to capture all of their incoming stuff, in which case it might be, I dunno, 3,000 at a guess? And it probably wouldn't be worth it to put _all of it_ in the electronic inbox, so I doubt we'd get to that kind of size.

200 is probably fine, but consider that many GTD users are procrastinators who probably, eventually, accumulate large inboxes.  (Why would they use GTD if they weren't natural procrastinators? haha)  In general, `org-map-entries` can be slower than `org-ql` when `org-ql` is appropriate, but use whichever is best for your needs.

> ### `outline-next-heading` type functions
> What I'm actually trying to do here is add a timestamp as text under the current org heading. I could have sworn there was a function for that but I completely failed to find one, soooooo I'm doing it like this. I'm open to better ways of doing it, though I don't think `outline-next-heading` will help here.

Ah, I see.  Yes, I think this is a missing function in Org.  You might find the `unpackaged/org-forward-to-entry-content` function useful, from https://github.com/alphapapa/unpackaged.el.  It's important to make sure that you don't insert lines between headings and their planning lines and drawers, which this function attempts to skip properly.

> 
> ### Refiling and strict hierarchies
> You're right, I canjust move this to the One-Offs instead. There's two reasons for having at least one separate hierarchy (the `Projects`):
> 
> * finding stuck projects

That's an example of where `org-ql` would probably be helpful.  See the `org-ql-view` command and examples in the `org-ql` readme.

> ### The `project` buffer, and erasing the whole buffer
> As written, yes, it's pretty smelly, however:
> 
> * this is a temp buffer, and it'll be used while the inbox gets processed. This means the user might see this buffer more than once, and it'll be for different purposes. It's either "erase the buffer" or "kill the buffer and create a new one"
> * I'm inserting "a whole buffer", though it's a buffer narrowed to a single org heading, so I'm inserting a single org heading in the buffer

I see, thanks for the explanation.

> ### `Private function` in the docstring
> I think they're already written imperatively, in no small part because flycheck / already pings me about this, but I'll check again.

Do you mean `checkdoc`?  AFAIK, Flycheck itself doesn't check docstrings.

BTW, you may find this useful.  I've been working on it recently, and I think it works pretty well: https://github.com/alphapapa/makem.sh

Trevoke:
Okay, here's the new status (not done, then done):

## NOT DONE

1. video demo. I'm working on the script and I've made some small updates to the readme in that direction
2. better tutorial / readme in the elisp file. When I have something I'm happy with in the readme.org, it'll go in the elisp file too.
4. org-brain documentation. Again, upcoming.
5. ~unpackaged/org-forward-to-entry-content~ -- I did not realize I could break org by adding something between the heading and the drawers, I will add a fix for this.

## DONE
1. docstrings. Yes, I meant checkdoc.
2. org-edna - that was moved to the user's config file through the readme
3. finding stuck projects - this is built-in to org-mode through `(org-agenda-list-stuck-projects)` and I found a way to not fully destroy the user's settings (basically save them, override them, call the function, restore them)
4. `:set`, `:risky`, etc - making the filenames un-customizable allowed me to remove a large amount of inscrutable code, so that's on longer necessary.
5. `cl-lib`, `cl-case`, `cl-find-if` - done.
6. `find-file-noselect` - done.
7. byte-compile cleanly -- done
8. override variables I don't own -- I no longer do this, this is in the hands of the user through the readme.

Okay, let me know what I should do that I haven't captured in the "not done" list above!

riscy:
Looks like this is progressing well, and thanks for all the work.  I'll just roll through with a quick pass:
### org-gtd.el
```
org-gtd.el:59:1:Warning: Unused lexical variable `org-stuck-projects'
org-gtd.el:59:1:Warning: Unused lexical variable `org-stuck-projects'
In org-gtd-show-stuck-projects:
org-gtd.el:61:31:Warning: reference to free variable `org-stuck-projects'
In end of data:
org-gtd.el:235:1:Warning: the following functions are not known to be defined:
f-dirname, f-this-file, f-join, f-file-p, org-brain-add-resource
```

```
Warning (emacs):
org-gtd.el:60: All interactive functions should have documentation
```

No issues! 💯

Other static checks:
- org-gtd.el#L23: Prefer `https` over `http` (if possible)

Feel free to ping when you're ready for another look.

Trevoke:
Alright, I've made a very large amount of adjustments. I ended up removing org-brain entirely and letting the user store the information however they want. The documentation isn't *perfect* but there's a few screenshots.

No byte compile errors, checkdoc is clean, package-lint is clean.

What's left? What else do I need?

alphapapa:
Here's one more idea for you to consider: I see you use `read-multiple-choice` for processing items.  That's fine, of course, and it comes with Emacs.  But you might consider using Hydra or Transient instead, as their UIs are generally prettier and more powerful.  `pretty-hydra` produces nice-looking choice UIs automatically with very little code, e.g. https://github.com/jerrypnz/major-mode-hydra.el#pretty-hydra-define

Trevoke:
Using transient (and/or learning how to use hydra) is both on this project's roadmap and on my personal growth roadmap. I don't think either are necessary for a 0.2 release, though - I'll likely get more value out of getting some folks using the package.

The only way I can think of to check whether org-edna has been loaded is to check whether [its implementation](http://bzr.savannah.gnu.org/lh/org-edna-el/view/head:/org-edna.el#L691) has changed the behavior of emacs. That seems very dirty. This is in part why I originally just called it myself. I'm open to recommendations on this topic.

alphapapa:
> The only way I can think of to check whether org-edna has been loaded is to check whether [its implementation](http://bzr.savannah.gnu.org/lh/org-edna-el/view/head:/org-edna.el#L691) has changed the behavior of emacs. That seems very dirty. This is in part why I originally just called it myself.

I think there's nothing wrong with checking whether org-edna's functions are added to the hooks it adds to when it's activated.  Ideally, org-edna would provide a global minor mode to do those, and then you could check the value of a variable.  But since it doesn't, checking those hooks should be fine.

Trevoke:
I sent a message to the author of org-edna suggesting the global minor mode and asking him what he thinks.

I'm uncomfortable with your suggestion, which seems like a direct violation of encapsulation principles.

alphapapa:
> I'm uncomfortable with your suggestion, which seems like a direct violation of encapsulation principles.

Welcome to the world of Emacs packages, where everything runs in a Lisp image, and there are no actual boundaries.

If the author wants to do a minor mode, great.  In the meantime, the existing org-edna code shows that those functions will be present in those hooks when org-edna is activated.  No harm is done by `(memq 'fn hook)`.  A variable would seem cleaner, but as long as the hooks are set like that, it would just be a proxy for whether the functions are in the hooks, anyway.  It doesn't make any difference in the end.

riscy:
Just quickly checking on the compilation status:

```
org-gtd.el:36:1:Error: Cannot open load file: No such file or directory, org-brain
```

Prefer cl-lib's `cl-destructuring-bind` to `destructuring-bind` (deprecated).

Trevoke:
Fixed and fixed.

riscy:
Alright, now I'm catching up. :)

I notice your README still mentions org-brain.

>> org-gtd needs org-edna to be loaded and using property inheritance, so you’ll need to add this to your config file:

> If that's actually needed, you should probably test that and warn the user if it's not the case. Should only be a couple of lines of code.

I'm concerned by this too.  It's hiding at the bottom of your README (which MELPA doesn't distribute), and isn't even present in your elisp file's commentary.  How would someone know they need to do that?  What happens if they forget to?  Will they know how to solve/diagnose the problem if they run into it?

It's unfortunate the author of org-edna didn't implement a minor mode, but if it is strictly needed then I think it is better to make the code violate encapsulation rather than force your users to, if you can see what I mean.  What about something like this?
```elisp
(defun org-gtd--validate-edna-state ()
    (if (or (not org-edna-use-inheritance)
(add-hook 'org-trigger-hook 'org-edna-trigger-function)(defun org-gtd--validate-edna-state ()
  "Check that edna has been correctly loaded.
Future: if edna implements a minor mode, this function will simplify."
    (if (or (not org-edna-use-inheritance)
            (not (memq 'org-trigger-hook #'org-edna-trigger-function))
            (not (memq 'org-blocker-hook #'org-edna-blocker-function)))
        (when (y-or-n-p "org-edna needs to be loaded.  Load?")
          (setq ...))
      (message "Please see README about this issue at http://...")))
```

Trevoke:
Would it be acceptable for _my code_ to call `(org-edna-load)` when one of my functions is called? 

As in, instead of checking _what it does_, why don't I just check _if it exists_ and then call it myself?

riscy:
Yeah, I think that's a better solution.  Thanks :)

riscy:
Ping me if that's the solution you decide on, and I think then this looks good to go.

Trevoke:
The author of org-edna implemented the minor mode - I'll use this to check whether org-edna is enabled, and probably throw an error if it isn't turned on. I'll reach out when that's done.

Trevoke:
I've cleaned up the readme and added the core content of the readme to the elisp package header.

I also have a function for "showing stuck projects", which means that while org-edna is still marked as a dependency, if it doesn't get enabled, it's not the end of the world.

It should also be much more clear to the user what the configuration is and why it's there.

In the end this still seemed cleaner than adding a check in my code, particularly because of the difficulty of doing any kind of intelligent wrapping around the `org-agenda` functions.

riscy:
Thanks!  I think after a couple quick fixes this should be ready to go:

### org-gtd.el ###

```
org-gtd.el:495:1:Warning: the following functions are not known to be defined:
    org-element-property, string-join, org-element-map
```
You will probably only see these when compiling in batch mode, rather than a running Emacs instance, but you can fix them with `(require 'org-element)` and `(require 'subr-x)`

- No issues!

```
1 issue found:
123:53: error: This key sequence is reserved (see Key Binding Conventions in the Emacs Lisp manual)
```
Can this one be addressed?  What about using `C-c C-c`?

doolio:
> but you can fix them with `(require 'org-element)` and `(require 'subr-x)`

@Trevoke I've included this as part of PR [#15](https://github.com/Trevoke/org-gtd.el/pull/15).

Trevoke:
@riscy This is a tricky one. I would like `C-c C-c` but that caused issues: https://github.com/Trevoke/org-gtd.el/issues/5

Honestly, any key binding that is a reasonable "confirmation" key binding would work for me. Unfortunately org-mode used quite a lot of them already. I could use `C-c C-<return>` but that might cause the user to move their hands more than they would want.

Got any suggestions?

riscy:
> Got any suggestions?

You have [some leeway](https://www.gnu.org/software/emacs/manual/html_node/elisp/Key-Binding-Conventions.html) --

> Sequences consisting of C-c followed by any other ASCII punctuation or symbol character are allocated for minor modes. Using them in a major mode is not absolutely prohibited, but if you do that, the major mode binding may be shadowed from time to time by minor modes. 

So you could try out <kbd>C-c '</kbd> or whatever you like.  It's only the letters you have to watch out for.

Trevoke:
I meeaaaaaaaaaaan this is a keymap for a minor mode.

Trevoke:
Re: `C-c '` -- this would be easier if I weren't doing this on an org-mode buffer, because org-mode has taken almost all of these keybindings.

I guess I could try to worm my way out of this and just make it a configurable option, but I don't know how well that would work for a keymap for a minor mode.

Trevoke:
Oh never mind, I understand what your message meant.  I needed to also read:

> Don't define `C-c` _letter_ as a key in Lisp programs. Sequences consisting of `C-c` and a letter (either upper or lower case) are reserved for users; they are the only sequences reserved for users, so do not block them.

riscy:
I was being terse. :)  Ping when this is ready for another look.

Trevoke:
@riscy okay, I've changed it up so that assigning the keybinding is the user's decision, and I've adjusted the README to indicate that. I'm ready for another look.

Trevoke:
@riscy ping ?

riscy:
Hi, apologies I let this slip.  I notice above @doolio has addressed the org-element nag I brought up in [this unmerged PR](https://github.com/Trevoke/org-gtd.el/pull/15), which is still throwing a warning -

```
org-gtd.el:485:1:Warning: the following functions are not known to be defined:
    org-element-property, org-element-map
```

There's also this checkdoc which is extremely minor and I'd consider non-blocking, but fyi:
```
Warning (emacs): 
org-gtd.el:229: Argument `GOTO' should appear (as GOTO) in the doc string
```
I'll also note that it's unusual to upcase variables in elisp (probably you'd want to name these variables `goto` and `keys`):

Trevoke:
Ah-ha, that makes sense.

1. Added the require statements
2. added GOTO and KEYS -- those aren't my variable names, they are `org-capture`'s, so I am using them and naming them the way that function does.

riscy:
Nice, thanks!  This looks good to go.

Trevoke:
Oh sweet! Thank you.

Now I just need to make sure I have the right build so that the quality doesn't decrease.

PR #6623, opened 2020-01-02
Add recipe for org-ml (formerly om.el)

ndwarshuis:

A functional API for org-mode in the vein of dash.el

https://github.com/ndwarshuis/om.el

Maintainer/developer

None

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/om
+++ b/recipes/om
@@ -1,1 +1,0 @@
-(om :repo "ndwarshuis/om.el" :fetcher github)
--- a/recipes/om
+++ b/recipes/om
@@ -1,0 +1,1 @@
+(om :repo "ndwarshuis/om.el" :fetcher github)
--- a/recipes/org-ml
+++ b/recipes/org-ml
@@ -1,0 +1,1 @@
+(org-ml :repo "ndwarshuis/org-ml" :fetcher github)


alphapapa:
This looks very interesting.  Have you seen `org-dp`?  it seems like a similar idea, but your package looks like it may be more comprehensive and "ergonomic" (I also like using Dash).

Your readme is quite comprehensive!  How long have you been working on this?  It's surprising to see it pop up out of nowhere in such an advanced state.  

I haven't even looked at the code yet, but I can tell that I'm going to want to spend some time in it.  I'll just make a quick suggestion for now: 

The package name should probably start with `org-`.  I think this is important, because if your package has as much potential as I think it does, the name should make clear that it's Org-related by following the convention.  I know that Org has some built-in exceptions to that rule, like `ox-`, and even `ol-` now, but 1) those are prefixes used by many filenames and packages, some of which are not in Org itself, and 2) I'm not sure it's really a good idea in the first place.  Anyway, I can't figure out what the M in `om` is for, so it has no mnemonic value to me.  :(  And if every Org related package eschewed the `org-` prefix, the MELPA library would be much harder to dig through to find Org-related packages.  So please consider renaming it.

As for naming suggestions, well, that's one of the 3 hard problems.  Some quick ideas are `org-fp` (for functional programming) and `org-fa` (for functional API), but we can probably come up with some other good possibilities.

ndwarshuis:
Thank you for the feedback and interest :) I've been working on it a few months after discovering that I was writing the same functions over and over again in my config. I'm aware of `org-dp` but like you said, I wanted something easily usable with dash, and it kinda snowballed into this really big thing.

To answer some of the questions and maybe provide more context when you read through:
1. naming: I was going for something like `s.el` or `f.el` which are packages that fall into the same category as this (just a library of functions). So 'om' just stands for 'org-mode' just like 's' stands for 'string' and 'f' stands for 'file'. Granted, I realize this is an org package coming into an established ecosystem already so I understand the need for an 'org-' prefix. Do you still think this is warranted here?
2. org-element bugs and contact: Have not contacted Nicolas yet but that's a good suggestion. I found quite a few inconsistencies when making this that I'm planning to fix/PR soon (mostly around `org-interpret-data` which you noted). For now the workarounds that I have a pretty important to make this usable as is (for example, converting any block-like thing to a drawer will have 'nil' in the middle if it's empty because of the way `format` is used). 
3. crazy macros: this is actually an area where I would appreciate more feedback (there's actually three areas):
   1. better cl-defun: I like the idea of adding it somewhere else. In the meantime, what alternative would you suggest for this? The reason why I like the keywords is because it makes it clear which properties are applicable to which type. I've thought about using only one builder function that just takes a type, plist, and children (ternary), but I think this would be more confusing. Alternatively, I could just take the child arguments as a single list and just use `cl-defun`
   2. anaphoric forms: some of the macros are to create a function and its anaphoric form simultaneously (which I do ALOT). The added benefit is all the docstrings and signatures are standardized. I used to use a helper function to make an anaphoric function from a given function symbol. This is more verbose but I can switch back to that if it makes the code clearer.
   3. type system: This is probably overkill but all the public functions basically have their own type checking system. I once did this with a bunch of `(unless (type fails) (throw error)` things but I converted those to macros since it's basically boilerplate. Most of the macros are there to parse 'typed' signatures (with keys like :p-int for positive integer). To make this even more confusing, not all typing is handled this way; some type errors are thrown deep within a function call, intercepted with a `condition-case` and rethrown (this is to make the error output simpler). If this isn't present, the error messages are extremely misleading (like I could get a 'can't add string to int' error when I actually just passed the wrong node type). Maybe I'm just wishing lisp were more Haskell-like :) but I'll admit taking this out would greatly reduce the macro complexity and `eval-when-compile` calls. That being said, I think this is still a problem but this may just not be the right solution. I think dash and s can get away with not worrying about this because they really only work with one data type (lists and strings), whereas this package deals with lists and plists with different possible types mixed together in a highly structured fashion.

If I had to choose one of those macro areas above it would be ii, since that can be done with one additional macro and saves a ton of space. The other two are debatable as to whether they are needed.

One idea I'm thinking about adding soon is a 'help-function' which has org-element rules readily available. One of the reasons for the added complexity is to steer the user in the right direction (both with my mutant cl-defun and the type system) but could this be a better solution for that?

Thanks for the Cask alternative, I'll check it out ;)

alphapapa:
> Thank you for the feedback and interest :) I've been working on it a few months after discovering that I was writing the same functions over and over again in my config. I'm aware of `org-dp` but like you said, I wanted something easily usable with dash, and it kinda snowballed into this really big thing.
> 
> To answer some of the questions and maybe provide more context when you read through:
> 
> 1. naming: I was going for something like `s.el` or `f.el` which are packages that fall into the same category as this (just a library of functions). So 'om' just stands for 'org-mode' just like 's' stands for 'string' and 'f' stands for 'file'. Granted, I realize this is an org package coming into an established ecosystem already so I understand the need for an 'org-' prefix. Do you still think this is warranted here?

I should have realized that `om` was short for `org-mode`.  :)  I'd generally recommend using the `org-` prefix, because the consistency is helpful for recognizing functions that are Org-related.  At the same time, I empathize with wanting to use shorter prefixes.  What do you think of `org-ea` for "Org element API"?  But it's your project, so of course it's up to you.

> 2. org-element bugs and contact: Have not contacted Nicolas yet but that's a good suggestion. I found quite a few inconsistencies when making this that I'm planning to fix/PR soon (mostly around `org-interpret-data` which you noted). For now the workarounds that I have a pretty important to make this usable as is (for example, converting any block-like thing to a drawer will have 'nil' in the middle if it's empty because of the way `format` is used).

Cool.  If you don't mind, please post to the org-mode mailing list so I can follow the discussion, as I'm also very interested in Nicolas's feedback.

Do you think it would be feasible to work toward putting this library in Org itself, assuming that Nicolas would be agreeable to it?  I think it would be great if it were part of Org itself going forward, to help ensure continued compatibility with changes in Org (sort of like the difference between in-tree and out-of-tree Linux drivers).

> 3. crazy macros: this is actually an area where I would appreciate more feedback (there's actually three areas):
>    
>    1. better cl-defun: I like the idea of adding it somewhere else. In the meantime, what alternative would you suggest for this? The reason why I like the keywords is because it makes it clear which properties are applicable to which type. I've thought about using only one builder function that just takes a type, plist, and children (ternary), but I think this would be more confusing. Alternatively, I could just take the child arguments as a single list and just use `cl-defun`

I don't think I can give detailed, specific feedback without carefully studying all of the function-defining macros, which I confess I'm not eager to do at the moment.  ;)  In general, I'd recommend either:

a.  Contributing changes upstream to Emacs, if appropriate.  I doubt they'd be interested in another `defun` variation, but if the argument handling of `defun` or `cl-defun` could be enhanced without breaking backward compatibility, I'd be happy to see that.  (I can't recall if the Common Lisp `defun` allows arguments similar to your enhancements; if it does, it would seem reasonable to enhance `cl-defun` to match.)
b.  Contributing changes to Dash.  For example, see this Dash issue I filed: https://github.com/magnars/dash.el/issues/268

If it's unsuitable or infeasible to do either of those, I'd generally suggest not using it in your package, because it raises the bar for contributors who need to understand your code.  But it's ultimately your call, of course.

>    2. anaphoric forms: some of the macros are to create a function and its anaphoric form simultaneously (which I do ALOT). The added benefit is all the docstrings and signatures are standardized. I used to use a helper function to make an anaphoric function from a given function symbol. This is more verbose but I can switch back to that if it makes the code clearer.

I'm certainly in favor of DRYer code.  :)  I tend to use macros in my own packages more than most people do, mainly for that purpose.  And I'm also a fan of anaphoric macros, in general.

In this case, I'd suggest that you consider refactoring and reorganizing your code, which could probably make it easier to understand.  For example, they're currently all in one big file, which is intimidating and a bit bewildering.  :)  Instead, you could spread the code out into a few separate files, organized by purpose, each with a short commentary and example, which would help users understand the code and find the code they're looking for.

>    3. type system: This is probably overkill but all the public functions basically have their own type checking system. I once did this with a bunch of `(unless (type fails) (throw error)` things but I converted those to macros since it's basically boilerplate. Most of the macros are there to parse 'typed' signatures (with keys like :p-int for positive integer). To make this even more confusing, not all typing is handled this way; some type errors are thrown deep within a function call, intercepted with a `condition-case` and rethrown (this is to make the error output simpler). Without such a system, the error messages are extremely misleading (like I could get a 'can't add string to int' error when I actually just passed the wrong node type). Maybe I'm just wishing lisp were more Haskell-like :) but I'll admit taking this out would greatly reduce the macro complexity and `eval-when-compile` calls. That being said, I think this is still a problem but this may just not be the right solution. I think dash and s can get away with not worrying about this because they really only work with one data type (lists and strings), whereas this package deals with lists and plists with different possible types mixed together in a highly structured fashion.

Hm, that's interesting.  :)  Here are a few thoughts:

1.  I'm thinking that your library could end up being widely used to programmatically modify Org files (e.g. imagine if the Org Agenda commands that modify entries in their source buffers were implemented with your library--maybe I could use it in `org-ql-view` to do that!).  In that case, performance could be a concern in some use cases, and all that type checking would reduce performance to some extent.
2.  For that reason, it might be preferable to omit most of the type checking at runtime and do as much as you can in the library's tests to ensure that its own code works properly.  Then, if you add Edebug support to your macros and functions, users could more easily trace their own mistakes while using it.
3.  Of course, reducing the code's complexity would help users and potential contributors understand it.  Speaking for myself, as the code currently stands, I feel like I would have to invest a lot of time to understand how it works before I could contribute anything, but I'm sure there are better programmers who could learn it more quickly.

I'm not sure how concerned you are with internal type-related bugs vs. mistakes by users, but it's something to think about.  The old axiom about premature optimization might apply to some extent as well, but only you know the code well enough to say.  :)

> One idea I'm thinking about adding soon is a 'help-function' which has org-element rules readily available. One of the reasons for the added complexity is to steer the user in the right direction (both with my mutant cl-defun and the type system) but could this be a better solution for that?

I'm not sure what you mean here.  Generally I would favor concise, well-organized documentation and plentiful, varied examples.  Seems like you have a good start on that.  I'd suggest adding more examples outside of the function/macro references, which should show *application* of the library as a whole.

ndwarshuis:
> I should have realized that om was short for org-mode. :) I'd generally recommend using the org- prefix, because the consistency is helpful for recognizing functions that are Org-related. At the same time, I empathize with wanting to use shorter prefixes. What do you think of org-ea for "Org element API"? But it's your project, so of course it's up to you.

Makes sense. In addition to `org-ea` and `org-fp` I also like `org-ml` (org meta-language, which I guess coincidentally om could also stand for :) ).

> Cool. If you don't mind, please post to the org-mode mailing list so I can follow the discussion, as I'm also very interested in Nicolas's feedback.

> Do you think it would be feasible to work toward putting this library in Org itself, assuming that Nicolas would be agreeable to it? I think it would be great if it were part of Org itself going forward, to help ensure continued compatibility with changes in Org (sort of like the difference between in-tree and out-of-tree Linux drivers).

Yes, I can work on starting that discussion, and I'm not opposed at all to getting this in org-mode itself for the reasons you mentioned.

> I don't think I can give detailed, specific feedback without carefully studying all of the function-defining macros, which I confess I'm not eager to do at the moment. ;) In general, I'd recommend either:

> a. Contributing changes upstream to Emacs, if appropriate. I doubt they'd be interested in another defun variation, but if the argument handling of defun or cl-defun could be enhanced without breaking backward compatibility, I'd be happy to see that. (I can't recall if the Common Lisp defun allows arguments similar to your enhancements; if it does, it would seem reasonable to enhance cl-defun to match.)
b. Contributing changes to Dash. For example, see this Dash issue I filed: magnars/dash.el#268

> If it's unsuitable or infeasible to do either of those, I'd generally suggest not using it in your package, because it raises the bar for contributors who need to understand your code. But it's ultimately your call, of course.

I don't blame you :)

I actually don't think its a good idea to add to `cl-defun` since the goal of the cl libraries in emacs is to mimic common lisp as much as possible. The behavior of `defun` in CL is to bind to the plist of key-val pairs to the symbol given after &rest, and then bind the values of each key to their own symbol. This specification makes it impossible to do what I'm doing which is to to accept a variable number of arguments and intelligently figure out which are keywords and which are "rest" arguments and bind to the appropriate symbol. Therefore I think I'll see if I can add it somewhere else.

> I'm certainly in favor of DRYer code. :) I tend to use macros in my own packages more than most people do, mainly for that purpose. And I'm also a fan of anaphoric macros, in general.

> In this case, I'd suggest that you consider refactoring and reorganizing your code, which could probably make it easier to understand. For example, they're currently all in one big file, which is intimidating and a bit bewildering. :) Instead, you could spread the code out into a few separate files, organized by purpose, each with a short commentary and example, which would help users understand the code and find the code they're looking for.

Fine idea

> Hm, that's interesting. :) Here are a few thoughts:

>    I'm thinking that your library could end up being widely used to programmatically modify Org files (e.g. imagine if the Org Agenda commands that modify entries in their source buffers were implemented with your library--maybe I could use it in org-ql-view to do that!). In that case, performance could be a concern in some use cases, and all that type checking would reduce performance to some extent.

True

>    For that reason, it might be preferable to omit most of the type checking at runtime and do as much as you can in the library's tests to ensure that its own code works properly. Then, if you add Edebug support to your macros and functions, users could more easily trace their own mistakes while using it.
>    Of course, reducing the code's complexity would help users and potential contributors understand it. Speaking for myself, as the code currently stands, I feel like I would have to invest a lot of time to understand how it works before I could contribute anything, but I'm sure there are better programmers who could learn it more quickly.

> I'm not sure how concerned you are with internal type-related bugs vs. mistakes by users, but it's something to think about. The old axiom about premature optimization might apply to some extent as well, but only you know the code well enough to say. :)

The "type system" is there to catch the end-user's mistakes. All the internal components should be validated using the unit test suite. In that sense there is still a performance hit but only for the arguments that the user supplies, since it will be assumed that everything works internally.

Granted, I've never really used edebug so I'm not familiar with that method of debugging. I tried adding it to get `undercover.el` to work and it was a pain and ultimately failed, probably because the macros are so complicated :( I can try adding debug support on a new branch without the type system and see how it behaves.

> I'm not sure what you mean here. Generally I would favor concise, well-organized documentation and plentiful, varied examples. Seems like you have a good start on that. I'd suggest adding more examples outside of the function/macro references, which should show application of the library as a whole.

I meant to make a function that spat out documentation for a given node and its properties (among other things), but then I remembered that I can make an info node to do the same thing.

And yes, would like to add more comprehensive examples but I think at this point I should start a github wiki since the README is already extremely long...just haven't gotten that far yet.

> P.S. I wonder if there's any potential to integrate ts.el support for easily modifying Org timestamps...

Ah beautiful :) I wish I knew about this before I started

alphapapa:
> Granted, I've never really used edebug so I'm not familiar with that method of debugging. I tried adding it to get `undercover.el` to work and it was a pain and ultimately failed, probably because the macros are so complicated :( I can try adding debug support on a new branch without the type system and see how it behaves.

Edebug is a very helpful tool.  The basics of stepping through forms is pretty easy: `edebug-defun` (e.g. `C-u C-M-x` on a `defun`), then run your code, and then `SPC` or `n` when the debugger starts.  `o` is also helpful for getting out of nested forms once you know what you need to know.  The rest is documented in the manual, and there are a few tutorials and blog articles that help introduce it as well.

For macros that define functions, it's necessary to tell Edebug what the macro's arguments are.  Specifications for simple macros are simple, but they can get complicated for powerful ones.  Here's an example in one of my macros: https://github.com/alphapapa/org-super-agenda/blob/1ac7f86ff0583014eaf98eb3dccaab0ff55dcd48/org-super-agenda.el#L355  The syntax is thoroughly documented in the manual section about the `declare` form, but I always have to use some trial-and-error to get complex ones working.  There aren't enough examples of complex uses, but maybe that one will help (there are other examples in that package as well).

Anyway, not everyone uses Edebug, but it's like the Elisp equivalent of CL debugging, so it's worth learning the basics (and probably `x`/`X` as well, which saves a lot of time).

> I meant to make a function that spat out documentation for a given node and its properties (among other things), but then I remembered that I can make an info node to do the same thing.

Yes, and you can even generate Texinfo manuals from Org files.  I do that for `org-super-agenda`, and I learned by copying the Magit manual, which does so in a very sophisticated way.  The Org manual is also generated in a similar way.

> Ah beautiful :) I wish I knew about this before I started

I'm beginning to think that "discovery of existing tools" should be known as the Fourth Hard Problem.  :)  If you haven't seen it, I'm trying to catalogue Elisp tools in a useful way here: https://github.com/alphapapa/emacs-package-dev-handbook  Feel free to make any suggestions!

ndwarshuis:
A few updates:
- macros: I took out the type system. Your intuition about edebug was correct; it is way simpler and intuitive than what I had built in. The better `cl-defun` and anaphoric form macros are still there but I moved them to a separate file. The former will hopefully be part of dash but for now its in this package (issues [here](https://github.com/magnars/dash.el/issues/268) and [here](https://github.com/magnars/dash.el/issues/329) for anyone else wanting to follow this)
- naming: I would like to keep `om.el` as opposed to other names starting with `org-`. It fits the naming convention of other 'library' packages like `f.el` and `s.el` and it has a nice contemplative ring to it :) Also, if the concern is that people won't find the package because it doesn't start with `org-`, anyone searching for 'org', 'org-mode', or 'org functional' in MELPA will come across this, and it will appear above the search results for `org-` packages because alphabetical ordering.
- documentation/info files: still working on it, there were some last minute functions I wanted to add for use in my other packages but that is hopefully settling down now
- including in org itself: I started a thread [here](https://lists.gnu.org/archive/html/emacs-orgmode/2020-01/msg00187.html). Not sure how much attention it will get since the org mailing list has lots of threads.

Have you had a chance to read the code further?

alphapapa:
> * macros: I took out the type system. Your intuition about edebug was correct; it is way simpler and intuitive than what I had built in. The better `cl-defun` and anaphoric form macros are still there but I moved them to a separate file. The former will hopefully be part of dash but for now its in this package (issues [here](https://github.com/magnars/dash.el/issues/268) and [here](https://github.com/magnars/dash.el/issues/329) for anyone else wanting to follow this)

Cool, I'll follow those issues.

> * naming: I would like to keep `om.el` as opposed to other names starting with `org-`. It fits the naming convention of other 'library' packages like `f.el` and `s.el` and it has a nice contemplative ring to it :) Also, if the concern is that people won't find the package because it doesn't start with `org-`, anyone searching for 'org', 'org-mode', or 'org functional' in MELPA will come across this, and it will appear above the search results for `org-` packages because alphabetical ordering.

I've gotten so used to Org-related code starting with `org-` that I feel like something's wrong when that's not the case.  ;)  Plus I like that it's easy to filter package names on `^org-` and get all the Org-related ones, similarly to how most Magit-related ones start with `^magit`.  I think the biggest benefit, though, is that it makes packages more discoverable.  There are so many packages available now that discoverability is a real issue.  We keep seeing people submit packages and post code that does things for which packages already exist, but they didn't know because they couldn't find them.

So, I still encourage you to rename it using a `org-` prefix.  But it's up to you.  I've done my best to convince you.  :)

> * including in org itself: I started a thread [here](https://lists.gnu.org/archive/html/emacs-orgmode/2020-01/msg00187.html). Not sure how much attention it will get since the org mailing list has lots of threads.

Yeah, I saw that.  I hope Nicolas notices, because he's the main author of org-element, but of course, he is busy with things as well.  He probably wouldn't mind if you sent him a personal mail drawing his attention to that thread.  You can blame it on me if you want.  :)

> Have you had a chance to read the code further?

I've been busy with other projects, so, no, not very much.  I looked at a few more examples yesterday, but it was hard for me to wrap my head around the API in some cases. 

I do remember one specific thing: in the querying functions, the way `:many` works confused me.  It might be the terminology that gets me.  More explanation in that area might help.

ndwarshuis:
> I've gotten so used to Org-related code starting with org- that I feel like something's wrong when that's not the case. ;) Plus I like that it's easy to filter package names on ^org- and get all the Org-related ones, similarly to how most Magit-related ones start with ^magit. I think the biggest benefit, though, is that it makes packages more discoverable.

So actually to this point, I just tried googling my own packages and couldn't find om.el (I could find my other `org-whatever` packages despite them probably getting less traffic). So in that case I will change it (not to sound like a total flip-flopper). Should I open a separate PR here with the new name?

> I've been busy with other projects, so, no, not very much. I looked at a few more examples yesterday, but it was hard for me to wrap my head around the API in some cases.

No worries, makes sense.

> Yeah, I saw that. I hope Nicolas notices, because he's the main author of org-element, but of course, he is busy with things as well. He probably wouldn't mind if you sent him a personal mail drawing his attention to that thread. You can blame it on me if you want. :)

Yup will do

> I do remember one specific thing: in the querying functions, the way :many works confused me. It might be the terminology that gets me. More explanation in that area might help.

`:many` is a wildcard pattern that matches zero or more levels in the parse tree. So if you wanted to match all timestamps under a headline (including those in all subheadlines) the pattern would just be `(:many timestamp)`. So all timestamps would be matched if this entire subtree were fed to a match function:

```org
* one
[2020-01-01 Wed]
** two
[2020-01-01 Wed]
*** three
[2020-01-01 Wed]
**** four
[2020-01-01 Wed]
```

Would it make more sense if I used some sort of regexp-like syntax? I can't really use a period because that's special in lisp, but I already have `:any` which matches any type at exactly one level. Then `:any *` could be the same as `:many` (match zero or more levels of any type). This also makes the syntax more modular so I could add other operators (like `:any +` to match one or more levels of any type) as needed.

alphapapa:
> So actually to this point, I just tried googling my own packages and couldn't find om.el (I could find my other `org-whatever` packages despite them probably getting less traffic). So in that case I will change it (not to sound like a total flip-flopper). Should I open a separate PR here with the new name?

Cool, I think it will help your package and users in the long run.  You can just adjust this PR and force-push.

> > I do remember one specific thing: in the querying functions, the way :many works confused me. It might be the terminology that gets me. More explanation in that area might help.
> 
> `:many` is a wildcard pattern that matches zero or more levels in the parse tree. So if you wanted to match all timestamps under a headline (including those in all subheadlines) the pattern would just be `(:many timestamp)`. So all timestamps would be matched if this entire subtree were fed to a match function:
> 
> ```
> * one
> [2020-01-01 Wed]
> ** two
> [2020-01-01 Wed]
> *** three
> [2020-01-01 Wed]
> **** four
> [2020-01-01 Wed]
> ```

That does look useful.  It still feels strange to me to call it `many` or `any`.  It feels to me more like a mapping with selection, i.e. mapping "select all timestamps in this entry" across each entry.  A similarly named function from Dash, `-any`, is a predicate.  I would tend to expect to use it more like (pseudocode example):
```el
(->> entries entry-timestamps -non-nil)
```
Where `entry-timestamps` is a function that returns all timestamps in an entry.  If you haven't already, you could probably use a macro to provide shorter function aliases and maybe implied use of `-non-nil`, e.g. (pseudocode again):
```el
(org-m->> entries timestamps)
```
Just some ideas.  :)  I went through a similar pattern with `org-ql`: when I first started on it, all of the symbols and functions had full prefixes, like `org-agenda-ng--todo` instead of `todo`, which was too awkward to be useful for writing queries.  (I'm not saying that's the case with your code--again, I don't understand it well.)

> Would it make more sense if I used some sort of regexp-like syntax? I can't really use a period because that's special in lisp, but I already have `:any` which matches any type at exactly one level. Then `:any *` could be the same as `:many` (match zero or more levels of any type). This also makes the syntax more modular so I could add other operators (like `:any +` to match one or more levels of any type) as needed.

AFAIK periods can be used, e.g. see `let-alist`.  But other than the suggestions I've already made, I feel like I don't understand the library well enough to offer more.  It may be that the API is unusual, or it may be merely that it feels unusual to me.   :)  Anyway, your library is an exciting development for Org that I expect will be the foundation for many interesting tools.

riscy:
Hi @ndwarshuis -- sorry for the long wait on this.  You can definitely update this PR with the name you chose.  In the meantime, here's a quick first pass over the checks (nothing looks major or really like a blocker):

### om-macs.el
- No issues! 💯

- No issues! 💯

- om-macs.el#L16: Prefer `https` over `http` (if possible)
### om.el
- No issues! 💯

```
Warning (emacs):
om.el:3234: Disambiguate org-log-into-drawer by preceding w/ function,command,variable,option or symbol.
Warning (emacs):
om.el:3253: Disambiguate org-log-into-drawer by preceding w/ function,command,variable,option or symbol.
Warning (emacs):
om.el:4783: All variables and subroutines might as well have a documentation string
Warning (emacs):
om.el:4809: All variables and subroutines might as well have a documentation string
```

- No issues! 💯

- om.el#L22: Prefer `https` over `http` (if possible)

tarsius:
Friendly ping!

tarsius:
Despite the lack of progress I am not *yet* closing this because of @alphapapa's positive comments.

ndwarshuis:
@tarsius sorry for the delay, thank you for keeping it open

@alphapapa Given Nicolas's [feedback](https://lists.gnu.org/archive/html/emacs-orgmode/2020-02/msg00021.html) I won't be able to include this in org-mode itself unless I totally take out the dash.el and s.el dependencies. A few points on this:
1) This isn't impossible but it would be alot of work and it would probably make the code harder to maintain since it is written in a functional style at its core.
2) Even if dash.el isn't in the code itself I would also still want the examples and tests to use dash since that was the intended use-case and style for this package to begin with.
3) It would also mean that I would be reinventing any function definition macros that eventually get submitted to dash (like the 'better cl-defun' discussed including in dash.el earlier, which I still need to submit).

Feel free to agree/disagree with any of these points, but I think at the very least the next decision to make is where this package should go since the work that needs be done is quite different for incorporating it into org-mode vs submitting here. In my mind the decision comes down to the advantages of relying on third-party (but well-established) packages such as dash and the potential gains in compatibility by being part of org (like an in-tree kernel driver as you mentioned before).

riscy:
> So in that case I will change it (not to sound like a total flip-flopper). Should I open a separate PR here with the new name?

Is there still any intention to change the name?  (I wouldn't consider it a blocker, but I would say the `org-` prefix would be an improvement in my hobgoblin mind)

conao3:
[ndwarshuis/om.el - Upcoming Breaking Changes for 2.0](https://github.com/ndwarshuis/om.el#upcoming-breaking-changes-for-20)
> This packages will be renamed org-ml (org-metalanguage)

It seems the package will be renamed (om.el -> org-ml.el). But the recipe change and actual rename are not executed.

ndwarshuis:
To confirm, yes it is going to be renamed to org-ml. But there are some other changes listed that I want to finish before bumping to version 2.

riscy:
Of course!  Feel free to ping when ready.

vv111y:
Hi @ndwarshuis, how far are you with version 2 and acceptance here? I want to use your package right away.

Thank you for making it, this is really useful and a big time-saver

ndwarshuis:
@vv111y version 2 will come within a few days if all goes according to plan. You can follow the latest commits on the develop branch

ndwarshuis:
name change complete

riscy:
Great!  Here's the latest since the last time:

### org-ml-macs.el ###

- No issues!

- No issues!

package-lint (using version 20200616.2257):
- No issues!

- org-ml-macs.el#L85: Don't quote lambdas; this prevents them from being compiled

### org-ml.el ###

```
org-ml.el:127:29:Error: Symbol's value as variable is void: org-ml-elements
```

```
Warning (emacs): 
org-ml.el:3298: Disambiguate org-log-into-drawer by preceding w/ function,command,variable,option or symbol.
Warning (emacs): 
org-ml.el:3317: Disambiguate org-log-into-drawer by preceding w/ function,command,variable,option or symbol.
Warning (emacs): 
org-ml.el:5052: All variables and subroutines might as well have a documentation string
Warning (emacs): 
org-ml.el:5077: All variables and subroutines might as well have a documentation string
```

package-lint (using version 20200616.2257):
- No issues!

- org-ml.el#L737: Consider `when ...` instead of `unless (null ...)`

ndwarshuis:
@riscy everything should be fixed in 2.0.1 except the lambda quoting in org-macs.el which is a false positive (that lambda shouldn't be compiled because it is being used as an unevaluated form to build a macro)

PR #6627, opened 2020-01-05
Add recipe for competitive-programming-snippets

sei40kr:

Competitive Programming snippets for [Yasnippet](https://github.com/joaotavora/yasnippet).

https://github.com/sei40kr/competitive-programming-snippets

Maintainer

None needed

### Checklist

### DIFF
--- a/recipes/competitive-programming-snippets
+++ b/recipes/competitive-programming-snippets
@@ -1,0 +1,3 @@
+(competitive-programming-snippets
+ :repo "sei40kr/competitive-programming-snippets"
+ :fetcher github)


purcell:
Thanks. [This](https://github.com/sei40kr/competitive-programming-snippets/blob/6146108c3babdae5a57831354098e4d7cb1a9372/competitive-programming-snippets.el#L51) will cause double-initialisation if people use your autoloaded setup function, so I think you should remove that line. Looks good otherwise.

sei40kr:
@purcell 
I removed ```;;;###autoload``` pragma from initialization.

purcell:
> I removed `;;;###autoload` pragma from initialization.

You didn't, as far as I can see - it's still there. But that's fine. I think [this](https://github.com/sei40kr/competitive-programming-snippets/blob/6146108c3babdae5a57831354098e4d7cb1a9372/competitive-programming-snippets.el#L51) is the line you should delete. Better to have an explicit setup function than to have loading the library cause side effects.

mem-YSeong:
@purcell 
Sorry I had failed to push.
I updated to remove the initialization on load. Please review.
Thanks for your advice.

sei40kr:
@purcell 
Sorry I failed to push the update.
I re-updated the package to remove the initialization on load. Please review.
Thanks for your advice!

https://github.com/sei40kr/competitive-programming-snippets/commit/29e9c0a2e2332caded31190b97f388ffa3a9218b

PR #6628, opened 2020-01-07
Add Verb package.

federicotdn:

Verb is a new HTTP client for emacs, based on Outline mode. It allows users to organize and send HTTP requests from Emacs.

https://github.com/federicotdn/verb

Maintainer

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/verb
+++ b/recipes/verb
@@ -1,0 +1,1 @@
+(verb :repo "federicotdn/verb" :fetcher github)


riscy:
Nice!  None of the following are blockers, but I'll leave them here for your consideration:

- verb.el#L804: Consider `unless ...` instead of `when (not ...)`
- verb.el#L1339: Consider `unless ...` instead of `when (not ...)`
- verb.el#L1347: Consider `unless ...` instead of `when (not ...)`
- verb.el#L1352: Consider `unless ...` instead of `when (not ...)`
- "HTP responses" -> "HTTP responses"
- "nunmber of seconds" -> "number of seconds"
- "delimeters" -> "delimiters"
- "occured" -> "occurred"
- "commments" -> "comments"

federicotdn:
Thanks! Will fix those.

PR #6629, opened 2020-01-09
add org-linkotron recipe

perweij:

The purpose of this package is to provide a way to open a group of org-links at once. This makes it easier to switch between work tasks that each require a number of links opened (in the web browser). 

Upon activation, the program locates all org-links in the current text block, then it calls `org-open-at-point` on each of them, with a configurable pause between each call (to allow Firefox to handle the load, in case they are browser links).

This is my first attempt at creating an Emacs package.

https://gitlab.com/perweij/org-linkotron

I am the maintainer.

**None needed**

### Checklist

### DIFF
--- a/recipes/org-linkotron
+++ b/recipes/org-linkotron
@@ -1,0 +1,1 @@
+(org-linkotron :fetcher gitlab :repo "perweij/org-linkotron")


alphapapa:
Hey, this looks like a useful package!  Here's some random feedback:

* I see you used some of my "scaffolding"!  :)  I'm glad it was useful to you.
  * For the test code, of course you're welcome to borrow from org-super-agenda's test suite, but I've actually come up with a better solution, which I'll be moving org-super-agenda to soon.  It's much simpler and easier to use.  https://github.com/alphapapa/makem.sh
    * Using it, in the test file, you should just `(require 'org-linkotron)` instead of using `load-file` with a path.
  * Here's some generally useful information: https://github.com/alphapapa/emacs-package-dev-handbook#best-practices
* https://gitlab.com/perweij/org-linkotron/blob/master/test/test.el#L4 I see in the commit log, `Revert from cl-lib to cl until finding equivalent of "every".`, but did you miss `cl-every`?  `cl-lib` is intended to be used instead of `cl` in every case.
* You shouldn't need to define your own `back-to-heading` and `forward-to-heading` functions, because there are, e.g. `org-back-to-heading`, `outline-next-heading`, etc.
* Rather than write your own regexp for links, you can use `org-bracket-link-regexp`.
* Prefer `when` instead of "one-armed `if`s".  :)
* Instead of `org-linkotron--find-region` you could use functions like `org-entry-beginning-position`, `org-entry-end-position`, etc.
* A full-second pause between opening links seems excessive and may make your package feel slow.
* The logic for finding and returning links feels a bit weird.  You could probably replace most of the functions with this simple one:
```el
(defun org-linkotron--entry-links ()
  "Return links from current entry."
  (let ((bound (org-entry-end-position)))
    (save-excursion
      (goto-char (org-entry-beginning-position))
      (cl-loop while (re-search-forward org-bracket-link-regexp bound t)
               collect (match-string 1)))))
```

weipe:
Big thanks for reading and providing feedback!

> https://github.com/alphapapa/makem.sh

This is incredibly convenient! Incorporated.

> https://github.com/alphapapa/emacs-package-dev-handbook#best-practices

Loads of good reading: after a couple of pages I have already learnt a lot - not least the handy cl-loop - while - collect pattern.

>  `cl-every`

Fixed! I could probably use a better way to compare the lists too (saw a chapter in the best-practices article).

> `org-back-to-heading`, `outline-next-heading`, etc.
> Rather than write your own regexp for links, you can use `org-bracket-link-regexp`.
> Prefer `when` instead of "one-armed `if`s".  :)
> Instead of `org-linkotron--find-region` you could use functions like `org-entry-beginning-position`, `org-entry-end-position`, etc.
> A full-second pause between opening links seems excessive and may make your package feel slow.

Good points, fixed!

> Or you could just do it all in one like this:
> ...
> That basically reduces the package to a single function and a `defcustom`. :)

Very cool - hope you don't mind if I copy that!

I updated the project (I could possibly trig the PR checks ti re-run here in Melpa).

alphapapa:
> Very cool - hope you don't mind if I copy that!

Please do!  Thanks.

weipe:
Good, thanks!

riscy:
Hi, just a quick pass --

### org-linkotron.el
```
org-linkotron.el:92:1:Error: Cannot open load file: No such file or directory, ol
```
`ol` isn't in your Package-Requires (and I'm not sure where to get it).  If I comment that out line out and try to byte-compile I also see this:
```
org-linkotron.el:110:41:Warning: reference to free variable
`org-link-bracket-re'
```
Is this related to `org-bracket-link-regexp`?  You require org 9.0, but as of org 9.1.9, that variable doesn't appear to be a part of org.

weipe:
Hi! Thanks for taking your time to check this PR!

> org-linkotron.el:92:1:Error: Cannot open load file: No such file or directory, ol
> org-linkotron.el:110:41:Warning: reference to free variable `org-link-bracket-re'
> Is this related to `org-bracket-link-regexp`? You require org 9.0

Hm, yes, I admit I misread the documentation for that variable, and was also confused by the free-variable-warning. 

I now removed the require 'ol, and bumped the package-require to 9.1 (removing a mismatch with the Cask file too). As far as I can tell, the compiler errors and warnings are now gone.

On a side note, I find it somewhat difficult to test the package with the given version requirements, as `package-install` always seems to install the latest available version (which for org would be 9.3 I guess?). Would the best practice be to just set the latest available version of all packages in the requirements?

riscy:
> bumped the package-require to 9.1

I don't think 9.1 has the variable you need either, but correct me if I'm wrong.  Is the variable introduced in org 9.3?

> Would the best practice be to just set the latest available version of all packages in the requirements?

org is a tricky one because it's simultaneously part of Emacs and sees very active development (with function/variable names changing and evolving quickly).  The best bet might be to specify a version that is the same version you're developing alongside.

weipe:
> I don't think 9.1 has the variable you need either, but correct me if I'm wrong. Is the variable introduced in org 9.3?

Ah, thanks, I now checked the org-repo's history, and you're right - the variable appeared in 9.3. I have  now set the reqs to 9.3.

> org is a tricky one because it's simultaneously part of Emacs and sees very active development (with function/variable names changing and evolving quickly). The best bet might be to specify a version that is the same version you're developing alongside.

Thanks for the good advice!

tarsius:
Friendly ping!

PR #6630, opened 2020-01-09
Add keycaster recipe

chuntaro:

This package is a keystroke visualizer for GUI version Emacs.
This allows you to display keystrokes without relying on external tools when creating screencasts.
There are screencasts in the repositories below to see how this works.

https://github.com/chuntaro/emacs-keycaster

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/async-await
+++ b/recipes/async-await
@@ -1,1 +1,0 @@
-(async-await :repo "chuntaro/emacs-async-await" :fetcher github)
--- a/recipes/async-await
+++ b/recipes/async-await
@@ -1,0 +1,1 @@
+(async-await :repo "chuntaro/emacs-async-await" :fetcher github)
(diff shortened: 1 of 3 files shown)


alphapapa:
Have you seen https://github.com/tarsius/keycast ?  It would be good if you could contribute to it as an alternate display style.

chuntaro:
Emacs has the ability to capture keystrokes, so most (close to 99%) of the keycaster code is spent displaying.
It should be the same for keycast.
So, since the display method is completely different, I think that there is no problem to say that it is a completely different package.
However, it is not good to have almost the same name, so can I change the package name and PR again?

riscy:
> so can I change the package name and PR again?

You can just update this PR if you like, and update the link to the package repository in your PR description.  It's all good. :)

riscy:
Ping when this is ready for another look.

PR #6634, opened 2020-01-12
Add recipe for `qtcreator-theme' package

LesleyLai:

An Emacs color theme that mimics the default theme of the Qt Creator IDE

https://github.com/LesleyLai/emacs-qtcreator-theme

maintainer

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/qtcreator-theme
+++ b/recipes/qtcreator-theme
@@ -1,0 +1,1 @@
+(qtcreator-theme :fetcher github :repo "lesleylai/emacs-qtcreator-theme")


riscy:
Hi!  Thanks for this.  Everything looks good with the checklist.  It's not a blocker, but I think it may be worth adding a license excerpt to your elisp file as well, just in case that's the only thing someone receives.

You can use the short form of the license at the bottom of [this](https://www.apache.org/licenses/LICENSE-2.0.html) page and add it right above your `;;; Commentary:` and right below your `;; This file is not part of GNU Emacs.` 

I'll hit merge now, and leave that for your consideration.

LesleyLai:
@riscy Thank you, I [add](https://github.com/LesleyLai/emacs-qtcreator-theme/commit/553595a4efbc6e454a5cbfc4b3464c48cc6eb1da) the license.

PR #6635, opened 2020-01-13
Add `keepass-mode` package

ifosch:

`keepass-mode` allows the user to open, read, and query a KeePass DB file.

https://github.com/ifosch/keepass-mode

I'm the maintainer to the package.

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/keepass-mode
+++ b/recipes/keepass-mode
@@ -1,0 +1,1 @@
+(keepass-mode :fetcher github :repo "ifosch/keepass-mode" :branch "melpa")


riscy:
Thanks for this.

Including keepass-mode-pkg.el is discouraged -- MELPA will create a `-pkg.el` file, so this may be worth excluding from your recipe.  More below:

### keepass-mode-autoloads.el
- No issues! 💯

- No issues! 💯

- keepass-mode-autoloads.el#L22: Prefer `https` over `http` (if possible)

### keepass-mode.el
- No issues! 💯

- No issues! 💯

- No issues! 💯

- keepass-mode.el#L49: It's safer to sharp-quote function names; use `#'`
- keepass-mode.el#L57: No `format` required; messages are already f-strings
- keepass-mode.el#L59: No `format` required; messages are already f-strings
- keepass-mode.el#L22: Prefer `https` over `http` (if possible)
- The function `keepass-mode-command` looks like it should be using `shell-quote-argument` for all of the arguments; otherwise if one of the arguments has special characters (|, &, etc.) then the command will fail

### Maintainer/license

- PR by ifosch: https://github.com/ifosch/keepass-mode
- Including keepass-mode-pkg.el is discouraged -- MELPA will create a `-pkg.el` file

### Summary

- (keepass-mode :fetcher github :repo "ifosch/keepass-mode" :branch "melpa")
- Package-Requires: emacs "27"
- _elisp/keepass-mode-autoloads.el (gpl): Manually generated autoloads
- _elisp/keepass-mode-pkg.el (gpl): Mode for KeePass DB
- _elisp/keepass-mode.el (gpl): Mode to open Keepass DB
</details>

ifosch:
@riscy I've implemented all your suggestions and fixed the checks in the `melpa` branch. Let me know if there's anything else. Thank you!

riscy:
Thanks, looks much better.

I've been advised to make sure packages are adding GitHub compatible LICENSE/COPYING files to the repository when possible.  You could use the same LICENSE.md file MELPA uses, for example, and then the GitHub API would pick up that you're GPL3.

I'm not so familiar with what's going on in keepass-mode-autoloads.el -- if we want to autoload keepass-mode, couldn't we just add an autoload cookie to keepass-mode.el and have one fewer file?  In reading the documentation, it seems like the "keepass.mode" argument in this:
```elisp
(autoload 'keepass-mode "keepass.mode" nil t)
```
should be "keepass-mode" (since that's the name of the corresponding file).

I may be missing something, since this is new to me, but maybe you can help clarify?

ifosch:
Now the upstream contains a LICENSE.md file with the GPL 3.0 text, and I've implemented the autoload cookie, removing the `keepass-mode-autoloads.el` file.

tarsius:
> Now the upstream contains a LICENSE.md file with the GPL 3.0 text

It would be better to use a `LICENSE` file which contains a copy of the license that has *not* been markdownified. Tools such as those by Github and the Emacsmirror can detect the latter but not (yet?) the former.

ifosch:
> It would be better to use a `LICENSE` file which contains a copy of the license that has _not_ been markdownified. Tools such as those by Github and the Emacsmirror can detect the latter but not (yet?) the former.

Fine, LICENSE.md removed, and replaced with a LICENSE with plain GPL text.

riscy:
Thanks for your patience with that, @ifosch -- I was not fully aware of the problems with a markdownified license before this.

ifosch:
> Thanks for your patience with that, @ifosch -- I was not fully aware of the problems with a markdownified license before this.

No problem :smile:

riscy:
Regarding your major mode definition:
```elisp
;;;###autoload
(define-derived-mode keepass-mode tabulated-list-mode "KeePass mode"
  "KeePass mode."
  (setq keepass-mode-db buffer-file-truename)
  (setq keepass-mode-password (keepass-mode-ask-password))
  (setq keepass-mode-group-path "")
  (kill-buffer)
  (switch-to-buffer (format "*KeePass %s*" keepass-mode-db))
  (use-local-map keepass-mode-map)
  (make-local-variable 'keepass-mode-db)
  (make-local-variable 'keepass-mode-password)
  (make-local-variable 'keepass-mode-group-path)
  (keepass-mode-open))

;; ...
(add-to-list 'auto-mode-alist '("\\.kdbx\\'" . keepass-mode))
(add-to-list 'auto-mode-alist '("\\.kdb\\'" . keepass-mode))
```
It feels like this is using a major mode as a "backdoor" to the *KeePass XYZ* buffer -- i.e., the major mode kills the buffer it was started in.  I can understand the usage pattern you're going for, though (and I like it).

I feel like killing the buffer and switching to a new one is what is forcing you to manually `(use-local-map keepass-mode-map)`, whereas that would be automatic if the buffer didn't change.  I also wonder if deriving from `tabulated-list-mode` is having an effect, since the buffer we end up in (*KeePass XYZ*") is not the buffer we started with?

I'll stop there and see what you think about that -- maybe there is a way to write a <kbd>M-x keepass</kbd> command that retains a lot of the workflow you're going for.

ifosch:
@riscy Thanks for your points. I just reduced the major mode removing these lines. This was a very early development stage, where in some point I found I needed this, but I just removed, and it works perfectly without the buffer switch and, as you felt, the mode map manual set.

I discard the implementation of such `M-x keepass` command for now, since I think it's better to work with this attached to using a specific file, but I was considering to have some way of creating files in some future, maybe using such a function.

riscy:
Okay, right on.  Then the next (and I think last!) thing I'll point out is that these variables
```elisp
(defvar keepass-mode-db "")
(defvar keepass-mode-password "")
(defvar keepass-mode-group-path "")
```
feel like they want to be using `defvar-local`, instead, especially the password. (Currently the password is asked for every time we open the buffer; then we set the *global* value of keepass-mode-password to that password -- which persists everywhere -- and then we make it local).  So instead we could have:
```elisp
(defvar-local keepass-mode-db "")
(defvar-local keepass-mode-password "")
(defvar-local keepass-mode-group-path "")

;; ..

;;;###autoload
(define-derived-mode keepass-mode tabulated-list-mode "KeePass"
  "KeePass mode for interacting with the KeePass DB.
\\{keepass-mode-map}"
  (setq-local keepass-mode-db buffer-file-truename)
  (setq-local keepass-mode-password (keepass-mode-ask-password))
  (setq-local keepass-mode-group-path "")
  (keepass-mode-open))
```
Ping me when ready and we can get this merged.

ifosch:
@riscy I see your point in this change also. I've changed to use `defvar-local` and `setq-local`. Thanks!

riscy:
Great thanks!  And, welcome to MELPA!

ifosch:
> Great thanks! And, welcome to MELPA!

Thank you very much!

PR #6638, opened 2020-01-15
Point 'undohist' to ideasman42's repository

ideasman42:

Store the undo history on disk.

This is an upgrade to the existing `undohist` package as the current repository has been archived by it's author and not updated since 2015, see: https://github.com/m2ym/undohist-el

The following improvements have been made.

- Store full undo/redo history, allowing `undo-only` to work as expected (this is the main improvement).
- Use a minor-mode<br>`(undohist-mode)` for a buffer-local mode.<br>`(global-undohist-mode)` to enable globally - to replace `(undohist-initialize)` which couldn't be disabled.
- Store history compressed (default to gz).
- Don't save history for ignored files.

See [changelog](https://gitlab.com/ideasman42/emacs-undohist/blob/master/changelog.rst) for a full list.

https://gitlab.com/ideasman42/emacs-undohist

New maintainer

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/undohist
+++ b/recipes/undohist
@@ -1,1 +1,1 @@
-(undohist :repo "m2ym/undohist-el" :fetcher github)
+(undohist :repo "ideasman42/emacs-undohist" :fetcher gitlab)


tarsius:
>> Relevant communications with the upstream package maintainer
>
> None needed.

Actually we would like to hear @m2ym's thoughts on this. He doesn't seem to be on Github much, so if he does not respond here in a few days, then you should send him an email asking to do so.

tarsius:
We have just transferred Tomohiro Matsuyama's packages to the Emacs orphanage, see #6644 

I would be happy to make you a maintainer.

Two requests though:

- Somehow you rewrote the original commits. Fix that by rebasing your own commits onto the old ones. (Also look into why that happened in the first place and try to avoid doing it in other repositories.)

- Please don't use your `elisp-autofmt` thing. IMO you should indent elisp the same way everyone else (+/- a few other rebels) has been doing it for a few decades now. You can of course indent differently in your own projects (though it will be an uphill battle because contributors will indent "wrong" or stay away completely), but please don't do this for projects you only contribute to or even take over.

ideasman42:
Since this is closed, what is the next step? Should I submit a new PR or comment on this one?

tarsius:
> Since this is closed,

We can still talk here though.

> what is the next step?

My plan was to use the repository in the Emacsorphanage as the new upstream. Does that not work for you?

tarsius:
> it seems like it could be best not to mutate undohist into something else.

Yes. So...

> what is the next step?

You'll have to wait for someone else to review your new package.

PR #6640, opened 2020-01-15
Add dired-git recipe

conao3:
----

Display Git repository information in dired.

https://github.com/conao3/dired-git.el

A author and maintainer.

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/dired-git
+++ b/recipes/dired-git
@@ -1,0 +1,1 @@
+(dired-git :fetcher github :repo "conao3/dired-git.el")


tarsius:
Other "git information in dired" packages exist. How does yours compare to those?

Please consider using plain old GPL instead of AGPL. The likelihood of someone creating some webservice based on this (with or without releasing any modifications) approaches zero. On the other hand using AGPL means this cannot be integrated into GPL code.

conao3:
Before developing this package,  I quickly search MELPA with `git dired` and here is [result](https://melpa.org/#/?q=git%20dired).
Yes, `dired-k` is "git information in dired", but this package purpose is "git repository information in dired".

I usually manage all my author repositories in `~/dev/repos` and fork repositories are located in `~/dev/forks`.
I want to see these repositories status in dired. I used `find . -mindepth 1 -maxdepth 1 -type d | xargs -I^ sh -c "cd ^; pwd; git status"`  before to complete that purpose.

---

And If you want to issue AGPL, please note this in CONTRIBUTING.
I check
> The package is released under a [GPL-Compatible Free Software License](https://www.gnu.org/licenses/license-list.en.html#GPLCompatibleLicenses)

and I decide all my packages distribute under AGPLv3 when start creating Emacs packages.

I have already 7 my packages in [MELPA](https://melpa.org/#/?q=conao3) and those packages are distributed under AGPLv3.
Please inform me to valid and legal way to changing AGPLv3 to GPLv3 on already released packages.

tarsius:
Actually the licenses are compatible, see https://en.wikipedia.org/wiki/Affero_General_Public_License#Compatibility_with_the_GPL, but I still feel it would be better use the GPLv3. Lets wait for some opinions from other people.

If you decide to change the license, then I recommend you create an issue in each of the repositories asking the contributors to give their blessing.

conao3:
I changed `dired-git` package license via my decision because it is built by only me.
I consider about other packages license later.

conao3:
@purcell, @riscy, please review it. 3 weeks passed after this PR submitted.
Or, please consider to add me as MELPA reviewer if you have no times.

riscy:
> On the other hand using AGPL means this cannot be integrated into GPL code.

I think this is a good reason to prefer GPL in this case.  Pragmatically, the extra provisions won't have any effect, and AGPL creates a gotcha for people who might want to reuse the code.

@conao3 also writes:

> please note this in CONTRIBUTING.

...which I can take a stab at later.  I think MELPA can be a bit more explicit about some things in there.

and

> please consider to add me as MELPA reviewer if you have no times.

Not my decision to make, but the easiest way to get started is to jump right in and start commenting on the open PR's  :)  Personally, I'm thrilled to see people taking a look at and using the packages that are up for PR, citing related packages, checking byte-compilation, checking for bugs, etc., regardless of whether they have the ability to merge.

--------

On this PR itself, I have a few non-blockers:

- `dired-git-disable-dirs` is meant to be a list of strings, but its type is `sexp` -- this is not a blocker, but you could set it to a list of strings
- dired-git.el#L260: It's safer to sharp-quote function names; use `#'`
- avairable -> available
- return value form -> return value from

But this looks good to merge.  And, thanks!

conao3:
I believe that just commenting on PR by just ordinary people without merge authority can make the PR discussion confusing.
In general, what my comments and other people point out does not match perfectly.

The package developer will then be fragmented and will have to deal with numerous reviews, and the experience will be worse.
But if you recommend it I would like to try it.

riscy:
I'd say give it a go.  It's already relatively common for this to happen on MELPA, and I think it's a good thing, so don't worry about the discussion becoming confusing.

If you're worried, you can always preface with: "I'm not a MELPA maintainer, but here's some feedback..."

conao3:
Good, thanks!

PR #6643, opened 2020-01-16
Add recipe for ivy-clipmenu.el

wpcarro:
Publishing github.com/wpcarro/ivy-clipmenu.el to Melpa.

[Ivy](https://github.com/abo-abo/swiper#ivy) integration with the clipboard manager, [clipmenu](https://github.com/cdown/clipmenu).

https://github.com/wpcarro/ivy-clipmenu.el

I'm the project creator and maintainer.

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/ivy-clipmenu
+++ b/recipes/ivy-clipmenu
@@ -1,0 +1,3 @@
+(ivy-clipmenu
+ :repo "wpcarro/ivy-clipmenu.el"
+ :fetcher github)


riscy:
Hello, thanks for this.  Here's a quick first pass:

### ivy-clipmenu.el
```
ivy-clipmenu.el:75:1:Warning: defcustom for `ivy-clipmenu-history-length'
fails to specify type
ivy-clipmenu.el:75:1:Warning: defcustom for `ivy-clipmenu-history-length'
fails to specify type
```

- No issues! 💯

- No issues! 💯

Other suggestions:
- ivy-clipmenu.el#L56: Use `temporary-file-directory` instead of /tmp in code
- It would be a good idea to add a license summary to the front-matter of your elisp file, just in case someone happens to _only_ receive your source file
- Your dependency list includes two versions of emacs (24 and 25.1)
- Your package name doesn't necessarily need to include the word "Emacs" in it - could it be "Client for clipmenu"?

wpcarro:
@riscy Thank you for the code review!

I've pushed the recommended changes upstream, so -- barring any additional feedback -- everything should be ready to go.

wpcarro:
I recently learned about `completing-read`. I think I should use that instead of `ivy-read` and re-brand this package as `clipmenu.el`: an Emacs integration with `clipmenu` and remove my dependency on `ivy`.

Unless there are any objections, I'm going to do this.

wpcarro:
Actually, I'm going to leave this as is, because I'd eventually like to support functions like removing entries from `clipmenu`, in which case, I'd like to use `ivy`.

Is there any work left to do before this gets published?

wpcarro:
Is there any work remaining for me to do? I'm unfamiliar with MELPA. How long does it typically take for a package to be published?

riscy:
Apologies for the long wait on this, I've had a lot in the backlog and have been looking at MELPA PR's in an arbitrary order (i.e. whatever was at the top of my inbox), and yours was unfairly delayed.

Everything looks great regarding the checklist.  I'm looking through the code now and the one last thing that leaps out at me is this:

```elisp
(defun ivy-clipmenu--checksum (content)
  "Return the CRC checksum of CONTENT."
  (s-trim-right
   (with-temp-buffer
     (call-process "/bin/bash" nil (current-buffer) nil "-c"
                   (format "cksum <<<'%s'" content))
     (buffer-string))))
```
Here you want to use `shell-quote-argument` on `content` in case it contains special characters (e.g. "&& rm -rf xyz").  But then again, if all we need is a checksum, why invoke bash?  Could we get away with just calling Emacs' built-in `md5` function?

Let me know what you think about making a change along those lines, and I'll take another look soon.

Another (small) item, not a blocker: if you add an `https://` before your references to `github.com/cdown/clipmenu` then <kbd>M-x find-file-at-point</kbd> will be able to identify those as URL's.

wpcarro:
@riscy - Thanks for the feedback.

> But then again, if all we need is a checksum, why invoke bash? Could we get away with just calling Emacs' built-in md5 function?

Unfortunately, this depends on the CRC checksum algorithm, which I couldn't find implemented in Emacs. I cannot use `md5` and be compatible with `cdown/clipmenu`.

- Using `shell-quote-argument` now.
- Added the fix for `ffap` support.

riscy:
Great, thank you for your patience.  And, welcome to MELPA. :)

PR #6645, opened 2020-01-18
Update the repository location of ruby-electric.el

knu:
I'm the maintainer of this package and moving the repository.

### DIFF
--- a/recipes/ruby-electric
+++ b/recipes/ruby-electric
@@ -1,1 +1,1 @@
-(ruby-electric :fetcher github :repo "knu/ruby-electric.el")
+(ruby-electric :fetcher github :repo "ruby/elisp-ruby-electric")


tarsius:
https://github.com/knu/ruby-electric.el => https://github.com/ruby/elisp

tarsius:
Lets talk about this first.

I generally encourage maintainers to use more than a single repository for related, yet distinct libraries/packages. The reason for this is that when those only partially related libraries are located in the same repository, then it becomes difficult to use just one of the libraries. Users of Melpa are not affected by this, but people who install packages by cloning their Git repository are. These likely are the users that contribute back, so you shouldn't make it harder for them. Users might end up loading outdated bundled copies of third-party libraries because of this.

This package used to be mirrored on the [Emacsmirror](https://emacsmirror.net), but I concluded that it *bundles third-party libraries* and therefore cannot be distributed on the mirror because one of the nice things about the mirror is that there are no feature conflicts between the mirrored packages.

So I moved it to the Emacsattic instead ([here](https://github.com/emacsattic/ruby-additional/commits/master)) and removed the bundled libraries `ruby-mode.el`, `inf-ruby.el` and `ruby-electric.el`. So now there are no conflicts but the remaining libraries also won't be updated when upstream makes updates. `ruby-mode` is part of Emacs and `inf-ruby` is maintained in a separate repository. (Same for `ruby-electric`, but you want to change that.) That leaves `ruby-additional.el`, `rdoc-mode.el`, `rubyydb2x.el` and `rubydb3x.el` all of which look very suspect to me. These libraries/packages would not be added to Melpa if they had to go through the review process now.

And now you want to add `ruby-electric.el` to this collection...
(`ruby-electric.el` is of better quality than the libraries mentioned above.)

The best outcome in my eyes would be if each of these libraries got their own repository and would then be improved enough to pass the Melpa review. The maintainers of https://github.com/ruby probably wouldn't want multiple repositories for minor ruby-related elisp-libraries under their umbrella, so a new organization https://github.com/ruby-elisp might have to be created.

Second best outcome would be if we continued the let `ruby-electric.el` have its separate repository and removed that file from the "various ruby thingies for emacs" repository again.

knu:
Let us think about it.  It's like we've had only a couple of people in the ruby core team to maintain these packages, so we've got little motivation to have separate repositories for each package under ruby/, nor to create a new organization just for them.  But if the latter works better for you and the users, I'm not unwilling to do that.

knu:
By the way, I don't completely see why ruby-additional.el had to be removed from MELPA when the features it provides are not available with GNU Emacs version of ruby-mode.el.

tarsius:
> By the way, I don't completely see why ruby-additional.el had to be removed from MELPA when the features it provides are not available with GNU Emacs version of ruby-mode.el.

We'll have to ask @purcell who removed it in 29ce18418cf8b5cbbaa7f299182613c03bd1f428.

purcell:
IIRC the files there appeared only to be of historical value and/or were not well packaged.

knu:
We (the ruby core team) agreed that there's no problem in creating one repository per elisp package right under https://github.com/ruby.  So I'll make that happen shortly.  Thanks for the feedback!

ruby-additional.el is still useful and I know there are many users who rely on it, so I'll file a PR for it after giving it a dedicated repository.

purcell:
Curious: why is `ruby-additional.el` helpful? Is none of the code there getting into Emacs releases? If it isn't, then the description is misleading. And if the code *is* getting into Emacs releases, I'd expect `ruby-additional.el` to include various `fboundp` or emacs version checks.

knu:
The description written six years ago indeed is misleading by now.  Unfortunately, there has been little to no communication between the maintainers on the both sides since then, when ruby-mode.el diverged.

I'll try, but it is unlikely that the code is getting into Emacs releases any time soon.

riscy:
I'll mark this one as awaiting upstream while we see what happens.

knu:
It has a new home: https://github.com/ruby/elisp-ruby-electric

tarsius:
Why did you add `:files ("ruby-electric.el"`? There are no other files in the repository. Do you plan to change that?

knu:
It's just a leftover of what I did in the original change.  Removed!

tarsius:
Thanks!

PR #6649, opened 2020-01-20
Add 'undo-fu-session' recipe

ideasman42:

Store the undo history on disk between sessions.

The intention is to make this a companion package for [undo-fu](https://gitlab.com/ideasman42/emacs-undo-fu), allowing them to remain separate as users wont necessarily want to use both.

I may eventually add hooks allowing additional `undo-fu` data to be stored by `undo-fu-session`, although this isn't urgent as they work fairly well together already.

Note that I submitted a [PR](https://github.com/melpa/melpa/pull/6638) to update `undohist`, which has been made orphaned.

While updating this package I think I made enough changes to warrant calling it a new package, since it's no longer compatible with files from `undohist`.

The following functionality is supported.

- Store full undo/redo history, allowing `undo-only` to work as expected.
- Use a minor-mode<br>`(undo-fu-session-mode)` for a buffer-local mode.<br>`(global-undo-fu-session-mode)` to enable globally.
- Store history compressed (default to gz).
- Fast early-exit on restore _(only decode undo data if buffer length & sha1 checksum match)._
- Option to limit number of undo-session files stored _(avoid adding files indefinitely until the user manually removes them)_.

https://gitlab.com/ideasman42/emacs-undo-fu-session

Maintainer.

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/undo-fu-session
+++ b/recipes/undo-fu-session
@@ -1,0 +1,3 @@
+(undo-fu-session
+ :repo "ideasman42/emacs-undo-fu-session"
+ :fetcher gitlab)


riscy:
Thanks :)  Quick first pass below for reference, but there are no blockers.

### undo-fu-session.el
- No issues! 💯

- No issues! 💯

- No issues! 💯

- undo-fu-session.el#L273: There is no need to quote lambdas (neither #' nor ')
- undo-fu-session.el#L270: Consider `unless ...` instead of `when (null ...)`
- undo-fu-session.el#L24: Prefer `https` over `http` (if possible)

### Maintainer/license

- PR by ideasman42: https://gitlab.com/ideasman42/emacs-undo-fu-session
- LICENSE excerpt: 'GNU GENERAL PUBLIC LICENSE...'

### Summary

- (undo-fu-session :repo "ideasman42/emacs-undo-fu-session" :fetcher gitlab) 
- Package-Requires: emacs "24.1"
- _elisp/undo-fu-session.el (gpl): Persistent undo, available between sessions
</details>

ideasman42:
Thanks, made the suggested updates.

PR #6650, opened 2020-01-21
Add svelte-mode package

leafOfTree:

Emacs major mode for Svelte.

https://github.com/leafOfTree/svelte-mode

I'm the maintainer.

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/svelte-mode.el
+++ b/recipes/svelte-mode.el
@@ -1,0 +1,3 @@
+(svelte-mode
+ :fetcher github
+ :repo "leafOfTree/svelte-mode")


riscy:
Hi!  Thanks for this - a quick first pass through the checklist.

### svelte-mode.el
```
svelte-mode.el:290:4:Warning: sgml-syntax-propertize called with 3 arguments,
but accepts only 2
In end of data:
svelte-mode.el:665:1:Warning: the function `string-join' is not known to be
defined.
```
For the latter, you can `(require 'subr-x)` -- the former seems like it may be a bug?

- No issues! 💯

- No issues! 💯

- svelte-mode.el#L311: You can use `not` or `null`

### Maintainer/license

- PR by leafOfTree: https://github.com/leafOfTree/svelte-mode
- License found (through API): 'GNU Lesser General Public License v3.0' (good)
- LICENSE excerpt: 'GNU LESSER GENERAL PUBLIC LICENSE...'

### Summary

- (svelte-mode :fetcher github :repo "leafOfTree/svelte-mode") 
- Package-Requires: emacs "26.1"
- _elisp/svelte-mode.el (gpl): Emacs major mode for Svelte
</details>

leafOfTree:
Hi, thanks for the review and suggestions, and sorry for my late response. Everything should be fine, except
> svelte-mode.el:290:4:Warning: sgml-syntax-propertize called with 3 arguments,
but accepts only 2

`sgml-syntax-propertize` accepts 3 arguments in emacs27, but 2 in emacs26. I hope this plugin works in both versions. So I make it like
```lisp
  (when (= emacs-major-version 26)
...
	  (sgml-syntax-propertize (point) end))))
  (when (> emacs-major-version 26)
...
    (sgml-syntax-propertize (point) end svelte--syntax-propertize)))
```
However, there is still the warning. I guess it's OK? 
Any suggestion is welcome.

riscy:
Ah, I see.  If you really wanted to quiet the Emacs 26 byte-compiler, you could use a hack (that might also inhibit some byte-compile optimizations on the function, but maybe it's not a bottleneck):
```elisp
(apply #'sgml-syntax-propertize (list (point) end svelte--syntax-propertize))
```

I notice you credit Emacs's `mhtml-mode` -- I see a number of commands ported over verbatim.  Excuse my ignorance and let me ask: would there be any sense in having this mode derive from `mhtml-mode`?  Would there be any sense in calling the functions directly instead of copying them over?

On the one hand you would get bugfixes (such as maybe the FIXME in `svelte--pre-command` and `mhtml--pre-command`) for free, but on the other hand you might end up tied down to a specific version of Emacs, so I can see both sides of the argument.

leafOfTree:
Thanks for the tips. The byte-compile warning has been quieted.
`mhtml-mode` is a great example. However, this mode has some differences from `mhtml-mode`. Some functions are copied with new names, but some are rewrited or added. 
I hope that this mode is consistent, extensiabe and dependent on basic modes. So I would prefer it is a different mode.

conao3:
Hi!

1. Please remove `Package-Version` header meta information.
https://github.com/leafOfTree/svelte-mode/blob/master/svelte-mode.el#L9
2. Why is there `Credits` header?
https://github.com/leafOfTree/svelte-mode/blob/master/svelte-mode.el#L11
3. Please remove top-level sexp. All packages should not have side-effects just require package for user-experience. see [[https://www.gnu.org/software/emacs/manual/html_node/elisp/Coding-Conventions.html#Coding-Conventions][Coding-conventions]].

leafOfTree:
Hi, thanks for the feedback, @conao3.

1. Removed.
1. Moved to commentary.
1. I know little about `top-level sexp`. Could you please show me an example and how to do with it? As I didn't find it in the link about `Coding-conventions`.

conao3:
https://github.com/leafOfTree/svelte-mode/blob/master/svelte-mode.el#L615-L618

If there is `advice-add`, it has side-effects just `require` this package.
If you really need it, please define global-minor-mode, or prepare `*-setup / *-teardown` function.

riscy:
Yeah, @conao3 has a good point.

One way I've dealt with this is to add the line like this:
```elisp
(advice-add 'emmet-detect-style-tag-and-attr :around #'svelte--emmet-detect-style-tag-and-attr-advice)
```
into the code that turns the major mode on.

Then you can add a function named `svelte-unload-function` that does the teardown and removes this advice.  The nice thing about naming it `svelte-unload-function` is that, if someone calls <kbd>M-x unload-feaure, svelte-mode</kbd>, then `svelte-unload-function` will be called to remove the advice.

leafOfTree:
Good idea. Thank you for your help. I've moved the advice and added a teardown function.
https://github.com/leafOfTree/svelte-mode/commit/486a3399ae818c6ed1ec7f024cb99a0997e527db

riscy:
Nice.  You should document that new function and maybe point users to the ``` `unload-feature' ``` function to explain why `svelte-unload-function` is named the way that it is.

I'm looking a little more closely at the code now and I have a couple more things to point out:

- for your `declare-files` such as `(declare-function emmet-check-if-between "emmet-mode")`, you can use `:ext` (e.g. `(declare-function emmet-check-if-between "ext:emmet-mode")`) to indicate to check-declare that the files should be checked if found, or otherwise ignored -- since your mode doesn't depend on them being there.   I personally like to see all `declare-function`'s at the top of the file.
- You should combine your top-level `defvar` and `setq`s.  Anywhere you have a `setq` at the top level should be avoided.  As an example: on line 140 you have `(defvar svelte-block-re)` and then all the way down on line 530 you have `(setq svelte-block-re "\\({[#:/]\\).*\\(}\\)$")`, but you can accomplish this all within one `defvar`.  Preferably these `defvar` go at the top of the `;;; Code` block.  (Incidentally, using `setq` also means that, if someone has deliberately modified this variable and they re-load your package, their modification will disappear -- whereas with `defvar` their modification remains intact.)
- It's just a personal opinion, but the change where you write "Mainly inspired by mhtml-mode." in your frontmatter could be strengthened to "Mainly based on mhtml-mode.", since you do take functions verbatim.

That should be all to wrap this up.

leafOfTree:
Thank you for your great suggestions on improving the code. I have now got a better sense of the code style. It's updated now. 
https://github.com/leafOfTree/svelte-mode/commit/998a587d77531b8c1b259175abe9d16574805382
https://github.com/leafOfTree/svelte-mode/commit/5ee10ede0df77cb5b1f52ad0065a9b19db35edaa

leafOfTree:
Hi, @riscy .  Is there anything else I should improve? Thank you.

riscy:
My fault, I missed the updates.

I notice a quick pair of byte-compile warnings on the latest:


```
svelte-mode.el:87:46:Warning: reference to free variable `svelte--css-submode'
svelte-mode.el:93:46:Warning: reference to free variable `svelte--js-submode'
```
which you could solve with a couple of `defvar`s I think.

leafOfTree:
Thanks for the quick response. I've added `defvar` to fix the warnings.

riscy:
No problem!  Welcome to MELPA, by the way.

leafOfTree:
Cool! Thanks a lot.

PR #6652, opened 2020-01-22
Add doct recipe

progfolio:
doct (declarative Org capture templates) defines an alternative, declarative syntax for Org capture templates

https://github.com/progfolio/doct

Maintainer

None needed

### Checklist

### DIFF
--- a/recipes/doct
+++ b/recipes/doct
@@ -1,0 +1,1 @@
+(doct :repo "progfolio/doct" :fetcher github)


alphapapa:
I've been following your progress in your Reddit posts.  Glad to see it progressing.

Quick feedback, the same I give to all Org-related packages: the name should have `org` in it.  `doct` doesn't suggest anything Org related; I would guess it referred to something related to Microsoft Word templates.  So I'd suggest naming the package `org-dct`.

BTW, have you made any progress regarding integrating it into Org proper?

progfolio:
I'm open to renaming the package if necessary.

I haven't started any discussion on the Org dev list just yet. Ideally I'd like to get some more feedback from users of the package to see if I've missed the mark on anything first.

As always, appreciate the feedback.

riscy:
The checklist looks good -- thank you for sticking closely to it!  A couple small checks I can quickly add before taking a deeper look:

- doct.el#L244: It's safer to sharp-quote function names; use `#'`
- doct.el#L247: It's safer to sharp-quote function names; use `#'`
- betewen -> between
- occurence -> occurrence
- Replace %doct(KEYOWRD)  -> KEYWORD
- poperty -> property
- creat -> create

@alphapapa has good advice about adding org to the name (my first guess was something to help with documenting source code or similar), although there are of course a few packages on MELPA that are org-related and lack org somewhere in their name.

progfolio:
Thank you for taking the time to look the code over.
Took care of the typos and the sharp quotes. Changes on latest master.

Let me sleep on the name change and I'll get back to you about it tomorrow.
Thanks again.

alphapapa:
It would be more consistent with the naming convention to have an `org-` prefix.  That prefix makes discovering Org-related packages very easy.  Consider `package-list-packages`, and that a name like `doct-org` will be sorted far from all the `org-` packages.

progfolio:
I think my main gripe with renaming it is having to read/maintain code with an even longer faux namespace prefix. Leaning toward keeping it as "doct", but I'll sleep on it and try to think of some more names tomorrow.

alphapapa:
You don't have to tell me.  At least `org-dct` isn't nearly as long as names like `org-super-agenda` and `org-recent-headings`.  ;)  You get used to it after a while.  And there's always https://github.com/Malabarba/Nameless if it bothers you that much.

progfolio:
@alphapapa Yeah, it's a shame namespaces haven't found their way into Elisp yet. Nameless does look like it could help a bit. Thanks for pointing that out.

@riscy I think I'd like to keep the name "doct" if that's OK. If it's a blocker, I'll rename. Thanks again.

riscy:
Okay -- it isn't a MELPA requirement (yet?).  Sorry @alphapapa (I think it's good to continue to encourage this convention even if it's not always successful).  In any case, since the package summary is `DOCT: Declarative Org capture templates` it will remain searchable under the keyword "org", which is a good minimum.

Not a blocker, but for your consideration:
- doct.el#L25: Prefer `https` over `http` (if possible)

...and, welcome to MELPA. :)

progfolio:
Thanks!

PR #6653, opened 2020-01-24
Add julia-snail recipe.

gcv:

Julia Snail is an IDE for Julia (https://julialang.org/), similar to SLIME for Common Lisp and CIDER for Clojure.

https://github.com/gcv/julia-snail

I am the author.

n/a

### Checklist

Please confirm with `x`:

- [ ] I have confirmed some of these without doing them *(very funny)*

### DIFF
--- a/recipes/julia-snail
+++ b/recipes/julia-snail
@@ -1,5 +1,4 @@
 (julia-snail
  :fetcher github
  :repo "gcv/julia-snail"
- :files ("*"
-         (:exclude ".dir-locals.el" ".gitignore" ".travis.yml" "tests")))
+ :files (:defaults "JuliaSnail.jl"))
--- a/recipes/julia-snail
+++ b/recipes/julia-snail
@@ -1,0 +1,5 @@
+(julia-snail
+ :fetcher github
+ :repo "gcv/julia-snail"
+ :files ("*"
+         (:exclude ".dir-locals.el" ".gitignore" ".travis.yml" "tests")))


alphapapa:
I don't use Julia, but this looks really cool, and the readme looks very nice.  One thing it lacks, though: screenshots!  It would be great if you could demonstrate some of the features.  Here are some helpful resources: https://github.com/alphapapa/emacs-package-dev-handbook#animations--screencasts

gcv:
Great suggestion, thanks! I made a screencast using some of your suggestions and a few inventions of my own. Scroll down in the README (https://github.com/gcv/julia-snail) to see it. It shows off the important things Snail can do.

alphapapa:
Cool!  How did you do the little caption box in the lower-left corner of the screencast?

gcv:
The caption box uses a small posframe hack: https://github.com/gcv/dotfiles/blob/9d573bb/emacs/interactives.el#L571-L605

I looked at keycast, and decided it would produce too much noise. I also wanted to add commentary, but with less effort than required for making a full-on video with an audio track. This was the result.

alphapapa:
Very cool!  If you have a blog, you should write an article showing that off.  A lot of package authors could find that trick useful in their demos.

recipes/julia-snail
```diff
@@ -0,0 +1,5 @@
+(julia-snail
+ :fetcher github
+ :repo "gcv/julia-snail"
+ :files ("*"
```
riscy: Could this recipe be simplified to use the `:defaults` recipe with one additional `:include`?  One reason for this is to make it easier for your package to grow without requiring additional changes to the recipe.
```elisp
(julia-snail
 :fetcher github
 :repo "gcv/julia-snail"
 :files (:defaults "JuliaSnail.jl"))
```
(I'm not sure if you also need snail.png, and generally MELPA doesn't distribute the LICENSE and README)

gcv: Thanks for all the comments, @riscy. I believe I resolved all of them with the latest batch of commits to the julia-snail repository. I also simplified the recipe as per your suggestion.

riscy:
I realize vterm is core to this package, but it's worth noting that:

> This package is in active development and, while being stable enough to be used as a daily-driver, it is currently in early alpha stage

and it might not be difficult to support another REPL, such as comint-mode.  In any case, this looks really cool.  Welcome to MELPA. :)

tarsius:
@gcv Please rename the parser library instead of relying on this trick `(require 'julia-snail-parser "parser.el")`. No other of the ~4400 packages on Melpa or the ~7400 on the Emacsmirror currently do this and as a result the metadata extraction tools do not support it.

gcv:
Done in https://github.com/gcv/julia-snail/commit/f37e53f9dfb535db7e3cb10304f359b40638855d

Though I must note that we already have to deal with the absence of namespaces in Elisp, and having to port the same nonsense to the area of directories and filenames feels like adding insult to injury. This "trick" is a [fully documented](https://www.gnu.org/software/emacs/manual/html_node/elisp/Named-Features.html#index-require-1170) use of the `require` form.

PR #6655, opened 2020-01-24
Add recipe for languagetool.el

PillFall:
Publishing languagetool.el to MELPA.

Use LanguageTool as grammar checker.

https://github.com/PillFall/Emacs-LanguageTool.el

I'm the maintainer and creator of this package.

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/languagetool
+++ b/recipes/languagetool
@@ -1,0 +1,4 @@
+(languagetool
+ :repo "PillFall/Emacs-LanguageTool.el"
+ :fetcher github)
+ 


riscy:
Thanks for this.  Rolling through with a quick pass:

### languagetool.el
```
languagetool.el:291:30:Warning: reference to free variable
`languagetool-hint-idle-delay'
languagetool.el:292:30:Warning: reference to free variable
`languagetool-hint-function'
languagetool.el:290:9:Warning: assignment to free variable
`languagetool-hint--timer'
In languagetool--clear-buffer:
languagetool.el:303:9:Warning: reference to free variable
`languagetool-hint--timer'
```

```
Warning (emacs):
languagetool.el:334: Argument `lang' should appear (as LANG) in the doc string
```

```
1 issue found:
1:0: error: Package should have a Homepage or URL header.
```

- languagetool.el#L242: It's safer to sharp-quote function names; use `#'`
- languagetool.el#L57: Prefer `https` over `http` (if possible)

Lastly I notice you are including `languagetool-pkg.el`.  This is discouraged, since MELPA will create a `-pkg.el` file of its own based on the contents of `languagetool.el`, so it may be worth excluding it from your recipe.

### Maintainer/license

- PR by PillFall: https://github.com/PillFall/Emacs-LanguageTool.el
- License found (through API): 'GNU General Public License v3.0' (good)

### Summary

- (languagetool :repo "PillFall/Emacs-LanguageTool.el" :fetcher github) 
- _elisp/languagetool.el (gpl): LanguageTool integration for grammar check
- _elisp/languagetool-pkg.el (unknown license): Couldn't parse header
</details>

PillFall:
In the last commits I push the changes for the advices and corrections made. There is only a thing:

- In languagetool.el#L58: There is no `https` version of that page.

riscy:
Hi!  I'm really sorry to have overlooked this for so long.  I'm taking a deeper dive now:

A quick look through the source code:
```elisp
    (dolist (arg languagetool-java-arguments)
      (when (stringp arg)
        (setq arguments (append arguments (list arg)))))
    (setq arguments (append arguments '("-jar")))
    (setq arguments (append arguments (list languagetool-language-tool-jar)))
    (dolist (arg languagetool-language-tool-arguments)
      (when (stringp arg)
        (setq arguments (append arguments (list arg)))))
    (setq arguments (append arguments '("-c")))
    (setq arguments (append arguments '("utf8")))
    (setq arguments (append arguments '("--json")))
```
I notice quite a few times you check `(stringp arg)`, but is that necessary if that thing should be a string (or nil) anyway?  (Note `string=` works with nil).

I also feel like this could all be replaced by a single `nconc` (or maybe a `cl-mapcon` if you really want to explicitly check whether something is a string?), for example this form:
```elisp
(nconc nil
       '("-jar" "xyz")
       nil
       '("-c" "utf8")
       nil
       '("--json"))
```
turns into `("-jar" "xyz" "-c" "utf8" "--json")`, so it may be exactly what you're looking for.

PillFall:
Hi!

I tried to change to `nconc` but the function always returns `nil`, I don't know what I'm doing wrong. But following your direction, I decided to delete all those `stringp` and minimize the code.

I pushed the changes in the last commit.

tarsius:
@riscy please take another look. Thanks!

riscy:
@PillFall I apologize for leaving this unanswered (twice now!) but I'm hopefully settling into a workflow where this won't happen as often.  Thanks for your patience. :)

This looks solid now.    One thing I would mention is that the repo name `Emacs-LanguageTool.el` might be renamed to match the case of the file, e.g. `Emacs-languagetool.el` or I might even suggest `languagetool.el` (or even `language-tool.el`, although you'd have to modify the recipe, which isn't necessary at this point).  A couple of other suggestions that are minor and non-blocking:
- languagetool.el#L22: Prefer `https` over `http` (if possible)
- Correct te buffer -> Correct the buffer
- recomends -> recommends
- LanguagetTool -> LanguageTool
- hightlights -> highlights
- pased as argument -> passed as an argument

All that said, welcome to MELPA! :)

PillFall:
@riscy, There's no problem, I'll keep your suggestions in mind and apply them in the next commits.

Thanks for all. I'm glad to be in here.

PR #6658, opened 2020-01-26
Add README and other org files

ouboub:
* recipes/matlab-mode (:fetcher): Add the org files

branch : matlab-readme

Provides support to matlab, for editing and running code via the elisp matlab-shell

https://github.com/matlab-mode/mirror.git

I am mainly the maintainer of the relevant repositories, and tester of new code.

### Checklist

Please confirm with `x`:

- [x ] I've built and installed the package using the instructions in [CONTRIBUTING.org](https://github.com/melpa/melpa/blob/master/CONTRIBUTING.org)

### DIFF
--- a/recipes/matlab-mode
+++ b/recipes/matlab-mode
@@ -1,7 +1,7 @@
 (matlab-mode
  :fetcher git
  :url "https://git.code.sf.net/p/matlab-emacs/src"
- :files ("*.el" "*.m" ("toolbox" "toolbox/*.m")
+ :files ("*.el" "*.m" "*.org" ("toolbox" "toolbox/*.m")
          ("toolbox/+emacs" "toolbox/+emacs/*.m")
          ("toolbox/+emacs/@Breakpoints" "toolbox/+emacs/@Breakpoints/*.m")
          ("toolbox/+emacs/@EmacsServer" "toolbox/+emacs/@EmacsServer/*.m")


purcell:
Generally we don't bundle arbitrary documentation into packages because it ends up in an obscure package installation directory that the user is not going to browse to when they need information. The exception to this is `.info` documentation, which is automatically added to the Info index. Is there functionality in `matlab-mode` to find and open this documentation for the user? If not, this PR will likely be declined.

ouboub:
I have been told my the GNU emacs maintainers, that there package infrastructure, needs a README or a README.org file, otherwise the following happens:
run
M-x package-list-packages
then when matlab is listed 
hit i
if there is no README file in the package, emacs (current master) gives a confusing error backtrace
if you want I can forward the maintainer's email.
I can remove the NEWS.org file if you require that

ouboub:
HI, 

so what's is up? I need the REAME.org to be part of the package, otherwise, the emacs package manager return an error. The current state of art is unclear to me.
Could someone please clarify? thanks

alphapapa:
Please quote the email here.

purcell:
> I have been told my the GNU emacs maintainers, that there package infrastructure, needs a README or a README.org file, otherwise the following happens:

Maybe they do, but we don't. MELPA extracts the `Commentary:` of your package into a `README.txt`, and that is what is displayed when hitting `?` in the `list-packages` view. (`i` marks packages for later installation, so I think you meant `?`.) `README.org` would be ignored in this regard, but bundled into the downloaded tarball where it would not be easily found.

riscy:
I'll mark this one awaiting @ouboub -- no hurry.

riscy:
Pinging @ouboub just to check the status on this one.

tarsius:
Another friendly ping!

PR #6661, opened 2020-01-28
add transient-dwim recipe

conao3:

Useful preset transient commands

https://github.com/conao3/transient-dwim.el

Author and maintainer

none needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/transient-dwim
+++ b/recipes/transient-dwim
@@ -1,0 +1,1 @@
+(transient-dwim :fetcher github :repo "conao3/transient-dwim.el")


riscy:
Thanks for adhering to the checklist!  Can you elaborate some use cases for me?  (I think some more documentation on the repo would be great too.)  I see by default it's loading up a magit interface, but what else could I potentially do?

conao3:
Provides a unified dispatch interface for major mode, minor mode, and packages.
Today,
- dired-mode
- image-dialed-thumbnail-mode
- dired-git-mode
- magit
but more major and minor modes and packages will be supported.
The user only has to set one item, and as you can see in the comments, only the following:.
```
(define-key global-map (kbd "M - =") 'transient-dwim-dispatch)
```

Once configured, simply press `M-=` and you'll get a context-sensitive transient menu.
There are currently few supported modes, so the `dired` buffer is a good way to see how to use it.
If you are using `magit`, all of its functions can be accessed from `M-= M-=` in all buffers.
Just press `M-=` three times and you will be able to commit using the `magit` interface.

riscy:
Okay, I see.  I think it'd be worth mentioning where these transients can be used somewhere in the frontmatter or in your README, but I may be missing it.  Also, I had to look up the "DWIM" acronym :) so it may be worth adding what it means to your frontmatter too (for folks like me), but it does seem like a good package name.

tarsius:
1. I recommend you discard `transient-dwim--declare-function`. The definition is eleven lines long, add three lines of "data" to that, and why? To avoid having to write three lines. So at best this is premature optimization. But even if you end up with a hundred `declare-function` forms I would still recommend just doing that. A little boilerplate hasn't killed anyone.

2. The `all` wrappers around Magit commands should not *override* the arguments, instead they should add `--all` to them. Have a look at the `interactive` form of `magit-commit-create`. Do something similar in your wrapper; cons `-all` onto the other arguments (but avoid duplicating it).

3. `transient-dwim--define-transient-command-multi` is a big mistake. That's how things were done in the days of `magit-key-mode.el`. Two implementations later things are much more flexible; lets not give that up.

   `transient-dwim--create-docstring` is fine and you have basically two ways of using that.
   * Modify it a bit so that it changes the doc-string of a command that has already been defined. Then define each prefix command using its own literal `define-transient-command`, followed by a call to this "update doc-string" function.
   * Define a macro like `transient-dwim--define-transient-command-multi` but change it so that it only defines a *single* prefix command.

4. Once you have stopped defining all prefixes using a single form its also easy to see that you should not use `declare-function` to avoid having to require `magit`. Instead create a second library, which requires `magit` and defines your magit wrapper.

conao3:
Thanks for your review, @tarsius!
I split package to core package and magit integration package. (solve your point 1, 4)
2should be solved after to submit `transient-dwim-magit` to MELPA.

Well, how to solve point 3?
It is important that I can easily define the relevant transient commands in order to support a number of major and minor modes.
You mentioned the docstring, but I'm leaving a way to override it if I want to specialize the docstring.

Also, you mentioned more flexible to use the transient macro directly than the wrapper, but I don't parse any arguments passed to transient and just pass it to transient.
Therefore, we enjoy the flexibility of transient as your result.

And you have known it fails this method from your experience.
However, I am convinced that this topical use is successful.
As suggested to stop premature-optimizing on function declaring, don't stop defining transient commands in this way with "premature-optimizing".
Let me make silly mistakes. I believe it will be sure of learning if it is failing.

riscy:
Just checking in on this one, I notice that the currently PR'd recipe `(transient-dwim :fetcher github :repo "conao3/transient-dwim.el")` will now include both elisp files, which I think is not deliberate.

conao3:
Thanks, recipe fixed!

riscy:
Cool. :)  One suggestion I have with this:
```
(transient-dwim :fetcher github :repo "conao3/transient-dwim.el"
                :files (:defaults (:exclude "transient-dwim-magit.el")))
```
is that every time you add a new transient-dwim subpackage, you'll have to update the `:exclude` list on this recipe (and any other similar recipe), which means another PR, etc.  Probably you should not use `:defaults` in this case, or you could use another repo.

Tarsius wrote above that `transient-dwim--define-transient-command-multi` is a mistake -- can I assume this is not resolved yet?

conao3:
I revert the package split and the recipe.
If `magit` is not installed, the `magit` related menu is not displayed.
This feature is why this package claims to be *do-what-i-mean*.

Also, I don't think `*-define-transient-command-multi` is a mistake.
It is a reasonable way to take full advantage of the Lisp's homoiconography.

conao3:
@purcell, @riscy 

While @tarsius says my `transient-dwim--define-transient-command-multi` is a big mistake, but I'm happy with the status quo of being able to deal with new packages with minimal code adding.
https://github.com/conao3/transient-dwim.el/commit/ad450448f9acf14623e562a4e73bb5b39b18b4f2

Being able to give data and generate instructions in this way is a feature of Lisp, and it's why I use it.
I'd love to hear your review.

riscy:
I think if you've considered the feedback and are happy with it, that's more important.  I'll go ahead and merge.

conao3:
Thanks!

PR #6663, opened 2020-01-30
Add recipe for virtual-auto-fill

luisgerhorst:

`virtual-auto-fill` is a package to [display unfilled text in a readable way](https://github.com/luisgerhorst/virtual-auto-fill/blob/master/README.md#screenshots).  It wraps the text as if you had inserted line breaks (e.g. using [`fill-paragraph`](https://www.gnu.org/software/emacs/manual/html_node/emacs/Fill-Commands.html) or `auto-fill-mode`) without actually modifying the underlying buffer.  It also indents paragraphs in bullet lists properly.

Internally, [Adaptive Wrap Prefix mode](http://elpa.gnu.org/packages/adaptive-wrap.html) mode, [Visual Fill Column mode](https://github.com/joostkremers/visual-fill-column), and [Visual Line mode](https://www.gnu.org/software/emacs/manual/html_node/emacs/Visual-Line-Mode.html) are employed to wrap paragraphs and bullet lists between the wrap prefix and the fill column. If you don't need all of the features provided by Virtual Auto Fill mode, it may be better for you to use a subset of these modes directly. However, if you often switch between filled and unfilled text, Virtual Auto Fill mode catches you when you accidentally invoke `fill-paragraph` (by default bound to `M-q`) out of habit in a virtually filled buffer. To prevent you from breaking the formatting convention without noticing it, you are asked whether you actually meant to fill the paragraph and if you want to be warned again next time.

https://github.com/luisgerhorst/virtual-auto-fill

I'm the original author and current maintainer.

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/virtual-auto-fill
+++ b/recipes/virtual-auto-fill
@@ -1,0 +1,1 @@
+(virtual-auto-fill :repo "luisgerhorst/virtual-auto-fill" :fetcher github)


conao3:
Hi!

I cannot find any issues for this package, what is a blocker to leave it, @riscy?

riscy:
This is of course non-blocking but you can add:
```elisp
(put 'virtual-auto-fill-visual-fill-column-warning-in-emacs-pre-26-1
     'risky-local-variable t)
```
after the respective variable if you want to be extra safe.

I like your solution in `virtual-auto-fill--restore-state`.

You require diminish so that you can `(diminish 'visual-line-mode)`.  I think there are a few (admittedly small) issues with this.  First, you don't do the same for `adaptive-wrap-prefix-mode` (whose lighter happens to be "" today, but maybe it won't always be).  Next, what if the user prefers using visual-line-mode with its lighter intact in another buffer -- and meanwhile you're diminishing it?  What if the user prefers using visual-line-mode with its lighter /diminished/, but your call to `(diminish-undo 'visual-line-mode)` undoes their customization?  These are small issues, but ultimately I think a mode should not automatically diminish other minor modes -- I'll wait to see what what you think.

- virtual-auto-fill.el#L253: Consider `unless ...` instead of `when (not ...)`
- virtual-auto-fill.el#L26: Prefer `https` over `http` (if possible)
- virtual-auto-fill.el#L160: You can use `not` or `null`

luisgerhorst:
Thanks for the suggestions! I was able to implement all of them and I agree that the diminishing is an issue. Maybe when I have time I will try to implement a save/restore for the diminishment and the key mapping too but for now I have removed it.

riscy:
Great, this looks like a thoughtful package, and with good documentation.

And, welcome to MELPA!

PR #6664, opened 2020-01-30
add recipe amread-mode

stardiviner:

A minor mode helping user reading like Spritz.

https://github.com/stardiviner/amread-mode

maintainer

**None needed**

### Checklist

Please confirm with `x`:


terlar:
How does this compare to spray? https://melpa.org/#/spray

stardiviner:
@terlar Similar with spray, but still different, It is very much like Spritz which only display one word in buffer. But I just display property on word and move forward. I keep buffer content what it is.

stardiviner:
I try to record a screencast for demo, but I found I can't record it correctly.

terlar:
Thank you, that would definitely help, I will also try it when I have time. I recently found the tool [peek](https://github.com/phw/peek) worked well to record such videos. Have you tried it?

stardiviner:
I tried `peek`, still recording weird. Can't record the property moving animation. Weird. @terlar Can you record one to help me if you can? Thanks for this.

terlar:
Here you go:
Peek 2020-01-30 23-18

Is this looking like it should for you?

I have some issue with using the keys, it worked a few times, but then I just got buffer is read-only on every key press q/SPC. So I had to quit by doing `M-x`.

stardiviner:
@terlar Thanks for your screencast gif and testing out. Can you tell me your latest using commit version? I bind `q` to quit amread-mode. and bind `[SPC]` to pause-or-resume. In the design, amread-mode will set buffer to read-only-mode, and exit read-only-mode when exit amread-mode.

stardiviner:
I updated code, added commands to enable/disable amread-mode, then bind `q` to `amread-mode-disable` explicitly. Should fix your problem now.

recipes/amread-mode
```diff
@@ -0,0 +1 @@
+(amread-mode :repo "stardiviner/amread-mode" :fetcher github :files ("amread-mode.el"))
```
riscy:
```suggestion
(amread-mode :repo "stardiviner/amread-mode" :fetcher github)
```

stardiviner: Thanks

stardiviner:
> amread-mode.el:44:1:Warning: Unused lexical variable `length'

I fixed this issue by remove this unused variable.

> prefer https over http

fixed too

Thanks for your tips. And other LICENSE  text, I'm not very clear understanding.

riscy:
> And other LICENSE text, I'm not very clear understanding.

No need to act on any of that stuff. :)

alphapapa:
FYI, for recording screencasts: https://github.com/alphapapa/emacs-package-dev-handbook#animations--screencasts

conao3:
Hi!

1. Please consider adding Cask file is good to easy resolve dependency, build, packaging etc.
3. Please remove unneeded `prog1` because return value is not used.
4. Please consider removing `amread-mode-{enable, disable}` function.
  Minor mode is safely enabling and disabling minor mode, `amread-mode` is already interactive toggle function.

stardiviner:
@conao3 Thanks, I added Cask file now. And .gitignore, and followed you advice by some modifications.

riscy:
A couple last items, but I think we're almost there:

- You use the keyword 'tools' but per `finder-known-keywords` that keyword is for "programming tools".  (Many of these keywords feel like they come from a different Emacs era, but I think we can still conform.)  I feel like "convenience" is a good match, or possibly "wp" (word processing).
- I don't see "speed-reading" as a keyword anywhere -- but isn't that sort of what this is?  I would have trouble finding this if I were searching for such facilities.
- Your `:lighter` shouldn't have a trailing space -- only a leading space -- and probably shouldn't have a space in it.  What about just `" amread"`, which makes it really clear which minor mode is active?
- Should `amread-start` and `amread-stop` get an `;;;###autoload` cookie?
- `(define-key map [remap keyaobrd-quit] 'amread-mode-disable)` -> `keyboard-quit`
- If narrowing is in effect on a buffer, I wonder if you want to use `(goto-char (point-min))` instead of `(goto-char 0)` (on line 71)

stardiviner:
@riscy Hi, I updated amread-mode, actually I updated many days ago on my laptop, but I'm offline, so now pushed up. Thanks for point out.

riscy:
Awesome, thank you. :)

I'm seeing a couple more tiny little nags:

### amread-mode.el ###

```
amread-mode.el:44:1:Warning: Unused lexical variable `length'
```

- No issues!

```
1 issue found:
```

And while I'm having a peek, you could sharp-quote the keybinding functions (which will help the byte-compiler detect if they get misnamed in the future), as in:
```elisp
    (define-key map (kbd "q") #'amread-mode-quit)
    (define-key map (kbd "SPC") #'amread-pause-or-resume)
    (define-key map [remap keyboard-quit] #'amread-mode-disable)
```
...and actually, applying that change revealed to me that `#'amread-mode-disable` does not exist as a function. :)

stardiviner:
> amread-mode.el:44:1:Warning: Unused lexical variable `length'

This is necessary step to move point forward. If I remove it, the point will not move forward.

Other modifications done.

riscy:
> This is necessary step to move point forward

I see.  You have a couple options here -- `let` forms let you use `progn` so you can:
```elisp
  (let* ((begin (point))
         ;; move point forward. NOTE This forwarding must be here before moving overlay forward.
         (end (progn
                (skip-chars-forward "^\s\t\n—")
                (skip-chars-forward "—")
                (point)))
```
or, if for some reason you want to keep that length calculation intact, you can tell the compiler to ignore it by prefixing with `_`:
```elisp
  (let* ((begin (point))
         ;; move point forward. NOTE This forwarding must be here before moving overlay forward.
         (_length (+ (skip-chars-forward "^\s\t\n—") (skip-chars-forward "—")))
         (end (point)))
```

I notice you still haven't fixed the issue where `amread-mode-disable` isn't defined -- please re-read that carefully.

riscy:
> please re-read that carefully.

Actually it's me who needs to re-read carefully :)  I had an old amread buffer open.

stardiviner:
:) I indeed renamed deleted command name.
I select the second solution incause I need to use the calculated length in future.

stardiviner:
Any update on this? I think this is stable for MELPA users. :)

riscy:
Sorry for the delay, I'll take a look again soon!

riscy:
Looking good -- I noticed you recently added some cl code so we're here with the byte-compile:
```
amread-mode.el:136:1:Warning: the following functions are not known to be
    defined: incf, decf
```
To resolve these, you should `(require 'cl-lib)` and use `cl-incf` and `cl-decf` (`cl` is deprecated, so we prefer the `cl-lib` equivalents).

riscy:
(Ping me and this is a quick merge after that.)

stardiviner:
Sorry for delay, I added `cl-lib` requirement, and updated GitHub commit. Thanks for reviewing my code :)

riscy:
Non-blockers (sorry for all the nags), but these just popped up.  This is just an fyi:

```
Warning (emacs): 
amread-mode.el:99: Lisp symbol `amread-mode' should appear in quotes
Warning (emacs): 
amread-mode.el:104: Lisp symbol `amread-mode' should appear in quotes
Warning (emacs): 
amread-mode.el:109: Lisp symbol `amread-mode' should appear in quotes
Warning (emacs): 
amread-mode.el:121: Lisp symbol `amread-mode' should appear in quotes
```

stardiviner:
Oh, I have not re-run `checkdoc` after added new code. Thanks for merging. I updated my docstrings. And I run `checkdoc`, now all passed. :100:

Issue #3493, opened 2016-01-13
Collect stats on popularity of different Emacsen


Wilfred:
I believe package.el has a user agent that includes the Emacs version.

Could you look at access logs and find out the popularity of the different Emacs versions? This would be useful data for package authors, to help us decide what versions we should support.

It'd also be interesting to know if there are any other Emacsen (eg XEmacs) using MELPA.

lunaryorn:
:+1: Would be awesome to have these!  

Sometimes I caught myself thinking about adding a little analytics to Flycheck to learn what Emacs version I should care for 😊

purcell:
There's `url-user-agent`, but that wasn't available back in Emacs 24.1, at least. The current default for that ("URL/Emacs") doesn't include the Emacs version, and that's indeed what in the access logs.

We might be able to get the default changed upstream, but that won't help much with distinguishing all the old versions.

lunaryorn:
> We might be able to get the default changed upstream, but that won't help much with distinguishing all the old versions.

I think that'd be great. Data for older version's lost then, but we could at least gather proper statistics for the upcoming Emacs 25 and onwards.

/cc @Malabarba

dunn:
Some upstream discussion: https://debbugs.gnu.org/cgi/bugreport.cgi?bug=18812

Malabarba:
Well...  who feels like bringing it up again? :-)

cpitclaudel:
Done here: https://lists.gnu.org/archive/html/emacs-devel/2016-03/msg01344.html

lunaryorn:
@cpitclaudel Thank you very much for bringing this issue to emacs-devel.  I hope that we end up with a solution that gives us at least some insights into our user base.

milkypostman:
very nice.

I haven't participated in many discussion on the mailing list but is it now
up to someone to suggest the code change?

cpitclaudel:
I can definitely submit a patch to the mailing list :) Even better, someone else could. Increasing the number of voices in support of this proposal would help (thanks Sebastian!)

cpitclaudel:
In particular, it would be nice to see this as a bugfix instead of a new feature; this would allow this to go into Emacs 25 instead of 25.1.

cpitclaudel:
[...]

PR #6666, opened 2020-01-31
Log processor rewrite

purcell:
See #3493. Filing as a pull request for visibility with @milkypostman. Will merge soon unless there's a blocker.

---

Okay, so I've aggressively reworked the log processor, and I have backfilled a pair of databases from the historical logs ready to switch over (which took ages!).

The new schema stores user agents, downloaded package versions and individual download times: basically all the interesting stuff in the log files. This will let us do all manner of per-package, per-time-period, per-version reporting, as well as continue to get the current per-package download counts.

Here's how it compares...

### Total number of downloads

```
Old: 107624260
New: 173760968
```

Our methodology has changed, in this regard: a download is now a successful GET of a specific versioned package file by a distinct IP / user agent pair. Previously it was unique `(package-root-name, ip)` pairs over all time. Some packages are heavily affected, e.g. `dash` will jump from around 1.05M downloads to around 2.5M.

### Space usage

```
Old: 3659710464 bytes
New: 3978551296 bytes
```
Surprisingly, the result is still half the size of the "old old" `WITH ROWID` table file, and only slightly larger than the "Old" `WITHOUT ROWID` version. This is achieved by aggressively moving strings out of the downloads tables into separate tables and referencing them by ID.

#### Query time for package totals to create JSON

```
Old: 215s
New: 232s
```

### Ingestion speed

```
Old: Unknown
New: Around 5000 (matching) log lines per second
```

I believe the speed to be comparable, though.

### DIFF
--- a/docker/logprocessor/process_log.py
+++ b/docker/logprocessor/process_log.py
@@ -2,20 +2,19 @@
 
 
 import argparse
-from datetime import datetime
+from datetime import datetime, timedelta
+from timeit import default_timer as timer
 import gzip
 import json
-import os
 import re
 import sys
 import time
-import tempfile
 import sqlite3
 from operator import or_
 
 LOGREGEX = r'^(?P<ip>[\d.]+) [ -]+ \[(?P<date>[\w/: +-]+)\] ' \
            r'"GET /+packages/+(?P<package>[^ ]+)-(?P<version>[0-9.]+).(?:el|tar) ' \
-           r'HTTP/\d.\d" 200'
+           r'HTTP/\d.\d" 200 \d+ "[^"]*?" "(?P<agent>[^"]*)'
 
 
 def json_handler(obj):
@@ -30,7 +29,7 @@ def json_handler(obj):
 
 def json_dump(data, jsonfile, indent=None):
     """
-    jsonfiy `data`
+    jsonify `data`
     """
     return json.dump(data, jsonfile, default=json_handler, indent=indent, encoding='utf-8')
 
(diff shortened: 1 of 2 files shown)


milkypostman:
Nice! Please let me have a day or two to process this. What a way to wake
up!

purcell:
Sure. Cutover plan would be to simply merge this, and then manually run the script against any logs for the last few days, to ensure none have been missed. And then delete the old DB files, freeing up a few GB again.

This was a good excuse to play with sqlite again.

milkypostman:
thanks Steve. not as involved as I expected, guess it depends on your RE skills

purcell:
I'm all about the regexps. :-)

purcell:
Yeah, I think the only thing I've realised is that using the rowids like this is a bit dangerous if existing rows get changed e.g. by `VACUUM`, because the rowids can be reassigned. I might tidy it up slightly in due course, but I think it's not a big risk right now: the DB is purely additive, and we don't vacuum it in normal operation, because we don't delete records.

dickmao:
> a download is now a successful GET of a specific versioned package file by a distinct IP / user agent pair. Previously it was unique `(package-root-name, ip)` pairs over all time.

So a package count gets incremented when the someone updates his emacs, when before it didn't?  Not that I believe the telemetry is super reliable or meaningful, but among the three counting methods (in descending order of spuriousness) 1. pure GET counts, 2. bucket by IP address and User Agent, and 3. bucket by IP address, I actually prefer 1 and 3 over 2.  Yes, 1 overrepresents the most, but it's also the least ambiguous.

purcell:
@dickmao All counting methods have different trade-offs and are variously inaccurate, yes. I'm content with this one for now.

alphapapa:
I hadn't run my melpa-stats script since before this change.  I ran it now and saw that the total download count for most of my packages has roughly doubled.  I don't know what to make of it, but it's interesting.  :)

PR #6667, opened 2020-01-31
Add recipe for `masm-mode' package

YiGeeker:

A major mode for editing MASM x86 and x64 assembly code.  It
includes syntax highlighting, automatic comment indentation and
various build commands.

https://github.com/YiGeeker/masm-mode

I'm the maintainer. This is my first Emacs package so I hope I did everything right.

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/masm-mode
+++ b/recipes/masm-mode
@@ -1,0 +1,3 @@
+(masm-mode :fetcher github
+            :url "https://github.com/YiGeeker/masm-mode"
+            :repo "YiGeeker/masm-mode")
--- a/recipes/masm-mode
+++ b/recipes/masm-mode
@@ -1,3 +1,1 @@
-(masm-mode :fetcher github
-            :url "https://github.com/YiGeeker/masm-mode"
-            :repo "YiGeeker/masm-mode")
+(masm-mode :fetcher github :repo "YiGeeker/masm-mode")


riscy:
Just a quick pass over the checklist -- also, you may want to add the GPL short-text to your front matter (i.e. "This program is free software; you can redistribute it and/or modify...")  just in case someone _only_ receives the file.

### masm-mode.el
```
masm-mode.el:542:1:Warning: Unused lexical argument `maj-mode'
masm-mode.el:542:1:Warning: Unused lexical argument `maj-mode'
masm-mode.el:542:1:Warning: Unused lexical argument `savep'
masm-mode.el:593:1:Warning: Unused lexical argument `maj-mode'
masm-mode.el:593:1:Warning: Unused lexical argument `maj-mode'
masm-mode.el:634:1:Warning: Unused lexical argument `maj-mode'
masm-mode.el:634:1:Warning: Unused lexical argument `maj-mode'
masm-mode.el:634:1:Warning: Unused lexical argument `savep'
```
These can be muted by prefixing with `_`.

```
Warning (emacs):
masm-mode.el:543: Argument `savep' should appear (as SAVEP) in the doc string
Warning (emacs):
masm-mode.el:594: Argument `command' should appear (as COMMAND) in the doc string
Warning (emacs):
masm-mode.el:635: Argument `savep' should appear (as SAVEP) in the doc string
```

```
2 issues found:
8:3: warning: You should include standard keywords: see the variable `finder-known-keywords'.
19:0: error: Customization groups should not end in "-mode" unless that name would conflict with their parent group.
```

- masm-mode.el#L554: There is no need to quote lambdas (neither #' nor ')
- masm-mode.el#L559: There is no need to quote lambdas (neither #' nor ')
- masm-mode.el#L576: There is no need to quote lambdas (neither #' nor ')
- masm-mode.el#L589: There is no need to quote lambdas (neither #' nor ')
- masm-mode.el#L600: There is no need to quote lambdas (neither #' nor ')
- masm-mode.el#L618: There is no need to quote lambdas (neither #' nor ')
- masm-mode.el#L631: There is no need to quote lambdas (neither #' nor ')
- masm-mode.el#L645: There is no need to quote lambdas (neither #' nor ')
- masm-mode.el#L660: There is no need to quote lambdas (neither #' nor ')
- masm-mode.el#L673: There is no need to quote lambdas (neither #' nor ')
- masm-mode.el#L570: It's safer to sharp-quote function names; use `#'`
- masm-mode.el#L572: It's safer to sharp-quote function names; use `#'`
- masm-mode.el#L583: It's safer to sharp-quote function names; use `#'`
- masm-mode.el#L585: It's safer to sharp-quote function names; use `#'`
- masm-mode.el#L612: It's safer to sharp-quote function names; use `#'`
- masm-mode.el#L614: It's safer to sharp-quote function names; use `#'`
- masm-mode.el#L625: It's safer to sharp-quote function names; use `#'`
- masm-mode.el#L627: It's safer to sharp-quote function names; use `#'`
- masm-mode.el#L654: It's safer to sharp-quote function names; use `#'`
- masm-mode.el#L656: It's safer to sharp-quote function names; use `#'`
- masm-mode.el#L667: It's safer to sharp-quote function names; use `#'`
- masm-mode.el#L669: It's safer to sharp-quote function names; use `#'`
- masm-mode.el#L699: Consider `unless ...` instead of `when (not ...)`

### Maintainer/license

- PR by YiGeeker: https://github.com/YiGeeker/masm-mode
- License found (through API): 'GNU General Public License v3.0' (good)
- LICENSE excerpt: 'GNU GENERAL PUBLIC LICENSE...'

### Summary

- (masm-mode :fetcher github :repo "YiGeeker/masm-mode") 
- _elisp/masm-mode.el (unknown license): MASM x86 and x64 assembly major mode
</details>

YiGeeker:
Thanks very much! I fixed up all the problems.

conao3:
Hi!

1. Please consider using line-{beginning, end}-position function without an actual moving point.
2. Please consider wrap `save-excursion` for sexp moving point.
3. The file is mixed index with tabs and spaces.
  Currently, indent with space is common. I unify indent with spaces.
  And you can add file local variable to ensure indent with spaces for the future.

riscy:
Thanks -- overall this looks good.

I am not sure about this line here:
```elisp
(add-hook 'after-change-major-mode-hook #'masm-mode-after)
```
This adds a global hook, so every other buffer that changes its major mode is also going to run `masm-mode-after`.  That doesn't seem like what you want --  `add-hook` has a LOCAL flag you can look into.

For patterns like this `(set (make-local-variable 'font-lock-defaults)` you can use `setq-local` instead of `set (make-local-variable`, as well, since you're using Emacs 25.1.

I believe this line is unnecessary:
```elisp
(use-local-map (nconc (make-sparse-keymap) masm-mode-map))
```
since the `define-derived-mode` macro should automatically invoke `use-local-map`.

This line:
```elisp
(add-to-list 'auto-mode-alist '("\\.asm\\'" . masm-mode))
```
will clobber Emacs's built-in `"\\.asm\\'" . asm-mode`, which is fine, but you might add a short documentation/notice about that.

Also for your consideration (non-blocking):
- masm-mode.el#L587: It's safer to sharp-quote function names; use `#'`
- masm-mode.el#L589: It's safer to sharp-quote function names; use `#'`

YiGeeker:
Thank you very much! I revised related code.

riscy:
This line:
```elisp
(add-hook 'after-change-major-mode-hook #'masm-mode-before :local t)
```
is actually passing the `:local` symbol in for what would actually be the APPEND parameter, so you probably want the line to look more like:
```elisp
(add-hook 'after-change-major-mode-hook #'masm-mode-before nil t)
```
(or if you do want to APPEND, use t instead of nil above).  I don't think that will actually create issues for you right now, but it might in the future if you're trying to debug the add-hook behavior.

Another small thing I notice as I take a look:
- The form beginning `(let ((process-environment (append (list...` in `masm-build` seems like it doesn't need to be making a call to `append`?

YiGeeker:
Thank you very much. I fixed the bug and passed t to the APPEND parameter. What I want is to make directory local variables take effect. If this line `(add-hook ...)` is deleted, directory local variables will be to late to change the mode behavior, for instance, the customized varible `masm-program-type` , changing key words between 32bits and 64bits. I don't know how to achieve this function alternatively, can you help me?
The LINK program relies on some Windows native environment variables, so a call to append is need. Not to influence environment variables of other buffers, I don't set these enviroment variables directly.

riscy:
Looks good --

> If this line (add-hook ...) is deleted, directory local variables will be to late to change the mode behavior ...

Okay, I see the issue.  I think this is fine, but you could add some comments for the next person to come along ("Delay this to a hook instead of running it immediately, due the order in which file local variables are processed." or whatever -- the "Make sure that file local variables work." docstring is sort of cryptic to me).

But I feel like this:
```elisp
    (if masm-program-type
        (setq-local font-lock-keywords masm-win64-font-lock-keywords)
      (setq-local font-lock-keywords masm-win32-font-lock-keywords))))
```
is buggy and should be doing a string comparison, since the options for that variable are:
```elisp
(defcustom masm-program-type t
  "Syntax for `masm-mode', t for Win64 and nil for Win32."
  :type '(choice (const :tag "Win64" t)
                 (const :tag "Win32" nil))
  :group 'masm-mode)
```

riscy:
I think neither of those issues are blockers on this being merged. :)  Welcome to MELPA!

YiGeeker:
Thanks for your advice and your enthusiasm, I updated the related code.

PR #6668, opened 2020-01-31
Add metronome recipe

jagrg:

A simple [metronome](https://en.wikipedia.org/wiki/Metronome) with accurate timing.

https://gitlab.com/jagrg/metronome

I'm the maintainer

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/metronome
+++ b/recipes/metronome
@@ -1,0 +1,1 @@
+(metronome :repo "jagrg/metronome" :fetcher gitlab)


riscy:
Just rolling through with a quick first pass over the checklist:

### metronome.el
```
metronome.el:72:1:Warning: Unused lexical variable `i'
In metronome-cancel-timers:
metronome.el:112:69:Warning: reference to free variable `fn'
In metronome-resume:
metronome.el:167:6:Warning: reference to free variable `last-bpm'
In end of data:
metronome.el:246:1:Warning: the following functions are not known to be
defined: when-let, fn, if-let, last-bpm
```
Looks like all of that is cleared up by adding `(require 'subr-x)` and renaming the `i` to `_i` (for example).

- No issues! 💯

- No issues! 💯

- metronome.el#L153: It's safer to sharp-quote function names; use `#'`
- metronome.el#L207: No `format` required; messages are already f-strings
- metronome.el#L217: No `format` required; messages are already f-strings
- metronome.el#L227: No `format` required; messages are already f-strings
- metronome.el#L21: Prefer `https` over `http` (if possible)

### Maintainer/license

- PR by jagrg: https://gitlab.com/jagrg/metronome

### Summary

- (metronome :repo "jagrg/metronome" :fetcher gitlab) 
- _elisp/metronome.el (gpl): A simple metronome with accurate timing
</details>

jagrg:
Thanks! The first warning is now fixed. What I don't understand is why
I'm not seeing the other warnings (I'm also using Emacs 26.3 BTW).
Anyhow those should be fixed too with your advice. Also, if you don't
mind me asking, is the sharp quote also needed for functions *defined*
in the Lisp file?

riscy:
Probably what's happening is you're compiling in an Emacs that has already loaded subr-x in, but there's no guarantee that will always be the case for others.

The sharp-quote isn't strictly needed, but it buys you compiler warnings when the function doesn't exist.  If you rename a function during refactoring, or you make a typo when you write its name, the sharp-quote forces the compiler to look for whatever you typed.

jagrg:
I see. Thanks for making those two points clear.

All done.

riscy:
Great!  Taking another look now.

Instead of a `defun` on `metronome-mode`, I'd strongly recommend using
```elisp
(define-derived-mode metronome-mode special "Metronome"
    "Major mode for displaying and controlling a metronome."
    (setq-local cursor-mode ...))
```
Using `define-derived-mode` means you don't have to `kill-local-variables` or `setq major-mode/mode-name` or even `use-local-map` -- all of that is handled implicitly.  I chose `special`-mode as the mode to derive off of, but in rare cases sometimes `fundamental-mode` is preferred (check the docs on each).

Also, I've recently been asked to be a bit more strict about the licenses on packages, so in light of that I'd ask you to also add a LICENSE file (or similar) to your repo -- one that appears in a format that the GitLab API can detect, for example, so that your license is automatically detectable.  The boilerplate you have in the elisp file should stay as is (and thanks for that).



- seconds ber beat -> seconds per beat
-

jagrg:
All done. Thanks again for the feedback.

conao3:
Hi!

```emacs-lisp
 metronom…    25     warning         There should be two spaces after a period (emacs-lisp-checkdoc)
 metronom…    26     warning         There should be two spaces after a period (emacs-lisp-checkdoc)
 metronom…    28     warning         There should be two spaces after a period (emacs-lisp-checkdoc)
 metronom…    36     warning         There should be two spaces after a period (emacs-lisp-checkdoc)
 metrono…    37     warning         There should be two spaces after a period (emacs-lisp-checkdoc)
 metrono…   112     warning         There should be two spaces after a period (emacs-lisp-checkdoc)
 metrono…   244     warning         There should be two spaces after a period (emacs-lisp-checkdoc)
```

alphapapa:
> The first warning is now fixed. What I don't understand is why I'm not seeing the other warnings (I'm also using Emacs 26.3 BTW).

FYI, this tool can help you spot errors that other users will see by testing your package in a clean, "sandboxed" Emacs config: https://github.com/alphapapa/makem.sh

riscy:
> Please resolve below package-lint/check-doc/byte-compiler warnings

@conao3 These are the very few nags that I actually mute in my setup, since so many people have `(sentence-end-double-space nil)` since it affects other editing behaviors, not just checkdoc, unfortunately.

This looks good to go @jagrg.  I'm glad we have finally filled the metronome gap in Emacs. 😎

PR #6670, opened 2020-02-01
Add weak-ref recipe

ebpa:

Exactly what it says on the box.  It contains a function for creating a single weak reference using a hash table with weak values and a function for dereferencing it.  Admittedly this is sugar, but standalone weak references are a feature missing from elisp and this package makes for better code.

https://github.com/skeeto/elisp-weak-ref

An enthusiastic user.

I exchanged emails with @skeeto asking if he minded my submitting it to Melpa.  He added the version header, version tag, and adjusted the name to follow the feature namespace.  Additional checkdoc docstring complaints were recently addressed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/weak-ref
+++ b/recipes/weak-ref
@@ -1,0 +1,1 @@
+(weak-ref :fetcher github :repo "skeeto/elisp-weak-ref")


ebpa:
Checkdoc docstring complaints have been addressed.

alphapapa:
It would be very helpful if there were some good examples of where it is useful.  The example in the readme shows how it works, but it does not show where would it be useful to have a variable's value disappear whenever Emacs happens to GC.

ebpa:
Thanks, @alphapapa.  That's a good point.  It might help to elaborate on the existing commentary to also note that the variable's value won't necessarily disappear after garbage collection the example might imply.  Maybe something like:

"Unlike the above demonstration, weak references are useful in circumstances where the weak reference isn't the only reference to an object.  The essential feature of weakness being that that reference doesn't _prevent_ garbage collection of an object.  For example, these types of references can be used to create caching structures that would otherwise create memory leaks."

Possibly at the beginning.  Maybe linking to make-hash-table in the docstring, https://www.gnu.org/software/emacs/manual/html_node/elisp/Creating-Hash.html, or https://wikipedia.org/wiki/Weak_reference.

I'll take a stab at this on Sunday.

alphapapa:
> "Unlike the above demonstration, weak references are useful in circumstances where the weak reference isn't the only reference to an object. The essential feature of weakness being that that reference doesn't _prevent_ garbage collection of an object. For example, these types of references can be used to create caching structures that would otherwise create memory leaks."

That's a good description, but I think some Emacs-specific examples are still called for.  I'm not an expert, but I'm trying to think of how this could be useful for a cache in Emacs, but I can't.  If there's also a non-weak reference to an object, it won't be GC'ed.  But if that non-weak reference goes away, the value could be GC'ed at any moment (Emacs GC's fairly often).  So it would seem like the only benefit would be a slim chance that another reference to the object would be created before it gets GC'ed.  It's virtually guaranteed that the object would be GC'ed before another reference would be created to it, in which case the weak reference wouldn't be useful.

Of course, Chris would not have created the library if he didn't think it would be useful, so I must be wrong.  So I really do want to see useful examples.  If I could use this to benefit my packages, I'd like to.  :)

skeeto:
That's a reasonable request, though unfortunately I don't think there's 
a good answer. I originally wrote this package back in 2012 as a proof 
on concept for this article:

Elisp Weak References
https://nullprogram.com/blog/2012/12/17/

It was "here's how this could be generalized" and not "here's how it's 
useful in practice." I mention Skewer, but I've never actually used weak 
references anywhere in Skewer. The "cache-table" relies on an explicit 
timeout.

Since the feature provided by the package is trivial, and I didn't think 
it was useful, I never bothered to get it onto MELPA. In 2016 Erik 
reached out to me about putting it on MELPA, which I said was fine. He 
must have had a use in mind, and it wouldn't really bother me. I cleaned 
it up a bit (though not enough), and Erik finally followed up here in 
2020.

In the four years since Erik and I first discussed adding this to MELPA, 
I've observed what a mess NPM has become. There left-pad, event-stream, 
is-odd, and other such issues resulting from an ecosystem built upon 
having hundreds or even thousands of transient, trivial dependencies. 
The average NPM user has no idea what all they're pulling in. Are these 
libraries trustworthy? Are they of questionable quality? How well are 
they maintained? (When I see my co-workers using NPM, these questions 
burn in my mind, and it frustrates me that they're not considering these 
issues.)

NPM is the worst case, but this attitude has spread to other package 
registries, including Cargo (Rust) and PyPI (Python). Even MELPA to some 
extent, such as date-at-point, flash-region, and hookify. The latter has 
two dependencies of its own, one of which it's not actually using and 
the other could be trivially removed. (Note: I'm not trying to pick on 
these packages in particular; they're just what I happened to find after 
a casual search. My own weak-ref package would go on this a list, too.)

I always cite Russ Cox's great article when talking about this issue:

Our Software Dependency Problem
https://research.swtch.com/deps

In short, adding a dependency should be done carefully and thoughtfully, 
and the decision should be routinely re-evaluated.

So, anyway, how are weak references useful? A lot of people — including 
naive me in that 2012 article — suggest using them for caches, but this 
is dead wrong. Weak references are not for caches, and *especially* not, 
as alphapapa pointed out, with Emacs' rudimentary garbage collector. 
Striking that from the list, what's left?

* Breaking reference count cycles. This is the primary use case for 
weak_ptr in C++. Emacs has a mark-and-sweep GC, so cycles don't matter.

* Canonicalization. This includes string and symbol interning, and this 
is probably the intended use-case for Emacs' weak hash tables. However, 
there's no reason to use my weak-ref package in a hash table. Just use 
the built-in weak reference it already has.

* Lapsed listeners. Say you're using the observer pattern, and observers 
subscribe to events. In Emacs these would be hooks and callbacks. You 
may want observers that are no longer in use to get garbage collected, 
but the listener list will keep them alive so long as they are 
subscribed. If the observed only uses weak references in its listener 
list, then it won't keep those objects alive.

So the only use case I can come up with is lapsed listeners. Still, it's 
not a good fit for Emacs. "Listeners" tend to be either function objects 
or symbols with a function binding: things that are funcall-able. But, 
in the former case, it's unusual to hold onto the listener function 
object outside the hook list anyway, especially if it's a closure. Emacs 
just doesn't really have the lapsed listener problem. And even if it 
did, you could just use a weak hash table to hold all the listeners 
rather than a literal list.

It feels weird that I'm essentially making an argument *against* adding 
my own package to MELPA, especially after I told Erik I'd help clean it 
up for prime time. Despite this being about my own package, I personally 
don't have a horse in this race. I gave up on package.el nearly three 
years ago — I've been meaning to write about this for some time now — so 
it really makes little difference to me. I'm fine with maintaining my 
packages in a package.el-friendly way so that they can live on MELPA, 
and the rest I leave up to you folks.

riscy:
Thanks @skeeto for the thoughtful reply.  @ebpa, maybe you have a specific use-case to add to the list?  MELPA isn't for indexing every elisp file that's ever been written, but I will say this seems like an elisp interesting artifact with value.

There are a couple items on the checklist to clean up in the meantime -- I think these are easy.

### weak-ref.el
- No issues! 💯

```
Warning (emacs):
weak-ref.el:0: You should have a section marked ";;; Commentary:"
```

```
3 issues found:
1:0: error: Package should have a ;;; Commentary section.
1:55: warning: You should depend on (emacs "24") if you need lexical-binding.
```

### Maintainer/license

- PR by ebpa: https://github.com/skeeto/elisp-weak-ref
  - NOTE: Repo and recipe owner don't match
- License found (through API): 'The Unlicense' (good)

### Possibly related
- https://melpa.org/#/org-re-reveal-ref
- https://melpa.org/#/org-ref

### Summary

- (weak-ref :fetcher github :repo "skeeto/elisp-weak-ref") 
- Package-Requires: n/a
- _elisp/weak-ref.el (unknown license): weak references for Emacs Lisp
</details>

alphapapa:
Thanks, Chris, I always learn something from you!  I look forward to your writing about how you manage your Emacs packages now.

ebpa:
That's quite a sales pitch, @skeeto!  I will of course defer to your wishes for the package and the Melpa team's judgement; expect no hard feelings here :-)  A few words of advocacy first-- for the general functionality afforded by weak references (since that's been brought up) and for the existence of the library.

@Skeeto, I'm very curious about why you consider weak references inappropriate for caches.  To my mind, weak references serve the goal of referring to an object without affecting that object's life cycle.  They are used in contexts that anticipate the lifetime of the referrer to be different than the referent.  Within a single library that controls the creation and references for its own objects, the use of weak references is very likely an anti-pattern.  However, it is also common that an object's life cycle is dictated by external logic not exposed via an API or not even well-defined.  The author of an extension library may thoughtfully consider the impact, for example, of annotating large objects by using a hash table.  Strong key references will extend the lifetime of their referents and affect the resource usage of the extended library.  A few manual strategies for maintaining this index (mitigating this memory leak) come to mind, but for establishing a policy, "Should key value pair (A, B) be pruned?" the simple criterion "Could A otherwise be freed?" is often correct and possible using a weakly-keyed map.

Unlike many functions that arrive in an arguably questionable position of being reified as a package, I feel that these two functions fill a specific and well-defined niche.  The generalized weak reference is a language feature missing from Emacs lisp that is found in a variety of other languages.  As such, I think it deserves some form of shared functional identity.

@riscy  The use case is any context you have a single referent / no natural index structure.  The generalized weak reference is certainly more obscure than weakness for hash tables, but I have at least personally found individual weak references useful in isolating different segments of logic (namely debug logic) operating on large objects.  It is incredibly convenient to be able to add a property, `:last-state`, to an object that points to its previous state and then be able to leave it there for your debugging convenience (and lazily avoid a massive memory leak) rather than going through the ceremony of maintaining an alternative.

alphapapa:
@ebpa would [finalizers](https://www.gnu.org/software/emacs/manual/html_node/elisp/Finalizer-Type.html) be relevant for any of these ideas?  I haven't used them before, and they're relatively new to Emacs.

ebpa:
> @ebpa would finalizers be relevant for any of these ideas?

Certainly neat (this is the first I've looked closely at the finalizer API), but: I don't think so- given that finalizer effects run after GC and only for GC'd finalizers.

skeeto:
Regarding caches: Weak reference are *too* weak. Just because the 
garbage collector runs does not mean the system is under actual memory 
pressure. Despite this, weakly-held memory will be freed anyway. For 
example, imagine parsing a million small XML documents one at a time. 
This creates absolute tons of garbage — all those conses and short 
strings — but only one, small document is ever held at a time. The 
garbage collector will run many times during this process, resetting 
every weak cache, but real memory usage remains low. (That isn't just 
some contrived example: Elfeed parses many small XML documents 
routinely, only ever holding one parsed document at a time.)

In the pathological case, Emacs might thrash on the cache. For instance, 
suppose you want to cache A, B, and C, and you keep accessing them in 
order regularly (A B C A B C …). The three don't fit in the current heap 
simultaneously, so when C is being loaded, both A and B are evicted. 
Then you need A, then B, but again, the heap is full, so A and C are 
evicted. If Emacs grew the heap just a little bit to fit all three 
simultaneously, it would be in a much better place, but the weak 
references prevent this from happening.

If Emacs had a generational garbage collector and the cache memory could 
be kept out of eden, or if it included, say, Java's concept of soft 
references, then it would be more reasonable for caching. A simple "we 
ran of out conses" wouldn't clear out all the caches, and they'd only be 
freed under actual (if light) memory pressure, when garbage collection 
is run more aggressively. In the thrashing case, the heap might be 
allowed to grow reasonably before the cache is wiped.

Even with better tools like that, I'd still avoid doing it with 
weak/soft references. Better to use a LRU cache tuned for the workload. 
With weak/soft references, either everything is wiped out, or some 
random subset is wiped out, neither of which are very useful. An LRU 
releases the least important things first.

dw-github-mirror:
> here left-pad, event-stream, is-odd, and other such issues resulting from an ecosystem built upon
having hundreds or even thousands of transient, trivial dependencies.
The average NPM user has no idea what all they're pulling in. Are these
libraries trustworthy? Are they of questionable quality? How well are
they maintained? 

An excellent point @skeeto!
>NPM is the worst case, but this attitude has spread to other package
registries, including Cargo (Rust) and PyPI (Python). 

While maybe not as extreme as NPM, I am tempted to add an older example to the list: (La)TeX with CTAN.   On top of there being spurious packages where **the package's author actively discourages people from using them** (hello, `xspace`) we have a plethora of nigh-identically named packages (`package`, `pkg`, `xpackage`, `package+`, `packagex`,  `xpkgx+` being likely released in late 2025). Half of are silently deprecated naturally, meaning if something suddenly breaks with a newer LaTeX version you eventually find some poor guy on stackoverflow with the same issue two years ago being told that package is long deprecated and that he should use and install `pkg2+` instead, of course using the argument `VER=1` for compatibility purposes.

riscy:
I wonder, @ebpa, if a lot of the discussion (and proposed use-cases, link to the article, etc.) could be added to the package's front-matter as well as its documentation? Maybe even adding a link to this PR to the package's front-matter would be a great way to give this package more context.

In my previous review I missed mentioning that the license boilerplate should be added above the `;;; Commentary` section -- also, this last package-lint nag:

```
1 issue found:
```
After those, I think we're good to go.

skeeto:
@riscy I've never noticed capitalized descriptions being the standard, so that linter warning seems like it's enforcing someone's personal style more than the community norms. In Emacs itself on master at the moment it's about break even: 847 starting lowercase and 880 starting uppercase. Uppercase has a slight, unfair advantage because descriptions that start with formal names (i.e. "PostScript") will be capitalized even when otherwise it would not be.

Maybe this is a newer trend? I did some wrangling with `git blame` to extract the first line of every .el file in Emacs with a standard header, sorted by the time that line was last changed, and computed a running sum of `lower - upper`. Here's what that looks like:

plot

For the overwhelming majority of Emacs' history, lower case descriptions was the standard. There was a significant shift towards capitalization in 2009 which I didn't investigate. There was also another major shift in mid 2017. This was due to the merging of Org v9.0.9 in emacs-mirror/emacs@5cecd275820. Org switched to capitalized style descriptions, tipping the balance towards that style for the first time (perhaps as a direct result of that linter warning?)

> the license boilerplate should be added above the `;;; Commentary` section

In Emacs itself, all public domain dedications are above the package variable block ([example](https://github.com/emacs-mirror/emacs/blob/f633e014aca727b9ccecaf8e6d69386f93c5073f/lisp/mail/feedmail.el#L3)).

riscy:
> I've never noticed capitalized descriptions being the standard, so that linter warning seems like it's enforcing someone's personal style more than the community norms.

Thanks for the stats -- that is interesting.  I believe the capitalization check relatively new to package-lint, and exceptions to the checklist are sometimes made.  But if everyone asks for exceptions, then the MELPA reviewers have to spend more time negotiating and less time looking at code.  So please consider it. :)

> In Emacs itself, all public domain dedications are above the package variable block

This was my oversight, I didn't realize "This is free and unencumbered software released into the public domain." /was/ the boilerplate.

alphapapa:
@skeeto That kind of glorious pedantry with data and graphs makes me smile.  :D  Would you mind publishing the code, too?  It could be very instructive on how to compile similar git-spelunking stats in the future.

skeeto:
Throw this in a script named `blame.sh`: 

```sh
#!/bin/sh
set -e
find -name '*.el' -print0 | \
    xargs -0 -P$(nproc) -n1 git blame -p -L 1,1 | \
    awk '/^committer-time/{t = $2} /^\t/{print t, $0}' | \
    LC_ALL=C sort -n | \
    awk '/ ---? [a-z]/{print $1, "+1"} / ---? [A-Z]/{print $1, "-1"}'
```

Note: It requires GNU Coreutils. The script will run multiple `git blame` in parallel using all available cores. Writing the parallel output into the same pipe like that is a little dubious and racy, [but it's probably safe in this case](https://nullprogram.com/blog/2016/08/03/). Always use `LC_ALL=C` with GNU sort [to avoid braindamaged behavior](https://utcc.utoronto.ca/~cks/space/blog/linux/BashLocaleScriptDestruction). Run that script at root of the Emacs repository checked out at the commit where you'd like to do the analysis (takes about 6 minutes):

    $ ./blame.sh >data

Then put this source in `blame.m`:

```octave
#!/usr/bin/env octave

data = dlmread('data');
years = (data(:, 1) - 694224000) / (365*24*60*60) + 1992; 
totals = cumsum(data(:, 2));

plot([1991 2021], [0, 0], 'r-');
hold('on');
plot(years, totals, 'linewidth', 2);

xlim([1992 2020.3]);
grid('minor');
title('Capitalized Descriptions');
ylabel('lower - upper');
xlabel('year');

print('-r150', 'plot.png');
```

Date handling is a little rough and doesn't account for leap days, but it's probably off by less than a pixel. Run it to produce `plot.png`:

    $ ./blame.m

alphapapa:
Brilliant!  Thanks, Chris.

ebpa:
> Just because the garbage collector runs does not mean the system is under actual memory pressure. Despite this, weakly-held memory will be freed anyway.

I wholly agree.  Mindful use of weak references should assume the garbage collector will run frequently and at any possible inopportune moment.  No logic should ever assume that GC _won't run_.

> In the pathological case, Emacs might thrash on the cache. For instance, suppose you want to cache A, B, and C, and you keep accessing them in order regularly (A B C A B C …). The three don't fit in the current heap simultaneously, so when C is being loaded, both A and B are evicted.

I didn't mean to suggest that weak maps are a replacement for general caching algorithms (although now my mind is stuck on perversely combining the two; thanks!).  My main point was that a weak map is a good data structure to use when you want a cache and don't want to break the memory management of some existing code that is outside your control.  It has the added benefit that the cache can rely on that external memory management for pruning.

> https://nullprogram.com/blog/2016/08/03/

_Very cool!_

> I wonder, @ebpa, if a lot of the discussion (and proposed use-cases, link to the article, etc.) could be added to the package's front-matter as well as its documentation? Maybe even adding a link to this PR to the package's front-matter would be a great way to give this package more context.

@riscy, Good idea to include a link to this thread.  The most concrete demonstrable example that comes to mind is a weakly-linked list like I mentioned.  The key is that *something else* has a well-defined lifecycle, and I'm racking by brain thinking of a meaningful example that doesn't rely on GC not running mid-example.  Maybe someone has some thoughts?

alphapapa:
FYI, a new [package](https://github.com/cireu/emacs-cell) relevant to this has been [proposed](https://lists.gnu.org/archive/html/emacs-devel/2020-02/msg00289.html) by @cireu on the emacs-devel list.

riscy:
Thanks all for putting up with my pedantry while you have an interesting discussion.  @ebpa this is good to go.

cireu:
Looks that Weak reference facility is really important for Elisp programmer :joy:

Fuco1:
I have another use-case for this.

In [Elsa](https://github.com/emacs-elsa/Elsa), I build a concrete syntax tree for the buffer with forms of various kinds, among them lists.  A list form holds buch of other forms which all have a parent reference back to the list.  This creates a loop of the list form referencing child forms each of them referencing the parent back => this will never be collected.

The parent reference can be broken with a weak ref, because I analyze "one defun at a time", the moment I move to the next one, there is no reference to the parent form and this will cause the entire tree to be collected.

I kind of ignored it but now with huge projects Elsa can consume gigabytes of memory which is not great.  I suspect this might be one of the issues...

Here's an example:

```elisp
;; create a float form
(defvar my-float (elsa-form-float :value 1))

;; let-bind a list form so it goes out of scope
(let ((lst (elsa-form-list :sequence (list my-float))))
  ;; weakly refer it as a parent on the float
  (oset my-float parent (weak-ref lst)))

;; eval this and check parent

(weak-ref-deref (oref my-float parent)) ;; => it's there
(garbage-collect)
(weak-ref-deref (oref my-float parent)) ;; => it's gone

;; ==================

;; now a demonstration that the loop is not collected.  Create a weak
;; reference to the float so this won't prevent GC.
(defvar my-weak-float (weak-ref (elsa-form-float :value 1)))

;; deref it here so it is a real reference
(let ((lst (elsa-form-list :sequence (list (weak-ref-deref my-weak-float)))))
  ;; add a real reference
  (oset my-float parent lst))

;; check if the float form exists
(weak-ref-deref my-weak-float) ;; exists
(garbage-collect)
(weak-ref-deref my-weak-float) ;; still exists, because I've created a
                               ;; self-reference loop
```

PR #6678, opened 2020-02-04
Add versuri recipe

mihaiolteanu:

The lyrics package.

https://github.com/mihaiolteanu/versuri

Author and maintainer.

None needed.

### Checklist

Please confirm with `x`:

------
### package-lint [solved]
My `package-lint` says stuff like,
```
7:39: warning: Version dependency for dash appears too high: try 2.12.0
7:55: error: Package request is not installable.
7:73: error: Package anaphora is not installable.
7:92: error: Package elquery is not installable.
7:110: error: Package s is not installable.
```
I see these kinds of warnings and errors for other MELPA packages as well. Can this be ignored?

### elquery [solved]
I've defined my own `versuri--elquery-read-string` function, similar to `elquery-read-string`. The difference is that it accepts multibyte characters and doesn't remove newlines from the parsed html response. Removing newlines creates chaos in the lyrics. The problem is, it uses a private function, `elquery--parse-libxml-tree`. I've created issues on the `elquery` page, but no response given: [AdamNiederer/elquery/issues](https://github.com/AdamNiederer/elquery/issues).

### lyrics.el
There is a lyrics.el package already. The last promise from its maintainer to move things forward for this package was two years ago: [emacs-pe/lyrics.el/issues/4](https://github.com/emacs-pe/lyrics.el/issues/4). The package seems unfinished/incomplete.

### DIFF
--- a/recipes/versuri
+++ b/recipes/versuri
@@ -1,0 +1,1 @@
+(versuri :fetcher github :repo "mihaiolteanu/versuri")


alphapapa:
> I see these kinds of warnings and errors for other MELPA packages as well. Can this be ignored?

This probably indicates a problem in your personal Emacs config.  You may want to look into that.  In the meantime, you should test your package outside of your Emacs config, e.g. with https://github.com/alphapapa/makem.sh or the MELPA `make sandbox` tool.

Regarding `elquery`, you might consider using an alternative, like `esxml`.  See https://github.com/alphapapa/emacs-package-dev-handbook#xml--html

> There is a lyrics.el package already. The last promise from its maintainer to move things forward for this package was two years ago: emacs-pe/lyrics.el/issues/4. The package seems unfinished/incomplete.

To be clear, is your package based on that package?  Did you copy any code from it?  I don't see any mention of it in your readme or commentary.

mihaiolteanu:
> or the MELPA make sandbox tool

I've started Emacs with `./emacs-sandbox.sh` and called `make sandbox INSTALL=<name>` via `shell-command`, where `<name>` is the path to this package file. I get the same warnings and errors as above from `package-lint-current-buffer`.

If I `list-packages`, for example, and **then** call `package-lint-current-buffer` again, I get `No issues found.` (same behavior in both sandboxed Emacs and normal usage).

> you might consider using an alternative, like esxml

I might try that, if I can mend it to my needs. `elquery` is still mentioned as the package to read and manipulate html on the page you provided.

> To be clear, is your package based on that package? Did you copy any code from it?

No, it is not. Nothing copied from there or inspired from that package. Just mentioning it as this package, versuri, is now implementing functionality that already exists, albeit in an unfinished form.

mihaiolteanu:
> you might consider using an alternative, like esxml

I've replaced `elquery` with `esxml`. Works like a charm and it seems easier to use, too. Thanks for your suggestion.

riscy:
### versuri.el
```
versuri.el:48:1:Error: Cannot open load file: No such file or directory, esqlite
```

- No issues! 💯

- No issues! 💯

- versuri.el#L382: No `format` required; messages are already f-strings
- versuri.el#L22: Prefer `https` over `http` (if possible)
- versuri.el#L215: Prefer `https` over `http` (if possible)
- versuri.el#L219: Prefer `https` over `http` (if possible)

License:
- Please add a [GitHub-compatible](https://github.com/licensee/licensee) LICENSE or COPYING file to your repository
- gpl license text appears in individual files (good)

mihaiolteanu:
All fixed.

riscy:
```
Package `esxml-query-' is unavailable
```
I think the package is just called `esxml`

conao3:
Hi!

3. Please indent the whole file because of some of the places broken indent.
4. Please use `expand-file-name` instead of `concat` to create file path.

Note:
  5. Do you really want to save `versuri.db` in `xdg-config-home`?
    Common Elisp packages save files in `user-emacs-directory` like `~/.emacs.d/` if needed.
    For that case, you can use `locate-user-emacs-file` to create file path.

alphapapa:
> Please add Cask file is good to easy resolve dependency, build, packaging etc.

Cask is a good tool, but I don't think we should expect people who don't use it to add a `Cask` file, which would require studying the Cask documentation, etc.

> Do you really want to save `versuri.db` in `xdg-config-home`?
> Common Elisp packages save files in `user-emacs-directory` like `~/.emacs.d/` if needed.
> For that case, you can use `locate-user-emacs-file` to create file path.

It would probably be best to store it in the cache directory, e.g. `~/.cache` (I assume it's `xdg-cache-home` in Emacs's XDG support).  It's not configuration, and its a cache of remote data, right?

Personally, I kind of hate it when packages store non-configuration data in `~/.emacs.d` and clutter my backups and my git status with cache data.  :)

conao3:
> Cask is a good tool, but I don't think we should expect people who don't use it to add a Cask file, which would require studying the Cask documentation, etc.

cask is worth studying.
The `require` error will also be resolved by managing the package with cask and installing `flycheck-cask`.
And I need it to build this package.
Ideally, I need a Makefile like you develop, but I can use `cask install` and` cask build` just by having a `Cask` file, which is a minimum requirement.

> Personally, I kind of hate it when packages store non-configuration data in ~/.emacs.d and clutter my backups and my git status with cache data. :)

Of course, I agree.
However, there are so many packages that save their own files in `.emacs.d`, which is a good place to store the files for the emacs package.
There are many packages that take this strategy will be obvious if you look at the [no-littering](https://github.com/emacscollective/no-littering) package.
I'll add my personal opinion, but this way the `.emacs.d` folder is makes dirty and dirty and should be  managed by `git` with whitelisted management.
Any package could store a file with sensitive information, accidentally commit it with `git`, and send it to remote.

alphapapa:
Cask is a large, complicated Python package.  We shouldn't encourage people to use it for every little Emacs package.  `package-lint` and `./makem.sh lint-compile` are simple ways to ensure that the proper features are `require`d.

> I'll add my personal opinion, but this way the `.emacs.d` folder is makes dirty and dirty and should be managed by `git` with whitelisted management.
> Any package could store a file with sensitive information, accidentally commit it with `git`, and send it to remote.

I don't do `git add .` in my `~/.emacs.d`.  But the extra files clutter up `git status`, and I have to ignore them manually.

We should encourage new packages to not store cache data in `~/.emacs.d`.

conao3:
> Cask is a large, complicated Python package.

Really? https://github.com/cask/cask/search?l=python
The python file is only 3 files and over 85% of LOC is Emacs-lisp in cask.

> We should encourage new packages to not store cache data in ~/.emacs.d.
I agree. `~/.cache` is good as you have commented for such a cache file.

mihaiolteanu:
> I think the package is just called esxml

I'm not sure how it's called anymore. Package-lint was not happy with `esxml` either. I've changed it a few times between the two [1].

I don't have access to a pc right now and I won't have in the next few days due to some personal problem. Closing this issue and the `vuiet` [2] recipe will take a while. Sorry to all the people involved for the inconvenience.

[1] https://github.com/mihaiolteanu/versuri/commit/114e407877846d13a3e543692831d6d64cb3f357
[2] https://github.com/melpa/melpa/pull/6684#issuecomment-584301009

riscy:
> I'm not sure how it's called anymore. Package-lint was not happy with esxml either. I've changed it a few times between the two [1].

Try this for your `Package-Requires`:
```elisp
;; Package-Requires: ((emacs "26.1") (dash "2.16.0") (request "0.3.0") (anaphora "1.0.4") (esxml "0.1.0") (s "1.12.0") (esqlite "0.3.1") (ivy "0.11.0"))
```
And then `(require 'esxml-query)` can stay intact (esxml installs exsml-query).

mihaiolteanu:
Fixed the `esxml` issue (thanks Chris).

What is the consensus for storing the db file? `(xdg-cache-home)`, or?

mihaiolteanu:
Hellooo! :)

riscy:
Apologies, there's quite a backlog. :)

riscy:
Looks impeccable. :)

How are you feeling about SQL injectability?  For example, I don't know exactly what this will do, but maybe this is something to consider:
```elisp
(let ((artist "riscy")
      (song "melpa")
      (lyrics "hello\"); DELETE from lyrics..."))
  (format "INSERT INTO lyrics(artist,song,lyrics) VALUES(\"%s\", \"%s\", \"%s\")"
          artist song (s-trim lyrics)))
```
The users seem to be fully in control of these values, so the risk is low.  But you could have songs or lyrics with quotation marks in them that could throw a wrench in the works.

One solution might be to replace every `"` in the string with a `""`, which is how quotation marks are safely escaped in sqlite databases I believe.

mihaiolteanu:
> ```emacs-lisp
> (lyrics "hello\"); DELETE from lyrics..."))
> ```

Those are some...funky lyrics! What music are you listening to, Chris? :))
Jokes aside, I agree, it makes no harm to escape the quotes. Done!

And yes, I can see you're pretty much the only maintainer who answers around here. At least to the PRs that I've been following. That's a lot of work, so no need for "apologies", I was just trying to make sure you're not actually waiting for my input on something.

riscy:
I feel like you want to quote the artist, too, and the lyrics.  I have at least one recent listen with quotation marks in the artist name (Jukio "Kozilek" Kallio) which could affect this:
```elisp
(defun versuri--db-save-lyrics (artist song lyrics)
  "Save the LYRICS for ARTIST and SONG in the database."
  (esqlite-stream-execute versuri--db-stream
   (format "INSERT INTO lyrics(artist,song,lyrics) VALUES(\"%s\", \"%s\", \"%s\")"
           artist song
           (esqlite-escape-string (s-trim lyrics) ?\"))))
```

mihaiolteanu:
No harm in escaping the artist and song too. Done!

riscy:
Great, thanks!

PR #6680, opened 2020-02-06
remove recipe for readability

zzkt:
The `readability` package provides an interface to a now defunct service. According to [Wikipedia](https://en.wikipedia.org/wiki/Readability_(service)) it closed down in 2016. The package is available in the [orphanage](https://github.com/emacsorphanage/readability) and appears abandoned.

### DIFF
--- a/recipes/readability
+++ b/recipes/readability
@@ -1,1 +1,0 @@
-(readability :fetcher github :repo "emacsorphanage/readability")


alphapapa:
Note as well that Emacs's EWW browser has a built-in `eww-readability` function that implements similar functionality natively.  It's used in `org-web-tools`.

conao3:
I am the maintainer of `emacsorphanage/readability`.
I agree with this conclusion, and if you consulted me, I would have said yes. But the flow was not appropriate.
@zzkt should raise the issue at the `emacsorphanage/readability` first.

Also, [emaccessorphanage/readability](https://github.com/emacsorphanage/readability) is now redirected to [emaccesstic/readability](https://github.com/emacsattic/readability).
As for the external behavior, repository transfer seems to have been executed.
I don't know the relationship between emaccattic and emacsorphanage. What was the purpose of emaccattic, @tarsius?

zzkt:
@conao3 what do you consider the appropriate flow for removing an obsolete package from MELPA?

conao3:
Already commented.
> I agree with this conclusion, and if you consulted me, I would have said yes. But the flow was not appropriate.
> @zzkt should raise the issue at the `emacsorphanage/readability` first.

PR #6681, opened 2020-02-06
Include changelog file in treemacs recipe.

Alexander-Miller:

The package is already on melpa. This change adds the changelog to the list of files included in the package.

https://github.com/Alexander-Miller/treemacs/
https://github.com/Alexander-Miller/treemacs/blob/master/Changelog.org

Maintainer

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/treemacs
+++ b/recipes/treemacs
@@ -2,6 +2,7 @@
  :fetcher github
  :repo "Alexander-Miller/treemacs"
  :files (:defaults
+         "Changelog.org"
          "icons"
          "src/elisp/treemacs*.el"
          "src/scripts/treemacs*.py"


tarsius:
@purcell Do we have a policy for this? Something like "we don't include CHANGELOG, AUTHOR etc. in the package, but if the maintainer submits a pull-request to add such a file, then we add it anyway" maybe?

Also maybe we should add extend the pull-request template with "This template is for adding a package to Melpa. If you merely modify an existing recipe, then you can you may wish to write a free-form description of your reasoning. Even in this case, we appreciate a link to the repository."

tarsius:
@purcell commented on a similar request: https://github.com/melpa/melpa/pull/6658#issuecomment-578540202.

Alexander-Miller:
> because it ends up in an obscure package installation directory that the user is not going to browse to when they need information.

That is not the point. An interactive function to open the changelog file has already been written and published long ago. It does of course needs an actual changelog to show, hence this PR.
I figured showing an org file instead of opening github is the more Emacs-y solution, and the bundled changelog would be synchronized with the version that's actually installed.

purcell:
> Do we have a policy for this? Something like "we don't include CHANGELOG, AUTHOR etc. in the package, but if the maintainer submits a pull-request to add such a file, then we add it anyway" maybe?

The (unwritten) policy is pretty much: don't include non-`info` documentation files in the package *unless*...

> An interactive function to open the changelog file has already been written and published long ago. It does of course needs an actual changelog to show, hence this PR.

... the package provides commands to show them. :-)

Alexander-Miller:
:+1: 

> The (unwritten) policy is pretty much:

I can open a PR to get that written down if you haven't started already.

purcell:
> I can open a PR to get that written down if you haven't started already.

Sure, thanks. There's only `CONTRIBUTING.md` for now, not a full policy manual. I'd want to stop short of actively implying that people should add such commands if they don't already have them: I think the ideal outcome for end users is if everything is in `info` format, impractical as that can sometimes be for maintainers.

alphapapa:
> I think the ideal outcome for end users is if everything is in `info` format, impractical as that can sometimes be for maintainers.

It's pretty easy to export Org files to info manuals, e.g. with an `after-save-hook` in a README.org file.  Maybe we should document that.

PR #6682, opened 2020-02-06
add somafm recipe

artenator:

`somafm.el` is a simple interface to [soma.fm](https://somafm.com) radio stream channels. It allows you to easily browse the available channels and stream them from the comfort of emacs. The overall goal is to eventually have all of the functionality and aesthetic appeal of the [soma.fm popup player](https://somafm.com/player) but from within emacs.

https://github.com/artenator/somafm.el

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/somafm
+++ b/recipes/somafm
@@ -1,0 +1,1 @@
+(somafm :fetcher github :repo "artenator/somafm.el")


mihaiolteanu:
Nice! I've installed it and it works nicely.

I do see you have very very long lines in your .el file. I think the standard is to have no more than 80 characters per line. This is especially true for comment strings, where the requirement is actually even stricter[1].

I'm not sure if you need the `somafm-pkg.el` file in your repository. This file is generated by the package manager when the package is installed.

If I hit <return> on the channel that is already playing, it stops and then starts again with a noticeable silent moment. Maybe don't do anything in such a scenario, or maybe toggle the play/pause status of the player?!

The somafm-mode is really nice and clean and easy to use. For my own taste I would also like a more streamlined approach, like

```
(ivy-read "Pick channel: "
          somafm-current-channel-order
          :action
          (lambda (c)
            (message (format "You picked %s" c))))
```
similar to what you have with counsel or helm. The user can choose whatever channel he wants without visiting additional buffers

screenshot1581059513

Again, nothing wrong with your approach, it's just a wild idea.

You can also maybe use `mode-line-misc-info` to put the currently playing song and/or station in the mode-line[2]. This would help to get somafm-mode out of the way, as you can choose what to play and can see what is playing from any other buffer.

That's some really relaxing music, by the way :)

[1] [Documentation Tips](https://www.gnu.org/software/emacs/manual/html_node/elisp/Documentation-Tips.html)
[2] [How to append strings to the mode-line](https://emacs.stackexchange.com/questions/13855/how-to-append-string-that-gets-updated-to-mode-line)

kidd:
@mihaiolteanu, maybe https://github.com/kidd/radios feels more like what you want (although it's a totally different approach codewise)

artenator:
@mihaiolteanu 
Thanks for checking it out, and for the great feedback!

I have gone ahead and added the function `somafm-by-completion`, which allows you to pick a channel to listen to the way you have described via `completing-read`.
image

I have also fixed the bug that causes the silence when picking a channel that is already playing.

I have also fixed my long lines.

Let me know what you think! 😎 

I haven't gotten to the `mode-line-misc-info` quite yet, but I plan to experiment with it and see what it looks like.

conao3:
Hi!

2. Maybe use `make-sparse-keymap` is common than `make-keymap`.
  Please search `make-sparse-keymap` in `simple.el.gz` in Emacs source files. There are many usage of `make-sparse-keymap` to define such a keymap
3. Most places indent with spaces, but one of place indent with tab.
  Please fix it and you can use file local variable to ensure to indent with space in the future.

artenator:
@conao3 

Thanks for the feedback and contribution! I have merged in your PR that contained these fixes.

artenator:
@mihaiolteanu 
I have implemented your suggestion to add the current song to `mode-line-misc-info`

Here's what it looks like:
image

mihaiolteanu:
Nice! I've tried your changes. 
Enter does not do anything now if the station is already playing. 
`somafm-by-completion` lets you pick a radio channel without visiting any buffers
The mode-line shows the current playing song.

I, for one, like it better than before. The only question is if you like it better. I think you should strive to make your player the way you like it and use it. The others can suggest things but you can, of course, ignore them if you will.

Again, for me, this is an improvement.

mihaiolteanu:
Oh, wait, wait. If you do decide to leave the `somafm-by-completion` the way it is right now, I think you will also need `somafm-pause` and `somafm-stop` commands. The user would probably want to bind some keys to those.

riscy:
Thanks all for the awesome feedback above.  As I'm looking this over, it looks ready to go to me.  @artenator -- welcome to MELPA! :)

PR #6684, opened 2020-02-07
Add vuiet recipe

mihaiolteanu:

The music player and explorer for Emacs.

https://github.com/mihaiolteanu/vuiet

Author and maintainer

None needed.

### Checklist

Please confirm with `x`:

### Additional Info
This package depends on [versuri](https://github.com/mihaiolteanu/versuri) for which there is an open [PR](https://github.com/melpa/melpa/pull/6678) for adding it to MELPA.

### DIFF
--- a/recipes/vuiet
+++ b/recipes/vuiet
@@ -1,0 +1,1 @@
+(vuiet :fetcher github :repo "mihaiolteanu/vuiet")


conao3:
Hi!

3. Please indent the whole file because of some of the places broken indent.
4. Please use `define-key` instead of `bind-keys`.

mihaiolteanu:
No comments on this yet?

riscy:
> No comments on this yet?

Just waiting for this before I can run tests:

> This package depends on versuri for which there is an open PR for adding it to MELPA.

riscy:
It's strange, I get this error when I try to byte compile:

```
vuiet.el:33:1:Error: Opening output file: No such file or directory, /workspace/.config/.lastfmrc
```
which I think is because the lastfm package [writes to a config file](https://github.com/mihaiolteanu/lastfm.el/blob/cff7699343467b36ff3dad90061b567025ff9b23/lastfm.el#L75) when it's loaded, but the directory may not even exist.

(Incidentally this doesn't stop the lastfm package from being _compiled_, just from being loaded when the directory doesn't exist.)

If I create the required directory, I get this:
```
vuiet.el:34:1:Error: esqlite: "sqlite3 not found"
```
which is because versuri does something similar by (a) depending on the directory existing and (b) [attempting to run an sqlite command](https://github.com/mihaiolteanu/versuri/blob/master/versuri.el#L53) immediately when the package is imported.

Would you consider fixing these?  One option would be to try to delay running these commands until they need to be ran, for example by turning `versuri--db-stream` into a function that returns a value, instead of a constant?

mihaiolteanu:
Hello Chris,

I've fixed both `lastfm` and `versuri` to not call any function during loading. I'm not sure why you're getting those errors or how these changes fixes them, but let's see.

I was under the impression that `xdg-config-home` writes to your `/home/username/.config` folder and not to `/workspace`. And shouldn't `.config` be available in every username folder?

The `sqlite3 not found` error looks to me like a `you don't have sqlite3 installed on your machine` error.

Thanks again for your input.

riscy:
Hey!  I was getting those errors by trying to 'require' the packages on a machine that wasn't prepared for them (i.e. didn't have a .config directory, didn't have sqlite3 installed as you mentioned).

I checked out the changes you made to lastfm and I think this raises a warning:

```
lastfm.el:106:27:Warning: reference to free variable `lastfm--config-file'
In lastfm-generate-session-key:
lastfm.el:147:25:Warning: reference to free variable `lastfm--config-file'
```
since it's a function instead of a variable.

That leads to an error since any package that `require`s on lastfm right now, or anyone that loads the package, will see an error due to this line at the top level:
```elisp
(lastfm--set-config-parameters)         ;set params on start-up.
```

Ideally you don't want any package attempting to change Emacs's state when it's merely loaded -- is it possible to delay that set-config-parameters line until the package is being used?

mihaiolteanu:
I've changed lastfm, I hope successfully, to not eval anything when loaded.

riscy:
Awesome, lastfm looks good now.  I guess the big benefit is that now it can be imported by other packages, and those packages automatically tested/compiled, without any setup required.

Since you use `bind-keys` it looks like you have to `(require 'bind-key)` and also add it to your   `Package-Requires` to support Emacs 26.  But that `bind-keys` call will also overwrite a user's keybindings if they reload vuiet, so the safest way to define keybindings that won't accidentally get overwritten would  be something like:

```elisp
(defvar vuiet-mode-map
    (bind-keys :map keymap
               ("C-m" . org-open-at-point)
               ("q"   . kill-current-buffer)
               ("j"   . next-line)
               ("k"   . previous-line)
               ("l"   . forward-char)
               ("h"   . backward-char))
    keymap))
```

mihaiolteanu:
> will also overwrite a user's keybindings if they reload vuiet

I see your point. Fixed.


> you have to `(require 'bind-key)` and also add it to your `Package-Requires`

Oups. Fixed as well.

riscy:
Great!  Thanks for your patience. :)

PR #6687, opened 2020-02-09
Add counsel-web

mnewt:

Dynamic web search and suggestions using counsel

https://github.com/mnewt/counsel-web

author/maintainer

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/counsel-web
+++ b/recipes/counsel-web
@@ -1,0 +1,1 @@
+(counsel-web :repo "mnewt/counsel-web" :fetcher github)


riscy:
Hi, thanks for this.  This is only a quick first pass.

### counsel-web.el
```
counsel-web.el:338:1:Warning: the following functions are not known to be
defined: browse-url-default-browser, json-read
```

- No issues! 💯

- No issues! 💯

- counsel-web.el#L25: Prefer `https` over `http` (if possible)

License:
- GitHub API found `Other`
  - Please use a [GitHub-compatible](https://github.com/licensee/licensee) format for your license file, e.g. no markup
- LICENSE.md excerpt: `# GNU GENERAL PUBLIC LICENSE...`
- gpl license text appears in individual files

mnewt:
@riscy Thanks for the feedback! I think I squashed all of those issues:

* Added requires for unknown functions
* https link to GPL
* Github compatible license file

Tested on Emacs 25.1, 26.3, and 27

conao3:
Hi!

1. Please add `Cask` file is good to easy resolve dependency, build, packaging etc.
2. Please add `.gitignore` file is good to ignore git stage `*.elc` files and `.cask` folder.
3. Please remove top-level sexp.
  All packages should not have side-effects just `require` package for user-experience.
  But as I don't know how reset `ivy-add-actions`, I cannot prepare `*-teardown` function.

mnewt:
@conao3 All great suggestions, thank you for your help. Merged with some changes.

riscy:
### counsel-web.el
- No issues! 💯

- No issues! 💯

```
1 issue found:
35:12: warning: This file is not in the `cl-lib' ELPA compatibility package: require `cl-lib' instead.
```

License:
- GitHub API found `GNU General Public License v3.0`
- counsel-web.el has GPL license text

---

I notice I have installed a slightly old version of ivy (also marked `Version: 0.13.0`) in which the interface to `ivy--recompute-index` takes 3 arguments (not the two you use).

The two dashes `--` indicate that this function is not part of ivy's public interface, so relying on its interface to never change, or for it to exist at all, is risky (I think ivy's private interface changes relatively frequently).  A similar argument can be made for using `ivy--set-candidates`; is there any way you can use functions from the public interfaces of counsel/ivy?

mnewt:
@riscy You have a good point about using the private API, and I too have noticed it often changes.

At first I thought I had to use the private API to accomplish this. But reading more in `counsel.el` makes it look like it is possible to do this using all (or at least mostly) public functions. See https://github.com/abo-abo/swiper/blob/82331ed4accaa032419257e659926584beb76261/counsel.el#L6447 for an example.

I'll work on that and get back to you.

mnewt:
Turns out it was easy. There is one function to call that replaces all of that internal API usage: `ivy-update-candidates`. 

But. It is a very new function so now `counsel-web` will only work with `ivy` from the last four months.

https://github.com/abo-abo/swiper/commit/881cbc516a7355226731f6772b5284fd12c55c1c

Suggestions? Should I write a shim?

riscy:
> Suggestions? Should I write a shim?

If you're okay with being a bit "bleeding edge" I think it's best to just rely on the (recent) version of ivy.  The shim would still have to reference private functions and I think be prone to breakage.

mnewt:
OK, `ivy-update-candidates` is part of the 0.13.0 release so I think it's fine to just require that version and leave it at that. 

I think it's ready. Unless of course you have any more issues or suggestions.

@riscy Thank you for all your help!

mihaiolteanu:
Hello Matthew. Sorry for actually checking `counsel-web` so late (I'm molteanu on reddit).

Maybe some users do not really care about using eww and just want to use their browsers?
The feature that I would care about is `counsel-web-suggest` with a default action of `browse-url-default-browser`. No showing the urls inside the minibuffer, no nothing.

I also see there is a customization for "counsel web search alternate action" that lets you browse the url in your default browser, but you can't do that as a default action. Why not? Wouldn't it be useful to search a word or phrase directly in your favorite browser?

In short, I, as a user, would only care for search words and not for links and pages. Others might have other needs, I agree, so don't take this as a bug. In the end, if the package does not do what one wants it to do, maybe that package is not suitable for that person.

mnewt:
@mihaiolteanu Thank you for spending time looking this package over and for your great feedback.

The main idea in `counsel-web` is to NOT have to browse search results in a browser--instead go directly to the page you were searching for. It accomplishes that in two steps:
1. `counsel-web-suggest`, which takes a search string, displays dynamic search suggestions, and returns the chosen suggestion. 
2. `counsel-web-search`, which takes a search string (possibly handed off from `counsel-web-suggest`) and browses a search candidate.

You are right in that I didn't envision a user wanting to use their web browser to view the results from `counsel-web-suggest`. I guess I should have since that is exactly what `counsel-search` (included in `counsel`) does. You may want to use that instead.

You can bypass `counsel-web-search` and get the same functionality from `counsel-web-suggest` by doing this:

```emacs-lisp
(defun browse-url-search (string)
  (browse-url
   (format "https://duckduckgo.com/?q=%s" (url-hexify-string string))))

(setq counsel-web-suggest-action #'browse-url-search)
```

Maybe that would be the start of a good enhancement to `counsel-web`.

If you want to change `counsel-web-search`'s default action to use your main browser, that is easier:

https://github.com/mnewt/counsel-web#change-the-default-browser

riscy:
These seem to have popped up in the latest -- possibly new additions?

```
10 issues found:
35:10: error: You should depend on (emacs "25.1") if you need `dom'.
162:27: error: You should depend on (emacs "25.1") if you need `alist-get'.
213:43: error: You should depend on (emacs "25.1") if you need `dom-attributes'.
215:12: error: You should depend on (emacs "25.1") if you need `dom-texts'.
217:7: error: You should depend on (emacs "25.1") if you need `dom-by-class'.
232:24: error: You should depend on (emacs "25.1") if you need `dom-by-tag'.
233:47: error: You should depend on (emacs "25.1") if you need `dom-attributes'.
238:16: error: You should depend on (emacs "25.1") if you need `dom-texts'.
239:48: error: You should depend on (emacs "25.1") if you need `dom-attributes'.
258:25: error: You should depend on (emacs "25.1") if you need `alist-get'.
```

mnewt:
Whoops, I actually changed the Emacs version required because I didn't realize when dom was added. Changed it back to 25.1 and those errors are gone. I just added package-lint to my flycheck config so I hopefully don't make that mistake again.

riscy:
Cool, this looks good to go -- and nice work supporting multiple search engines. :)

PR #6688, opened 2020-02-09
Add additional instructions re: licensing

riscy:
This extends CONTRIBUTING.org with some additional instructions regarding licenses in the source code.  Hopefully it is capturing recent issues such as use of AGPL (e.g. #6640) and idiosyncratically formatted licenses that aren't detected by tooling (e.g. #6599).


purcell:
Great!

PR #6689, opened 2020-02-10
Add auctex-cluttex recipe

tsuu32:

Add [ClutTeX](https://www.ctan.org/pkg/cluttex) support for [AUCTeX](https://www.gnu.org/software/auctex/) package.

https://github.com/tsuu32/auctex-cluttex

Author and maintainer

None needed.

### Checklist

Please confirm with `x`:

### Note
package-lint warns about function prefix such as `TeX-run-ClutTeX`, but this follows AUCTeX package's naming convention. Is this acceptable?

### DIFF
--- a/recipes/auctex-cluttex
+++ b/recipes/auctex-cluttex
@@ -1,0 +1,1 @@
+(auctex-cluttex :fetcher github :repo "tsuu32/auctex-cluttex")


conao3:
Hi!

1. Please remove top-level `setq` and `advice-add`.
  Instead of that, please prepare `*-setup` and `*-teardown` function for the user experiment.
  All packages should not have any side-effect cause `require` package.
2. Please use space indent instead of space and tab mixed indent.
  As add `indent-tabs-mode: nil` to file footer, you can ensure space indent in the future.
3. Please consider using `process-exit-status` instead of parse process output.
  You can see `diff-sentinel` as a usage sample.
4. Please consider using `user-error` to inform user error instead of `message`.
5. You can clean this `if` statement https://github.com/tsuu32/auctex-cluttex/blob/2d55265f7877ae5d152f7bb2130606f0e5e57efe/auctex-cluttex.el#L102-L106
  Please reference this commit https://github.com/tsuu32/auctex-cluttex/pull/1/commits/7aa1d59e033649e87bf856ff99987cf2e7358a53
6. Please consider using `when` if there are no else sexp

conao3:
Quote @tsuu32 comment from https://github.com/tsuu32/auctex-cluttex/pull/1

> I cannot merge this PR.
> 
> faf687a: No need to merge. Users who want to use this package should load this package by require, and users who don't want to this package should not download or load this package.
> 
> 4900840: This is not issue.
> 
> ac58971 a2c245e: My package follows AUCTeX package convention. No need to merge.
> 
> Please Don't create such a large PR.
> Thanks.

I'm purely sad from such aggressive comments.

---
In CONTRIBUTING include below statement
https://github.com/melpa/melpa/blob/master/CONTRIBUTING.org#making-your-package-ready-for-inclusion
> Coding style and conventions
The Emacs Lisp files should follow the [Emacs Lisp conventions](https://www.gnu.org/software/emacs/manual/html_node/elisp/Tips.html) and the [Emacs Lisp Style Guide](https://github.com/bbatsov/emacs-lisp-style-guide).

I add some references from my comment.
1. Maybe undocumented...?
Anyway, changing Emacs variables and functions just by requiring is a bad design.
If you need to change the global value of Emacs, prepare `*-setup` function and change it according to the user's intention.
When changing global values with `*-setup`, it is better to prepare a `*-teardown` function and leave room for users to restore.
@riscy is commented the same thing in other packages. https://github.com/melpa/melpa/pull/6615#issuecomment-583882774
2. See https://github.com/bbatsov/emacs-lisp-style-guide#source-code-layout--organization
3. Just suggestion and introduction `process-exit-status`
4. In https://www.gnu.org/software/emacs/manual/html_node/elisp/Programming-Tips.html#Programming-Tips

> When you encounter an error condition, call the function error (or signal). The function error does not return. See Signaling Errors.
Don't use message, throw, sleep-for, or beep to report errors. 

5. Just suggestion clean `if`. Don't forget all sexp return value
6. See https://github.com/bbatsov/emacs-lisp-style-guide#syntax
7. Just suggestion proper Cask file management

conao3:
related 1, in https://www.gnu.org/software/emacs/manual/html_node/elisp/Coding-Conventions.html#Coding-Conventions
> Simply loading a package should not change Emacs's editing behavior. Include a command or commands to enable and disable the feature, or to invoke it.

riscy:
Hi!  I think this looks cool.  A couple comments on the exchange above:

> Please Don't create such a large PR

For what it's worth, I didn't feel this was overly aggressive. :)  Many of the comments made are good, but I think this is a fair request since it's easier for the repo owner to accept or reject many small PR's, rather than parse a single large one.

------

A key issue raised above is that this package `setq`'s some variables when the package is loaded, which is discouraged.  The preference is for minor modes that can be easily be turned on (or off) and whose presence is easily seen by running `describe-mode`.  Tons of older packages don't conform, but it helps the growing Emacs ecosystem scale if we can work in this direction.

It also gives a new user an easy way to remove the package's effects (without restarting Emacs) if they don't like it.

So I think what we should do is aim to turn this pattern in your Commentary:
```elisp
(with-eval-after-load 'tex (require 'auctex-cluttex))
```
...into this:
```elisp
(with-eval-after-load 'tex (auctex-cluttex-mode))
```
by defining an autoloaded global minor mode that adds and removes from `TeX-command-list`  and `TeX-expand-list-builtin` when the mode is activated/deactivated.

Let me know if you need some more guidance on the minor mode approach.

Beyond that, a quick walk through the checklist:

### auctex-cluttex.el
- Per CONTRIBUTING.org, consider [using lexical binding](https://github.com/melpa/melpa/blob/master/CONTRIBUTING.org#fixing-typical-problems) (you will need at least emacs 24)
- Lexical binding has been enabled in what follows
- No issues! 💯

- No issues! 💯

```
3 issues found:
98:0: error: "TeX-run-ClutTeX" doesn't start with package's prefix "auctex-cluttex".
108:0: error: "TeX-ClutTeX-filter" doesn't start with package's prefix "auctex-cluttex".
116:0: error: "TeX-ClutTeX-sentinel" doesn't start with package's prefix "auctex-cluttex".
```
The rationale for sticking to your package-prefix is, as above, helping the elisp ecosystem as it continues to scale.  It could be the case that the author(s) of tex-mode decide to add a `TeX-ClutTex-sentinel` (or something) in the future, and per convention that "namespace" is theirs.

License:
- GitHub API found `GNU General Public License v3.0`
- _elisp/auctex-cluttex.el has gpl license text
(good)

Hope that helps!

tsuu32:
Thanks for the review!

> Many of the comments made are good, but I think this is a fair request since it's easier for the repo owner to accept or reject many small PR's, rather than parse a single large one.

That's what I meant.

---
> by defining an autoloaded global minor mode that adds and removes from TeX-command-list and TeX-expand-list-builtin when the mode is activated/deactivated.

I define global minor mode `auctex-cluttex-mode`. https://github.com/tsuu32/auctex-cluttex/commit/5b863ec1f0bb2e5f5c27efd7bf29e7380ece42e7  
So this is fixed upstream.

 What I do:
* add ` -*- lexical-binding: t -*-`
* Use `auctex-cluttex-` prefix
* Make `auctex-cluttex-mode`
* Do not `advice-add` globally

conao3:
The following issue has remained.

- Please remove top-level sexp. All packages should not have side-effects just require package for user-experience. see [Coding-conventions](https://www.gnu.org/software/emacs/manual/html_node/elisp/Coding-Conventions.html#Coding-Conventions).

tsuu32:
@conao3 Thanks for your PR.
I marged.

tsuu32:
> Please remove top-level sexp. All packages should not have side-effects just require package for user-experience. see Coding-conventions.

Hmm...  I marged but I worry about it. I understand top-level sexp is discouraged, but I think it should be acceptable.
I don't want users who don't have requirements to use this package.

And many packages such as major mode and theme packages write top-level sexp for instance `add-to-list` for `auto-mode-alist` and `custom-theme-load-path`. I think this is for user experience.

@riscy Isn't top-level `(unless (executalbe-find "cluttex") (error "..."))` appropriate as package on MELPA?

conao3:
There is no conflict. see [Packaging Basics](https://www.gnu.org/software/emacs/manual/html_node/elisp/Packaging-Basics.html).
> They are typically used to autoload the principal user commands defined in the package, but they can also perform other tasks, such as adding an element to auto-mode-alist (see Auto Major Mode).

So, I suggest leaving add-to-list `auto-mode-alist` but suggest removing that top-level sexp (and it generate error!)

But, my comments are just suggestion. 
It is up to the package owner to decide whether to adopt it.

tsuu32:
> But, my comments are just suggestion.
It is up to the package owner to decide whether to adopt it.

OK. This is my decision https://github.com/tsuu32/auctex-cluttex/commit/73188333b52445b8374a1b2d971748143b7fc411.

riscy:
This is looking great to me.  Thanks both.

And @tsuu32, welcome to MELPA! :)

tsuu32:
thanks.😀

PR #6690, opened 2020-02-11
Add center-layout.

laec:

Provides a global minor mode center-layout-mode for centering
windows by applying left and optionally right margins.

https://gitlab.com/lae/emacs-center-layout

I'm the maintainer.

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/center-layout
+++ b/recipes/center-layout
@@ -1,0 +1,4 @@
+(center-layout
+ :fetcher gitlab
+ :repo "lae/emacs-center-layout"
+ :branch "stable")


conao3:
Hi!

1. Please add `Keywords:`, `URL` meta information. see https://www.gnu.org/software/emacs/manual/html_node/elisp/Library-Headers.html
```diff
see
https://www.gnu.org/software/emacs/manual/html_node/elisp/Library-Headers.html

1 file changed, 2 insertions(+)
center-layout.el | 2 ++

modified   center-layout.el
@@ -3,6 +3,8 @@
 ;; Homepage: https://gitlab.com/lae/emacs-center-layout
 ;; Package-Version: 0.1.1
 ;; Package-Requires: ((emacs "25.1"))
+;; Keywords: tools
+;; URL: https://gitlab.com/lae/emacs-center-layout
 
 ;;; Commentary: 
```

2. Please concider using `non-nil` instead of `on`.
```diff
use 'non-nil' instead of 'on'

1 file changed, 5 insertions(+), 5 deletions(-)
center-layout.el | 10 +++++-----

modified   center-layout.el
@@ -104,11 +104,11 @@ ARGS will be passed to the function.
 `window-min-size' caculates the width by including the horizontal
 margins, and since function `center-layout-mode' always
 dynamically sets the margins to fill available space (especially
-when `center-layout-apply-right-margin' is on), this causes Emacs
-to think that there is no available horizontal space anymore so
-spliting windows will always occur below, never to the side,
-which is undesirable, especially when trying to use side windows
-like neotree.
+when `center-layout-apply-right-margin' is non-nil), this causes
+Emacs to think that there is no available horizontal space
+anymore so spliting windows will always occur below, never to the
+side, which is undesirable, especially when trying to use side
+windows like neotree.
 
 This advice will remove the horizontal margins from the result
 which is correct behaviour since margins will be dynamically
```

3. I don't familiar with buffer-local-variables, but `with-current-buffer` is needed to get it.
```diff
get buffer local variable using with-current-buffer

1 file changed, 4 insertions(+), 2 deletions(-)
center-layout.el | 6 ++++--

modified   center-layout.el
@@ -37,7 +37,8 @@ neotree, treemacs etc, return their content width if it is
 greater, as typically they are used as dedicate bottom/side
 windows and have their own layout, limiting their width will
 cause undesired line wrapping/truncation."
-  (let ((pixels (* (with-selected-window window center-layout-columns)
+  (let ((pixels (* (with-current-buffer (get-buffer-window window)
+                     center-layout-columns)
                    (window-font-width window))))
     (if (window-dedicated-p window)
         (max
@@ -69,7 +70,8 @@ cause undesired line wrapping/truncation."
   (let* ((margins (center-layout--free-columns window))
          (margin-left (ceiling margins 2))
          (margin-right
-          (if (with-selected-window window center-layout-apply-right-margin)
+          (if (with-current-buffer (get-buffer-window window)
+                center-layout-apply-right-margin)
               (- margins margin-left)
             0)))
     `(,margin-left . ,margin-right)))
```
---
BTW, @purcell, issue 1 is not warned package-lint.
We will be warned if there are no keywords included in `finder-known-keywords` as `:Keyward`, but pacakge-lint will not warne if file have not written this field in the first place.
Also. URL meta information is also used in `describe-package` etc. and should be described in every package.

walseb:
I'm not a melpa maintainer either but it would be nice if the readme included a comparison between this package and olivetti https://github.com/rnkn/olivetti

terlar:
There is also:
- https://melpa.org/#/centered-window
- https://melpa.org/#/visual-fill-column

Which also provide similar functionality as I understand it. Might be worth mentioning the differences and benefits as well.

laec:
@conao3 thanks for the review, I've [updated the docs](https://gitlab.com/lae/emacs-center-layout/commit/8b6a5a7be193741c9929d8935d16728f1459d2a6), didn't add `URL` because `Homepage` is already there and is the same as URL. `with-selected-window` is equivalent to `with-current-buffer` in this case, it's just that the former takes a window and the later takes a buffer as input.

@walseb @terlar thanks for the suggestions, I might have looked at some of those in the past but now I don't remember how they work anymore, my package is very minimal and mainly designed around my habit so the differences might just be in the details. I don't want to include comparisons to other packages in the readme because the comparisons will require quite a bit of time and effort to create and keep up to date as those packages get updated in the future, otherwise the comparison will get outdated and mislead the viewers, hope this make sense.

conao3:
Sorry, I overlooked `homepage` keyword. and I haven't known `with-selected-window` technic.
Thanks!

alphapapa:
> 1. I don't familiar with buffer-local-variables, but `with-current-buffer` is needed to get it.

Consider using `buffer-local-value` instead.  See https://github.com/alphapapa/emacs-package-dev-handbook#accessing-buffer-local-variables

alphapapa:
> I don't want to include comparisons to other packages in the readme because the comparisons will require quite a bit of time and effort to create and keep up to date as those packages get updated in the future, otherwise the comparison will get outdated and mislead the viewers, hope this make sense.

If you're going to add a package to MELPA that duplicates functionality in existing packages, you should document how it compares to those packages.  If the other packages change in the future, then when you become aware of it, you can update your comparisons.

If you don't compare them, then every user is left to do so individually, which is a waste of the user's time and effort in doing a task that the package author should have done.

If you're not sure that your package is distinctive enough from these other packages to be worth comparing them, then it's probably not worth publishing on MELPA.  MELPA is intended to be at least somewhat curated so that users don't have to research and distinguish between several packages that do same thing.  That's already a problem in several cases, and we don't want to keep doing that.

If your package is only slightly different from others, it would be better to contribute to those other packages so that users can benefit from the improvements without having the expense of extra choices to wade through.

alphapapa:
@laec Additionally, you checked the box indicating that you read `CONTRIBUTING.org`, but you have missed important requirements mentioned in it.

laec:
> If you're going to add a package to MELPA that duplicates functionality in existing packages, you should document how it compares to those packages. If the other packages change in the future, then when you become aware of it, you can update your comparisons.
>
> If you don't compare them, then every user is left to do so individually, which is a waste of the user's time and effort in doing a task that the package author should have done.

If I were to do such comparison, the responsible thing to do would be to evaluate all those packages in detail, and keep a close eye on them and re-compare when they get updated in the future, to ensure the comparison is as accurate as possible. Unfortunately it will require much of my free time, and my time is quite limited since work and off-work life are quite busy for me already. So I prefer to not do it at all instead of doing a potentially bad job at it. I think packages like this are highly based on personal tastes anyway, users can't tell if it suits their personal taste until they try it, and I already have all the features of my package clearly documented, which should be enough for the user to decide if they even want to try it.
 
> If you're not sure that your package is distinctive enough from these other packages to be worth comparing them, then it's probably not worth publishing on MELPA. MELPA is intended to be at least somewhat curated so that users don't have to research and distinguish between several packages that do same thing. That's already a problem in several cases, and we don't want to keep doing that.

I respect that and will happily accept this merge request to be rejected.

alphapapa:
> If I were to do such comparison, the responsible thing to do would be to evaluate all those packages in detail, and keep a close eye on them and re-compare when they get updated in the future, to ensure the comparison is as accurate as possible. Unfortunately it will require much of my free time, and my time is quite limited since work and off-work life are quite busy for me already. So I prefer to not do it at all instead of doing a potentially bad job at it.

It's okay if your comparison becomes out-of-date someday.  It doesn't need to be perfect forever.  It just needs to help guide users toward making a choice between packages that, to them, probably appear to be nearly the same.  You can always say something like, "This comparison was last updated on X.  Corrections and updates are welcome."  You can even share the comparison with authors of the other packages, so if they update their packages, they can let you know if the comparison needs updating.

But, really, don't overthink it.  Adding a comparison in your readme is not making a solemn commitment to ensure its accuracy forever.  There are no guarantees of any kind in this system. 
 Something is better than nothing here.  A sentence or two about other packages, like, "This one is similar but has more features and requires more configuration," is all you need.

laec:
Sorry for the late reply, I've been quite busy recently, so I think I'll close this for now. Thank you everyone for your feedback, much appreciated.

PR #6691, opened 2020-02-11
Add the client for the Meyvn build tool.

danielsz:
Meyvn is a Clojure live coding environment & build tool. 

The Meyvn client complements the cider nREPL with additional facilities such as reload on save, teardown and rebuild world, and also the ability to add Clojure dependencies at runtime.

https://github.com/danielsz/meyvn-el

I am the maintainer.

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/meyvn
+++ b/recipes/meyvn
@@ -1,0 +1,3 @@
+(meyvn
+ :fetcher github
+ :repo "danielsz/meyvn-el")


conao3:
Hi!

0. Please test your package ensure build via MELPA clean build see https://github.com/melpa/melpa/blob/master/CONTRIBUTING.org#test-your-recipe
  I need to add many dependency packages to ensure build your package.
1. Require `ivy` is not good idea. How about `helm` user? You want to provide ivy integrate feature, please separate package like `meyvn`, `meyvn-ivy`, (`meyvn-helm`) see https://github.com/melpa/melpa/blob/master/CONTRIBUTING.org#fixing-typical-problems
>     MELPA only looks at the main .el file (or -pkg.el file, if provided, though we discourage that). If you have different optional parts of a large package, each of which has different dependencies, then these should really be published as separate packages.

2. Please consider adding Cask file is good to easy resolve dependency, build, packaging etc.
3. Please add .gitignore file is good to ignore git stage *.elc files and .cask folder.
4. Please indent the whole file because of some of the places broken indent.
5. Please use `expand-file-name` instead of `concat` to create a file path.
6. Please remove top-level sexp.
All packages should not have side-effects just require package for user-experience.

danielsz:
Thank you for the feedback. It is helpful.

danielsz:
I changed the installation instructions to a one-liner. I hope it's alright.

``` emacs-lisp
(add-hook 'cider-mode-hook #'meyvn-setup)
```

conao3:
Good!

But this issue remained.
> Require ivy is not good idea. How about helm user? You want to provide ivy integrate feature, please separate package like meyvn, meyvn-ivy, (meyvn-helm) see https://github.com/melpa/melpa/blob/master/CONTRIBUTING.org#fixing-typical-problems
>> ELPA only looks at the main .el file (or -pkg.el file, if provided, though we discourage that). If you have different optional parts of a large package, each of which has different dependencies, then these should really be published as separate packages.

Please consider resolving this issue if you have time. (separate core feature and ivy integration feature)

danielsz:
OK, no problem. If you can point to a reference project that implements that well, that can be helpful. If not, no worries. Thank you.

conao3:
Thanks replay!
Please reference related [conversation](https://github.com/melpa/melpa/issues/6262) and the issued [package](https://github.com/akirak/org-starter).

riscy:
Very much appreciate the use of short functions in this code -- nice style.

I agree with @conao3 about the reliance on ivy.  I'm sure helm or ido or icicles users would appreciate generic calls to `completing-read` or specialized sub-packages, but altogether that's not a blocker.

You capitalize Mevyn in your docstrings but not in your package summary line -- maybe capitalize it in the latter?

You make a call to `(parseedn-read)` which is transitively imported by one of your dependencies, but you shouldn't rely on that being the case.  You should `Package-Requires` on [parseedn](https://github.com/clojure-emacs/parseedn) and `(require 'parseedn)` accordingly.

Minor:
- meyvn.el#L83: No `concat` required; messages are already f-strings
- meyvn.el#L118: No `concat` required; messages are already f-strings
- meyvn.el#L267: Consider `unless ...` instead of `when (not ...)`
- meyvn.el#L22: Prefer `https` over `http` (if possible)
- pacakge -> package
- "Does a Mevyn config exists?" -> "exist?"

danielsz:
I took note of all your comments, recommandations and feedback, and will do the work as soon as possible. You guys rock. Thank you so much!

danielsz:
Hi, 

I believe I've integrated all your recommendations. 
The dependency on `ivy` has been removed in favor of a generic `completing-read` implementation without loss of functionality. Please let me know if there is anything else. Thank you!

riscy:
Great, thanks!

danielsz:
It was a pleasure to work with you on this, you provided excellent guidance and in such a good spirit. This means a lot to me. Thanks a lot @riscy and @conao3.

PR #6693, opened 2020-02-12
add nndiscourse

dickmao:

A Gnus backend for Discourse

https://github.com/dickmao/nndiscourse

Maintainer

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/nndiscourse
+++ b/recipes/nndiscourse
@@ -1,3 +1,3 @@
 nndiscourse :repo "dickmao/nndiscourse"
             :fetcher github
-            :files ("nndiscourse.el" ("nndiscourse" "nndiscourse/Gemfile" "nndiscourse/Gemfile.lock" "nndiscourse/nndiscourse.gemspec" "nndiscourse/nndiscourse.thor" "nndiscourse/lib"))
+            :files ("nndiscourse.el" ("nndiscourse" ("nndiscourse/Gemfile" "nndiscourse/Gemfile.lock" "nndiscourse/nndiscourse.gemspec" "nndiscourse/nndiscourse.thor" "nndiscourse/lib")))
--- a/recipes/nndiscourse
+++ b/recipes/nndiscourse
@@ -1,3 +1,3 @@
-nndiscourse :repo "dickmao/nndiscourse"
-            :fetcher github
-            :files ("nndiscourse.el" ("nndiscourse" ("nndiscourse/Gemfile" "nndiscourse/Gemfile.lock" "nndiscourse/nndiscourse.gemspec" "nndiscourse/nndiscourse.thor" "nndiscourse/lib")))
+(nndiscourse :repo "dickmao/nndiscourse"
+             :fetcher github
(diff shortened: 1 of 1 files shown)


conao3:
Hi!

1. Please use proper indent in the whole Elisp file.
  Proper indent helps us understand Lisp code quickly.
2. Please use `prin1-to-string` instead of an empty `format`.
3. Please concider using `awhen` instead of `aif` which have no `else` form.
4. Please remove top-level sexp.
All packages should not have side-effects just require package for user-experience. see https://www.gnu.org/software/emacs/manual/html_node/elisp/Coding-Conventions.html#Coding-Conventions
> Simply loading a package should not change Emacs's editing behavior. Include a command or commands to enable and disable the feature, or to invoke it. 

(There are many top-level sexp, I cannot arrange `*-setup` and `*-teardown` function...)

riscy:
Quick first pass on this, but I agree with @conao3  -- *most* of the time, when you're adding hooks (or fsetting), you want to be doing these changes with the togglability of a minor mode.

The recipe looks complicated, maybe moreso than it needs to be?  Maybe it would be simplest if it only included the elisp files, and then did a `bundle-install` using the GitHub URL to your gem, rather than having the gem packaged with the elisp files.  Is that a possibility?

Other than that:

### nndiscourse.el ###

```
nndiscourse.el:50:1:Error: Cannot open load file: No such file or directory, json-rpc
```
Missing a `Package-Requires` in this case.

- No issues!

- No issues!

...except!:
```
;; Evade package-lint!
(fset 'gnus-user-format-function-S
      (symbol-function 'nndiscourse--format-time-elapsed))
```
Should this be override advice?  Should it be added to a global minor mode, to let the users toggle the behavior on and off in case they run into issues?

- nndiscourse.el#L371: It's safer to sharp-quote function names; use `#'`
- nndiscourse.el#L985: It's safer to sharp-quote function names; use `#'`
- nndiscourse.el#L1081: It's safer to sharp-quote function names; use `#'`
- nndiscourse.el#L1082: It's safer to sharp-quote function names; use `#'`

dickmao:
> The recipe looks complicated, maybe moreso than it needs to be?

I'd like to beg off a two-stage install.  I'm a beginner rubyist and bundler confuses me generally.  I also don't think the recipe is complicated, and as I mentioned in `nnreddit`, the ruby code makes no sense without the elisp.  That is, the ruby code should not comprise an independent gem.

> nndiscourse.el:50:1:Error: Cannot open load file: No such file or directory, json-rpc

Fixed, thank you. https://github.com/dickmao/nndiscourse/commit/2721fdff07d793e0b8019fc1e5d34264a2ef7fad

> (fset 'gnus-user-format-function-S
>       (symbol-function 'nndiscourse--format-time-elapsed))
> Should it be added to a global minor mode?

No?  There is so much that is "wrong" (by modern standards) about Gnus that this small liberty seems *de minimus*.  There are precedents for this namespace pollution in `gnus-diary.el` and `gnus-registry.el`.

dickmao:
> ;; Evade package-lint!
> (fset 'gnus-user-format-function-S

Hum I now remember trying to work around this for nnreddit.  I forget the details but something about Gnus made this very hard -- like the `gnus-parameters` (Gnus's contortionist mechanism for decoupling backend namespaces) didn't allow hygienically setting function bindings, only value bindings (and Gnus line formatting needs to get at a function binding).

riscy:
Okay, this looks good then.  Thanks!

PR #6694, opened 2020-02-12
update edts

getong:
The edts has moved to https://github.com/sebastiw/edts .
And this has changed to use rebar3 to compile.

### DIFF
--- a/recipes/edts
+++ b/recipes/edts
@@ -1,40 +1,31 @@
 (edts
  :fetcher github
- :repo   "tjarvstrand/edts"
+ :repo   "sebastiw/edts"
  :files  ("*.el"
           "COPYING"
           "COPYING.LESSER"
           "Makefile"
           "README.md"
           "start"
+          "start.bat"
+          "rebar.config"
+          ("config"                     "config/*")
 
           ("elisp/edts"                 "elisp/edts/*.el")
           (:exclude                     "elisp/edts/*-test.el")
 
-          ("lib/edts"                   "lib/edts/Makefile")
-          ("lib/edts"                   "lib/edts/rebar")
-          ("lib/edts"                   "lib/edts/rebar.config")
-          ("lib/edts/priv"              "lib/edts/priv/app.config")
+          ("lib/edts/include"           "lib/edts/include/*")
           ("lib/edts/priv"              "lib/edts/priv/dispatch.conf")
           ("lib/edts/src"               "lib/edts/src/*")
 
-          ("plugins/edts_debug"         "plugins/edts_debug/*.el")
-          (:exclude                     "plugins/edts_debug/*-test.el")
-          ("plugins/edts_debug"         "plugins/edts_debug/Makefile")
-          ("plugins/edts_debug"         "plugins/edts_debug/rebar")
-          ("plugins/edts_debug"         "plugins/edts_debug/rebar.config")
-          ("plugins/edts_debug/src"     "plugins/edts_debug/src/*")
+          ("lib/edts_debug"             "lib/edts_debug/*.el")
+          (:exclude                     "lib/edts_debug/*-test.el")
+          ("lib/edts_debug/src"         "lib/edts_debug/src/*")
 
-          ("plugins/edts_dialyzer"      "plugins/edts_dialyzer/*.el")
-          (:exclude                     "plugins/edts_dialyzer/*-test.el")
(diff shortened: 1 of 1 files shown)


tarsius:
... from https://github.com/tjarvstrand/edts before, which redirects to the new repository. So this looks official. Still it would be nice if you could link to the conversation that you surely had about this move.

getong:

Erlang Development Tool Suite

https://github.com/sebastiw/edts

I am An enthusiastic user.

None needed

### Checklist

Please confirm with `x`:

- [x] The package is released under a [GNU Lesser General Public License v3.0](https://www.gnu.org/licenses/license-list.en.html#GPLCompatibleLicenses).

getong:
@sebastiw
Would you take a look at of this?
I test it with
```
make recipes/edts
tar xf packages/edts-20200209.1523.tar -C ~/.emacs.d/elpa
cd ~/.emacs.d/elpa/edts-20200209.1523/ && make
```
And I restart the emacs, it works fine.

recipes/edts
```diff
@@ -1,40 +1,29 @@
...
-          ("lib/edts"                   "lib/edts/Makefile")
-          ("lib/edts"                   "lib/edts/rebar")
-          ("lib/edts"                   "lib/edts/rebar.config")
-          ("lib/edts/priv"              "lib/edts/priv/app.config")
+          ("lib/edts/include"           "lib/edts/include/*")
           ("lib/edts/priv"              "lib/edts/priv/dispatch.conf")
           ("lib/edts/src"               "lib/edts/src/*")
 
-          ("plugins/edts_debug"         "plugins/edts_debug/*.el")
-          (:exclude                     "plugins/edts_debug/*-test.el")
```
sebastiw: Strip this test as well

getong: The two lines are deleted.

sebastiw:
> Still it would be nice if you could link to the conversation that you surely had about this move.

It was private email conversation between me and Thomas in Swedish, so I doubt it helps much.

getong:
https://github.com/sebastiw/edts/commit/6e08bc97ce8a923a6ec4e0af442e861ad5010298
The edts project does not maintain for a long time, it just needs to update.

recipes/edts
sebastiw: Yes, I can see that, but we should keep the line `(:exclude "plugins/edts_debug/*-test.el")` because we don't want to include the build tests in the package.

getong: Done

recipes/edts
```diff
@@ -1,40 +1,30 @@
...
-          ("plugins/edts_debug"         "plugins/edts_debug/Makefile")
-          ("plugins/edts_debug"         "plugins/edts_debug/rebar")
-          ("plugins/edts_debug"         "plugins/edts_debug/rebar.config")
-          ("plugins/edts_debug/src"     "plugins/edts_debug/src/*")
+          ("lib/edts_debug"             "lib/edts_debug/*.el")
+          (:exclude                     "lib/edts_debug/*-test.el")
+          ("lib/edts_debug/src"         "lib/edts_debug/src/*")
 
-          ("plugins/edts_dialyzer"      "plugins/edts_dialyzer/*.el")
-          (:exclude                     "plugins/edts_dialyzer/*-test.el")
```
sebastiw: I think also future tests in edts_dialyzer should be excluded.
`(:exclude                     "lib/edts_dialyzer/*-test.el")`

getong: This directory does not contain any test .el files.

sebastiw: No, but if it is included in the future there will have to be a new PR for melpa. My point is that it is easy to forget that and now all files are included. `:exclude` won't do any harm.

getong: done

riscy:
> It was private email conversation between me and Thomas in Swedish, so I doubt it helps much.

Except for the "private email" part, language is basically not an issue these days with translation services being where they're at. :)

I think this is good to merge, but while I have the attention of the maintainers:

- edts-mode.el has no detectable license text
- edts-start.el has no detectable license text
- You may also want to bring the byte-compilation up to a working state, so that each file can byte-compiled individually.  For instance:
```
edts-shell.el:25:1:Error: Symbol's value as variable is void: edts-erl-sname
```
In this case, edts-shell.el requires edts-api.el which uses `edts-erl-sname`, but that variable is defined in edts-mode.el (which isn't imported by edts-api.el).  I think you can solve it by adding `(require 'edts-mode)` to edts-api.el.

PR #6695, opened 2020-02-12
add geolocation recipe

gonewest818:

This package obtains your current location by obtaining the MAC addresses of nearby wifi access points and then submitting that list to third-party geolocation APIs. Those geolocation services use the known locations of the wifi access points and the relative strength of each signal to triangulate your latitude and longitude.

https://github.com/gonewest818/geolocation.el

I am the author.

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/geolocation
+++ b/recipes/geolocation
@@ -1,0 +1,1 @@
+(geolocation :fetcher github :repo "gonewest818/geolocation.el")


conao3:
Hi!

1. Please consider adding Cask file to easy resolve dependency, build etc.
2. Please add .cask entry to .gitignore file.
3. Please concider using when-let instead of if-let which have no else form.
4. Please use quote instad of backquote end of symbol

alphapapa:
> * Please consider adding Cask file to easy resolve dependency, build etc.
> * Please add .cask entry to .gitignore file.

@conao3 You've been asking nearly every recently submitted package to add Cask files.  This is not in the MELPA contributing guidelines.  Although useful, Cask is a non-standard, idiosyncratic tool.  Adding correct Cask files requires consulting extensive documentation to learn how it works.  

In fact, you recently [added](https://github.com/mtekman/remind-bindings.el/commit/a93f76d29ac12b1addd0908036f3101a3e21968b) a Cask file to another author's project, which he merged without question, yet the Cask file does not seem to be correct: it does not list the dependencies which the package depends on, while they are listed in the Elisp file's headers.  Surely adding an incomplete, incorrect Cask file does more harm than good.

As well, Cask files are not necessary for the simple case of declaring package dependencies.  This is already covered by the Elisp `Package-Requires` header, which is [not difficult to parse](https://github.com/alphapapa/makem.sh/blob/master/makem.sh#L402).

So please stop asking authors of new packages to add unnecessary Cask files to their projects.  If they want to use Cask, great.  If you recommend Cask to them to fulfill a need they have, great.  But please do not blindly tell all authors to add Cask files when they don't even use or need Cask.  All that does is clutter up git repos with useless files and mislead tools which read Cask files expecting them to be correct.  And it implies to new authors that Cask is a standard, necessary tool for Emacs packages, which it is not.

gonewest818:
Just a note that I'm aware of Cask, but a few years back I found Cask was not actively maintained (https://github.com/cask/cask/issues/376).  For my purposes I found it just as easy to include a local `.emacs/init.el` and a `Makefile`. That's what I'm using to manage dependencies, run tests, generate code coverage reports, run `elisp-lint` and so on.  It's not flashy but it works well for me.

conao3:
Managing dependencies in two places is a bad idea.
In fact, it is possible to manage if only the dependency list is described in Cask and not the version, but if the dependent version changes, it is difficult for humans to change the main Elisp file header and Cask file. There is always a leak in the correction.
Cask can resolve dependencies by referring to the header of the file specified in `package-file`. There is no need to manage the same information in two places.

Needless to say, parsing is easy. However, it is important that `cask; cask build` can build my first-time packages.
My comments are not enforceable. Package author can just ignore the commits if he doesn't need them, and it's easy to do it with git.
However, it doesn't seem to be welcome, so I'll refrain from adding Cask comments in the future.

gonewest818:
The `Makefile` mechanism I have here is only for continuous integration tests. Agree there is the risk that dependencies get out of sync, but actually, what I'm testing in CircleCI is always the latest from MELPA (and as we are all aware, MELPA doesn't serve versions prior to the latest anyway so you have to support what's out there in the wild).

I'm happy to consider alternative tools and stop supporting my own Makefile technique, but it hasn't been a burden so far and switching is not a priority for me right now.

gonewest818:
FYI, the other stylistic feedback (use of when-let, misquoted symbol in docstring, etc) was addressed. So if the MELPA maintainers have feedback to add, just let me know.
thank you -

gonewest818:
For what it's worth, the library has had a significant update since I first submitted this pull request. 
* interactions with system utilities ("netsh" on windows, "airport" on macos) run asynchronously
* service requests to external services such as Google Maps Geolocation are also asynchronous
* background process to watch your position and call a hook when there is a new position available (parity with the feature in `osx-location` but in a cross platform way).

Would be great to get this published on MELPA at this point because I expect the development to quiet down for a while as I move to my next side project.

riscy:
>  I'll refrain from adding Cask comments in the future.

@conao3 I know your heart is in the right place.  Thanks. :)

I was just about to mention [osx-location](https://github.com/purcell/osx-location/)!  Okay, cool.  This looks solid to me.

I think the reliance on `auth-source` for `auth-source-pick-first-password` should be made an explicit `require` (rather than relying on it being transitively imported, just in case).  I won't consider that a blocker, but it's probably worth tuning up.  

The ``` `back/front quoting' ``` in docstrings, such as here:
```elisp
(defvar geolocation-location nil
  "The most recently scanned location.

Value is nil if the position is unknown.  Otherwise the value is
an alist with the following keys:
  `latitude'  : latitude of the current position
  `longitude' : longitude of the current position
  `accuracy'  : accuracy of the estimate in meters
  `timestamp' : timestamp when this position was obtained")
```
is usually to create hyperlinks in the Emacs help to the help pages of other symbols -- not a blocker, but just an fyi.

And ultra minor:
- utilites -> utilities
- utilty -> utility
- Technologoies -> Technologies
- `calendar-latitiude` -> `calender-latitude`

gonewest818:
Thank you.  I’ll fix up the spelling errors (I’ve enabled fly spell) and the dependency on auth-source soon.  I’ll probably go back and deal with the symbol quoting later.

riscy:
No worries, those were some of my pickiest comments. :)

gonewest818:
All good. 

I’m also going to check why the elisp-lint didn’t flag the transitive dependency. Seems like that ought to have been caught.

PR #6702, opened 2020-02-17
Add org-pdftools

fuxialexander:

Add support for org links from pdftools buffers with more precise location control.

https://github.com/fuxialexander/org-pdftools

Maintainer

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/org-pdftools
+++ b/recipes/org-pdftools
@@ -1,0 +1,1 @@
+(org-pdftools :repo "fuxialexander/org-pdftools" :fetcher github)
--- a/recipes/org-pdftools
+++ b/recipes/org-pdftools
@@ -1,1 +1,2 @@
-(org-pdftools :repo "fuxialexander/org-pdftools" :fetcher github)
+(org-pdftools :repo "fuxialexander/org-pdftools" :fetcher github
+              :files ("org-pdftools.el"))


riscy:
### org-pdftools.el
```
org-pdftools.el:387:1:Warning: the following functions are not known to be
defined: string-empty-p, org-pdftools-open-custom-open,
```
For org-pdftools-open-custom-open, you can't sharp-quote line 208 since the symbol is a variable, not a function (so just leave it without any quoting).  

```
Warning (emacs):
org-pdftools.el:99: All variables and subroutines might as well have a documentation string
Warning (emacs):
org-pdftools.el:202: All variables and subroutines might as well have a documentation string
Warning (emacs):
org-pdftools.el:281: `y-or-n-p' argument should end with "? "
```

```
3 issues found:
78:1: warning: `eval-after-load' is for use in configurations, and should rarely be used in packages.
80:6: error: You should depend on (emacs "26.1") if you need `org-link-set-parameters'.
95:2: error: Aliases should start with the package's prefix "org-pdftools".
```
Line 80 is okay since you depend on a later version of `org`.  For the alias error, instead of this:
```elisp
;; org 9.3 and later use org-link-store-props
;; before that it was org-store-link-props, also used by this code
;; alias if required so that this code works on old or new orgmode
(unless (fboundp 'org-link-store-props)
  (defalias 'org-link-store-props 'org-store-link-props))
```
...it's better to write your own function that encapsulates this `fboundp` check and calls the correct function, rather than defaliasing something outside your "namespace".


- org-pdftools.el#L216: It's safer to sharp-quote function names; use `#'`
- org-pdftools.el#L95: It's safer to sharp-quote function names; use `#'`
- org-pdftools.el#L22: Prefer `https` over `http` (if possible)

License:
- ✔ GitHub API found `GNU General Public License v3.0`
- ✔ org-pdftools.el has GPL license text

riscy:
Quick first pass:

### fit-text-scale.el
- No issues! 💯

- No issues! 💯

```
1 issue found:
105:10: warning: This file is not in the `cl-lib' ELPA compatibility package: require `cl-lib' instead.
```

- fit-text-scale.el#L28: Prefer `https` over `http` (if possible)

License:
- LICENSE excerpt: `Copyright (C) 2017-2020 Marco Wahl...`
- ✔ fit-text-scale.el has GPL license text

I notice your LICENSE file is just the GPL boilerplate, rather than the full version of the GPL.

riscy:
(Oops, I pasted a review in here that was for another package)

fuxialexander:
@riscy Thanks a lot for the comments! Actually I was not sure about how to deal with some of those problems. I'll try to revise them based on your suggestions.

fuxialexander:
@riscy May I ask how should I fix 'All variables and subroutines might as well have a documentation string'?

randomwangran:
It's nice to see experts helping one of my favourite org-mode extension.

Really appreciate all of you.

riscy:
> May I ask how should I fix 'All variables and subroutines might as well have a documentation string'?

Sure!  Just adding a docstring to the function (or variable) in question is all you need to do:
```elisp
(defun xselect-convert-to-length (_selection _type value)
  "Here is a docstring for a function that takes a VALUE."  ; <-------
  (let ((len (cond ((stringp value)
		    (length value))
		   ((setq value (xselect--selection-bounds value))
		    (abs (- (nth 0 value) (nth 1 value)))))))
    (if len
	(xselect--int-to-cons len))))
```

conao3:
Hi!

```emacs-lisp
 org-pdft…    33   1 error           Cannot open load file: No such file or directory, org-refile (emacs-lisp)
 org-pdf…    67     warning         There should be two spaces after a period (emacs-lisp-checkdoc)
 org-pdf…   369     warning         There should be two spaces after a period (emacs-lisp-checkdoc)
```

riscy:
Sorry for such a long turnaround - I think the first error @conao3 brings up is because `org-refile` is a function and not a package (correct me if I'm wrong?).

If I fix that, I run into another compiler error:
```
org-pdftools.el:37:1:Error: pdf-info-epdfinfo-program is not executable
```
This is because, unfortunately, `pdf-annot` defines some default values that require epdfinfo to be installed (ideally that package would delay this check until runtime, but I digress).  This is not your fault or something you need to fix -- I'm just leaving it here so I have a note for myself to come back to.

Instead of installing epdftools in my sandbox, I decided to just comment out the `(require 'pdf-annot)` line and I arrive at the following (edited to remove the pdf-annot nags):

byte-compile and load-file (using Emacs 26.3):
```
org-pdftools.el:295:1:Warning: Unused lexical variable `link'
In org-pdftools-store-link:
org-pdftools.el:329:24:Warning: reference to free variable `desc'
In end of data:
org-pdftools.el:391:1:Warning: the following functions are not known to be
    defined: string-empty-p, acons
```
For `string-empty-p` you should `(require 'subr-x)`.  Instead of `acons` you should use `cl-acons` and `(require 'cl-lib)`.

Also a note about defcustoms -- when `nil` is a valid value, instead of something like the following:
```elisp
(defcustom org-pdftools-open-custom-open nil
  "Custom function to open linked pdf files."
  :group 'org-pdftools
  :type 'function)
```
You can use:
```elisp
(defcustom org-pdftools-open-custom-open nil
  "Custom function to open linked pdf files."
  :group 'org-pdftools
  :type '(choice function nil))
```

I think addressing the above should patch everything up for this PR.

Some other minor non-blockers:
- Seperator -> Separator

fuxialexander:
@riscy 
`org-refile` is a package (https://github.com/bzg/org-mode/blob/master/lisp/org-refile.el). Adding `(require 'org-refile)` fixed a compile error in my machine. Not sure why it cannot be load in the above case.

I'll fix the problems you mention ASAP. I was dealing with my work and not being very responsive recently.

riscy:
You're correct, sorry - in the org that comes with Emacs 26.3 org-refile is in org.el, but now it's moved. :)

riscy:
And feel free to ping when ready. :)

riscy:
I'm just browsing through the org-mode commit history and I notice version 9.3.6 still has `org-refile` inside org.el:

https://github.com/bzg/org-mode/blob/66810d2121aa1d936238fc53cc155e6eda425313/lisp/org.el#L9228

and there's no org-refile.el file yet at this point:

https://github.com/bzg/org-mode/tree/66810d2121aa1d936238fc53cc155e6eda425313/lisp

so I believe you want to require on org 9.4

fuxialexander:
@riscy Hi! Sorry for the long delay. I think I've fixed all the problems you mentioned above.

riscy:
I'm just noticing, and this isn't really a blocker, but it's very difficult for someone to install org 9.4 -- the current version tagged `20200323` available from http://orgmode.org/elpa is 9.3.6!

As a result I worry a lot of people would have unnecessary trouble using your package.

And! your package seems to work fine without the `require 'org-refile` at least on older versions of org.  Does it strictly need to be imported on org 9.4 for your package to compile?  Could you safely use `(require 'org-refile nil t)`?

purcell:
Agreed, there's no blocker here, so I'll merge now.

However, please consider the points in @riscy's final comment above.

Additionally, you should add a `defgroup` form, since you refer to the `org-pdftools` custom group but haven't defined it anywhere.

fuxialexander:
@purcell @riscy Thanks for your comment. I've changed the dependency of org to 9.3.6 and removed the org-refile line. Somehow that works for me now (Previously a function was missing when org-refile is not included). 

I've also just added the `defgroup`.

Thanks a lot for your kind help again!

PR #6703, opened 2020-02-17
Add share2computer

tumashu:

an elisp helper of android ShareToComputer app.

https://github.com/tumashu/share2computer

maintainer

 **None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/share2computer
+++ b/recipes/share2computer
@@ -1,0 +1,1 @@
+(share2computer :repo "tumashu/share2computer" :fetcher github)


conao3:
Hi!

1. Please add some keywords to `Keywords` header.
2. Please define custom variables group before use.

I pick-up `convenience` keyword for your package, but there is a better keyword maybe.

riscy:
Hello!  Mostly this looks good, just a few quick patches.  First, please use a LICENSE file that GitHub can detect -- see: https://github.com/licensee/licensee.  

Would it be worth adding a `(require 'org)` since org comes with Emacs?  That way you won't have to stub the org functions with these:
```elisp
(declare-function org-back-to-heading "org" (&optional invisible-ok))
(declare-function org-attach-dir "org" (&optional create-if-not-exists-p no-fs-check))
```
if you want to delay pulling in org, you could always inline the require:
```elisp

;;;###autoload
(defun share2computer-org ()
  "Download files shared by Android ShareToComputer to org attach dir."
  (interactive)
  (require 'org)
  (let (c marker)
    (when (eq major-mode 'org-agenda-mode)
      (setq marker (or (get-text-property (point) 'org-hd-marker)
		       (get-text-property (point) 'org-marker)))
      (unless marker
	(error "No task in current line")))
    (save-excursion
      (when marker
	(set-buffer (marker-buffer marker))
	(goto-char marker))
      (org-back-to-heading t)
      (share2computer--internal (org-attach-dir t)))))
```

Truly the license-in-repo is the most important, the rest is for your consideration.  This looks good.

And some non-blockers (for your consideration):
- I noticed @conoa3's comment about additional keywords, and `comm` feels like a nice fit: `(comm . "communications, networking, and remote file access")`

tumashu:
@riscy should fixed

riscy:
I should have tested my suggestion, but it looks like if you inline the `(require 'org)` (which I think you should still do) you still need the `declare-function` lines to resolve compiler warnings:

```
share2computer.el:270:1:Warning: the following functions are not known to be
    defined: org-back-to-heading, org-attach-dir
```

I don't think that's worth blocking on since I know you can fix it -- and apologies for the oversight on my part.

PR #6705, opened 2020-02-17
Add native-complete recipe

CeleritasCelery:

This package interacts with a shell's native completion functionality to provide the same completions in Emacs that you would get from the shell itself.

https://github.com/CeleritasCelery/emacs-native-shell-complete

I am the maintainer

None Needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/native-complete
+++ b/recipes/native-complete
@@ -1,0 +1,3 @@
+(native-complete :fetcher github
+                 :repo "CeleritasCelery/emacs-native-shell-complete"
+                 :files ("native-complete.el"))


terlar:
How does this compare to https://melpa.org/#/fish-completion which despite it’s name supports both bash and fish completions.

CeleritasCelery:
`fish-completion` is similar to the more popular [bash-completion](https://github.com/szermatt/emacs-bash-completion) package. They both start a separate shell process and then use that to try and do a "best-effort" completion. But neither of these can know the actual state of your shell, since they are running in a different process. This means that they won't be synchronized with your environment variables, aliases, functions, or PATH. They will only know the initial environment that is part of your shell startup. 

Another issue that this package handles correctly is subshells. Within a shell I can invoke a new shell that is totally different type. The other "best-effort" completion packages can't deal with this.

This package interacts with your actual running process to get completions so that is always 100% correct, and will handle any changes to the environment or subshells correctly. Basically it gives you what you would get from using `TAB` in an interactive terminal.

manuel-uberti:
FWIW, I've been using this for a while now and it works as expected.

alphapapa:
@CeleritasCelery This looks very interesting.  If you are so-inclined, a short GIF demo would be very helpful to prospective users.  https://github.com/alphapapa/emacs-package-dev-handbook#animations--screencasts

CeleritasCelery:
@alphapapa good idea. I added a screencast gif. That utility you pointed to is top notch.

alphapapa:
@CeleritasCelery Great demo!  It really shows what the package is capable of.  Very cool.

riscy:
Woah, cool!  Here's a quick pass:

### native-complete.el ###

```
native-complete.el:79:18:Warning: reference to free variable
    `comint-terminfo-terminal'
native-complete.el:80:13:Warning: assignment to free variable
    `comint-terminfo-terminal'
native-complete.el:124:1:Warning: Unused lexical variable
    `comint-redirect-perform-sanity-check'
In native-complete--get-prefix:
native-complete.el:140:29:Warning: reference to free variable
    `comint-prompt-regexp'
native-complete.el:155:1:Warning: Unused lexical variable `ansi-color-context'
In native-complete-at-point:
native-complete.el:206:36:Warning: reference to free variable
    `comint-redirect-completed'
In end of data:
native-complete.el:217:1:Warning: the following functions are not known to be
    defined: comint-redirect-cleanup, ansi-color-filter-apply
```
I think adding `(require 'comint)` clears _all_ of that.  Easy.

- No issues!

```
1 issue found:
78:3: warning: `with-eval-after-load' is for use in configurations, and should rarely be used in packages.
```
I think for this function:
```elisp
(defun native-complete-setup-bash ()
  "Setup support for native-complete enabled bash shells.
This involves not sending the `--noediting' argument as well as
setting `TERM' to a value other then dumb."
  (interactive)
  (with-eval-after-load 'shell
    (when (equal comint-terminfo-terminal "dumb")
      (setq comint-terminfo-terminal "vt50"))
    (setq explicit-bash-args
          (delete "--noediting" explicit-bash-args))))
```
you could just `(require 'shell)` and do away with the `with-eval-after-load`?  I almost feel like this could be a minor mode (where these variables' values were toggled in and out), but I'll leave that for your consideration.

- native-complete.el#L122: It's safer to sharp-quote function names; use `#'`

### License ###

- GitHub API found `GNU Lesser General Public License v3.0`
- native-complete.el has no detectable license boilerplate

For this one, can you add some of the standard GNU (Lesser) GPL boilerplate above your `Commentary` section?

CeleritasCelery:
Thanks @riscy . I have addressed your feedback. I ended up just using `(require 'shell)` since that cleaned up all the byte-compiler warnings and removed the need for `with-eval-after-load`.

riscy:
Thanks, taking another look now.

This seems to work great for me, but looks like it should probably be a minor mode due to the `advice-add` call (i.e., ideally it's possible for a user to turn the completion off without restarting their emacs session).  Turning the minor mode on would add the advice, and turning the minor mode off would remove the advice.

Another optional but encouraged thing to do would be to define a function named exactly: `native-complete-unload-function`.  This function, if defined, is automatically called when someone uses <kbd>M-x unload-feature</kbd>, and this is a good place to remove any advice that gets added.  If you don't define this function, someone might `unload-feature` your package, but the advice will stick around, which could cause some breaking behavior for them.

tarsius:
Probably also merge #6706 when merging this.

CeleritasCelery:
I added an unload function. However I am still holding off on adding the an explicit minor mode. Normally I would completely agree with you, but in this case I feel like it adds unnecessary setup complexity. This advice just aborts redirection if the user tries to send another command, which should be the default behavior of Emacs anyways (I might suggest that on the devel mailing list). If you don't do that you end up in nested redirections, which breaks the shell and you have to start a new one. If a user finds a use-case where this advice is causing a real issue I will create a minor mode then.

riscy:
Great, thanks!

PR #6708, opened 2020-02-18
Add ivy-emoji

Sbozzolo:

Insert emoji in Emacs buffers using the ivy selection framework.

https://github.com/sbozzolo/ivy-emoji

Author and Maintainer

**None needed** 

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/ivy-emoji
+++ b/recipes/ivy-emoji
@@ -1,0 +1,1 @@
+(ivy-emoji :fetcher github :repo "sbozzolo/ivy-emoji")


conao3:
Hi!

```
 ivy-emoj…    30     warning         There should be two spaces after a period (emacs-lisp-checkdoc)
 ivy-emoj…    31     warning         There should be two spaces after a period (emacs-lisp-checkdoc)
```

Sbozzolo:
Thanks, @conao3. I addressed your comments. How did you detect those warnings? `checkdoc` and `package-lint-current-buffer` produce clean outputs.

conao3:
The checkdoc warnings can be customized to return what errors depending on the global value.
You need to remove those customizations to see all the standard errors.

riscy:
Sorry for the long turnaround -- the particular checkdoc warning is coming because you have `sentence-end-double-space` set to nil, and by default it's t.  (Some Emacs starter packages even set this value to nil!)  As far as warnings go, that one's not a big deal, so more of a "for your consideration" thing.

This is cool!  A couple small suggestions:

It may be more efficient to call `char-to-string/char-from-name`  than to create a temporary buffer from which to pull out each emoji (although later on I think the function could be even more performant if written in as a mapcar, but it's not important).  I also think you can reduce the let-nesting in the function as follows:
```elisp
(defun ivy-emoji--create-list ()
  (let (emoji-list)
    (dolist (range ivy-emoji-codepoint-ranges)
      (dotimes (i (- (cdr range) (car range)))
        (let* ((codepoint (+ (car range) i))
               (name (get-char-code-property codepoint 'name)))
          (when name
            (setq emoji-list
                  (append emoji-list
                          (list (format "%s %s"
                                        (char-to-string (char-from-name name))
                                        (ivy-emoji--clean-name name)))))))))
    emoji-list))
```

Also I often nag people to use an https link rather than an http link (on line 23) to the license page.

Sbozzolo:
Thank you for your response.

I updated the package according to your comments.

riscy:
Everything looks great with the code now -- I just noticed this package:

https://github.com/dunn/company-emoji/

Have you had a chance to check that out or considered collaborating?  The code and framework appear to be different, but ultimately these two packages occupy a similar niche.  It's okay if the answer is no, but some explanation of your package's novelty and a link in your README to this existing similar package are good to have.

Sbozzolo:
Yes, I tried the package but I found that It gets on my way when I type. For example, if I want to type :D, `company-emoji` would try to autocomplete with ✡. Most of the time, when I type colon I don't want to insert an emoji, so the completion is distracting and not useful. `ivy-emoji` is for inserting an emoji when it is needed.

I'll update the README with some more information about the two packages.

Thanks.

riscy:
Okay, looks good!

PR #6710, opened 2020-02-18
Add mini-frame

muffinmad:

Place minibuffer at the top of the current frame while reading user input

https://github.com/muffinmad/emacs-mini-frame

Maintainer

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/mini-frame
+++ b/recipes/mini-frame
@@ -1,0 +1,1 @@
+(mini-frame :fetcher github :repo "muffinmad/emacs-mini-frame")


conao3:
Hi!

```
 mini-fr…   239   1 warning         Unused lexical variable ‘resize-mini-frames’ (emacs-lisp)
```

muffinmad:
> 1. Please resolve below package-lint/check-doc/byte-compiler warnings

Thanks! Emacs 27 was not complaining about `resize-mini-frames`, forgot to check on 26.

conao3:
good!
And the following comments help you and your package.

1. Please remove unneeded progn

---

What a difference from ivy-posframe or helm-posframe?

And if you want to use childframe, require posframe make your package simple may be.

- https://github.com/tumashu/ivy-posframe
- https://github.com/tumashu/helm-posframe

muffinmad:
> 1. Please remove unneeded progn

Done. Thanks for PR!

> What a difference from ivy-posframe or helm-posframe?

Well, I don't know if we can compare mini-frame to ivy/helm-posframe because mini-frame isn't targeted for specific completion system.

I don't consider mini-frame as ivy-posframe competitor. But mouse clicks is working in mini-frame with ivy unlike in ivy-posframe :)

mini-frame can be used without completion system at all. It's just minibuffer in child frame.

> And if you want to use childframe, require posframe make your package simple may be.

I don't see how can posframe can be used because posframe create its frames with `minibuffer . nil` while I need `minibuffer . only`.

What advantages of using posframe instead of vanilla Emacs functions to work with frames?

alphapapa:
This looks really cool!  A little feedback:

* I'd suggest naming the package `minibuffer-in-frame`, or maybe `minibuffer-on-top`, because `mini-frame` doesn't necessarily suggest that its purpose is to change the minibuffer.  When I saw the name, I figured it must be something to create very small frames--which it kind-of is, but it's really about the minibuffer.  :)  I think this package may become very popular, but discovery is one of the hardest problems, so it's important to use as descriptive a name as possible.

* The specified type of this defcustom is probably incorrect:

```el
(defcustom mini-frame-show-parameters '((left . 0.5)
                                        (top . 0)
                                        (width . 1.0)
                                        (height . 1))
  "Frame parameters wich will be applied to mini frame on show.
Unless background color is specified it will be set to result of
`mini-frame-background-color-function'."
  :type '(choice alist function))
```
It should probably just be `:type 'alist`, unless you want to get more specific about frame parameters.  For example, you can see how some specific frame parameters are specified in this defcustom in Yequake: https://github.com/alphapapa/yequake/blob/16dd29c68f4b59ee2da7b7c13d17a687c9c79e16/yequake.el#L100

* The same applies to the defcustom for `mini-frame-completions-show-parameters`.

* It's really slick the way you change the background color to make the minibuffer frame stand out!

* And using another frame for completions is great!

Great work on this package!

alphapapa:
One more quick suggestion: in the minibuffer frame, you might want to do something with the fringe colors.  For example, here's a screenshot of how it looks for me.  Notice that the fringes are not lightened, so they appear as dark columns next to the buffer content, which looks a little jarring.  When I first saw them in the minibuffer frame, I didn't know what they were; only after seeing the completions buffer did I realize they were the fringes.

miniframe

This is really cool!   It even works well with Helm as far as I can tell.

tumashu:
I don't consider mini-frame as ivy-posframe competitor. But mouse clicks is working in mini-frame with ivy unlike in ivy-posframe :)

really? wow

muffinmad:
@alphapapa Thanks!

> * I'd suggest naming the package `minibuffer-in-frame`, or maybe `minibuffer-on-top`, because `mini-frame` doesn't necessarily suggest that its purpose is to change the minibuffer.

Naming is hard indeed :) `mini-frame` is like shortened "**mini**buffer in child-**frame**". Also it is follow the mini-window concept (see docstring of `resize-mini-window` variable). I'd like to keep it because it's pretty short, if there will be no principal objections.

> * The specified type of this defcustom is probably incorrect:
> It should probably just be `:type 'alist`, unless you want to get more specific about frame parameters.

Before `resize-mini-frames` option discovering, the `height` frame parameter was calculated like `(if minibuffer-completion-table icomplete-prospect-height 1)`. This way mini-frame was tall enough to fit completions.

By allowing to choose between alist and function one can do simple frame positioning like
```elisp
(custom-set-variables
 '(mini-frame-show-parameters
   '((top . 10)
     (width . 0.7)
     (left . 0.5))))
```
or provide parameters alist in more complex ways, e.g.:
```elisp
(custom-set-variables
 '(mini-frame-show-parameters
   (lambda ()
     `((width . 0.5)
       (left . ,(pcase (random 3)
                  (0 0)
                  (1 0.5)
                  (3 1.0)))
       (top . ,(pcase (random 3)
                 (0 0)
                 (1 0.5)
                 (3 1.0)))))))
```
Docstring should mention use cases of the function as value of the option. But docstrings are almost as hard as naming :)

muffinmad:
> One more quick suggestion: in the minibuffer frame, you might want to do something with the fringe colors.

Oh, my theme set `unspecified` as fringe background, so I haven't notice this issue. I will look into changing fringe color for those frames. I may suggest workarounds for you:
- Set background color of fringe to `unspecified` so it will follow frame background color like:
```elisp
(custom-set-faces '(fringe ((t (:background unspecified :foreground "gray26")))))
```
- Disable fringe for those frames by adding `(left-fringe . 0)` and `(right-fringe . 0)` to `mini-frame-show-parameters`

> This is really cool! It even works well with Helm as far as I can tell.

Helm buffer appears in the bottom of current frame while minibuffer "moved" to top. Maybe helm-posframe will handle Helm better.
Or maybe some workarounds must be added to mini-frame to show Helm buffer in child frame like it does for "`*Completions*`" buffer.

muffinmad:
> > One more quick suggestion: in the minibuffer frame, you might want to do something with the fringe colors.
> 
> Oh, my theme set `unspecified` as fringe background, so I haven't notice this issue. I will look into changing fringe color for those frames.

Done. `background` attribute of `fringe` is set to `nil` for those frames.

alphapapa:
> Naming is hard indeed :) `mini-frame` is like shortened "**mini**buffer in child-**frame**". Also it is follow the mini-window concept (see docstring of `resize-mini-window` variable). I'd like to keep it because it's pretty short, if there will be no principal objections.

I do understand how the name was formed by shortening.  :)  The issue is that, when a potential user, who might be very interested in moving the minibuffer to the top of the frame, is looking for interesting packages to install, the name `mini-window` will not convey any information about the package affecting the minibuffer.  So the package will probably not find nearly as large an audience as it would if it had a more descriptive name.  Of course, it's your package, so it's your decision.  But I can't encourage you enough to choose as descriptive a name as possible.  :)

> By allowing to choose between alist and function one can do simple frame positioning like...or provide parameters alist in more complex ways, e.g.:

Thanks for the example, that's very cool.  What I mean is, the `alist` specification could be more specific, to guide users toward setting those specific keys and values.

Thanks for fixing the fringe issue!

tarsius:
I for one like the name. Reminds me of my abandoned [ministrut](https://github.com/emacsattic/ministrut).

riscy:
Looks great!  The only thing MELPA would add is to ask you to add a LICENSE (or COPYING) file to your repository that the GitHub API can pick up (e.g. https://github.com/licensee/licensee - compatible).  After that this is an easy merge.

Quickly spellchecking docstrings caught a couple small things:
- mini-frame-show-paremeters -> parameters
- ommited -> omitted

muffinmad:
@riscy Thanks!

> Looks great! The only thing MELPA would add is to ask you to add a LICENSE (or COPYING) file to your repository

Done.

> Quickly spellchecking docstrings caught a couple small things:

Fixed.

PR #6713, opened 2020-02-20
Recipe: Add sbuffer

alphapapa:

Sbuffer is like Ibuffer, but using magit-section to group buffers in a very flexible way.

https://github.com/alphapapa/sbuffer.el

Author and maintainer.

### Checklist

Please confirm with `x`:

### Notes

* Sbuffer is named like Ibuffer, with an "S" for "sections."  However, there is an unrelated package on MELPA named `s-buffer`, which I just discovered.  It is not depended on by any other MELPA package, and it doesn't have very many installs, so I'm not sure if the similar name is a problem.

  I'm willing to consider renaming this package, but naming things is hard...  :)  I thought about `mbuffer`, since it uses Magit sections, but there is also an `m-buffer` package, which is similar to `s-buffer` (!).  So I'm open to suggestions here.

Thanks.

### DIFF
--- a/recipes/sbuffer
+++ b/recipes/sbuffer
@@ -1,0 +1,1 @@
+(sbuffer :fetcher github :repo "alphapapa/sbuffer.el")


conao3:
Hi!

alphapapa:
@conao3 Not sure whether to be glad there were no issues found, or mildly annoyed at noise from your bot posting that.  :P

conao3:
I'm not a bot :)
Manually clone, byte-compile and read your whole code.

But I think such bots are useful for speeding up the flow of MELPA.
A typical suggestion that can be analyzed statically. I want to use the Webhook feature, but only the MELPA maintainer has permission...

alphapapa:
Haha, okay, I guessed you had written a bot since the responses were so similar.

I'm guessing it wouldn't be too difficult to write a GitHub Actions script that would run makem.sh or whatever on new package recipes (i.e. it would checkout the new package into a temporary directory and run against its code).  Then much of these reviews could be automated and viewable by recipe submitters.

conao3:
Yes exactly.
I have noticed in the last few weeks that the warnings I am seeing are different from the warnings the package owner is seeing sometimes.

I think they probably customize the presence or absence of warnings.
Because of these issues, it is important to build the package in a clean, standard environment and check for warnings.

alphapapa:
Right, that's why I suggest, e.g. running `./makem.sh --sandbox lint`, which always uses the latest version of `package-lint` and `relint` and is unaffected by the user's local config.

alphapapa:
I'm withdrawing this PR temporarily because I'm expanding the scope of the package and am looking for a better name.  If anyone's interested, it's going to have a kind of frame-based workspace-switching functionality using rules to categorize buffers into workspaces automatically (i.e. the Sbuffer groups).  Naming suggestions are welcome.

PR #6715, opened 2020-02-20
Document policy on inclusion of files like readmes and changelogs.

Alexander-Miller:
As discussed in https://github.com/melpa/melpa/pull/6681


purcell:
Thanks Alex!

PR #6717, opened 2020-02-22
Add recipe for buffer-wrap.

jcs090218:

Wrap the beginning and the end of buffer.

https://github.com/jcs-elpa/buffer-wrap

The maintainer.

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/buffer-wrap
+++ b/recipes/buffer-wrap
@@ -1,0 +1,1 @@
+(buffer-wrap :repo "jcs-elpa/buffer-wrap" :fetcher github)


alphapapa:
What does this package do?

jcs090218:
I actually do not know how to explain this. This is a simple (maybe stupid) package that make the buffer wrap from the beginning and the end of line, like the gif below.

bw_demo

I think this is quite useful for any buffer similar to this. For example, the `neotree` (the gif above), `buffer menu`, `package-list`, etc. Basically just an interface that opens to all buffer. I found people doing this over and over again is quite stupid.

From doing this:



To this:



---

Just the idea of sharing code and interface. Thanks!

alphapapa:
So it's basically this?

```el
(forward-line 1)
(when (eobp)
  (goto-char (point-min)))
```

jcs090218:
And the opposite, yes.

alphapapa:
Have any of those packages you mentioned expressed interest in depending on another package to implement this functionality?

jcs090218:
> Have any of those packages you mentioned expressed interest in depending on another package to implement this functionality?

I think it depends on the user not the package itself. You can choose to have it implemented or not. And I think this is the best part.

riscy:
In my humble estimation this is okay, because this sort of behavior is orthogonal to the behavior of other modes, and could make it easy for someone to add hooks into whatever modes (magit-mode, dired-mode, whatever) to enable this.

I will mention though that `line-number-at-pos` is quite slow on large buffers (if you dig through the source, it effectively runs a linear scan through an open buffer)!  I can only really see people using this on smaller menu-like buffers, so I don't think that's an issue.  But it's often a good exercise to try to avoid `line-number-at-pos` and compare locations in terms of points, when possible.

A couple other small (non-blocking) comments:
- Re: docstrings like e.g. ``` "Post command for `buffer-wrap'." ``` -- the back-front quoting (``` ` ' ```) is typically used to create a link in the online Emacs help, so should be used around a symbol that actually exists.
- Re: docstrings like ``` "Enable 'buffer-wrap-mode." ``` -- the leading apostrophe seems like a trick to get around checkdoc's nagging.  But I think it would be fine to just write "Enable buffer-wrap.", no tricks required.

PR #6718, opened 2020-02-22
Add recipe for learn-ocaml

erikmd:

This package is a minor mode to interact directly from Emacs (`>= 25.1`) with the [learn-ocaml](http://ocaml-sf.org/learn-ocaml/) platform, developed by the OCaml software foundation to provide students with automated feedback in OCaml laboratory sessions.

(in particular, we plan to start using this mode on next week in Toulouse :)

https://github.com/pfitaxel/learn-ocaml.el

I am the maintainer of `learn-ocaml.el`.

None needed.

### Checklist

Please confirm with `x`:

- [x] I've used the latest version of [package-lint](https://github.com/purcell/package-lint) to check for packaging issues, and addressed its feedback (→ [done](https://github.com/pfitaxel/learn-ocaml.el/commit/099a539ead361f03784de0c421ee4a6ab85df328))
- [x] `M-x checkdoc` is happy with my docstrings (→ 0 error)

Cc @yurug @leunam217 FYI

### DIFF
--- a/recipes/learn-ocaml
+++ b/recipes/learn-ocaml
@@ -1,0 +1,4 @@
+(learn-ocaml
+ :fetcher github
+ :repo "pfitaxel/learn-ocaml.el"
+ :files (:defaults "LICENSE"))
--- a/recipes/learn-ocaml
+++ b/recipes/learn-ocaml
@@ -1,4 +1,3 @@
 (learn-ocaml
  :fetcher github
- :repo "pfitaxel/learn-ocaml.el"
- :files (:defaults "LICENSE"))
+ :repo "pfitaxel/learn-ocaml.el")


riscy:
Great, and thanks for your close adherence to the checklist!

There's a couple things you might want to clean up around conditional handling (my heuristic is to search for `progn` and, often, when I see that, there's a chance to make the code simpler):
- You may want to flip around the `if` on line 743 -- i.e. remove the `not` from the guard and handle the `new-server-value`-non-nil case first, which will also let you remove the `progn`.
- The `if` on line 772 could be a `when` -- then you don't need `progn`, and you don't need to specify an "else" part using `nil`.
- The collection of `ifs` spanning lines 274 to 291 explicitly return `nil` but that's not necessary (leaving that part empty implies a nil).  Even better, you could use a `when` instead of an `if`, e.g.
    ```elisp
        (server-option (when server (concat "--server=" server)))
        (token-option (when token (concat "--token=" token)))
        (local-option (when local "--local")) ;...
    ```
- I believe this check `(if (bound-and-true-p learn-ocaml-mode)` in `define-minor-mode learn-ocaml-mode` is overkill, since `learn-o-caml-mode` will definitely be bound.  You can just check its truthiness without the `bound-and-true-p` wrapper.

Minor:
- Utilitary -> Utility I believe
- occured -> occurred
- I notice the warning message `"An error occured when executing the last command, Do you want to open the log to have more information?"` is quite long (longer than 80 characters) -- could it be a one-liner, like `"`learn-ocaml.el' encountered an error.  Open log?"`

erikmd:
Thanks @riscy for your feedback! Really appreciate. So I pushed a new version in `master`.

erikmd:
Hi @riscy, does it look good to you with the latest changes? (sorry to ping you once more… but it'd be really great if `learn-ocaml` could be on MELPA for the first lab sessions of this week :+1:)

riscy:
I'm quite backlogged with the MELPA PR's, but this was a fair reason to give it a bump in priority.

I think it may be a good idea to look into using `shell-quote-argument` within `learn-ocaml-command-constructor` as well such as on the `token`, `server`, and `id` options to handle cases where people might inadvertently add characters like & or ' into these fields -- if that's even possible? (Apologies for overlooking this on the first pass.)

In any case, have fun with the course. :)

erikmd:
Great, thanks @riscy ! :-)

regarding your comment

> I think it may be a good idea to look into using `shell-quote-argument` within `learn-ocaml-command-constructor` as well such as on the `token`, `server`, and `id` options to handle cases where people might inadvertently add characters like & or ' into these fields -- if that's even possible? 

even if values with those chars wouldn't be valid, my initial thought was OK 'better safe than sorry', so I've tried adding `shell-quote-argument` there, but it doesn't work because `learn-ocaml-command-constructor` just constructs a list that is ultimately used by `make-process`, which does not require escaping (instead of `shell-command-to-string` that would just read a string), and actually fails if the `:` in the URL is quoted → `(shell-quote-argument "https://example.com") = "https\\://example.com"`
so to sum up, the current version of that function sounds OK.

PR #6722, opened 2020-02-25
added krain recipie

Knarkzel:

This is a package for organizing org-mode notes automagically. It is currently
in beta.

https://gitlab.com/knarkzel/krain

I am the creator and maintainer of this package.

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/krain
+++ b/recipes/krain
@@ -1,0 +1,1 @@
+(krain :repo "knarkzel/krain" :fetcher gitlab)


alphapapa:
This seems lo have significant overlap with existing tools like `org-journal`, `org-datetree`, `org-velocity`, `org-rifle`, `org-ql`, `deft`, `helm`, `swiper`, `grep`, etc.  It's probably not appropriate for publishing on MELPA at such an early stage of development.  If you develop it further into a unique package, you could resubmit it then.  However, please consider exploring existing tools and contributing to them before publishing a very similar package.

BTW, a few Elisp tips:

* Consider `format-time-string` rather than `(shell-command-to-string "date ...`.
* Keymaps should be defined with `defvar`.  Generally there shouldn't be any top-level `setq` forms.
* Consider `directory-files-recursively` rather than `(shell-command-to-string (concat "find...`.
* Consider `(krain-mode)` rather than `(funcall 'krain-mode)`.

conao3:
Hi!

```diff
diff --git a/Cask b/Cask
new file mode 100644
index 0000000..c6695a4
--- /dev/null
+++ b/Cask
@@ -0,0 +1,4 @@
+(source gnu)
+(source melpa)
+
+(package-file "krain.el")
```

```shell
$ cask; cask build
Package operations: 1 installs, 0 removals
  - Installing [ 1/1] emacs (26)... already present
Compiling /home/conao/dev/forks/krain/krain.el...

In toplevel form:
krain.el:49:1:Error: Cannot open load file: No such file or directory, counsel
```

Cannot build your package. Maybe missing `Package-Requires`?

riscy:
I agree with @alphapapa that this looks like it's in early development and there's quite a few mature packages in the orgverse with similar patterns.

That's not to discourage you from continuing to develop this, though.  And in the future, if this progresses, this could definitely be reopened.

In the meantime, here's some quick feedback:

- Since you rely on hydra and counsel, you'd need to add those to your `Package-Requires` lines too
- `(defvar krain-directory "~/downloads/krain" "Directory where `krain' stores notes.")` seems like it would by default require someone to have a downloads directory, so maybe too specific of a default for a general audience?
- I love ripgrep, but not everyone will have it installed.  What does `counsel-rg` do when it's not installed?  Maybe more importantly, you send ` "rg -N --no-filename topic " ` directly to the shell, which will fail unless the user has installed ripgrep, so maybe worth adding to your docs.
- Variable/function names, especially for MELPA packages, shouldn't have special characters like asterisks around them (e.g. *krain-highlight*)
- Rather than using a top-level `setq` for e.g. *krain-highlight* consider checking out `defvar`

I think the latter point would solve these byte-compile warnings:
```
krain.el:114:7:Warning: assignment to free variable `*krain-highlight*'
In krain-mode:
krain.el:123:31:Warning: reference to free variable `*krain-highlight*'
```

Hope that helps -- feel free to comment here and reopen if any of us has missed something, e.g. there's a use case we've missed that krain is cornering and other existing packages are missing.

PR #6723, opened 2020-02-25
Add separedit

twlz0ne:

Edit comment or string/docstring or code block inside them in separate buffer with your favorite mode.

Usage:

- Move the cursor to comment or string or doc
- Press <kbd>C-c C-'</kbd>

https://github.com/twlz0ne/separedit.el

Author

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/separedit
+++ b/recipes/separedit
@@ -1,0 +1,4 @@
+(separedit
+ :fetcher github
+ :repo "twlz0ne/separedit.el"
+ :files ("separedit.el"))


alphapapa:
I'm pretty sure there's already a package on MELPA that does this, but I can't recall the name.

twlz0ne:
@alamaison Thanks for telling me this package.

And then I found out that other people had metioned it in my [post](https://emacs-china.org/t/separedit-el-comment-edit-el-docstring/11196/16?u=twlz0ne) but I didn't notice.

The `poporg` seems great, if I had known it earlier, I might not have started the `separedit.el`.

But after after a simple trial, I found there are some shortcomings:

- Dependency on the `extradoc.py` tool to extracting text. It makes the editing process a bit less fluit.
- Cannot edit code blocks inside string/comments directly. 
- Not support for nested editing.
- Cannot select edit mode manually.

alphapapa:
> * Dependency on the `extradoc.py` tool to extracting text. It makes the editing process a bit less fluit.

I can't find any references to that script or calling external processes in https://github.com/QBobWatson/poporg/blob/master/poporg.el.

twlz0ne:
@alphapapa Sorry for not checking carefully.

But I ’m sure I felt a little bit less fluent during the trial. And I see it creates temporary files every time:

```
Saving file /var/folders/86/9youk1oabz8ng7hc4sc6j58h0000gn/T/poporg-SkR0x0...
Wrote /var/folders/86/9youk1oabz8ng7hc4sc6j58h0000gn/T/poporg-SkR0x0
```

Maybe that is the problem.

alphapapa:
I guess that's caused by this code:

https://github.com/QBobWatson/poporg/blob/2c58d68c81ecca4140bf179f19ed153ec804b65a/poporg.el#L649-L654

I don't think that's necessary to use undo, so if that's all it's for, that would be a good candidate to send a patch to improve.

fbergroth:
> I'm pretty sure there's already a package on MELPA that does this, but I can't recall the name.

`edit-indirect` and `literal-string` has some similarities to this package.

twlz0ne:
@fasheng I used `edit-indirect` as an library in `separedit` (same as [`markdown-mode`@e9dff50d#L8670](https://github.com/jrblevin/markdown-mode/blob/e9dff50d572caa96b68a7466c18c97a8d6ed651c/markdown-mode.el#L8670)).

Strictly speaking, the two are not at the same level.  `edit-indirect` is more primitive, and `separedit` provides more convenient features. I prefer to compare `separedit` with `poporg`.

twlz0ne:
> so if that's all it's for, that would be a good candidate to send a patch to improve.

If I didn't start the `separedit`, maybe I will send the patch to `poporg`.

But since the beginning of `separedit`, I have more ideas. If let me consider `poporg` at this point, I find that it doesn't quite match my expectations, I've also listed some of them [above](https://github.com/melpa/melpa/pull/6723#issuecomment-591265351), not only the "creating temporary files" problem.

tarsius:
Also [edit-indirect](https://github.com/Fanael/edit-indirect).

fasheng:
I read the Emacs China forum [post](https://emacs-china.org/t/separedit-el-comment-edit-el-docstring/11196) above just now, it shows people like some unique features provide by this package. And I like the `nested editing` and `graph code block` the most:

```
,--- elisp
| (foo \"bar\")
`---
```

Consider the packge is well maintained and very likely some new ideas will be implemented in the future, and contrib/debug such features to `poporg` will take a lot of extra time. I prefer to accept `separedit` in such case, IMHO.

conao3:
Hi!

```shell
$ cask; cask build
  - Installing [ 1/2] edit-indirect (latest)... done
  - Installing [ 2/2] dash (latest)... done
Compiling /home/conao/dev/forks/separedit.el/cask-bootstrap.el...

In end of data:
cask-bootstrap.el:22:1:Warning: the function
    ‘display-warning@ignore-load-path-warning’ is not known to be defined.
Compiling /home/conao/dev/forks/separedit.el/separedit-test-helper.el...

In toplevel form:
separedit-test-helper.el:43:1:Warning: Unused lexical argument ‘end’

In xah-syntax-color-hex:
separedit-test-helper.el:150:4:Warning: ‘font-lock-fontify-buffer’ is for
    interactive use only; use ‘font-lock-ensure’ or ‘font-lock-flush’ instead.

In end of data:
separedit-test-helper.el:280:1:Warning: the following functions are not known
    to be defined: sh-indent-line, separedit, should, separedit--block-info,
    separedit--string-region, separedit--remove-escape,
    separedit--restore-escape, separedit--remove-comment-delimiter,
    separedit--comment-delimiter-regexp
Compiling /home/conao/dev/forks/separedit.el/separedit.el...
```

twlz0ne:
@conao3 these files are test resources.

- `cask-bootstrap.el` &`cask-checkinstall.el` scripts for people who don't have cask installed but want to try running tests
- `separedit-test-helper.el` some utils for helping generate test case or make interference.
- `run-test.example.sh` an example script for testing package in multiple version of Emacs.

dakra:
Maybe someone should make a comparison of all those string-edit libraries.
I looked at them a while ago but never really got around to really compare them.
I found:

- [poporg](https://github.com/QBobWatson/poporg)
- [string-edit](https://github.com/magnars/string-edit.el)
- [Edit Indirect](https://github.com/Fanael/edit-indirect)
- [Fence Edit](https://github.com/aaronbieber/fence-edit.el)
- [see-mode](https://github.com/marcelino-m/see-mode)
- [literal-string](https://github.com/joodie/literal-string-mode)

riscy:
Great discussion above! :)  I'm feeling this package is well documented and substantial enough to stand on its own and develop its own direction, even though there are a few similar packages.  It even looks like it already has a nontrivial userbase.

The list @dakra provides is bigger than I thought -- I agree it would be great if someone went on a safari and tried them all out...

Minor, but non-blocking notes for @twlz0ne:

- separedit.el#L308: It's safer to sharp-quote function names; use `#'`
- separedit.el#L314: It's safer to sharp-quote function names; use `#'`
- separedit.el#L724: It's safer to sharp-quote function names; use `#'`
- separedit.el#L430: It's safer to sharp-quote function names; use `#'`
- separedit.el#L23: Prefer `https` over `http` (if possible)
- determing -> determining
- eidt buffer -> edit
- prevous -> previous
- funcitons -> functions
- otherwish -> otherwise
- edting -> editing

tarsius:
@twlz0ne The `cask-bootstrap.el` file is weird. I recommend you remove it.

I am not sure what purpose it serves but if its important to your workflow, then make these changes:

* Do not name it `cask-bootstrap.el`. Cask itself comes with a file by that name and your file conflicts with that.
* Do not provide *any* feature in the file. This is not a library but some maintainance script and it therefore should not provide a feature.

twlz0ne:
@riscy Thanks for your careful review.

@tarsius Thanks for your suggestion, I will make adjustments next commit.

twlz0ne:
@riscy @tarsius How can I edit the summary of the package on MELPA?

Screenshot_2020-03-19_at_3 54 14_PM

tarsius:
[Here](https://github.com/twlz0ne/separedit.el/blob/master/separedit.el#L1)

tarsius:
Like `cask-bootstrap.el` (see above) `test-bootstrap.el` is a test script not an end-user library and should therefore not provide a feature. This is especially so because both of your `playonline` and `separedit` packages come with such a file. Please also look at your other testing scripts in these and other repositories and make sure they don't provide a feature and are instead loaded using `load-file` or `load`.

twlz0ne:
Thank you for reminding again. All `provide` statements are now removed from test script:

``` sh
⋊> curl -G 'https://api.github.com/search/code' \
        --data-urlencode 'q=provide user:twlz0ne filename:bootstrap.el filename:checkinstall.el'
{
  "total_count": 0,
  "incomplete_results": true,
  "items": [
  ]
}
```

PR #6725, opened 2020-02-27
Add srfi recipe

lassik:

Provides quick access to Scheme Requests for Implementation (SRFI)
documents from within Emacs:

* `M-x srfi-list` brings up a `*SRFI*` buffer listing all SRFIs.

* `M-x srfi` does the same, but allows live-narrowing the list by
  typing numbers or words into the minibuffer.

* The `*SRFI*` buffer provides one-key commands to visit the SRFI
  document, its discussion archive (mailing list archive), and it
  version control repository. These commands open the right web
  pages in your web browser using `browse-url`.

https://github.com/srfi-explorations/emacs-srfi

Wrote it.

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/srfi
+++ b/recipes/srfi
@@ -1,0 +1,1 @@
+(srfi :fetcher github :repo "srfi-explorations/emacs-srfi")


riscy:
Quick first pass --

### srfi-data.el ###

- No issues!

- No issues!

### srfi.el ###

- No issues!

- No issues!

- No issues!

- srfi.el#L116: This define-key could overwrite a user's keybindings.  Try: `(defvar my-map (let ((km (make-sparse-keymap))) (define-key ...) km))`
- srfi.el#L117: This define-key could overwrite a user's keybindings.  Try: `(defvar my-map (let ((km (make-sparse-keymap))) (define-key ...) km))`
- srfi.el#L118: This define-key could overwrite a user's keybindings.  Try: `(defvar my-map (let ((km (make-sparse-keymap))) (define-key ...) km))`
- srfi.el#L119: This define-key could overwrite a user's keybindings.  Try: `(defvar my-map (let ((km (make-sparse-keymap))) (define-key ...) km))`
- srfi.el#L120: This define-key could overwrite a user's keybindings.  Try: `(defvar my-map (let ((km (make-sparse-keymap))) (define-key ...) km))`
- Prefer the default recipe, especially for small packages.  I think you can get away with it in this case?

lassik:
> This define-key could overwrite a user's keybindings.

Fixed. Thanks!

lassik:
> Prefer the default recipe, especially for small packages. I think you can get away with it in this case?

srfi-data-convert.el is only useful for package maintainers who need to run it in order to generate srfi-data.el based on the upstream data file srfi-data.scm. (srfi-data-convert.el translates Scheme syntax to Emacs Lisp syntax and throws away information that srfi.el does not need). That's why I specified the MELPA recipe to exclude srfi-data-convert.el

recipes/srfi
```diff
@@ -0,0 +1,2 @@
+(srfi :fetcher github :repo "srfi-explorations/emacs-srfi"
+      :files ("srfi.el" "srfi-data.el"))
```
conao3: `:files (:default (:exclude "srfi-data-convert.el"))` is better.
Manage with whitelist, you need to modify recipe if you split package file.
Manage with blacklist is you don't need to modify the recipe.

riscy:
Good point re: `:exclude` @conao3.  Another option would be to put that special file inside its own directory, which MELPA would automatically ignore, but that's not a blocker.

I think to be evenhanded in my insistence for license boilerplate, we should consider srfi-data.el to be code, and as such it should have some boilerplate or a SPDX identifier.

Beyond that I think this is good to go, so ping me when ready. :)

lassik:
> Good point re: `:exclude` @conao3. Another option would be to put that special file inside its own directory, which MELPA would automatically ignore, but that's not a blocker.

OK, I moved `srfi-data-convert.el` (and the trivial shell script wrappers assisting it) into an `admin` subdirectory so they are excluded by the default recipe. Force-pushed this PR to update the recipe.

> I think to be evenhanded in my insistence for license boilerplate, we should consider srfi-data.el to be code, and as such it should have some boilerplate or a SPDX identifier.

OK, the generator now outputs `Copyright` and `SPDX-License-Identifier` lines into `srfi-data.el`.

> Beyond that I think this is good to go, so ping me when ready. :)

Thanks a lot!

lassik:
@riscy Oh, and here's the ping :)

riscy:
Thanks!

PR #6728, opened 2020-03-01
Test a PR against CI (this is only a test)

riscy:
Opened this PR to dip my toe into MELPA CI.  (merge target is `test_melpazoid` -- this won't go into master).


riscy:
Ok, no python 3.6 on the current melpa build.

PR #6730, opened 2020-03-05
update location of minsk-theme

jlpaca:
recipe originally added in https://github.com/melpa/melpa/pull/6556.

the repository is expected soon also to contain implementations of the colour theme for other applications, so the recipe is updated to point to the files specific to emacs.

### Checklist

### DIFF
--- a/recipes/minsk-theme
+++ b/recipes/minsk-theme
@@ -1,1 +1,1 @@
-(minsk-theme :repo "jlpaca/minsk-theme" :fetcher github)
+(minsk-theme :fetcher github :repo "jlpaca/minsk-theme" :files ("theme/emacs/minsk-theme.el"))


riscy:
This can be merged fast but I question (without accusation) the choice to have a separate branch for each application, in this case "emacs".  Would you have a separate branch for every application?  If you update your README or your license, will you have to cherry-pick the commit into the branch for each application you support?

I would strongly recommend sticking to the default (master) branch, but let me know what you think.

jlpaca:
@riscy that was the intention, yes, but I now feel like it's indeed not worth the trouble. the recipe has been updated to pull from the default branch instead.

the travis build failed, I think, because it was trying to pull the file from the new location, which did not exist in the old `master`. that should now also be fixed.

riscy:
Great!  While I've got your attention, it may be worth adding license boilerplate above the `;;; Commentary` section to your .el file, just in case that's the *only* document they receive, even something like:
```
This work is licensed under the terms of the MIT license.  
For a copy, see <https://opensource.org/licenses/MIT>.
```

Very cool theme, btw. :)

PR #6731, opened 2020-03-06
Add recipe for emoji-github.

jcs090218:

Display list of GitHub's emoji. (cheat sheet)

https://github.com/jcs-elpa/emoji-github

The maintainer.

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/emoji-github
+++ b/recipes/emoji-github
@@ -1,0 +1,1 @@
+(emoji-github :repo "jcs-elpa/emoji-github" :fetcher github)


conao3:
Hi!

Hi!

1. Please fix major-mode customize-group.

---

2. Please consider removing note from summary.

package-lint warn if the summary ends with period.
```emacs-lisp
 emoji-git…     1   1 warning         The package summary should not end with a period. (emacs-lisp-package)
```

I suggest this diff.
```diff
modified   emoji-github.el
@@ -1,4 +1,4 @@
-;;; emoji-github.el --- Display list of GitHub's emoji.  (cheat sheet)  -*- lexical-binding: t; -*-
+;;; emoji-github.el --- Display cheat sheet of GitHub's emoji.  -*- lexical-binding: t; -*-
 
 ;; Copyright (C) 2020  Shen, Jen-Chieh
 ;; Created date 2020-03-05 10:58:18
```

riscy:
I had a quick look at the gist on which this is based, and it looks like some of the emojis are outdated - for example the commenters pointed out`:couplekiss:`.  Other emojis that GitHub _does_ support are missing from the list, for example :zambia:.  

Helpfully, one commenter in the gist pointed out that GitHub supports an API that directly lists the supported GitHub emojis -- here: https://api.github.com/emojis.

What do you think about making this package use the API to get an up-to-the-minute list of GitHub's emoji, rather than relying on a potentially outdated list?  I think that would be more useful altogether.

jcs090218:
Thanks for the feedback! I implemented the way you expected here => [a0f2e6](https://github.com/jcs-elpa/emoji-github/blob/a0f2e61a80ebd8a3595e376b62f0d50a4150b3d1/emoji-github.el) but I realized I like how emoji are sorted by category! Unless I figured it out the way to sort these emoji or github api has the sorted emoji api call. I might have to say NO this time.

Furthermore discuss the usefulness of this package is subjective. In my own experience, I like how it sorted by the type and not by alphabetic order.

riscy:
I hear you -- but I think for this package to make sense for MELPA, you should be listing all of the emojis available on GitHub and removing the emojis that GitHub doesn't currently support (even though they may be listed in that gist).

What about this as a solution: pull down the emoji list from the API.  Your package puts each emoji in a category (people, nature, etc.) based on a list that you maintain.  If your package finds emojis that don't fit in a category it knows about, it puts them in an "Uncategorized" category.  Then, maintaining the package implies maintaining the categories.  What do you think?

jcs090218:
This make sense to me. I will implement this ASAP. Thanks!

jcs090218:
Okay, done! I have all the unlisted emoji under the `Others` category. Thanks! :D

riscy:
Thanks, looks good!  Perhaps you have plans to sort the other emojis?  It would make things even more useful, and should probably considered part of maintaining a package like this.

Another couple of non-blockers:
- emoji-github.el#L375: There is no need to quote lambdas (neither #' nor ')
- Also you may want to explicitly `(require 'subr-x)` for string-empty-p,

PR #6732, opened 2020-03-06
Add sexp-diff recipe

xuchunyang:

Diff two sexps

https://github.com/xuchunyang/sexp-diff.el

I'm the maintainer.

None needed

### Checklist

Please confirm with `x`:

<s>I fixed some checkdoc warnings, but for other warnings, I don't plan to fix, make checkdoc 100% happy is too strict.</s> All checkdoc warnings is now fixed.

### DIFF
--- a/recipes/sexp-diff
+++ b/recipes/sexp-diff
@@ -1,0 +1,1 @@
+(sexp-diff :fetcher github :repo "xuchunyang/sexp-diff.el")


conao3:
Hi!

```emacs-lisp
 sexp-di…   109  31 error           You should depend on (emacs "26.1") if you need `record'. (emacs-lisp-package)
 sexp-di…   117  38 error           You should depend on (emacs "26.1") if you need `record'. (emacs-lisp-package)
 sexp-di…   129  29 error           You should depend on (emacs "26.1") if you need `record'. (emacs-lisp-package)
```

But this warning is false-positive.
Please consider using `obj` instead of `record` to suppress warnings.

xuchunyang:
> Please consider using `obj` instead of `record` to suppress warnings.

Thanks for the suggestion, I probably follow this suggestion on another day, but right now I am not very happy to change my code just to walk around package-lint's bug.

conao3:
Of course, I agree.
However, since passing package-lint is a MELPA requirement, you should be aware of package-lint output.

And, Of course, @riscy will agree and merge this PR that this warning is a false positive, but keeping this warning visible will make it harder to notice other warnings.
I think Elisp package authors should always enable `package-lint` and follow `package-lint` suggestions.

tarsius:
> I think Elisp package authors should always [...] follow `package-lint` suggestions.

Not when those suggestions are wrong, no.

----

Only headings should begin with three or more semicolons. Please remove the access semicolons used in the library headers. That way `oulline-minor-mode` won't treat each line of the commentary as a heading.

xuchunyang:
> Only headings should begin with three or more semicolons. Please remove the access semicolons used in the library headers. That way `oulline-minor-mode` won't treat each line of the commentary as a heading.

Done.

riscy:
This is a cool package!

The `record` thing is obviously not a blocker, but another very small reason to rename the variable might be to play ball with myopic font-locking:

image

This issue also happens in e.g. python (in which variables and functions share a namespace, but still) where `dir` is a special function but also a convenient variable name, and the consensus there is to write `dir_`.

PR #6733, opened 2020-03-06
Add curl-to-elisp recipe

xuchunyang:

Convert curl command into elisp code

https://github.com/xuchunyang/curl-to-elisp

I'm the maintainer.

None needed

### Checklist

Please confirm with `x`:

<s>I am happy with the result of `M-x checkdoc`, though checkdoc isn't, it suggests curl at the beginning in `message` should to Curl , I understand the meaning and choose to ignore this suggestion.</s> Now checkdoc is happy, I just fixed all the warnings.

### DIFF
--- a/recipes/curl-to-elisp
+++ b/recipes/curl-to-elisp
@@ -1,0 +1,1 @@
+(curl-to-elisp :fetcher github :repo "xuchunyang/curl-to-elisp")


conao3:
Hi!

```emacs-lisp
 curl-to…   244  19 error           `mm-url-encode-multipart-form-data' was removed in Emacs version 24.3. (emacs-lisp-package)
```

xuchunyang:
Not sure why I missed this error message when I made this PR, though the error from package-lint does not make sense to me, `mm-url-encode-multipart-form-data` is provided `mm-url.el`, and I have CI to ensure this function is available for all versions Emacs I want to support, maybe it's a bug of package-lint?

conao3:
As package-lint database, `mm-url-encode-multipart-form-data` was added at `24.1`, was removed at `24.3`, and added at `25.1`.
https://github.com/purcell/package-lint/blob/5cfa10041a467ca122f5bda31119a5904d574631/data/stdlib-changes
Is this data is mistaked?

xuchunyang:
Let's assume the data is correct, my package depends on 25.1, maybe package-lint doesn't consider this?

conao3:
It seems to be a problem of `package-lint`.
I thought the `package-lint` problem would be solved elsewhere, so the quality of the other code was very good.
And the parsing of the curl command was very informative to me. Thanks!

riscy:
Looks great.

PR #6741, opened 2020-03-08
added recipie for modern-fringes

SpecialBomb:

modern-fringes is meant to replace the existing emacs fringe bitmaps to what I feel are better looking ones.

https://github.com/SpecialBomb/emacs-modern-fringes

I am the author, and the maintainer.

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/modern-fringes
+++ b/recipes/modern-fringes
@@ -1,0 +1,1 @@
+(modern-fringes :repo "specialbomb/emacs-modern-fringes" :fetcher github)


SpecialBomb:
A note I should make, I think package-lint didn't work for me properly, but I'm not very sure. I was using the latest version of it.

SpecialBomb:
Someone also suggested I should make this a minor mode, rather than defining an init function. I'm unsure of whether that would be the right thing to do or not.

SpecialBomb:
apparently, I can't spell recipe.

alphapapa:
> Someone also suggested I should make this a minor mode, rather than defining an init function. I'm unsure of whether that would be the right thing to do or not.

That is the right thing to do.  Emacs doesn't use `init` and `revert` functions; that's what minor modes are for.

> A note I should make, I think package-lint didn't work for me properly, but I'm not very sure. I was using the latest version of it. I also use emacs 26.3, so that may be part of it.

I recommend using this to lint your packages in a clean environment: https://github.com/alphapapa/makem.sh

SpecialBomb:
The thing was about the minor modes, since I have to modify global data, I
wasn't sure how I would possibly do a buffer specific minor mode. It's more
than possible to make it a global minor mode, however. I'll try working on
that soon, not at home at the moment.

Also, thank you for the build script, I'll mess with it soon.

SpecialBomb:
@alphapapa question, which group in emacs do you think modern-fringes should fit into?

I've implemented the minor mode and I'd like it to show up on the customization screen properly, but I don't really know where to put it.

alphapapa:
Both `convenience` and `frames` seem okay to me.  The official keywords don't seem to be updated often, and they don't seem to cover everything well, so it's not always clear what to use.

SpecialBomb:
convenience it is then!

SpecialBomb:
Alright, the minor mode has been implemented, and all the groups are in place. I bumped the version up to 4.

SpecialBomb:
Alright, I've got everything set up finally. package-lint gave no errors in the end, but one warning that does not affect anything. It warns I should have a keymap for my minor mode, but my minor mode doesn't use any keybinds, so it's fine.

conao3:
Hi!

1. Please consider indent with spaces see [[https://github.com/bbatsov/emacs-lisp-style-guide#source-code-layout--organization][Emacs lisp style guide]]
2. Please consider make {init, revert} function internal function.
3. Please fill paragraph in 80 column.

SpecialBomb:
@conao3 ill start work on that today.

SpecialBomb:
Those fixes are finished, though I would like some elaboration on what "making something an internal function" means. I was guessing that meant not to autoload them, but I'm unsure.

conao3:
Elisp uses a prefix followed by two hyphens to indicate that the function is internal to the package.
The user calling init manually is confusing because the minor mode is already exposed.

SpecialBomb:
@conao3 I understand, working on it now. It should be as simple as renaming the functions.

alphapapa:
To be clear, `--`-prefixed functions are merely a naming convention; it's not enforced by code.  It's a good idea, but it mainly communicates to users which functions are intended to be called from code internal to the package.

SpecialBomb:
Ah, I see.

I have it all sorted on a branch already, I just have to merge it.

Also, thank you all for being so helpful 💜

SpecialBomb:
Merged it to master

riscy:
Great!  Thank you for accommodating the minor mode, it's great for the health of the Emacs ecosystem.

The last thing I'd ask is to add some MIT license boilerplate above your `;;; Commentary` section, or simply [an SPDX line](https://spdx.org/using-spdx-license-identifier) like`SPDX-License-Identifier: MIT` within your author/keywords/URL/version block.

After that this is good to go.

SpecialBomb:
@riscy sure thing, ill do it now.

SpecialBomb:
@riscy should be merged to master

PR #6742, opened 2020-03-08
Add emacs-bazel-mode recipe

jkaye2012:

Provides "official" support for the Bazel build system within Emacs.

https://github.com/bazelbuild/emacs-bazel-mode

A user. 

https://github.com/bazelbuild/emacs-bazel-mode/issues/13 - the package maintainer is aware of this and is in support of it.

### Checklist

Please confirm with `x`:

- [ ] I've used the latest version of [package-lint](https://github.com/purcell/package-lint) to check for packaging issues, and addressed its feedback (I was unable to get this installed on my machine because I do not use ELPA and it appears to carry a dependency there)
- [x] `M-x checkdoc` is happy with my docstrings (there are 5 total errors. 4 are disambiguations that do not seem to make sense in the context of the documentation, and one is that some lines are over 80 columns. Everything else is fine)

### DIFF
--- a/recipes/emacs-bazel-mode
+++ b/recipes/emacs-bazel-mode
@@ -1,0 +1,3 @@
+
+(emacs-bazel-mode :repo "bazelbuild/emacs-bazel-mode" :fetcher github
+  :files ("lisp/*.el" (:exclude "lisp/*-test.el")))


jkaye2012:
Also, it's probably worth noting that the 80+ width checkdoc errors are URLs that cannot really be split. Forgot to mention that, sorry.

conao3:
Hi!

0. Please fill package header correctly.
Main Elisp file is `lisp/bazel-mode.el` isn't it?
Please fill `Version`, `URL`, `Package-Requires` header please.

@jkaye2012 is just a user, you have no push permission?
@bazelbuild, @phst solve this issue? or please inform your thought of adding MELPA to us.

jkaye2012:
Right, I don't have push permissions. I don't mind creating a pull request
for these changes though. Whatever is easiest for everyone.

riscy:
I think this one needs some figuring out --

MELPA has an [existing bazel-mode](https://github.com/codesuki/bazel-mode/) that was updated about 4 months ago.  It has [an open issue](https://github.com/codesuki/bazel-mode/issues/1) in which @codesuki looked open to the code being moved, but I don't think that happened. (@codesuki -- it would be great to hear your thoughts on this.)

Indexing two different packages that use the same prefix for their functions (`bazel-mode`) is something we should avoid, so as not to inflict confusing debugging sessions on users.  But the bazel-mode presented here looks to be more actively developed.

I personally think the best outcome would be for @codesuki to see about merging any enhancements from the "old" bazel-mode into the "new" bazel mode and then ceding maintainership to the bazelbuild org, but let's see how this goes.

In the meantime, the bazel-mode presented here byte compiles cleanly and is appropriately licensed, but needs some packaging metadata that `package-lint` would help out with.

jkaye2012:
Is there a way to install package-lint without elpa? I disabled elpa due to
gpg issues some time ago.

riscy:
I know the issues you're talking about -- it's on MELPA, not ELPA, so that should work.  In the worst case scenario, it has [a GitHub repo](https://github.com/purcell/package-lint) where you can grab the package-lint.el file.

riscy:
I'll remove the "awaiting upstream" so this doesn't fall off the radar.

codesuki:
> I think this one needs some figuring out --
> 
> MELPA has an [existing bazel-mode](https://github.com/codesuki/bazel-mode/) that was updated about 4 months ago. It has [an open issue](https://github.com/codesuki/bazel-mode/issues/1) in which @codesuki looked open to the code being moved, but I don't think that happened. (@codesuki -- it would be great to hear your thoughts on this.)
> 
> Indexing two different packages that use the same prefix for their functions (`bazel-mode`) is something we should avoid, so as not to inflict confusing debugging sessions on users. But the bazel-mode presented here looks to be more actively developed.
> 
> I personally think the best outcome would be for @codesuki to see about merging any enhancements from the "old" bazel-mode into the "new" bazel mode and then ceding maintainership to the bazelbuild org, but let's see how this goes.
> 
> In the meantime, the bazel-mode presented here byte compiles cleanly and is appropriately licensed, but needs some packaging metadata that `package-lint` would help out with.

Thanks for reaching out! Since `bazelbuild/emacs-bazel-mode` is actively developed I am fine with this.

riscy:
Thanks so much @codesuki!  I think in this case we should consider replacing the current `bazel-mode` recipe and getting the other checks up to date.

purcell:
>  I think in this case we should consider replacing the current `bazel-mode` recipe and getting the other checks up to date.

Yes, definitely, let's look to replace. That would involve changing this PR. Want to update the original `bazel-mode` recipe instead, @jkaye2012, and then force-push to update this PR?

>     * I've used the latest version of [package-lint](https://github.com/purcell/package-lint) to check for packaging issues, and addressed its feedback (I was unable to get this installed on my machine because I do not use ELPA and it appears to carry a dependency there)
> Is there a way to install package-lint without elpa? I disabled elpa due to gpg issues some time ago.

Many packages on MELPA depend on ELPA packages, so you should get both working. Try installing the `gnu-elpa-keyring-update` package from there manually, and/or change the `package-check-signature` variable.

jkaye2012:
Yeah, I'll do this tomorrow or Sunday. I ended up fixing the gpg issues
this week so no worries there.

riscy:
@jkaye2012 I have some changes pending: https://github.com/bazelbuild/emacs-bazel-mode/pull/53

riscy:
If you modify the recipe and update the filename and the recipe name to `bazel-mode` then we should be good to go, if that PR is accepted.

riscy:
Closed per 10f3918 -- thanks @jkaye2012

jkaye2012:
No problem - sorry I didn't get to this, the world is a crazy place right
now 🙂

riscy:
Totally understandable.

PR #6751, opened 2020-03-12
Add recipe for git-invoke

lechten:

This project offers a small Emacs Lisp package with functions to invoke Git commands (clone, pull, update of submodules).  Some commands invoke Git and direct output into an output buffer, others return a string; see doc strings of contained functions.  This is not meant for end users but for programmers.

I refactored the functionality contained in this proposed package from [oer-reveal](https://melpa.org/#/oer-reveal) as I also need this for another project.

https://gitlab.com/oer/git-invoke

Maintainer

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/git-invoke
+++ b/recipes/git-invoke
@@ -1,0 +1,4 @@
+(git-invoke
+ :repo "oer/git-invoke"
+ :fetcher gitlab
+ :files (:defaults "LICENSES"))


conao3:
Hi!

However, there are many packages for `git` in MELPA.
For example, `magit` provides an Elisp binding for almost any `git` command.
What are the originality and benefits?
(Most of the answers to this question are going to give to MELPA maintainers (@purcell, @tarsius, @riscy).
I can't judge whether it is useful or not, but I know that there is a similar package in MELPA)

And since users rarely see this PR, the package description included in the repository README.

lechten:
Thank you for looking at this!

Yes, I had a brief look at Git packages but did not see one that meets my needs.  I want to provide this for Emacs 24.5 with Debian 9 (note the [comment in git-invoke--invoke](https://gitlab.com/oer/git-invoke/-/blob/master/git-invoke.el#L57)), while `magit` for example requires Emacs 25.  Also, I’d like to provide functions that can return `string` values.

I’m not sure what you suggest concerning README.  MELPA’s CONTRIBUTING.org recommends not to include such files.

conao3:
Yes. This is not a MELPA requirement.
Users downloading from MELPA will see your repository, not this PR.

Therefore, the package description in this PR can be summarized from the description in the repository.
Currently, the description in this PR is more detailed than the repository README, so I suggest that you write that information in the repository.

### Brief summary in this PR
This project offers a small Emacs Lisp package with functions to invoke Git commands (clone, pull, update of submodules). Some commands invoke Git and direct output into an output buffer, others return a string; see doc strings of contained functions. This is not meant for end users but for programmers.

I refactored the functionality contained in this proposed package from oer-reveal as I also need this for another project.

### Repository readme - [Link](https://gitlab.com/oer/git-invoke/-/tree/c7627e8935ed90aee5e7b22b75f1ec02eddcb6c5)
This project offers a small Emacs Lisp package with functions to
invoke Git commands (clone, pull, update of submodules).
Some commands invoke Git and direct output into an output buffer,
others return a string; see doc strings of contained functions.  This
is not meant for end users but for programmers.

lechten:
Thanks for the suggestion. I [improved](https://gitlab.com/oer/git-invoke/-/commit/05b7bd3d5872ebea1ed2bd18b48c60342bfc3ef6) the commentary section in the Lisp file and the README.

alphapapa:
This package seems to be simply wrappers around `shell-command-to-string` and `call-process`, which doesn't seem to qualify for publishing on MELPA according to the guideline, "Reasonably innovative package."  Anyone who needs to call these git commands in a package can do so with a few lines of code.  The desire to support a 5-year-old version of Emacs on a 3-year-old Debian oldstable release probably does not justify it either.

BTW, you should either use `shell-quote-argument` or avoid `shell-command-to-string` altogether (I recommend the latter).  See e.g. https://github.com/alphapapa/elexandria/blob/a22b12f3472baa617545d2f247ea41f5ef70a488/elexandria.el#L212

lechten:
@alphapapa Thanks for your feedback, `shell-command-to-string` is gone [now](https://gitlab.com/oer/git-invoke/-/commit/35be582509ef31ba60d78eef9a76c05354563cb4). 

It seems that I'm an example that not "anyone" can do this properly. Even if, should we do this repeatedly?

alphapapa:
> It seems that I'm an example that not "anyone" can do this properly. Even if, should we do this repeatedly?

By that logic, every conceivable shell command should have a corresponding `defun` and be in a MELPA package.  Obviously that's not realistic.

Every dependency also comes with a cost.  Simply wrapping a few shell commands is not worth the cost of adding an extra dependency to a package that needs to get the output of `git`.  It's a couple of lines of code.

The best thing you could do, probably, would be to write a blog article reminding people of how to properly get the output of commands as a string and how to handle arguments to them.  The problem of how to get people to read it remains unsolved.  ;)

tarsius:
I agree with @alphapapa.

lechten:
Thanks for the feedback, everyone!

PR #6754, opened 2020-03-15
nndiscourse recipe fix

dickmao:
recipe fix


riscy:
I could have sworn it was this way when I merged. :)  Sorry about that.

PR #6755, opened 2020-03-15
Add xwidget-plus recipe.

canatella:
Enhance xwidget-webkit navigation.
See https://github.com/canatella/xwidget-plus

This package provides enhancements to the xwidget-webkit browser. It provides an easy way to navigate a web page links using the keyboard. It also provides helper functions. I intend to add other features here.

https://github.com/canatella/xwidget-plus

I'm the maintainer.

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/xwwp
+++ b/recipes/xwwp
@@ -1,0 +1,1 @@
+(xwwp :fetcher github :repo "canatella/xwwp" :files ("xwwp.el" "xwwp-follow-link.el" "xwwp-follow-link-ido.el"))
--- a/recipes/xwwp-follow-link-helm
+++ b/recipes/xwwp-follow-link-helm
@@ -1,0 +1,1 @@
+(xwwp-follow-link-helm :fetcher github :repo "canatella/xwwp" :files ("xwwp-follow-link-helm.el"))
--- a/recipes/xwwp-follow-link-ivy
+++ b/recipes/xwwp-follow-link-ivy
@@ -1,0 +1,1 @@
+(xwwp-follow-link-ivy :fetcher github :repo "canatella/xwwp" :files ("xwwp-follow-link-ivy.el"))


conao3:
Hi!

```shell
$ cask; cask build
Package operations: 6 installs, 0 removals
  - Installing [ 1/6] emacs (25.3)... already present
  - Installing [ 2/6] helm (latest)... done
  - Installing [ 3/6] ivy (latest)... done
  - Installing [ 4/6] with-simulated-input (latest)... done
  - Installing [ 5/6] ert-async (latest)... done
  - Installing [ 6/6] ert-runner (latest)...   - Installing [ 6/6] ert-runner (latest)... done
Compiling /home/conao/dev/forks/xwidget-plus/flycheck_xwidget-plus-common.el...

In end of data:
flycheck_xwidget-plus-common.el:160:1:Warning: the function
    ‘xwidget-webkit-execute-script’ is not known to be defined.
Compiling /home/conao/dev/forks/xwidget-plus/xwidget-plus-common.el...

In end of data:
xwidget-plus-common.el:160:1:Warning: the function
    ‘xwidget-webkit-execute-script’ is not known to be defined.
Compiling /home/conao/dev/forks/xwidget-plus/xwidget-plus-follow-link.el...
Compiling /home/conao/dev/forks/xwidget-plus/xwidget-plus.el...

In end of data:
xwidget-plus.el:65:1:Warning: the function ‘xwidget-buffer’ is not known to be
    defined.
```

canatella:
Thanks for your feedback. 

I can't trigger those compile error in the sandbox. Does your build supports xwidget ? Because I do `(require 'xwidget)` which should define those function.

Also maybe I should do something if the build does not support xwidget ? Although I think calling any of these function on an unsupported build will trigger an error message saying just that.

riscy:
Cherry-picking a few package-lints that need your attention:

### xwidget-plus-common.el ###

```
50:0: error: "--js" doesn't start with package's prefix "xwidget-plus"
```

### xwidget-plus-follow-link.el ###

```
273:3: error: You should depend on (emacs "26.1") or the seq package if you need `seq-sort-by'.
141:1: warning: `with-eval-after-load' is for use in configurations, and should rarely be used in packages.
164:1: warning: `with-eval-after-load' is for use in configurations, and should rarely be used in packages.
190:1: warning: `with-eval-after-load' is for use in configurations, and should rarely be used in packages.
```
At a quick glance, it looks like these are indeed lines that should be called in a configuration file, rather than when your package is loaded.

Ultra minor, but you seem to have different license boilerplate in your files -- they're all GPL, but some list the FSF address, others give a URL -- I'd use the URL one (and use https).

And as a last quick note I'd recommend adding `(require cl-lib)` in files where you use `cl-defmethod` just to be safe and not rely on transitive imports.

canatella:
Thank you for the review. Apparently I completely messed up the package-lint run. I fixed that and now I also got all those warnings. Will fix those and get back. Thanks!

canatella:
> ### xwidget-plus-follow-link.el
> 
> ```
> 141:1: warning: `with-eval-after-load' is for use in configurations, and should rarely be used in packages.
> 164:1: warning: `with-eval-after-load' is for use in configurations, and should rarely be used in packages.
> 190:1: warning: `with-eval-after-load' is for use in configurations, and should rarely be used in packages.
> ```
> 
> At a quick glance, it looks like these are indeed lines that should be called in a configuration file, rather than when your package is loaded.

The goal here is to avoid depending on every completion backend. Let's say the user uses ido. I do not want to enforce a dependency on `ivy` and `helm`. I could do a simple `(when (featurep 'ivy))` but then It means that the user as to load the dependencies in the right order. First `(require 'ivy)` then `(require 'xwidget-plus)`. Using `with-eval-after-load` here allows me to ensure both that we do not trigger the unneeded dependency and that the code is always evaluated in the correct order. Is this ok or should I just check `(featurep ...)` and let the user manage the load order ?

mattiasb:
I believe `xwidget` refers to being able to embed [X11](https://en.wikipedia.org/wiki/X_Window_System) widgets in Emacs and your packages seems to work solely on the `xwidget-webkit-*` stuff. Maybe your package should be called `xwidget-webkit-plus` to not clash with a potential future package that adds functionality to the more general xwidget case?

riscy:
@mattiasb raises a good point.

> Using with-eval-after-load here allows me to ensure both that we do not trigger the unneeded dependency and that the code is always evaluated in the correct order. Is this ok or should I just check (featurep ...) and let the user manage the load order ?

I see what's going on now.

Anecdotally, I'm looking at my Emacs session right now and I have loaded `ido` and `ivy` (in a previous Emacs configuration, I was loading helm as well).  These were loaded in some arbitrary order by my config, so I might suffer some unintended function redefinitions in becoming a user.

If what I'm describing seems like it is indeed a problem, maybe we can come up with a solution?  Is simply using completing-read (with the user's `completing-read-function` set appropriately) an option?

canatella:
> Maybe your package should be called `xwidget-webkit-plus` to not clash with a potential future package that adds functionality to the more general xwidget case?

I could change that indeed but then what about using xwwp as the namespace? `xwidget-webkit-plus-then-my-function-name` starts getting really long.

canatella:
> If what I'm describing seems like it is indeed a problem, maybe we can come up with a solution? 

I'm all for it but I would need help on what to do in that case.

> Is simply using completing-read (with the user's `completing-read-function` set appropriately) an option?

I don't think so, the whole point of this feature is to highlight the available completion in the browser as you type them. `completing-read` does not offer any hooks that I know of that would work over all completion backends and that is triggered when the candidate list is updated. If you look at the code you'll see that for each backend there is a different way :(

mattiasb:
> I could change that indeed but then what about using xwwp as the namespace? 
> `xwidget-webkit-plus-then-my-function-name` starts getting really long.

It's only five chars (`plus-`) longer than other `xwidget-webkit-*` functions, but I'll leave this to @riscy :)

canatella:
> > I could change that indeed but then what about using xwwp as the namespace?
> > `xwidget-webkit-plus-then-my-function-name` starts getting really long.
> 
> It's only five chars (`plus-`) longer than other `xwidget-webkit-*` functions, but I'll leave this to @riscy :)

I know, but one has to set a limit, when is long too long ;)

canatella:
I'm going to split up the package so that the the framework to run define and run javascript function in xwidget-webkit is in its own package, the follow-link feature is in its own package too and ivy and helm support are also different packages so that I can remove the `with-eval-after-load`. Just waiting to see which namespace I should use. Given that `xwidget-webkit-plus-follow-link-and-then-my-function-name` is way too long to work with...

riscy:
> so I might suffer some unintended function redefinitions in becoming a user.

It's totally possible I'm missing something, but if I say `(require 'helm) (require 'ivy) (require 'xwidget-plus-follow-link)` then both of these should get executed, I think:
```elisp
  (cl-defmethod xwidget-plus-follow-link-candidates ((_ xwidget-plus-completion-backend-ivy))
    (with-current-buffer (ivy-state-buffer ivy-last)
      (let* ((collection (ivy-state-collection ivy-last))
             (current (ivy-state-current ivy-last))
             (candidates (ivy--filter ivy-text ivy--all-candidates))
             (result (cons current candidates)))
        (seq-map (lambda (c) (cdr (nth (get-text-property 0 'idx c) collection))) result))))

  (cl-defmethod xwidget-plus-follow-link-candidates ((backend xwidget-plus-completion-backend-helm))
    (let* ((candidates (oref backend candidates))
           (selection (helm-get-selection))
           (selected (when selection (cdr (elt (oref backend collection) selection))))
           (result (seq-map #'cdr candidates)))
      (cons selected result)))
```
and I guess they could be run in arbitrary order depending on the order in which I loaded each package.  This might be a moot point since --

> I'm going to split up the package ... so that I can remove the with-eval-after-load.

which sounds like a great idea.

I think `xwwp` is a good prefix for all of the packages, although make sure you've got `xwidget` somewhere in the summary lines to help discoverability in the search.

Thanks so much for being accommodating!

canatella:
> > so I might suffer some unintended function redefinitions in becoming a user.
> 
> It's totally possible I'm missing something, but if I say `(require 'helm) (require 'ivy) (require 'xwidget-plus-follow-link)` then both of these should get executed, I think:
> 
> ```emacs-lisp
>   (cl-defmethod xwidget-plus-follow-link-candidates ((_ xwidget-plus-completion-backend-ivy))
> ...
> 
>   (cl-defmethod xwidget-plus-follow-link-candidates ((backend xwidget-plus-completion-backend-helm))
>     ...
> ```

Ho I see, but it is a `cl-defmethod` here, not a `defun`, it is creating a method on the helm and ivy backend class, not redefining a function here. But as you said, moot point now ;)
> 
> > I'm going to split up the package ... so that I can remove the with-eval-after-load.
> I think `xwwp` is a good prefix for all of the packages, although make sure you've got `xwidget` somewhere in the summary lines to help discoverability in the search.

Let's do that then

> Thanks so much for being accommodating!
Thank you for taking the time to review my code ;)

canatella:
So I splitted the package, now package-lint is complaining about package not being installable, but that seems to be normal: https://github.com/purcell/package-lint/issues

purcell:
You've ended up with more packages here than you need: at the very least, you could (should) put `xwwp-follow-link-ido.el` in `xwwp-follow-link`, because every Emacs has `ido`. And you could even put those two along with `xwwp.el` in a base package called `xwwp`. I'd probably recommend the latter.

canatella:
> you could (should) put `xwwp-follow-link-ido.el` in `xwwp-follow-link`, because every Emacs has `ido`. And you could even put those two along with `xwwp.el` in a base package called `xwwp`

I splat ido so that user could choose to use ido the same way they choose to use helm or ivy: by requiring the matching package. If I forcibly `(require 'ido)` in the `xwwp-follow-link` package it means that I can no longer rely on `(when (featurep 'ido) ...)` to select which backend to use by default. And it would also means that anybody requiring my package would also require ido. But if you think it still not worth it I can merge it.

As for the split of the `xwwp` package, that one defines a framework to define javascript function and matching lisp binding that I hope for me or anybody to reuse to implement such kind of enhancement. The `follow-link` one is implemented on top of that. The next one would probably be the filling of text input using the minibuffer. So this might be premature optimization but I'd like to avoid having to split the package again next time I add an orthogonal feature to the `follow-link` package. What do you think?

purcell:
> I splat ido so that user could choose to use ido the same way they choose to use helm or ivy: by requiring the matching package.

> You can still put it in a separate elisp file, just inside the same package. Perhaps that's the thing I failed to make clear?

> If I forcibly `(require 'ido)` in the `xwwp-follow-link` package it means that I can no longer rely on `(when (featurep 'ido) ...)` to select which backend to use by default.

You can't do that anyway, actually. Presence of a feature has nothing to do with what a user wants. One of the principles of packages is that - like the emacs stdlib - you can have lots of packages installed but not have them all activated. So in my emacs I might find `(featurep 'helm)` and `(featurep 'ivy)` both return `t`, but I only want to use `ivy`.

> And it would also means that anybody requiring my package would also require ido. But if you think it still not worth it I can merge it.

That doesn't apply with a separate .el file in the same package.

purcell:
> As for the split of the `xwwp` package, that one defines a framework to define javascript function and matching lisp binding that I hope for me or anybody to reuse to implement such kind of enhancement. The `follow-link` one is implemented on top of that. The next one would probably be the filling of text input using the minibuffer. So this might be premature optimization but I'd like to avoid having to split the package again next time I add an orthogonal feature to the `follow-link` package, but then indeed I wanted to have everything in the same package before so it does not really matter to me. What do you think then?

I would split into as few packages as possible given the dependencies, so you might have an `xwwp` package containing `xwwp.el`, `xwwp-follow-link.el` and `xwwp-follow-link-ido.el`, and then separate `xwwp-follow-link-helm` packages etc. If, later, you wanted to split `xwwp` out, that would still be possible by introducing a package `xwwp-follow-link` containing `xwwp-follow-link.el` and `xwwp-follow-link-ido.el`, and then having the `helm`/`ivy`-specific packages depend on that instead of `xwwp`.

canatella:
Done!

riscy:
This is looking good to me and aligns with @purcell's comments.  I double-checked and I think the package-requires all line up correctly.  Nobody mentioned it here and it's not a blocker, but in the future it's better to split three such recipes into three PR's so the dependencies can be merged first, which helps with automated testing on my end.

PR #6760, opened 2020-03-16
Add brightscript-mode recipe

viseztrance:

This is a major mode that supports the BrightScript programming language from BrightSign and Roku devices.

https://github.com/viseztrance/brightscript-mode/

Maintainer

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/brightscript-mode
+++ b/recipes/brightscript-mode
@@ -1,0 +1,3 @@
+(brightscript-mode
+ :repo "viseztrance/brightscript-mode"
+ :fetcher github)


conao3:
Hi!

- Please concider using when/unless instead of if which have no else form.

And why do you use advice feature for advice your package function?
https://github.com/viseztrance/brightscript-mode/blob/master/brightscript-mode.el#L59-L62

Addtional, the three files are small enough to combine into a single file. (And so do other little major modes.)

viseztrance:
> And why do you use advice feature for advice your package function?
> https://github.com/viseztrance/brightscript-mode/blob/master/brightscript-mode.el#L59-L62

I started with a generic mode, and this was the way I made it case insensitive. Having this said it's fixed now, and I also moved things to a single file.

In addition to this reran `checkdoc`, `package-lint-current-buffer` and byte compile against my mode el file.

purcell:
Nice work, looks great - thanks. Merging now: welcome to MELPA! :tada:

PS. It's not obvious why you're declaring the dependency on Emacs 26.3: perhaps lower that if you can, so the package is available to users of older Emacsen too.

PR #6764, opened 2020-03-18
add rime recipe

DogLooksGood:

RIME input method in Emacs.

https://github.com/DogLooksGood/emacs-rime

Maintainer

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/rime
+++ b/recipes/rime
@@ -1,3 +1,3 @@
 (rime :repo "DogLooksGood/emacs-rime"
-      :files (:defaults "rime.el" "lib.c" "Makefile")
+      :files (:defaults "lib.c" "Makefile")
       :fetcher github)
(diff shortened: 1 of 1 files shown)


DogLooksGood:
Have tried to build and installed following the insturctions. 
But only el file is installed.

mihaiolteanu:
> But only el file is installed.

Yes, I'm not sure your "lib.c" and "Makefile" files need to be in the recipe.
I can only guess, but those two files seem to be part of a library that the user needs to install separately and are not part of this package.

DogLooksGood:
@mihaiolteanu 
These two files are used to build the dynamic module which is necessary for this package.

DogLooksGood:
The CI failed due to `File already exists: /tmp/rime6412HWe/rime-20200317.1828/rime.el`

riscy:
The `:defaults` part of the recipe already gets `rime.el`.  I'll modify.

DogLooksGood:
I see, thanks

purcell:
Looks good, @DogLooksGood :-)

Merging now, thanks.

PR #6770, opened 2020-03-23
Add company-org-roam package

jethrokuan:

`company-org-roam` is a [company][company] backend for use with [org-roam]. In
Org-roam buffers.

Related to #6738 .

https://github.com/jethrokuan/company-org-roam

Maintainer.

None-needed.

### Checklist

Please confirm with `x`:

[company]: https://company-mode.github.io/
[org-roam]: https://github.com/jethrokuan/org-roam

### DIFF
--- a/recipes/company-org-roam
+++ b/recipes/company-org-roam
@@ -1,0 +1,2 @@
+(company-org-roam :fetcher github
+                   :repo "jethrokuan/company-org-roam")


purcell:
You're using `dash` extensively, it seems, but haven't declared a dependency on it. 

Looks good otherwise!

riscy:
I also notice you're using s (also with no explicit dependency) around #L88 but using `string-prefix-p` may allow dropping the dependency on `s` entirely.  It seems to be brought in transitively so it's not a blocker.

With dash here this looks ready to go.

jethrokuan:
Thanks @riscy , will make the amendment.

purcell:
> [s] seems to be brought in transitively so it's not a blocker.

Always better to declare all dependencies explicitly in case upstream dependencies change to drop the necessary but transitive ones.

PR #6773, opened 2020-03-24
Add recipe for blackout

raxod502:

Blackout is a package which allows you to hide or customize the display of major and minor modes in the mode line. It can replace diminish.el, delight.el, and dim.el.

The reason that Blackout is a better alternative to all three of the listed existing packages [is documented in the README](https://github.com/raxod502/blackout/tree/34adf3d4ea03010861fe6b2b3f28966784a75416#comparison).

https://github.com/raxod502/blackout

I am the maintainer.

**None needed**

### Checklist

Please confirm with `x`:

Note: `package-lint` complains about the `use-package` handlers and the associated `with-eval-after-load`. However, both of these things are necessary for `use-package` integration, so the warnings are erroneous.

### DIFF
--- a/recipes/blackout
+++ b/recipes/blackout
@@ -1,0 +1,1 @@
+(blackout :fetcher github :repo "raxod502/blackout")


conao3:
Hi!

```emacs-lisp
 blackout…    15     warning         There should be two spaces after a period (emacs-lisp-checkdoc)
 blackou…    97     warning         There should be two spaces after a period (emacs-lisp-checkdoc)
 blackou…   109   1 error           "use-package-normalize/:blackout" doesn't start with package's prefix "blackout". (emacs-lisp-package)
 blackou…   109   1 error           `use-package-normalize/:blackout' contains a non-standard separator `/', use hyphens instead (see Elisp Coding Conventions). (emacs-lisp-package)
 blackou…   112     warning         There should be two spaces after a period (emacs-lisp-checkdoc)
 blackou…   136   1 error           "use-package-handler/:blackout" doesn't start with package's prefix "blackout". (emacs-lisp-package)
 blackou…   136   1 error           `use-package-handler/:blackout' contains a non-standard separator `/', use hyphens instead (see Elisp Coding Conventions). (emacs-lisp-package)
 blackou…   149   2 warning         `with-eval-after-load' is for use in configurations, and should rarely be used in packages. (emacs-lisp-package)
```

raxod502:
Thanks for your feedback.

* *There should be two spaces after a period*: I've now explicitly disabled `sentence-end-double-space` in a file-local variable, thanks for the heads-up.
* *"use-package-normalize/:blackout" doesn't start with package's prefix*: There's no avoiding this; it's standard practice for defining `use-package` integration.
* *`with-eval-after-load' is for use in configurations*: There's also no avoiding this, because we must wait for `use-package` to be loaded in order to define the keyword.

conao3:
> "use-package-normalize/:blackout" doesn't start with package's prefix: There's no avoiding this; it's standard practice for defining use-package integration.

I agree. Sorry, that useless report.

conao3:
The `use-package` integration has a `use-package-` prefixing convention. [MELPA search](https://melpa.org/#/?q=use-package-).

Since users who don't use `use-package` may also want to use your package, I think it's a good idea to separate the parts that are relevant to `use-package` from the parts that are not.
That way, the `with-eval-after-load` warning disappears and you can claim that your core package is doing function naming along the `package-lint` lines.

purcell:
Thanks! This looks good overall.

.The `use-package` integration was the only part that stood out to me as requiring a little thought.

Generally the rule is that a package must not auto-enable itself. Autoloading the `use-package-handler/:blackout` function is fine.

The `with-eval-after-load` warnings are there for a number of carefully considered reasons, and normally my default position would be that you should absolutely remove the `with-eval-after-load` block, because the user can _trivially_ add `(add-to-list 'use-package-keywords :blackout)` to their config: they'll have a `use-package` clause for `blackout` anyway, I would think.

Alternatively, you could expose an autoloaded `blackout-use-package-enable` the user can call.

I would _strongly_ suggest this approach, but I'm not going to block the merge on it (this time!), which I'll go ahead with now. :-)

PR #6771, opened 2020-03-23
Add recipe for selectrum

raxod502:
### Brief summary of what the package does

Selectrum is a better solution for incremental narrowing in Emacs, replacing Helm, Ivy, and IDO. Its design philosophy is based on choosing the right abstractions and prioritizing consistency and predictability over special-cased improvements for particular cases. As such, Selectrum follows existing Emacs conventions where they exist and are reasonable, and it declines to implement features which have marginal benefit compared to the additional complexity of a new interface.                                    

### Direct link to the package repository

https://github.com/raxod502/selectrum

### Your association with the package

I am the maintainer.

### Relevant communications with the upstream package maintainer

**None needed**

### Checklist

Please confirm with `x`:

- [x] The package is released under a [GPL-Compatible Free Software License](https://www.gnu.org/licenses/license-list.en.html#GPLCompatibleLicenses).
- [x] I've read [CONTRIBUTING.org](https://github.com/melpa/melpa/blob/master/CONTRIBUTING.org)
- [x] I've used the latest version of [package-lint](https://github.com/purcell/package-lint) to check for packaging issues, and addressed its feedback
- [x] My elisp byte-compiles cleanly
- [x] `M-x checkdoc` is happy with my docstrings
- [x] I've built and installed the package using the instructions in [CONTRIBUTING.org](https://github.com/melpa/melpa/blob/master/CONTRIBUTING.org)
- [ ] I have confirmed some of these without doing them

Note: `package-lint` reports erroneous warnings about autoloading, as explained [in a comment](https://github.com/raxod502/selectrum/blob/210e85697c60befaa03891d282fe593e64fe62b8/selectrum.el#L964-L968).

PR #6774, opened 2020-03-24
Add recipe for ctrlf

raxod502:

CTRLF (pronounced "control F) is an intuitive and efficient solution for single-buffer text search in Emacs, replacing packages such as Isearch, Swiper, and helm-swoop. Taking inspiration from the widely-adopted and battle-tested ctrl+F interfaces in programs such as web browsers, but following the flow and keybindings of Isearch, CTRLF improves on existing text search solutions in convenience, robustness, and consistency.

https://github.com/raxod502/ctrlf

I am the maintainer.

**none needed**

### Checklist

Please confirm with `x`:

As documented in the README, the unusual autoloading conventions improve startup time by allowing the user to enable `ctrlf-mode` without loading CTRLF until a command is actually used. Thus the corresponding `package-lint` warnings are erroneous.

### DIFF
--- a/recipes/ctrlf
+++ b/recipes/ctrlf
@@ -1,0 +1,1 @@
+(ctrlf :fetcher github :repo "raxod502/ctrlf")


conao3:
Hi!

```diff
 ctrlf.el     1   1 warning         The package summary should not end with a period. (emacs-lisp-package)
 ctrlf.el     1   1 warning         The package summary should not end with a period. (emacs-lisp-package)
 ctrlf.el     1   1 warning         The package summary should not end with a period. (emacs-lisp-package)
 ctrlf.el     1   1 warning         The package summary should not end with a period. (emacs-lisp-package)
 ctrlf.el     1   1 warning         The package summary should not end with a period. (emacs-lisp-package)
 ctrlf.el     1   1 warning         Including "Emacs" in the package summary is usually redundant. (emacs-lisp-package)
 ctrlf.el     1   1 warning         The package summary should not end with a period. (emacs-lisp-package)
 ctrlf.el     1   1 warning         Including "Emacs" in the package summary is usually redundant. (emacs-lisp-package)
 ctrlf.el     1   1 warning         The package summary should not end with a period. (emacs-lisp-package)
 ctrlf.el    16     warning         There should be two spaces after a period (emacs-lisp-checkdoc)
 ctrlf.el    78     warning         There should be two spaces after a period (emacs-lisp-checkdoc)
 ctrlf.el    80     warning         There should be two spaces after a period (emacs-lisp-checkdoc)
 ctrlf.el   102     warning         There should be two spaces after a period (emacs-lisp-checkdoc)
 ctrlf.el   104     warning         There should be two spaces after a period (emacs-lisp-checkdoc)
 ctrlf.el   107     warning         There should be two spaces after a period (emacs-lisp-checkdoc)
 ctrlf.el   133     warning         There should be two spaces after a period (emacs-lisp-checkdoc)
 ctrlf.el   135     warning         There should be two spaces after a period (emacs-lisp-checkdoc)
 ctrlf.el   136     warning         There should be two spaces after a period (emacs-lisp-checkdoc)
 ctrlf.el   216  13 error           You should depend on (emacs "26.1") if you need `isearch'. (emacs-lisp-package)
 ctrlf.el   216  13 error           You should depend on (emacs "26.1") if you need `isearch'. (emacs-lisp-package)
 ctrlf.el   216  13 error           You should depend on (emacs "26.1") if you need `isearch'. (emacs-lisp-package)
 ctrlf.el   216  13 error           You should depend on (emacs "26.1") if you need `isearch'. (emacs-lisp-package)
 ctrlf.el   216  13 error           You should depend on (emacs "26.1") if you need `isearch'. (emacs-lisp-package)
 ctrlf.el   216  13 error           You should depend on (emacs "26.1") if you need `isearch'. (emacs-lisp-package)
 ctrlf.el   216  13 error           You should depend on (emacs "26.1") if you need `isearch'. (emacs-lisp-package)
 ctrlf.el   221  13 error           You should depend on (emacs "26.1") if you need `isearch'. (emacs-lisp-package)
 ctrlf.el   221  13 error           You should depend on (emacs "26.1") if you need `isearch'. (emacs-lisp-package)
 ctrlf.el   221  13 error           You should depend on (emacs "26.1") if you need `isearch'. (emacs-lisp-package)
 ctrlf.el   221  13 error           You should depend on (emacs "26.1") if you need `isearch'. (emacs-lisp-package)
 ctrlf.el   221  13 error           You should depend on (emacs "26.1") if you need `isearch'. (emacs-lisp-package)
 ctrlf.el   221  13 error           You should depend on (emacs "26.1") if you need `isearch'. (emacs-lisp-package)
 ctrlf.el   221  13 error           You should depend on (emacs "26.1") if you need `isearch'. (emacs-lisp-package)
 ctrlf.el   286     warning         There should be two spaces after a period (emacs-lisp-checkdoc)
 ctrlf.el   288     warning         There should be two spaces after a period (emacs-lisp-checkdoc)
 ctrlf.el   290     warning         There should be two spaces after a period (emacs-lisp-checkdoc)
 ctrlf.el   326     warning         There should be two spaces after a period (emacs-lisp-checkdoc)
 ctrlf.el   349     warning         There should be two spaces after a period (emacs-lisp-checkdoc)
 ctrlf.el   350     warning         There should be two spaces after a period (emacs-lisp-checkdoc)
 ctrlf.el   359     warning         There should be two spaces after a period (emacs-lisp-checkdoc)
 ctrlf.el   792     warning         There should be two spaces after a period (emacs-lisp-checkdoc)
 ctrlf.el   810     warning         There should be two spaces after a period (emacs-lisp-checkdoc)
 ctrlf.el   878     warning         There should be two spaces after a period (emacs-lisp-checkdoc)
 ctrlf.el   898     warning         There should be two spaces after a period (emacs-lisp-checkdoc)
 ctrlf.el   933     warning         There should be two spaces after a period (emacs-lisp-checkdoc)
 ctrlf.el   974     warning         There should be two spaces after a period (emacs-lisp-checkdoc)
 ctrlf.el   983     warning         There should be two spaces after a period (emacs-lisp-checkdoc)
 ctrlf.…   992     warning         There should be two spaces after a period (emacs-lisp-checkdoc)
 ctrlf.…  1001     warning         There should be two spaces after a period (emacs-lisp-checkdoc)
 ctrlf.…  1009     warning         There should be two spaces after a period (emacs-lisp-checkdoc)
 ctrlf.…  1017     warning         There should be two spaces after a period (emacs-lisp-checkdoc)
 ctrlf.…  1025     warning         There should be two spaces after a period (emacs-lisp-checkdoc)
 ctrlf.…  1033     warning         There should be two spaces after a period (emacs-lisp-checkdoc)
 ctrlf.…  1040   1 warning         Private functions generally should not be autoloaded. (emacs-lisp-package)
 ctrlf.…  1040   1 warning         Private functions generally should not be autoloaded. (emacs-lisp-package)
 ctrlf.…  1040   1 warning         Private functions generally should not be autoloaded. (emacs-lisp-package)
 ctrlf.…  1040   1 warning         Private functions generally should not be autoloaded. (emacs-lisp-package)
 ctrlf.…  1040   1 warning         Private functions generally should not be autoloaded. (emacs-lisp-package)
 ctrlf.…  1040   1 warning         Private functions generally should not be autoloaded. (emacs-lisp-package)
 ctrlf.…  1040   1 warning         Private functions generally should not be autoloaded. (emacs-lisp-package)
 ctrlf.…  1042     warning         There should be two spaces after a period (emacs-lisp-checkdoc)
 ctrlf.…  1047   3 error           Global minor modes should be autoloaded or, rarely, `:require' their defining file (i.e. ":require 'ctrlf"), to support the customization variable of the same name. (emacs-lisp-package)
 ctrlf.…  1047   3 error           Global minor modes should be autoloaded or, rarely, `:require' their defining file (i.e. ":require 'ctrlf"), to support the customization variable of the same name. (emacs-lisp-package)
 ctrlf.…  1047   3 error           Global minor modes should be autoloaded or, rarely, `:require' their defining file (i.e. ":require 'ctrlf"), to support the customization variable of the same name. (emacs-lisp-package)
 ctrlf.…  1047   3 error           Global minor modes should be autoloaded or, rarely, `:require' their defining file (i.e. ":require 'ctrlf"), to support the customization variable of the same name. (emacs-lisp-package)
 ctrlf.…  1047   3 error           Global minor modes should be autoloaded or, rarely, `:require' their defining file (i.e. ":require 'ctrlf"), to support the customization variable of the same name. (emacs-lisp-package)
 ctrlf.…  1047   3 error           Global minor modes should be autoloaded or, rarely, `:require' their defining file (i.e. ":require 'ctrlf"), to support the customization variable of the same name. (emacs-lisp-package)
 ctrlf.…  1047   3 error           Global minor modes should be autoloaded or, rarely, `:require' their defining file (i.e. ":require 'ctrlf"), to support the customization variable of the same name. (emacs-lisp-package)
 ctrlf.…  1064   6 warning         `with-eval-after-load' is for use in configurations, and should rarely be used in packages. (emacs-lisp-package)
 ctrlf.…  1064   6 warning         `with-eval-after-load' is for use in configurations, and should rarely be used in packages. (emacs-lisp-package)
 ctrlf.…  1064   6 warning         `with-eval-after-load' is for use in configurations, and should rarely be used in packages. (emacs-lisp-package)
 ctrlf.…  1064   6 warning         `with-eval-after-load' is for use in configurations, and should rarely be used in packages. (emacs-lisp-package)
 ctrlf.…  1064   6 warning         `with-eval-after-load' is for use in configurations, and should rarely be used in packages. (emacs-lisp-package)
 ctrlf.…  1064   6 warning         `with-eval-after-load' is for use in configurations, and should rarely be used in packages. (emacs-lisp-package)
 ctrlf.…  1064   6 warning         `with-eval-after-load' is for use in configurations, and should rarely be used in packages. (emacs-lisp-package)
```

raxod502:
Thanks for the feedback.

* *Package summary should not end in a period*: It doesn't, must be a bug in the linter.
* *There should be two spaces after a period*: I've explicitly disabled `sentence-end-double-space` in a file-local variable, thanks for the heads-up.
* *You should depend on (emacs "26.1")*: Fixed by removing the explicit `(require 'isearch)`. Now compatible with Emacs 25.
* *Private functions generally should not be autoloaded*: This usage is an exception to the rule, as I noted in the pull request and code.
* *`with-eval-after-load' is for use in configurations*: This usage is again an exception; it improves startup performance significantly.

conao3:
MELPA's [CONTRIBUTING](https://github.com/melpa/melpa/blob/master/CONTRIBUTING.org#making-your-package-ready-for-inclusion) asks you to follow the `package-lint` warning.
If you claim it's a linter bug, please file a PR with `package-lint` and ask for @purcell's consent.

Double-space is for information only. This is now in accordance with the preferences of the package author. Sorry for the noise report.

The use of `with-eval-after-load` is prohibited in the emacs manual.
This, too, requires the consent of the MELPA administrator if you want to use it.

(same comment at #6771, #6773)

purcell:
Thanks! Another interesting package from you. :-)

- The `with-eval-after-load` is totally nonsensical here. It is called in code that will only be executed after the file has been loaded.
- As `package-lint` points out, there's no need to autoload the private `ctrlf--keymap`, and anyway, you should make that keymap public and call it `ctrlf-mode-map`. Also there should be no need at all to redefine it every time you enable the mode. Not the least of the reasons for this is that a user should be able to customise the keymap. The related machinery with the `ctrlf-mode-bindings` custom var and its `:set` function is also extremely unusual. You should be able to do all of this with a proper `defcustom` that contains a regular keymap, even if you have to do a little merging/augmentation of that when applying it in the minibuffer later.
- The `progn` wrapper for the `define-minor-mode` form is also redundant.
- You shouldn't `(setq ctrlf-mode t)`: that var is managed by the code generated by `define-minor-mode`, so there's no need to mess with it. Equally, explicitly toggling the mode off when toggling it on is highly unusual and I am 99% certain it is unnecessary.

Hope that helps!

riscy:
I'll also add that your .el file needs some license boilerplate (just in case someone _only_ receives the file) or an SPDX license identifier.  Looks cool.

raxod502:
@riscy: thanks, I added SPDX license identifiers to all 11 of my packages :)

@purcell: I got rid of the `(setq ctrlf-mode t)`, that was cruft.

However, I'd like to make an argument for why the other three things you commented on ought to stay the same. Actually, they are all part of the same system, which allows for the following feature: you can enable `ctrlf-mode` at startup in your init-file without actually loading CTRLF. This is phenomenal for improving startup time, because everything that is _actually_ needed at startup is autoloaded, and the package manager can cache every package's autoloads in the same file -- thus every package that does this will result in one fewer disk seek during startup. So: the `progn` is used so that the actual definition of the minor mode is available in an autoload; the `with-eval-after-load` has an effect because the minor mode is available even when `ctrlf` is not loaded; and the keymap must be autoloaded because it's used in the minor mode definition.

Now regarding the `ctrlf-mode-bindings`: I understand that it is nonstandard, but I actually really like it. I used to use standard keymap variables like you suggest, but I always hated how opaque and hard to inspect and modify they were. With an alist, the bindings are trivial to inspect and modify programmatically. I think it's a step forward for the same reason that the immutable data structures in Clojure are so fun to work with: the `ctrlf-mode-bindings` use a format that is more plain-data-oriented, and so all the usual ways of manipulating data can be applied.

riscy:
> Actually, they are all part of the same system

I wonder if it would be possible to formalize this system and make it into some kind of a package?  I don't know if that's possible, but the package could make the required code a little bit more comprehensible.  Either way, that doesn't feel like a blocker and actually sounds really interesting.

The mode bindings feel a bit stranger to me, but I experimented with them in practice and they don't appear to conflict much with normal Emacs workflows, except for perhaps `describe-mode` where keybindings don't get listed (correct me if I'm wrong).  Unorthodox but not a blocker.

One automated nag pops up:
- ctrlf.el#L883: You can use `not` or `null`

raxod502:
Funny story, that `eq t` thing used to look smarter:

```elisp
  (let ((inhibit-history (or (not (eq style ctrlf--style))
                             (not (eq t     ctrlf--backward-p)))))
```

I thought that the parallelism helped illustrate that what we were doing was checking if either variable changed from its default. Unfortunately, the indentation got changed later...

> I wonder if it would be possible to formalize this system and make it into some kind of a package?

It would be great, but my intuition is this would not be a good idea. It seems to me like you really need quite a bit of human intervention to figure out the right implementation. And then the issue is I'm not sure how a package can add to another package's autoloads. I think the most promising approach might be a "package" for general startup optimization that provides code for the things that can be automated, and clear documentation+examples for the things that can't be. A standard, if you will.

PR #6776, opened 2020-03-24
Add vlc recipe

xuchunyang:

Control VLC Media Player using its HTTP API, for example, `M-x vlc-next` to play next item in the playlist.

https://github.com/xuchunyang/vlc.el

I'm the maintainer

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/vlc
+++ b/recipes/vlc
@@ -1,0 +1,1 @@
+(vlc :fetcher github :repo "xuchunyang/vlc.el")


alphapapa:
This looks cool.

> To use this package, you're going to need to enable the web interface, go to Preferences → HTTP web interface → check "Enable HTTP web interface", and don't forget to set a non-empty password, then restart VLC.

Is there a way to launch VLC from Emacs with that interface enabled?  For example, in `vlc -H` I see this:

```
      --access {any,satip,vcd,vdr,filesystem,filesystem,cdda,access,upnp,concat,udp,imem_access,dvb,tcp,http,smb,dtv,attachment,dv1394,sftp,ftp,imem,sdp,v4l2,avio,access_mms,none} 
                                 Access module
          This allows you to force an access module. You can use it if the
          correct access is not automatically detected. 
```

conao3:
Hi!

xuchunyang:
> Is there a way to launch VLC from Emacs with that interface enabled? For example, in `vlc -H` I see this:

This shell command does it, that is, it launches VLC plus HTTP interface on http://localhost:8080/ (you can change to another port with `--http-port 9999`)

    $ /Applications/VLC.app/Contents/MacOS/VLC --extraintf=http --http-password=secret
    # For macOS, you can also use open(1)
    $ open -a VLC --args --extraintf=http --http-password=secret

see also https://wiki.archlinux.org/index.php/VLC_media_player#Web_interface

By the way, `vlc -H` says there is `--daemon`, but I found it does not work here (macOS), I was hoping I can run VLC without any UI.

alphapapa:
> I was hoping I can run VLC without any UI.

You can, e.g. I use the `oldrc` interface in a script by running `vlc -I oldrc --rc-unix` and sending/receiving commands over a socket.

My idea was for you to launch VLC from your Emacs package rather than requiring the user to reconfigure VLC and launch it manually.  In some cases, that would be preferable; of course, in others, controlling an existing VLC instance might be preferred.

xuchunyang:
> My idea was for you to launch VLC from your Emacs package rather than requiring the user to reconfigure VLC and launch it manually.

That's a good feature, the one possible issue is that if you launch VLC from Emacs, the parent process is going to be Emacs, so if then when you close Emacs, VLC will be closed as well, though this might be wanted in some cases. On other hand, macOS can work around this issue using the `open` command, but it seems other OSes do not have the equivalent.

xuchunyang:
I added `M-x vlc-start` to address this, this command will launch VLC with web interface enabled.

alphapapa:
> On other hand, macOS can work around this issue using the `open` command, but it seems other OSes do not have the equivalent.

See `nohup`.

> I added M-x vlc-start to address this, this command will launch VLC with web interface enabled.

Nice!

purcell:
Nice work as usual, @xuchunyang! Maybe add some `;;;###autoload` cookies?

Merging now, in any case. :-)

xuchunyang:
Thanks.

> Maybe add some `;;;###autoload` cookies?

Done.

PR #6779, opened 2020-03-25
Add recipe for sxiv.el

contrapunctus-1:

Runs sxiv, the Simple X Image Viewer

https://gitlab.com/contrapunctus/sxiv.el

Author and maintainer

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/sxiv
+++ b/recipes/sxiv
@@ -1,0 +1,3 @@
+(sxiv
+ :fetcher gitlab
+ :repo "contrapunctus/sxiv.el")


conao3:
Hi!

```diff
new file   Cask
@@ -0,0 +1,4 @@
+(source gnu)
+(source melpa)
+
+(package-file "sxiv.el")
```

```
$ cask; cask build
  - Installing [ 1/2] emacs (25.1)... already present
  - Installing [ 2/2] dash (2.16.0)... done
Compiling /home/conao/dev/forks/sxiv.el/sxiv.el...

In sxiv-dired-marked-files-p:
sxiv.el:53:32:Warning: reference to free variable ‘dired-re-mark’

In sxiv-dired-mark-files:
sxiv.el:71:28:Warning: reference to free variable ‘dired-re-dot’

In end of data:
sxiv.el:160:1:Warning: the following functions are not known to be defined:
    dired-insert-subdir, dired-mark-if, dired-get-filename,
    dired-file-name-at-point, dired-get-marked-files
```

contrapunctus-1:
@conao3 thanks! Didn't realize I'd have to depend on Dired too. I've added and pushed a `(require 'dired)`, should I be adding it to the Package-Requires too?

conao3:
I think it isn't needed. Other dired-related packages registered in MELPA don't seem to list `dired` as a dependency either. This is fulfilled by the requirements of Emacs version.

purcell:
Thanks! Quick notes:

- Don't use `eq` to check `major-mode`, e.g. [here](https://gitlab.com/contrapunctus/sxiv.el/-/blob/dae46e6f5890f3d97d45eaadf0194b7ff01f6baf/sxiv.el#L51) - prefer `derived-mode-p`. That whole function could just be:
```el
(defun sxiv-dired-marked-files-p ()
  "Return t if there are marked files in the current Dired buffer.
With no marked files, or if not in a Dired buffer, return nil."
  (and (derived-mode-p 'dired-mode)
       (save-excursion
         (goto-char (point-min))
         (re-search-forward dired-re-mark nil t))))
```
- Looks good otherwise (putting aside my strong dislike of `dash`'s anaphoric macros)

In any case, this can be merged and you can consider the above in your own time. :-)

PR #6782, opened 2020-03-25
Add ascii-table recipe

lassik:

Interactive ASCII table

Show a character map of the ubiquitous 7-bit ASCII character set (128 characters in total).

Do `M-x ascii-table` to bring up a window with the ASCII table.  Press 'b' for binary, 'o' for octal, 'd' for decimal and 'x' for hexadecimal. Press TAB to change the way control characters are shown.

https://github.com/lassik/emacs-ascii-table

Wrote it

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/ascii-table
+++ b/recipes/ascii-table
@@ -1,0 +1,1 @@
+(ascii-table :fetcher github :repo "lassik/emacs-ascii-table")


conao3:
Hi!

```
 ascii.el    17     warning         There should be two spaces after a period (emacs-lisp-checkdoc)
 ascii.el    27     warning         There should be two spaces after a period (emacs-lisp-checkdoc)
```
Note:
`ascii` is a common word and may have a good package name like `ascii-table` or `ascii-interactive-table`.

lassik:
Thanks. Fixed the checkdoc warnings. My copy of checkdoc did not warn about them. Maybe it's sensitive to the value of `sentence-end-double-space`.

lassik:
Please vote on this comment:

* 👍 - change the package name to `ascii-table`
* 👎 - keep the package name `ascii`

lassik:
The current command name is `M-x ascii`. Changing the package name to `ascii-table` would mean that the command has to be `M-x ascii-table` as well. Of course, you could still do `(defalias 'ascii 'ascii-table)` in your personal Emacs config, but it's an extra step.

* If the package name is kept as `ascii`, perhaps other ASCII-related things besides the table can also be collected in the same package.
* On the other hand, it could be more polite towards GNU to let them have all symbols that start with `ascii`.
* On the third hand, Emacs has existed for 35 years and they still haven't coined any such symbols.

purcell:
Definitely suggest `ascii-table`. We all use completing `M-x` or keybdindings anyway, so command name length isn't really an issue, nor the longer internal identifiers. Looks good though!

lassik:
Thanks for the review.

OK, the request to change the name to `ascii-table` is unanimous. Changed.

purcell:
Great, merged now - thanks!

PR #6783, opened 2020-03-25
Added calc-at-point

walseb:
This package allows you to perform calculations on buffer contents

https://github.com/walseb/calc-at-point

Author

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/calc-at-point
+++ b/recipes/calc-at-point
@@ -1,0 +1,1 @@
+(calc-at-point :fetcher github :repo "walseb/calc-at-point")


conao3:
Hi!

```diff
new file   Cask
@@ -0,0 +1,4 @@
+(source gnu)
+(source melpa)
+
+(package-file "calc-at-point.el")
```

```
$ cask; cask build
  - Installing [ 1/2] dash (2.12.0)... done
  - Installing [ 2/2] emacs (26)... already present
Compiling /home/conao/dev/forks/calc-at-point/calc-at-point.el...

In toplevel form:
calc-at-point.el:30:1:Error: Cannot open load file: No such file or directory, dash-functional
```

```
 calc-at-…    65     warning         There should be two spaces after a period (emacs-lisp-checkdoc)
 calc-at…    96     warning         There should be two spaces after a period (emacs-lisp-checkdoc)
 calc-at…   120     warning         There should be two spaces after a period (emacs-lisp-checkdoc)
```

missing `dash-functional` dependency?

walseb:
Thanks it should work now. What did you use to generate the last warnings? `M-x checkdoc` doesn't report any warnings.

purcell:
This looks great. I guess my only point of feedback here is that you probably shouldn't try to provide the authoritative definition of `'number` for `think-at-point`. It certainly wouldn't pick up numbers in certain locales. How about `calc-number`, to make it a bit more specific to the purpose?

walseb:
Good point. The thing is that the normal `thing-at-point` function already has a definition of `'number`, but the function `bounds-of-thing-at-point` doesn't. So I just copied the regex from the already defined `'number` function and put it into `bounds-of-thing-at-point`. But I will think of a more modular solution or maybe just rename the package. And this is unrelated but do you know of a function that works like `string-to-number` but returns numbers with higher precision? Currently it seems to round to `float` which messes up numbers with many decimal places

purcell:
> And this is unrelated but do you know of a function that works like `string-to-number` but returns numbers with higher precision? Currently it seems to round to `float` which messes up numbers with many decimal places

Not sure, sorry, but I'd probably look somewhere like the `calc` source code for clues.

walseb:
I'm now no longer modifying `thing-at-point` and instead allow the user to pass in functions that returns the boundary of the thing at point when run. I still depend on `thing-at-point` because the default implementation of this function uses it to get the boundaries

riscy:
Cool idea!  Looks good to go.  Non-blocking, but I find sharp-quoting can help me detect problems earlier by causing the compiler to warn, e.g. --

- calc-at-point.el#L67: It's safer to sharp-quote function names; use `#'`
- calc-at-point.el#L222: It's safer to sharp-quote function names; use `#'`
- calc-at-point.el#L214: It's safer to sharp-quote function names; use `#'`

walseb:
Thanks! It should be fixed now, although I didn't sharp quote any operators partially because it made it look more complex and because it would be very strange for them to be missing

PR #6786, opened 2020-03-26
Add basic-ide

Sasanidas:

This package provide IDE like feature for editing c64 basic files, it also interacts with VICE emulator

https://gitlab.com/sasanidas/emacs-c64-basic-ide

Maintainer

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/basic-ide
+++ b/recipes/basic-ide
@@ -1,0 +1,1 @@
+(basic-ide :fetcher gitlab :repo "sasanidas/emacs-c64-basic-ide")


purcell:
Fun! Thanks for submitting this. A few notes:

- Would this work for any `basic` interpreter? If not, a more specific name would be preferable.
- It's dangerous to use `concat` to add file paths into command strings: at least use `shell-quote-argument` first. Note also that not every BASIC buffer will have a `buffer-file-name` -- buffers in Emacs don't have to be backed by a file, so you might like to handle this case explicitly.
- Don't hardcode the `/tmp` directory - there are emacs functions for producing temporary file and directory names on various platforms.
- The file doesn't seem to be syntactically valid: where's the closing paren [here](https://gitlab.com/sasanidas/emacs-c64-basic-ide/-/blob/11a9c711817acd136ffc40d6cb615d0216cd7764/basic-ide.el#L51)? I note that you haven't confirmed that you can byte-compile this code cleanly, so perhaps this is why.

Hope that helps!

Sasanidas:
> Fun! Thanks for submitting this. A few notes:
> 
>     * Would this work for any `basic` interpreter? If not, a more specific name would be preferable.
> 
>     * It's dangerous to use `concat` to add file paths into command strings: at least use `shell-quote-argument` first. Note also that not every BASIC buffer will have a `buffer-file-name` -- buffers in Emacs don't have to be backed by a file, so you might like to handle this case explicitly.
> 
>     * Don't hardcode the `/tmp` directory - there are emacs functions for producing temporary file and directory names on various platforms.
> 
>     * The file doesn't seem to be syntactically valid: where's the closing paren [here](https://gitlab.com/sasanidas/emacs-c64-basic-ide/-/blob/11a9c711817acd136ffc40d6cb615d0216cd7764/basic-ide.el#L51)? I note that you haven't confirmed that you can byte-compile this code cleanly, so perhaps this is why.
> 
> 
> Hope that helps!


Thank you for your response!

- For now it only supports basic for the c64, but in theory you can change the x64 executable for another and just send the buffer content. But still, it is intended for c64 development
-  Thank you for your tips , I'm not an expert in lisp, I'll try to fix those problems soon.
-  Yes, the compiler is throwing me some errors, and maybe that can take me a little time to fix.

One more thing is that some features have some know bugs, I'm going to put those bugs in the README.md file, some are easy to fix but other may take some time.

riscy:
Some quick notes from me on the byte-compilation output:

```
basic-ide.el:120:38:Warning: defcustom for `basic-ide-completion-backend'
    fails to specify type
basic-ide.el:120:42:Warning: reference to free variable
    `basic-ide-completion-backend'
basic-ide.el:143:57:Warning: assignment to free variable `file--name'
basic-ide.el:141:8:Warning: `end-of-buffer' is for interactive use only; use
    `(goto-char (point-max))' instead.
basic-ide.el:153:59:Warning: reference to free variable
    `basic-ide-cbmbasic-executable'
basic-ide.el:153:59:Warning: reference to free variable `file--name'
basic-ide.el:144:8:Warning: `end-of-buffer' is for interactive use only; use
    `(goto-char (point-max))' instead.
basic-ide.el:179:6:Warning: `(setq)' is a malformed function
basic-ide.el:168:10:Warning: assignment to free variable `local-buffer-name'
basic-ide.el:162:111:Warning: assignment to free variable `pos1'
basic-ide.el:161:104:Warning: assignment to free variable `pos2'
basic-ide.el:160:85:Warning: reference to free variable `etq'
basic-ide.el:160:89:Warning: reference to free variable `bds'
basic-ide.el:160:120:Error: missing value for `(cdr bds)' at end of setq
basic-ide.el:162:87:Warning: reference to free variable `pos1'
basic-ide.el:162:116:Warning: reference to free variable `pos2'
basic-ide.el:170:15:Warning: assignment to free variable `command-output'
basic-ide.el:167:10:Warning: reference to free variable `local-buffer-name'
basic-ide.el:169:8:Warning: reference to free variable `command-output'
basic-ide.el:180:6:Warning: reference to free variable
    `basic-ide-x64-executable'
basic-ide.el:180:6:Warning: reference to free variable `basic-ide-x64-kernal'
basic-ide.el:195:6:Warning: reference to free variable
    `basic-ide-vice-simon-disk'
In end of data:
basic-ide.el:201:1:Warning: the following functions are not known to be
    defined: basic-ide-local-execute, eshell-kill-input, eshell-send-input
```
When you see "reference/assignment to free variable", you can solve those ones easily with a `defvar` at the top of your file.

In other cases, when you're using `setq` to define the variable inside the scope of a function, you can use `let`, e.g.:
```elisp
(setq local-buffer-name (if output-buffer-name (format output-buffer-name) "cbmbasic-output"))
(do some stuff with local-buffer-name)
```
becomes:
(let ((local-buffer-name (if output-buffer-name (format output-buffer-name "cbmbasic-output"))))
  (do some stuff with local-buffer-name))
```
Some other errors up there point to, I think, legitimate errors, so it's worth going through those.

You've also put all of your function definitions under `define-minor-mode`, but all of these should be at the top-level.  Check out other minor modes (for example <kbd>C-h f hl-line-mode</kbd>) for a pattern to follow.

And some other suggestions:
- basic-ide.el#L163: Use `temporary-file-directory` instead of /tmp in code
- basic-ide.el#L164: Use `temporary-file-directory` instead of /tmp in code
- basic-ide.el#L185: Use `temporary-file-directory` instead of /tmp in code
- basic-ide.el#L187: Use `temporary-file-directory` instead of /tmp in code
- basic-ide.el#L99: Prefer `https` over `http` (if possible)
- basic-ide.el#L178: Prefer `https` over `http` (if possible)

**Comment by Sasanidas** [comment]
> Some quick notes from me on the byte-compilation output:
> 
> byte-compile-file (using Emacs 26.3):
> 
> ```
> basic-ide.el:120:38:Warning: defcustom for `basic-ide-completion-backend'
>     fails to specify type
> basic-ide.el:120:42:Warning: reference to free variable
>     `basic-ide-completion-backend'
> basic-ide.el:143:57:Warning: assignment to free variable `file--name'
> basic-ide.el:141:8:Warning: `end-of-buffer' is for interactive use only; use
>     `(goto-char (point-max))' instead.
> basic-ide.el:153:59:Warning: reference to free variable
>     `basic-ide-cbmbasic-executable'
> basic-ide.el:153:59:Warning: reference to free variable `file--name'
> basic-ide.el:144:8:Warning: `end-of-buffer' is for interactive use only; use
>     `(goto-char (point-max))' instead.
> basic-ide.el:179:6:Warning: `(setq)' is a malformed function
> basic-ide.el:168:10:Warning: assignment to free variable `local-buffer-name'
> basic-ide.el:162:111:Warning: assignment to free variable `pos1'
> basic-ide.el:161:104:Warning: assignment to free variable `pos2'
> basic-ide.el:160:85:Warning: reference to free variable `etq'
> basic-ide.el:160:89:Warning: reference to free variable `bds'
> basic-ide.el:160:120:Error: missing value for `(cdr bds)' at end of setq
> basic-ide.el:162:87:Warning: reference to free variable `pos1'
> basic-ide.el:162:116:Warning: reference to free variable `pos2'
> basic-ide.el:170:15:Warning: assignment to free variable `command-output'
> basic-ide.el:167:10:Warning: reference to free variable `local-buffer-name'
> basic-ide.el:169:8:Warning: reference to free variable `command-output'
> basic-ide.el:180:6:Warning: reference to free variable
>     `basic-ide-x64-executable'
> basic-ide.el:180:6:Warning: reference to free variable `basic-ide-x64-kernal'
> basic-ide.el:195:6:Warning: reference to free variable
>     `basic-ide-vice-simon-disk'
> In end of data:
> basic-ide.el:201:1:Warning: the following functions are not known to be
>     defined: basic-ide-local-execute, eshell-kill-input, eshell-send-input
> ```
> 
> When you see "reference/assignment to free variable", you can solve those ones easily with a `defvar` at the top of your file.
> 
> In other cases, when you're using `setq` to define the variable inside the scope of a function, you can use `let`, e.g.:
> 
> ```emacs-lisp
> (setq local-buffer-name (if output-buffer-name (format output-buffer-name) "cbmbasic-output"))
> (do some stuff with local-buffer-name)
> ```
> 
> becomes:
> 
> ```emacs-lisp
> (let ((local-buffer-name (if output-buffer-name (format output-buffer-name "cbmbasic-output"))))
>   (do some stuff with local-buffer-name))
> ```
> 
> Some other errors up there point to, I think, legitimate errors, so it's worth going through those.
> 
> You've also put all of your function definitions under `define-minor-mode`, but all of these should be at the top-level. Check out other minor modes (for example C-h f hl-line-mode) for a pattern to follow.
> 
> And some other suggestions:
> 
>     * basic-ide.el#L163: Use `temporary-file-directory` instead of /tmp in code
> 
>     * basic-ide.el#L164: Use `temporary-file-directory` instead of /tmp in code
> 
>     * basic-ide.el#L185: Use `temporary-file-directory` instead of /tmp in code
> 
>     * basic-ide.el#L187: Use `temporary-file-directory` instead of /tmp in code
> 
>     * basic-ide.el#L99: Prefer `https` over `http` (if possible)
> 
>     * basic-ide.el#L178: Prefer `https` over `http` (if possible)

Hello!
Thank you very much for those tips, I tried to implement all of them and clean the compiler buffer.
For now, there is no more free variable assignment and as you guys recommended I used make-temp-file instead of the plain concat.
But there is only one more error i cannot figure out how to solve 
```

Compiling` file /mnt/Documentos/Documentos/Programming/emfibasic/basic-ide.el at Tue Mar 31 22:07:33 2020

In end of data:
basic-ide.el:216:1:Warning: the following functions are not known to be defined:
    eshell-kill-input, eshell-send-input
 
```
Thank you

**Comment by riscy** [comment]
Those ones should be easy -- just add a `(require 'eshell)` and you should be set.  After that we can start taking a closer look and see if there are any more suggestions.

You've also got one very minor package-lint:

package-lint-current-buffer (using version 20200313.2338):
```
1 issue found:
83:7: warning: Closing parens should not be wrapped onto new lines.
```

**Comment by Sasanidas** [comment]
> Those ones should be easy -- just add a `(require 'eshell)` and you should be set. After that we can start taking a closer look and see if there are any more suggestions.
> 
> You've also got one very minor package-lint:
> 
> package-lint-current-buffer (using version 20200313.2338):
> 
> ```
> 1 issue found:
> 83:7: warning: Closing parens should not be wrapped onto new lines.
> ```

Ok, I think I addressed most of the major problems, but of course I'm open to suggestions.
Thank you!

**Comment by riscy** [comment]
I thought I'd double-check on this one:

>> Would this work for any basic interpreter? If not, a more specific name would be preferable.
> it is intended for c64 development

Would basic-ide-c64 be a reasonable name for it then?  If you decide to expand it to other BASICs in the future, you could create a basic-ide package that basic-ide-c64 extended from as a derived-mode?

Beyond that, I'm looking at this next pass of your code, and it looks much improved and compiles.  Some other items I'm seeing now on a more careful pass:

- You should flatten your code a bit more -- you shouldn't run `flycheck-define-checker` inside `define-minor-mode` but rather before `define-minor-mode`, and similarly you shouldn't `defun basic-ide-company-backend` inside `define-minor-mode` -- ideally they lay flat.
- A lot of your code is indented (as if it were nested under the `define-minor-mode`) but doesn't need to be.  You can easily use <kbd>M-x indent-region</kbd> in your elisp buffer to indent everything perfectly.  I'd also aim to try to put one blank line between function definitions (basic-ide-vice-load-simon-basic and basic-ide-vice-reset)
- Along the same lines, try to ensure your docstrings and comments are wrapped more or less consitently around 80 characters.  There's a meaningful benefit to doing this: the automatically-generated help across various Emacs facilities (e.g. <kbd>M-x apropos</kbd>) will display correctly if you adhere to this.  You can use <kbd>M-x fill-paragraph</kbd> to achieve this on the lines in question.
- Quite a few variables seem to be autoloaded but on first glance they don't need to be.  See [when to autoload](https://www.gnu.org/software/emacs/manual/html_node/elisp/Autoload.html).
- As a small thing I notice, is it worth having `"as"` in your `basic-ide-completion-backend`?  It seems like autocompleting that after `a` would be counterproductive for a user (similar for if, to, fn, on, ...)

**Comment by Sasanidas** [comment]
> I thought I'd double-check on this one:
> 
> > > Would this work for any basic interpreter? If not, a more specific name would be preferable.
> 
> > it is intended for c64 development
> 
> Would basic-ide-c64 be a reasonable name for it then? If you decide to expand it to other BASICs in the future, you could create a basic-ide package that basic-ide-c64 extended from as a derived-mode?
> 
> Beyond that, I'm looking at this next pass of your code, and it looks much improved and compiles. Some other items I'm seeing now on a more careful pass:
> 
>     * You should flatten your code a bit more -- you shouldn't run `flycheck-define-checker` inside `define-minor-mode` but rather before `define-minor-mode`, and similarly you shouldn't `defun basic-ide-company-backend` inside `define-minor-mode` -- ideally they lay flat.
> 
>     * A lot of your code is indented (as if it were nested under the `define-minor-mode`) but doesn't need to be.  You can easily use M-x indent-region in your elisp buffer to indent everything perfectly.  I'd also aim to try to put one blank line between function definitions (basic-ide-vice-load-simon-basic and basic-ide-vice-reset)
> 
>     * Along the same lines, try to ensure your docstrings and comments are wrapped more or less consitently around 80 characters.  There's a meaningful benefit to doing this: the automatically-generated help across various Emacs facilities (e.g. M-x apropos) will display correctly if you adhere to this.  You can use M-x fill-paragraph to achieve this on the lines in question.
> 
>     * Quite a few variables seem to be autoloaded but on first glance they don't need to be.  See [when to autoload](https://www.gnu.org/software/emacs/manual/html_node/elisp/Autoload.html).
> 
>     * As a small thing I notice, is it worth having `"as"` in your `basic-ide-completion-backend`?  It seems like autocompleting that after `a` would be counterproductive for a user (similar for if, to, fn, on, ...)

Hello and thanks for all the tips, I was changing some important features in the development branch so, let's update some things.

So right now it can support others basic interpreters such as the basic 7.0 and Simon's basic, and in the no so distant future other vice implementation such as the commodore 128 and 16.

Let's go with the issues:

1 - I move most of the code from inside of the minor mode to above as suggested but still keep the add-to-list company backend function.

2 - As suggested, the code was indented with indent-region.

3 - As suggested, the document string was shorted with fill-paragraph.

4 - As suggested, there is no need for autoload custom statements.

5 - I change that in the new update, now it is up to the use to decide the completion backend, the idea is to support all the petcat possibilities.


Open to more tips, thank you!

**Comment by riscy** [comment]
Having another look, thanks!

One thing that I've just noticed is that you `package-requires` on helm which is quite large, but then don't appear to use it.  Unless I'm missing something, you should remove that -- otherwise people who don't have helm will be installing a large dependency.

```
basic-ide.el:206:42:Warning: assignment to free variable
    `basic-ide-completion-backend'
```
That comes from `basic-ide-company-backend` where you setq the variable but you could also just:
```elisp
(defun basic-ide-company-backend (command &optional arg &rest ignored)
  "Provide the necessary backend for company mode.
The completion requires COMMAND &optional ARG &rest IGNORED."
  (let ((basic-ide-completion-backend (basic-ide--commands)))
    (cl-case command
      (interactive (company-begin-backend 'basic-ide-completion-backend))
      (prefix (and (eq major-mode 'basic-mode)
                   (company-grab-symbol)))
      (candidates
       (-when-let*
           ((basic-ide-backend (bound-and-true-p basic-ide-completion-backend)))
         (cl-remove-if-not
          (lambda (c) (string-prefix-p arg c))
          basic-ide-backend))))))
```
Noting I changed the `if-let` to a `when-let` so you don't have to use`progn`.

I also notice you have:
```elisp
(defcustom basic-ide-cbmbasic-executable "/home/fermin/Programming/emfibasic/cbmbasic/cbmbasic" ...
```
which seems fairly particular to you - maybe it would be better to set this to simply `cbmbasic` and let the user customize any path information if it's not on their PATH?

Lastly, I notice you don't package-requires on `s`, but you do use it, so one of your dependencies must be installing it.  This might be unsafe to rely on.  Better to use Emacs's build in `split-string` than to use the similar `s-split` in this case.

riscy:
And actually, can't we assume `(bound-and-true-p basic-ide-completion-backend)` will always be true in that context?

Sasanidas:
> Having another look, thanks!
> 
> One thing that I've just noticed is that you `package-requires` on helm which is quite large, but then don't appear to use it. Unless I'm missing something, you should remove that -- otherwise people who don't have helm will be installing a large dependency.
> 
> byte-compile-file (using Emacs 26.3):
> 
> ```
> basic-ide.el:206:42:Warning: assignment to free variable
>     `basic-ide-completion-backend'
> ```
> 
> That comes from `basic-ide-company-backend` where you setq the variable but you could also just:
> 
> ```emacs-lisp
> (defun basic-ide-company-backend (command &optional arg &rest ignored)
>   "Provide the necessary backend for company mode.
> The completion requires COMMAND &optional ARG &rest IGNORED."
>   (let ((basic-ide-completion-backend (basic-ide--commands)))
>     (cl-case command
>       (interactive (company-begin-backend 'basic-ide-completion-backend))
>       (prefix (and (eq major-mode 'basic-mode)
>                    (company-grab-symbol)))
>       (candidates
>        (-when-let*
>            ((basic-ide-backend (bound-and-true-p basic-ide-completion-backend)))
>          (cl-remove-if-not
>           (lambda (c) (string-prefix-p arg c))
>           basic-ide-backend))))))
> ```
> 
> Noting I changed the `if-let` to a `when-let` so you don't have to use`progn`.
> 
> I also notice you have:
> 
> ```emacs-lisp
> (defcustom basic-ide-cbmbasic-executable "/home/fermin/Programming/emfibasic/cbmbasic/cbmbasic" ...
> ```
> 
> which seems fairly particular to you - maybe it would be better to set this to simply `cbmbasic` and let the user customize any path information if it's not on their PATH?
> 
> Lastly, I notice you don't package-requires on `s`, but you do use it, so one of your dependencies must be installing it. This might be unsafe to rely on. Better to use Emacs's build in `split-string` than to use the similar `s-split` in this case.

Hello!

So I change the company backend function with a variable, so it doesn't load every time (it doesn't make too much sense to load the same thing over and over)

Change the personal path, that was a mistake, sorry for that.

And deleted unused libraries like helm and s, it compiles clean now.

Open for more tips, thank you!

riscy:
That looks like everything, and the checklist looks clean.

One thing I would recommend is having a look over your documentation and removing reference to "helm support" (since it seems like that is gone, at least for now).  A couple typos I'm also seeing while scanning the docs:

- Executa -> Execute
- wih -> with
- You write "Drupal configuration." in `defgroup basic-ide` but I think that docstring needs to be updated relative to the mode.

riscy:
Oh and looks like this is your first package on MELPA.  Welcome. :)

PR #6788, opened 2020-03-28
Add cpu-sos

oitofelix:
----

‘cpu-sos’ is a buffer-local minor-mode designed to track the
visibility of buffers associated with sub-processes or EXWM managed
windows and send a SIGSTOP signal to those processes (and their
related ones) as soon as their buffers become buried, eventually
reverting this by sending a SIGCONT signal as soon as they become
visible again.  This has the effect of limiting CPU consumption of
processes managed by Emacs at the user’s discretion.  Useful for
programs whose background processing the user is not interested in.
For example, web-browsers running JavaScript aggressively on
background for no good reason.  Other legitimate use is to forcibly
disable background app notifications while one’s attention focus is
elsewhere.

CAVEATS: Notice that the concept of "visibility" used by this package
is defined by the semantics of the value ‘visible’ given to the
parameter ‘ALL-FRAMES’ of function ‘get-buffer-window’.  This is
necessary, but not sufficient for actual view of the buffer at hand.
For instance, if a buffer is in a window of a frame that is totally
occluded by another it still is regarded as "visible", although one
can’t actually see it.  Aside from imprecise detection of visual
interaction, there is no attempt to detect sound interaction.
Therefore, buffers running music players or recording programs should
not have this mode enabled.  The same is true if one wants to have
asynchronous processes delivering notifications at arrival.  Keep also
in mind that trying to yank selection from stopped processes is
problematic.

https://github.com/oitofelix/cpu-sos

I’m the author.

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/cpu-sos
+++ b/recipes/cpu-sos
@@ -1,0 +1,1 @@
+(cpu-sos :fetcher github :repo "oitofelix/cpu-sos")


purcell:
Thanks! Quick notes:

- `package-lint` should have cautioned you that `cl-macs` is not available in Emacs 24.1, and that anyway you shouldn't `require` that feature directly, but instead require `cl-lib`. There are also several other issues I can see which `package-lint` would raise. Did you really run `package-lint`, as you claimed in the checklist above?
- Please don't use emoji in mode lighters by default. Instead, use a variable for easy customisation, and make the default lighter string something more boring and ASCII, like `" CPUSOS"`.

oitofelix:
Hi Steve,

>   • package-lint should have cautioned you that cl-macs is not
>   available in Emacs 24.1, and that anyway you shouldn't require that
>   feature directly, but instead require cl-lib. There are also several
>   other issues I can see which package-lint would raise. Did you
>   really run package-lint, as you claimed in the checklist above?
>   • Please don't use emoji in mode lighters by default. Instead, use a
>   variable for easy customisation, and make the default lighter string
>   something more boring and ASCII, like " CPUSOS".

Done!

--

riscy:
Thanks for checking that.  :)  There are some byte-compile warnings:
```
cpu-sos.el:129:33:Warning: Unknown slot `value'
cpu-sos.el:131:62:Warning: reference to free variable `exwm--connection'
In end of data:
cpu-sos.el:235:1:Warning: the following functions are not known to be defined:
    xcb:+request-unchecked+reply, exwm--buffer->id
```
but I feel like these are non-blocking because they only show up when exwm isn't installed.  However, since exwm is only an optional dependency, I think you should fix them so that all your intended users can experience a warning-free byte-compile.  You can add something like this:

```elisp
(defvar exwm--connection)
(declare-function xcb:+request-unchecked+reply "ext:exwm.el" t t)
(declare-function exwm--buffer->id "ext:exwm.el" t t)
```
and you can stop the compiler from compiling the call to `slot-value` down and creating a warning by replacing it with:
```elisp
                   (funcall 'slot-value
                            (xcb:+request-unchecked+reply
                             exwm--connection
                             (make-instance 'xcb:ewmh:get-_NET_WM_PID
                                            :window
                                            (exwm--buffer->id buffer)))
                            'value))))))
```
(noting that I'm explicitly *not* sharp-quoting 'slow-value) but you may want to double-check my work on that.

PR #6789, opened 2020-03-28
Add rego mode

psibi:

Rego mode is a major mode for Rego language. More details about rego here: https://www.openpolicyagent.org/docs/latest/policy-language/#what-is-rego

https://github.com/psibi/rego-mode

Author 

** None Needed **

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/rego-mode
+++ b/recipes/rego-mode
@@ -1,0 +1,3 @@
+(rego-mode
+ :repo "psibi/rego-mode"
+ :fetcher github)


riscy:
Hi, thanks for this. :)  A quick run over the checklist:

```
rego-mode.el:122:9:Warning: reference to free variable `rego-format-at-save'
In rego-format-region:
rego-mode.el:141:9:Warning: reference to free variable `rego-format-arguments'
```
I think these are easily solved with some `defvar`s at the top of the file.

- No issues!

package-lint-current-buffer (using version 20200313.2338):
- No issues!

- rego-mode.el#L182: It's safer to sharp-quote function names; use `#'`
- rego-mode.el#L25: Prefer `https` over `http` (if possible)


### License ###

- Add a LICENSE file that GitHub can detect (e.g. no markup).

psibi:
@riscy Thanks for the comments. I'm using the same emacs version:
```
GNU Emacs 26.3 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.10)
```
But I'm unable to reproduce the `byte-compile-file` issues.  Apart from that I have fixed other suggestions by you:
* using `#'`
* Prefer `https`
* Add LICENSE file

psibi:
@riscy What should be the next steps done to push this along ?

riscy:
There are other PR's, but yours will get attention again soon. :)

riscy:
Probably you're not seeing the warnings because you're byte-compiling in an Emacs instance that has already evaluated the buffer, and so those variables exist in that session before they're called on.

I took a closer look and you can fix by reordering your declarations so that the "free variables" come before they're used:
```elisp
(defcustom rego-format-at-save t
  "If non-nil, the Rego buffers will be formatted after each save."
  :type 'boolean
  :group 'rego
  :safe 'booleanp)

;; The main mode functions
;;;###autoload
(define-derived-mode rego-mode prog-mode
  "Rego"
  "Major mode for editing Rego files."
  :group 'rego
  :keymap rego-mode-map
  :syntax-table rego-mode-syntax-table
  (setq font-lock-defaults '(rego-mode-font-lock-keywords))
  (setq-local indent-tabs-mode t)
  (setq-local tab-width 4)
  (setq-local comment-start "# ")
  (setq-local comment-end "")
  (when rego-format-at-save
    (rego-format-on-save-mode)))

(defcustom rego-opa-command "opa"
  "Command used to normalize Rego files.
Should be opa or the complete path to your opa executable,
  e.g.: /home/sibi/.local/bin/opa"
  :type 'file
  :group 'rego
  :safe 'stringp)

(defcustom rego-format-arguments '("fmt")
  "Provide a list of arguments for the formatter e.g. '(\"--ascii\")."
  :type 'list
  :group 'rego
  :safe 'listp)

(reformatter-define rego-format
  :program rego-opa-command
  :args rego-format-arguments
  :group 'rego
  :lighter " OpaFmt")
```

Taking a closer look, I see in your major mode definition that you `(setq-local indent-tabs-mode t)` but this seems like it should be up to users, not up to your mode.  Similarly with `(setq-local tab-width 4)`, which also seems like a personal preference that users should retain control over.

And one last thing, but minor -- should the comment "Same as rego-repl but asks DATA-FILE argument interactively which will be passed to the repl." be:
```elisp
"Same as `rego-repl-show' but asks DATA-FILE argument interactively which will be passed to the repl."
```

Otherwise the code looks nice and tidy.  And thank you for the link to what Rego is in the PR description -- very helpful.

psibi:
@riscy Thanks for the feedback. I have pushed a fix which addresses all of them.

> Probably you're not seeing the warnings because you're byte-compiling in an Emacs instance that has already evaluated the buffer, and so those variables exist in that session before they're called on.

I tested with the new Emacs instance and also by removing the existing `elc` file, but it doesn't seem to work. I guess I'm missing something else.

riscy:
> I tested with the new Emacs instance and also by removing the existing elc file, but it doesn't seem to work. I guess I'm missing something else.

No worries, everything seems to check out now.  Running flycheck on elisp code is often really good at picking these kinds of things up too.  I run all my checks on here with this tool https://github.com/riscy/melpazoid/ which does approximate cleanroom byte compilation too.

psibi:
>  I run all my checks on here with this tool https://github.com/riscy/melpazoid/ which does approximate cleanroom byte compilation too.

That's nice. I will see if I can integrate that with my CI. Thanks for merging!

PR #6791, opened 2020-03-29
Add elisp-depmap recipe

mtekman:

A library to generate a dependency map for elisp projects. Generates a GraphViz-compatible DOT file representation of all the top-level definitions in an elisp project, and represents the connections between them.

https://github.com/mtekman/elisp-depmap.el

Maintainer

**None needed**

### Checklist

### DIFF
--- a/recipes/elisp-depmap
+++ b/recipes/elisp-depmap
@@ -1,0 +1,1 @@
+(elisp-depmap :fetcher github :repo "mtekman/elisp-depmap.el" )


purcell:
Thanks! A few quick notes here:

- You need to be using `shell-quote-argument` when you're assembling command lines in `package-map-exec.el`, otherwise filenames with spaces or other special shell characters will certainly break the code.
- `package-map` is a confusing name for this, because "packages" are installable units that contain one or more files, as in MELPA. Perhaps `elisp-dep-map` or similar?
- From `package-lint`, I have a decent amount of experience with code for finding elisp symbol definitions. I wouldn't advise doing this with regexps, and I can see that your existing patterns miss some common cases like the `cl-` variants of `defun` and `defmacro` etc. Using existing facilities like `imenu` can help here.
- Reliably finding occurrences of where symbols are used is very difficult in practice. It's possible that the elisp `xref` backend could help, at least for code that has already been loaded.
- `autoload`ed symbols will presumably break this code, because you wouldn't need to `require` a package in order to access the symbol.
- `projectile` is probably an unnecessary dependency, and it constrains the usage model excessively IMO: I'd expect to just find a function which is given a number of elisp file paths (or a `load-path`-style list of directories). `projectile` would just be a way of getting such a list of paths, but it needn't be used directly by the package. (Also, you mention different versions of `projectile` in different `.el` files.)

Hope that helps! I'm pleased you're working on something like this, because it would be nice to suggest it as a tool for users who have split up their package haphazardly and ended up with a broken symbol dependency graph.

mtekman:
Hey thanks -- I'll rethink the name, drop `projectile` functions, and I'll play around with `xref` or `imenu` (or I'll just peek at `package-lint`).

> autoloaded symbols will presumably break this code, because you wouldn't need to require a package in order to access the symbol.

I didn't follow this part so well... I think you're saying that the package looks at the `require` symbol and follows the named library back to the source file, and parses that in a recursive fashiong, right? I don't do that. My package literally only scans the files in a directory tracked by version control (although I used projectile) and parses them one by one. So autoloaded symbols will still be correctly parsed even their source library is not `require`'d, if I have understood correctly

purcell:
>  So autoloaded symbols will still be correctly parsed even their source library is never `require`'d elsewhere, if I have understood correctly

That makes sense, and yes, I misunderstood the approach. :-)

mtekman:
Hey, so I explored `xref` and `imenu`, and `imenu` seems the better choice because `xref` appears the `xref-find-backend` appears to rely on `grep`, which I have tried to use previously in this package with some async finish-hook issues, as well as `projectile-ripgrep` which had a [similar issue](https://github.com/nlamirault/ripgrep.el/issues/33). 

That being said, as I started to play around with `imenu`, it became clear to me that my regex approach, naive though it may be, was actually not that much worse than the method that `imenu` uses to build it's index.

Specifically, `(imenu--make-index-alist)` generates the alist-equivalent of my hashtable, but the method to get this is not better than my own:

```elisp
(defun imenu-default-create-index-function ()
   ;;   The most general method is to move point to end of buffer, then repeatedly call
   ;;   `imenu-prev-index-position-function' and `imenu-extract-index-name-function'.
   ;;    All the results returned by the latter are gathered into an index alist.
   ;;    This method is used if those two variables are non-nil.
```
By default `imenu-prev-index-position-function`  is `(beginning-of-defun)` which is too general for my purposes (since it halts on function calls too, e.g. `(make-variable-buffer-local 'imenu-syntax-alist)`), and the  `imenu-extract-index-name-function` which I began to work on amounted to essentially the same code as the inner loop of my `package-map-parse--alltopdefs-file` function.

>  I can see that your existing patterns miss some common cases like the `cl-` variants of `defun` and `defmacro`

This can be easily alleviated in my current code by customizing the custom variable that lists the types it should search for:

```elisp
(defcustom package-map-parse-function-shapes
  '((setq . underline) (defvar . underline) (defcustom . plain) (defun . tab) (defsubst . component) (defmacro . trapezium))
  "Define variables to look for and graphviz shapes."
  :type 'list
  :group 'package-map)
```

I am happy to proceed with the `imenu` approach if you feel it would greatly improve this package, but otherwise I think my current approach is - maybe not better - but not worse. :D

purcell:
> I am happy to proceed with the `imenu` approach if you feel it would greatly improve this package, but otherwise I think my current approach is - maybe not better - but not worse. :D

Makes sense. I'm in favour of whatever you find works best! Re. specific additions to your list, look for the `*`-suffixed variants of `defun`, `defmacro`, `defgeneric`, `defmethod`, and all the `cl-`-prefixed versions of those. And of course you have the option of macroexpanding/compiling the analysed code if that helps, provided you trust it (since that can lead to arbitrary code execution).

mtekman:
Okay, so I made the following changes:

* [X] Remove projectile dependency (current buffer directory assumed to be root)
* [X] Use `shell-quote-argument` (see: `elisp-depmap-exec.el`)
* [X] Change the project name (`package-map` → `elisp-depmap`)
* [X] Extend symbol definitions (see: `elisp-depmap-secondhelp.el`)
   - Partial. I added a `cl-` prefix check, but opted not to extend backwards using `*` or `[a-z]+` because it just felt messy.
   - If users want to add more custom types they can just list them explicitly in the `elisp-depmap-parse-function-shapes` custom variable.
   - Added missing `defgeneric` and `defmethod` (and their `cl-` counterparts).
* [ ] Implement a better definition/reference backend via either `xref` or `imenu`
   - Attempted, but it did not seem more better than my current solution as previously discussed.

 Ready for round two review!

mtekman:
Force pushed to amend rename conflict

riscy:
You need a `Package-Requires` on dash (since several of the el files make use of it) and then this looks good to go.  A couple sharp-quote suggestions:

- elisp-depmap-graph.el#L95: It's safer to sharp-quote function names; use `#'`
- elisp-depmap-graph.el#L96: It's safer to sharp-quote function names; use `#'`

Besides that I believe everything has been addressed.  Looks super cool!

mtekman:
Thanks, I just made the changes :D

riscy:
Great!

PR #6793, opened 2020-03-30
Add gtk-variant recipe

bepvte:

This package sets the theme variant (usually dark or light) of graphical Emacs' GTK theme. This is useful on desktop environments where the decorations are provided by GTK (like GNOME) to match Emacs' title bar with the rest of its window.

For example, if my GNOME desktop is on light theme, and I use a dark Emacs theme, running `gtk-variant-set-frame 'dark` will cause the windows title bar to match my Emacs theme.

https://github.com/bepvte/gtk-variant.el

I maintain the package.

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/gtk-variant
+++ b/recipes/gtk-variant
@@ -1,0 +1,1 @@
+(gtk-variant :fetcher github :repo "bepvte/gtk-variant.el")


riscy:
Since you bring in `dash`, you need a Package-Requires on it in package-map.el.

When dash is installed, there are a few minor warnings:

### package-map-graph.el ###

```
package-map-graph.el:103:14:Warning: reference to free variable
    `package-map-graph-decoratesubgraph'
package-map-graph.el:107:1:Warning: Unused lexical variable `numlines'
package-map-graph.el:107:1:Warning: Unused lexical variable `color'
```
Required but unused arguments can be prefixed with an underscore to indicate they're unused, but otherwise it's better to remove them.

```
Warning (emacs): 
package-map-graph.el:122: White space found at end of line
```

### package-map-parse.el ###

```
package-map-parse.el:39:1:Warning: defcustom for
    `package-map-parse-subclustergroups' fails to specify type
package-map-parse.el:39:1:Warning: defcustom for
    `package-map-parse-subclustergroups' fails to specify containing group
package-map-parse.el:39:1:Warning: defcustom for
    `package-map-parse-subclustergroups' fails to specify type
package-map-parse.el:39:1:Warning: defcustom for
    `package-map-parse-subclustergroups' fails to specify containing group
```
These look like you've already got a handle on fixing.

bepvte:
@riscy, I believe you replied to the wrong issue :)

riscy:
Gosh, you're right.

I'd recommend you remove the backslash in your package summary (so it's just `Set the GTK theme variant (titlebar color)`).  The other backslashes in the docstring are fine, and I know checkdoc will complain if they aren't there.

The one other thing you should check is to use `shell-quote-argument` whenever you send strings to the shell, just to handle strange input (for example is it possible that `(frame-parameter frame 'outer-window-id)` could have special characters?  or `variant`?  I may of course be missing something).

bepvte:
All fixed. I also added tags for stable.

riscy:
Great, thanks!

PR #6794, opened 2020-03-30
Add flyspell-visible package

ideasman42:

Use flyspell to show all incorrect spelling on the screen (instead of only the text just typed or any text the cursor moves over).

This means you can use flyspell _(a popular package which many users already have configured)_, displaying all visible errors (as most word processors & web browsers do for example).

https://gitlab.com/ideasman42/emacs-flyspell-visible-mode

Maintainer.

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/flyspell-visible-mode
+++ b/recipes/flyspell-visible-mode
@@ -1,0 +1,3 @@
+(flyspell-visible-mode
+ :repo "ideasman42/emacs-flyspell-visible-mode"
+ :fetcher gitlab)


riscy:
Interesting -- is this similar to running `flyspell-buffer`, which doesn't toggle the mode but still makes the spellchecks visible?

In the meantime, a quick run through the byte-compiler gives:

```
flyspell-visible-mode.el:103:1:Warning: Unused lexical variable
    `flyspell-issue-message-flag'
In end of data:
flyspell-visible-mode.el:217:1:Warning: the following functions are not known
    to be defined: flyspell-small-region, flyspell-word,
    flyspell-post-command-hook, flyspell-pre-command-hook
```

which I think is due to a missing `(require 'flyspell)`

ideasman42:
@riscy thanks for looking into this, added `(require 'flyspell)`.

The issue with `flyspell-buffer` is it's a one-off operation, if you run it on a load hook, it causes significant lag (5+ seconds in my experience) when loading large.
Even then, text insertion wont always update properly (if you paste text it wont know to update for e.g.).
----

So this package aims to solve all that and make flyspell show all on-screen errors like most other editors with spell-checking already do.

ideasman42:
Closing this, I ended up using my own spell-checker that doesn't use flyspell, will submit this instead once it's ready.

PR #6797, opened 2020-03-30
add cort recipe

conao3:

Simplify extended Emacs unit test framework

https://github.com/conao3/cort.el

Author

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/cort
+++ b/recipes/cort
@@ -1,0 +1,1 @@
+(cort :fetcher github :repo "conao3/cort.el")


riscy:
For the package-summary, I believe it should be "Simplified extended unit test framework" (rather than `Simplify`).

Beyond that everything looks perfectly fine.  The only other thing I'd add is that the `;;; Commentary` is super terse, and maybe you could add some of the text from [your usage](https://github.com/conao3/cort.el#usage) section to give people a quick heads up on what's compelling about the package.

riscy:
I don't think either of those are worth blocking on.

conao3:
Thanks!!

PR #6798, opened 2020-03-30
add display-buffer-control (dbc) package

seregaxvm:

This package is a wrapper around emacs `display-buffer-alist` facility. It allows to write declarative style rules about how to open (new/old frame, window position etc) specified buffers, matched by name, major/minor modes.

https://gitlab.com/matsievskiysv/display-buffer-control

I am the author

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/dbc
+++ b/recipes/dbc
@@ -1,0 +1,1 @@
+(dbc :fetcher gitlab :repo "matsievskiysv/display-buffer-control")


riscy:
This looks like a great idea.

It would be worth expanding your `;;; Commentary` section to give some quick examples (such as those in your README) just so people can get a quick glimpse of how the package works from the MELPA index, or even including a link to your README in the commentary would be beneficial.

The only other suggestion I have is typographic, but not blockers at all:
- accorging -> according

PR #6800, opened 2020-04-01
Add mix.el

ayrat555:

Emacs Minor Mode for Mix, a build tool that ships with Elixir.

https://github.com/ayrat555/mix.el

Author and maintainer

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/mix
+++ b/recipes/mix
@@ -1,0 +1,1 @@
+(mix :fetcher github :repo "ayrat555/mix.el")


riscy:
In case someone only receives your source file, it's a good idea to add the MIT license boilerplate (or an SPDX license line) to your source file.

Your adherence to the checklist is great (thanks!) and in the meantime some quick feedback.  It seems you're defining a minor mode for the purpose of adding keybindings, but major modes can take keybindings as well.  In fact `define-derived-mode` will automatically employ any keybindings defined in the variable `mix-mode-map`.  Moreover, since you're using `define-derived-mode`, you don't need to setq the major mode, nor the mode-name (although instead of "Mix Mode." you might want to write "Mix" on line 83).

Also - mix.el#L276--291: `define-key` at the top-level could overwrite a user's keybindings.  Try: `(defvar my-map (let ((km (make-sparse-keymap))) (define-key ...) km))`

Here is some much more minor feedback:

- mix.el#L264: It's safer to sharp-quote function names; use `#'`
- mix.el#L96: It's safer to sharp-quote function names; use `#'`
- mix.el#L151: It's safer to sharp-quote function names; use `#'`
- mix.el#L256: It's safer to sharp-quote function names; use `#'`
- mix.el#L145: It's safer to sharp-quote function names; use `#'`
- mix.el#L89: It's safer to sharp-quote function names; use `#'`
- mix.el#L47--75: Customizable variables shouldn't be private (double-dashed) -- customizable is as public as you can get :)

Hope this helps!

ayrat555:
@riscy Thanks for the feedback.

I addressed all issues except for 
> major modes can take keybindings as well. In fact define-derived-mode will automatically employ any keybindings defined in the variable mix-mode-map

Do you imply that I should set keymap in the Mix major mode? I tried that. Compilation mode completely overwrites elixir-mode if add 
```
(add-hook 'elixir-mode-hook 'mix-mode)
```

riscy:
> Do you imply that I should set keymap in the Mix major mode? ... Compilation mode completely overwrites elixir-mode if I add `(add-hook 'elixir-mode-hook 'mix-mode)`

Ah right, I misunderstood.

A couple last (non-blocking) suggestions:
- Mention `(add-hook 'elixir-mode-hook 'mix-minor-mode)` in the elisp file's Commentary section
- Perhaps alter your summary line ("Mix Major Mode") to mention Elixir ("Build Elixir using Mix") for better discoverability

Since none of those are blocking this is good to merge.  Welcome to MELPA! :)

PR #6805, opened 2020-04-03
Rename repo multi-libvterm to multi-vterm

suonlight:

This package is merged to melpa from https://github.com/melpa/melpa/pull/6512.

This PR is to update repo name from `multi-libvterm` to `multi-vterm`


tarsius:
Thanks.

You should never merge `master` into a feature branch. Rebase onto the upstream branch instead. In magit you should be able to do that by pressing `r u`.

I have now checked that out in magit to fix it and notice that my latest fix was merged before you started fixing. I am just discarding this and fix it myself as that is less work.

PR #6806, opened 2020-04-04
Add recipe for org-runbook

tyler-dodge:

org-runbook makes heirarchical runbook commands from org files accessible directly from buffers.

https://github.com/tyler-dodge/org-runbook

Maintainer and Author

N/A

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/org-runbook
+++ b/recipes/org-runbook
@@ -1,0 +1,1 @@
+(org-runbook :fetcher github :repo "tyler-dodge/org-runbook")


riscy:
### org-runbook.el ###

```
org-runbook.el:423:1:Warning: the function `projectile-project-name' is not
    known to be defined.
```
One way to deal with this is to add at the top of your Code something like `(declare-function project-name "ext:projectile.el" (&optional project))`

- No issues!

package-lint-current-buffer (using version 20200404.317):
- No issues!

Other:
- I'm curious, how come you use `eshell-command` instead of the simpler `shell-command`?
- You call `eshell-command`, but I don't see any calls to `shell-quote-argument` -- maybe you don't want to do any escaping of commands since the user is defining them directly... just double-checking.
- org-runbook.el#L344: `define-key` at the top-level could overwrite a user's keybindings.  Try: `(defvar my-map (let ((km (make-sparse-keymap))) (define-key ...) km))`
- org-runbook.el#L327: Consider `unless ...` instead of `when (not ...)`
- org-runbook.el#L334: Consider `unless ...` instead of `when (not ...)`
- org-runbook.el#L415: Consider `unless ...` instead of `when (not ...)`
- org-runbook.el#L20: Prefer `https` over `http` (if possible)

### Loadability ###

Running #'load on each file:
```
Loading org-runbook.el
Done.
```

### License ###

- GitHub API found `Other`
  - Use a GitHub-compatible format for your license file.

tyler-dodge:
Thanks for the feedback.

I pushed a new version updated with the changes that you've recommended:

https://github.com/tyler-dodge/org-runbook/compare/1.0.2...1.0.3

* I'm curious, how come you use eshell-command instead of the simpler shell-command? 

It was more or less an arbitrary choice. I use eshell a lot otherwise, so that's why I defaulted to that version. I have added an additional function for using async-shell-command and made that the default.

* You call eshell-command, but I don't see any calls to shell-quote-argument -- maybe you don't want to do any escaping of commands since the user is defining them directly... just double-checking.

Yeah, there isn't a need for shell-quote-argument since it is intended to be provided multiline script blocks.

riscy:
Okay, great.  I like `-doto` for the keybindings.

Altogether looks like an interesting package.  Welcome to MELPA!

PR #6808, opened 2020-04-04
recipes: added the soria theme

mssola:

**Soria** is a color theme which mixes:

- [xoria256](http://www.vim.org/scripts/script.php?script_id=2140): a color theme by Dmitry Zotikov, originally for Vim. I did not do the initial porting from Vim to Emacs, I took the initial code from [suxue](https://github.com/suxue/xoria256-emacs). That being said, [this](https://github.com/suxue/xoria256-emacs) project has not received any update in years and the project doesn't look maintained. Moreover, this theme does not want to be a port of xoria256 (since it also adds colors from the openSUSE guidelines). Considering all this, contributing directly to that port did not look like a good option.
- openSUSE: I took some colors from [openSUSE's Brand Guidelines](http://opensuse.github.io/branding-guidelines/).

https://github.com/mssola/soria

Maintainer.

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/soria-theme
+++ b/recipes/soria-theme
@@ -1,0 +1,1 @@
+(soria-theme :repo "mssola/soria" :fetcher github)


riscy:
Thanks for this -- nice theme, and code mostly looks good.

Some quick feedback --

### soria-theme.el ###

- soria-theme.el#L1392: Don't set `no-byte-compile: t`.
  Byte-compiling is enabled in what follows
```
soria-theme.el:64:30:Warning: assignment to free variable
    `font-lock-number-face'
In soria-theme-add-font-lock-numbers:
soria-theme.el:78:13:Warning: reference to free variable
    `font-lock-number-face'
```
These will disappear if you consider my advice down below --

- Rather than a line like `(unless (>= emacs-major-version 24) (error "The soria theme requires Emacs 24 or later!"))`, you can just add this constraint in a `Package-Requires` line at the top of your file.
- This `(add-hook 'prog-mode-hook 'soria-theme-add-font-lock-numbers)` on soria-theme.el#L1373 will be problematic for people who switch themes -- they'll be stuck with highlighting that may be arbitrarily bad for the next them they choose.  Moreover there is a popular package on MELPA that achieves this in a very comprehensive way -- https://melpa.org/#/highlight-numbers -- so a better approach would be to define faces for that mode, instead of hooking prog-mode.  
- soria-theme.el#L16: Prefer `https` over `http` (if possible)
- soria-theme.el#L27: Prefer `https` over `http` (if possible)

mssola:
Thanks a lot for the review!

I've fixed all the issues regarding lexical binding and byte compilation, plus I've also changed `http` sites to `https`. Moreover, I've removed all that number parsing I had in favor of the `highlight-numbers` package (for which I was totally unaware of, thanks!).

Thus, I think I've cleared up all your comments :smile:

riscy:
Awesome, thanks for this.  And welcome to MELPA!

Some minor non-blockers I'll leave here:
- I would recommend making `soria-theme-purple-identifiers` (interactive) so that people can reach it (discover it even) with <kbd>M-x</kbd>
- discrepencies -> discrepancies

mssola:
@riscy done, thanks!

PR #6809, opened 2020-04-05
Add spell-fu package

ideasman42:

Use show all incorrect spelling on the screen (unlike flyspell which only highlights text as you type).

Unlike other spell-checkers I've tried, this loads the dictionary into memory, making it fast enough to run as part of syntax highlighting, although by default it uses an idle timer at 1/4 of a second, as spell checking all words as you type can become irritating, and this means it doesn't add overhead (however small) when performing edits.

This package has been posted to reddit, see: [RFC](https://www.reddit.com/r/emacs/comments/futthm/rfc_spellfu_fast_elisp_highlighter_for_misspelled/).

https://gitlab.com/ideasman42/emacs-spell-fu

Maintainer.

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/spell-fu
+++ b/recipes/spell-fu
@@ -1,0 +1,3 @@
+(spell-fu
+ :repo "ideasman42/emacs-spell-fu"
+ :fetcher gitlab)


riscy:
> This package has been posted to reddit, see: RFC.

I _love_ seeing this because it looks like you got some great feedback and some code review.  I hope more authors feel encouraged to do this (and linking too).

A few very minor comments:
- spell-fu.el#L596: It's safer to sharp-quote function names; use `#'`
- and

        Warning (emacs): 
        spell-fu.el:150: Argument `fn-orig' should appear (as FN-ORIG) in the doc string

And I see spell-fu.el is GPL-licensed, but it may be worth adding a LICENSE/COPYING file that GitLab can detect using common tooling (e.g. the licensee package).

The third is the most important but I don't think any of these issues block, so this looks good to merge.

ideasman42:
Thanks for the review, all issues should be addressed now.

PR #6811, opened 2020-04-05
Add org-treeusage recipe

mtekman:

This library provides a minor mode for peeking at the line or character usage of each heading with respect to the parent heading in an org-mode file, allowing users with large org files to see the distribution of heading content and make informed decisions on where to prune, refile, or archive.

https://github.com/mtekman/org-treeusage.el

Maintainer

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/org-treeusage
+++ b/recipes/org-treeusage
@@ -1,0 +1,1 @@
+(org-treeusage :fetcher github :repo "mtekman/org-treeusage.el")


alphapapa:
:open_mouth: It's like `du` for Org trees.  Amazing.

Great animation, BTW!  ;)

mtekman:
Heh, I was actually tempted to call it `org-ncdu`

Also, thanks for showing me that lorem overlay snippet of yours -- I reworked it to do a full text replacement of my org file (otherwise I would have had +10000 overlays!)  and now I have something I can actually screenshot in live demos

> Do you want more feedback like that? It's your code, of course. :)

Always -  whatever helps me learn!

> Finally, I realize that I gave you some advice about how to organize your code, so some of this may be my fault, but anyway: you may want to consider putting all of the code in one file. None of them are very large, and having them spread out makes it a bit harder to study the code, because the reader must jump between files. Also, in the future, the Emacs native-comp compiler will be able to better optimize code that's in a single file, rather than code that calls functions spread among several files.

Ah, I did wonder why most of the packages I've seen of yours tend to be single file endeavours -- now I get it haha

So, this was actually all originally one file... and then I spent a good portion of yesterday splitting it out and making sure it compiled clean, and now I am too far in to go back. Mmm, I will definitely take this into account with future packages, but not sure if I can face putting it all this into one file again, we'll see :D  

> You use nested let forms, which is unnecessary: you can just use a single let*.

I thought it was good to use `let` where operations can be run in parallel and `let*` when they cannot, so that's why I have nested forms like that. Is there no real performance benefit in mixing the forms like that?

> The functions org-density-parse--makekey and org-density-parse--makevalues should probably be inlined rather than defined as top-level functions, because they're so simple (the --makekey function is just (cons arg1 arg2)).
> The docstrings are not quite descriptive. For example, org-density-parse--makeroot's docstring is Perform once. Get full bounds, and put into HASHER. I don't know what that means, what the function does, or what value it returns. Generally function docstrings should emphasize what value the function returns, if it returns one; otherwise it should simply say what side effects it has.

Yep, I'll make them more descriptive and I'll inline or defsubst the short functions

Thanks!

alphapapa:
> Heh, I was actually tempted to call it `org-ncdu`

If you're not too attached to the name `org-density`, I recommend seriously considering using that name instead.  On the other hand, someone might think that it brings `ncdu` output into Org.  Naming things is hard.  :)

> Also, thanks for showing me that lorem overlay snippet of yours -- I reworked it to do a full text replacement of my org file (otherwise I would have had +10000 overlays!) and now I have something I can actually screenshot in live demos

Yeah, I noticed that.  I didn't know you modified it like that, though.  That could be useful as an alternative mode.  Would you mind posting it to the repo as an issue so I can see about adding it?

> Ah, I did wonder why most of the packages I've seen of yours tend to be single file endeavours -- now I get it haha

Well, look at Emacs's own code, and you'll see many large files.  It's mostly a matter of style and preference AFAIK.  But the native compiler can, or will be able to, perform more optimizations in a single "compilation unit," which is usually a source file, so that's at least one technical reason.

> So, this was actually all originally one file... and then I spent a good portion of yesterday splitting it out and making sure it compiled clean, and now I am too far in to go back. Mmm, I will definitely take this into account with future packages, but not sure if I can face putting it all this into one file again, we'll see :D

Haha, well, like I said, mostly a matter of style, which is up to you.

> > You use nested let forms, which is unnecessary: you can just use a single let*.
> 
> I thought it was good to use `let` where operations can be run in parallel and `let*` when they cannot, so that's why I have nested forms like that. Is there no real performance benefit in mixing the forms like that?

The parallelness is a matter of semantics; the bindings are not executed on parallel cores or threads.  You could benchmark if you want, but there shouldn't be any significant performance difference either way, AFAIK.  But the code will be clearer and more idiomatic without nested `let` forms.  AFAIK there's never any reason to use them like that.

> > The functions org-density-parse--makekey and org-density-parse--makevalues should probably be inlined rather than defined as top-level functions, because they're so simple (the --makekey function is just (cons arg1 arg2)).
> > The docstrings are not quite descriptive. For example, org-density-parse--makeroot's docstring is Perform once. Get full bounds, and put into HASHER. I don't know what that means, what the function does, or what value it returns. Generally function docstrings should emphasize what value the function returns, if it returns one; otherwise it should simply say what side effects it has.
> 
> Yep, I'll make them more descriptive and I'll inline or defsubst the short functions

The inlining may improve performance, because function call overhead is a thing in Elisp, but I mainly suggested it as a matter of clarity.  So I'd suggest manually inlining them rather than using `defsubst` or `define-inline`.

mtekman:
Okay I've clean it up a little:

* [X] Removed `-mathpos.el` as all the content was inlined into `-parse.el`
* [X] Inlined a few other functions in `-parse.el` which are called within loops or maphash lambdas
* [X] Removed most nested `let` forms
* [ ] Changed the name...
    - So I am not sold on `org-density` since density sounds like it's trying to be a new org framework, same as if I had called it `org-react` or something.... 
    - I like the idea of having `du` in the name, but both `org-du` and `org-ncdu` feel short and non-descriptive. 
    - Suggestions?

alphapapa:
Of the choices you gave, org-du seems like the most descriptive and least misleading.  :)

mtekman:
> org-treeusage.

Sold!

mtekman:
Force pushed to amend rename conflict

riscy:
Awesome.  It looks like you need a package-requires on dash in your main file (in this case your main file is `org-treeusage.el`).  Actually your main file should probably be the one and only place you list Package-Requires since right now I see:

  - Package-Requires mismatch between org-treeusage-cycle.el and org-treeusage.el!
  - Package-Requires mismatch between org-treeusage-overlay.el and org-treeusage.el!

Only a few minor things besides that:

```
Warning (emacs): 
org-treeusage-overlay.el:89: Lisp symbol `org-cycle-hook' should appear in quotes
```

Sharp-quote recommendations:
- org-treeusage-cycle.el#L68: It's safer to sharp-quote function names; use `#'`
- org-treeusage.el#L71: It's safer to sharp-quote function names; use `#'`
- org-treeusage.el#L81: It's safer to sharp-quote function names; use `#'`
- org-treeusage.el#L75: It's safer to sharp-quote function names; use `#'`

I think you should reconsider using ` ɗʋ` as your minor mode lighter -- it may show up poorly in simple terminals, and macOS has/had(?) terrible unicode performance when the user's font doesn't support a character, since Emacs will noticeably lag while the OS loads another font that does support the character -- at least that's my understanding of that.

mtekman:
Thanks, I just made the changes.

>  Actually your main file should probably be the one and only place you list Package-Requires

I have been wondering about this.  So I don't actually need to write `Package-Requires` in each file of the same project, but I *should* keep the header info (Author,License,Commentary,etc)?

alphapapa:
> I have been wondering about this. So I don't actually need to write `Package-Requires` in each file of the same project, but I _should_ keep the header info (Author,License,Commentary,etc)?

Yes.

riscy:
I also recommend dropping the Package-Requires from org-treeusage-cycle.el but that's not a blocker at all.  Thanks!

novoid:
> Of the choices you gave, org-du seems like the most descriptive and least misleading. :)

Aw, man. That hurts: https://github.com/novoid/org-du

alphapapa:
> > Of the choices you gave, org-du seems like the most descriptive and least misleading. :)
> 
> Aw, man. That hurts: https://github.com/novoid/org-du

Oops, I forgot about that.  :)

PR #6812, opened 2020-04-06
Add lisp-local recipe

lassik:

Allows different Lisp indentation in each buffer via a `lisp-indent-function` wrapper that temporarily replaces global indentation-related symbol properties with local values.

https://github.com/lispunion/emacs-lisp-local

Wrote it

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/lisp-local
+++ b/recipes/lisp-local
@@ -1,0 +1,1 @@
+(lisp-local :fetcher github :repo "lispunion/emacs-lisp-local")


lassik:
The package could also be `lisp-local`. In that case, the user-settable variable would be `lisp-local-indent` instead of the slightly confusing `lisp-buffer-local-indent`.

lassik:
The package name could also be `lisp-local`. In that case, the user-settable variable would be `lisp-local-indent` instead of the slightly confusing `lisp-buffer-local-indent`.

purcell:
> The package name could also be `lisp-local`. In that case, the user-settable variable would be `lisp-local-indent` instead of the slightly confusing `lisp-buffer-local-indent`.

Yeah, that would probably be good, or perhaps `lisp-local-indent`.

lassik:
I thought `local` would be good instead of `local-indent`, since other buffer-local Lisp behavior can perhaps be added to the package later. Then people don't have to install new hooks for another package if they want the new behavior.

My goal is to keep this package minimalist so it only provides some variables for the settings, and is unopinionated about where those settings come from. Other packages could supply them.

lassik:
OK, changed package name from `lisp-buffer-local` to `lisp-local`. Checked that it works using `make recipes/lisp-local` followed by `make sandbox INSTALL=lisp-local`.

I took the opportunity to move the package from my personal GitHub account to [lispunion/emacs-lisp-local](https://github.com/lispunion/emacs-lisp-local) while at it. I have an ongoing project to collect a bunch of cross-Lisp-compatibility stuff under that organization.

riscy:
Nice.

PR #6817, opened 2020-04-07
Add recipe for icomplete-vertical

oantolin:

This package defines a global minor mode to display icomplete candidates vertically.

https://github.com/oantolin/icomplete-vertical

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/icomplete-vertical
+++ b/recipes/icomplete-vertical
@@ -1,0 +1,1 @@
+(icomplete-vertical :repo "oantolin/icomplete-vertical" :fetcher github)


riscy:
Thanks, this looks good to me -- the only thing I'd add is we generally ask for a LICENSE file that GitHub can detect (e.g. no markup) if possible (see: https://github.com/licensee/licensee )

And as a small suggestion, you might be able to shorten your summary line down from `Global minor mode to display icomplete candidates vertically` to `Display icomplete candidates vertically` (since I think a properly implemented package would use a global minor mode for this behavior anyway).

Also I'm curious, what's `(declare (indent 1))`?  The Emacs help for `defun-declarations-alist` doesn't provide a description.

oantolin:
Thanks for taking a look at the package and for your suggestions! I've added the text of the GPL 3 as a file called LICENSE and I shortened the package description at the top of `icomplete-vertical.el` as you suggested.

What `(declare (indent n))` (where `n` is some positive integer) does is tell Emacs that the first `n` arguments of the macro are special and must be indented further than the remaining ones. It is used for macros that end in an implicit `progn`, to specifiy where the implicit `progn` starts.

This is documented in `(info "(Elisp) Indenting Macros")`.

riscy:
Ah right it's a macro.  Thank you for enlightening me.

Also, this looks good to go.  Thanks, and welcome to MELPA! :)

oantolin:
Amazing, thanks!

PR #6821, opened 2020-04-09
Add OpenSearch descriptor

nagy:
This adds an [OpenSearch] descriptor to the site.

[OpenSearch]: https://developer.mozilla.org/en-US/docs/Web/OpenSearch


purcell:
Yes, very cool - thanks!

PR #6822, opened 2020-04-09
Telega recipe

zevlg:
Added "contrib" directory to the files.

"contrib" directory contains add-ons contributed to the telega project.  Such as:

* ol-telega.el - Links to chats and messages in Org mode
* telega-url-shorten.el - Global minor mode to shorten urls in messages
* telega-status-history.el - Global minor mode to collect online user status information

### DIFF
--- a/recipes/telega
+++ b/recipes/telega
@@ -1,3 +1,3 @@
 (telega :fetcher github
         :repo "zevlg/telega.el"
-        :files (:defaults "etc" "server" "Makefile"))
+        :files (:defaults "etc" "server" "contrib" "Makefile"))
--- a/recipes/telega
+++ b/recipes/telega
@@ -1,4 +1,3 @@
 (telega :fetcher github
         :repo "zevlg/telega.el"
-        :branch "master"
         :files (:defaults "etc" "server" "Makefile"))


riscy:
I may have to defer to someone with more experience on this one, but this feels like it should be a separate package (maybe telegra-contrib).

But probably the best possible option is that, when someone makes a "contrib" package, they just create their own package on MELPA, specify its dependencies, and make sure that telega itself is a dependency too.  They could `@` you in their PR and have you chime in on any conventions you'd expect, etc.

The other problem here is I think dependency management is going to be more challenging, just because of the uncontrolled nature of a "contrib" directory.  For example right now I see ol-telega.el importing `ol` but that isn't declared as a Package-Requires (I think it needs to be), and the functions it contains start with `org-` instead of your expected prefix (`telega-`).  And telega-url-shorten.el has a requirement on all-the-icons, and so on.

zevlg:
Yeah, I was in doubt about adding "contrib/" to telega recipe, but few people asked me to do so.  I was planning to spread "contrib/" files only by github sources or by separate MELPA packages with correct dependencies

I'm closing this PR, thanks!

PR #6828, opened 2020-04-12
Add org-multi-wiki

akirak:

Org-based wiki system for multiple workgroups with namespace support.

https://github.com/akirak/org-multi-wiki

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/org-multi-wiki
+++ b/recipes/org-multi-wiki
@@ -1,0 +1,2 @@
+(org-multi-wiki :fetcher github :repo "akirak/org-multi-wiki"
+                :files ("org-multi-wiki.el"))


riscy:
`(add-hook 'org-mode-hook 'org-multi-wiki-check-buffer)` should be something a user adds to their own configuration, usually not something done by a package.  Alternatively this looks like a good opportunity to implement a globalized minor mode.

And the automated stuff:

### org-multi-wiki.el ###

```
org-multi-wiki.el:43:1:Error: Cannot open load file: No such file or directory, ol
```
I think this needs a `Package-Requires` on a later version of org-mode, since Emacs 26.3 doesn't have it.

- No issues!

- No issues!

- org-multi-wiki.el#L277: It's safer to sharp-quote function names; use `#'`
- org-multi-wiki.el#L26: Prefer `https` over `http` (if possible)

akirak:
Thanks for the comment. I think I've fixed those issues.

riscy:
Thanks!

PR #6830, opened 2020-04-12
Add owcmd package

fishyfriend:

This package provides a command to switch to the next window, run a command, then switch back to the original window automatically. Use it to save keystrokes when you need to do just one thing in the other window, like scroll the buffer, jump to a definition, paste some text, or quit a temporary buffer. It is a sort of generalization of the built-in commands `scroll-other-window` and `scroll-other-window-down`.

https://github.com/fishyfriend/execute-other-window

I am the author.

**None needed**

### Checklist

Please confirm with `x`:

### Note on prior art

There is an existing package, [owdriver](https://github.com/aki2o/owdriver), that provides similar functionality. I believe execute-other-window provides unique value and should stand on its own. **TLDR:** execute-other-window avoids owdriver's keymap-driven approach and this offers better compatibility and convenience for some use cases.

#### Details

owdriver maintains a keymap of bindings the user wants available in the other window, where each key (outside of some defaults) must be configured individually. This offers some finer control over what keys are available in the other window, but has several disadvantages:
* labor-intensive if all you want is for all keys to work by default.
* would require extra dispatch code to set up keybindings for multiple modes that bind the same keys.
* hard to integrate with tools like Evil that override existing keymaps.

The keymap-driven approach does not seem ideal for these cases; there are probably solutions but they would require adding a lot of code complexity. Moreover, even with solutions to these problems, as a user of both Evil and God Mode, my life is greatly simplified if I can avoid adding yet another "weird" keymap to an already challenging keymap scheme. Thus, for me anyway, execute-other-window makes sense as an alternative.

### DIFF
--- a/recipes/owcmd
+++ b/recipes/owcmd
@@ -1,0 +1,1 @@
+(owcmd :fetcher github :repo "fishyfriend/owcmd")


riscy:
The license and code and documentation look nice on a glance.  And the comparison to prior art above is hugely appreciated.  Here's a quick run over the checklist:

### execute-other-window.el ###

- No issues!

```
Warning (emacs): 
execute-other-window.el:0: The first line should be of the form: ";;; package --- Summary"
```
Actually you'll want:
`;;; execute-other-window.el --- Run a single command in the other window -*- lexical-binding: t; -*-`

```
12 issues found:
84:0: error: "eow-cancel-reselect-on-new-window" doesn't start with package's prefix "execute-other-window".
89:0: error: "eow-cancel-reselect-commands" doesn't start with package's prefix "execute-other-window".
100:0: error: "eow-ignore-commands" doesn't start with package's prefix "execute-other-window".
115:0: error: "eow-ignore-prefix-commands" doesn't start with package's prefix "execute-other-window".
138:0: error: "eow--ignore-prefix" doesn't start with package's prefix "execute-other-window".
141:0: error: "eow--calling-window" doesn't start with package's prefix "execute-other-window".
144:0: error: "eow--target-window" doesn't start with package's prefix "execute-other-window".
147:0: error: "eow--match-in-list" doesn't start with package's prefix "execute-other-window".
161:0: error: "eow--handle-command-pre" doesn't start with package's prefix "execute-other-window".
166:0: error: "eow--handle-command-post" doesn't start with package's prefix "execute-other-window".
180:0: error: "eow--clean-up-state" doesn't start with package's prefix "execute-other-window".
189:0: error: "eow-execute-other-window" doesn't start with package's prefix "execute-other-window".
```
Perhaps the package should just be called eow?

fishyfriend:
Thank you very much for the feedback!

I've fixed the first line comment issue.

To unify the package name and prefix, I have renamed this package "owcmd" as I feel it will help people remember what the package does and hints at its similarity to owdriver. The linter passes fully now.

I have amended this PR's commit to use my renamed package repo. I think my force-push may have caused the Travis CI build to fail. However, I was able to build, install, and use the revised package in a sandboxed Emacs locally.

riscy:
I think the explanation you gave above as to the difference between your package and owdriver warrants being added to your README or possibly even the package commentary -- it's valuable to the community to see thos comparisons -- but I'll mention again that it's appreciated.  I'll have to give both packages a shot. :)

And welcome to MELPA!

fishyfriend:
Happy to be here! You raised a good point re: documenting the comparison to the other similar package. I've done as you suggest.

PR #6831, opened 2020-04-12
Add leaf-manager recipe

conao3:

Configure manager for leaf based init.el

https://github.com/conao3/leaf-manager.el

Author.

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/leaf-manager
+++ b/recipes/leaf-manager
@@ -1,0 +1,1 @@
+(leaf-manager :fetcher github :repo "conao3/leaf-manager.el")


riscy:
Just some minor notes.

- leaf-manager-template-copyright-to looks like its :type should be `:type '(choice regexp nil)` rather than the `sexp` catchall

Otherwise, the code looks good but I can offer some editing hints:

- "Configure manager for leaf based init.el" -> "Configuration manager
- "Value as string use straightly." -> "When the value is a string it is used directly."
- editting -> editing
- "invoke this in other than" -> "invoke this except in"
- confierm -> confirm
- lefa-manger-header-line -> leaf-manager-header-line
- manger -> manager (a few occurrences)
- leaf-manager-templater-feature-name -> leaf-manager-template-feature-name
- cannot writable -> cannot be written
- File exist (%s) -> File exists (%s)
- Unknown sexp exist -> Unknown sexp exists

riscy:
Your animations are great, btw.

conao3:
Thanks!

PR #6834, opened 2020-04-13
New recipe: prf-shell

p3r7:

More user-friendly shell APIs for Emacs.

Split in 3 files:
 - `prf-shell-command.el` for single shell commands ([blog post](https://www.eigenbahn.com/2020/01/19/painless-emacs-shell-commands))
 - `prf-shell.el` for interactive shells ([blog post](https://www.eigenbahn.com/2020/01/21/painless-emacs-interactive-shells))
 - `prf-remote-shell.el` for interactive remote shells

Please note that I'm unsure if these need be distributed as a single package or split in 3 and need some guidance here (package authoring n00bie spotted).

https://github.com/p3r7/prf-shell

Maintainer.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/friendly-remote-shell
+++ b/recipes/friendly-remote-shell
@@ -1,4 +1,4 @@
 (friendly-remote-shell :repo "p3r7/friendly-remote-shell"
                        :fetcher github
                        :files (:defaults
-                               (:exclude "friendly-shell-command.el" "friendly-remote-shell.el")))
+                               (:exclude "friendly-shell-command.el" "friendly-shell.el")))
--- a/recipes/friendly-remote-shell
+++ b/recipes/friendly-remote-shell
@@ -1,0 +1,4 @@
+(friendly-remote-shell :repo "p3r7/friendly-remote-shell"
+                       :fetcher github
+                       :files (:defaults
+                               (:exclude "friendly-shell-command.el" "friendly-remote-shell.el")))
--- a/recipes/friendly-shell
+++ b/recipes/friendly-shell
@@ -1,0 +1,4 @@
+(friendly-shell :repo "p3r7/friendly-shell"
+                :fetcher github
+                :files (:defaults
+                        (:exclude "friendly-shell-command.el" "friendly-remote-shell.el")))
--- a/recipes/friendly-shell-command
+++ b/recipes/friendly-shell-command
@@ -1,0 +1,4 @@
+(friendly-shell-command :repo "p3r7/friendly-shell-command"
+                        :fetcher github
+                        :files (:defaults
+                                (:exclude "friendly-shell.el" "friendly-remote-shell.el")))
--- a/recipes/prf-remote-shell
+++ b/recipes/prf-remote-shell
@@ -1,4 +1,0 @@
-(prf-remote-shell :repo "p3r7/prf-remote-shell"
-                  :fetcher github
-                  :files (:defaults
-                          (:exclude "prf-shell-command.el" "prf-remote-shell.el")))
--- a/recipes/prf-remote-shell
+++ b/recipes/prf-remote-shell
@@ -1,0 +1,4 @@
+(prf-remote-shell :repo "p3r7/prf-remote-shell"
+                  :fetcher github
+                  :files (:defaults
+                          (:exclude "prf-shell-command.el" "prf-remote-shell.el")))
--- a/recipes/prf-shell
+++ b/recipes/prf-shell
@@ -1,4 +1,0 @@
-(prf-shell :repo "p3r7/prf-shell"
-           :fetcher github
-           :files (:defaults
-                   (:exclude "prf-shell-command.el" "prf-remote-shell.el")))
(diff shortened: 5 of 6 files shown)


p3r7:
### One package VS several

As stated in first message, I'm unsure if these need be distributed as a single package or split in 3.

I've declared `prf-shell` as a dependency of `prf-remote-shell` but this naturally doesn't please checkdoc: `prf-remote-shell.el:8:53: error: Package prf-shell is not installable.`


### Package(s) name(s)

I'm unsure about the package(s) name(s).

I took the habit to prefix my private packages in my config with `prf-` (shorthand for my pseudonym) but this in not very explicit.

Maybe something like `friendly-shell-command` / `friendly-shell` / `friendly-remote-shell` would be better (and a nod to dependency [friendly-tramp-path](https://github.com/p3r7/friendly-tramp-path))?

### Taming linting of tramp-make-tramp-file-name

Starting with Emacs 26, `tramp-make-tramp-file-name` went from needing 4-5 arguments to 6-7.

I've dealt with it like so:

```el
(if (>= emacs-major-version 26)
        (setq path (tramp-make-tramp-file-name method user domain host port localname))
      (setq path (tramp-make-tramp-file-name method user host localname)))
```

But this does not please the linter / byte-compiler. I guess some `eval-when-compile` sorcery is needed here but still need to investigate.

### prf-async-shell-command VS prf-shell-command-async

Single shell commands in Emacs are spawned with `shell-command` and `async-shell-command`.

As a result, I've named my wrappers `prf-shell-command` and `prf-async-shell-command`, in the same file `prf-shell-command.el`.

This does not please the linter: `error: "prf-async-shell-command" doesn't start with package's prefix "prf-shell-command".`

I hesitate between rename it to `prf-shell-command-async` (breaking the naming scheme of parent functions) or moving it to a separate file / package.

riscy:
Nice, I like this.  Quick run-through --

### prf-remote-shell.el ###

```
prf-remote-shell.el:44:1:Error: Cannot open load file: No such file or directory, friendly-tramp-path
```
I think this one's just a missing package-requires.

- No issues!

package-lint-current-buffer (using version 20200415.606):
```
4 issues found:
52:0: error: "prf-remote-shell-buffer-name-regexp" doesn't start with package's prefix "prf-shell".
58:0: error: "prf-remote-shell--make-tramp-file-name" doesn't start with package's prefix "prf-shell".
86:0: error: "prf-remote-shell" doesn't start with package's prefix "prf-shell".
115:0: error: "prf-remote-shell-register-display-same-window" doesn't start with package's prefix "prf-shell".
```
These should probably be fixed since the package's "namespace" is `prf-shell` (just in case in the future someone made a `prf-remote` package for example).

### prf-shell-command.el ###

- No issues!

- No issues!

package-lint-current-buffer (using version 20200415.606):
```
1 issue found:
59:0: error: "prf-async-shell-command" doesn't start with package's prefix "prf-shell".
```


### prf-shell.el ###

- No issues!

- No issues!

package-lint-current-buffer (using version 20200415.606):
- No issues!




### Packaging ###


- Package-Requires: cl-lib "0.6.1", with-shell-interpreter "0.1.0", emacs "24.1"
  - Package-Requires mismatch between prf-remote-shell.el and prf-shell.el!

It's best if only the main file has the package-requires in it, since those are the only ones that will take effect.

One other thing I'll add is that the `Commentary` in prf-shell.el could be fleshed out a bit, even with a link.  One reason for that is that MELPA uses that in the index of its packages, e.g. [here](https://melpa.org/#/with-shell-interpreter).   An actual link to the README is more useful, but a summary is the most useful.

Hope this helps!

riscy:
Oh and some other feedback.

> I'm unsure if these need be distributed as a single package or split in 3.

I think based on the above (different prefixes, different package-requires) this is starting to look like a good candidate to be split into different packages -- especially if your users can use the packages in a modular fashion, and so can avoid dependencies to minimize their Emacs packages if they desire.

> Maybe something like friendly-shell-command / friendly-shell / friendly-remote-shell would be better (and a nod to dependency friendly-tramp-path)?

Actually I like these names better.  I don't think there's anything wrong with prefixes that indicate the author per se, but I don't think they add much for users (and so I would personally steer away from them).

> I hesitate between rename it to prf-shell-command-async (breaking the naming scheme of parent functions) or moving it to a separate feature / package.

One option if you want to not break the naming convention of the parents would be to use names like `friendly-shell-async-shell-command`.  Yeah, it's wordy, but it's strict about the prefix (which is important for MELPA packages) and also includes a full copy of the function it's modeled after.  And it's not _that_ bad.

p3r7:
The git log started to get dirty so I created https://github.com/melpa/melpa/pull/6851 to respect the convention of 1 commit per recipe.

> different prefixes, different package-requires [...] good candidate to be split into different packages

OK!
I was struggling to find best practices regarding package / feature splitting. Thanks for the clear explanation.

> Actually I like these names better. 

I've renamed the package with the `friendly-` prefix.

> use names like `friendly-shell-async-shell-command`

Yeah, that's having fully-qualified names VS short ones.
I guess I have to choose one or the other for a public (user-facing) fns in a package.

I went with the short solution (i prefer them in elisp as we have no file-local aliases like in python).

So I've renamed `prf-async-shell-command` into `prf-shell-command-async`.

PR #6835, opened 2020-04-13
Fix competitive-programming-snippets recipe

sei40kr:
# Brief summary of what the package does

Please see https://github.com/melpa/melpa/pull/6627.

https://github.com/sei40kr/competitive-programming-snippets

A maintainer

### Checklist


tarsius:
Please follow this advice from the readme:

> If the package merely requires some additional files, for example for bundling external dependencies, but is otherwise fine with the defaults, it's recommended to use :defaults as the very first element of this list, which causes the default value shown above to be prepended to the specified file list.

PR #6840, opened 2020-04-15
Make bibtex-completion a separate package

tmalsburg:
Useful for org-roam-bibtex and generally cleaner.

bibtex-completion.el used to be part of packages helm-bibtex and ivy-bibtex. First this is redundant, second there are now other packages that need bibtex-completion but do not depend on helm-bibtex or ivy-bibtex.

https://github.com/tmalsburg/helm-bibtex

This repo contains bibtex-completion, helm-bibtex, and ivy-bibtex since they are tightly related and developed in tandem.

I'm the maintainer.

None needed

### Checklist

Please confirm with `x`:

I did not manage to make checkdoc 100% happy (some lines are too long).  Shame on me.

### DIFF
--- a/recipes/bibtex-completion
+++ b/recipes/bibtex-completion
@@ -1,0 +1,1 @@
+(bibtex-completion :fetcher github :repo "tmalsburg/helm-bibtex" :files ("bibtex-completion.el"))
--- a/recipes/helm-bibtex
+++ b/recipes/helm-bibtex
@@ -1,1 +1,1 @@
-(helm-bibtex :fetcher github :repo "tmalsburg/helm-bibtex" :files ("helm-bibtex.el" "bibtex-completion.el"))
+(helm-bibtex :fetcher github :repo "tmalsburg/helm-bibtex" :files ("helm-bibtex.el"))
--- a/recipes/ivy-bibtex
+++ b/recipes/ivy-bibtex
@@ -1,1 +1,1 @@
-(ivy-bibtex :fetcher github :repo "tmalsburg/helm-bibtex" :files ("ivy-bibtex.el" "bibtex-completion.el"))
+(ivy-bibtex :fetcher github :repo "tmalsburg/helm-bibtex" :files ("ivy-bibtex.el"))


riscy:
While here I notice that bibtex-completion.el needs a `Package-Requires` on biblio.  Non-blocking since it's already on MELPA of course, but definitely should be added.  And thanks!

tmalsburg:
Thanks. I've just added the missing biblio dependency.

tarsius:
I would rename the repository, but that's just me (an ivy user :stuck_out_tongue_winking_eye: ).

tmalsburg:
Yeah, the name is not ideal, but there are a ton of links to this repo on webpages and in people's configs and elsewhere. By renaming we'd create a lot of unnecessary work for everyone.

tarsius:
In case you didn't know; web and git clients are redirected, api clients however are not.

tmalsburg:
Wasn't aware, thanks!

tmalsburg:
Are you suggesting "bibtex-completion"? Just to avoid misunderstandings.

tarsius:
Yes.

bdarcus:
> I would rename the repository, but that's just me (an ivy user  ).

This is long merged, but just to weigh in on this:

I think ideal longer term is a repo for each project.  Could even create a github org to hold them if there might be some advantage to that.

tmalsburg:
> But I also think ideal longer term may be a repo for each project. Could even create a github org to hold them if there might be some advantage to that.

I'm not strictly against it, but what do you think would be the benefits?

BTW, I will try to catch up with recent related discussions over the next couple of days. Thanks for the patience.

bdarcus:
> I'm not strictly against it, but what do you think would be the benefits?

"It" meaning the org, or the separate repos?

The latter just makes it easier to have more focused docs, issue trackers, even development.

I think the org is less obvious actually, maybe more if you want to be able to manage different groups of people with different permissions.

tmalsburg:
> "It" meaning the org, or the separate repos?

Separate repos. I think for myself it would slightly complicate workflows and not really buy me anything.  Helm-bibtex and ivy-bibtex especially are so similar that I find it useful to have them in the same repo.  They even share the documentation. Another reason against new repos is that they'd lose the git history, historic issues, and so on.  Also a lot of links out on the web would suddenly become outdated.

PR #6841, opened 2020-04-16
Add pdb-capf

muffinmad:

completion-at-point function for pdb (Python debugger)

https://github.com/muffinmad/emacs-pdb-capf

Maintainer

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/pdb-capf
+++ b/recipes/pdb-capf
@@ -1,0 +1,1 @@
+(pdb-capf :fetcher github :repo "muffinmad/emacs-pdb-capf")


riscy:
Thanks!  The other checks are good, but when byte-compiling the file I get:

```
pdb-capf.el:76:1:Error: Symbol's value as variable is void: rx-constituents
```
This is actually due to the `python` package not having `(require 'rx)` in it, so not your bug per se, but for a clean byte-compile you will need to add `(require 'rx)` to your package.  Also, since you use `string-empty-p` you'll need `(require 'subr-x)`.

Meanwhile the package looks unique to me, but have you had a chance to check out and compare to the other completion packages listed [here](https://www.emacswiki.org/emacs/PythonProgrammingInEmacs#toc8)?

purcell:
> have you had a chance to check out and compare to the other completion packages listed [here](https://www.emacswiki.org/emacs/PythonProgrammingInEmacs#toc8)?

I _think_ this is fairly unique, in that it targets `pdb`. One thing I noticed is that completion will crash if triggered after `'''`, or maybe even just `'`: probably want to escape the completion string somehow when interpolating it into the snippet.

muffinmad:
> Thanks! The other checks are good, but when byte-compiling the file I get:
> 
> byte-compile-file (using Emacs 26.3):
> 
> ```
> pdb-capf.el:76:1:Error: Symbol's value as variable is void: rx-constituents
> ```
> 
> This is actually due to the `python` package not having `(require 'rx)` in it, so not your bug per se, but for a clean byte-compile you will need to add `(require 'rx)` to your package. Also, since you use `string-empty-p` you'll need `(require 'subr-x)`.

Thanks! Fixed now.

> Meanwhile the package looks unique to me, but have you had a chance to check out and compare to the other completion packages listed [here](https://www.emacswiki.org/emacs/PythonProgrammingInEmacs#toc8)?

@purcell is right. This package targets not `python-mode` but `pdb`. Completions `pdb` debugging session can be used in e.g. `gud-mode`.

muffinmad:
> One thing I noticed is that completion will crash if triggered after `'''`, or maybe even just `'`: probably want to escape the completion string somehow when interpolating it into the snippet.

Good catch! Never tried to complete `'''` while debugging :) I think those characters must not be passed to completion function at all. Fixed now. Thanks!

riscy:
Ok, great - I'm looking forward to trying this out.

PR #6842, opened 2020-04-16
Add recipe for orderless

oantolin:

Completion style that matches multiple regexps in any order

https://github.com/oantolin/orderless

I am the author and maintainer

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/orderless
+++ b/recipes/orderless
@@ -1,0 +1,1 @@
+(orderless :repo "oantolin/orderless" :fetcher github)


riscy:
Quick first pass --

```
orderless.el:184:1:Warning: the function `orderless--highlight-matches' is not
    known to be defined.
```

The form that contains the defun seems a little bit unusual too (but I may not be familiar with what's going on there) so the issue might be lurking in there.

oantolin:
I just wanted to avoid another global variable, when it is a constant used only in that one function. But if the byte-compiler complains about not finding the function definition inside the `let`, I guess I could flatten the structure. (I think the real fix is making the byte-compiler slightly smarter, but I'm definitely not going to try doing that... :D.)

I'll comment back here when I've made the change.

oantolin:
Done, @riscy . Sorry about that. I guess I only byte-compiled it when that function was already defined, so I never noticed the byte-compiler's blindness to let forms.

riscy:
No worries!  This looks great.  Thanks. :)

PR #6845, opened 2020-04-17
Recipe for new org-special-block-extras package

alhassy:

Provides new custom blocks for org-mode;
supported backends are HTML and LaTeX.

https://github.com/alhassy/org-special-block-extras

Maintainer

**None needed** 

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/org-special-block-extras
+++ b/recipes/org-special-block-extras
@@ -1,0 +1,1 @@
+(org-special-block-extras :fetcher github :repo "alhassy/org-special-block-extras")


riscy:
### org-special-block-extras.el ###

```
org-special-block-extras.el:57:1:Error: Cannot open load file: No such file or directory, dash-functional
```
This just looks like a missing package-requires on dash-functional.

But even if I fixed that my byte-compiler halts with `Not enough arguments for format string` which is due to this string:
```elisp
(cl-loop for colour in org-special-block-extras--colors
      do (eval (read (format
                      "(defun org-special-block-extras--%s (backend contents)
                     (format (pcase backend
                     (`latex \"\\\\begingroup\\\\color{%s}%%s\\\\endgroup\")
                     (`html  \"<div style=\\\"color:%s;\\\">%%s</div>\")
                     (t      \"org-special-block-extras: Unsupported backend\"))
                     contents))"
                      colour colour colour))))
```
One more thing for right now - inside the code I also notice instead of the deprecated`loop` you should use `cl-loop`.

```
Warning (emacs): 
org-special-block-extras.el:174: White space found at end of line
```

package-lint-current-buffer (using version 20200415.606):
- No issues!

- org-special-block-extras.el#L22: Prefer `https` over `http` (if possible)

purcell:
> This just looks like a missing package-requires on dash-functional.

Just a quick drive-by note: definitely consider using `pcase-let*` and `apply-partially` instead of `-let*` and `-partial`. Since you're already using `cl-lib`, there's no particular need for `dash` here -- nice to keep things standard if you can. :-)

alhassy:
@purcell I find the number of backquotes and commas a bit much with `pcase-let` and instead prefer `seq-let`, but that one also seems to have less utility than `-let` ---although `seq-let` is more general, working on arbitrary sequences unlike `-let`.
I did not know about `apply-partially`; thanks!

@riscy Thanks for the feedback.

Using *Emacs 26.3* and *Checkdoc 0.6.1*,

0. I do not see any error regarding a shortage of arguments for that cited string.
    - I personally do not see an error there; perhaps you could point it out to me.
    - I have removed the `t` case; in case that makes things better on your end.

1. According to `C-h o RET loop`, it seems that `cl-loop` and `loop` are aliases.
   - Regardless, I have performed the suggested change.

Thanks for your time ^_^

riscy:
Whatever you've changed seems to have made byte-compilation on my end work -- here's the latest scroll.

```
org-special-block-extras.el:206:11:Warning: assignment to free variable
    `org-export-allow-bind-keywords'
org-special-block-extras.el:214:7:Warning: assignment to free variable
    `org-export-allow-bind-keywords'
In end of data:
org-special-block-extras.el:252:1:Warning: the following functions are not
    known to be defined: org-html-special-block, org-latex-special-block
```
It looks like at least one of these is due to having the `(setq org-export-allow-bind-keywords t)` form at the top level.

The other not-known errors can be solved by not sharp-quoting but I think those `advice-add` forms are problematic to begin with as I take a deeper look at the code -- packages shouldn't modify (i.e. advise) Emacs functions simply when a package is loaded.  Instead it should be up to the user to toggle the behavior off or on again.  It looks like org-special-block-extras should be a global minor mode, instead?

- org-special-block-extras.el#L124: It's safer to sharp-quote function names; use `#'`
- org-special-block-extras.el#L125: It's safer to sharp-quote function names; use `#'`

alhassy:
The idea to use a minor mode is definitely the way to go; I completely forgot about that!

Is there anything I could do to see the warnings you see? ---I've shifted to Emacs 28.

To avoid the `org-export-allow-bind-keywords` free variable issue, I thought of
wrapping that call in `with-no-warnings`, but have opted to simply ignore setting it altogether
and instead inform users to do so themselves. What would you advise for such a situation?

Thanks for the hard work!

riscy:
Here's the latest (I don't believe these are due to using Emacs 28) and I'll inline some suggestions, too --

### org-special-block-extras.el ###

```
org-special-block-extras.el:78:15:Warning: assignment to free variable
    `org-export-allow-bind-keywords'
org-special-block-extras.el:174:1:Warning: Unused lexical argument `object'
org-special-block-extras.el:174:1:Warning: Unused lexical argument `window'
In org-special-block-extras--latex-definitions:
org-special-block-extras.el:198:9:Warning: reference to free variable `for'
org-special-block-extras.el:202:43:Warning: reference to free variable `that'
org-special-block-extras.el:198:25:Warning: reference to free variable `in'
org-special-block-extras.el:202:9:Warning: reference to free variable `do'
org-special-block-extras.el:202:28:Warning: reference to free variable `this'
In org-special-block-extras--edcomm:
org-special-block-extras.el:285:11:Warning: assignment to free variable
    `org-export-allow-bind-keywords'
org-special-block-extras.el:293:1:Warning: Unused lexical variable `cadr'
org-special-block-extras.el:293:1:Warning: Unused lexical variable `&plist'
org-special-block-extras.el:293:1:Warning: Unused lexical argument `object'
org-special-block-extras.el:293:1:Warning: Unused lexical argument `window'
org-special-block-extras.el:304:60:Warning: reference to free variable `path'
org-special-block-extras.el:343:1:Warning: Unused lexical argument
    `description'
org-special-block-extras.el:381:1:Warning: Unused lexical argument
    `description'
org-special-block-extras.el:452:7:Warning: reference to free variable `for'
org-special-block-extras.el:459:23:Warning: reference to free variable `link'
org-special-block-extras.el:452:25:Warning: reference to free variable `in'
org-special-block-extras.el:459:11:Warning: reference to free variable `link′'
org-special-block-extras.el:459:17:Warning: reference to free variable `='
org-special-block-extras.el:462:36:Warning: reference to free variable
    `social'
org-special-block-extras.el:460:7:Warning: reference to free variable `do'
In end of data:
org-special-block-extras.el:469:1:Warning: the following functions are not
    known to be defined: org-html-special-block, org-latex-special-block,
    org-special-block-extras--parallel, org-special-block-extras--2parallel,
    org-special-block-extras--parallelNB,
    org-special-block-extras--2parallelNB, org-link-set-parameters,
    org-element-context, loop, this, first, second, third, social
```
For any unused lexical arguments, it's easy.  If you can remove it, remove it -- but if you can't, you can prefix the variable name with `_` (e.g. `_window`) and the byte-compiler will be happy.

You should replace your deprecated call to `loop` with `cl-loop` (same with `cl-first`, `cl-second`, and `cl-third`).

Since you're using `org-link-set-parameters`, you should `(require 'org)`, but you should probably also `(require 'ox-latex)` and `(require 'ox-html)`.

I think the remaining byte-compile issues point to possible bugs in the implementation (for example `(defalias #'org-special-block-extras--parallel #'org-special-block-extras--2parallel)` should probably be written `(defalias 'org-special-block-extras--parallel #'org-special-block-extras--2parallel)` since only one of them is a function at that moment, but I can't find either symbol anywhere else in the file).

Package-lint looks fine although I'm not familiar with noweb-refs.

- No issues!

alhassy:
I've made the given suggestions.

- Checkdoc: No issues
- Package Lint: Two warnings about closing parens on new lines ---necessary for noweb-refs.
- Byte-compile-file: No issues

The `org-special-block-extras--2parallel` function and some 20+ others are defined using `eval`.

Thank-you for your time!

riscy:
I'm curious, is the byte-compiler giving you nothing at all?  Here is the latest.

```
org-special-block-extras.el:650:1:Warning: Unused lexical argument
    `description'
org-special-block-extras.el:673:14:Warning: reference to free variable `_'
In end of data:
org-special-block-extras.el:698:1:Warning: the following functions are not
    known to be defined: org-special-block-extras--2parallel,
    org-special-block-extras--2parallelNB
```

The description variable could just be _description and that will mute that warning, similar to the discussion above.  The free variable `_` looks like a bug to me, since the `cond` form uses `t` to indicate `else`, and `_` is bound to nothing at that point.

As for the defaliased functions -- on a closer look, now I see where they come from.  The metaprogrammy loop that defines those functions is not evaluated during compilation but you can solve that by wrapping their definitions in an [eval-when-compile](https://www.gnu.org/software/emacs/manual/html_node/elisp/Eval-During-Compile.html) form.  That form will also have to come _before_ you refer to the functions in the code, and then you'll be there.

Lastly --

- org-special-block-extras.el#L651: No `format` required; messages are already f-strings

Hope this helps. :)

alhassy:
Hey Chris, thanks man: I've made the changes and none of checdoc, the linter, nor the compiler mention any issues except the closing parens on newlines due to the noweb-refs.

Thanks for the eval-when-compile and the message tip ;-)

riscy:
Oops, I see my oversight -- I think you are compiling the file in an Emacs session that has already loaded the package, so that's why you're probably not seeing the compiler warnings.  Currently there's even a byte-compile _error_.

I was going to open a PR but, just to be quick, you can remove the `eval-when-compile` (sorry about that, that's incorrect), so the code will look like this:
```elisp
(defvar org-special-block-extras--colors
  '(black blue brown cyan darkgray gray green lightgray lime
          magenta olive orange pink purple red teal violet white
          yellow)
  "Colours that should be available on all systems.")

(cl-loop for colour in org-special-block-extras--colors
      do (eval (read (format
                      "(defun org-special-block-extras--%s (backend contents)
                     (format (pcase backend
                     (`latex \"\\\\begingroup\\\\color{%s}%%s\\\\endgroup\\\\,\")
                     (_  \"<span style=\\\"color:%s;\\\">%%s</span>\"))
                     contents))"
                      colour colour colour))))
```
Then at the top of your file, you can just forward-declare the two functions that you create aliases for to keep the compiler clean:

```elisp
(require 'org)
(require 'ox-latex)
(require 'ox-html)

(declare-function org-special-block-extras--2parallel "org-special-block-extras" t t)
(declare-function org-special-block-extras--2parallelNB "org-special-block-extras" t t)
```
I'd recommend looking into flycheck for emacs lisp -- it will give you a heads-up about issues you might not otherwise see.

alhassy:
Hey Chris, I did the following
```
emacs -batch -f batch-byte-compile *.el
```
But that resulted in
```
org-special-block-extras.el:57:1:Error: Cannot open load file: No such file or directory, s
```

This is the same error that flycheck gives me.

Any suggestions?

---

I've removed both `eval-when-compile` and placed the forward declarations at the place indicated ---leaving the `defalias` already there.

riscy:
Are you still facing that error?  I'm not seeing it anymore, and this is looking largely good to go, but thought I'd check before I hit merge.

By the wya, in case it is helpful, I've written up a little GitHub actions snippet that you can add to your repo (fairly painlessly) to get the same checks I'm running working there: https://github.com/riscy/melpazoid#add-it-to-github-actions

alhassy:
Thanks for the link man ---all is good on my end ^_^

riscy:
Great, thanks.  And, welcome to MELPA! :)

PR #6848, opened 2020-04-18
Add scratch-comment

conao3:

Insert Elisp result as comment in scratch buffer.

https://github.com/conao3/scratch-comment.el

Author.

None needed.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/scratch-comment
+++ b/recipes/scratch-comment
@@ -1,0 +1,1 @@
+(scratch-comment :fetcher github :repo "conao3/scratch-comment.el")


alphapapa:
This function from my config seems to do the same thing in one line of code:
```el
(defun ap/lisp-eval-prev-to-comment ()
  "Eval previous form and insert as comment."
  (interactive)
  (insert "  ;;=> " (format "%S" (eval (elisp--preceding-sexp)))))
```

purcell:
> This function from my config seems to do the same thing in one line of code:

Yes, and if you want to preserve the useful underlying prefix behaviour of `eval-print-last-sexp` then you could do this:
```el
(defvar eval-print-as-comment-prefix ";;=> ")

(defun eval-print-as-comment (&optional arg)
  (interactive "P")
  (let ((start (point)))
    (eval-print-last-sexp arg)
    (save-excursion
      (goto-char start)
      (save-match-data
        (re-search-forward "[[:space:]\n]*" nil t)
        (insert eval-print-as-comment-prefix)))))
```
As it is, this package looks extremely over-engineered.

conao3:
@alphapapa 
For some functions, output by `princ` may be the main purpose.
For example, `pp` outputs a well-formed S-expression by `princ`. How about such output?

@purcell 
`eval-print-last-sexp` seemed to be doing a lot more work than I might think. See source.
My package respects the behavior of the original function and captures the `princ` output and the return value of the S-expression, while executing it in the same way.
Then insert it by performing a rectangular edit to show it as a comment.
It's not over-engineering, this is necessary work. I think.

alphapapa:
> For some functions, output by `princ` may be the main purpose.
> For example, `pp` outputs a well-formed S-expression by `princ`. How about such output?

What exactly do you want?  Something like this?

```el
(defun pp-insert-as-comment ()
  (interactive)
  (let ((current-prefix-arg '(4))
        end)
    (save-excursion
      (call-interactively #'pp-eval-last-sexp)
      (setq end (point)))
    (comment-region (point) end)
    (insert "\n")))
```
Capturing `standard-output` is not necessarily the same as the result an expression evaluates to.

riscy:
Hi @conao3 -- just checking in on your response to the feedback on this one.  One additional point to consider is who might actually use this?

I imagine the target audience would be elisp hackers.  But for many elisp hackers -- while this package would be interesting to look at -- the functionality is easy and almost "natural" for for an elisp hacker to re-engineer (as @alphapapa showed).  I myself have a few small elisp packages that I'm still holding off on indexing to MELPA because I'm waiting to get a better feel for their usefulness over time.

conao3:
As for @alphapapa's recent comments, they have a different purpose than mine.
`pp-insert-as-comment` tries to insert the evaluation value as a comment on the `pp`, but my package doesn't touch the return value.
Simply capture the return value and the output in `princ` and insert the string for the comment with a rectangle edit.
Again, simply `eval` is not the same behavior as the original `eval-print-last-sexp`.
The original `eval-print-last-sexp` did more work than we thought it would, and my package respects that behavior.

We all have our own ideas about what level of Elisp hackers to assume, but I would recommend using this package even for beginners who are just starting out with Elisp today.
There's a lot of discussion about what granularity to create a package at. (Perhaps humanity can't get to the answer). However, since this package has a clear purpose and requires a certain amount of code (about 100 lines) to handle the purpose well, I think it can be established as one package.

riscy:
I think you make a fair point.  I think part of the confusion comes from our collective inability to see: "why would you use this?" and part of that might be that this fairly young repository has a very terse README.  The commentary itself says

> Insert Elisp result as comment in scratch buffer.

Maybe the conversation we've had here will inspire adding some more to your README and/or package commentary so that the average Emacser can see this package and say "ah-ha, this is what I'm looking for" or similar.

I'll hold off on merging (from my side) for a couple days to let the other reviewers mull it over.  But in the meantime, I have this --

```
1 issue found:
85:9: error: You should depend on (emacs "26.1") if you need `eval-expression-get-print-arguments'.
```
Since you're hosting it on GitHub, I recently added a way to [add melpazoid to your GitHub actions](https://github.com/riscy/melpazoid#add-it-to-github-actions) which'll catch this.  (melpazoid only fails on honest-to-goodness errors.)

riscy:
@conao3 I think that last package-lint is legitimate -- feel free to ping when ready!

conao3:
I refined the package description and make the package require Emacs-26.1. Thanks!

conao3:
Almost 2 weeks passed. ping?

riscy:
Okay, I will go ahead and say we're good on this one. :)

PR #6851, opened 2020-04-19
Add friendly-shell, friendly-shell-command, friendly-remote-shell

p3r7:
Cleaner attempt of https://github.com/melpa/melpa/pull/6834.

More user-friendly shell APIs for Emacs.

3 packages:
 - `friendly-shell-command.el` for single shell commands ([blog post](https://www.eigenbahn.com/2020/01/19/painless-emacs-shell-commands))
 - `friendly-shell.el` for interactive shells ([blog post](https://www.eigenbahn.com/2020/01/21/painless-emacs-interactive-shells))
 - `friendly-remote-shell.el` for interactive remote shells

https://github.com/p3r7/friendly-shell

Maintainer.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/friendly-remote-shell
+++ b/recipes/friendly-remote-shell
@@ -1,0 +1,4 @@
+(friendly-remote-shell :repo "p3r7/friendly-remote-shell"
+                       :fetcher github
+                       :files (:defaults
+                               (:exclude "friendly-shell-command.el" "friendly-shell.el")))
--- a/recipes/friendly-remote-shell
+++ b/recipes/friendly-remote-shell
@@ -1,4 +1,4 @@
-(friendly-remote-shell :repo "p3r7/friendly-remote-shell"
+(friendly-remote-shell :repo "p3r7/friendly-shell"
                        :fetcher github
                        :files (:defaults
                                (:exclude "friendly-shell-command.el" "friendly-shell.el")))
--- a/recipes/friendly-shell-command
+++ b/recipes/friendly-shell-command
@@ -1,4 +1,4 @@
-(friendly-shell-command :repo "p3r7/friendly-shell-command"
+(friendly-shell-command :repo "p3r7/friendly-shell"
                         :fetcher github
                         :files (:defaults
                                 (:exclude "friendly-shell.el" "friendly-remote-shell.el")))
--- a/recipes/friendly-shell-command
+++ b/recipes/friendly-shell-command
@@ -1,0 +1,4 @@
+(friendly-shell-command :repo "p3r7/friendly-shell-command"
+                        :fetcher github
+                        :files (:defaults
+                                (:exclude "friendly-shell.el" "friendly-remote-shell.el")))
--- a/recipes/friendly-shell
+++ b/recipes/friendly-shell
@@ -1,0 +1,4 @@
+(friendly-shell :repo "p3r7/friendly-shell"
+                :fetcher github
+                :files (:defaults
+                        (:exclude "friendly-shell-command.el" "friendly-remote-shell.el")))


recipes/friendly-remote-shell
```diff
@@ -0,0 +1,4 @@
+(friendly-remote-shell :repo "p3r7/friendly-remote-shell"
```
riscy, line 1: This repository doesn't seem to exist at the moment. :)

recipes/friendly-shell-command
```diff
@@ -0,0 +1,4 @@
+(friendly-shell-command :repo "p3r7/friendly-shell-command"
```
riscy, line 1: This repository doesn't exist too.

p3r7:
> Sorry it's taken a while to get around to this second go round.

No problem. I understand the amount of work this represent :wink: 

> Usually when there are multiple recipes, it's much preferred to open three different PR's given three different recipes.

Oh my bad, I made a mistake : the 3 packages are in fact hosted on the same repo (as they are very closely related) which exists. I guess I blindly did a search / replace all :facepalm: 

I've fixed the repo reference in the recipes with https://github.com/melpa/melpa/pull/6851/commits/5e8f9070aa858b7ca121844e934b653606abfdbd.

If originally (attempted at least) to follow this guideline from [CONTRIBUTING.org](https://github.com/melpa/melpa/blob/master/CONTRIBUTING.org) : `If the packages are very closely related and maintained in the same repository (see below for why it might not be a good idea), you might open only one pull request for them all; in this case, keep adding one recipe per commit.`.

How does that look now?

recipes/friendly-remote-shell
p3r7: See my comment bellow.

recipes/friendly-shell-command
p3r7: See my comment bellow.

riscy:
Ah you're right about CONTRIBUTING.org :)  I might see if there's a possibility to amend that, because if the packages depend on one-another then it's harder to test if their dependencies are accessible from MELPA (an important check).  Also the yes/no decision whether to merge becomes easier when it's just one package.

That's not a blocker, but fyi.  In the meantime:

```
friendly-shell-command.el:24:1:Error: Cannot open load file: No such file or directory, dash
```
I think a `Package-Requires` is needed on dash in this case.

Minor stuff I see:
- proccess -> process
- necessayr -> necessary
- according -> according
- DIPLAY -> DISPLAY

p3r7:
- [x] Added missing dash dependency in the `Package-Requires` header.
 - [x] Fixed typos (I'm again sorry for those :facepalm:) 
 - [x] Completed documentation with recently added features

> if the packages depend on one-another then it's harder to test if their dependencies are accessible from MELPA (an important check)

So should I split this PR into several? Just tell me.

riscy:
No, it's fine. :)

riscy:
Okay, these look good to me!

The one (non-blocking) suggestion I might be to remove the underline in your commentary (or at least be aware it'll look like this (note the extra space) in `finder-commentary`:
```elisp
Commentary:
 -----------

For detailed instructions, please look at the README.md
```
and, worse, I'm not sure whether that underline will appear in the [MELPA index's "description"](  That will be helpful to people looking at the "description" part of the MELPA index ([example](https://melpa.org/#/bind-key)) section for the packages.  I'd also recommend giving a link to that README for ease of use (so somebody can just use find-file-at-point to open the link) or -- even better -- offer a few quick details about the package too.

I think that's not worth blocking on, though.  These look cool.

p3r7:
Indeed, I've added a more comprehensive package header doc, which was lacking.

Thanks for the tip about the underline.
Indeed it wasn't working so nice. I guess I originally took inspiration from another package's source but did not investigate how they would work with other packages.
I've removed them from all my other packages.

Thanks again for all your feedback!

PR #6852, opened 2020-04-19
Add Setup

phikal:

This package provides the `setup` macro that helps in configuring Emacs, and the `setup-define` that helps in configuring-configuring Emacs. The former is comparable in idea to `use-package`, but places more emphasis on user-extensibility via the latter.

https://git.sr.ht/~zge/setup.el

I am the maintainer.

None needed.

### Checklist

Please confirm with `x`:

**Note:** There are two `package-lint` complaints I cannot fix (`with-eval-after-load` usage), due to the nature of the package. I should note that both are in macros that only get used in real code after macro-expansion, not by the package itself.

### DIFF
--- a/recipes/setup
+++ b/recipes/setup
@@ -1,0 +1,1 @@
+(setup :url "https://git.sr.ht/~zge/setup.el" :fetcher git)


conao3:
It may sound strange for me saying, but MELPA already has [leaf](https://github.com/conao3/leaf.el) as a `use-package` alternative.
Are there any comparisons and advantages from two?

phikal:
Well since when is having two alternatives Emacs an argument against having a third?

I guess the obvious one, is that  `setup` isn't trying to be a use-package re-implementation (not in a bad way), but just a macro that simplifies common configuration practices. 

Leaf says 

> leaf has keywords almost identical to use-package, but some of usage of the keywords is different.

while `setup` has no interest in preserving syntactical similarities, and avoids the "complexity" of parsing variable-length plists by using alists, that also simplify indentation.

As I state in the README:

> It should be noted that this package doesn't intend to be an exhaustive collection of keywords, but rather encourage the user to create their own.

So for example, by default `setup` doesn't have a `:init` and `:config` analogue. If a user really wants `:init` or `:config`, it's trivial to implement:

```lisp
(setup-define config ((args ,body) (body after))
  body)
```

Yet I won't add it to the packages, as 90% of the cases I have seen `:config` being used, is to set variables, when `:custom` should have (imo) been preferred.

Also, `setup` is small. The two main macros, `setup` and `setup-define` make up around 100 lines of (quasiquote-extravaganza) elisp. Since it's public domain, users might even decide to copy the file without using `package.el`. Leaf has a lot more, but from a quick glance, I couldn't quite tell what part could be compared to my two macros. 

Just like in your case, 

> leaf solves the stress that I feel while using the use-package for 2.5 years. By developing from scratch, we have a cleaner and more predictable implementation than use-package.

I also had my annoyances with `use-package`, and implemented something that simplifies my kind of configuration-approach. But beyond that, it should (as a long-term goal) serve as a framework for anyone to _setup_ packages, in whatever way they like, without having to force their approach into the vocabulary of some third maintainer.

---

Btw, do you know of more `use-package` clones or approaches? I would like to list some "alternatives"  in my

conao3:
> Well since when is having two alternatives Emacs an argument against having a third?

No. However, MELPA requires a reasonable and innovative package in order to register recipes.

https://github.com/melpa/melpa/blob/master/CONTRIBUTING.org#what-to-know-before-opening-a-pull-request
> Reasonably innovative package
MELPA provides a curated set of Emacs Lisp packages, not an exhaustive list of every single Emacs Lisp file ever created. By default, MELPA maintainers will reject packages that duplicate functionality provided by existing packages. Please try to improve existing packages instead of creating new ones when possible.

What I feel your package does is simply pass arguments to macros in alist, and is only a small difference in superficial DSL.
If you want to use `use-package` via `alist`, I think it makes more sense to have a little macro like this in your init.el rather than creating a package called `setup`.
```emacs-lisp
(defmacro setup (name &rest args)
  (declare (indent 1))
  (let ((keys (mapcar #'car args)))
    `(use-package name
       ,@(mapcan
          (lambda (key)
            `(,(intern (format ":%s" key)) ,@(alist-get key args)))
          keys))))
```
And the `use-package` keyword is extensible, and so is the `leaf`.
So is your packaging reasonably innovative?

Also, regarding `:config`, I believe it is used to enable global minor mode, for example.
How would we configure this in `setup`?

```emacs-lisp
(use-package autorevert
  :config (global-auto-revert-mode +1))
```

phikal:
There is no doubt that my package isn't a radical innovation, but as I said before, the noteworthy difference is the user-facing `setup-define`, not the syntax or choice of default keywords. Size is a plus. I am biased, but it seems to me that `use-package` and Leaf have have a different design philosophy than `setup`, and I would regard that as a significant enough difference.

> Also, regarding :config, I believe it is used to enable global minor mode, for example.
How would we configure this in setup?

Easy, a minor mode defined using `define-minor-mode` is also a user option, and can as such be enabled using `customize-set-variable`, which is exposed in the default `setup` keyword list.

riscy:
Thanks both.

> use-package and Leaf have have a different design philosophy than setup, and I would regard that as a significant enough difference.

I think this is okay, although this --

> I would like to list some "alternatives"

-- is a great idea, and obviously leaf should be among them.  This attic package has a similar name https://github.com/emacsattic/emacs-setup and has an emacs-setup-require feature that might be in a similar niche.

riscy:
Actually, before I hit merge, there is also this on the Emacs mirror: https://github.com/emacsmirror/emacswiki.org/blob/master/setup.el (I was surprised the name hadn't been taken _somewhere_).

@tarsius may have some advice in this situation.

tarsius:
I would advice against the name `setup`.

> There is no doubt that my package isn't a radical innovation, but as I said before, the noteworthy difference is the user-facing `setup-define`, not the syntax or choice of default keywords. Size is a plus. I am biased, but it seems to me that `use-package` and Leaf have have a different design philosophy than `setup`, and I would regard that as a significant enough difference.

Maybe. But you also just created this five days ago and it might turn out to be a dead end. You should use and improve this for a few months before adding it to Melpa.

phikal:
> I was surprised the name hadn't been taken somewhere

That's actually why I chose the name. How about `configure`? `prepare`? I don't see these being used by either emacswiki or MELPA.

> But you also just created this five days ago and it might turn out to be a dead end.

@tarsius, could you elaborate on this?

tarsius:
> But you also just created this five days ago and it might turn out to be a dead end.

@tarsius, could you elaborate on this?

It appears that you have implemented some dsl and expect that to stay about as simple as it is now. But I don't think that `use-package` is all accidental complexity. Eventually your implementation will grow much more complex too and then you might decide that it isn't worth it to you and that you would rather use `use-package`, which while not perfect, you at least don't have to maintain yourself.

Or not. But I have seen many packages like this and almost all of them have been abandoned by now. It's at least possible that the same will happen with yours. I suggest you keep using this for a few months to see whether this package can at least take that first hurdle.

phikal:
I get your point, and if that's the consensus I'll bump this issue in a few months.

My hope was to at least get the package published on regular MELPA to gather input from users.

tarsius:
> My hope was to at least get the package published on regular MELPA to gather input from users.

While feedback and suggestions can be helpful, it can also be beneficial to *not* get any of that for a while. Gives you an opportunity to refine what *you* want, without having to compromise or fending of distracting suggestions. (You might still attract some early users by announcing this somewhere. Those that are willing to install manually might give better feedback anyway.)

a13:
> Also, regarding `:config`, I believe it is used to enable global minor mode, for example.
> How would we configure this in `setup`?
> 
> 
> ```emacs-lisp
> (use-package autorevert
>   :config (global-auto-revert-mode +1))
> ```

minor-modes have corresponding defcustoms, you know?

PR #6854, opened 2020-04-21
Remove @proofit404's disappeared packages

adisbladis:
These modes are gone from the authors github profile.
Discovered through the Nix emacs-overlay hydra: https://hydra.nix-community.org/eval/4237.

None. Hoping the author @proofit404 will chime in here.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/company-edbi
+++ b/recipes/company-edbi
@@ -1,3 +1,0 @@
-(company-edbi
- :fetcher github
- :repo "proofit404/company-edbi")
--- a/recipes/company-tern
+++ b/recipes/company-tern
@@ -1,1 +1,0 @@
-(company-tern :fetcher github :repo "proofit404/company-tern")
--- a/recipes/edbi-database-url
+++ b/recipes/edbi-database-url
@@ -1,3 +1,0 @@
-(edbi-database-url
- :fetcher github
- :repo "proofit404/edbi-database-url")
--- a/recipes/edbi-django
+++ b/recipes/edbi-django
@@ -1,1 +1,0 @@
-(edbi-django :fetcher github :repo "proofit404/edbi-django")
--- a/recipes/edbi-minor-mode
+++ b/recipes/edbi-minor-mode
@@ -1,1 +1,0 @@
-(edbi-minor-mode :fetcher github :repo "proofit404/edbi-minor-mode")
--- a/recipes/edbi-sqlite
+++ b/recipes/edbi-sqlite
@@ -1,4 +1,0 @@
-(edbi-sqlite
- :fetcher github
- :repo "proofit404/edbi-sqlite"
- :old-names (edbi-sqlite3))
--- a/recipes/envdir
+++ b/recipes/envdir
@@ -1,1 +1,0 @@
-(envdir :fetcher github :repo "proofit404/envdir-mode")


riscy:
Thanks for this, I confirmed all of the recipes in this point to 404s.

There's a slight chance this may have been a mistake, since the author still has [one MELPA-indexed package](https://github.com/proofit404/relative-buffers) up.

riscy:
Since these are gone there's no point in indexing them -- at least for now.  Easy revert if this turned out to be a mistake.  Thanks again @adisbladis

tarsius:
`widgetjs` still depends on `company-tern`.

@DamienCassou it appears you are maintaining `widgetjs` now; would you be willing to maintain `company-tern` as well? Otherwise you should probably remove the dependency.

DamienCassou:
@tarsius: I can see that `epkg-describe-package` gives `company-tern` as a dependency for `widgetjs`. But melpa doesn't: https://melpa.org/#/widgetjs. The melpa's recipe is:

```emacs
(widgetjs :fetcher github :repo "foretagsplatsen/emacs-js" :files ("widgetjs/widgetjs.el"))
```

The [widgetjs/widgetjs.el](https://github.com/foretagsplatsen/emacs-js/blob/79098cd4cdfa144807cb710af4589c13a7033842/widgetjs/widgetjs.el) doesn't reference `company-tern` either.

On the contrary, `emacs-js.el` at the root of the repository does reference `company-tern`. But this one is not on melpa.

tarsius:
> On the contrary, `emacs-js.el` at the root of the repository does reference `company-tern`. But this one is not on melpa.

Okay, so it doesn't affect Melpa, but as the maintainer of the Emacsmirror I would still like to know if you plan to do anything about this. :stuck_out_tongue:

DamienCassou:
I don't use this package and don't plan to maintain it.

tarsius:
> emacs-js
> 
> Emacs configuration files for editing JavaScript
> 
> Description
> 
> This package aims to simplify the installation and setup of a good development environment for editing JavaScript files.
> 
> Some of the settings might specific to how we do JavaScript at Företagsplatsen, but most of it should be generic enough to be reused.
> 
> ...
>
> This package also provides a helper mode for editing JavaScript projects that use WidgetJS.

Please consider maintaining `widgetjs` in a separate repository. Whether you do that does not matter for Melpa, but for people who install packages by cloning the respective git repository it does. Such users might want to get `widgetjs` without it being wrapped in a configuration tailored for Företagsplatsen (especially because that configuration is not symbol-prefix clean and all, and now depends on a package that is no longer maintained).

DamienCassou:
> Please consider maintaining widgetjs in a separate repository.

I know the reasons for doing that and, being a borg user, I agree. But this time, I don't think it matters at all. This package is only useful for users of the widgetjs JS framework that is made by Företagsplatsen and probably only used by us. The JS framework is not even on npmjs.

tarsius:
If `widgetjs` is only used by one company, then *maybe* it should not be on Melpa. I wonder what @purcell thinks about that.

FYI as long as this on Melpa, I will also keep it on the Emacsmirror, which now extracts limits the history to the `widgetjs/` directory. I am fine with that. :smiley:

DamienCassou:
> If `widgetjs` is only used by one company, then *maybe* it should not
> be on Melpa. I wonder what @purcell thinks about that.


This would make installation a bit harder for our new employees that we
try to convert to Emacs :-). But I understand that fully and will make a
removal PR if you want me to.


> FYI as long as this on Melpa, I will also keep it on the Emacsmirror,
> which now extracts limits the history to the `widgetjs/` directory. I
> am fine with that. :smiley:


thank you Jonas.

-- 
Damien Cassou

abend:
I can volunteer to maintain `company-tern` at least.

tarsius:
> There's a slight chance this may have been a mistake, since the author still has [one MELPA-indexed package](https://github.com/proofit404/relative-buffers) up.

That's gone too now, see #6982.

PR #6855, opened 2020-04-21
info-buffer: Move repo to gitlab

adisbladis:
~The repository is gone.~
Upstream moved to gitlab

cc @llvilanova 

### Checklist

Please confirm with `x`:


llvilanova:
The repository has moved to https://gitlab.com/llvilanova/info-buffer

adisbladis:
@llvilanova Thanks for the reply! I've updated the PR to move to the new upstream instead of removing the package.

PR #6856, opened 2020-04-21
Remove syndicate

adisbladis:
The authors entire Github profile has been removed.

None.

### Checklist

Please confirm with `x`:


riscy:
Thanks again! :)

PR #6861, opened 2020-04-22
Add recipe for message-attachment-reminder

alexmurray:

Reminders user of possible forgotten attachments in message-mode

https://github.com/alexmurray/message-attachment-reminder

Maintainer

N/A

### Checklist

### DIFF
--- a/recipes/message-attachment-reminder
+++ b/recipes/message-attachment-reminder
@@ -1,0 +1,1 @@
+(message-attachment-reminder :fetcher github :repo "alexmurray/message-attachment-reminder")


innerout:
Your link to the package repository results to 404 error.

alexmurray:
Gah apologies for the typo - should have been https://github.com/alexmurray/message-attachment-mode (I have edited this above)

riscy:
Nice.  I remember seeing people using the word "enclosed" as well, which you may want to add to your `message-attachment-reminder-regexp`.  I'd also recommend downcasing these and adding a let form to set case-fold-search (case sensitivity) to nil, e.g. `(let ((case-fold-search nil)) (re-search-forward ... ))` -- together those changes will ensure that you can find the upper/lowercase variations of those phrases.

The message `This message contains '%s' but appears to be missing an attachment, are you sure you want to send it?` is over 100 characters long, but I think it's safer to assume some of your users might be working in slightly narrower frames.  Could the message be made more terse, e.g. `This message contains '%s' but has no attachment; send it?` or even `Message appears to be missing an attachment.  Send anyway?`

One thing we've been checking for more carefully is a LICENSE (or COPYING) file in the repo that is machine-detectable (see: https://github.com/licensee/licensee).  This lets people use the API to get package licenses.

alexmurray:
Thanks for the feedback @riscy - FYI - I think you meant `case-fold-search` should be let bound to `t` to match case-insensitively. Please see https://github.com/alexmurray/message-attachment-reminder/commit/02d7e66e55729d294f92881530f58a6a7364a78c which should address your comments.

riscy:
You're right! Thanks. :)

PR #6866, opened 2020-04-24
Add neuron-mode recipe

felko:

Major mode for editing notes using the [neuron](https://neuron.zettel.page/) zettelkasten manager.

https://github.com/felko/neuron-mode

Author

None needed

### Checklist

### DIFF
--- a/recipes/neuron-mode
+++ b/recipes/neuron-mode
@@ -1,0 +1,1 @@
+(neuron-mode :fetcher github :repo "felko/neuron-mode")


riscy:
Checklist looks good on a very quick first pass.  I only see --

```
neuron-mode.el:148:1:Warning: Unused lexical argument `show-tags'
```

which looks like it should be used in the function body, based on the docstring.

felko:
Thanks, that should be fixed now.

riscy:
The latest --

```
neuron-mode.el:408:1:Warning: the function `neuron-query-tag' is not known to
    be defined.
```
(called in a couple places)

felko:
Fixed 👍 

`byte-compile-file` seems to raise warnings when a function is not in the environment, but actually doesn't care about whether it's in the current file or not. I evaluate the buffer during the development of the package (and `neuron-query-tag` existed at one point) so the warning didn't trigger. Took me a while to understand.

riscy:
Yeah, this is one of the biggest gotchas when developing Emacs packages.  `flycheck` for emacs lisp mode can give you great feedback that can prevent these sorts of things.

In any case this looks good to me.  I also took a peek at your webpage and like what you've done with the place re: neuron. :)  And, welcome to MELPA!

felko:
Thank you, I am not the author of neuron itself, nor the author of [neuron.zettel.page](https://neuron.zettel.page) though

PR #6867, opened 2020-04-25
Add go-noisegate recipe

ks888:

The client for Noise Gate (the golang test runner).

https://github.com/go-noisegate/go-noisegate.el

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/go-noisegate
+++ b/recipes/go-noisegate
@@ -1,0 +1,1 @@
+(go-noisegate :repo "go-noisegate/go-noisegate.el" :fetcher github)


riscy:
Hi and thanks for this. :)

Here's some quick feedback --

### go-noisegate.el ###

```
go-noisegate.el:58:40:Warning: reference to free variable
    `go-noisegate-changes'
In go-noisegate--record-change:
go-noisegate.el:65:29:Warning: reference to free variable
    `go-noisegate-changes'
go-noisegate.el:70:56:Warning: assignment to free variable
    `go-noisegate-changes'
In go-noisegate--reset-changes:
go-noisegate.el:74:9:Warning: assignment to free variable
    `go-noisegate-changes'
```
I believe you only need to `(defvar go-noisegate-changes)` at the top of your file (although a docstring would be nice to add too).

- No issues!

- No issues!

- go-noisegate.el#L58: It's safer to sharp-quote function names; use `#'`
- go-noisegate.el#L25: There's no need for this -- define-derived-mode will set that for you (as well as the mode-name).  Regarding the `(setq-local truncate-lines t)` that follows -- is this necessary, or just your preference?  If the latter, sometimes it's better to just give users the option to add hooks if they want to set such variables.

One last issue I see is that:
```elisp
(defvar go-noisegate-gate-cmd (go-noisegate--find-gate-command)
  "The path to the gate command.")
```
requires the `go` binary to be installed to succeed, otherwise the package can't be loaded without erroring.  This is fine for now, but if someone creates a package that has go-noisegate as a dependency, it won't be possible to compile _their_ package unless go is installed on the testing machine.  If this can be avoided, it is definitely worth it, because it makes packages easier to test.

### Packaging ###

-Your  `;;; Commentary` is empty which will unfortunately lead to a "blank" entry on MELPA.  The Commentary is used to fill in the Description field on MELPA, for example see [here](https://melpa.org/#/company-jedi).
- I see GitHub API found `Apache License 2.0` but please add license boilerplate above your `;;; Commentary` or an [SPDX-License-Identifier](https://spdx.org/using-spdx-license-identifier) to go-noisegate.el

ks888:
Thank you for your feedback! I believe I fixed all the issues.
* go-noisegate.el: https://github.com/go-noisegate/go-noisegate.el/commit/70b5fd8541c7e35a5f15b6d2dcdf1907c47e2c92
* Packaging: https://github.com/go-noisegate/go-noisegate.el/commit/a2cec1ca1c91ed8d0a8961aadcf0835783aeb4f9

riscy:
There's one small byte-compile nag:

```
go-noisegate.el:64:1:Warning: Lexical argument shadows the dynamic variable
    mode-name
```
Since the variable isn't used, maybe it would be more appropriate to call it `_mode-name` per convention.  (You could also rename the variable `length` in `go-noisegate--record-change` to `_length`, similarly, and then you don't even have to say "LENGTH is not used" to pacify checkdoc -- it's just assumed to be not used by convention.)

Taking a deeper look --

Linewrapping your `;;; Commentary` is probably a good idea, because the `finder-commentary` for the package looks like this for me (and similarly may not display well in the MELPA index):
image

And one last note I'll make is that the Summary line ` --- Emacs plugin for Noise Gate` could be tweaked for the better.  First, obviously it's for Emacs, so you could drop that word. :)  Typically they're called "packages" (or features?) instead of "plugins" as well, but I think saying that it's a package isn't necessary either.  What about something like: "Run Golang tests with Noise Gate"?

Those are the only major issues I'm seeing at the moment. :)

ks888:
Thank you again for the detailed feedback. It's great to know these conventions!
* Fixed variable name: https://github.com/go-noisegate/go-noisegate.el/commit/825d1fb05ec329f938c4c5bed23592f54d326f80
* Fixed summary and commentary: https://github.com/go-noisegate/go-noisegate.el/commit/825d1fb05ec329f938c4c5bed23592f54d326f80

riscy:
Great, thanks!

PR #6869, opened 2020-04-26
Add package inform

dieter-wilhelm:
Package inform: Symbol links in Info buffers to their help documentation

This library provides links of symbols (functions, variables,
faces) within Emacs' Info viewer to their help documentation.  This
linking is done, when the symbol names in texinfo documentations
(like the Emacs and Elisp manual) are `quoted', which is done
customarily.  And the quoted names must be known to Emacs i.e. the
definitions are loaded in its memory.

You can follow these additional links with the usual Info
keybindings.  The customisation variable
`mouse-1-click-follows-link' is influencing the clicking behaviour
(and the tooltips) of the links, the variable's default is 450
(milli seconds) setting it to nil means only clicking with mouse-2
is following the link .

https://github.com/dieter-wilhelm/inform

I'm the maintainer

**None needed**

### Checklist

### DIFF
--- a/recipes/inform
+++ b/recipes/inform
@@ -1,0 +1,3 @@
+(inform
+:fetcher github
+:repo "dieter-wilhelm/inform")
--- a/recipes/treemacs
+++ b/recipes/treemacs
@@ -2,6 +2,7 @@
  :fetcher github
  :repo "Alexander-Miller/treemacs"
  :files (:defaults
+         "Changelog.org"
          "icons"
          "src/elisp/treemacs*.el"
          "src/scripts/treemacs*.py"


riscy:
This appears to not have any new recipe committed -- maybe a missing push?

dieter-wilhelm:
> This appears to not have any new recipe committed -- maybe a missing push?

Pushed now the recipe, thanks for reminding me!

riscy:
Neat, I like it --

- inform-pkg.el -- these -pkg.el files are (very very mildly) discouraged since, for example, they have poor compatibility with package-lint.  Also, you list a `Package-Requires` both here and in inform.el, so you'll just have to be careful to keep these in sync.
- I know you're aware that there is a large amount of commented-out code in the function `inform-make-xrefs` -- you might consider commenting out (or adding to a development branch) to make the file a little more habitable for users
- `with-syntax-table` doesn't need a `progn` in it, I believe.
- A thought on rewording your summary: `Symbol links in Info buffers to their help documentation` -> `Link symbols in Info buffers to their help documentation`

None of that feedback is quite a blocker, so this is good to merge!

dieter-wilhelm:
> Neat, I like it --

Thanks, I've got also positive feedback from the emacs-devel mailing list.  But only a few took the pains to comment at all so I think it is a good idea to mellow this package on Melpa.

>     * inform-pkg.el -- these -pkg.el files are (very very mildly) discouraged since, for example, they have poor compatibility with package-lint.  Also, you list a `Package-Requires` both here and in inform.el, so you'll just have to be careful to keep these in sync.

Maybe I confused something when I first read the Melpa Readme and Emacs' package documentation.  Arey you telling me that  I can completely remove -pkg.el files for Melpa?

>     * I know you're aware that there is a large amount of commented-out code in the function `inform-make-xrefs` -- you might consider [deleting this] (or adding to a development branch) to make the file a little more habitable for users

done

>     * `with-syntax-table` doesn't need a `progn` in it, I believe.

done

>     * A thought on rewording your summary: `Symbol links in Info buffers to their help documentation` -> `Link symbols in Info buffers to their help documentation`

done
 
> None of that feedback is quite a blocker, so this is good to merge!

Thank you, thank you, thank you very much! :-)

riscy:
> Maybe I confused something when I first read the Melpa Readme and Emacs' package documentation. Arey you telling me that I can completely remove -pkg.el files for Melpa?

Yep -- everything can be specified in the main file, and package-lint will tell you if anything is missing.

PR #6871, opened 2020-04-29
Add recipe for vega-view

jackrusher:

This is a package for viewing Vega visualizations in emacs. There's an extensive README at the repo in case this doesn't immediately make sense.

https://github.com/applied-science/emacs-vega-view

Author/maintainer.

The Makefile fails on my recipe, but it also seems to fail on every other recipe that `require`s `cl-lib`, despite said packages already being in MELPA. Everything else is squeaky clean.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/vega-view
+++ b/recipes/vega-view
@@ -1,0 +1,1 @@
+(vega-view :repo "applied-science/emacs-vega-view" :fetcher github)


riscy:
Hi!  This looks interesting.  Here's a quick first pass --

### vega-view.el ###

```
vega-view.el:98:11:Warning: assignment to free variable
    `cider-popup-output-marker'
In end of data:
vega-view.el:166:1:Warning: the following functions are not known to be
    defined: json-encode, cider-interactive-eval, nrepl-make-response-handler,
    cider-emit-interactive-eval-output,
    cider-emit-interactive-eval-err-output, cider-emit-into-popup-buffer,
    cider-sexp-at-point
```
I think a `Package-Requires` and a `require` need to be set for the above.  Also a `(require 'json)`, which is part of Emacs already.

- No issues!

- No issues!

- vega-view.el#L6: Prefer `https` over `http` (if possible)

### Packaging ###






- Use a LICENSE file that GitHub can detect (e.g. no markup) if possible
- Please add a LICENSE or COPYING file to the repository

jackrusher:
Hey! Thanks for the feedback and patience. I've used emacs since 1985, but this is my first time publishing to MELPA. :) Hopefully all the issues are now resolved.

riscy:
> I've used emacs since 1985

Amazing that we're still hacking away on a 44 year old piece of software!

Package-lint is having trouble with this one currently --

```
9:0: error: Expected (package-name "version-num"), but found (cider 0\.24\.0).
```

If you just change that to `;; Package-Requires: ((emacs "25") (cider "0.24.0") (parseedn "0.1"))` that should get it fixed.

As I take a look at the source, I'll also mention it's a good idea to wrap your long docstring lines (at something like 80 characters, but anything within reason).  The reason for doing this is since when you create these docstrings, you're also creating help files for your users, and they can end up displaying like this:

image

As the author of the docstrings, only _you_ know how the text should be wrapped (Emacs can't guess, since sometimes strings should be long, e.g. in the case of URLs).

And a very minor typo I spotted:
- spcifications -> specifications

jackrusher:
Amusingly, I only unwrapped those doc strings because one of the automatic checkers had complained about they were wrapped (using `paredit-reindent-defun` which re-fills the paragraph in the case of docstrings, which sometimes indents the filled text).

Anyway, current check-in should hopefully resolve these issues. :)

riscy:
Yes, all looks good to merge.  You will see a couple checkdoc nags (`First line is not a complete sentence`) but these aren't worth blocking on in my opinion.  I believe that nag is most important in the case of interactive commands, for which <kbd>M-x apropos</kbd> will only render the first line of the docstring.

If you want to keep an eye on these, I have some [simple instructions up](https://github.com/riscy/melpazoid#add-it-to-github-actions) for running MELPA checks on each push (it only "fails" when there is a bona fide error).

And, thanks!  Welcome to MELPA. :)

PR #6873, opened 2020-04-30
add multistate package

seregaxvm:

This package provides `multistate-mode` minor mode. This mode allows modal editing and may be described as `evil-mode` without vi. It does not predefine any states, key bindings, does not use hooks, so it's very useful when user wants to configure modal editing but does not Emacs to emulate Vim (or spend time removing Vim out of `evil-mode`).

<https://gitlab.com/matsievskiysv/multistate>

I'm an author.

**None needed**

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/multistate
+++ b/recipes/multistate
@@ -1,0 +1,1 @@
+(multistate :repo "matsievskiysv/multistate" :fetcher gitlab)


mattiasb:
How does this compare to [control-mode](https://github.com/stephendavidmarsh/control-mode)?

seregaxvm:
I don't know much about `control-mode`, but from its manual and [`modalka-mode`](https://github.com/mrkkrp/modalka) manual I understood that it has only one normal mode (or state).
It generates keymaps dynamically by removing <key>Ctrl</key> from Emacs global bindings.
In this sense `multistate` mode is closer to `modalka` mode (which has static keybindings) but with multiple modes.

mattiasb:
Thanks! :)

seregaxvm:
Changed mode so that all states are buffer-local.

riscy:
Another package to compare to is god-mode https://github.com/emacsorphanage/god-mode which is in the orphanage but still seems to get an update now and then.

A quick pass over the checklist gives --
```
Warning (emacs): 
multistate.el:50: Name emacs should appear capitalized as Emacs
Warning (emacs): 
multistate.el:266: Argument `write' should appear (as WRITE) in the doc string
```
where the first is a false positive, and the second looks like an easy fix.

Minor:
- You use a variety of unicode glyphs for your lighter(s).  That's not a blocker but you should be aware that they will render poorly (or won't render at all) on older terminals (and I can't imagine a package like this _not_ being useful on an older terminal) and can sometimes clobber display performance on some versions of macOS due to it cycling through different fonts until it finds one that supports the glyph.
- I notice in a couple places you've written "mutistate" instead of "multistate"

seregaxvm:
Sorry, forgot to do a checkdoc after last commit. Double checked everything again, fixed typos.

> Another package to compare to is god-mode https://github.com/emacsorphanage/god-mode which is in the orphanage but still seems to get an update now and then.

Modalka mode has an extensive [other-solutions](https://github.com/mrkkrp/modalka#other-solutions) section which summarizes similar packages. So, right off the bat I'd remove god mode, boon, fingers and xah fly keys because they all define their versions of ergonomic keybindings. 

This package may be compared to [`evil-core.el`](https://github.com/emacs-evil/evil/blob/master/evil-core.el).
There's been a [demand](https://github.com/emacs-evil/evil/issues/1286) for extracting its functionality from evil package but I got the impression that it's a low priority task to evil maintainers.

riscy:
I see.  Okay, neat.  Everything looks good to go -- thanks for this!

PR #2398, opened 2015-01-12
remove several problematic packages from the wiki

tarsius:
See #2128.

Issue #2128, opened 2014-10-30
Get emacswiki files from the git repo


purcell:
EmacsWiki is [backed by git](http://www.emacswiki.org/emacs/Git_repository), so rather than scraping EmacsWiki.org, we could just grab `:fetcher wiki` files from there.

https://github.com/emacsmirror/emacswiki.org

tarsius:
So I have had a look at the differences. Looks like it might be a while until you can switch to use my repos. Here are the findings:

* Melpa has 27 wiki packages missing from the Emacsmirror.
* Melpa mirrors 27 wiki packages, which the Emacsmirror gets from elsewhere.
* Emacsmirror has 22 wiki packages, which Melpa gets from elsewhere.

In most cases "elsewhere" is probably preferable over the wiki. Here's the list of packages you might want to check first:

```
(("cal-china-x" "git@github.com:xwl/cal-china-x.git")
 ("dummy-h-mode" "git@github.com:yascentur/dummy-h-mode-el.git")
 ("eimp" "git@github.com:nicferrier/eimp.git")
 ("elisp-depend" "git@github.com:tehom/elisp-depend.git")
 ("fic-mode" "git@github.com:lewang/fic-mode.git")
 ("iy-go-to-char" "git@github.com:doitian/iy-go-to-char.git")
 ("legalese" "git@github.com:jorgenschaefer/legalese.git")
 ("pastebin" "git@github.com:nicferrier/elpastebin.git")
 ("scpaste" "git@github.com:technomancy/scpaste.git")
 ("unbound" "git@github.com:haxney/unbound.git")
 ("wc-mode" "git@github.com:bnbeckwith/wc-mode.git"))
```

tarsius:
I went through the list of packages which Melpa mirrors from the
Emacswiki and which are completely missing from the Emacsmirror.

I have reservations about most of these libraries.  When I first
created the Emacsmirror I quickly reviewed every elisp file to be
found on the Emacswiki.  Obviously I could not spend a lot of time
on each and every library, so it is quite possible that I failed to
import some libraries which are perfectly fine.  This time again I
have reviewed the libraries quickly, take the criticism with a grain
of salt, and try not to be offended - maintaining something like
Melpa and Emacsmirror, and what might seem like nit-picking to
outsiders is essential for such collections to be of any value.

# Good change of being added

## to recent

Generally I think we should start refusing to add any packages which
are created now and only distributed on the Emacswiki (to either the
[...]

PR #6874, opened 2020-04-30
Add apt-utils recipe

Sasanidas:

This package can be used to manage apt packages with in emacs, it has a lot of features such as 
get package information, search package with regex and so on.

This package was found in emacswiki.org and slightly modify to work with modern emacs,  there is
still some work to do to optimize the code. 

https://gitlab.com/sasanidas/apt-utils

Maintainer

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/apt-utils
+++ b/recipes/apt-utils
@@ -1,0 +1,1 @@
+(apt-utils :fetcher gitlab :repo "sasanidas/apt-utils")


tarsius:
We used to distribute this package but removed it in #2398, which references #2128, which contains this:

> apt-utils file corrupted on wiki (fixed)
> But more importantly this isn't upstream, the best I could find is http://ftp.de.debian.org/debian/pool/main/a/apt/apt_0.9.7.9+deb7u7.tar.gz but my tools can no longer import tarballs.

But then in #5758 we added the `debian-el` package, which contains `apt-utils.el`. Also see #5449.

This is upstream and therefore what Melpa should distribute. You should probably contribute to that at https://salsa.debian.org/emacsen-team/debian-el.

Sasanidas:
> We used to distribute this package but removed it in #2398, which references #2128, which contains this:
> 
> > apt-utils file corrupted on wiki (fixed)
> > But more importantly this isn't upstream, the best I could find is http://ftp.de.debian.org/debian/pool/main/a/apt/apt_0.9.7.9+deb7u7.tar.gz but my tools can no longer import tarballs.
> 
> But then in #5758 we added the `debian-el` package, which contains `apt-utils.el`. Also see #5449.
> 
> This is upstream and therefore what Melpa should distribute. You should probably contribute to that at https://salsa.debian.org/emacsen-team/debian-el.

Oh I didn't know that, the apt file is more for general apt distributions, that package works with Ubuntu as well?

Thank you.

tarsius:
Seems so: https://packages.ubuntu.com/eoan/debian-el.

Sasanidas:
> Seems so: https://packages.ubuntu.com/eoan/debian-el.

Ok then, no need for this package on Melpa.

Thank you.

PR #6875, opened 2020-04-30
Fix README homebrew link

Sasanidas:
Replace the homebrew github link with the current one.


tarsius:
The old link takes us to:

> # Homebrew (Legacy)
> 
> This repository was deprecated and split into two repositories:
> 
> - Homebrew formulae/packages: [Homebrew/homebrew-core](https://github.com/Homebrew/homebrew-core) (former contents of `Library/Formula`)
> - Homebrew package manager: [Homebrew/brew](https://github.com/Homebrew/brew)

You suggest that we use the second link but since Melpa is a package archive not a package manager, the first link seems more appropriate.

By the way, I think Melpa should split the recipes into a separate repository too.

Sasanidas:
Ok, no problem then, request closed.

PR #6876, opened 2020-04-30
Add auth-source-xoauth2

ccrusius:

Adds an OAUTH2 backend to auth-sources, enabling it to be used with services such as GMail.

https://github.com/ccrusius/auth-source-xoauth2

Author and user.

**None needed**

### Checklist

Please confirm with `x`:

- [ ] My elisp byte-compiles cleanly: There is a free variable warning that I'm not sure can be removed without unnecessarily (requre 'nnimap), which would bring all of Gnus with it.

### DIFF
--- a/recipes/auth-source-xoauth2
+++ b/recipes/auth-source-xoauth2
@@ -1,0 +1,3 @@
+(auth-source-xoauth2
+ :fetcher github
+ :repo "ccrusius/auth-source-xoauth2")


riscy:
Quick first pass --

### auth-source-xoauth2.el ###

```
auth-source-xoauth2.el:181:30:Warning: reference to free variable
    `nnimap-authenticator'
```
For this, since it looks like you're deliberately putting off loading nnimap, you could always `(defvar nnimap-authenticator)` I believe.

```
Warning (emacs): 
auth-source-xoauth2.el:107: Symbols t and nil should not appear in single quotes
```

```
12 issues found:
38:10: error: You should depend on (emacs "25.1") or the seq package if you need `seq'.
40:10: error: You should depend on (emacs "26.1") if you need `auth-source-pass'.
130:3: error: You should depend on (emacs "25.1") if you need `when-let'.
138:5: error: You should depend on (emacs "25.1") if you need `when-let'.
142:7: error: You should depend on (emacs "25.1") if you need `when-let'.
193:5: error: You should depend on (emacs "25.1") or the cl-lib package if you need `cl-defmethod'.
224:3: error: You should depend on (emacs "25.1") if you need `when-let'.
249:53: error: You should depend on (emacs "26.1") if you need `auth-source-pass--find-match'.
257:6: error: You should depend on (emacs "26.1") if you need `auth-source-pass--find-match'.
271:14: error: You should depend on (emacs "26.1") if you need `auth-source-pass-get'.
287:3: error: You should depend on (emacs "25.1") if you need `when-let'.
288:5: error: You should depend on (emacs "25.1") if you need `when-let'.
```
I think these are all good to look at.

- auth-source-xoauth2.el#L180: It's safer to sharp-quote function names; use `#'`
- auth-source-xoauth2.el#L198: It's safer to sharp-quote function names; use `#'`
- auth-source-xoauth2.el#L222: Consider `unless ...` instead of `when (not ...)`
- auth-source-xoauth2.el#L9: Prefer `https` over `http` (if possible)

ccrusius:
Thank you Chris for the review - I fixed the problems except for the lint asking me to require "26.1" for auth-source-pass: that package is available from Melpa for Emacs 25.x and below.

riscy:
Hi!  I think two changes are needed:

- `(require 'auth-source-pass)` instead of `(require 'auth-source)`
- Update your package-requires to `((emacs "25.1") (auth-source-pass "5.0.0"))` (or, of course,  depend on Emacs 26.1)

Other than that I see no blockers, so this is ready for a quick merge after the above fixes.

One last comment is that your wrapper (`auth-source-xoauth2-pass--find-match`) is bound to be fragile, since the function it wraps is not part of the public interface of auth-source-pass.  One defense against this might be to raise an issue in the auth-source-pass and see if the wrapped function can be made public.

ccrusius:
Thanks for the review!

I tried the changes but the linter still complains. It does not seem to like the transfer of `auth-source-pass` from MELPA to Emacs. But there's also the problem that `auth-source-pass` is really optional for this package, so I actually do not want to require it. The `(require 'auth-source)` at the top is just for the base `auth-source` functionality. Let me know if you have any suggestions.

Good point about `...-find-match`. I'll raise it with the `auth-source-pass` people.

riscy:
Yeah, I think package-lint will complain about it under the assumption that it's coming from Emacs.  Let's disregard those.

> auth-source-pass is really optional for this package, so I actually do not want to require it. The (require 'auth-source) at the top is just for the base auth-source functionality.

Ah I see.  I think `auth-source` is not necessary for this package, too!  At least, I'm not seeing anywhere that its functions or variables are required, but it's a completely harmless import in any case.

I think auth-source-pass is necessary for the macro expansion of `auth-source-xoauth2-pass--find-match` -- and unfortunately the compiler expands macros before it even starts reading forward declarations and before running any `byte-compile-eval-before-compile` forms (so we can't use e.g. declare-function to pacify the compiler).

There's lots of workarounds, but I think the _best_ thing you could do is Package-Requires on Emacs 26.1 and do `(require 'auth-source-pass nil t)`.  Add to your README and/or commentary that the package will work with Emacs 25.1 and even without auth-source-pass installed, but that users may see compilation errors.

ccrusius:
Thanks again for the review!

I implemented the changes you suggested. I left `(require 'auth-source)` there, I need `auth-source-backend` to define `defvar auth-source-xoauth2-backend`.

riscy:
Right, I see it now -- thanks for being accommodating.  And, welcome to MELPA! :)

ccrusius:
Thanks a lot!

PR #6879, opened 2020-05-01
Include images for sly info manual

### LINKED ISSUE #6870: Include images for info docs in SLY recipe
I am not sure if this is appropriate or not, but the info docs for the SLY package include some images that aren't part of the MELPA recipe for the package. Is there a good way to include these as part of the recipe so that installing the package can also install the info docs along with the included images?

aplaice:
Fix #6870.

Include `doc/images/` since they're used in the sly info manual (for instance [here](https://github.com/joaotavora/sly/blob/master/doc/sly.texi#L644)).

(I tested installing the package before and after updating the recipe, and they were indeed absent before and present after.)

The gif files in `doc/animations/` do not seem to be referenced in the info manual, so I'm not including them.

https://github.com/joaotavora/sly/

The issue was mentioned upstream [here](https://github.com/joaotavora/sly/issues/316), though obviously it was out of scope there.

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/sly
+++ b/recipes/sly
@@ -4,4 +4,5 @@
              "lib"
              "slynk"
              "contrib"
+             "doc/images"
              (:exclude "sly-autoloads.el")))


riscy:
Based on the issue you linked, I think there's an appetite from the users to have these included -- and none of the images are very large at all.  Another argument in favor of doing this is that, by default, MELPA packages .info files, so they might as well be packaged in a presentable way.

Let's wait a couple days to see if anyone else has any opinions on this, but I'm for it.

PR #6882, opened 2020-05-04
Add org-roam-server

goktug97:

A Web Application to Visualize the [Org-Roam](https://github.com/org-roam/org-roam) Database

https://github.com/org-roam/org-roam-server

Maintainer

**None needed**

### Checklist

Please confirm with `x`:

EDIT: Change direct github link because it is now part of org-roam

### DIFF
--- a/recipes/org-roam-graph-server
+++ b/recipes/org-roam-graph-server
@@ -1,0 +1,4 @@
+(org-roam-graph-server
+ :repo "goktug97/org-roam-graph-server"
+ :fetcher github
+ :files ("*.el" "index.html" "assets"))
--- a/recipes/org-roam-graph-server
+++ b/recipes/org-roam-graph-server
@@ -1,4 +1,0 @@
-(org-roam-graph-server
- :repo "goktug97/org-roam-graph-server"
- :fetcher github
- :files ("*.el" "index.html" "assets"))
--- a/recipes/org-roam-server
+++ b/recipes/org-roam-server
@@ -1,0 +1,4 @@
+(org-roam-server
+ :repo "goktug97/org-roam-server"
+ :fetcher github
+ :files ("*.el" "index.html" "assets"))
--- a/recipes/org-roam-server
+++ b/recipes/org-roam-server
@@ -1,4 +1,4 @@
 (org-roam-server
- :repo "goktug97/org-roam-server"
+ :repo "org-roam/org-roam-server"
  :fetcher github
  :files ("*.el" "index.html" "assets"))


riscy:
Hi!  Thanks for this.  Quick first pass --

### org-roam-server.el ###

```
org-roam-server.el:169:1:Warning: Unused lexical argument `path'
In org-roam-server-export-server-id:
org-roam-server.el:199:95:Warning: Pattern t is deprecated.  Use `_' instead
org-roam-server.el:206:1:Warning: Unused lexical variable `file-from'
org-roam-server.el:206:1:Warning: Unused lexical variable `bls'
org-roam-server.el:206:1:Warning: Unused lexical argument `file'
In org-roam-server-insert-citelinks:
org-roam-server.el:206:42:Warning: `(ref (with-temp-buffer
    (insert-file-contents file) (org-roam--extract-ref)))' is a malformed
    function
org-roam-server.el:212:58:Warning: reference to free variable `ref'
org-roam-server.el:215:34:Warning: reference to free variable `key-backlinks'
org-roam-server.el:224:28:Warning: reference to free variable
    `grouped-backlinks'
org-roam-server.el:240:1:Warning: Unused lexical variable `file-from'
In org-roam-server-insert-backlinks:
org-roam-server.el:240:42:Warning: `(backlinks (org-roam--get-backlinks
    file))' is a malformed function
org-roam-server.el:246:38:Warning: reference to free variable `backlinks'
org-roam-server.el:248:34:Warning: reference to free variable
    `grouped-backlinks'
In end of data:
org-roam-server.el:286:1:Warning: the following functions are not known to be
    defined: if-let*, org-ref-p, key-backlinks, grouped-backlinks, first
```
These are mostly due to failed macro expansion due to what I think are some missing imports.  for `if-let*`, `(require 'subr-x)`; instead of `first`, prefer the non-deprecated`cl-first` (and `(require 'cl-lib)` or just use `car`.  For `org-ref-p`, you need an explicit require on `'org`.  The "Pattern t is deprecated"

Doing all of the above will bring you down to:
```
org-roam-server.el:169:1:Warning: Unused lexical argument `path'
org-roam-server.el:206:1:Warning: Unused lexical variable `file-from'
org-roam-server.el:206:1:Warning: Unused lexical variable `bls'
org-roam-server.el:240:1:Warning: Unused lexical variable `file-from'
```
If you're not using these variables, you can prefix them as `_variable` to silence the byte-compiler -- however, it seems like in some cases the code makes it "look" like they're intended to be used (possibly a bug?) and in other places they're defined and then redefined shortly after (as is the case with `bls`).

- No issues!

- No issues!

- org-roam-server.el#L33: Prefer `https` over `http` (if possible)

goktug97:
Fixed byte-compile warnings. Also, there was a bug where a loop is repeated twice. I think I put it accidentally while editing.

> Suggestions/experimental static checks:
org-roam-server.el#L33: Prefer https over http (if possible)

https didn't work.

riscy:
Apologies for the https pointer on a localhost url, I didn't double-check.  Everything looks much improved, but this one still could use a fix:

> prefer the non-deprecated cl-first (and (require 'cl-lib) or just use car.

due to byte-compile-file (using Emacs 26.3):
```
org-roam-server.el:283:1:Warning: the function `first' is not known to be
    defined.
```

goktug97:
Sorry, I forgot to fix the second occurrence of the `first`. I don't know why but I don't get it as a warning even though my emacs version is also 26.3. Changed it to `car`, hopefully, everything is fixed now.

riscy:
Looks good to go.  Welcome to MELPA!

PR #6884, opened 2020-05-05
Add ytel recipe

gRastello:

Search for youtube videos from Emacs in an [elfeed](https://github.com/skeeto/elfeed)-like buffer.

https://github.com/grastello/ytel

Maintainer

**None needed**

### Checklist

### DIFF
--- a/recipes/ytel
+++ b/recipes/ytel
@@ -1,0 +1,1 @@
+(ytel :repo "grastello/ytel" :fetcher github)


xuchunyang:
> * [x]  My elisp byte-compiles cleanly

You're using [seq-do](https://github.com/gRastello/ytel/blob/aea205c694ae9aea15d62c3e24a7a1332bc459b9/ytel.el#L142) without loading the seq.el library, thus the byte compiler should give a warning on this. A good idea is byte compiling from Emacs -Q.

> (cl-defstruct ytel-video

This creates functions named make-NAME and copy-NAME, they aren't using the package prefix `ytel-`, you can avoid this with :constructor and :copier, e.g.,

```elisp
(cl-defstruct (person (:constructor person-create)
                      (:copier nil))
  name age)
```

gRastello:
Ok I've fixed those two things. However the byte compiler does not seem to complain about any missing `require`, even when it is run from an `emacs -Q`.

Thank you man!

xuchunyang:
> However the byte compiler does not seem to complain about any missing `require`, even when it is run from an `emacs -Q`.

You're right (at least for some versions of Emacs, I tried Emacs 27), that's because json.el requires map.el since Emacs 25.1, and map.el requires seq.el, thus seq.el is implicitly required.

riscy:
Hi!  Thanks for this -- looks cool.

I notice you're defining a major mode with a `defun`, but there is a helpful (and now idiomatic) macro for this: `define-derived-mode`.  It will handle the setting of `major-mode`, `mode-name`, killing local variables, running hooks, automatically using `ytel-mode-map` as its keymap, and so on.  You should also typically avoid turning on/off some opinionated settings (such as enabling hl-line-mode and truncate-lines, which users may have set to values that already work well for them).

The other thing that strikes me is that this is called `ytel` (which I assume is short for maybe youtube-emacs-lisp), but primarily uses Invidious which from my reading is a YouTube frontend.  Maybe Invidious and YouTube should at least appear together in the summary line -- "Query YouTube via invidio.us"?  I'm also capitalizing the T in Tube per the branding.

Some minor stuff:
- ytel.el#L143: There is no need to quote lambdas (neither #' nor ')
- ividious -> Indivious

gRastello:
Thank you for the stuff you picked up.

I didn't know about `define-derived-mode`; that sure is convenient!

I'm not completely sold on not setting `truncate-lines` and `hl-line-mode` but I realize that's just my personal taste coming through. Users can easily set up a hook to achieve that "look and feel" so in the end I decided to avoid setting them by default.

riscy:
> I'm not completely sold on not setting truncate-lines and hl-line-mode but I realize that's just my personal taste coming through. Users can easily set up a hook to achieve that "look and feel" so in the end I decided to avoid setting them by default.

Now that I think about it some more (sorry for not having done that earlier ;) I may be wrong in suggesting those shouldn't be turned on, especially now that I've gone and seen how the mode operates.  Apologies if that's caused any trouble.

This looks good to go.  Welcome to MELPA! :)

gRastello:
> Now that I think about it some more (sorry for not having done that earlier ;) I may be wrong in 
suggesting those shouldn't be turned on, especially now that I've gone and seen how the mode operates. Apologies if that's caused any trouble.

It's really not a big problem. I'm keeping it like this for the time being. Might switch it back to how it was before if people seem to like it more like that.

Thank you again and have a nice day!

PR #6886, opened 2020-05-06
Migrate digit-groups from Bitbucket to Github

adamsmd:
Migrate my package(s) from Bitbucket to Github

As @tarsius has requested in #6484 I have migrated my packages
from Bitbucket to Github.  The old repositories were located at:

* https://bitbucket.org/adamsmd/digit-groups/


tarsius:
Thank you!

PR #6887, opened 2020-05-06
Migration of my packages from GitHub to https://depp.brause.cc

wasamasa:
I've completed moving my packages to my own Git hosting at https://depp.brause.cc/ and would like to update the recipes to point to the respective new locations.

### DIFF
--- a/recipes/eyebrowse
+++ b/recipes/eyebrowse
@@ -1,1 +1,1 @@
-(eyebrowse :fetcher github :repo "wasamasa/eyebrowse")
+(eyebrowse :fetcher git :url "https://depp.brause.cc/eyebrowse.git")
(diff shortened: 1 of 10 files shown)


CsBigDataHub:
@wasamasa 
Just a request , Can you please not archive more popular repositories like `eyebrowse` in Github.
I cannot access https://depp.brause.cc/ as the URL is blocked by company proxy/VPN. I am not sure if I am the only one facing this issue.

wasamasa:
GitHub doesn't give me the means to disable all user interaction otherwise, so no. I can only speculate why your work would block the URL, could it be it doesn't like the .cc (could be misinterpreted as credit card fraud) TLD?

That aside, I'm sure most of my packages landed on https://github.com/emacsmirror/ already.

PR #6893, opened 2020-05-09
Add org-books recipe

lepisma:

`org-books` assists in managing reading list in org-mode.

https://github.com/lepisma/org-books

Author

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/org-books
+++ b/recipes/org-books
@@ -1,0 +1,1 @@
+(org-books :fetcher github :repo "lepisma/org-books")


riscy:
Thanks for this!  I would strongly recommend adding "with helm" (or similar) to your summary line, due to it not being compatible with other completion managers.

As a very minor comment, I might flip around the `(if (null` on line 268 and remove the `progn` as:
```elisp
(if headers
    (helm :sources (helm-build-sync-source "org-book categories"
                     :candidates (-map (lambda (h) (cons (car h) (marker-position (cdr h)))) headers)
                     :action (lambda (pos)
                               (org-content)
                               (goto-char pos)
                               (let ((level (or (org-current-level) 0)))
                                 (org-books-goto-place)
                                 (insert "\n")
                                 (org-books--insert (+ level 1) title author props)
                                 (save-buffer))))
          :buffer "*helm org-books add*")
  (goto-char (point-max))
  (org-books--insert 1 title author props)
  (save-buffer))
```
and I might even recommending turning that nontrivial lambda into a named private function as a small readability enhancement.  And a last couple things:
- org-books.el#L88: Consider `when ...` instead of `unless (null ...)`
- org-books.el#L29: Prefer `https` over `http` (if possible)

None of these are blockers, however, and this looks like a very cool feature.  Thanks!

lepisma:
thanks for the review @riscy! onto these for next release.

PR #6894, opened 2020-05-10
Added recipe osa

atomontage:

osa.el is an Emacs Lisp to macOS OSA (AppleScript / JavaScript) bridge.

Requires Yamamoto Mitsuharu's Emacs Mac port.
Includes several useful automation scripts.

https://github.com/atomontage/osa

Author.

None needed.

### Checklist

Please confirm with `x`:

- [] I've used the latest version of [package-lint](https://github.com/purcell/package-lint) to check for packaging issues, and addressed its feedback

### DIFF
--- a/recipes/osa
+++ b/recipes/osa
@@ -1,0 +1,1 @@
+(osa :repo "atomontage/osa" :fetcher github :files ("*.el" ("scripts" "scripts/*.js" "scripts/*.applescript")))
--- a/recipes/osa
+++ b/recipes/osa
@@ -1,1 +1,1 @@
-(osa :repo "atomontage/osa" :fetcher github :files ("*.el" ("scripts" "scripts/*.js" "scripts/*.applescript")))
+(osa :repo "atomontage/osa" :fetcher github :files (:defaults ("scripts" "scripts/*.js" "scripts/*.applescript")))


riscy:
Hi, this looks interesting.

You should prefer a recipe like ```(osa :repo "atomontage/osa" :fetcher github :files (:defaults ("scripts" "scripts/*.js" "scripts/*.applescript")))```.  This will prevent you from shipping with your `-test.el` file.  Similarly osa-example.el should probably be excluded (or even better, create a directory called `examples/` in your repo) so that MELPA is only distributing the necessary code.

Also, if you `(require 'mac nil t)` instead of `(require 'mac)` the package will be able to be automatically checked (e.g. byte-compiled) on non-macOS operating systems.

Lastly, the GitHub API is detecting your repo's license as "Other" -- if possible, it's a good idea to use a GitHub-compatible format for your license file (See: https://github.com/licensee/licensee)

Here's a quick pass over the checklist --

### osa.el ###

```
osa.el:290:15:Warning: `string-to-multibyte' is an obsolete function (as of
    26.1); use `decode-coding-string'.
In end of data:
osa.el:335:1:Warning: the function `mac-osa-script' is not known to be
    defined.
```
That last warning you could silence by not sharp-quoting the function or using a `declare-function` form.  Neither of the above are blockers.

- No issues!

```
27 issues found:
6:12: warning: "1.0 - 2020-04-29" is not a valid version. MELPA will handle this, but other archives will not.
54:10: error: You should depend on (emacs "25.1") or the seq package if you need `seq'.
58:10: error: You should depend on (emacs "25.1") or the cl-generic package if you need `cl-generic'.
68:1: error: You should depend on (emacs "25.1") or the cl-lib package if you need `cl-defgeneric'.
84:1: error: You should depend on (emacs "25.1") or the cl-lib package if you need `cl-defgeneric'.
108:1: error: You should depend on (emacs "25.1") or the cl-lib package if you need `cl-defmethod'.
109:1: error: You should depend on (emacs "25.1") or the cl-lib package if you need `cl-defmethod'.
111:1: error: You should depend on (emacs "25.1") or the cl-lib package if you need `cl-defmethod'.
118:1: error: You should depend on (emacs "25.1") or the cl-lib package if you need `cl-defmethod'.
121:1: error: You should depend on (emacs "25.1") or the cl-lib package if you need `cl-defmethod'.
128:1: error: You should depend on (emacs "25.1") or the cl-lib package if you need `cl-defmethod'.
135:1: error: You should depend on (emacs "25.1") or the cl-lib package if you need `cl-defmethod'.
145:1: error: You should depend on (emacs "25.1") or the cl-lib package if you need `cl-defmethod'.
152:1: error: You should depend on (emacs "25.1") or the cl-lib package if you need `cl-defmethod'.
174:1: error: You should depend on (emacs "25.1") or the cl-lib package if you need `cl-defmethod'.
178:1: error: You should depend on (emacs "25.1") or the cl-lib package if you need `cl-defmethod'.
182:1: error: You should depend on (emacs "25.1") or the cl-lib package if you need `cl-defmethod'.
185:1: error: You should depend on (emacs "25.1") or the cl-lib package if you need `cl-defmethod'.
189:1: error: You should depend on (emacs "25.1") or the cl-lib package if you need `cl-defmethod'.
196:1: error: You should depend on (emacs "25.1") or the cl-lib package if you need `cl-defmethod'.
203:1: error: You should depend on (emacs "25.1") or the cl-lib package if you need `cl-defmethod'.
209:1: error: You should depend on (emacs "25.1") or the cl-lib package if you need `cl-defmethod'.
213:1: error: You should depend on (emacs "25.1") or the cl-lib package if you need `cl-defmethod'.
228:1: error: You should depend on (emacs "25.1") or the cl-lib package if you need `cl-defmethod'.
232:1: error: You should depend on (emacs "25.1") or the cl-lib package if you need `cl-defmethod'.
289:14: error: You should depend on (emacs "25.1") or the seq package if you need `seq-contains'.
313:14: error: You should depend on (emacs "25.1") or the seq package if you need `seq-contains'.
```

atomontage:
I updated the recipe and package version metadata. I also changed the require so that it doesn't error. I'd like to keep the license as is, since it makes it clear that the entire project is dual licensed. Github probably expects the stock BSD text to be in the file and nothing else.

Thanks for taking the time to review this.

riscy:
> Github probably expects the stock BSD license text to be in the file and nothing else.

Ok, sounds good.  Thanks for this -- and welcome to MELPA. :)

tarsius:
> Also, if you `(require 'mac nil t)` instead of `(require 'mac)` the package will be able to be automatically checked (e.g. byte-compiled) on non-macOS operating systems.

And on macOS, where does the `mac` library come from then?

atomontage:
It's provided by Yamamoto Mitsuharu's Emacs Mac .

>> Also, if you `(require 'mac nil t)` instead of `(require 'mac)` the package will be able to be automatically checked (e.g. byte-compiled) on non-macOS operating systems.
>
> And on macOS, where does the `mac` library come from then?

tarsius:
> It's provided by Yamamoto Mitsuharu's Emacs Mac .

I am not sure we should distribute packages that depend on a fork of Emacs. (Could this package be added to that fork instead?)

> Also, if you `(require 'mac nil t)` instead of `(require 'mac)` the package will be able to be automatically checked (e.g. byte-compiled) on non-macOS operating systems.

But why would we want to do that? It seems `mac` is the most important dependency and we should avoid implying that it isn't even really required.

atomontage:
>> It's provided by Yamamoto Mitsuharu's Emacs Mac .
>
> I am not sure we should distribute packages that depend on a fork of Emacs. (Could this package be added to that fork instead?)

It is unfortunate that the two ports haven't yet been unified but the fact
remains that Emacs Mac supports advanced features that have been blocked
from NS Emacs for non-technical reasons (parity with Linux is one of them).

I don't see how MELPA blocking Emacs Mac packages is good for the
Emacs community and Emacs in general. Yamamoto regularly contributes
code to core, announces his Emacs Mac updates in emacs-devel and has been
part of the Emacs core development team for many years.

riscy:
> But why would we want to do that? It seems mac is the most important dependency and we should avoid implying that it isn't even really required.

I lumped this in with packages that require a database, or to have some version of Ruby installed, or a specific bit of linting software that requires (say) 200MB of installed prerequisites.  If a package can be automatically checked (byte-compiled, loaded, etc) without a heavy and sometimes extremely precise setup, I think that's a good thing.  For example it is more likely to be automatically checked for something as trivial as syntax errors.

But I agree with you that `mac` is the most important dependency and should probably throw something.  A compromise might be something like:

```elisp
(eval-and-compile
  (unless (require 'mac nil t)
    (warn "osa.el only works with Yamamoto Mitsuharu's macOS Emacs.
Please see: xyz...")))
```

But I'm definitely open to more ideas here...

Another possible concern is that this breaks the dependency tree, since `mac` isn't on any package manager.  But I don't think we can realistically enforce, since the various Emacs package managers operate more or less independently.

tarsius:
> I lumped this in with packages that require a database, or to have some version of Ruby installed, or a specific bit of linting software that requires (say) 200MB of installed prerequisites.

Makes sense.

PR #6895, opened 2020-05-10
Add a recipe for literate-calc-mode

sulami:

Major/minor mode for inline `calc` results.

https://github.com/sulami/literate-calc-mode.el

I'm the creator/maintainer.

None needed

### Checklist

Please confirm with `x`:

### DIFF
--- a/recipes/literate-calc-mode
+++ b/recipes/literate-calc-mode
@@ -1,0 +1,1 @@
+(literate-calc-mode :fetcher github :repo "sulami/literate-calc-mode.el")


riscy:
This is neat, and thanks for the clarifying screenshot.

I have only some small feedback:

- literate-calc-mode.el#L125: Consider `when ...` instead of `unless (null ...)`
- literate-calc-mode.el#L24: Prefer `https` over `http` if possible ([why?](https://news.ycombinator.com/item?id=22933774))

And, more importantly -- even though you have GPL listed in your file (good), we're also striving for a LICENSE file that GitHub can detect (e.g. no markup) if possible (See: https://github.com/licensee/licensee).  Once that's sorted this looks good to merge.

sulami:
Done.

riscy:
Thank you!  I won't block on this, but it seems that GitHub is unable to detect your license (it's detecting "other") which may be because your license also includes your copyright statement.  If you reuse something [very standard (such as the original text)](https://github.com/riscy/shx-for-emacs/blob/master/COPYING) then it can be detected by common tools.  Have a look at fixing that when you can.

And, welcome to MELPA! :)

